Issue #153, opened 2014-12-24
udp transport layer


Stasik0:
i looked a bit into udp and it seems not as trivial as changing a flag :) we need to cnage the server logic a bit as e.g. stated here http://www.cs.dartmouth.edu/~campbell/cs60/socketprogramming.html

so is it worth creating a new layer and than maybe to listen on both TCP and UDP?

PR #155, opened 2015-01-04
Udp transport

Stasik0:
I'd like to merge it in as long as it is not too far apart from master, relates #153

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -441,8 +441,6 @@ void UA_Server_processBinaryMessage(UA_Server *server, UA_Connection *connection
                 if(connection->state == UA_CONNECTION_OPENING &&
                 		connection->channel == UA_NULL) {
                 	processMessage(connection, server, msg, &pos);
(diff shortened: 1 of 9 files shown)


src/server/ua_server_binary.c
```diff
@@ -54,6 +54,7 @@ static void processHEL(UA_Connection *connection, const UA_ByteString *msg, UA_U
     UA_TcpMessageHeader_encodeBinary(&ackHeader, &ack_msg, &tmpPos);
     UA_TcpAcknowledgeMessage_encodeBinary(&ackMessage, &ack_msg, &tmpPos);
     UA_ByteStringArray answer_buf = { .stringsSize = 1, .strings = &ack_msg };
+    // the string is freed internall in the (asynchronous) write
```
jpfr: The string does not have to be freed at all.
UA_alloca just decreases (the stack grows from the top) the esp register and returns a pointer.
When the function returns, esp is reset to the value it had before the function was called.

This works for the ACK message because its size is small enough.
We would have to change this asynchronous writes...

PR #220, opened 2015-04-16
Rewrite of the tcp network layer. It became a lot easier.

jpfr:
plz review.


Stasik0:
todo: fix udp in server and client
todo: fix readme
@jpfr you okay with that?

PR #223, opened 2015-04-17
Enabling External NodeStores for path translation

LEvertz:
Up to now the service translateBrowsePathToNodeId was always called on the internal nodeStore. When only external nodeStores were present this resulted in a NULL-Pointer deref (the ns-Pointer in the server-strucutre is not set as there is no internal nodeStore).
If the starting point for a path translation lies within an external namespace, it makes sense to call this namespace's translate function. Therefore, I added the translateBrowsePathToNodeId function to the external nodeStore function pointer table and implemented the check for external namespaces for the path starting points, calling the external function when necessarry. The looping and checks are analog to the other services.


LEvertz:
Produces an error while testing with an external namespaces, so closed the request.
Turned out the error was on the test side. Reopened.

PR #237, opened 2015-04-30
add a hint which file is appended

HolgerJeromin:
This makes debugging with the amalgamated files much easier.

### DIFF
--- a/tools/amalgamate.py
+++ b/tools/amalgamate.py
@@ -75,7 +75,7 @@ else:
 for fname in args.inputs:
     if not "util.h" in fname:
         with open(fname) as infile:
-            file.write("/*********************************** amalgamated original file \"" + infile + "\" ***********************************/\n")
+            file.write("/*********************************** amalgamated original file \"" + fname + "\" ***********************************/\n")
             for line in infile:
                 inc_res = include_re.match(line)
                 guard_res = guard_re.match(line)
--- a/tools/amalgamate.py
+++ b/tools/amalgamate.py
@@ -75,6 +75,7 @@ else:
 for fname in args.inputs:
     if not "util.h" in fname:
         with open(fname) as infile:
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
We should add an info about the version, too. Debugging an outdated file is no real fun :-)

The current time is probably the easiest to get (but does not give us a [reproducible build](http://martinfowler.com/bliki/ReproducibleBuild.html)), but an git version info would be much better. open62541_VERSION_MAJOR is another nice candidate, but is not in use right now.

Stasik0:
@HolgerJeromin it is actually used in the server itself, just not set yet https://github.com/acplt/open62541/blob/265bf9cdd3301ad7140923eb5fee949c0a188895/src/server/ua_server.c#L127
cant one just run "git rev-parse HEAD" from python script to get a unique commit hash? Of course this call need to be safe if git is not installed etc.

The plan: Build time -  OPEN62541_VERSION_MAJOR "." OPEN62541_VERSION_MINOR "." OPEN62541_VERSION_PATCH - "git rev-parse HEAD"

okay? could you please implement it?

HolgerJeromin:
Sorry, my comment was not clear. I meant open62541_VERSION_MAJOR is not maintained right now. My python and git skills are not good enough for this to change.

uleon:
I found a quite nice piece of code at http://stackoverflow.com/questions/1704907/how-can-i-get-my-c-code-to-automatically-print-out-its-git-version-hash. In particular I like the approach
```
GIT_VERSION := $(shell git describe --abbrev=4 --dirty --always --tags)
CFLAGS += -DVERSION=\"$(GIT_VERSION)\"
```

Stasik0:
gratz @uleon, i worked it in 419c05c73f33bc5f10df62cc6bc3bac746d3bd66, will merge the req now

uleon:
@Stasik0 : :+1: cmake hero :+1:

PR #244, opened 2015-05-05
Improve compatibility for old / uclibc systems endian conversion

LEvertz:
if there are no appropriate functions defined and the endianness does not fit from the start an error is thrown.


Stasik0:
merged and squashed via 085f5eea88ddd35c6cb529e6d686c0200307dcb8

PR #247, opened 2015-05-06
gcc version check for pragma diagnostic

LEvertz:
the diagnostic push / pop pragma is new since gcc version 4.6
this pull adds a test for the gcc version to avoid errors in older versions

### DIFF
--- a/examples/logger_stdout.c
+++ b/examples/logger_stdout.c
@@ -18,7 +18,7 @@ static void print_time(void) {
 const char *LogLevelNames[6] = {"trace", "debug", "info", "warning", "error", "fatal"};
 const char *LogCategoryNames[4] = {"communication", "server", "client", "userland"};
 
-#if ((__GNUC__ >= 4 && __GNUC_MINOR__ >= 6) || defined(__clang__))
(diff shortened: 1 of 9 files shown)


examples/logger_stdout.c
```diff
@@ -18,7 +18,7 @@ static void print_time(void) {
 const char *LogLevelNames[6] = {"trace", "debug", "info", "warning", "error", "fatal"};
 const char *LogCategoryNames[4] = {"communication", "server", "client", "userland"};
 
-#if defined(__GNUC__) || defined(__clang__)
+#if ((__GNUC__ >= 4 && __GNUC_MINOR__ >= 6) || defined(__clang__))
```
HolgerJeromin: This would not match for example gcc 5.3

LEvertz: Right, my bad.

I’ll fix it and recommit

Stasik0:
Is it tested for an older gcc? I suspect that we need to turn "treat warnings as errors" off if we cannot use pragma statements.

Stasik0:
I meant the -Werror flag

LEvertz:
It seemed to work anyway on my gcc (4.5.3)

(at first with out-commented pragmas)

Stasik0:
@LEvertz i have pulled and squashed your commits, please synch your branch

PR #252, opened 2015-05-21
splitting server_run s.t. the server can run in one thread with other applications

Stasik0:
Hey guys,

I have separated UA_Server_run into three parts s.t. we can run the server without an endless loop as proposed by @LEvertz.

Is it okay for everyone? I'd also merge it into v0.1 since the interface is only extended and not changed.

Best Regards
Sten

### DIFF
--- a/include/ua_server.h
+++ b/include/ua_server.h
@@ -74,6 +74,20 @@ UA_Logger UA_EXPORT UA_Server_getLogger(UA_Server *server);
 UA_StatusCode UA_EXPORT UA_Server_run(UA_Server *server, UA_UInt16 nThreads, UA_Boolean *running);
 
 /**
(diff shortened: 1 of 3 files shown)


jpfr:
top!

Stasik0:
@Jpfr:  is it correct that the nl will sleep in select? Do we need a boolean parameter in nl's getWork if we do not want it to sleep?

jpfr:
The last networklayer sleeps with the timeout required to precisely hit the next timed work (or a max timeout). All other networklayers only collect what they already have and don't sleep.
```c
            if(i == server->networkLayersSize-1)
                workSize = nl->getWork(nl->nlHandle, &work, timeout);
            else
                workSize = nl->getWork(nl->nlHandle, &work, 0);
```

PR #257, opened 2015-06-03
Fix Visual Studio build

stephanhensel:
Pull request as discussed in #256

### DIFF
--- a/examples/server.c
+++ b/examples/server.c
@@ -8,7 +8,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h> //access
+#include <io.h> //access
 
 #define __USE_XOPEN2K
 #ifdef UA_MULTITHREADING
--- a/include/ua_config.h.in
+++ b/include/ua_config.h.in
@@ -29,6 +29,13 @@
 #  endif
 #endif
 
+/* Visual Studio needs __restrict */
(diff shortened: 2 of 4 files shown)


Stasik0:
@stephanhensel: did it work for MSVS? as you see the build breaks for GCC. I cannot edit your PR easily, could you please add this to your repository? 
```
#ifdef _MSC_VER
    #include <io.h> //access
#else
    #include <unistd.h> //access
#endif
```
Then I will merge it in.

Stasik0:
@stephanhensel 
I would also advice you to edit ua_config.h.in by adding
```
#if defined(_MSC_VER) && _MSC_VER >= 1400
#   define UA_RESTRICT __restrict
#else
#   define UA_RESTRICT restrict
#endif
```

and use `US_RESTRICT` consistently

jpfr:
Thanks for the pull!

@Stasik0 What about putting it into `src/ua_util.h`?
`ua_config.h.in` contains the macros that are set by cmake.

@stephanhensel 
Do you need to replace unistd.h, a standard POSIX library, by the windows-specific io.h?

Stasik0:
Yea, np.

The access function is actually not relevant for windows, so we could also comment it out completely

stephanhensel:
I changed the ua_config.h.in as Stasik0 told. This works for Visual Studio. Can you now merge it in?

Stasik0:
it still break, i will merge it on my own, wait a second

PR #259, opened 2015-06-07
Generated ns0

Stasik0:
Okay, here is one of the contributions by @ichrispa.

Cmake becomes two options: 
* ENABLE_GENERATE_NAMESPACE0 to enable or disable ns0 generation
* GENERATE_NAMESPACE0_FILE to select the xml file for ns0

The xml templates for ns0 are stored here: https://github.com/acplt/open62541/tree/generated_ns0/tools/schema/namespace0. The Opc.Ua.NodeSet2.Minimal.xml file reflects the currently used set of nodes that is sufficient to pass CTT tests.

The server without generation is about 84k, with generated minimal node set about 130k. One reason is that the pre-defined functions like ```addVariableTypeNode_organized``` are not *yet* used. The aim is to get it down to 84k, then we can always generated the namespace.


Stasik0:
@ichrispa i merged in via a squash-merge https://github.com/acplt/open62541/commit/52a08a9b1942685f56ad4c8827f83911f53283ef

PR #260, opened 2015-06-08
Matlab Mex Macros and Memory Presistence

hfaham:
- Mex compilation check flag 
- Matlab Memory macros
- Memory presistence


Stasik0:
perfect!
@hfaham just wondering, no includes are needed? like #include "mex.h" or something? Who defines mexMakeMemoryPersistent? Or is it in the toolchain magic?

PR #261, opened 2015-06-08
fixed memleaks in external namespace handling for browse

Stasik0:


Stasik0:
@LEvertz i am good with it. Anyone else has any objectives?

PR #268, opened 2015-06-10
python 3 fixes for generate_datatypes.py and amalgamate.py

HolgerJeromin:
The output seems to be the same with python 3.


Stasik0:
it does not compile here on 2.7.4

PR #269, opened 2015-06-10
The use of ${PYTHON_EXECUTABLE} is more generic.

bilach:
It is possible that python is not included in the PATH variable.


bilach:
Sorry, I didn't see that you already discussed this solution. So please ignore this pull request.

HolgerJeromin:
Yes, I experimented a lot making python 2 and 3 happy and making the build more robust. 
But thanks anyway for your work.

PR #277, opened 2015-06-15
removing init from the decodeBinary -> brings up to 10% of performance

Stasik0:
the cleanup needs to trace the position of the last allocation + calloc in array decoding
no valgrind varnings or memleaks, tests do pass

however, I am still not sure... @jpfr review needed

### DIFF
--- a/src/ua_types.c
+++ b/src/ua_types.c
@@ -1117,11 +1117,18 @@ UA_StatusCode UA_copy(const void *src, void *dst, const UA_DataType *dataType) {
 }
 
 void UA_deleteMembers(void *p, const UA_DataType *dataType) {
+    UA_deleteMembersUntil(p, dataType, -1);
+}
+
+void UA_deleteMembersUntil(void *p, const UA_DataType *dataType, UA_Int32 lastMember) {
     uintptr_t ptr = (uintptr_t)p;
     if(dataType->fixedSize)
         return;
     UA_Byte membersSize = dataType->membersSize;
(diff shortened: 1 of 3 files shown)


Stasik0:
this and other optimizations bring already about 10% performance increase when compared to v0.1

100000 iterations took 4451.838000 ms - this branch
100000 iterations took 4737.178000 ms - master
100000 iterations took 5172.686000 ms - 0.1

src/ua_types.c
```diff
@@ -1117,11 +1117,18 @@ UA_StatusCode UA_copy(const void *src, void *dst, const UA_DataType *dataType) {
...
+    UA_deleteMembersUntil(p, dataType, 0);
+}
+
+void UA_deleteMembersUntil(void *p, const UA_DataType *dataType, size_t maxOffset) {
     uintptr_t ptr = (uintptr_t)p;
     if(dataType->fixedSize)
         return;
     UA_Byte membersSize = dataType->membersSize;
     for(size_t i=0;i<membersSize; i++) {
+        if(maxOffset >0 && ptr-(uintptr_t)p >=maxOffset){
```
jpfr: You compare `uintptr_t` with a `size_t`.
Imho we could make offset an `uintptr_t` for semantics.
Or we treat the offset as the member until which we want to delete.
`if(i > maxOffset) return;`

Stasik0:
@jpfr i changed to member number... how is it now?

PR #287, opened 2015-07-01
simplfy closing connections in the network layer

jpfr:
sockets are only closed in the main loop. if a worker wants to close a connection, the socket is only shutdown, not closed.
this is then picked up by the select-call in the next main loop iteration.
when a connection is closed, two jobs are returned. one to immediately detach the securechannel and a delayed job that eventually frees the connection memory.

### DIFF
--- a/examples/networklayer_tcp.c
+++ b/examples/networklayer_tcp.c
@@ -57,21 +57,30 @@ static UA_StatusCode socket_write(UA_Connection *connection, UA_ByteString *buf,
     return UA_STATUSCODE_GOOD;
 }
 
+static void socket_close(UA_Connection *connection) {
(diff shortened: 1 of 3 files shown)


Stasik0:
@LEvertz, how does it feel like?

HolgerJeromin:
Refactoring the library is not easy. :-) 
It's a moving target. *looking at the worker code* 
But go ahead, I am back on the office on Monday.

LEvertz:
Seems ok to me - reminds me of the "old" structure I used in our binding - so yes, interface is well adaptable. Simplifies our nl as well

Stasik0:
okay, i will pull in then

PR #333, opened 2015-08-17
Update ua_subscription.c

StalderT:
fix mem leak


StalderT:
fix mem leak in ua_subscription.c

Issue #348, opened 2015-08-27
Memory leak connect/new client or SIGSEGV


tpiat:
I have next code
If I use 
			UA_Client_disconnect(node.client);
			UA_Client_delete(node.client);
I receive SIGSEGV, if I'll comment <B>UA_Client_disconnect(node.client);</B>, I'll get memory leak.

Even if I don't use UA_Client_delete/UA_Client_new, I get memory leak.

Windows 7, GCC.

Probably I wrote incorrect program. 
Code is below.

```
#include "open62541.h"
#include <string>

using namespace std;

struct OPCUA_Nodes
{
	string ip;
	UA_Client *client;
	int client_connected;
	UA_ReadRequest *rReq;
};

OPCUA_Nodes node;

void core_Logger(UA_LogLevel level, UA_LogCategory category, const char *msg, ...)
{	
}

int localget()
{
	string uri("opc.tcp://");
	uri+=node.ip;

	if(!node.client_connected)
	{		
		UA_StatusCode retval = UA_Client_connect(node.client, ClientNetworkLayerTCP_connect,(char *)uri.c_str());
		if(retval != UA_STATUSCODE_GOOD)
		{
			UA_Client_disconnect(node.client);
			UA_Client_delete(node.client);
			node.client = UA_Client_new(UA_ClientConfig_standard, core_Logger);
			node.client_connected=0;
			return retval;
		}
		else node.client_connected=1;
	}

	UA_ReadResponse rResp = UA_Client_read(node.client, node.rReq);
	if(rResp.responseHeader.serviceResult != UA_STATUSCODE_GOOD)
	{
		node.client_connected=0;		
	}
	UA_ReadResponse_deleteMembers(&rResp);

	return 0;
}

void mainLoop()
{
	while (1)
	{
		localget();
//		sleepMicro(1000000); // Sleep 1s
	}
}

int main(int argc, char* argv[])
{
	node.ip="127.0.0.1:16664";
	node.client_connected=0;
	node.client = UA_Client_new(UA_ClientConfig_standard, core_Logger);
	node.rReq = new UA_ReadRequest;
	UA_ReadRequest_init(node.rReq);
	node.rReq->nodesToRead = (UA_ReadValueId *)UA_Array_new(&UA_TYPES[UA_TYPES_READVALUEID], 1);
	node.rReq->nodesToReadSize = 1;
[...]

PR #349, opened 2015-08-27
Update ua_types.c

tpiat:
Without this line must be SIGSEGV


Stasik0:
no we will not merge it, see comment #348

PR #353, opened 2015-08-28
OS X compilation

ngoossens:
* OS X specific code
  * Updated `networklayer_tcp.c` to use SO_NOSIGPIPE on OS X as MSG_NOSIGNAL is not available
  * Added include of `<machine/endian.h>` rather than `<endian.h>` on OS X
  * Added workaround for missing `clock_gettime` on OS X
* Updated CMakeList.txt
  * Added checks for compiler flag differences
  * Updated `exported_headers` to compile `client_stateless.c`
  * Added urcu libraries to samples where needed
  * Added amalgamation dependency to samples where needed
  * Removed rt library dependency for OS X
* Python and Lua
  * Updated CMakeList.txt to reference updated output name
  * Updated CMakeList.txt for Python and Lua library dependency
  * Updated `bindings/open62541.i` to fix init warnings
  * Cast `UA_INT64_(MAX/MIN)` and `UA_UINT64_(MAX/MIN)` to export with correct precision
  * Changed `UA_NodeIdType` to be external enum for exporting
  * Updated UA_StatusCode type handling functions
* Stateless Client
  * Updated to remove dependency on `_calcSizeBinary`
* UPD
  * Updated `examples/networklayer_udp.c` and `examples/server_udp.c` to use updated code
* Tests
  * Updated tests to `ck_assert_ptr_eq` instead of `use ck_assert_int_eq` where appropriate
  * Updated localization test to check for correct locale when using generated namespaces
* Namespace Generation
  * Updated `tools/pyUANamespace/ua_node_types.py` to ensure no duplicate UA_Server_addNode entries occur


src/server/ua_services_attribute.c
```diff
@@ -411,9 +411,7 @@ void Service_Read(UA_Server *server, UA_Session *session, const UA_ReadRequest *
 		size_t offset = 0;
         UA_Connection *c = UA_NULL;
         UA_SecureChannel *sc = session->channel;
-        if(sc)
```
Stasik0: are you sure?

Stasik0:
Woooow! Great job, @ngoossens.

Unfortunately I am not happy with the changes in exported and internal headers in CMakeLists. May I ask you to restore the current state?

I know, it will break the stateless_client compilation with the amalgate. However this is only a test client for an unofficial feature that we used for a paper. Therefore, it is even not expected to compile with amalgamated files (as you noticed, it breaks the userspace API and digs into the "hidden" server API). I just do not want our users to rely on "new" visible API for the further releases.

Besides that, I left one comment.

CMakeLists.txt
```diff
@@ -48,37 +48,48 @@ if(CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID}" STREQUAL "xClang")
...
+        set(CMAKE_MACOSX_RPATH 1)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_DARWIN_C_SOURCE=1")
+    endif()
 endif()
 
 # build the main library
 set(exported_headers ${PROJECT_BINARY_DIR}/src_generated/ua_config.h
                      ${PROJECT_SOURCE_DIR}/include/ua_statuscodes.h
                      ${PROJECT_SOURCE_DIR}/include/ua_types.h
+                     ${PROJECT_SOURCE_DIR}/src/ua_util.h
```
Stasik0: these are the problematic lines

src/server/ua_services_attribute.c
ngoossens: No, I was a little unsure about this one. I couldn't find any reference to `sc` off session in the history so it looked a bit out of place. I probably should have left it as is but was trying to get it all compiling. Happy to revert, unless you can offer some guidance?

CMakeLists.txt
ngoossens: Hi @Stasik0,

Glad you are happy :) No problem, I will revert the headers and add a condition to remove the stateless client from the build when ENABLE_AMALGAMATION=ON.

src/server/ua_services_attribute.c
Stasik0: The problem is that I am not sure myself ;) Can you just, revert this line and mark it with "TODO" or "FIXME" and/or open an issue

CMakeLists.txt
Stasik0: Yea

examples/networklayer_udp.c
```diff
@@ -71,41 +68,44 @@ typedef struct {
 } ServerNetworkLayerUDP;
 
 /** Accesses only the sockfd in the handle. Can be run from parallel threads. */
-static void writeCallbackUDP(UDPConnection *handle, UA_ByteStringArray gather_buf) {
-	UA_UInt32 total_len = 0, nWritten = 0;
-	struct iovec iov[gather_buf.stringsSize];
-	for(UA_UInt32 i=0;i<gather_buf.stringsSize;i++) {
```
jpfr: Are you sure you want to delete this for-loop?

ngoossens: I am new to the codebase so was following history in a number of these changes. Happy for any insight you can provide.

In this case gather_buf  is a UA_ByteStringArray which is no longer in the codebase. When it was it was an array of UA_ByteString with a length of stringsSize, seeing as we can now only pass a single UA_ByteString to the method the for-loop seemed unnecessary.

CMakeLists.txt
ngoossens: Reverted.

src/server/ua_services_attribute.c
ngoossens: Reverted, FIXME added and issue https://github.com/acplt/open62541/issues/354 opened.

examples/client_stateless.c
```diff
@@ -81,28 +84,22 @@ int main(int argc , char *argv[])
...
-			UA_SequenceHeader_calcSizeBinary(&reqSequenceHeader)+
-			UA_NodeId_calcSizeBinary(&reqRequestType) +
-			UA_ReadRequest_calcSizeBinary(&req);
-
 	UA_TcpMessageHeader_init(&reqTcpHeader);
 	reqTcpHeader.messageTypeAndFinal = UA_MESSAGETYPEANDFINAL_MSGF;
-	reqTcpHeader.messageSize = messageEncodedLength;
 
-	UA_TcpMessageHeader_encodeBinary(&reqTcpHeader, &message, &messagepos);
+	messagepos = 8;
```
jpfr: nice!

jpfr:
There are many good optimizations in this pull request. Thank you, @ngoossens!

But the PR should be split into separate topics and merged individually. The smaller the PR, the faster they get reviewed+merged.

- Apple fixes separately
- Unit test updates separately
- UDP network layer separately
- All other fixes

Also, you introduce UA_NO_AMALGAMATION. That is not needed in most places. The amalgamation process uses a regex to filter out "internal" includes that are not needed in the amalgamated release. The reduces the ifdef-hell quite  a bit. By the way, I don't see UA_NO_AMALGAMATION defined in the cmake files.

jpfr:
A quick way for selective commits is to

- create a new branch from master
- `git checkout <dev-branch> <filename>` or cherry-pick the relevant commits
- selectively add the chunks in the file with `git add -p <filename>
- push the new branch and open a PR

ngoossens:
@jpfr, no problem happy to help. Thanks for the feedback. You are absolutely right this should be broken up, I started out just trying to get it building on OS X and ended up going down the rabbit hole.

I will be afk soon but will be sure to re-submit as smaller PRs by Monday.

ngoossens:
Closing in favour of a set of smaller PRs as per @jpfr suggestion.

PR #357, opened 2015-08-31
OS X compilation

ngoossens:
* OS X specific code
  * Updated `networklayer_tcp.c` to use SO_NOSIGPIPE on OS X as MSG_NOSIGNAL is not available
  * Added include of `<machine/endian.h>` rather than `<endian.h>` on OS X
  * Added workaround for missing `clock_gettime` on OS X
* Updated CMakeList.txt
  * Added checks for compiler flag differences
  * Added urcu libraries to samples where needed
  * Removed rt library dependency for OS X


Stasik0:
:+1: from me, even if i have no osx to test

ngoossens:
Thanks @Stasik0, I will update the wiki.

PR #358, opened 2015-08-31
Updated  to ensure no duplicate UA_Server_addNode entries occur

ngoossens:
There are two paths that call `self.getFirstParentNode()` (Line #656 & Line #668), these can both generate the same UA_Server_addNode call which can cause a double free issue in `nodeEntryFromNode` in `src/server/ua_nodestore.c`. 

Added an if statement to ensure that the duplicate does not occur.


Stasik0:
@ichrispa didn't you fix it in hi_level_api branch?

ichrispa:
The pull request is valid for the master branch, but  entire code segment in that function has been changed/fixed in the hl-api.

Stasik0:
@ngoossens sorry, Nick, but it was already fixed in hl-api that is going to be merged asap

ngoossens:
@Stasik0, no problem. hl_api looks like quite a big change set so I am keen to test it out.

Issue #337, opened 2015-08-18
0.2 release meta issue


Stasik0:
Hey folks,

i'd like to evaluate how 0.2 can evolve. From my point of view we have more-or-less enough feature to create a new stable release. Afaik these are:

* methods
* subscriptions
* xml-nodeset generation
* new user-api
* MSVS compatibililty
* performance improvements
* documentation and tutorials.

I'd like to have an alpha tag due to our workshop on the 7.9.15. Imho only the api_hl_abstractions branch needs to be reintegrated in order to create an 0.2-alpha. Is it correct, in this case @ichrispa: would you please estimate the deadline or the blockers for it?

Is there anything else we should/can bring 0.2 until it is feature-frozen?

ichrispa:
@Stasik0: The single major blocker right now is the serverside api and services support. Some isolated services like the nodemanagement and attributes are not fully matures (ex: you can create 4 node types and delete none using a client). Also, many of the hl-features use the server's nodestore instead of reusing the service apis. The two major blockers I currently see are:

  * Nodemanagement Service does allow creation of all node types
  * Nodemanagement Service can not delete nodes yet
  * Attributes Service only allows the value attribute to be written to

These three are majorly counter-intuitiv. I will fix them asap, as I want the hl-branch to be merged into master before they deviate too much.

I would like to add a series of tutorials to the list of 0.2 features-freezes. We are getting many new users and a lot of "how can I do this/that" questions. Many of these questions have answers hidden in  the examples, but a conclusive tutorial will propably be better.

Stasik0:
Okay, so the commitment is to do the merge before 7.9.? So we can present the 0.2-alpha version there? Would you prefer to write tutorials in the doc-system or via a wiki in github, or both?

jpfr:
>  Nodemanagement Service does allow creation of all node types
    Nodemanagement Service can not delete nodes yet
    Attributes Service only allows the value attribute to be written to

We probably won't achieve complete coverage of this for 0.2.
Creation of all node types requires that we handle object modelling rules, i.e. required and optional components and properties (which we don't consider so far).
The same applies to the deletion of objects and *type*-nodes.

Another blocker is that subscriptions (to my knowledge) have not been verfied with the official testing tools and have no unit tests. For the 0.1 release it was a lot of work to get all tests go green.

> Would you prefer to write tutorials in the doc-system or via a wiki in github

Maybe we can switch to the sphinx+doxygen documentation between 0.2-pre and 0.2...
[...]

PR #359, opened 2015-08-31
Api hl abstractions

Stasik0:
should we do it? #337


Stasik0:
imho just merge it in and stabilize, @jpfr @ichrispa what is with the documentation? should we go over to sphinx.

PR #367, opened 2015-09-01
include C runtime into the exe for MSVC

HolgerJeromin:
With this change the files are executable without installing the c runtime enviroment


Stasik0:
@HolgerJeromin can you squash the commits?

PR #368, opened 2015-09-01
move SecureChannel expire check into internal function

HolgerJeromin:
With this change a client could call `UA_Client_renewSecureChannel` for keeping the session open. 
He can not check the expiry himself!


Stasik0:
it is completely non-invasive so I merge in with a good feeling ;)

PR #376, opened 2015-09-03
Travis OS X Build

ngoossens:
Not sure if this is required (or wanted) since there is already a clang build under Linux but thought I would submit for discussion. What do you think?


Stasik0:
@ngoossens Woooow cool stuff! What i see here is that the tests actually failed to compile (due to my change in the api that is already fixed upstream)

https://travis-ci.org/acplt/open62541/jobs/78572727#L2191
https://travis-ci.org/acplt/open62541/jobs/78572728#L2189

The original build of this commit failed:
https://travis-ci.org/acplt/open62541/jobs/78566059#L2354
**The command "make && make test" exited with 2.**

I think the problem is that the scripts do not return the right value, or they do not return the value at all?

I it probably a good idea to test it with your state i.e. not to pull from master

What does this syntax mean?
"if [ ${TRAVIS_OS_NAME:-'linux'} = 'linux' ]; then . ./tools/travis_linux_script.sh; fi"

is it the same?

"if [ ${TRAVIS_OS_NAME} == 'linux' ]; then ./tools/travis_linux_script.sh; fi" ?

Stasik0:
@ngoossens ping?

ngoossens:
Hi @Stasik0,

Sorry this totally slipped my mind. Was digging into some cross compiling for an old (glibc 2.3) system and then went on leave for a short while. Will get back to onto this today.

ngoossens:
Hi @Stasik0,

When you get a chance if you could please review the changes I would appreciated it. I have updated the if syntax to be simpler and added a sha-bang to force bash usage which makes the `set -ev` command correctly exit on failed exit codes. This is as per the Travis CI docs: http://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps

Stasik0:
@ngoossens looks good, did you try to inject an error to verify that the script actually fails?

Stasik0:
@ngoossens okay looks good! Could you remove the last commit - i will pull afterwards! good job

ngoossens:
@Stasik0, thanks. I did have some errors while testing but just to confirm did an additional test which you can see here: https://travis-ci.org/acplt/open62541/builds/80815349

ngoossens:
Commit removed, thanks!

PR #378, opened 2015-09-04
rename get/setNodeAttribute; use inline methods for type safety

jpfr:
Some proposed changes to the server API.

- Use const pointers for setting atttributes
- Rename to UA_Server_getNodeAttribute and similar
- Use static inline functions (with the function body in the header) instead of macros for type safety. Don't use ptrs but const values for specific attribute set functions. That makes clear that values are copied and the user needs to clear the memory on his side.

@ichrispa 
@Stasik0 

What do you say? Is it much work to update the tutorials?


examples/CMakeLists.txt
```diff
@@ -36,8 +36,10 @@ add_custom_command(OUTPUT ${PROJECT_BINARY_DIR}/src_generated/nodeset.h ${PROJEC
                        ${PROJECT_SOURCE_DIR}/tools/pyUANamespace/NodeID_Blacklist_FullNS0.txt
                        ${PROJECT_SOURCE_DIR}/examples/server_nodeset.xml)
                
-add_executable(server_nodeset server_nodeset.c ${PROJECT_BINARY_DIR}/src_generated/nodeset.c)
-target_link_libraries(server_nodeset ${LIBS})
+if(ENABLE_GENERATE_NAMESPACE0)
```
Stasik0: why this one? nodeset uses its own nodest.xml and is not dependend on compiled ns0

Stasik0:
@jpfr make sure you can compile with BUILD_EXAMPLES

Stasik0:
@jpfr are examples compiling? BUILD_EXAMPLES?

jpfr:
@ichrispa 
As discussed, we let the userspace functions use the normal services.
That removes a lot of code duplication. Can you review fe2996b?

jpfr:
@ichrispa I'd like to merge this PR soon.
Can you please review the changed include/ua_server.h?

Further, I propose that we use the AddNodes-service more.
When we add an object or variable node, the service should add the children. And the userspace function makes use of the service.

jpfr:
@ichrispa 
@FlorianPalm 

The server API was redesigned and reviewed together with @Stasik0.
Can you have a look here? https://github.com/acplt/open62541/blob/inline_setattribute/include/ua_server.h

The nodeset generation is working. But might need some polish.
Alle unit tests and the example servers are working.

So this is ready for merging if you agree.

PR #395, opened 2015-09-10
Static inline type handling functions and standard bitfields

jpfr:
- don't macro-generate function definition in ua_type.h
- use static inline functions on integer types (saves 4kb)
- use only (unsigned) int and boolean bitfields (costs 4kb). nonstandard extensions bork various ffi interfaces

### DIFF
--- a/include/ua_statuscodes.h
+++ b/include/ua_statuscodes.h
@@ -1,7 +1,8 @@
 #ifndef OPCUA_STATUSCODES_H_
 #define OPCUA_STATUSCODES_H_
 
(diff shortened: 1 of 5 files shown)


Stasik0:
Cant we add an option for bitfields? Or is it too complex?

jpfr:
We have bitfields now. On more compilers than before.
And we are standard compliant.

Please review. This branch is ready from my side.

Stasik0:
Okay, can you squash? Then I will merge

jpfr:
done

PR #400, opened 2015-09-14
Correct declarations (missing const, ...) and 2 small bugs in ua_server_worker()

joerg555:
- correct def of UA_Variant_decodeBinary() missing *UA_RESTRICT
- no more Bug in void processJobs() jobsSize may be -1
- Timeout calculation in processRepeatedJobs() must be 32 Bit (if 16 bit overflow at 65 s)
- some comments for UA_DateTime variables to check code
- no "#ifdef _MSC_VER" in Subscription_registerUpdateJob thats stuff


jpfr:
Looks good.
Thank you for the PR.

Issue #300, opened 2015-07-15
Use PCG Random Number generator


jpfr:
The latest cygwin patch worked around the differences in the random number generator across operating systems.

We could simply roll our own PCG-based RNG.
http://www.pcg-random.org/download.html

- literally 5 lines of code
- nicer randomness than the builtin RNG
- removes another friction point for porting

Currently, we store the seed in the server struct. For concurrency, we should declare it as thread local storage. So threads don't interfere / cache-congest.

ichrispa:
@jpfr: Would this also replace rand_r from stdlib.h in ua_types.c?

I am trying to compile open62541 for android-17 and the subminimal support of libc is turning out to be a problem...

But I am happy to report that the stack runs great on my Kobo Touch eBook Reader - I don't know what use that will be, but it proves we are in principal still ok on ARM targets :)

Stasik0:
@ichrispa afaik, it will

DuhaNN:
@ichrispa can you help me with android? Do you have wrapper? Thanks

jpfr:
The change was implemented in 85cd217.

PR #402, opened 2015-09-15
Replace random_r with RNG from pcg32

markusgraube:
First step to tackle issue #300
just replacing random_r funtion from stdlib with minimal c
implementation from http://www.pcg-random.org/using-pcg-c-basic.html

No thoughts about thread-safety yet

### DIFF
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -28,11 +28,6 @@ extern "C" {
 #endif
 #include "ua_config.h"
 #include "ua_statuscodes.h"
-  
-/*************************/
-/* External Dependencies */
-/*************************/
-#include "pcg_basic.h"
 
 
 /** A two-state logical value (true or false). */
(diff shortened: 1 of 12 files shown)


jpfr:
/deps/pcg_basic.h and /deps/pcg_basic.c are missing from the commit.

Good job otherwise!

include/ua_types.h
```diff
@@ -355,7 +361,7 @@ void UA_EXPORT UA_Guid_delete(UA_Guid *p);
 static UA_INLINE void UA_Guid_deleteMembers(UA_Guid *p) { }
 static UA_INLINE UA_StatusCode UA_Guid_copy(const UA_Guid *src, UA_Guid *dst) { *dst = *src; return UA_STATUSCODE_GOOD; }
 UA_Boolean UA_EXPORT UA_Guid_equal(const UA_Guid *g1, const UA_Guid *g2);
-UA_Guid UA_EXPORT UA_Guid_random(UA_UInt32 *seed); ///> Do not use for security-critical entropy!
+UA_Guid UA_EXPORT UA_Guid_random(pcg32_random_t* rngptr); ///> Do not use for security-critical entropy!
```
jpfr: This breaks the amalgamated build. pcg_basic.h is currently not included in open62541.h.
We can keep pcg internal without exposing it in the "public headers".
Just use UA_Guid_random(UA_UInt32 *seed) and cast internally.

include/ua_types.h
```diff
@@ -28,6 +28,12 @@ extern "C" {
 #endif
 #include "ua_config.h"
 #include "ua_statuscodes.h"
+  
+/*************************/
+/* External Dependencies */
+/*************************/
+#include "pcg_basic.h"
```
jpfr: Now we don't need this include in the public interface. It can be moved to ua_types.c.

Otherwise, ready to merge!

PR #406, opened 2015-09-16
prevent a segfault in ClientNetworkLayerGetBuffer

HolgerJeromin:
every call to this function tests the return value, so this is probably a good place for the check.

### DIFF
--- a/examples/networklayer_tcp.c
+++ b/examples/networklayer_tcp.c
@@ -450,6 +450,8 @@ UA_ServerNetworkLayer ServerNetworkLayerTCP_new(UA_ConnectionConfig conf, UA_UIn
 
 static UA_StatusCode ClientNetworkLayerGetBuffer(UA_Connection *connection, UA_ByteString *buf) {
 #ifndef UA_MULTITHREADING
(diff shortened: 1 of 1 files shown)


examples/networklayer_tcp.c
```diff
@@ -450,6 +450,8 @@ UA_ServerNetworkLayer ServerNetworkLayerTCP_new(UA_ConnectionConfig conf, UA_UIn
 
 static UA_StatusCode ClientNetworkLayerGetBuffer(UA_Connection *connection, UA_ByteString *buf) {
 #ifndef UA_MULTITHREADING
+    if(connection->state == UA_CONNECTION_CLOSED)
+        return UA_STATUSCODE_BADCONNECTIONCLOSED;
```
Stasik0: is buf kept undefined in this case?

HolgerJeromin: in all our own code there is 
`UA_ByteString message;`
before the call of `getbuffer(&c->connection, &message)`. 
IMO this is the responsibility of the caller.

Stasik0:
okay

PR #410, opened 2015-09-17
Save Handling of Subscriptionlist and Workerjobs

joerg555:
For testing i made 2 connectons to my server with subscriptions in UaExpert and had crashes with invalid pointers.
Please don's use:
 - LIST_FOREACH_SAFE()

Make your own loop an then you can see what is done.
I will try to check the other loops later.
73 Jörg


Stasik0:
@joerg555 hey! well I was trying to understand the changes in the queue-handling but I did not get it, the problem I see is: consider three jobs (a),(b),(c) okay, now after the execution of (a) the order becomes (b),(c),(a). And then you call the LIST_NEXT(a), but it is the last entry in the list, so (b) and (c) are never executed.

What is with the idea to get the new item with LIST_FIRST only as long as `tw->nextTime <= current`, and not use LIST_NEXT at all?

joerg555:
Hi Stasik0
- ## in processRepeatedJobs(UA_Server *server):
processJobs() will change the sort-order of _server->repeatedJobs_
so you have to do _LIST_NEXT(tw, pointers)_ after processJobs().
if you do it before, and you have more than one job, the netxt pointer may be invalid. 

- ### To get control of what you do its good do make your own loop like processRepeatedJobs()

- ## UA_Subscription_deleteMembers()
before there was a _LIST_FOREACH_SAFE()_ .
You kan't see that it takes the _LIST_NEXT()_ at start of the loop.
Because _Subscription_deleteUnpublishedNotification()_ deletes some entries of _subscription->unpublishedNotifications()_ it dos't work because after that LIST_NEXT() - pointer may be invalid.
So the only solution is to say _Subscription_deleteUnpublishedNotification(UA_UInt32 seqNo, UA_Boolean bDeleteAll, UA_Subscription *sub)_ what to do and give one more parameter _bDeleteAll_.
 Because i dont know exact about _seqNo_ we cant use seqNo == 0 for this meaning

- ## I think we have a big problem with memory usage.
I work now for a server implementation in VS2013 / 2015 in C++
its very unsave to do malloc, realloc and free in different parts of the programm.
I Have only 3 entrypoints:
OnAccept()
OnReceive()
OnTimer()
and no simple Socket variable.
I started to make a new network-layer without changing your basic code.
I will push it to my fork when ist's working.

jpfr:
> processJobs() will change the sort-order of server->repeatedJobs
so you have to do LIST_NEXT(tw, pointers) after processJobs().

What @stasik0 meant is to not use list_next at all. If we always take the first entry from the list, it will be the next job to execute. No matter if and where entries are added or removed from within a job.

joerg555:
Sorry, no i have checked it.

I think ist's the right way.
The list is not so long, and with queue-List ist's very fast to walk throu.
My suggestion is to make a new file:
ua_server_repeatedjobs.c / .h
because i need it without the ua_server_worker.c/h.
So you have a nice interface and can see what's going in and out.
regards jörg.

Stasik0:
making a new file is okay, still taking the first entry from the list should be the same speed as the next entry

joerg555:
Hi,
sorry, now i have some more small changes and don't know how to merge because this is outstanding.
can you pull / merge other requests from joerg555 without this one ?

Stasik0:
probably you will have to make an internal branch and cherry-pick the desired changes there, and then to open a pull-request from the branch

Stasik0:
hey @joerg555 , hey @jpfr 

I have added a different approach of making repeated work more safe (i pick always the head of the list now)

https://github.com/acplt/open62541/commit/4af6f485de036cb8424d56fd672dc8770f5b0c6d in branch https://github.com/acplt/open62541/compare/worker

could you please test it?

jpfr:
The changes were pull via https://github.com/acplt/open62541/commit/e58d410a5a3f3f935297d5377e1df8d9e548ea9e

Thanks @joerg555 !

PR #409, opened 2015-09-17
remove multiple calls to WSAGetLastError

reza-ebrahimi:
remove multiple calls to WSAGetLastError()

PR #411, opened 2015-09-17
remove multiple calls to WSAGetLastError()

reza-ebrahimi:
remove multiple calls to WSAGetLastError()


reza-ebrahimi:
@Stasik0 Is it ok?

Stasik0:
well, not really, since you have all the random stuff now also included in it https://github.com/acplt/open62541/pull/411/files... probably you squashed one commit too much.... hmm what should we do?

reza-ebrahimi:
@Stasik0 I think the solution is remove my master repository and fork it again, then apply my changes and commit.

HolgerJeromin:
perhaps it is easier to bring the old #409 in shape :-)

PR #413, opened 2015-09-17
Fixing compiler warning C4244 about data type conversions

reza-ebrahimi:
Fixing compiler warning C4244 about data type conversions


jpfr:
Thank you for the PR.

PR #414, opened 2015-09-18
fixing pass by reference in two functions and replace sprintf call to…

reza-ebrahimi:
fixing pass by reference in two functions and replace sprintf call to its safe version snprintf because of its known buffer size

### DIFF
--- a/examples/networklayer_tcp.c
+++ b/examples/networklayer_tcp.c
@@ -56,7 +56,8 @@ static UA_StatusCode socket_write(UA_Connection *connection, UA_ByteString *buf,
         do {
 #ifdef _WIN32
             n = send((SOCKET)connection->sockfd, (const char*)buf->data, buflen, 0);
-            if(n < 0 && WSAGetLastError() != WSAEINTR && WSAGetLastError() != WSAEWOULDBLOCK){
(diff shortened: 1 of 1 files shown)


jpfr:
I'd leave the function signatures as-is. The pointer also has 64bit.

Using snprintf is cool.

reza-ebrahimi:
@jpfr What's the problem with 64bit variable pointer? the pointer itself is 32bit.

reza-ebrahimi:
I see linker errors in AppVeyor generated output that can't link _snprintf function, it seems to be libcmtd .lib (for debug) library not found by linker, how to solve the problem?

HolgerJeromin:
I think jpfr wanted to say, that there is no memory advantage by using pointers, as the data and its pointer are both 64 bit.

`examples/logger_stdout.c` uses `UA_DateTime_toString`, too.
And perhaps other users of this library. They are forced to adapt their code (without a gain, see above).

reza-ebrahimi:
@HolgerJeromin I have a test code which is simple functions pass by value and pass by reference in C++ and disassemble it, which was there is no difference in optimization mode between two functions, so there is some extra works in pass by reference function disassembled code in debug mode.
I'll remove those changes and commit again.

PR #415, opened 2015-09-18
replace sprintf with its safer version (snprintf) because of known buffer size

reza-ebrahimi:
replace sprintf with its safer version (snprintf) because of known buffer size


FlorianPalm:
I think it is also essential to define macros, which encapsulate the proper printf-function for each platform

PR #428, opened 2015-10-01
Small bugfix and changes to allow C++ Code

joerg555:
- bug in networklayer_tcp.c with UA_LOGLEVEL 100
- move UA_Server_processBinaryMessage() to ua_server.h because ua_server_internal.h is inkompatibel with C++ (may be fixed)
- minor changes an optimations in ua_types.h to be C++ clean
- UA_Server_processBinaryMessage() change "const UA_ByteString *msg" to const and add comment what is done

Now you can check my branch "simplsmsvc" for my VC 2015 projekt but now it's not complete.
if you have python installed just open "JsmOpen62541.sln" and run.


include/ua_server.h
```diff
@@ -508,6 +508,13 @@ typedef struct UA_ServerNetworkLayer {
  */
 void UA_EXPORT UA_Server_addNetworkLayer(UA_Server *server, UA_ServerNetworkLayer *networkLayer);
 
+/**
+ * process binary message you received from socket
+ * dose not modify UA_ByteString, you have to alloc and free it youself.
+ * use of connection->getSendBuffer() and connection->sent() to answer Message
+ */
+void UA_Server_processBinaryMessage(UA_Server *server, UA_Connection *connection, const UA_ByteString *msg);
```
Stasik0: hi @joerg555 here you break our internal api a bit, what is the exact reason you move it out? (you are not exporting the function as far as i see)

joerg555:
I have my own receive-proc:
https://github.com/joerg555/open62541/blob/simplemsvc/simplemsvc/MfcUaServer/UaClientData.cpp

int UaClientData::CheckReceive()
Line 63 - 102

src/server/ua_server_internal.h
```diff
@@ -74,8 +74,6 @@ struct UA_Server {
 #endif
 };
 
-void UA_Server_processBinaryMessage(UA_Server *server, UA_Connection *connection, UA_ByteString *msg);
```
jpfr: @joerg555 
Can we please leave this function in server_internal?
We don't want to expose it to the end user.

Making the bytestring const is good, though.

PR #435, opened 2015-10-08
match member order definition in UA_ExpandedNodeId struct

jgilje:
To use the UA_EXPANDEDNODEID_NUMERIC macro with g++ 4.9, I had to match the order the members were defined in the UA_ExpandedNodeId struct.

I've tested the server-example after changing the macro, it compiles and works fine after the change.

### DIFF
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -409,8 +409,8 @@ UA_StatusCode UA_EXPORT UA_ExpandedNodeId_copy(const UA_ExpandedNodeId *src, UA_
 UA_Boolean UA_EXPORT UA_ExpandedNodeId_isNull(const UA_ExpandedNodeId *p);
 #define UA_EXPANDEDNODEID_NUMERIC(NSID, NUMERICID) (UA_ExpandedNodeId) {            \
         .nodeId = {.namespaceIndex = NSID, .identifierType = UA_NODEIDTYPE_NUMERIC, \
-                   .identifier.numeric = NUMERICID },                               \
-        .serverIndex = 0, .namespaceUri = {.data = (UA_Byte*)0, .length = -1} }
+                   .identifier = {.numeric = NUMERICID }},                          \
+        .namespaceUri = {.length = -1, .data = (UA_Byte*)0}, .serverIndex = 0 }
 
 /* StatusCode */
 UA_StatusCode UA_EXPORT * UA_StatusCode_new(void);


Stasik0:
okay, seems not to endanger anything

joerg555:
I think its right, because i allways have problems with C++ Code in VS2015 and have to change it in the following way:
```C++
#define UA_STRING(CHARS)                            {(UA_Int32) strlen(CHARS), (UA_Byte*)CHARS }
#define UA_QUALIFIEDNAME(NSID, CHARS)               { NSID, {(UA_Int32) strlen(CHARS), (UA_Byte*)CHARS } }
#define UA_LOCALIZEDTEXT(LOCALE, TEXT)              { UA_STRING(LOCALE), UA_STRING(TEXT) }
#define UA_EXPANDEDNODEID_NUMERIC(NSID, NUMERICID)  { { NSID,  UA_NODEIDTYPE_NUMERIC, NUMERICID }, { -1, NULL}, 0 }

**Comment by Stasik0** [comment]
@joerg555 does it work without the changes now?

**Comment by jgilje** [comment]
@Stasik0 The UA_STRING change mentioned by @joerg555 is because strlen() returnes size_t, which is an unsigned int. g++ gives me a warning about this, because the length member of UA_String is a signed int. Explicitly casting to UA_Int32 gets rid of that warning.

**Comment by Stasik0** [comment]
@jgilje we need signed int, since we descriminate between string of length (length=0) and no string at all (length-1)... I wonder why gcc is not waning us. Hmm should we add it to the code?

**Comment by joerg555** [comment]
Hi @Stasik0 
for me this works, i have checked it with my debugger.

I have forgotten this
```C++
#define UA_STRING_NULL {-1, (UA_Byte*)0 }
```
or 
```C++
const UA_NodeId UA_NODEID_NULL =
{
    0, UA_NODEIDTYPE_NUMERIC, { 0 }
};
```
If you want to save Code you can do it with this, what you think about it ?
```C++
//
// this works only with char [] strings
// Sample:
// const char m_strLang[] = "en_US";
// const char m_strCurrTime[] = "currenttime";
// const UA_QualifiedName m_dateName = { 1, UA_STRINGCONST(m_strCurrTime) };
// const UA_LocalizedText m_dateNameBrowseName = { UA_STRINGCONST(m_strLang), UA_STRINGCONST(m_strCurrTime) };
#define UA_STRINGCONST(s) {sizeof(s)-1, (UA_Byte*)s }

**Comment by jpfr** [comment]
In the `inline_setattribute` branch, most macros in ua_types.h are replaced with static inline functions.
The "static" values, e.g. UA_STRING_NULL are actual variables defined as extern const.

This version should be C++ clean.
https://github.com/acplt/open62541/blob/inline_setattribute/include/ua_types.h

Master will be merged on top before the pull request comes in.

**Comment by joerg555** [comment]
@jpfr 
Hi,
why you use for example
```c++
static UA_INLINE void UA_Boolean_init(UA_Boolean *p) { *p = UA_FALSE; }
```
and not
```c++
UA_INLINE void UA_Boolean_init(UA_Boolean *p) { *p = UA_FALSE; }
```
inline is inline, if the compiler can't do it, you can modyfy UA_INLINE macro ? :)

Stasik0:
@jpfr static and inline are orthogonal attributes, static is the scope, while inline is just a hint to a compiler

jpfr:
There are some differences between C and C++ for handling inline functions.
There are even differences between C versions (ANIS, 99, with GNU/MSVC extensions, ..).
http://www.greenend.org.uk/rjk/tech/inline.html

statinc inline does the right thing on every system we have encountered so far, including C++.

joerg555:
OK
thanks for the info

joerg555:
OK 
i have checked "ua_types.h / .c" inline version it in VS2013 / VS2015  and its working.
Only one error in ua_client.c i have fixed
```c++
        if(!UA_String_equal(&endpoint->securityPolicyUri,
                            &UA_STRING("http://opcfoundation.org/UA/SecurityPolicy#None")))
```
to 
```c++
        UA_String url = UA_STRING("http://opcfoundation.org/UA/SecurityPolicy#None");
        if(!UA_String_equal(&endpoint->securityPolicyUri, &url))
``
And what you think about UA_STRINGCONST(s)

PR #436, opened 2015-10-09
use UA_DataValue instead of UA_Variant for MonitoredItem_queuedValue

jgilje:
Using UA Expert, I noticed the source timestamp gets the server timestamp for monitored items. This patch fix this by changing the value field of MonitoredItem_queuedValue to UA_DataValue.


Stasik0:
its good, @ichrispa you own the code :)

Stasik0:
@jgilje could you squash the two commits into one?

jgilje:
I'll reopen the pull request with the commits squashed (could probably have done this easier, but I'm not familiar with all the aspects of git :))

PR #439, opened 2015-10-12
use UA_DataValue instead of UA_Variant for MonitoredItem_queuedValue (commits squashed)

jgilje:
Commits from pull request #436 squashed. Comment from previous pull request: Using UA Expert, I noticed the source timestamp gets the server timestamp for monitored items. This patch fix this by changing the value field of MonitoredItem_queuedValue to UA_DataValue.


Stasik0:
@ichrispa still has to sign off - he owns the subscription code

ichrispa:
Hi @Stasik0  and @jgilje,

the code looks good to me too. Note that we can't benefit from the source timestap right now for the lack of events, but this is a good change anyway :+1: 

Best regards,

Chris

PR #430, opened 2015-10-03
adding UA_Client_NamespaceGetIndex

HolgerJeromin:
fixes #405

PR #448, opened 2015-10-19
Aligning the server with the Highlevel API

jpfr:
Due to popular request (@ichrispa), this branch proposes a similar API for the server and client.

1. Aligned between client and server are
   - Node and references creation + deletion
   - Attribute reading and writing
2. The client highlevel API sits "on top" of the raw services and is branched out into separate files
3. Bugfix: Objects are now properly destroyed if their ObjectType has a destructor

@ichrispa Maybe we can have a 30min call to review the changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,7 @@ set(exported_headers ${PROJECT_BINARY_DIR}/src_generated/ua_config.h
                      ${PROJECT_SOURCE_DIR}/include/ua_log.h
                      ${PROJECT_SOURCE_DIR}/include/ua_server.h
                      ${PROJECT_SOURCE_DIR}/include/ua_client.h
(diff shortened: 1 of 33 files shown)


HolgerJeromin:
Do we have or want the high level api behind an ifdef? 
This question blocks #430

jpfr:
If the goal is to just reduce the binary size, we can leave it in.
Size-conscious users will use static binaries. And every unused highlevel function has zero overhead.

HolgerJeromin:
Ok. Good point. 
Then we could merge #430?

ichrispa:
Very nice. I will go over the changes to both APIs ASAP (theres a lot of changes), but at first sight the functionality I need for my projects seems to have been maintained.

jpfr:
Merged in c9ae4288d830fbcfb1bb65f526f4393b87dc45d4

PR #456, opened 2015-10-26
Update ua_subscription_manager.c

StalderT:
SubscriptionManager_deleteSubscription call SubscriptionManager_getSubscriptionByID to remove the subscription but SubscriptionManager_getSubscriptionByID don't use LIST_FOREACH_SAFE

### DIFF
--- a/src/server/ua_subscription_manager.c
+++ b/src/server/ua_subscription_manager.c
@@ -62,8 +62,8 @@ void SubscriptionManager_addSubscription(UA_SubscriptionManager *manager, UA_Sub
 
 UA_Subscription *SubscriptionManager_getSubscriptionByID(UA_SubscriptionManager *manager,
                                                          UA_Int32 subscriptionID) {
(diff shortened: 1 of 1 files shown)


Stasik0:
@StalderT I think "_SAFE" does not relate to thread-safety, but is the way to use a list iterator if a list-entry may be deleted during iterating over it. I do not see any deletion of list elements in SubscriptionManager_getSubscriptionByID, therefore the use of "_SAFE" Iterator only adds overhead.

StalderT:
SubscriptionManager_deleteSubscription call SubscriptionManager_getSubscriptionByID : 
UA_Subscription *sub = SubscriptionManager_getSubscriptionByID(manager, subscriptionID);

After call : 
    LIST_REMOVE(sub, listEntry);
    UA_Subscription_deleteMembers(sub, server);
    UA_free(sub);

StalderT:
Sorry I just understand. Thanks for the explanation.

PR #462, opened 2015-11-04
using strtoul instead of atoi in networklayer_tcp

HolgerJeromin:
refer to https://www.securecoding.cert.org/confluence/x/qoAyAQ


HolgerJeromin:
i could do the same in https://github.com/acplt/open62541/blob/master/examples/client_legacy.c#L503-L546 but this is probably not so important as the tcp layer (which may be used in production).

jpfr:
Thank you @HolgerJeromin :-)

PR #478, opened 2015-12-08
Consistent cmake option naming

jpfr:
@Stasik0 @FlorianPalm 
Are you ok with the renaming? It should be easy to adapt local build scripts.


jpfr:
@ichrispa
I hope the cmake option naming won't break anything on your end.

If nobody objects, this will be merged tomorrow.

Stasik0:
@jpfr probably a good time to make CI compile the examples aswell

doc/mainpage.dox
```diff
@@ -1,110 +0,0 @@
-/**
```
Stasik0: is this one beein copied somewhere else?

Stasik0:
oh no, it cannot be merged anymore

jpfr:
Master was merged. So the PR is currently read.
Only coveralls complains that the test coverage went down a bit.

PR #479, opened 2015-12-09
Adding context to a monitored item handling function.

Aleskey78:
Useful for passing extra data to callback, e.g. this pointer from a wrapper c++ class.


Stasik0:
looks good for me, same approach as we use in the callbacks on the server side, @jpfr @ichrispa do we have some code-style regarding the order and naming of parameters?

ichrispa:
@Stasik0: I always try to shift the "void *handle" to the end of the call, but it varies. I.e. methodcalls have them right after the callback, datasources in the struct.

jpfr:
We use such ```void *handle``` pointers in quite a few places now.

Maybe we can find a unified approach to userland "contexts" that are linked to nodes/sessions/subscriptions, such as

```c
typedef struct {
   void *handle;
   UA_Guid contextId;
   void (*destroyCallback)(UA_Context *ctx); // call with "&self"
} UA_Context;
```
Then we can tell the userspace when the context is deleted.
Ojecttypes/variabletypes could have custom context "constructor" function pointers.

ichrispa:
Though I don't see the requirement to change void *handles, I would like to request that we don't get a new struct and rewrite half the API again. I am having a hell of a job changing work-in-progress applications every time the stack gets fix A but breaks at several dozen places.

jpfr:
@ichrispa Not any node. Only those with callbacks into userland. Further, I have the use case in a project, were we want a node destructor callbacks. Currently we just leave a few bytes of "context memory" on the heap when the node is removed from inside open62541.

But let's not change the API as you said and do this post 0.2.

@Aleskey78 Looks good. I will merge as-is.

PR #483, opened 2015-12-14
Avoiding conditional directives that split up statements.

RomeroMalaquias:


RomeroMalaquias:
Ok, sorry for that.
Fixing it

Stasik0:
okay, this seems to be a non-functional coding-style issue @jpfr, @FlorianPalm what do you think?

RomeroMalaquias:
Hi @Stasik0,
It is only aesthetic refactoring, that compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

* [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
* [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

Stasik0:
+1 on that in this case

Issue #480, opened 2015-12-09
processRepeatedJobs, again non-RTC-problems, but in the future


NetoLe:
Hello again, and sorry again.

So it works for now, if the server starts in the past. But if it starts in the future, like 2037, it gets problems, when it gets synchronized to RT.

the if-part:
`if(tw->nextTime > current)`
breaks the loop that could set the time back.
the return then is the `MAXTIMEOUT`, because the `timeout` has values like over 3 billion, without setting the `tw->nextTime` to a valid value with which could be worked in the next iteration.

To explain the problem a bit clearer:
I see, that the timeout is needed for `ServerNetworkLayerTCP_getJobs` and the inherent select function. But I am adding my own repeated Job (Type: UA_JOBTYPE_METHODCALL) that I use for a workaround, which updates my Variable-Values every 250 ms. So it is part of the repeated jobs but not related to the network functions. 
I don't know how to get it working correctly without breaking other parts in the code. Maybe an iteration through the list setting back the nextTimes would work.

jpfr:
We can fix that.
Most systems let us use a monotonically increasing clock to measure elapsed time.
See for example here: http://linux.die.net/man/3/clock_gettime

That means, we would use a clock to measure time between repeated jobs and not the system time.
The elapsed time can jump forward, e.g. if the system was in energy-saving mode, but never backwards.

Stasik0:
porting clock_gettime to windows with performance counters http://stackoverflow.com/questions/5404277/porting-clock-gettime-to-windows

NetoLe:
ok, thanks. The monotonic clock sound's like a good idea. 
I noticed my problem was also caused by a NTP additionally. it starts at 1900 and not at 1970. I had an offset to 1970 and subtracted it always. but the counter started at 0 as an unsigned long. Subtracting then always got me values larger then the RT, which was the root of the problem.
For now I will just subtract if the current time value is greater then the offset to 1970, which fixes the problem temporary. 
There's another simple problem with the moment, when it get's synchronized: It loses connection for about 15-20 seconds. UaExpert turns to 'ConnectionErrorApiReconnect' and creates a new session if it gets a new connection. A client I programmed with the Client-Api from the stack can't reconnect automatically. I need to restart it. This happens just, when it starts in the past and gets synched to RT.
When it starts in the future and gets synched to RT it doesn't lose connection but is not capable of working any of its jobs, besides that the ServerTime is running correct then. 
[...]

Issue #472, opened 2015-12-04
Problem with non RTC


NetoLe:
Hello, I think this is more a minor issue, but I want to tell you that.

We try to implement an open62541-Server on an embedded device (and it works fine so far), which has no RTC.
That means it starts with a date 1.1.1970. It synchronizes when it gets connected to a PC, which causes to change the date by 45 years. That knocks out the server-application. Maybe this is a part that should be done in the future for embedded devices. We got it to work, commenting this part out (at ua_server.c):

  `  // UA_Job cleanup = {.type = UA_JOBTYPE_METHODCALL,
                      // .job.methodCall = {.method = UA_Server_cleanup, .data = NULL} };
    // UA_Server_addRepeatedJob(server, cleanup, 10000, NULL); `

We are not using the latest stable build because we want to use method calls (which are working fine) and the UA_Int..-casts on the byte-arrays in the decoder and encoder-parts caused the system to crash all the time, when reading casted 0xFF-Values on the first Byte (working on a 32-bit ARM); we implemented a shifting-operation, which we then saw in the latest-git-Version, which we are using now.

Stasik0:
Hi,

commenting server cleanup will maybe temporaliy fix your issue, but
probably cause some ressource leaks, since the server will never cleanup
the expired sessions and secure channels.
Could you please trace which line of the scheduler knocks the server out?
Actually it should be a good idea to fix it.

The second issue is probably the unaligned access. Did you try undefined
this variable manually
https://github.com/acplt/open62541/blob/9dd20578a6333e33f21a2c6e85c48b3e50c30f6c/include/ua_config.h.in#L69
?

Best Regards

PR #490, opened 2015-12-23
types: add a monotonic cpu timer to measure delays

jpfr:


Stasik0:
+1, affects #472

jpfr:
And #480 ...

PR #494, opened 2016-01-04
check datasource callback before invoking

Wtango:


jpfr:
Cool, that looks like a sensible change.
Can you please also change that documentation to match the implemented behavior?

ua_server.h currently says
```c
/**
 * Datasources are the interface to local data providers. It is expected that
 * the read and release callbacks are implemented. The write callback can be set
 * to null.
 */
```

Stasik0:
Yes, this is cool :) A unit-test to test this feature, or don't you have the time?

Wtango:
Unit-test is added.

jpfr:
Thanks @Wtango!

PR #495, opened 2016-01-04
single server config struct

jpfr:
The server configuration is folded into a single struct.
That includes the applicationDescription and buildInformation.

Startup currently looks like this:
```c
UA_ServerConfig config = UA_ServerConfig_standard;
config.running = &running;
config.logger = Logger_Stdout;
config.serverCertificate = loadCertificate();
UA_Server_addNetworkLayer(server, ServerNetworkLayerTCP_new(UA_ConnectionConfig_standard, PORT));
UA_StatusCode retval = UA_Server_run(server);
```

Ideas, comments?


Stasik0:
so instead of saving it in the "server" you save it in the "config". Well it is just the matter of taste, but, I would not prefer to have "running" inside of the config and better to pass it as an argument to the Server_run to indicate that this value is needed for the control.

PR #504, opened 2016-01-14
Subscription callback: using MonitoredItemId instead of ClientHandle

Aleskey78:
E.g. MonitoredItemId can be used for node name lookup in a callback function.
UA_Client_Subscriptions_addMonitoredItem returns indeed a "UA_UInt32 *newMonitoredItemId", not a ClientHandle.


ichrispa:
I used the clientSide handle in these function because... well... it's the client side and it has a handle. If it proves more advantageous to completely hide the handle and only work with the serverside monitoredItemId, I have nothing against that.

Aleskey78:
The reason for the change was, that ClientHandle and MonitoredItemId are not „synchronized“ ones "UA_Client_Subscriptions_addMonitoredItem" was called for a non-existing node.

Stasik0:
@ichrispa so you are okay with the merge?

Stasik0:
@StalderT @ichrispa bump?

Stasik0:
@ichrispa bump?

Issue #505, opened 2016-01-14
Client loses connection to Codesys server after ca. 3 mins.


Aleskey78:
Hello,

after around 3 minutes a client loses a connection to server, see the picture below.
Testet on 476af29985.

![opcua_verbindungsabbruch](https://cloud.githubusercontent.com/assets/3206586/12319031/5b1c1faa-ba9f-11e5-8dda-a0ba424cf3b6.png)

Stasik0:
Can be a token renewal problem, can you get us a wireshark dump?

Aleskey78:
here is a dump:
[connectionlost.pcapng.zip](https://github.com/open62541/open62541/files/90122/connectionlost.pcapng.zip)

NetoLe:
so your open62541 client sends a correct renew token but the codesys Server (is that on Port 4840, right?) response is an Error; maybe the Codesys server has no renewal implementation for SecureChannels?
I would suggest that, because the open62541-client gets into an Error-State and initializes a new SecureChannel following with a new Hello-Message, which it only does if the response is not the response which was expected
Can you get an log output for debug purposes, when `#define UA_LOGLEVEL 200` is set. this is placed in the generated ua_config.h. 
I think you will get a `"Reply answers the wrong request. Expected OpenSecureChannelResponse."` if I'm not totally wrong.

Aleskey78:
Hi, thanks for your prompt response.
I set UA_LOGLEVEL to 200, as you suggested, now a compiler brings me an error: 'UA_Server' has no member named 'logger'.
I use an amalgamated version.
It looks like "server->logger" should be changed to "server->config.logger".

Aleskey78:
I've got these two messages:
[01/14/2016 11:09:49.597] debug/channel Requesting to renew the SecureChannel
[01/14/2016 11:09:49.610] debug/channel Receiving OpenSecureChannelResponse failed

Stasik0:
Alexey, could you please also dump the three minutes of communication with uaexpert? I would like to see how it renewes the token, it should be the OpenSecureChannel request.

Please open a new issue for the logger problem - we probably need to include the debug level into the CI, you can also open a new pull request if you can fix it ad hoc.

Aleskey78:
@Stasik0, uaexpert dump probably won't help:[uaexpert_3min.pcapng.zip](https://github.com/open62541/open62541/files/90354/uaexpert_3min.pcapng.zip)

OpenSecureChannel message: OpenSecureChannelRequest
[...]

PR #506, opened 2016-01-14
Server logger

Aleskey78:
small change: "server->logger" to "server->config.logger"
see: https://github.com/open62541/open62541/issues/505#issuecomment-171613839


Stasik0:
seems to be fixed by dfec630 by @jpfr

PR #515, opened 2016-01-20
Changed addNetworkLayer method. Not included in source

Pro:
The source in the tutorial doesn't compile.
I changed it to use the same as in the server.c example. Please check and merge if correct.


Pro:
Since there are many other commits, I'll create a new pull request with cleaner commits

PR #519, opened 2016-01-26
Memory leak in MonitoredItem_CopyMonitoredValueToVariant

jgilje:
I'm using open62541 built with address sanitizer, which reports a memory leak in `MonitoredItem_CopyMonitoredValueToVariant`.

The solution is to not touch the returned data, and let `UA_DataValue_deleteMembers` do all the cleanup (equivalent to how this is done in `getVariableNodeDataType` and `getVariableNodeArrayDimensions` in ua_services_attribute.c).


jgilje:
Just wanted to add a note that appveyor failed with a FTP timeout in early setup.

jpfr:
Great, thx for the PR.

PR #520, opened 2016-01-26
Resize newValueAsByteString (in MonitoredItem_QueuePushDataValue) as needed

jgilje:
I'm hitting a bug where some of my items are larger than what fits inside the staticly sized newValueAsByteString in MonitoredItem_QueuePushDataValue - causing the item to report as updated for every iteration this method is run.

I see there's been some talk about `calcSizeBinary` which doesn't yet exist. This is just a suggestion until a better solution is implemented.

At the moment, I let newValueAsByteString grow indefinitely at the moment. An upper bounds check could easily be added if wanted / required.

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -263,6 +263,7 @@ UA_MonitoredItem * UA_MonitoredItem_new() {
     TAILQ_INIT(&new->queue);
     UA_NodeId_init(&new->monitoredNodeId);
     new->lastSampledValue.data = 0;
+    new->lastSampledSize = 512;
     return new;
 }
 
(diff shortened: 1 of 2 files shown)


jpfr:
Good idea to realloc.

However, I would not pull in this PR.
The "calcSize" solution just landed in 5fbf79ca39f1b3901acee602e13d315734707221.
Can you verify that this works for you?

Anyway, keep up the good work!

jgilje:
No worries :) My intention was just to have something in place until calcSize landed.

I just tested with the new calcSize code, and it works (some bugs discovered, I'll open a new PR for this).

PR #523, opened 2016-01-27
calcSizeBinary fixes

jgilje:
* calcSizeBinaryJumpTable was missing an entry for boolean types
* type was not updated before calling into calcSizeBinaryJumpTable, which meant that the calculated value was wrong (I have several types for my monitored items when this was discovered)


jpfr:
*ouch*
Added unit tests in e3062b2524962b8e0870ca757966f1321a2aa516.

Issue #492, opened 2015-12-25
Towards chunking while encoding


Stasik0:
xmas is here, time to get chunking working

@jpfr @FlorianPalm @ichrispa: I need some feedback on the architectural questions, do you think its gonna be okay, since the changes affect encoding and will make it instable again

There are 2 strategies possible in general:

* encode full message and cut it in chunks -> encoder can be left unchanged, however we need a buffer for the full message which may be arbitrary large

* make encoder to be able to "skip" the already encoded chunk(s) while it runs recursively -> this is what I have chosen, it means:
    * every encode function becomes a "skip" size_t parameter, as long as skip is greater than 0 the function does not do anything
    * we interpret the error messages ```UA_STATUSCODE_BADENCODINGERROR``` of the encoder as "buffer has ended, but there is more stuff to encode in the message"

So the strategy to make chunks in ```ua_securechannel.c``` will be to call encoder several times on the actual message in [this line](https://github.com/open62541/open62541/blob/master/src/ua_securechannel.c#L138) (I will assume that no chunking will happen in the header stuff, since a minimal chunk size is 8192), as long as we receive ```UA_STATUSCODE_BADENCODINGERROR``` from the encoder we just accumulate offsets from the previous chunks as the skip parameter. Getting the offest from the previous chunk will automatically ensure the skip is always valid e.g. we do not try to skip a half of an integer.

okay, i started to implement the stuff in a separate branch, please look into the array encoder to see how it works: https://github.com/open62541/open62541/commit/4309d6e21093bf20222b0571a6524bf4386686ca#diff-c3b90ce2fd11b5aac86ded3a6d7ac74cR336, __CODE NOT REALLY TESTED__

so guys, is it a valid approach to go? Regarding performance,  it is actually an additional check ```skip > 0``` per single encoder call.

HolgerJeromin:
Do you want to release version 0.2 before or after this change?

Stasik0:
Do we have to decide now? User api is not changed.

HolgerJeromin:
But stability probably :-)

jpfr:
1. What is the interplay with encryption? Is the entire message encrypted and then cut into chunks or is it encrypted separately?
2. We have a max chunk size and a max chunks count. So messages always have an upper bound on size.
3. If you activate the "readspeed" test case, you can see the effects on speed.

[...]

PR #524, opened 2016-01-28
Support for sending chunked messages

jgilje:
I've added support for **sending** chunked messages. Receiving chunked messages is not handled in this PR. Tested against UAExpert and wireshark.

Note that the maxMessageSize and maxChunkCount in UA_ConnectionConfig must be updated to notify remote it supports chunked transfers (UAExpert closes the session if not).


FlorianPalm:
Hi,
there is an discussion about chunking in #492. We have already a solution which does not need to allocate a buffer for a complete message. Check out the chunking_callback branch, for example.

jgilje:
Oh right!  I had actually read that PR a couple of weeks ago - and forgot all about it. Closing this issue then.

PR #530, opened 2016-02-01
modify UA_Client_Subscriptions_manuallySendPublishRequest to return status

StalderT:
Hello,

If I modify the client (in example directory) with something like that : 

```
while(1)
{
    UA_Client_Subscriptions_manuallySendPublishRequest(client);
    sleep(1);
}
```
I don't have possibility to know if the client are connected to the server.

You can find a PR to correct this problem.

### DIFF
--- a/include/ua_client_highlevel.h
+++ b/include/ua_client_highlevel.h
@@ -275,7 +275,7 @@ UA_Client_Subscriptions_new(UA_Client *client, UA_SubscriptionSettings settings,
 UA_StatusCode UA_EXPORT
 UA_Client_Subscriptions_remove(UA_Client *client, UA_UInt32 subscriptionId);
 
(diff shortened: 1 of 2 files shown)


Stasik0:
+1 here, @ichrispa owns the code
@StalderT could you squash your commits to one?

StalderT:
@Stasik0 do you know how I can squash commits?

jpfr:
Squashing can be done via rebase -i.
https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

Then `push --force` the branch to overwrite your PR with the single commit.

StalderT:
Sorry, I have tried, but I can't squash commits.

ichrispa:
Seems reasonable to me :+1:

PR #531, opened 2016-02-02
recv don't support timeout under cygwin

StalderT:
Workaround for https://cygwin.com/ml/cygwin/2013-07/msg00107.html


jpfr:
Great, thanks for the PR.

PR #532, opened 2016-02-02
add SO_KEEPALIVE option

StalderT:
https://en.wikipedia.org/wiki/Keepalive#TCP_keepalive


Stasik0:
@jpfr this has no "real" impact on our tcp stack, does it?

jpfr:
@StalderT What is the motivation for the TCP keepalive?
SecureChannels already need to be renewed after a timeout.

StalderT:
Ok, no real motivation just a suggestion.

PR #534, opened 2016-02-02
Chunking callback

jpfr:
I just merged master into this branch.
We were diverged a bit. But now I can connect with UAExpert again..

From my point of view, three things are still left open.
- Aborting after too many chunks / total message size did not survive the merge. Sorry!
- Testing! For example with the official CTT.
- We should move /examples/server_chunking into a proper unit test.


Stasik0:
Unit tests on the service level?

FlorianPalm:
Good! I will try to get these things to work.

FlorianPalm:
Please review this pull request @jpfr .
I have added two unit tests for encoding big strings and big arrays (credits goes to @Stasik0). Still, there is a demo application server_chunking which proves the chunking mechanism. For clients it is now possible to read AND write big data values.

Some effort needs to be put into the connection buffer negotiation and error handling of chunked messages.
I would prefer to merge this pull request as soon as possible to reduce merging overheads.

jpfr:
Yay! Like it.
There are two smaller things remaining.
- Sending aborts is marked as todo. Under which circumstances do we need that?
- The examples/server_chunking.c is needlessly big. Can we reduce that to the essential, i.e. a big string to test chunking with?

Stasik0:
@FlorianPalm the client seems not to work i.e. it trys to send 65k buffer to a 8k limited server :)

jpfr:
@FlorianPalm I see you kept working on the branch.
It all looks very good to me. But I'm not that deep into the details any more.
Are we ready to merge?

FlorianPalm:
I will need to have a look at the client code. @Stasik0 mentioned that there were some issues. I will do it in the next few days.

FlorianPalm:
The client is now able connect to the standard "server" application again. I changed the connection settings locally within the server_chunking example to force chunking. I would say that this branch is ready to get merged.

jpfr:
Merged.
Great job @FlorianPalm @Stasik0!

FlorianPalm:
puhh...finally, nice!

Stasik0:
Does it mean we are feature-complete for a 0.2-RC?

PR #540, opened 2016-02-05
fix possible memory leak

StalderT:
Hello,

I don't have test case and I don't know if it is needed but MonitoredItem_QueuePushDataValue always call UA_DataValue_deleteMembers(&newvalue->value) before UA_free(newvalue). "We cannot remove the oldest value and theres no queue space left. We're done here." is peraps for that?

It is just a question/suggestion. If no simply close the PR.


Stasik0:
@ichrispa will know surely

PR #544, opened 2016-02-11
Chunked receive

jgilje:
This PR makes it possible to receive chunked messages. As far as I can tell, the current chunking branch and PR only deals with sending chunked messages.

There is no background cleanup of partially received chunked messages, if a client has stopped sending (rogue clients).

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -320,7 +320,6 @@ processMSG(UA_Connection *connection, UA_Server *server, const UA_ByteString *ms
     UA_StatusCode retval = UA_UInt32_decodeBinary(msg, pos, &secureChannelId);
     retval |= UA_UInt32_decodeBinary(msg, pos, &tokenId);
     retval |= UA_SequenceHeader_decodeBinary(msg, pos, &sequenceHeader);
-    retval |= UA_NodeId_decodeBinary(msg, pos, &requestTypeId);
     if(retval != UA_STATUSCODE_GOOD)
         return;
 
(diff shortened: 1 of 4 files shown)


jpfr:
Great news!
There are couple of minor things that might be improved. Otherwise, perfect. :+1: 

- We probably don't need the additional datatype in Custom.Opc.Ua.Transport.bsd. The final 'F' is already bit-filtered out in `UA_MESSAGETYPEANDFINAL_MSGF & 0xffffff`. Or have a look at the Custom.Opc.Ua.Transport.bsd in the chunking PR.
- The code currently does not handle abort chunks 'A'. That may happen e.g. if the first chunks were sent out optimistically, but then there are more chunks required than the remote side can handle.

Stasik0:
@jgilje could you also tell us whether and how you have tested it?

jpfr:
Thinking in terms of the overall architecture.
For example, we have UA_Connection_completeMessages to join split packages into a complete chunk.
Similarly, we could put a UA_SecureChannel_completeRequests into the `ua_securechannel.c`.

This would also provide the ideal starting point for add security. That needs to happen in the securechannel and chunks are encrypted individually.

jgilje:
@jpfr about UA_MESSAGETYPEANDFINAL_MSGF, my fault! I started implementing against not up-to-date sources. I'll fix this and add A(bort) handling as well, and update my branch.

jgilje:
@Stasik0 Sure, I first came across this when an in-house app sent large messages to my server built with open62541. After that I was able to create a simpler test-case using UAExpert, by subscribing to lots of signals, so it would require chunking to send a CreateMonitoredItemsRequest. After this, I verified the message was indeed chunked by logging network traffic in wireshark (which has opcua decoding built-in).

jpfr:
Great. I suppose we can merge once you have made those changes.

jgilje:
@jpfr branch updated with suggestions. added some cleanup details as well

jpfr:
We're getting there. Some small things remain:

- `ch` is used after beeing freed: https://github.com/open62541/open62541/pull/544/files#diff-f7732748b7e13e607783dbd36f54eb82R402
- The network layer might give us multiple chunks at once. We check that the position offset has been increased correctly in `UA_Server_processBinaryMessage`. But you set it to zero at https://github.com/open62541/open62541/pull/544/files#diff-f7732748b7e13e607783dbd36f54eb82R403
If that is too difficult to track, we can change the behavior of `UA_Server_processBinaryMessage`.
- When a request is aborted, the chunk is removed, but the buffer is never freed. https://github.com/open62541/open62541/pull/544/files#diff-f7732748b7e13e607783dbd36f54eb82R412

jpfr:
- You decode the length and memcpy the content. Please check that the length stays within the boundaries of the msg. Classic case of a buffer overrun. https://github.com/open62541/open62541/pull/544/files#diff-f7732748b7e13e607783dbd36f54eb82R322

jpfr:
- The commented out error messages are quite nice. Maybe give them a lower log level so that they can be filtered out. Such as `UA_LOG_DEBUG(server->config.logger, UA_LOGCATEGORY_SECURECHANNEL, "Chunk message");`

jgilje:
@jpfr Great feedback! I'll get onto it tomorrow.

jgilje:
- ch use-after-free fixed
- cleanup bytes on aborted messages
- log chunk packets: I set the log-level at trace, and switched the category to securechannel
- the position: I deliberately set it to 0 after receiving the final chunked message because we read from the reconstructed message. I corrected *pos at the cleanup point. At the cleanup point however, I did the assumption that this message was the only one in msg. That has been fixed now.
- length and memcpy, the line before the memcpy is a realloc of size len, so it should be impossible to overflow I guess?  I could add a check that the realloc succeeds.

Stasik0:
+1 for checking realloc return since we might run out of memory

Stasik0:
@jgilje any idea, whether we can unit-test some of the code?

jgilje:
@Stasik0 We could possibly create some kind of dummy message and artificially split the message using the upcoming message chunker. After that, I would have verified we got the exact same message reconstructed using the code in this PR. That's the best I can come up with at the moment.

I'll try to get something done regarding realloc failures.

jpfr:
> length and memcpy, the line before the memcpy is a realloc of size len, so it should be impossible to overflow I guess? I could add a check that the realloc succeeds.

The culprit lies here: `memcpy(&ch->bytes.data[ch->bytes.length], &msg->data[*pos], len);`
`ch->bytes` will not overflow after the malloc. But `msg->data` still can.
Imagine a chunk of size 200 that advertises itself as being of size 250.

I resurrected unit tests with binary messages.
We could cut a dumped request into chunks by hand. The hex-format is quite readable and is converted to a binary file for the unit tests with a small python script.
https://github.com/open62541/open62541/blob/master/tests/dumps/client_HELOPN.hex

That also allows us to fuzz the server. We have currently no exploitable error for HEL/OPN/CreateSession that can be found by afl: http://lcamtuf.coredump.cx/afl/.

jgilje:
@jpfr Got it. Added a check for realloc as well, as suggested by @Stasik0

jpfr:
The PR was squashed to a single commit and merged to master.
See 5986df3f0ee3cd3318593b47ebe547b5877539b4.

jpfr:
Thanks @jgilje !

Issue #391, opened 2015-09-09
Migrate travis to new infrastructure


jpfr:
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

HolgerJeromin:
I am not sure if this qualifies as easy hack ;-P

Issue #371, opened 2015-09-02
add static code analysis on travis


HolgerJeromin:
For someone with free time. :-)

The `.travis.yaml` should probably cleaned up anyway...
For reference this blog entry could be helpful
https://legalizeadulthood.wordpress.com/2014/12/07/adding-static-analysis-to-your-c-github-repository/

PR #551, opened 2016-03-01
Migrate to new travis infrastructure

Pro:
I changed the travis files to use the new travis (docker) infrastructure (see #391).
Addionally I also added static code analysis using clang and cppcheck (see #371).

NOTE:
* Compilation runs through, it only fails with coveralls (because it's a fork), should be fine in the original repo
* There are some errors which cppcheck reveals. Maybe needs further fine tuning
* The build for OS X still fails because there's still a problem with liburcu


Stasik0:
:top:, great job! i will look into it in the evening!

Stasik0:
@Pro could you please squash your commits to one?
Question to all:
* what should we do with OSX? 
* cppcheck warning needs to be also fixed before we can go live.

So probably the way would be to to accept PR as a branch and tune code until everything works

Stasik0:
Probably cppcheck is right: the memory allcated by the decode https://github.com/open62541/open62541/blob/fb0266dd12c71e4776881293daa18c92813f51b9/src/ua_types_encoding_binary.c#L667 is never freed, we probably did not see it since nodeIds of NS0 are always numeric, @jpfr what do you mean?

Stasik0:
@Pro https://travis-ci.org/Pro/open62541/jobs/112856598 could you please restart this one? there seems to be a problem with amalgamation

Stasik0:
@Pro could you also allow failing for builds with ANALYZE=true (https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail)

Pro:
@Stasik0 done as requested.
The job can't be restarted since the commits are now squashed.

Pro:
It also looks like that when using `clang` amalgamation randomly fails... (This happend to me a few times when I ran all the tests on travis)

Stasik0:
amalgamation is just a python script, probably it gets terminated to early and does not finish the output, hmm is it really non-deterministic?

Stasik0:
@Pro getting picky: is it possibly to do cross-compilation stuff only if gcc is set as compiler?

Pro:
I couldn't see any deterministic behaviour.
https://travis-ci.org/Pro/open62541/jobs/112865574 failed for the first time, then I just restarted the build and now it seems to work

Stasik0:
@Pro 
* looking through files, imho "make -j8" causes with the amalgamation, probably we have not documented some dependency, for now can you just call "make" when using amalgamation?
* There are some more fixes I will need to look into, e.g. locating lxml in CmakeLists.txt? Why did you removed the old way to do it?

src_extra/networklayer_tcp.c
```diff
@@ -35,7 +35,6 @@
 
 /* workaround a glibc bug where an integer conversion is required */
 #if !defined(_WIN32)
-# include <features.h>
```
Stasik0: @Pro y this?

CMakeLists.txt
```diff
@@ -8,16 +8,10 @@ project(open62541 C)
 
 # Find Python and Python-lxml
 find_package(PythonInterp REQUIRED)
-execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"
```
Stasik0: @Pro is it really better? =)

src_extra/networklayer_tcp.c
Pro: It isn't used (as far as I could see) and this file doesn't exist on OS X (it caused compile errors)

CMakeLists.txt
Pro: Jep, because the get_python_lib command only returns the global path, i.e. if you install packages as root.
Since the new travis containers don't allow sudo installs, we have to use `pip install --user` and thus the lxml package isn't by this snippet.
That's why I changed it to `import lxml` and check the return value. Then you are 100% sure that python can find lxml or can't find it.

Pro:
Regarding the `make -j8` and amalgamation:
I would suggest trying to fix the dependency problem, otherwise you may run into different problems, e.g. on my system make uses by default j8
If you want, I can also change it back to `make` (if really needed)

Stasik0:
I agree, solving dependency problem should be better but, it postpones the merging of the PR in this case :/ I do not have time in short term. Maybe we can elaborate where the problem is exactly?

CMakeLists.txt
Stasik0: agreed

src_extra/networklayer_tcp.c
Stasik0: https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a00863.html#_details

hmm it seems to be related to some included features of c++ stdlib, it came into https://github.com/open62541/open62541/commit/2ae7fef1e4ae33e94a95c6f7b653655c6ff0d328 by @jpfr, can you help us?

Pro:
I think I found the amalgamation problem:
https://github.com/Pro/open62541/blob/master/CMakeLists.txt#L378

Shouldn't there be the same for `server_static`, i.e., `ADD_DEPENDENCIES(server_static amalgamation)`
(it's already like this for `client_static`)

Should I change it?

Stasik0:
@Pro yep, change it, it should not harm. It may be the problem indeed

Pro:
@Stasik0 unfortunately it didn't solve the race condition issue. Needs further investigation. For now I switched back to single core make.

Stasik0:
do you have "build_examples" enables when you experience race condidtion? Because exmaple servers do not depend on amalgamation. I am also not sure if we need this target anymore, since open62541.c and .h are included into open62541-object.

src_extra/networklayer_tcp.c
jpfr: features.h defines __GNU_LIBRARY__ et al.
We need to detect the glibc version, since the -Wconversion flag uncovers a bug in old glibc headers.

So far I am unaware of a non-Windows libc that does not come with this header.
Even embedded-stuff like musl has it: https://github.com/rofl0r/musl/blob/master/include/features.h

@Pro No features.h on Mac??? That's odd. What compiler do you use?

jpfr:
Wow! That's quite an improvement to our build infrastructure.
Great stuff @Pro.

All checks are passed. So let's squash and merge this.

On a related note, Coverity checks didn't run since a long time ago.
Automated documentation updates use a similar mechanism in travis and work correctly.
But let's get this PR in first.

jpfr:
> Probably cppcheck is right: the memory allcated by the decode https://github.com/open62541/open62541/blob/fb0266dd12c71e4776881293daa18c92813f51b9/src/ua_types_encoding_binary.c#L667 is never freed, we probably did not see it since nodeIds of NS0 are always numeric, @jpfr what do you mean?

Nothing wrong here. All internal DataTypes have a numeric NodeId.
If we decode a string NodeId, then we will not find the matching DataType. And return an extensionobjct with the binary data + the decoded nodeid.
So the decoded nodeid is freed together with the extensionobject we return.

src_extra/networklayer_tcp.c
Pro: @jpfr I tested compilation using travis and there I had the problem that features.h is not available on Mac. Maybe it's best to use a precompiler definition and include features.h only for linux.

Shall I change it back so we have a clean PR?

Pro:
@jpfr if you need anything from me (e.g. squash all commits of this PR into one) let me know.

HolgerJeromin:
If you add the string
fixes #391
fixes #371 
to the first comment, both issues will be closed automatically on merge. :)

Sadly i lost my admin rights here with the move out of the acplt organisation, so i cannot change it by myself. :-)

Pro:
@HolgerJeromin jep, that was my intent to close those issues with the commit :+1:

src_extra/networklayer_tcp.c
jpfr: It looks as if features.h is included with stdio.h when glibc is used.
And since stdio.h is used, just remove the line as-is and it should be fine everywhere.

Pro:
@jpfr there was still an error in the build script (the examples were always built with gcc, even if clang was selected). This is now fixed (see https://github.com/open62541/open62541/commit/f0170c06ea9bc11d9a0156464a1c57ce1e3dd244#diff-cd96240c93b4c91b51c86b9e774296a4R76), but now valgrind and clang fail (see https://travis-ci.org/open62541/open62541/jobs/112993383).

Stasik0:
valgrind seems to detect a segfault? 
@Pro side-issue: could you try this simplification of the cmake file https://github.com/open62541/open62541/compare/simpify_amalgate? I cannot reproduce any race-condition locally with it.

Pro:
@Stasik0 merged as requested. I also re-enabled the make -j8 flags. Looks good so far (at least for this CI build)

jpfr:
I guess the valgrind error is due to the docker-based infrastructure. I ran the example with the same clang-3.4 (but a current valgrind) and did get good results.

```
==9374== Memcheck, a memory error detector
==9374== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==9374== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==9374== Command: ./server
==9374==
[03/02/2016 12:43:39.394] info/network  TCP network layer listening on opc.tcp://virtualbox:16664
[03/02/2016 12:43:40.203] info/server   Received Ctrl-C
[03/02/2016 12:43:40.206] info/network  Shutting down the TCP network layer with 0 open connection(s)
==9374==
==9374== HEAP SUMMARY:
==9374==     in use at exit: 0 bytes in 0 blocks
==9374==   total heap usage: 1,761 allocs, 1,761 frees, 3,972,076 bytes allocated
==9374==
==9374== All heap blocks were freed -- no leaks are possible
==9374==
==9374== For counts of detected and suppressed errors, rerun with: -v
==9374== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

On travis we don't even get the "received ctrl-c" output. So sth is wrong on the system level.

Stasik0:
Why is it working with gcc? Different libstdc for signal handling?

Pro:
I now upgraded the CI scripts to use the newer valgrind version `3.11` and there it seems to work. Looks like the SIGSEGV was a problem in the older valgrind version. (It now downloads the valgrind, builds it and installs it, whereas valgrind will be cached to avoid building it every time)

In my opinion this PR can now be merged :+1:

jpfr:
Squashed and merged in a6fbd5c.
Nice work @Pro.

PR #557, opened 2016-03-15
Adds dockerfile

btrepp:
This uses alpine linux and attempts to create a minimal install with tools to develop using open62541
Copies the amalgamated header file into /usr/include and the dynamic and static libraries into /usr/lib
This should allow default gcc to pick them up easy enough

EDIT: also my C/CMake is a little rusty. I wanted to add the files in standard OS locations so people didn't need to mess with installs. If a cmake install target is added it would probably be better to run that instead of the manual CP steps I've done for the docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM alpine:3.3
+RUN apk add --no-cache cmake gcc musl-dev python py-lxml make && rm -rf /var/cache/apk/*
+ADD . /tmp/open62541
+WORKDIR /tmp/open62541/build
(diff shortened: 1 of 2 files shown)


Stasik0:
I'd suggest to merge it and to create a decent cmake [install target](https://cmake.org/cmake/help/v3.0/command/install.html) in a consequent step

btrepp:
I have added a tiny dockerfile to the PR. I've set up my private repo fork to build it automatically on dockerhub.

PR #564, opened 2016-03-21
Create documentation from headers

jpfr:
We need good docs. So it should be simple to write good docs.
This PR proposes a very simple way to generate documentation from our public header files.

Everything within `/** */` comments becomes restructured text.
The rest is included as "C" code with syntax highlighting.
So yes, the entire header is part of the documentation. As an additional motivation to write clean code. :smile:

The result for types.h can be seen here: https://gist.github.com/jpfr/0e18ffef3cbb5a8436ac

If this change gets approval, we can integrate the conversion with c2st.py in cmake and fold some of the existing documentation into the header files. Of course, we can still have pure .rst files, such as the tutorials.


jpfr:
This produces quite good results now: https://gist.github.com/jpfr/0e18ffef3cbb5a8436ac

Anybody not okay with merging this?

Stasik0:
ths gist ist looking good, why is the build failing?

PR #565, opened 2016-03-22
Added file for additional types to the codegenerator

Root-Core:
This was a failed pull request for an UTF8 Test.. now it is for a patch of the codegenerator.

Now deprecated:
Can't find more right now, the full shema seems to be not affected.
(Seems to be an test for special chars, so it _could_ be useful to someone.)


Stasik0:
Yea, its seems to be a UTF-8 test, @ichrispa okay to accept?

Root-Core:
I modified the type codegenerator to support additional types.
I am using it to generate Range and EUInformation.

It was not intended to be added or part of this small pull request.
Nevertheless it would be good to provide such an option, or was it there before?

Sorry, for that mistake. I am fairly new to github.

jpfr:
Hey @Root-Core,
imho adding more types should be a separate pull request.
But that's a good idea.

I don't think we need to add small grained selection besides "minimal" and "all" types...
Makes the code easier also.
But maybe somebody else has a strong opinion on this.

Root-Core:
Hello @jpfr,
you are right, it was intended to be a second pull request (or for the moment a workaround for me until i come up with something better), but i messed some things up. Sorry.

It might be a good idea to generate all types and use a text file as blacklist (as we exclude some types already). It would be possible to use a second, extended blacklist to enable an minimal set.

Root-Core:
I have separated the two pull requests, wich may help to get one of them merged easier.

jpfr:
With the last refactoring, we use a separate file with the types that shall be created.
https://github.com/open62541/open62541/blob/master/tools/schema/datatypes_minimal.txt

It should be fairly easy to append additional names to that list in your build process.
Are you okay with that solution or do you need a separate file for additional types?

Root-Core:
I am absolutely fine with that change. Thanks a lot for the support.

PR #571, opened 2016-03-31
Separate Publish Job for Subscriptions

jpfr:
This change separates the subscriptions callbacks into two sepate repeated jobs:
- sampling values (job per monitoreditem)
- sending out notifications (job per subscription)

The network traffic is greatly reduced since we use asynchronous responses.

Subsriptions are working fine as per UAExpert.
But some edge cases are not finished. For example Republish is incomplete.

Needs approval from @ichrispa before merging.
Feedback is highly appreciated.


jpfr:
Okay. This PR on async subscriptions getting serious.
Reviews welcome.

**However** our current client might block with async subscriptions.
Because we wait until a response for the publish request comes back. That may be **a lot later** since the server can put the publish on hold (and not answer or answer to a later request) until a notification appears.

That means, we will have to add async handling to the client. Otherwise, we cannot have standard-compliant subscriptions.

PR #573, opened 2016-04-01

Root-Core:
As DataSources may be based on hardware or event based channels
that could be subscribed (instead of an slower polling approach), it
makes sense to support "enabling" and "disabling" them.

Other typed Nodes do not need this functionality, as they are meant to
provide data by themself or from a source in the instantiating class.

Examples are provided.


Root-Core:
Sorry for that commit / check spam.. have to use the browser to commit changes atm.. it sucks.

PR #575, opened 2016-04-01

Root-Core:
As DataSources may be based on hardware or event based channels
that could be subscribed (instead of an slower polling approach), it
makes sense to support "enabling" and "disabling" them.

Other typed Nodes do not need this functionality, as they are meant to
provide data by themself or from a source in the instantiating class.

Examples are provided.

// Sorry: As the last pull request was flooded by checks, i wanted to open a new one.


jpfr:
Let me try to summarize the use case.
- In datasources we have the usual `read/write` callbacks.
- You have hardware where reading needs to be *enabled*.
- Enabling takes time. You don't want to do it for every read.
- You do not want reading enabled all the time when no subscriptions are present.

Is that correct for the motivation?

Root-Core:
Hello @jpfr,

the hardware provides an event driven output of data with filters which raise an event, if there is a new Value / Data. The event are blazing fast, but if you poll a "channel" (the hardware) cyclic it gets really slow and in some configurations, there are thousands of channels.
(It would be neck-breaking to get this event from every single channel)

To put it all in a nutshell, you have summarized it pretty well. We do not have to enable reading every time, we can just "read".. but this in fact is really slow. That's why the server should subscribe all needed "channels", as long as the subscription (UA) is present.

I think there are also other use cases, where such an callback could be useful, for example something you just summarized.

jpfr:
Ok. That sounds like something we should support.

But we might want some more machinery in the background to make it solid.
For example, there can be multiple clients that open a MonitoredItems on a value.
Form what I see in your code, you do not count the number of current subscriptions on a variable.

Do you plan on doing that manually on the userspace side?

Root-Core:
I implemented the stack as such a channel (as in this project everything is a channel), and in this channel i am counting the instances of monitored items.

It was the first iteration of this callback. I wanted to provide some logic like counting the instances in further iterations, but i was not sure if that makes sense in first place.. it heavily depends on the project, that uses the stack. Some could have a counter in there "subscribe" logic, some not. It would not hurt, to count it twice, though.

I was not sure how big this callback should get for this patch. This is why i hoped for responses to this pull request. ;)
Implementing the logic besides of the clean callback, may should lay at the person, that implements the DataSource, but if you can support the engineer, you should support the engineer.

So you may only want to trigger this callback, if the first UA subscription is requested and if no UA subscription is left. The DataSource could use the return value to tell if the source was activated.. in some cases, some hardware might fail or something like that..

Root-Core:
Updated version: #603

PR #591, opened 2016-04-21
Add static analysis with amalgamation enabled

Pro:
Add additional check with `scan-build` and amalgamation


Pro:
Travis fail will be fixed by https://github.com/open62541/open62541/pull/592

PR #603, opened 2016-05-10

Root-Core:
An updated version of #575, as the subscriptions were refactored.

It should now prevent segfaults by setting all "monitored" members of datasources to NULL and check against it. The only problem i can see is, that it could cause problems in existing implementations.

Any ideas about that issue?
We could use a boolean to activate the events, but maybe someone comes up with an better idea.


Pro:
@Root-Core do you have time to merge master into your PR and solve the conflicts?

Additionally can you remove your changes to the appveyor.yml file (and if needed create a new PR which applies the changes for appveyor). I think that change is a result of the merge.

And you can also enable the option "Allow edits from maintainers" at your PR so that we can directly push to your branch (In this view, at the right bottom side of the PR).

It would be nice to merge this PR before it comes more and more abandoned.

As discussed in #575 there are still some open points:

@jpfr wrote:
> Ok. That sounds like something we should support.
> 
> But we might want some more machinery in the background to make it solid.
> For example, there can be multiple clients that open a MonitoredItems on a value.
> Form what I see in your code, you do not count the number of current subscriptions on a variable.
> 
> Do you plan on doing that manually on the userspace side?

@Root-Core response:

> I implemented the stack as such a channel (as in this project everything is a channel), and in this channel i am counting the instances of monitored items.
> 
> It was the first iteration of this callback. I wanted to provide some logic like counting the instances in further iterations, but i was not sure if that makes sense in first place.. it heavily depends on the project, that uses the stack. Some could have a counter in there "subscribe" logic, some not. It would not hurt, to count it twice, though.
> 
> I was not sure how big this callback should get for this patch. This is why i hoped for responses to this pull request. ;)
> Implementing the logic besides of the clean callback, may should lay at the person, that implements the DataSource, but if you can support the engineer, you should support the engineer.
> 
> So you may only want to trigger this callback, if the first UA subscription is requested and if no UA subscription is left. The DataSource could use the return value to tell if the source was activated.. in some cases, some hardware might fail or something like that..

@jpfr is there something we should add to the PR, or is this sufficient for a first implementation?

jpfr:
Currently, we check for data changes "stupid" by reading after every sampling interval.

- We don't have a performance problem (in my experience)
- We have to do cylic checking anyway to ensure that the timestamps are according to the sampling interval

I see the improvements from this PR with rather with regard to events (triggered by objects).
There, we have notifications that are triggered manually. Similar to the `monitored` callback we see here.

Root-Core:
@Pro I will have a look into it. I should find some time next week.
The change to the `appveyor.yml` seems to be non intentional. So this will be reverted, I guess.

You should have the rights to edit the PR now.

@jpfr I do not fully understand your point. It seems, I am missing the link regarding the data change check. This callback here is atm only a hint for data sources, that they need to provide data from now on until they are not longer subscribed (well "monitored").

The logic should be improved in future, so that we count internally and we also should think about how to handle disconnection. This might reduce the count, but the spec allows subscriptions to last beyond disconnection (if not, please correct me). It has been a while, since I looked into the spec.. so I am not sure what is actually happening (or should happen) with the data changes between disconnection and reconnection of an client.

In our example, we have some really fast hardware channels, that should know if they are used. Therefore it would be rather nice to have such hints regarding a data source. It will help managing them, which will be more and more important with a growing share of hardware based implementations in the future.

jpfr:
@Root-Core 
Ok, got it.
You want to keep "sensors" that are continuously monitored in an "alerted" mode.
Thank you for the clarification. That's a good feature!

What about multiple monitoreditems on the same attribute?
Should we keep a counter of how many monitoreditems are active on the value?

Pro:
@jpfr I cleaned up the PR and rebased it.

What is still missing is a count on currently active monitored items on the same attribute.
I.e. the monitor callback should only be called on the first monitored item creation and on the last removal.
Do we want to handle that in the stack or on the user side?

Root-Core:
I rebased the PR, but it seems to have gone horrible wrong. Sorry about that.
If you compare the `Root-Core:Subscription_Events` branch and the `open62541:master` recursively, everything looks okay. If I look into my branch, it says `This branch is 4 commits ahead of open62541:master.` and a comparison via `New pull request` looks fine. No idea, why GitHub thinks the PR does include 232 changed files. 

About the counter, I would vote for stackside counting as it wouldn't make sense to count it by the user IMHO. I implemented it in the `UA_DataSource` struct, but I don't like the idea, that you can and have to initialize the member manually, which always should be 0.

A node store could be handy here, as I also thought about a hashmap. I exposed the ref count, but if I try calling `findOccupiedSlot()`, on a empty map I get an exception. This also is true for `UA_NodeMap_getNode()` This seems to be a problem with the node store plugin?

I have to investigate further.
Regards.

Pro:
I just did a cherry pick of your commit and cleaned it up a bit, then force-pushed, so the PR is clean now. So please update/pull before you continue.

Regarding the counter it's probably best if we add an additional variable to the UA_Node struct which is a list of all the active subscriptions on the node. This can be used to detect if it is the first subscription on the node and may also be useful for other use-cases in the future.

@Root-Core can you propose a corresponding change? It should suffice to add the subscription to the list when adding a monitored item, and remove it when the monitoring item is deleted or destroyed.

Root-Core:
@Pro Thank you. I've added a list to the UA_Node. I hope this works as expected. Some things might be improvable.

jpfr:
Do we need a list? That's quite some overhead.

An uint32 counter could be enough.
What do you say @Root-Core ?
We can then call the callback when the counter drops to zero.

Root-Core:
Memory wise it should not make a huge difference as it only stores the pointers. But a list has a bigger footprint indeed. Might be interesting to analyse nodes, that are not subscribed. If many nodes have an empty list, it could be some overhead. I've not compared it.

@Pro said the list 
> may also be useful for other use-cases in the future

So I've no strong opinion about this implementation. As the variable is available for all kind of nodes, I thought about making the callback work more generic, like a single function in the server config. This could make some use cases more difficult though. It might be good to have both possibilities?

Pro:
Hmmmm, @jpfr is right about the overhead. It also makes the code more complicated. Probably it's best if we first only use a counter. @Root-Core sorry for the additional request, but could you change it from the list to counter? Thank you very much!
If we really need the list in the future we can still add it.

Root-Core:
@Pro Done. This solution is way easier to implement, but I learned some things the other way. :)

We may discuss the API as we use a single callback - separated by the deleted flag - at the moment. It might be cleaner, if it was separated into two functions (`onMonitored` / `onUnmonitored` or something like this). Just asking for your opinion, as the current solution is working for me too.

Also we could add an callback to the server config (?) as we are counting on all nodes right now, which enables us to trigger an callback on all node types via the server (config) object with the node or id as parameter. The data source should keep a separate callback nevertheless.

include/ua_plugin_nodestore.h
```diff
@@ -81,7 +81,9 @@ typedef struct {
     UA_NodeReferenceKind *references;           \
                                                 \
     /* Members specific to open62541 */         \
-    void *context;
+    void *context;                              \
+    UA_UInt16 monCounter;
+    // FIXME: Should this list only be included, if UA_ENABLE_SUBSCRIPTIONS is active?
```
Pro: yes

Pro:
And rebase on master to include recent changes, Thanks!

src/server/ua_subscription.c
```diff
@@ -75,15 +75,37 @@ UA_Subscription_deleteMonitoredItem(UA_Server *server, UA_Subscription *sub,
...
     UA_MonitoredItem *mon;
-    LIST_FOREACH(mon, &sub->monitoredItems, listEntry) {
+    LIST_FOREACH(mon, &sub->monitoredItems, listEntry_store) {
         if(mon->itemId == monitoredItemId)
             break;
     }
     if(!mon)
         return UA_STATUSCODE_BADMONITOREDITEMIDINVALID;
 
+    /* Find the Node, to increase its monitored counter */
```
jpfr: ... decrease the monitored counter

include/ua_plugin_nodestore.h
```diff
@@ -70,7 +70,7 @@ typedef struct {
     UA_ExpandedNodeId *targetIds;
 } UA_NodeReferenceKind;
 
-#define UA_NODE_BASEATTRIBUTES                  \
+#define __UA_NODE_BASEATTRIBUTES                \
```
jpfr: The counter is added to all nodes.
But only variable nodes with a datasource are considered for the callback.

- The counter and the callback should appear always in pairs.
- In the future we might want to use the same mechanism for non-variable nodes as well. For example for event monitoring on objects or for other node attributes (e.g. user access rights that are checked with other systems in the background)

Can we put counter + callback in every node?
And count any type of monitoreditem, not only the value.

include/ua_plugin_nodestore.h
```diff
@@ -83,6 +83,14 @@ typedef struct {
     /* Members specific to open62541 */         \
     void *context;
 
+/* Add counter only if subscriptions are enabled */
+#ifdef UA_ENABLE_SUBSCRIPTIONS
```
jpfr: I find it a bit more readable to ifdef-define a macro `UA_NODE_BASEATTRIBUTES_SUBSCRIPTIONS` first and then include that in `UA_NODE_BASEATTRIBUTES`.

include/ua_server.h
```diff
@@ -695,6 +699,20 @@ typedef struct {
...
+     *
+     * @param server Allows the access to the server object
+     * @param sessionId The session id, represented as an node id
+     * @param sessionContext An optional pointer to user-defined data for the specific data source
+     * @param nodeid Id of the node in question
+     * @param removed Determines, if the monitoring on the node was removed or created.
+     * @return Returns a status code for logging. It could be used to determine,
+     *         if the subscription was successfull and might be returned to the user. */
+    UA_StatusCode (*monitored)(UA_Server *server, const UA_NodeId *sessionId,
+                              void *sessionContext, const UA_NodeId *nodeId,
```
jpfr: indentation

include/ua_server.h
```diff
@@ -695,6 +699,20 @@ typedef struct {
...
+
+    /* Triggered, if the the node gets (un)monitored by a subscription.
+     *
+     * @param server Allows the access to the server object
+     * @param sessionId The session id, represented as an node id
+     * @param sessionContext An optional pointer to user-defined data for the specific data source
+     * @param nodeid Id of the node in question
+     * @param removed Determines, if the monitoring on the node was removed or created.
+     * @return Returns a status code for logging. It could be used to determine,
+     *         if the subscription was successfull and might be returned to the user. */
```
jpfr: The return value is not actually logged even if it indicates an error.
Either add the logging or make it a void return value.

Imho, an error code can also be returned with the notification values once the monitoreditems runs.
And removing a subscription should never fail. We cannot prevent clients from disconnecting.

Root-Core: I added FIXMEs, where the callback returns. I was not sure, if it is okay to modify the result. 
If an hardware failure or something like this happens, it would be nice to be able to prevent the subscription in the first place.

You say the unsubscribe return value should not be modified, but it could be by `UA_Nodestore_getCopy()`. Do you consider this to be a valid change?

include/ua_plugin_nodestore.h
Root-Core: I thought about adding a more generic callback to the server config. It could be triggered on every subscription with the node id or the node itself as identification.. the data source should keep that "opt in" callback tough.

Root-Core:
Thank you for the feedback. 👍 

Open points:
1. Shall we use the return value of the callback?
2. How to handle non data source subscriptions?

To 1: https://github.com/open62541/open62541/pull/603#discussion_r170456508
To 2: https://github.com/open62541/open62541/pull/603#discussion_r170456633

jpfr:
> 1.  Shall we use the return value of the callback?

I'm fine with either. If you don't want to take a decision, then make it a void return value.
The user can add logging code inside the callback.

> 2.  How to handle non data source subscriptions?

You write

> I thought about adding a more generic callback to the server config. It could be triggered on every subscription with the node id or the node itself as identification..

This is even better!

- Remove the counter from the basenode (can be handled by the user in the node context).
- Move the `monitored` callback to the server config.
- Simplify the code by calling the `monitored` callback always. Not just on variable+datasource.

> the data source should keep that "opt in" callback tough.

As always, I'm on a crusade for simplicity. So I want to say no here.
But I can be convinced otherwise.
Is there a use case that cannot be covered by the above approach? If yes, please describe.

If you want an additional callback for some nodes (e.g. with a datasource), you can put that into the node context. Or specialize inside the `monitored` callback based on the NodeId.

jpfr:
Hello @Root-Core ,

any news?
We can also schedule a short conference call to get everybodys ideas.
Maybe with @Pro in the loop.

Root-Core:
I was short on time lately, not working on my default machine. I would use the status code as feedback. There might be some reason, why we can not start the subscription. Not sure about stopping (unsubscribe).

If we decide to make the counting into user space, we might be able to count not only value attribute subscriptions. While this is nice, it is less convenient. It seems to be more reusable though and therefore preferable, I guess.

The data source sounds to be more "complete" with the additional callback, as it reminds me of some kind of hardware or software feature, that should know if it needs to be 'alerted'. But I'm fine with the server callback solution as it is simpler, code wise for the stack. It could add some overhead to the "user side" code, but the most use cases should be coverable.

jpfr:
Ok, cool @Root-Core !

With this, the PR only adds API but does not break any.
Let's merge after the first release candidate. This buys us some time.

Pro:
@Root-Core any update on this?

Root-Core:
@jpfr I've refactored the whole thing again to only call the monitored callback of the server config.

The builds in travis / appveyor are failing though, which don't seem to be related with my changes. There are no useful outputs in the build log. Tests are passing. I can't figure it out right now.

jpfr:
Cool! Good to see you are still working on it.
I'll have a look when I get a free time slot.

Let's target master for now. We had an API freeze for 0.3.

Root-Core:
@jpfr Yes, sorry for not working on things lately.
With the recent changes, it not really alters the API beyond adding one variable to the server config (which can be left `NULL`, which is default).

The reason for switching to `master` was that I haven't seen @Pro changed it to `0.3`. I've pulled `master` and wondered why it would not be automatically mergeable after commiting. I ended up, changing it back to master. It seems to confuse CI for no obv. reason, sorry for that.

Pro:
@jpfr I merged master and fixed some compile issues. CI is all green again. We IMHO we can merge this.

examples/server_ctt.c
```diff
@@ -100,6 +100,43 @@ outargMethod(UA_Server *server,
...
+    if (removed && *monCount > 0) {
+        (*monCount)--;
+    } else if (!removed) {
+        (*monCount)++;
+    }
+    UA_Server_setNodeContext(server, *nodeId, (void*)monCount);
+
+    /* Check, if the node was monitored the first time or not monitored anymore */
+    if (!removed && *monCount == 1) {
+        // assure the nodeid is numeric (for logging in this example)
```
jpfr: Use C comments (/* */) instead of C++ comments (//)

examples/tutorial_server_datasource.c
```diff
@@ -130,6 +130,46 @@ writeCurrentTime(UA_Server *server,
     return UA_STATUSCODE_BADINTERNALERROR;
 }
 
+
```
jpfr: Spurious newline

examples/tutorial_server_datasource.c
```diff
@@ -130,6 +130,46 @@ writeCurrentTime(UA_Server *server,
...
+    /* Check, if the node was monitored the first time or not monitored anymore */
+    if (!removed && *monCount == 1) {
+        // assure the nodeid is a string (for logging in this example)
+        if (nodeId->identifierType == UA_NODEIDTYPE_STRING)
+            UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND, "Started monitoring on Node %.*s",
+                        (int)nodeId->identifier.string.length, nodeId->identifier.string.data);
+    }
+    else if (removed && *monCount == 0) {
+        // assure the nodeid is a string (for logging in this example)
```
jpfr: Use C comments (/* */) instead of C++ comments (//)

include/ua_plugin_nodestore.h
```diff
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
jpfr: Move to a dedicated PR with cosmetic changes.

include/ua_server.h
```diff
@@ -681,13 +682,14 @@ typedef struct {
      * `value->value.storageType` to `UA_VARIANT_DATA_NODELETE` to prevent the
      * memory being cleaned up. Don't forget to also set `value->hasValue` to
      * true to indicate the presence of a value.
-     * 
+     *
```
jpfr: The entire file should be moved to a dedicated PR with cosmetic changes.

include/ua_server.h
```diff
@@ -8,6 +8,7 @@
  *    Copyright 2015-2016 (c) Chris Iatrou
  *    Copyright 2015-2016 (c) Oleksiy Vasylyev
  *    Copyright 2016-2017 (c) Stefan Profanter, fortiss GmbH
+ *    Copyright 2018 (c) Fabian Arndt, Root-Core
```
jpfr: Add your author tag to all files with relevant changes.

include/ua_server_config.h
```diff
@@ -167,11 +167,31 @@ struct UA_ServerConfig {
...
+                                          const UA_NodeId *sessionId, void *sessionContext,
+                                          const UA_NodeId *nodeId, void *nodeContext,
+                                          const UA_UInt32 attibuteId, const UA_Boolean removed);
+#endif
+
     /* Historical Access */
 #ifdef UA_ENABLE_HISTORIZING
     UA_Boolean accessHistoryDataCapability;
     UA_UInt32  maxReturnDataValues; /* 0 -> unlimited size */
-
```
jpfr: Move cosmetic changes to a separate PR.

src/client/ua_client_internal.h
```diff
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
jpfr: Move cosmetic changes to a separate PR.

src/server/ua_server_internal.h
```diff
@@ -177,10 +177,10 @@ struct UA_Server {
 #define UA_Nodestore_get(SERVER, NODEID)                                \
     (SERVER)->config.nodestore.getNode((SERVER)->config.nodestore.context, NODEID)
 
-#define UA_Nodestore_release(SERVER, NODEID)                            \
-    (SERVER)->config.nodestore.releaseNode((SERVER)->config.nodestore.context, NODEID)
```
jpfr: Changes are not used for the PR. Please move to a separate PR.

src/server/ua_server.c
```diff
@@ -128,8 +128,8 @@ void UA_Server_delete(UA_Server *server) {
 
 #ifdef UA_ENABLE_SUBSCRIPTIONS
     UA_MonitoredItem *mon, *mon_tmp;
-    LIST_FOREACH_SAFE(mon, &server->localMonitoredItems, listEntry, mon_tmp) {
```
jpfr: Revert renaming to listEntry_store.
This adds only noise to the PR.

This can be added to a "cosmetic changes" PR.

jpfr:
By the way, `git rebase -i` and `git add -p` are your friends for cleaning up PR.

Pro:
>  By the way, git rebase -i and git add -p are your friends for cleaning up PR.

I wanted to avoid a force push since I do not know if @Root-Core is still working on it, where force pushes are a bad idea. For these cases I rather prefer squash-merges.

@Root-Core can you address the comments by @jpfr? Shouldn't be that much of remaining work

Root-Core:
@jpfr The most noise is due to the commit history here, a lot of it actually had some real use. As the PR further progressed due to the input given in the PR, it was abandoned. :+1: 

@Pro Thank you. I think the broken part about my last commit was the wrong revision of the ua_nodeset sub module. I presumably could easily have fixed it, but never had the time to do it myself. 

I try to look into it tomorrow. This PR doesn't really change the public API, as it only adds an optional pointer to the server config. So should this be part of  `0.3` as targeted first or `master`?

jpfr:
Cool!

> So should this be part of 0.3 as targeted first or master?

Let's merge it to master first.

Any specific reason why you want it in 0.3?
We should have a consensus among all core devs if we want to bring it in after the API freeze.

jpfr:
Hey @Root-Core ,

can you give as write-permissions to the branch in you repository?
We can then do a rebase and merge.

It would be a pity to have this complete feature bit-rot.

Pro:
@jpfr I already have write access to @Root-Core's branch.

I quickly checked if the branch can be easily rebased, but it results in many conflicts, so probably it's easier if we just squash-merge on github. Otherwise you can try to rebase and push to a separate branch, and I will force push that to @Root-Core repo.

jpfr:
Ok. Let's do that.

The PR is good.
But there is a lot of noise in the history and non-functional changes that are not needed any more.

Pro:
Ok, I leave it up to you to merge it, if you still want to address your review comments. Otherwise you can directly merge it.

jpfr:
The clean version is in #1954.

jpfr:
Merged via #1954.

Issue #576, opened 2016-04-08
dynamic memory constantly increasing on Server-side with every new Client.


NetoLe:
I am using an embedded device with an ARM-architecture and my open62541-Stack is included in an C++-Framework. It is working so far. But there is a small restriction I experienced. 

I created a "debug-node" to which I can subscribe and which shows me the used dynamic memory-size (using tlsf for allocation etc.) and gets me an overview. I wrote a "debugging-client" with the open62541-stack too with which I can subscribe to that node and get me the used memory output if its changing (which works with UaExpert too very well now). 

And third I have another MFC-using-Client with the open62541-Stack which subscribes to all the other relevant data of the embedded device.

Then I saw something unusual. Everytime I connect to the server on the embedded device, the memory is increasing - ok, that's normal, monitored items are created etc. But when I disconnect it keeps an offset of ~200 Byte. So the more Clients I connect and disconnect, the more dynamic memory is used.

The memory pool itself is not a global pool but a pool just for the Stack-working stuff. I wrapped the networklayer-functions of your tcp_networklayer (great work btw) to functions, which I am using in the device, which implicates all the mallocs and frees are UA-mallocs etc. too, using the same memory-pool. 

And I am not using multithreading.

Stasik0:
Sounds like a memleak, are you able to run your your client in valgrind with leak-check=full to show where exactly its leaking?

NetoLe:
I can't run valgrind (using windows).

But I figured out, that the problem are the sockets. they are not free'd after disconnecting
`static void socket_close()` Isn't called once. I found this line: `/* only "shutdown" here. This triggers the select, where the socket is "closed" in the mainloop */´ which seems it doesn't for me.

NetoLe:
ok, this relates to https://github.com/open62541/open62541/issues/517 .

It seems the networklayer in my device does not work as expected. When the client sends a TCP with FIN,ACK my device just answers ACK which results in a RST, ACK after 2 min, but that doesn't reset the ports. I have to search the problem on my device.

Issue #517, opened 2016-01-22
open62541-server does not close session with UAExpert.


NetoLe:
Ok maybe the title is not totally correct because I can't locate the problem for that.

What happens:
If I want to connect to an open62541-Server with UaExpert 1.4.1 it takes 2 min till the connection is established (while that time UaExpert just freezes). once this happened I can close and open my server as often as I want, it does immediately reconnect. If I want to disconnect it takes 2 min too!
This problem didn't occure with UaExpert 1.3.1, where it did immediately connect to the server and disconnect.
So I took a wireshark trace and saw that it took exactly 2 min till the connection worked. 

[uaexpert2minprob.zip](https://github.com/open62541/open62541/files/100763/uaexpert2minprob.zip)
(192.168.60.111 is client, ....60.8 is Server)
So I looked in the UaExpert configuration for a number with "120" in it. I found the session timeout which had 1200000,00 as value.
But changing this to a lower number didn't change the problem. it are still exactly 2 minutes and I can't locate which member is the problem here. Is this an UaExpert-problem?

jpfr:
Hey @NetoLe,

I can't reproduce the behavior with UAExpoer 1.4.1 on my box.
The messages on wireshark look clean too.
However there are some artifacts on the TCP level. Reordered packets, duplicate ACK messages, and so on.

Is this behavior reproducible?

NetoLe:
Hey @jpfr 
so I am using the open62541-Server on an embedded device and PC is linked over normal ethernet with it, there it is always reproducible.
I tried to run a PC-Application, so it's just localhost to localhost. What works is the Port 16664 which is given with the tutorial without any problems. What doesn't work is the Port 4840, which I am using on the embedded device too. The server immediately crashes with "Error during socket binding".
At that point UaExpert is not connected to the embedded device anymore.
So I looked up in cmd with "netstat -o" but the port isn't used at all. 
that is really weird. any ideas what I could try?

NetoLe:
oh, sorry, I should've used "netstat -ao"

There is a Service using this Port opcualds.exe "UA Local Discovery Server Windows Service".
Ok, what is this?

NetoLe:
so I installed 1.3.1 again and made a capture. It seems like the problem is caused by UaExpert itself?
because with 1.3.1 there is an immediately SYN, even before the ACK for the FIN+ACK comes. 
[...]

PR #605, opened 2016-05-10
fixing networklayer_tcp select()

NetoLe:
fixed socket-overflow on some devices

This is made regarding an embedded device with proprietary
networklayer-functions. It is geared to socket-functions from the
Unix-man page of select(), which says that it only counts the events
which are watched. If no error-event is watched it won't be counted,
which leads to open sockets even if they should be closed because they
were shutdown'd before, because the select isn't triggered because it
doesn't watch for errors, when the parameter for the error-set is NULL.
It now works fine and closes the specific sockets as it should. Sureley
this isn't a noticable problem on big desktop-machines with a lot of
sockets but if you are on an embedded device reduced to 30 or something
you will experience a run out of free sockets very fast after some
connects and disconnects. This fixes prior problems regarding conflicts
for those who had problems with UaExpert 1.4.1 and newer. This relates
to #576 and #517

### DIFF
--- a/plugins/networklayer_tcp.c
+++ b/plugins/networklayer_tcp.c
@@ -391,11 +391,12 @@ ServerNetworkLayerTCP_start(UA_ServerNetworkLayer *nl, UA_Logger logger) {
 static size_t
 ServerNetworkLayerTCP_getJobs(UA_ServerNetworkLayer *nl, UA_Job **jobs, UA_UInt16 timeout) {
     ServerNetworkLayerTCP *layer = nl->handle;
-    fd_set fdset;
+    fd_set fdset, errset;
     UA_Int32 highestfd = setFDSet(layer, &fdset);
+    setFDSet(layer, &errset);
     struct timeval tmptv = {0, timeout * 1000};
     UA_Int32 resultsize;
-    resultsize = select(highestfd+1, &fdset, NULL, NULL, &tmptv);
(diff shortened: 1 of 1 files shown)


plugins/networklayer_tcp.c
```diff
@@ -426,8 +427,10 @@ ServerNetworkLayerTCP_getJobs(UA_ServerNetworkLayer *nl, UA_Job **jobs, UA_UInt1
     size_t j = 0;
     UA_ByteString buf = UA_BYTESTRING_NULL;
     for(size_t i = 0; i < layer->mappingsSize && j < (size_t)resultsize; i++) {
-        if(!UA_fd_isset(layer->mappings[i].sockfd, &fdset))
-            continue;
+        if(UA_fd_isset(layer->mappings[i].sockfd, &errset)) {}
```
jpfr: @NetoLe Can you explain line 430?
If the errset is set for the socket, we want to continue (and close the socket)?
I would fold line 430 and line 431 into a single if. The empty brackets can be overseen.

Stasik0: @jpfr I think the idea is, to call socket_recv on an 'errored' socket, to receive an error message and to free the socket starting from line 440

NetoLe: yep, that's right. changed the line.

jpfr:
Great, thanks for the PR @NetoLe !

PR #606, opened 2016-05-11
Added callback to authentification

Root-Core:
This callback might help you to integrate the OPC UA-Server into an existing enviroment.

It provides the entered username, password and the endpoint, that requested the authentification.
You provide an UA_Boolean to allow / deny the authentification.

If the standard server configuration is used as origin of your own configuration, there shouldn't be an issue. **If not, the configs member .authCallback should be set to NULL.**


Stasik0:
good idea, can we fold the current lists of usenames/passwords into a 'standard' callback? We also need a triage - ist this feature in 0.2 or post 0.2 @open62541/open62541-core

Root-Core:
Sorry for all this commits.. i just can't test it, as it compiles on all my systems, but not on Travis.
I can't figure out, what is going wrong right now..

In my opinion, we do not need the current list, as it is more difficult to maintain it, then it is for your own method / callback.

Stasik0:
yea, exactly, i meant to wrap the list into a "standard" auth-callback with the same functionality

jpfr:
Yes! Great idea.
I see the authentication closely related to access rights management.
Maybe the authentication can return a session context (void*) that is then used to determine access rights for read/write/call, ...

I suggest we take this PR as the basis to develop a plugin mechanism for rights management pos 0.2.

cmorty:
@Root-Core You can just activate Travis for your own Repo. No need to make a pull-request. It should work out of the box without any configuration.

Pro:
@Root-Core can you try to merge master into this PR so we have a clean basis to continue working on this? Thank you very much!

Root-Core:
@Pro As there is some plugin called `access control`, I am not sure if this is important any more?
If you could reuse some of this code here, I would superseed the PR, as there are too many conflicts by now.

The only thing missing plugin wise would be the access to the IP address, as everything else should fit well into the whole plugin theme, I guess.

Root-Core:
@Pro Does closing this PR mean, that there is no need for merging?

Pro:
You are right, that the current access control plugin already includes the authentication callback (i.e., similar to your first two commits).
See https://github.com/open62541/open62541/blob/394df3be8d6a19781927f0790d9c2a9142a2d6a8/plugins/ua_accesscontrol_default.c#L29

As far as I can see, the rest of your commits is already (somehow) fixed in the master.
But thank you a lot for your valuable contribution!

Are you missing a specific feature which is in your PR and not currently in master?

jpfr:
Thanks @Root-Core !
Your PR was the inspiration for the access control layer.
Users can now attach data to the authenticated session and use that for fine-grained access control.

Root-Core:
@Pro, @jpfr Thank you guys for your amazing work here.

The only thing missing is an easy access to the clients ip address, as that allows detailed logging or handling of specific ip ranges etc. This might be a bit specific, but I could definitely think of some use cases.

jpfr:
Hey @Root-Core ,

the UA_Connection struct contains the socket id.
With that, you can get the name of the remote connection via getpeername.

https://linux.die.net/man/3/getpeername

We currently write only the IP address of new remote connections into the log. But not the peername.
In local networks without DNS the peername lookup can lead to serious timeouts...

jpfr:
But, you might have to use an internal API to get from session->channel->connection->ip

Root-Core:
Okay, I will look into it, if I find the time. I have been on to other projects the last months, but I might be able to figure out, how to get the session object and its channel. The peer name is not be necessary, the `getpeername()` call was actually a try to get the ip address, which seems to be easier with the internal ip.

Thank you for your responses I really appreciate that. 👍

jpfr:
On master there is a method to retrieve session objects by their identifier nodeid.
You get that in the callback.
All you have to do is include the internal session-manager header.

Good luck!

PR #613, opened 2016-05-18
fixes for call service

FlorianPalm:
-it is now possible to pass enumerations as argument of a call
-method nodes can be interconnected with any subtype of hasComponent reference


jpfr:
Rebased and merged.

Issue #622, opened 2016-05-31
Fails to build on centos6


otronarp:
I'm trying to build the 0.2 branch on centos6 with gcc 4.4.7 and it fails with
```
/usr/bin/gcc  -DUA_DYNAMIC_LINKING -DUA_NO_AMALGAMATION -DVERSION=v0.2.0-RC1-27-g3583b4e -I/home/jenkins/open62541/b/src_generated -I/home/jenkins/open62541/include -I/home/jenkins/open62541/src -I/home/jenkins/open62541/plugins -I/home/jenkins/open62541/deps  -O3 -DNDEBUG   -std=c99 -pipe -Wall -Wextra -Werror -Wformat -Wno-unused-parameter -Wno-unused-function -Wno-unused-label -Wpointer-arith -Wreturn-type -Wsign-compare -Wmultichar -Wstrict-overflow -Wcast-qual -Wmissing-prototypes -Wstrict-prototypes -Winit-self -Wuninitialized -Wformat-security -Wformat-nonliteral -Wshadow -Wconversion -fvisibility=hidden -fPIC -ffunction-sections -fdata-sections -fno-stack-protector -fno-unwind-tables -fno-asynchronous-unwind-tables -fno-math-errno -fmerge-all-constants -fno-ident -o CMakeFiles/open62541-object.dir/src/ua_types.c.o   -c /home/jenkins/open62541/src/ua_types.c
In file included from /home/jenkins/open62541/src/ua_util.h:4,
                 from /home/jenkins/open62541/src/ua_types.c:1:
/home/jenkins/open62541/b/src_generated/ua_config.h:179:3: error: #warning Slow Float Encoding
make[2]: *** [CMakeFiles/open62541-object.dir/src/ua_types.c.o] Error 1
```

I can get it to compile by removing the `-Werror`, but the Slow float encoding warning doesn't sound good. How can avoid that with this gcc version?

jpfr:
This warning can be removed safely. It is only an indicator that the float encoding could not be detected.

The warning message was improved in c3014dd38071c379e720b00ace1501e66bbc7195.

When you that your system is little-endian, you can just force `UA_BINARY_OVERLAYABLE_FLOAT` to true. Normally, gcc should detect this alright however...

PR #623, opened 2016-06-01
Gcc44 centos6

otronarp:
This is what I needed to be able to compile with -Werror on centos6 with gcc 4.4 (see issue #622).


jpfr:
Looks good! Thanks for the PR.

PR #624, opened 2016-06-03
changing session id to random GUID, updated logging of session ids

Stasik0:
this fixes some tests, since session id do not collide with nodes in NS0

### DIFF
--- a/src/server/ua_session_manager.c
+++ b/src/server/ua_session_manager.c
@@ -1,12 +1,9 @@
 #include "ua_session_manager.h"
 #include "ua_server_internal.h"
 
-#define STARTSESSIONID 1
-
 UA_StatusCode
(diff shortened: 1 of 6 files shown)


src/server/ua_session_manager.c
```diff
@@ -72,7 +69,7 @@ UA_SessionManager_createSession(UA_SessionManager *sm, UA_SecureChannel *channel
 
     sm->currentSessionCount++;
     UA_Session_init(&newentry->session);
-    newentry->session.sessionId = UA_NODEID_NUMERIC(1, sm->lastSessionId++);
+    newentry->session.sessionId = UA_NODEID_GUID(1, UA_Guid_random());
```
Stasik0: maybe going for guids in ns=0?

jpfr: nah. ns1 is good. those sessions are not standard-defined...

PR #635, opened 2016-06-14
One way reference ofr external namespaces

LEvertz:
Ran into a Problem when trying to bridge from NS0 to an external NS with a Reference. The referenced node in the external NS is obviously not found by findNode (called from editNode, called from addReferences_Single).

The standard says, it is ok for references to external nodes when only the source of the reference is edited in the local server. As my problem occurred on an external NS, not on real external nodes, this does not fully apply, but at least it is close. The fix I included in this pull request does its job quite well in first tests, but I'm happy to learn a better alternative.

### DIFF
--- a/include/ua_server_external_ns.h
+++ b/include/ua_server_external_ns.h
@@ -42,6 +42,9 @@ typedef UA_Int32 (*UA_ExternalNodeStore_addReferences)
 (void *ensHandle, const UA_RequestHeader *requestHeader, UA_AddReferencesItem* referencesToAdd,
  UA_UInt32 *indices,UA_UInt32 indicesSize, UA_StatusCode *addReferencesResults,
  UA_DiagnosticInfo *diagnosticInfos);
+ 
+ typedef UA_Int32 (*UA_ExternalNodeStore_addOneWayReference)
+(void *ensHandle, const UA_AddReferencesItem *item);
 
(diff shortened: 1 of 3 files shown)


Stasik0:
the first commit is not relevant, you can just call
`git revert 5d9838f`
on the branch and push

Stasik0:
@LEvertz is it possible to disable the check only for namespace ids that are in external namespace array of the server?

LEvertz:
@Stasik0 yep, it is. But it needs a for-loop then which takes time. If this doesn't matter I'll put that in.

Stasik0:
@LEvertz since it happens only on node-addition, I think it is not a big performance-problem

LEvertz:
OK then. Will include that this afternoon.

 

Von: Sten Grüner [mailto:notifications@github.com] 
Gesendet: Mittwoch, 15. Juni 2016 10:52
An: open62541/open62541
Cc: Evertz, Lars; Mention
Betreff: Re: [open62541/open62541] One way reference ofr external namespaces (#635)

 

@LEvertz <https://github.com/LEvertz>  since it happens only on node-addition, I think it is not a big performance-problem

LEvertz:
Looked over it again:
we will need onw-way references to and from external nodestores anyways. So I will add an addOneWayReference function to the external nodestore interface and commit that to this pull request.

Issue #509, opened 2016-01-14
Verify tutorials (relates #508)


Stasik0:
Before freezing 0.2 we need to make sure tutorials are aligned with current api

ichrispa:
Hmmm... I have two "volunteers" that can be assigned this task (with ETA mid february). Would that be sufficient?

Stasik0:
Yea,  but first we need to be sure that the api is stable

Pro:
See also: PR https://github.com/open62541/open62541/pull/516
I'll go through the tutorials and try to fix any errors when I see them using the current version.

Stasik0:
Can we actually include .c files in sphinx? Kinda, "include xyz.c:3-15"? Would be much better than double maintenence.

jpfr:
Totally.
Maybe we can write large parts of the documentation as comments in tutorial c-files and in the public header files. You know, take the code, detect comments in `/** */` double-star comments and wrap everything else in code blocks.

Otherwise we will run into impedence mismatches between code and docs. At least between releases.

Stasik0:
isn't there some sphinx solution?

FlorianPalm:
We would also need Doxygen, wouldn't we?  [Look here](http://stackoverflow.com/questions/11246637/using-sphinx-apidoc-to-generate-documentation-from-c-code)

Stasik0:
I thought about "including" parts of source code into tutorial articles which is, imho, not doxygen anymore.

memaikey:
First steps with open62541-client tutorial doesn't compile.
(trying to upload fix, but new to git)

memaikey:
Tutorial "Adding nodes to a server and connecting nodes to user-defined values" mentions 3 cases under  "Variants", but only lists 2.

Stasik0:
@memaikey its even worth: the listed items do not match the values in the struct: the truth is in the code: https://github.com/open62541/open62541/blob/38d4e7fbad8d11826521cecda630046ffa20035d/include/ua_types.h#L535

Stasik0:
@memaikey did you check all the tutorials?

memaikey:
@Stasik0 I finished toturial 1&2 and just made a pull request for my latest updates. I've read tutorial 3&4, but they basically say "go look at the code". I am currently testing these two. I don't think I'll check the final tutorial in a while.

schmidmaxa:
[...]

PR #641, opened 2016-06-17
Fix Tutorial_client_firstSteps

memaikey:
Eddited client tutorial, basic client exmpale did not compile
- Changed dir sturcture to be sorted correctly
- Added include to basic client example
- Removed Logger_Stdout parameter from UA_Client_new
- Removed UA_ClientCOnnectionTCP from UA_Client_connect


Stasik0:
@memaikey this is good, have you tried another tutorials or only this one?

memaikey:
@Stasik0 I did the server and client tutorial. I will probably do the others to and report any strange behavior I find.

Stasik0:
Very good, please tell us in #509 when you are done

PR #643, opened 2016-06-20
Fix "First steps with open62541-client" tutorial

memaikey:
* Added 2 headers
* Commented out 1 header
* Changed 2 parameters in client creation
* Unified space usage


jpfr:
Thank you for the improvements!

Issue #647, opened 2016-06-23
Subscription with UA_STATUSCODE_GOODLOCALOVERRIDE status


StalderT:
Hello,

I added
 
```
    value->hasStatus = UA_TRUE;
    value->status = UA_STATUSCODE_GOODLOCALOVERRIDE;

```

in readTimeData function in server.c

In UA_Expert, I can read the attributes : 

![opc2](https://cloud.githubusercontent.com/assets/13799456/16300453/4f234898-3940-11e6-89a5-705595ff19d2.PNG)

But if I try to monitor the node I have the error : 

Item [XXX] has invalid MonitoredItemId 0

![opc3](https://cloud.githubusercontent.com/assets/13799456/16300465/5e7f930a-3940-11e6-9274-e09502aa75dc.PNG)

PR #648, opened 2016-06-23
Update ua_services_subscription.c

StalderT:
fix https://github.com/open62541/open62541/issues/647


Stasik0:
I am not sure what to do here, since we do not discriminate between the status of the actual datavalue and the internal service Service_Read_single, is it okay for everybody without any check @open62541/open62541-core? A trivial idea would be to check for any UA_STATUSCODE_GOOD* and let the values pass?

StalderT:
I make some more tests and this PR is not a good idea. I think add a UA_STATUSCODE_GOOD*  check is better. I close this PR.

StalderT:
something like that is better : 

```
    if(v.hasStatus != UA_STATUSCODE_GOOD) {
        result->statusCode = v.status;
        if (result->statusCode & 0xFF000000) {
            UA_DataValue_deleteMembers(&v);
            return;
        }   
    }   
    UA_DataValue_deleteMembers(&v);

```

Stasik0:
@StalderT will you add a new pr?

PR #649, opened 2016-06-23
Fix tutorial

memaikey:


jpfr:
Great! Can you remove the commented-out includes completely?
I'll then do a squash-merge so only a single commit appears in the git history.

memaikey:
Will do, pull request coming right up. 
I don't know if I can safely remove it from tutorial_noderelations.rst, so I left it there.

PR #653, opened 2016-06-23
adding a separate switch to enable thread_local storage without getti…

Stasik0:
…ng lock-free datastructures, e.g. for creating multi-threaded clients


jpfr:
Always activated in #654

PR #669, opened 2016-07-06
Use standard python libraries for logging and argparsing in namespace compiler

markusgraube:
I have rewritten the namespace compiler in order to use the standard python libraries logging and argparse.

Now, it is possible to specify on the command line which log message should be printed. Furthermore, the code looks (hopefully) more clean.

However, this raises the dependencies of python to version 2.7


jpfr:
`/CMakeLists.txt` still checks for logger.py, hence the build fails.
Otherwise, this looks really good!

jpfr:
```
  File "C:/projects/open62541/tools/pyUANamespace/generate_open62541CCode.py", line 153, in <module>
    ns.allocateVariables()
  File "C:\projects\open62541\tools\pyUANamespace\ua_namespace.py", line 434, in allocateVariables
    n.allocateValue()
  File "C:\projects\open62541\tools\pyUANamespace\ua_node_types.py", line 1005, in allocateValue
    self.value().parseXML(self.__xmlValueDef__)
  File "C:\projects\open62541\tools\pyUANamespace\ua_builtin_types.py", line 171, in parseXML
    log(self, "parsing xmlvalue for " + self.parent.browseName() + " (" + str(self.parent.id()) + ") according to " + str(self.parent.dataType().target().getEncoding()))
NameError: global name 'log' is not defined
```

Maybe you just build with `UA_BUILD_EXAMPLES` enabled on your side. There's still something going on in the namespace generation.

jpfr:
Now it fails with some utf8 problems.
Can you just monitor the travis and appveyor results yourself?

jpfr:
thank you! merged.

PR #674, opened 2016-07-11
Bugfixes

Torben-D:
-Delete in case of SequenceNumberCheck failt
-"void UA_Client_Subscriptions_forceDelete(UA_Client *client, UA_Client_Subscription *sub);" the 
UA_Client_Subscription only exist if Subscription are enabled

### DIFF
--- a/src/client/ua_client_internal.h
+++ b/src/client/ua_client_internal.h
@@ -40,6 +40,7 @@ typedef struct UA_Client_Subscription_s {
     LIST_HEAD(UA_ListOfClientMonitoredItems, UA_Client_MonitoredItem_s) MonitoredItems;
 } UA_Client_Subscription;
 
+void UA_Client_Subscriptions_forceDelete(UA_Client *client, UA_Client_Subscription *sub);
 #endif
 
 /**********/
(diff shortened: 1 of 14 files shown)


src/server/ua_server_binary.c
```diff
@@ -552,6 +552,9 @@ processMSG(UA_Connection *connection, UA_Server *server, const UA_TcpMessageHead
                             sequenceHeader.sequenceNumber, channel->receiveSequenceNumber + 1);
         sendError(channel, msg, *offset, &UA_TYPES[UA_TYPES_SERVICEFAULT],
                   sequenceHeader.requestId, UA_STATUSCODE_BADSECURITYCHECKSFAILED);
+        Service_CloseSecureChannel(server, channel);
+		connection->close(connection);
```
Stasik0: tabs

src/client/ua_client_internal.h
```diff
@@ -40,6 +40,7 @@ typedef struct UA_Client_Subscription_s {
     LIST_HEAD(UA_ListOfClientMonitoredItems, UA_Client_MonitoredItem_s) MonitoredItems;
 } UA_Client_Subscription;
 
+void UA_Client_Subscriptions_forceDelete(UA_Client *client, UA_Client_Subscription *sub);
```
jpfr: Sorry, but no.
That will fail when subscriptions are disabled.

Why are you moving `UA_Client_Subscriptions_forceDelete` into the ifdef for subscrptions?
Why move the definition at all?

jpfr: please ignore my previous comment o_O

Issue #675, opened 2016-07-12
Local Discovery Server


Pro:
We need the concept of a centralized LDS (Local Discovery Server) (see http://www.automation.com/automation-news/article/how-opc-ua-clients-discover-servers-part-3)

Does open62541 support this kind of setup?

The required methods for this are:
- register server
- find servers
- get endpoints

It looks like the last two are already included in open62541:
* https://github.com/open62541/open62541/blob/0.2/src/server/ua_services_discovery.c
* https://github.com/open62541/open62541/blob/0.2/src/client/ua_client.c#L378

I couldn't find the `register server`. Is there any plan to support LDS?

I would be happy to help you add the functionality by creating some example code. Just let me know how you would approach this.

ichrispa:
Hi @Pro,

I am currently using a custom UDP broadcast mechanism to implement server discorvery. Nice to see someone's taking up the job of implementing the spec-conformant discovery service.

Best regards,

ichrispa

Pro:
The first simple implementation of an LDS supporting `RegisterServer` and `FindServers` is in the current master branch. Not yet supported (but planned) is LDS-ME and GDS.

PR #677, opened 2016-07-12
Service RegisterServer and Discovery

Pro:
The current implementation is missing the RegisterServer service for Local Discovery Server (see #675)

This pull request includes the implementation for the RegisterServer service and modification of the FindServers service to also return the registered servers.

Examples are also included in the PR.

I'm open for suggestions and improvements.


src/server/ua_services_discovery.c
```diff
@@ -118,3 +232,65 @@ void Service_GetEndpoints(UA_Server *server, UA_Session *session, const UA_GetEn
...
 }
+
+
+void Service_RegisterServer(UA_Server *server, UA_Session *session,
+                         const UA_RegisterServerRequest *request, UA_RegisterServerResponse *response) {
+    UA_LOG_DEBUG_SESSION(server->config.logger, session, "Processing RegisterServerRequest");
+
+    registeredServer_list_entry *registeredServer_entry = NULL;
+
+    {
```
jpfr: the extra block is not required in C99

jpfr:
That's already looking really good!
The PR could go into master real soon.

What else do you want to add?

src/server/ua_services_discovery.c
Pro: I added the extra block to hide the `current` variable created in the next line from the rest of the code to keep it clean.

Is that not needed?

Pro:
I think that the functionality on the server side is already pretty much done. So there shouldn't be any big changes any more.

What I will do next, is to create some example code to test the functionality of RegisterServer and FindServers. Maybe this example code can then be included in the tutorial.

Do we need additional tests for CI to test the RegisterServer and FindServers services?

Pro:
@jpfr the server needs a way to call the RegisterServer service from the discovery server.
Am I correct that the server itself can't call services on another server and thus I need to create a temporary client to call the RegisterServer service?

Stasik0:
I'd say yes: a temporary client is needed.

Pro:
@jpfr This PR is now ready to merge. It also includes some examples how to use the Discovery Server.
Unfortunately AppVeyor fails with some strange timeout when trying to download cygwin.

There's one thing not yet 100% clear to me:
Does the server need some way of mutex for the registered server array?
If the server is multithread and a client calls RegisterServer, the FindServer should be mutexed to avoid corrupt data?

examples/discovery/server_register.c
```diff
@@ -0,0 +1,167 @@
...
+	UA_LOG_INFO(logger, UA_LOGCATEGORY_SERVER, "Waiting for register thread to finish...");
+	pthread_join(registerServerThread, NULL);
+
+	// UNregister the server from the discovery server.
+	retval = UA_Server_unregister_discovery(server, "opc.tcp://localhost:4048" );
+	if (retval != UA_STATUSCODE_GOOD) {
+		UA_LOG_ERROR(logger, UA_LOGCATEGORY_SERVER, "Could not unregister server from discovery server. StatusCode 0x%08x", retval);
+		UA_Server_delete(server);
+		nl.deleteMembers(&nl);
+		return (int)retval;
```
FlorianPalm: Good work! Could you please review your indentation (->4 spaces). Thanks

Pro: Done, in latest commit.

Stasik0:
@Pro hey, I have seen you are registering periodically via a separate thread. Are you aware of periodic jobs of the server? e.g. here: https://github.com/open62541/open62541/blob/a123f230ed66ca8e93d70edd0f84e730365a20c0/examples/server_repeated_job.c#L38. So my API idea would be to let the _register() function add a periodic job with a user-defined (via API) or default period instead of creating a thread.

Pro:
Didn't know about this job feature. Very nice!!
Will update the example to use those instead of a separate thread

Stasik0:
@Pro we also have a "maintenance" job in the server which does the cleanup of stale sessions etc.:https://github.com/open62541/open62541/blob/6d23db1acf0bd77416b2bfa68fe8c858f12be5af/src/server/ua_server.c#L274. Do you also have a concept for cleaning up stale registrations at LDS? if not - you can probably extend this function

@open62541/open62541-core is a precopiler feature desired for LDS-Features? Unfortunately, I suppose that users may wish to disable LDS hard (at least the server side)

jpfr:
:+1: for a enabling LDS with a build switch

Pro:
Ok, then I'll modify the server.c to only include the RegisterServer service and LDS specific code in FindServers only when the build switch is enabled.
Also the corresponding examples will only be built if the switch is enabled.

Any preference for the build switch?
- `UA_ENABLE_LDS`
- `UA_ENABLE_DISCOVERY`

Pro:
@Stasik0 for LDS the servers should register periodically. At least that's what I found in the documentation here:
http://caxapa.ru/thumbs/272820/OPC_UA_Part_4_-_Services_Draft_1.01.05_S.pdf (Page 14)

Here is the description:

> Under normal conditions, Servers must periodically register with the DiscoveryServer as long as they are able to receive connections from Clients. If a Server goes offline then it must register itself once more and indicate that it is going offline. The registration frequency should be configurable, however, the default is 10 minutes.
> 
> If an error occurs during registration (e.g. the Discovery Server is not running) then the Server must periodically re-attempt registration. The frequency of these attempts should start at 1 second but gradually increase until the registration frequency is the same as what it would be if not errors occurred. The recommended approach would double the period each attempt until reaching the maximum.
> 
> When a Server registers with the a DiscoveryServer it may choose to provide a semaphore file which the DiscoveryServer can use to determine if the Server has been uninstalled from the machine. The DiscoveryServer must have read access to the file system that contains the file. 

This means that for the cleanup the server should check the semaphore file. If this isn't set, then it doesn't mention that servers should be removed automatically/after some specific time period. But there should definitely be some cleanup, e.g. after 60 Minutes, that at least after some time stale registrations are removed.

Unfortunately I don't have access to the complete most recent specification. Maybe you can check the corresponding page.

Stasik0:
So we need a cleanup mechanism for registered servers, dont we?

Pro:
Yes, I'll implement the cleanup like this:
If the semaphore file path is set, then it just checks the existance of the file. When it is deleted, the registration is removed.
If there is no semaphore file, then the registration will be removed if it is older than 60 minutes.

Stasik0:
Okay, its good.

jpfr:
> This means that for the cleanup the server should check the semaphore file.

Currently, everything that is not pure C is outside of the main library.
That includes all the architecture-specific things. Networking, logging, etc. are all exchangeable plugins.
Hence we are very portable. And people make use of that.

Do we want to support LDS features also on embedded systems that may not even have a filesystem?

Stasik0:
I mean these files only work locally. If a semaphore file is provided by the client -> a filesystem will probably exist.

src/server/ua_services_discovery.c
jpfr: ANSI C (89) required variables to be declared at the top of a block.
With C99 that is no longer the case. And most compilers allowed it anyway.

The variable is already "hidden" as it is local to that function.
So the bock seems mostly cosmetic to me.

Pro:
@jpfr @Stasik0  everything is now implemented as suggested:
- There's a compilation switch `UA_ENABLE_DISCOVERY`  
  It is `ON` by default. If OFF the code parts for `RegisterServer` and cleanup is excluded
- The RegisterServer stores all the registered servers in a double linked list
- In `UA_Server_cleanup` stale registrations are cleaned up if:
  * the registration included a semaphore file and this file is deleted
  * if there's no semaphore file, then the registration is removed if it is older than 60 Minutes
- There are three new examples in the `discovery` subdirectory:
  * `server_discovery.c` shows how to create a simple LDS server
  * `server_register.c` shows how to register a server with a LDS. Uses repeated jobs to re-register the server every 10 Minutes
  * `client_find_servers.c` shows how  to query the LDS and list the endpoints.

If you have any suggestions for improvement let me know. Otherwise this PR can be merged.

At some point in the future we should also implement some tests for CI to test discovery.

jpfr:
Thank you @Pro. That's a textbook pull request so far. Great job!

One last thing before merging. The PR should go into master and not into 0.2.
We froze the addition of new feature that are not required for the CTT tests.
Otherwise, we will never have a stable v0.2. :smile:

Pro:
Closed in favour of #687 which targets master branch.

PR #678, opened 2016-07-12
implementation of GetMonitoredItems for CTT

Stasik0:
I needed some hacks here

* a thread-local variable for a session-context of a method call
* I found an API-flaw of UA_Server_addMethodNode -- we cannot force nodeids of InputArgumetns and OutputArguments properties. Added a workaround in order not to break the 0.2 API.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -1303,7 +1303,7 @@ UA_Server * UA_Server_new(const UA_ServerConfig config) {
         UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT),
         UA_QUALIFIEDNAME(0, "GetMonitoredItems"), addmethodattributes,
         GetMonitoredItems, /* callback of the method node */
-        server, /* handle passed with the callback */
+        NULL, /* handle passed with the callback */
         1, &inputArguments,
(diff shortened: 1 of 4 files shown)


src/server/ua_server.c
```diff
@@ -1235,6 +1274,41 @@ UA_Server * UA_Server_new(const UA_ServerConfig config) {
...
+    addmethodattributes.displayName = UA_LOCALIZEDTEXT("", "GetMonitoredItems");
+    addmethodattributes.executable = true;
+    addmethodattributes.userExecutable = true;
+
+    UA_Server_addMethodNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_GETMONITOREDITEMS),
+        UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER),
+        UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT),
+        UA_QUALIFIEDNAME(0, "GetMonitoredItems"), addmethodattributes,
+        GetMonitoredItems, /* callback of the method node */
+        server, /* handle passed with the callback */
```
Stasik0: actually not needed, we can set to NULL :/

PR #682, opened 2016-07-13
Refactor appveyor.yml

Pro:
This is only a refactoring. The compilation tests are still the same as before.

It should also fix random timeouts when downloading cygwin


Pro:
Ready to merge

PR #687, opened 2016-07-14
Service RegisterServer and Discovery

Pro:
The Discovery Service is now implemented as suggested in PR #677:
- There's a compilation switch `UA_ENABLE_DISCOVERY`  
  It is `ON` by default. If OFF the code parts for `RegisterServer` and cleanup is excluded
- The RegisterServer stores all the registered servers in a double linked list
- In `UA_Server_cleanup` stale registrations are cleaned up if:
  * the registration included a semaphore file and this file is deleted
  * if there's no semaphore file, then the registration is removed if it is older than 60 Minutes
- There are three new examples in the `discovery` subdirectory:
  * `server_discovery.c` shows how to create a simple LDS server
  * `server_register.c` shows how to register a server with a LDS. Uses repeated jobs to re-register the server every 10 Minutes
  * `client_find_servers.c` shows how  to query the LDS and list the endpoints.

At some point in the future we should also implement some tests for CI to test discovery.

This PR targets the master branch (not `0.2` as in #677)


jpfr:
:cake:

PR #694, opened 2016-07-15
Include code from server_firstSteps.c in documentation

markusgraube:
Including code from example for first tutorial. I omitted all the non-amalgamtion stuff.
It can be used to help fixing #509 

If this is okay for you I would do the same for the rest of the tutorials.

Was there a reason for the whitespace between *#* and *include*?


jpfr:
Good idea!

The whitespace is just indendation for readability.

markusgraube:
So, this would be all from my side right now

PR #695, opened 2016-07-15
LDS customizable timeout for automatic de-register

Pro:
Added field in config to modify default timeout for de-registering a server.
`config.discoveryCleanupTimeout` in seconds

PR #697, opened 2016-07-15
 Enable valgrind for unit tests and fix memory leaks

Pro:
Enables valgrind during unit tests. Can be switched off in the CMake file.

Also includes #695 and #696 (so maybe you should merge them first).

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -7,7 +7,7 @@ else()
     add_definitions(-DUA_NO_AMALGAMATION)
 endif()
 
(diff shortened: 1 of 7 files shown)


Stasik0:
+1 here, @jpfr any comments?

jpfr:
:+1: Yes, very good idea.

Some of the unit tests take a long time with valgrind. So we have to see whether it can be used by default on travis also.

Pro:
Its already enabled by default on travis. Shouldn't be that much additional time.
The discovery test takes 12 secs because it also tests timeout cleanup.

Pro:
merged with #695

PR #703, opened 2016-07-18
Change order of deleting and saving incomplete message

nikita158:
Change order of deleting and saving incomplete message

### DIFF
--- a/src/ua_connection.c
+++ b/src/ua_connection.c
@@ -82,16 +82,6 @@ UA_Connection_completeMessages(UA_Connection *connection, UA_ByteString * UA_RES
         offset += length;
     }
 
-    /* throw the message away */
-    if(delete_at == 0) {
-        if(!*realloced) {
-            connection->releaseRecvBuffer(connection, message);
-            *realloced = true;
-        } else
-            UA_ByteString_deleteMembers(current);
-        return UA_STATUSCODE_GOOD;
-    }
-
     /* no complete message at all */
     if(offset == 0) {
         if(!*realloced) {
@@ -100,6 +90,16 @@ UA_Connection_completeMessages(UA_Connection *connection, UA_ByteString * UA_RES
             connection->releaseRecvBuffer(connection, message);
             *realloced = true;
         }
+        return UA_STATUSCODE_GOOD;
+    }
+    
+    /* throw the message away */
+    if(delete_at == 0) {
+        if(!*realloced) {
+            connection->releaseRecvBuffer(connection, message);
+            *realloced = true;
+        } else
+            UA_ByteString_deleteMembers(current);
         return UA_STATUSCODE_GOOD;
     }
 


Stasik0:
A larger comment would be helpful, @nikita158!

nikita158:
Many OPC UA clients(e.g. WinCC, TeslaScada) splits HEL-message into 2 packets.

Copying data into incompleteMessage buffer after deleting data makes no sense.
`/* throw the message away */
if(delete_at == 0) {...}

/* no complete message at all */
    if(offset == 0) {
        if(!*realloced) {
            /* store the buffer in the connection */
            UA_ByteString_copy(current, &connection->incompleteMessage);
            connection->releaseRecvBuffer(connection, message);
            *realloced = true;
        }
        return UA_STATUSCODE_GOOD;
    }`

now it works with WinCC; TeslaScada still can't connect.

jpfr:
> Copying data into incompleteMessage buffer after deleting data makes no sense.

That is not happening. We return after deleting a garbage message.

Can you provide a wireshark dump of the split HEL?

nikita158:
Thankx for reply,

printf debug(client is WinCC 7.4):
```C
static UA_StatusCode
socket_recv(UA_Connection *connection, UA_ByteString *response, UA_UInt32 timeout) {
    ...
    printf("socket_recv: response->length=%i, response->data=%s\n", response->length, response->data);
    ...
}

UA_StatusCode
UA_Connection_completeMessages(UA_Connection *connection, UA_ByteString * UA_RESTRICT message, UA_Boolean * UA_RESTRICT realloced) {
    ...
    /* throw the message away */
    if(delete_at == 0) {
        printf("UA_Connection_completeMessages: about to delete message. message->length=%i, message->data=%s\n", message->length, message->data);

        if(!*realloced) {
            connection->releaseRecvBuffer(connection, message);
            *realloced = true;
        } else
            UA_ByteString_deleteMembers(current);
        return UA_STATUSCODE_GOOD;
    }

    /* no complete message at all */
    if(offset == 0) {
        printf("UA_Connection_completeMessages: about to copy message into incompleteMessage buffer. message->length=%i, message->data=%s\n", message->length, message->data);

        UA_ByteString_copy(current, &connection->incompleteMessage);
        if(!*realloced) {
            /* store the buffer in the connection */
            UA_ByteString_copy(current, &connection->incompleteMessage);
            connection->releaseRecvBuffer(connection, message);
            *realloced = true;
        }
        return UA_STATUSCODE_GOOD;
    }
    ...
}
```
Output:
```
18:46:59.795 DEBUG: open62541: Connection 8 | New connection over TCP from 172.16.4.157:1835
socket_recv: response->length=1, response->data=H
UA_Connection_completeMessages: about to delete message. message->length=1, message->data=H
socket_recv: response->length=58, response->data=ELF;
UA_Connection_completeMessages: about to delete message. message->length=58, message->data=ELF;
```
<br>
So, as you said, if we return after deleting a garbage message, than
```C
/* no complete message at all */
    if(offset == 0) {...} 
```
can never be reached(seems for me)

jpfr:
Hey @nikita158,

Wow. you uncovered an edge case nobody thought would occur.
WinCC is sending a 1-byte TCP packet followed by a larger packet.
I have no idea how they came up with that. We classified the 1-byte packet as garbage since we couldn't even get to the point where the chunk size is stored.

Thanks for your PR.
However, it would be better to clean up that code on a bigger scale.
It has become just too difficult to understand.

I hope you are okay if I close this PR and we finish the issue in PR #707.

PR #711, opened 2016-07-19
The checks have to be relaxed to allow methods with dynamic array len…

Stasik0:
…gths

UaExpert passes arrayDimensionsSize of 1 with a number of array elemens.


jpfr:
`mathods`?

So the problem is that we receive a dimensionsize where we don't expect one?
Maybe we should just allow the case where no dimensionsize is defined on our side...

Stasik0:
Yea, thats why I removed the explicit comparisson. If you want one - do it in a callback.

PR #720, opened 2016-07-20
fix building with namespace compiler

markusgraube:
* generating NS0 need to know UA_NO_AMALGAMATION
* set src_generated/nodeset.c to be automatically generated in order to allow Cmake to handle the target correct

By the way, do we need
```
if(UA_BUILD_EXAMPLES_NODESET_COMPILER)
  add_custom_target(generate_informationmodel ALL
                    DEPENDS ${PROJECT_BINARY_DIR}/src_generated/nodeset.h ${PROJECT_BINARY_DIR}/src_generated/nodeset.c)
endif()
```
in the main CMakeLists.txt?

I would also prefer to put lines 314-329 (generating the example information model) in the example-CMake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,6 +336,7 @@ if(UA_ENABLE_AMALGAMATION)
     add_library(open62541-object OBJECT ${PROJECT_BINARY_DIR}/open62541.c ${PROJECT_BINARY_DIR}/open62541.h)
     target_include_directories(open62541-object PRIVATE ${PROJECT_BINARY_DIR})
 else()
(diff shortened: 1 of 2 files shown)


Stasik0:
maybe we need 

if(UA_BUILD_EXAMPLES_NODESET_COMPILER AND UA_BUILD_EXAMPLES) ?

Stasik0:
@markusgraube i pushed generation of nodeset in example CMakeLists of #725, is it the same? Can we close this one?

markusgraube:
@Stasik0 only partly. The generation of ns0 without amalgamtion stills fails. I opened a new pr for that #727

PR #721, opened 2016-07-20
speedup namespace compiler

markusgraube:
reordering of nodes takes quite a big amount of time when also including
all nodes from ns0. Thus, first remove all ignored nodes will make the
sorting much faster

@ichrispa can I do this or are there some side effects?

### DIFF
--- a/tools/pyUANamespace/generate_open62541CCode.py
+++ b/tools/pyUANamespace/generate_open62541CCode.py
@@ -132,7 +132,8 @@ ns.linkOpenPointers()
 
 # Remove nodes that are not printable or contain parsing errors, such as
 # unresolvable or no references or invalid NodeIDs
-ns.sanitize()
+logger.warn("Sanitizing is not performed until it's clear if sanity of ns0 nodes is clear")
+#ns.sanitize()
 
 # Parse Datatypes in order to find out what the XML keyed values actually
 # represent.
--- a/tools/pyUANamespace/ua_namespace.py
+++ b/tools/pyUANamespace/ua_namespace.py
@@ -547,12 +547,11 @@ class opcua_namespace():
         minweightnd = row[0]
     return (rind, minweightnd, minweight)
 
-  def reorderNodesMinDependencies(self):
-    # create a matrix represtantion of all node
-    #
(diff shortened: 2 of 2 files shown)


Stasik0:
@markusgraube IT seems to work, but I did not notice any speedup - is it really faster? I do not mind if we merge, though.

markusgraube:
@Stasik0 It is only for `UA_BUILD_EXAMPLES_NODESET_COMPILER`. Then it should be faster. For me it reduces the namespace compiler time for server_nodeset.xml from around 70 sesconds to 30 seconds.

Stasik0:
@markusgraube okay, namespace compiler is a black-box for me, probably you or @ichrispa can decide at best what to merge and what not

Stasik0:
@markusgraube okay, same here down from 70 to 30 secs, @ichrispa you okay with the code? Changes seem to be trivial

Stasik0:
it seems that the topological sorting is one of the bottlenecks... can't we use some heuristics like this one? http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.47.7745 it is dead-simple to implement, however it is unclear for me what the "edges" actually are. Are these references with isForward flag?

markusgraube:
@Stasik0 Nice idea to fasten the sort algorithm itself.

However, your cited paper is about a FAS problem (feedback arc set), isn't it? I thought, we just need a topologic sort since the `HasTypeDefinition` and `HasSubtype` relations should already be an DAG (directed acyclic graph), shouldn't they?

But it seems that the current algorithm can be improved for example by Kahn's algorithm (https://en.wikipedia.org/wiki/Topological_sorting). Or we just take an existing implementation https://pypi.python.org/pypi/topological_sort or https://pypi.python.org/pypi/toposort/1.0

or just follow a python recipe: https://algocoding.wordpress.com/2015/04/05/topological-sorting-python/

Stasik0:
@markusgraube okay, that is correct it thought the sorting considers all relations e.g. Ogranizes aswell, by the way, the xml nodesets coming from UaModeller seem already to be sorted, so no reorganization is needed

markusgraube:
Okay. I think we can close this and continue completely in #725

PR #724, opened 2016-07-21
Add accessLevel and userAccessLevel for variables in namespace compiler

markusgraube:
Should it be also set for `VariableType`?


jpfr:
> Should it be also set for VariableType?

No. Variables don't have that attribute.
https://github.com/open62541/open62541/blob/0.2/tools/schema/Opc.Ua.Types.bsd#L847

PR #725, opened 2016-07-22
Speedup namespace compiler #2 - Kahn's topological sort (relates #721)

Stasik0:
With this reimplementation the generation time goes down from 70 to 30 seconds on nodeset example

### DIFF
--- a/tools/pyUANamespace/ua_namespace.py
+++ b/tools/pyUANamespace/ua_namespace.py
@@ -541,12 +541,15 @@ class opcua_namespace():
       subTypeRefs.append(tn)
       subTypeRefs = subTypeRefs + self.getSubTypesOf(currentNode=tn)
     
+    def is_relevant(ref):
+      return (( ref.referenceType() in typeRefs and ref.isForward() )
+           or ( ref.referenceType() in subTypeRefs and not ref.isForward() ))
+    
     in_degree = { u : 0 for u in self.nodes }     # determine in-degree
     for u in self.nodes:                          # of each node
       for ref in u.getReferences():
        if isinstance(ref.target(), opcua_node_t):
-         if(( ref.referenceType() in typeRefs and ref.isForward() )
-         or ( ref.referenceType() in subTypeRefs and not ref.isForward() )):
+         if(is_relevant(ref)):
            in_degree[ref.target()] += 1
          if( ref.isForward() ):
            in_degree[u] += 1
@@ -564,8 +567,7 @@ class opcua_namespace():
       L.insert(0, u)
       for ref in u.getReferences():
        if isinstance(ref.target(), opcua_node_t):
-         if(( ref.referenceType() in typeRefs and ref.isForward() )
-         or ( ref.referenceType() in subTypeRefs and not ref.isForward() )):
+         if(is_relevant(ref)):
            in_degree[ref.target()] -= 1
            if in_degree[ref.target()] == 0:
              Q.appendleft(ref.target())
--- a/tools/pyUANamespace/ua_namespace.py
+++ b/tools/pyUANamespace/ua_namespace.py
@@ -530,13 +530,12 @@ class opcua_namespace():
         temp = temp + self.getSubTypesOf2(self.getNodeByBrowseName(t))
     relevant_types = temp
 
-    #in_degree means "being target of some reference in relevant_types"
     in_degree = { u : 0 for u in self.nodes }     # determine in-degree
     for u in self.nodes: # of each node
       if u not in printedExternally:
         for ref in u.getReferences():
          if isinstance(ref.target(), opcua_node_t):
-           if(ref.referenceType() in relevant_types and not ref.isForward()):
+           if(ref.referenceType() in relevant_types and ref.isForward()):
              in_degree[ref.target()] += 1
     
     Q = deque()                 # collect nodes with zero in-degree
@@ -548,10 +547,10 @@ class opcua_namespace():
     
     while Q:
       u = Q.pop()          # choose node of zero in-degree
-      L = [u] + L          # and 'remove' it from graph
+      L.append(u)          # and 'remove' it from graph
       for ref in u.getReferences():
        if isinstance(ref.target(), opcua_node_t):
-         if(ref.referenceType() in relevant_types and not ref.isForward()):
+         if(ref.referenceType() in relevant_types and ref.isForward()):
            in_degree[ref.target()] -= 1
            if in_degree[ref.target()] == 0:
              Q.appendleft(ref.target())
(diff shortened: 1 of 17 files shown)


Stasik0:
@markusgraube bump

Stasik0:
there is an arror :(

tools/pyUANamespace/ua_namespace.py
```diff
@@ -527,86 +529,38 @@ def printDotGraphWalk(self, depth=1, filename="out.dot", rootNode=None, followIn
...
+      u = Q.pop()          # choose node of zero in-degree
+      L.append(u)          # and 'remove' it from graph
+      for ref in u.getReferences():
+       if isinstance(ref.target(), opcua_node_t):
+         if ref.referenceType().browseName() in relevant_references and ref.isForward():
+           in_degree[ref.target()] -= 1
+           if in_degree[ref.target()] == 0:
+             Q.appendleft(ref.target())
+    if len(L) == len(self.nodes):
+        self.node = L
```
markusgraube: Shouldn't this be `self.nodes = L`?

tools/pyUANamespace/ua_namespace.py
```diff
@@ -527,86 +529,38 @@ def printDotGraphWalk(self, depth=1, filename="out.dot", rootNode=None, followIn
...
+        Q.appendleft(u)
+ 
+    L = []     # list for order of nodes
+    
+    while Q:
+      u = Q.pop()          # choose node of zero in-degree
+      L.append(u)          # and 'remove' it from graph
+      for ref in u.getReferences():
+       if isinstance(ref.target(), opcua_node_t):
+         if ref.referenceType().browseName() in relevant_references and ref.isForward():
```
markusgraube: The previous code took also the subtypes of the relevant references into account. Maybe, we should add this here too

Stasik0:
@markusgraube note that I had to add HasComponent to relevant references, since othervise the order was broken (components were initialized before their parents)

markusgraube:
@Stasik0 Really nice


Maybe, we could add all subtypes of HierarchicalReferences to relevant_references if this makes the sorting not too slow
```
self.getSubTypesOf(self.getNodeByBrowseName("HierarchicalReferences")
```
At least this shoul be still a DAG

Stasik0:
I have added subtypes now, however the getSubTypesOf function crashes if i give HasComponent to it (i had to woraround this) the same is for self.getSubTypesOf(self.getNodeByBrowseName("HierarchicalReferences")

but the generator works on my example and on NS0, how bout you?

> At least this shoul be still a DAG
i am not sure ) HasComponent and HasComponentOf are both hierarchical, but the graph is cyclic...

Stasik0:
unfortunately, "subtype_refrences" is also empty on my examples :(

Stasik0:
I got some other small optimizations, compile time is down to 17 seconds, @markusgraube could we also bake your improvements in? then we will get nscompiler on speed

jpfr:
I got some other small optimizations, compile time is down from 70 to 19 seconds.
:+1: Really nice!

markusgraube:
@Stasik0 

The sorting doesn't work for ns0
> Node graph is circular on the specified references

Thus, I can't push the sorting after the ignoring phase

jpfr:
> The sorting doesn't work for ns0

There are some well-known circular dependencies in ns0.
Since we check for circular dependencies during AddNodes, we need some "bootstrapping" in the server.
E.g. here: https://github.com/open62541/open62541/blob/0.2/src/server/ua_server.c#L762

We somehow need to whitelist those nodes. Imho @ichrispa already did that in his current sort mechanism.

By the way, if we could assume a minimal set of "boostrapped" nodes in the server, the namespace generation could use the public API only. That would make things a lot easier to handle for end-users bringing their own namespace definitions.

Stasik0:
@jpfr this bootstrapping fails somehow (shadowed definition etc)
@markusgraube I reproduced some @ichrispa's code I dont fully understand, but now it seems to work, could you try?

Stasik0:
btw ns0 generation seems a bit broken in 0.2 and here -> look at duplicate objects under "Server"

markusgraube:
@Stasik0 Yes, ns0 is compiling now. However, this solution is not great because the topological sorting just sorts 76 elements of 1516, which are exactly these elements which have no dependencies at all.
Then, we could just omit the sorting at all.

By the way: How much slower is bootstrapping node generation than using the high-level API? Or is it a matter of code and memory size that we want to avoid bootstrapping?

jpfr:
Bootstrapping should be faster than going by the AddNodes service. By adding nodes directly to the nodestore with no consistency checks.
But access to the nodestore is an internal API that is not intended for use by the end users.

Stasik0:
@markusgraube well, the behaviour is exactly as it was with the current implementation, probably we need to return to the specification of HasSubtype, HasTypeDefinitoin, Organizes, HasComponent and some more to get it better... However in this case the bootstrapping breaks as you have seen and I have no clue how to or time to fix it ATM.

Removing sorting completely gives a gain of about 2 sec with server_nodeset.xml. For complete ns0 the gain is not even measurable (i suspect it is also about 2 sec). I'd let it in and consider restricting relations and reworking bootstrapping at a later point of time.

Stasik0:
i give up atm, the bootstrapping is broken and code is too complex

Stasik0:
okay: summary of bugs 

(1)
bootstrapping has to be fixed at first.

if you comment out this line
https://github.com/open62541/open62541/pull/725/commits/2cb12485e8e1d1f98781926e53a26b42ab3e6a06#diff-81712a77990bcdaef01316de80879f87R555

the compilation of ns0 will break, however it should not be the case since NS-compiler has to cope with any order

(2) self.getSubTypesOf(currentNode=tn) return empty lists

I vote for letting the "sort" (which is actually not sorting) inside, work on these bugs and make a proper sorting later, by removing the marked line

 @open62541/open62541-core okay?

jpfr:
> I vote for letting the "sort" (which is actually not sorting) inside, work on these bugs and make a proper sorting later, by removing the marked line

When is later? Post 0.2?

Stasik0:
okay, guys. After the last spin, I made a proper topological sort, it works at least for server_nodeset.xml pretty good. Even the generation of NS0 seems to work as before.

However, I could not add a fix by @markusgraube, somehow sorting only not-ignored nodes breaks the order (just try by using a re-sorted server_nodeset.xml)

for nodeset_example the time is down to 15s

okay to merge? Please give it a try with some reordered XML files

Stasik0:
@ichrispa :+1: or :-1: ?

markusgraube:
@Stasik0 #what does this PR for the [minimal example model from #729](https://gist.github.com/markusgraube/058f0535100e7bf9c9ea448b0253289f)? It has a circular dependency, but should be able to create with high level api

ichrispa:
As with #729: Please don't replace djikstra, just add an alternative. A commandline switch to pick the sorting algorithm would be nice.

Pro:
@ichrispa @markusgraube @Stasik0 any update on this?
Would be really nice if we can integrate all the Namespace Compiler PRs or close them if they are not relevant anymore.

Pro:
Should we just close this in favour of #739 ?

Stasik0:
My parts seem to be in 739, so +1 to close

PR #729, opened 2016-07-26
Namespace compiler alternative introducing a --high-level-api option

markusgraube:
I have put some more effort into understanding the namespace compiler. To my understanding, we have to differentiate two situations:
1. Generating namesepaces directly into the stack (usually done with ns0 via `UA_ENABALE_GENERATE_NAMESPACE_0`). Here, we can use the internal functions of open62541and have no advantage of using the high level API or sorting nodes.
2. Generation of namespaces in userspace. Here, we can only use the high level API. For example, this is done with `UA_BUILD_EXAMPLES_NAMESPACE_COMPILER` and `examples/nodeset_examples.xml`

Thus, I have introduced a command line option `--high-level-api`. If it is turned off, everything is the same as before (just without reordering). If turned on, we don't use a complex sorting but following simple algorithm:

``` python
already_printed = list(printedExternally)
while unPrintedNodes:
    node_found = False
    for node in unPrintedNodes:
        for ref in node.getReferences():
            if ref.referenceType() in already_printed and ref.target() in already_printed:
                node_found = True
                code = code + node.printOpen62541CCode_HL_API(ref, supressGenerationOfAttribute)
                header = header + codegen.getNodeIdDefineString(node)
                unPrintedRefs.remove(ref)
                unPrintedNodes.remove(node)
                already_printed.append(node)
                break
    if not node_found:
        logger.critical("no code generation without bootstrapping possible")
        break
code.append("// creating references")
for r in unPrintedRefs:
    code = code + codegen.getCreateStandaloneReference(r.parent(), r)
```

The interesting changes are in 0aaa83b. 

@Stasik0 @ichrispa : What do you think? Is this a feasible approach?

If you are okay with this concept, I would clean up the code a little bit (especially in the non-`--high-level-api` section) and cherry pick all appropriate fixes from #725.

Furthermore, I'd like to make it unnecessary for the `--high-level-api` option to load the complete ns0 but use the blacklist of ns0 instead.


ichrispa:
Hi @markusgraube ,

as we just discussed: it should work, but I am not sure what kind of drawbacks/limitations might creep up in complex models. Maybe you could just test it with some userside model (CAE, ADI, FDI, ...). 

> Furthermore, I'd like to make it unnecessary for the --high-level-api option to load the complete ns0 but use the blacklist of ns0 instead.

That is not possible. The Compiler needs to have type definitions, struct definitions, reference types, etc. at hand to 'understand' the model. Not having NS0 in memory will result in an invalid model.

Best regards,

ichrispa

Stasik0:
hi @markusgraube,

well from what I see you just created another topological sorter. Imho it is in the same complexity or even slower as the Kahn's algorithm (yours is n^3, where Kahn' should be quadratic). From this point of view, I do not see any advantages. 

I have to agree with @ichrispa regarding ns0 access, e.g. to determine the subtypes of some type it needs to have a "HasSubtype" reference type. What we can try is to feed Opc.Ua.NodeSet2.Minimal.xml instead of the full NS0 which should be already sufficient for most "small" user-models.

markusgraube:
@Stasik0 You are right, that this topological sorter may not have the highest performance and could perhaps be exchanged with Khan or something else. 

However, there are fundamental difference to #725. The first thing is the differentiation between high level API and internal functions. 

The other important different aspect is the approach for sorting. 
> The problem is that even user-space models may have circular dependencies. So a perfectly valid model may require the use of internal bootstrapping-interface and there is no way to fix it.

Correct, however there are situation where it can also be possible to generate the this model without using internal functions of open6241. in #729, we are creating all nodes at first and afterwards all references.  Creating nodes with the high level API needs for every node a parent node (this dependency seems to come from the appropriate service). Thus, it's just enough to iterate over all nodes. Then, we can create for every node that has *at least one reference* that is already created (or blacklisted e.g. in ns0) the high level API call. We don't have to know how many other references are belonging to this node (as in 0.2 and #725).

I try to explain with a small example. Let's have nodes `A`, `B`, `C` and `NS0_1` with following references (the reference type is not important):
```
A->B
B->C
C->A
B->NS0_1
```
`NS0_1` may already be printed or blacklisted. `A`, `B` and `C` have a circular dependency, all with `in_degree=1`. #725 will fail to sort and create the nodes in the following order `A, B, C` and thus require bootstrapping since the reference `A->B` can not be resolved in order to use the high level API.

The code from #729 will iterate over the nodes and detect that there is an already existent reference target for `B->NS0_1`. Thus, `B` can be created using the high level API (and `NS0_1` as parent) and then the rest of the nodes can be also reordered to use the high level API: `B, A, C`. At the end, the reference `B->C` is added.

Or do I miss something important?

I know, that there might be some situations where the algorithm would break. However, these should be the cases when there is really no possibility to create the nodes with high level API. The amount of these situation should be really low, if we suppose an already created ns0 in the stack, since all OPC UA models I know are connected to namespace 0.

Stasik0:
okay this is a good point. However I have a feeling that already printed and blacklisted is not the same. In your example it is not clear for me how NS0_1 gets printed this is the quote "NS0_1 may already be printed or blacklisted". In #725 the sorting happens before printing and is not interleaved into the print process as in this PR.

If it is ignored it is clear - in this case we need to improve #725 by setting in_degree=0 for all ignored nodes. This will resolve your example as well. 

Should we just reverse the directions of references in Kahn's algorithm?

Stasik0:
@markusgraube could you give a new commit in #725 a try? is it the same? I would still vote for #725 for performance reasons (is it actually true) and a flexible "relevant_relations" definition (we need to find out which reference circles trigger bootstrapping-api)?

markusgraube:
@Stasik0 Sorry, I mixed up blacklisting and ignoring in the description. This PR is just about ignoredNodes which are handled as already_pinted.

> In your example it is not clear for me how NS0_1 gets printed this is the quote "NS0_1 may already be printed or blacklisted"

`NS0_1` would be in an ignore_file (e.g. `-i NodeID_Blacklist_FullNS0.txt`)

> a flexible "relevant_relations" definition (we need to find out which reference circles trigger bootstrapping-api)

I don't understand what the advantages of "relevant_relations" are. If there is a possibility to find a parentNode (in order to use node creation method of the high level api) it doesn't matter if this parentNode is referenced by "relevant_relation" or any other, do it? It's just important that there is a reference to an already_existent node.


> @markusgraube could you give a new commit in #725 a try? is it the same?

@Stasik0 The new commit is working for the server_example.xml, but not for other models.
If you try to insert some of the companion information models (ISA95, AMLBaseTypes, AMLLibraries), #725 will fallback to internal function calls while #729 still can solve this with the high level api.

```
  add_custom_command(OUTPUT ${PROJECT_BINARY_DIR}/src_generated/nodeset.h ${PROJECT_BINARY_DIR}/src_generated/nodeset.c
                  PRE_BUILD
                  COMMAND ${PYTHON_EXECUTABLE} ${PROJECT_SOURCE_DIR}/tools/pyUANamespace/generate_open62541CCode.py
                                                -i ${PROJECT_SOURCE_DIR}/tools/pyUANamespace/NodeID_Blacklist_FullNS0.txt
						${PROJECT_SOURCE_DIR}/tools/schema/namespace0/Opc.Ua.NodeSet2.xml
                                                ${PROJECT_SOURCE_DIR}/examples/server_nodeset.xml
                                                ${PROJECT_SOURCE_DIR}/examples/Opc.ISA95.NodeSet2.xml
                                                ${PROJECT_SOURCE_DIR}/examples/Opc.Ua.AMLBaseTypes.NodeSet2.xml
                                                ${PROJECT_SOURCE_DIR}/examples/Opc.Ua.AMLLibraries.NodeSet2.xml
                                                ${PROJECT_BINARY_DIR}/src_generated/nodeset)
```

#729 is slower than #725 (which might result from the fact that #725 at some point just gives up to sort further).  However, I see possibilties to improve the sorting (perhaps by seperating sorting and printing or maybe using kahn's algorithm).

However, we first need to agree if this approach is feasible at all.

tools/pyUANamespace/open62541_MacroHelper.py
```diff
@@ -106,8 +106,6 @@ def getCreateStandaloneReference(self, sourcenode, reference):
 
     if reference.isForward():
       code.append("UA_Server_addReference(server, " + self.getCreateNodeIDMacro(sourcenode) + ", " + self.getCreateNodeIDMacro(reference.referenceType()) + ", " + self.getCreateExpandedNodeIDMacro(reference.target()) + ", true);")
-    else:
-      code.append("UA_Server_addReference(server, " + self.getCreateNodeIDMacro(sourcenode) + ", " + self.getCreateNodeIDMacro(reference.referenceType()) + ", " + self.getCreateExpandedNodeIDMacro(reference.target()) + ", false);")
```
Stasik0: are you sure you also cover the bootstrapping case?

Stasik0:
you can try to merge "poor_py_opt" branch into your pr, to get some more performance increase. At the end I'd prefer a more robust code that handles more examples with userspace code over two variants. So if its true :+1: for this PR and :-1: for #725

tools/pyUANamespace/open62541_MacroHelper.py
markusgraube: As far as I can overview it, i would say yes. If the bootstrapping method is using `UA_Server_addReference`, it should also be there enough to execute it for the forward references.

markusgraube:
@ichrispa 
> Maybe you could just test it with some userside model (CAE, ADI, FDI, ...). 

I have tested this PR with the models for DI, ADI, AML and ISA-95 and it seems to work.
Do you have a node xml information model for FDI and PLCOpen?
What is CAE?

ichrispa:
@markusgraube: 
For the available models, I suggest https://github.com/OPCFoundation/UA-Nodeset

As far as the algorithm goes: It should not replace the current djikstra, but overwrite it only for --high-level. The essential difference here being the generation of a fatal message when bootstrapping is used. After a merge, I would also have any other algorithm fail if high-level is enabled.

An algorithm choice switch would come in handy (--sort-forward, --sort-djikstra, --sort-khan).

jpfr:
@ichrispa Do you have technical reasons for wanting to keep the current topological sorting?
Imho, all valid orderings should be essentially equivalent.

Imho, it requires a very good reason to take on additional complexity (selectable sort-algorithms) in a module that people already barely understand.

Pro:
@ichrispa @markusgraube @Stasik0 any update on this?

@markusgraube can you also please sign the CLA? Thanks!
https://cla-assistant.io/open62541/open62541?pullRequest=729

Pro:
Should we just close this in favour of #739 ?

Stasik0:
+1

PR #732, opened 2016-07-31
Code review: @Pro's MDNS implementation

jpfr:
PR for merging @Pro's MDNS Discovery implementation.

**Open points list**
- [x] Merge schema changes into master beforehand to lighten the PR
- [x] Merge Travis/Appveyor improvements into master beforehand to lighten the PR
- [ ] Get rid of POSIX networking includes in the main library (invent a plugin interface for general purpose UDP?)


jpfr:
I made a list of open points in the first commit. Feel free to edit.

Imho the UA-discovery functionality is solid. But I would opt for more modularity. That takes effort but should result in a cleaner architecture in the long run.

Eventually, we may have up to three usages of (multicast) UDP in open62541.
- discovery
- clock sync
- pub/sub extension

@Pro Do you think it is is possible to have a single "plugin" style UDP interface in open62541 for various uses? The upside would be that we keep the main library free from system-dependent networking includes (also POSIX). That's always a problem for embedded targets.

You hacked on mdnsd itself quite a bit (https://github.com/Pro/mdnsd). Of course it also depends on what can be done on that side.

Pro:
I'll look into it.
Also there is still a double free bug if the server runs longer than 10 minutes. Should be easy to fix, will do it when I'm back.

jpfr:
I rebased this branch on 0.2.
Please checkout to avoid merging pains.

The current travis fails are due to a remaining memleak that is picked up by valgrind...

Pro:
Travis is fixed now. The problem was that lastSeen was using non monotonic clock, and the timeout check monotonic. Fixed in PR #749

jpfr:
Hey @Pro,

this seems to have stabilized. At least, I haven't seen any bugfixes coming in the last month or so.
Shall we do a rebase + squash and finalize the merge?

Pro:
Yes, this can be merged.
I would recommend doing a merge instead of rebase. I already tried a rebase once and after the tenth `git rebase --continue` and conflict fixing I gave up. Merging should be easier.

I can merge master and 0.2 into this branch and fix the current conflicts.

Pro:
@jpfr finally I merged master into this branch. It still needs integration of your PR https://github.com/open62541/open62541/pull/932

In the meantime I'll check why travis and appveyor fail. Probably some error during merging

src/server/ua_server.c
```diff
@@ -241,12 +241,43 @@ void UA_Server_delete(UA_Server *server) {
                     &UA_TYPES[UA_TYPES_ENDPOINTDESCRIPTION]);
 
 #ifdef UA_ENABLE_DISCOVERY
-    registeredServer_list_entry *current, *temp;
-    LIST_FOREACH_SAFE(current, &server->registeredServers, pointers, temp) {
-        LIST_REMOVE(current, pointers);
-        UA_RegisteredServer_deleteMembers(&current->registeredServer);
-        UA_free(current);
+    {
```
jpfr: I don't think we need that additional blocks.
Maybe they are indicators for functionality that can be separated off in a small static function.
That's considered good style.

tests/check_discovery.c
```diff
@@ -411,9 +411,10 @@ static void GetEndpointsAndCheck(const char* discoveryUrl, const char* filterTra
 
     ck_assert_uint_eq(endpointArraySize , expectedEndpointUrlsSize);
 
-    for(size_t j = 0; j < endpointArraySize; j++) {
+    for(size_t j = 0; j < endpointArraySize && j < expectedEndpointUrlsSize; j++) {
         UA_EndpointDescription* endpoint = &endpointArray[j];
         char *eu = UA_String_to_char_alloc(&endpoint->endpointUrl);
+        ck_assert_ptr_ne(eu, NULL); // clang static analysis fix
```
jpfr: Hey @Pro ,

embedded applications are often on a tight budget ram-wise. We can never know when memory runs out.
So the rest of the SDK always checks whether malloc succeeds.

The assertion patches up the clang-analyzer issue.
But it will still segfault the moment malloc actually returns NULL.

Once clang-analyzer is happy with the code (and you are too), we can push the branch to coverity.
Coverity will probably uncover a couple more issues. That's normal for a feature this big. 😄

src/server/ua_server.c
Pro: I used the additional blocks to keep `*current` and `*temp` local. IMHO it does not make much sense to move the functionality in those blocks into a small function, because it's less than 20 lines and it is only cleaning up

Pro:
@jpfr this PR is now ready for final review and to be merged.
The build still fails because in master there's a build error: https://travis-ci.org/open62541/open62541/builds/200158980

Pro:
@jpfr all green :+1: 

I refactored the code and cleaned it up.
Any additional comments?

include/ua_client.h
```diff
@@ -114,6 +114,15 @@ UA_Client_getEndpoints(UA_Client *client, const char *serverUrl,
...
 UA_Client_connect(UA_Client *client, const char *endpointUrl);
 
+/* Connect to the selected server.
+ * This will not create a session.
+ *
+ * @param client to use
+ * @param endpointURL to connect (for example "opc.tcp://localhost:16664")
+ * @return Indicates whether the operation succeeded or returns an error code */
+UA_StatusCode UA_EXPORT
+UA_Client_connect_no_session(UA_Client *client, const char *endpointUrl);
```
jpfr: Is there a way to not expose this function?
Normal users should not have to care about this.

plugins/ua_network_tcp.c
```diff
@@ -24,10 +24,6 @@
 # ifndef __clang__
 #  include <malloc.h>
 # endif
-/* Fix redefinition of SLIST_ENTRY on mingw winnt.h */
```
jpfr: Don't we need the same `pragma push_macro("SLIST_ENTRY")` as in ua_clock.c?

jpfr:
Getting there. I refactored a bit without changing the functionality.
Splitting functionality out of large methods and such...

I made some last comments. Can you check?

src/server/ua_services.h
```diff
@@ -53,10 +53,24 @@ void Service_GetEndpoints(UA_Server *server, UA_Session *session,
                           UA_GetEndpointsResponse *response);
 
 #ifdef UA_ENABLE_DISCOVERY
+
+# ifdef UA_ENABLE_DISCOVERY_MULTICAST
```
jpfr: Can you add a small description copied from the service definition in part 4?
Compare the other services...

src/server/ua_server_worker.c
```diff
@@ -563,6 +569,50 @@ UA_StatusCode UA_Server_run_startup(UA_Server *server) {
...
 
+
+#ifdef UA_ENABLE_DISCOVERY_MULTICAST
+    if (server->config.applicationDescription.applicationType == UA_APPLICATIONTYPE_DISCOVERYSERVER) {
+
+        char *appName = (char *)malloc(server->config.mdnsServerName.length +1);
+        memcpy(appName, server->config.mdnsServerName.data, server->config.mdnsServerName.length);
+        appName[server->config.mdnsServerName.length] = '\0';
+
+        for(size_t i = 0; i < server->config.networkLayersSize; i++) {
```
jpfr: Split the inner part of the for-loop in a separate method.
And add comments. What does the code do? Why for every networklayer? Why here?

src/server/ua_server_worker.c
```diff
@@ -563,6 +569,50 @@ UA_StatusCode UA_Server_run_startup(UA_Server *server) {
...
+            char hostname[256]; hostname[0] = '\0';
+            const char *path;
+            {
+                char* uri = (char *)malloc(sizeof(char) * nl->discoveryUrl.length + 1);
+                strncpy(uri, (char*) nl->discoveryUrl.data, nl->discoveryUrl.length);
+                uri[nl->discoveryUrl.length] = '\0';
+                UA_StatusCode retval;
+                if ((retval = UA_EndpointUrl_split(uri, hostname, &port, &path)) != UA_STATUSCODE_GOOD) {
+                    if (retval == UA_STATUSCODE_BADOUTOFRANGE)
+                        UA_LOG_WARNING(server->config.logger, UA_LOGCATEGORY_NETWORK, "Server url '%s' size invalid", uri);
```
jpfr: Overlong lines can be trivially shortened.

src/server/ua_server_internal.h
```diff
@@ -275,6 +323,30 @@ void Service_Call_single(UA_Server *server, UA_Session *session,
...
+void UA_Discovery_multicastDestroy(UA_Server* server);
+
+typedef enum {
+    UA_DISCOVERY_TCP,     /* OPC UA TCP mapping */
+    UA_DISCOVERY_TLS     /* OPC UA HTTPS mapping */
+} UA_DiscoveryProtocol;
+
+UA_StatusCode UA_Discovery_multicastQuery(UA_Server* server);
+
+UA_StatusCode UA_Discovery_addRecord(UA_Server* server, const char* servername, const char* hostname, unsigned short port, const char* path,
```
jpfr: Overlong lines.

include/ua_client.h
Pro: I think it's still needed, e.g. to call FindServers service:
https://github.com/open62541/open62541/blob/feature_mdns/examples/discovery/client_find_servers.c#L20

plugins/ua_network_tcp.c
Pro: I also wondered the same, but it worked without the push macro. Anyways it's better to use it as you already implemented: https://github.com/open62541/open62541/commit/987c910efd13ed3fa1a03bb7f6e21a6e6de425fe

src/server/ua_services.h
Pro: Done: https://github.com/open62541/open62541/pull/732/commits/daf95c42605c0c4043d117de14d6bda50233fc27

src/server/ua_server_worker.c
Pro: This return value may also mean that the URL is invalid, e.g. port too long, not only number of chars

src/server/ua_server_worker.c
Pro: Done https://github.com/open62541/open62541/pull/732/commits/bce3281eb36cda7d18a61d5ebb4d8a41eb909c89

src/server/ua_server_internal.h
Pro: Done https://github.com/open62541/open62541/pull/732/commits/679d588081a582f46f648d483fbd6680742d63cf

Pro:
@jpfr all green (except AppVeyor due to null byte problem)

jpfr:
> I think it's still needed, e.g. to call FindServers service:
https://github.com/open62541/open62541/blob/feature_mdns/examples/discovery/client_find_servers.c#L20

I understand why we need to connect without a session for some services.
But that function is a huge magnet for bugs. What is the state of the client after running it?
What if we already have a session? What if we want to create a session afterwards?

Either we come up with a really clean and really well documented state machine for the client state/connection/session stuff **or** we just create a temporal new connection for the stuff that runs without a session.

Can we just wrap the findservers service such that it uses `UA_Client_connect_no_session` internally on a new connection that is closed afterwards?

Pro:
@jpfr ready to merge

jpfr:
Congratiulations @Pro!

Issue #748, opened 2016-08-15
UA_Server_getDataType memory leak


kyursen:
There is a server memory leak when `UA_Server_getDataType` results in calling `UA_Variant_setScalarCopy`.

The issue can be reproduced by modifying server_datasource.c example:
```
    // Insert these lines after UA_Server_addDataSourceVariableNode at line 71
    UA_NodeId nodeId;
    UA_NodeId_init(&nodeId);
    UA_Server_readDataType(server, myIntegerNodeId, &nodeId);
    UA_NodeId_deleteMembers(&nodeId);
```

Valgrind report:
```
==3488== 24 bytes in 1 blocks are definitely lost in loss record 1 of 1
==3488==    at 0x4C2AB80: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==3488==    by 0x4034DE: UA_Variant_setScalarCopy (ua_types.c:524)
==3488==    by 0x4194EB: getVariableNodeDataType (ua_services_attribute.c:191)
==3488==    by 0x419A8E: Service_Read_single (ua_services_attribute.c:310)
==3488==    by 0x412382: __UA_Server_read (ua_server.c:1445)
==3488==    by 0x4014DD: UA_Server_readDataType (ua_server.h:308)
==3488==    by 0x401A3B: main (server_datasource.c:107)
```

jpfr:
fixed in 0.2 branch

PR #751, opened 2016-08-17
Fix UA_Server_readDataType memory leak

kyursen:
Here is a quick fix related to #748.

Fix is quite ugly:

1. Casting `void*` to `NodeId*`.
I'm not sure that typeId is always numeric. If it is, then `UA_NodeId_copy` can be safely replaced with `memcpy`.

2. Adding unneeded malloc in case of UA_VALUESOURCE_VARIANT.
`getVariableNodeDataType` always makes a new copy of typeId to be deterministic for `__UA_Server_read`. One of the proposals is to give `__UA_Server_read` some flag indicating whether it should free memory or not.

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -506,7 +506,10 @@ __UA_Server_read(UA_Server *server, const UA_NodeId *nodeId,
     if(attributeId == UA_ATTRIBUTEID_VALUE ||
        attributeId == UA_ATTRIBUTEID_ARRAYDIMENSIONS)
         memcpy(v, &dv.value, sizeof(UA_Variant));
-    else {
+    else if(attributeId == UA_ATTRIBUTEID_DATATYPE) {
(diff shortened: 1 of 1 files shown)


jpfr:
1. No, the nodeid may contain an allocated string.
2. We should aim for consistent behavior across all attribute reads. I think there are cases where a variant with pointers into the existing node are returned. Those are not freed during a _deleteMembers call.

I will have a closer look at you PR tomorrow.

kyursen:
@jpfr I've noticed that `forceVariantSetScalar` sets UA_VARIANT_DATA_NODELETE storage type, just as you mentioned. So I have amended PR to revert changes in `getVariableNodeDataType`.

jpfr:
Thanks for the PR!
Your commit was merge here: https://github.com/open62541/open62541/commit/bfc01857daf7195337e2868cf15da4ec39c9199b

PR #764, opened 2016-08-23
Add install section for CMake

markusgraube:
To make it easier to integrate open62541 within external toolchains.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,3 +371,19 @@ if(UA_BUILD_EXAMPLES_NODESET_COMPILER)
   add_custom_target(generate_informationmodel ALL
                     DEPENDS ${PROJECT_BINARY_DIR}/src_generated/nodeset.h ${PROJECT_BINARY_DIR}/src_generated/nodeset.c)
 endif()
+
+
+##########################
+# Installation           #
+##########################
+
+install(TARGETS open62541
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib
+    )
+if(UA_ENABLE_AMALGAMATION)
+    install(FILES ${PROJECT_BINARY_DIR}/open62541.h DESTINATION include)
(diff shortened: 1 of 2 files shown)


jpfr:
Integrate with external toolchains or provide an installation à la RPM or debian package?

For the latter, we could think of always installing the amalgamated header open62541.h instead of copying many ua_* files to /usr/include.

Any opinions on this?

markusgraube:
Currently, the first option.

but debian or rpm packages are also a good idea. I'll add a first draft using [CPack](https://cmake.org/Wiki/CMake:Packaging_With_CPack)
And here you are right, `open62541.h` is enough

CMakeLists.txt
```diff
@@ -299,16 +299,17 @@ add_custom_command(OUTPUT ${PROJECT_BINARY_DIR}/open62541.h
                    COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/tools/amalgamate.py
                            ${GIT_COMMIT_ID} ${CMAKE_CURRENT_BINARY_DIR}/open62541.h ${exported_headers}
                    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/tools/amalgamate.py
-                           ${exported_headers}
-                           ${internal_headers})
```
jpfr: Great stuff!
Last small issue before merging.
I'd keep the exported_headers, internal_headers and lib_sources in the DEPENDS list.
Otherwise the amalgamation will not be rebuild automatically if, say a .c-file, changes during development.

markusgraube: @jpfr I don't see the point why the generation of open62541.h is dependant on the internal_headers and lib_sources. If a c-file changes which has no impact on the external_headers, there is no need to regenerate the open62541.h. Or do I miss something?

jpfr:
Great!

PR #770, opened 2016-08-25
Fix #769: Wrong NodeIDs offset for BinaryEncoding

### LINKED ISSUE #769: Wrong NodeIDs offset for BinaryEncoding
In commit https://github.com/open62541/open62541/commit/952072a3469f12dbb2c406d9dbd8689182fbc2b1 @jpfr  updated the `NodeIds.csv` to the most recent version.

Most of the (old) node IDs look something like this:
```
 FindServersRequest,420,DataType
 FindServersRequest_Encoding_DefaultXml,421,Object
 FindServersRequest_Encoding_DefaultBinary,422,Object
```
So encoding binary is just the main ID+2.
That's also used when constructing the binary message:
https://github.com/open62541/open62541/blob/0.2/include/ua_constants.h#L104
and
https://github.com/open62541/open62541/blob/0.2/src/ua_securechannel.c#L222

Unfortunately this does not work anymore for new types. If you check the new NodeIds.csv, there is for example:
```
ServerOnNetwork,12189,DataType
FindServersOnNetworkRequest,12190,DataType
FindServersOnNetworkResponse,12191,DataType
RegisterServer2Request,12193,DataType
RegisterServer2Response,12194,DataType
ServerOnNetwork_Encoding_DefaultXml,12195,Object
FindServersOnNetworkRequest_Encoding_DefaultXml,12196,Object
FindServersOnNetworkResponse_Encoding_DefaultXml,12197,Object
RegisterServer2Request_Encoding_DefaultXml,12199,Object
RegisterServer2Response_Encoding_DefaultXml,12200,Object
OpcUa_XmlSchema_ServerOnNetwork,12201,Variable
OpcUa_XmlSchema_ServerOnNetwork_DataTypeVersion,12202,Variable
OpcUa_XmlSchema_ServerOnNetwork_DictionaryFragment,12203,Variable
ServerOnNetwork_Encoding_DefaultBinary,12207,Object
FindServersOnNetworkRequest_Encoding_DefaultBinary,12208,Object
FindServersOnNetworkResponse_Encoding_DefaultBinary,12209,Object
RegisterServer2Request_Encoding_DefaultBinary,12211,Object
RegisterServer2Response_Encoding_DefaultBinary,12212,Object
```

Especially look at (https://github.com/open62541/open62541/blob/0.2/tools/schema/NodeIds.csv#L4822)
```
RegisterServer2Request,12193,DataType
RegisterServer2Response,12194,DataType
ServerOnNetwork_Encoding_DefaultXml,12195,Object
```
The node ID 12195 is NOT `RegisterServer2Request_Encoding_DefaultBinary` but `ServerOnNetwork_Encoding_DefaultXml`

So just offsetting the NodeID by 2 to get the ID for Binary Encoding does not work anymore, and wrong node IDs are sent.

Do you have any idea/suggestion how to fix this?

Pro:
See #769 for additional description of the problem.

This is my proposed solution for the problem.
Each DataType struct stores the correct NodeID for binary/XML encoded.

Open for feedback or direct merge

### DIFF
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -745,6 +745,8 @@ struct UA_DataType {
                                     pointers */
     UA_Boolean overlayable  : 1; /* The type has the identical memory layout in
                                     memory and on the binary stream. */
+    UA_UInt32  xmlEncodingId;    /* NodeId of datatype when encoded as XML */
+    UA_UInt32  binaryEncodingId;    /* NodeId of datatype when encoded as XML */
     UA_DataTypeMember *members;
 };
 
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -745,8 +745,8 @@ struct UA_DataType {
                                     pointers */
     UA_Boolean overlayable  : 1; /* The type has the identical memory layout in
                                     memory and on the binary stream. */
-    UA_UInt32  xmlEncodingId;    /* NodeId of datatype when encoded as XML */
-    UA_UInt32  binaryEncodingId;    /* NodeId of datatype when encoded as binary */
+    //UA_UInt16  xmlEncodingId;    /* NodeId of datatype when encoded as XML */
+    UA_UInt16  binaryEncodingId;    /* NodeId of datatype when encoded as binary */
     UA_DataTypeMember *members;
 };
 
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -746,7 +746,7 @@ struct UA_DataType {
     UA_Boolean overlayable  : 1; /* The type has the identical memory layout in
                                     memory and on the binary stream. */
     UA_UInt32  xmlEncodingId;    /* NodeId of datatype when encoded as XML */
-    UA_UInt32  binaryEncodingId;    /* NodeId of datatype when encoded as XML */
+    UA_UInt32  binaryEncodingId;    /* NodeId of datatype when encoded as binary */
     UA_DataTypeMember *members;
 };
 
--- a/src/ua_types_encoding_binary.c
+++ b/src/ua_types_encoding_binary.c
@@ -697,6 +697,26 @@ LocalizedText_decodeBinary(UA_LocalizedText *dst, const UA_DataType *_) {
     return retval;
 }
 
+static UA_StatusCode findDataType(const UA_NodeId *typeId, const UA_DataType **findtype) {
+    for(size_t i = 0; i < UA_TYPES_COUNT; i++) {
+        if(UA_NodeId_equal(typeId, &UA_TYPES[i].typeId)) {
+            *findtype = &UA_TYPES[i];
+            return UA_STATUSCODE_GOOD;
+        }
+    }
+    return UA_STATUSCODE_BADNODEIDUNKNOWN;
+}
+
+static UA_StatusCode findDataTypeByBinary(const UA_NodeId *typeId, const UA_DataType **findtype) {
+    for(size_t i = 0; i < UA_TYPES_COUNT; i++) {
+        if (UA_TYPES[i].binaryEncodingId == typeId->identifier.numeric) {
+            *findtype = &UA_TYPES[i];
+            return UA_STATUSCODE_GOOD;
+        }
+    }
+    return UA_STATUSCODE_BADNODEIDUNKNOWN;
+}
+
 /* ExtensionObject */
 static UA_StatusCode
 ExtensionObject_encodeBinary(UA_ExtensionObject const *src, const UA_DataType *_) {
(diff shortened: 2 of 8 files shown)


src/ua_types_encoding_binary.c
```diff
@@ -778,8 +786,15 @@ ExtensionObject_decodeBinary(UA_ExtensionObject *dst, const UA_DataType *_) {
         /* helping clang analyzer, typeId is numeric */
         UA_assert(typeId.identifier.byteString.data == NULL);
         UA_assert(typeId.identifier.string.data == NULL);
-        typeId.identifier.numeric -= UA_ENCODINGOFFSET_BINARY;
-        findDataType(&typeId, &type);
+        size_t i;
+        for (i = 0; i<UA_TYPES_COUNT; i++) {
```
jpfr: I don't get why you loop over the types and then call findDataType.
That's a bit redundant imho.

I see three loops going over the types. Imho we should only use findDataTypes in each of these places.

include/ua_types.h
```diff
@@ -745,6 +745,8 @@ struct UA_DataType {
                                     pointers */
     UA_Boolean overlayable  : 1; /* The type has the identical memory layout in
                                     memory and on the binary stream. */
+    UA_UInt32  xmlEncodingId;    /* NodeId of datatype when encoded as XML */
+    UA_UInt32  binaryEncodingId;    /* NodeId of datatype when encoded as XML */
```
jpfr: ... when encoded as **binary**

src/ua_types_encoding_binary.c
Pro: This loop checks against `binaryEncodingId` not (as findDataType) against typeId.identifier.numeric.

But for the loop above (line 711) you are right.

I'll extract it to a separate method and avoid double-loops. But first I wanted to see if this solution would be OK in general.

jpfr: Part 6 says:
> The identifier for the DataTypeEncoding node in the Server's AddressSpace.
ExtensionObjects defined by the OPC UA specification have a numeric node
identifier assigned to them with a NamespaceIndex of 0. The numeric
identifiers are defined in A.3.

As I understand the text, we should compare for the binary encoding identifier as we do in the other two places.
Please correct me if I missunderstood.

jpfr: FindDataTypes only applies when we get a type in binary encoding.
Otherwise, we already have the pointer to the type, don't we. :smirk:

src/ua_types_encoding_binary.c
```diff
@@ -708,13 +728,19 @@ ExtensionObject_encodeBinary(UA_ExtensionObject const *src, const UA_DataType *_
         UA_NodeId typeId = src->content.decoded.type->typeId;
         if(typeId.identifierType != UA_NODEIDTYPE_NUMERIC)
             return UA_STATUSCODE_BADENCODINGERROR;
-        typeId.identifier.numeric += UA_ENCODINGOFFSET_BINARY;
+        const UA_DataType *type = NULL;
+        /* helping clang analyzer, typeId is numeric */
+        UA_assert(typeId.identifier.byteString.data == NULL);
+        UA_assert(typeId.identifier.string.data == NULL);
+        if (findDataType(&typeId, &type)!= UA_STATUSCODE_GOOD)
```
jpfr: We should not do a findDataType here.
We know the type. It is `src->content.decoded.type`.

Otherwise, ready to merge!

include/ua_types.h
```diff
@@ -745,6 +745,8 @@ struct UA_DataType {
                                     pointers */
     UA_Boolean overlayable  : 1; /* The type has the identical memory layout in
                                     memory and on the binary stream. */
+    UA_UInt32  xmlEncodingId;    /* NodeId of datatype when encoded as XML */
```
jpfr: These are 64bit * 200 types -> 1,5kb added binary size.
- Currently we don't use the xmlEncodingId. The line can be commented out.
- There is no nodeid in ns0 with an identifier > 15000.  The binaryEncodingId could become an UInt16 integer.

Pro: I used UInt32 because UA_NodeId also uses UInt32 for the numeric representation. I'll change it here to UInt16

What about the UInt32 in UA_NodeId? Should we also change it there?

src/ua_types_encoding_binary.c
Pro: Yes, correct. So we don't need the `findDataType` at all.

Pro:
@jpfr done. Ready to merge.

jpfr:
Looks great! Merged.

PR #781, opened 2016-08-30
Extract EndpointUrl Splitting

### LINKED ISSUE #778: connect issue - DNS lookup of 192.163.1.193 failed with error Servname not supported for ai_socktype
Hello

I use open62541 SDK v0.2 today to connect to the OPC UA Foundation Server.
Open62541 client is on the ubuntu(virtualbox)
Server is on the windows PC.
I can do this connection before.

Error message : (when connect to the Unified Automation Server)
[08/29/2016 06:31:21.456] warning/network	DNS lookup of tq021110740 failed with error Servname not supported for ai_socktype

Error message : (when connect to the OPC UA Foundation Server)
[08/29/2016 07:10:09.713] error/client	GetEndpointRequest failed with statuscode 0x80b90000

Pro:
Fixes #778 and improves handling of invalid urls.

Can you think of any other valid/invalid combination to test in the unit tests?

Otherwise ready to merge.

### DIFF
--- a/include/ua_connection.h
+++ b/include/ua_connection.h
@@ -127,10 +127,20 @@ void UA_EXPORT UA_Connection_deleteMembers(UA_Connection *connection);
  * @param endpointUrl The endpoint URL to split up
  * @param hostname the target array for hostname. Has to be at least 256 size.
  * @param port set to the port of the url or 0
- * @param path pointing to the end of given endpointUrl or to NULL if no path given. The starting '/' is included in path
+ * @param path pointing to the end of given endpointUrl or to NULL if no path given. The starting '/' is NOT included in path
  * @return UA_STATUSCODE_BADOUTOFRANGE if url too long, UA_STATUSCODE_BADATTRIBUTEIDINVALID if url not starting with 'opc.tcp://', UA_STATUSCODE_GOOD on success
  */
 UA_StatusCode UA_EXPORT UA_EndpointUrl_split(const char *endpointUrl, char *hostname, UA_UInt16 * port, const char ** path);
+
+/**
+ * Convert given byte string to a number. Returns the number of valid digits.
+ * Stops if a non-digit char is found and returns the number of digits up to that point.
+ * @param buf
+ * @param buflen
+ * @param number
+ * @return
+ */
+size_t UA_EXPORT UA_readNumber(UA_Byte *buf, size_t buflen, UA_UInt32 *number);
 
 #ifdef __cplusplus
 } // extern "C"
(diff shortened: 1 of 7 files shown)


jpfr:
Nice! With the tests and all, this is a posterboy PR. :+1: 
Some last (last) comments before merging.

- Some embedded targets don't have atoi. In `/src/server/ua_services_attributes.c` there is a function
```c
static size_t readNumber(UA_Byte *buf, size_t buflen, UA_UInt32 *number)
```
doing the same job. You could make that an "utility function" and use it in both places.
- Why return the path with a leading slash? Imho most users would have to remove it. If there are use cases needing the slash, we can keep it.

ua_connection.c:236
- The case with and without pathTmp differ only in the value of portLen. Make the if-else part short and do the work in a shared block.
- Is it possible to use memcpy instead of strncpy? That saves the nasty macro switches.

HolgerJeromin:
It is quite common that in an URI the Path includes the leading /

HolgerJeromin:
Your code does not seem to support IPv6 Syntax `opc.tcp://[2001:0db8:85a3::8a2e:0370:7334]:1234/path`

Supporting this syntax is not fun. I did it for the [ACPLT Runtime](https://github.com/acplt/rte/blob/5f7c61c5795083767828b395c7cb49fcdcf3eedb/syslibs/comm/kshttp/source/httpClient_genericClient.c#L153-L321). But this code is not plain C.

Pro:
@HolgerJeromin yes, you are right, IPv6 is not supported.
I'll add it here using a simple state variable, but IPv6 is not supported anywhere else in open62541.

Pro:
@HolgerJeromin IPv6 is now included in the PR (without the username and password notation)

@jpfr this should now be ready to merge.

We maybe should add somewhere a check if the hostname is `IPv6` and then show an error message, that IPv6 is not supported in open62541 (at least for now).

HolgerJeromin:
Nice! You could test for `"opc.tcp://hostname:1234"` and `"opc.tcp://hostname:1234?query"`

Pro:
The unit test for `opc.tcp://hostname:1234` is already there?!
Endpoint urls should not contain query parameters, thus I think we shouldn't care about those. If an URL with query is passed, it will return a bad status code.

jpfr:
Great!
I did some last cosmetic changes (untabify) and squashed into one commit with @Pro as the author.

All merged in 0.2 via 0fdfcc43d467d0ea4bf003733c0fd50549677279.

HolgerJeromin:
Sorry, I missed the test. :-) 
You are right with the query. Good job

PR #782, opened 2016-08-30
Array dimension 0 means variable length

Pro:
Input/Output Arguments may have arrayDimensions set to 0 which means variable length.
See OPC Specification Part 3, page 61

So it should be fixed like this?


jpfr:
You're right. :smile: 
Merged.

PR #796, opened 2016-09-14
DataValue and independent datatype, valuerank, array dimensions in variable node

jpfr:
In order to support all CTT tests, this PR brings the following changes

- variable node (and variabletype node) contain a datavalue instead of only a variant
- datatype, valuerank, array dimensions can be set independently from the value itself
- the constraints imposed by datatype, valuerank, array dimension are checked for every
   - value write (checked against the same variable)
   - variable instantiation (checked against the variabletype)
   - write of datatype, valuerank, array dimension (internal consistency and with the variabletype)
- better documentation of datatype, valuerank, array dimension


jpfr:
This PR solves the last open CTT issues...

PR #798, opened 2016-09-16
Compile on OpenBSD

mklingn:
I had to introduce the modifications in commits ceb27d6 and 0932f47 to be able to compile on OpenBSD 5.8 without warnings. 
Peculiarities of the build process on OpenBSD are described in building.rst commited in 529bb14 and 
7a2071b.
Furthermore fixed a typo in doc/tutorial_noderelations.rst (a343e49) and modified some flag names also in there (6fff64c) that seem to have changed (?) - not entirely sure the latter is complete.

### DIFF
--- a/plugins/ua_network_tcp.c
+++ b/plugins/ua_network_tcp.c
@@ -42,6 +42,9 @@
 # ifdef __QNX__
 #  include <sys/socket.h>
 # endif
+# ifdef __OpenBSD__
+#  include <sys/socket.h>
+# endif
 # ifndef __CYGWIN__
 #  include <netinet/tcp.h>
 # endif
--- a/plugins/ua_network_tcp.c
+++ b/plugins/ua_network_tcp.c
@@ -51,7 +51,7 @@
 #endif
 
 /* unsigned int for windows and workaround to a glibc bug */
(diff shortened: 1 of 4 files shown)


plugins/ua_network_tcp.c
```diff
@@ -42,13 +42,16 @@
 # ifdef __QNX__
 #  include <sys/socket.h>
 # endif
+# ifdef __OpenBSD__
+#  include <sys/socket.h>
+# endif
```
HolgerJeromin: Perhaps we could have something [like that](http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#BSD)
````
#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
#include <sys/param.h>
#if defined(BSD)
	include<sys/socket.h>

#endif
#endif
````
This would support DragonFly BSD, FreeBSD, OpenBSD, NetBSD and Apple OSX/iOS

jpfr: Sounds good.
If we are aren't aware of differences between the BSD.

Do you want to prepare a PR?

jpfr:
Thanks for the PR!

plugins/ua_network_tcp.c
HolgerJeromin: Done #807

Issue #800, opened 2016-09-19
Stress testing the server causes crash


StalderT:
Hello,

I have add to server.c :
```
static UA_StatusCode
readTest(void *handle, const UA_NodeId nodeId, UA_Boolean sourceTimeStamp,
             const UA_NumericRange *range, UA_DataValue *value) {
    if(range) {
        value->hasStatus = UA_TRUE;
        value->status = UA_STATUSCODE_BADINDEXRANGEINVALID;
        return UA_STATUSCODE_GOOD;
    }   
    UA_UInt32 val = (UA_UInt32)handle;
    UA_DateTime currentTime = UA_DateTime_now();
    UA_Variant_setScalarCopy(&value->value, &val, &UA_TYPES[UA_TYPES_UINT32]);
    value->hasValue = UA_TRUE;
    if(sourceTimeStamp) {
        value->hasSourceTimestamp = UA_TRUE;
        value->sourceTimestamp = currentTime;
    }   
    return UA_STATUSCODE_GOOD;
}
```
and
```
    for(UA_UInt32 i=0; i<10000; i++)
    {
            UA_DataSource dateDataSource_test = (UA_DataSource) {.handle = (void*)i, .read = readTest, .write = NULL};
            UA_VariableAttributes v_attr_test;
            UA_VariableAttributes_init(&v_attr_test);
            char name[100];
            snprintf(name, 100, "TEST%04d", i);
            v_attr_test.description = UA_LOCALIZEDTEXT("en_US",name);
            v_attr_test.displayName = UA_LOCALIZEDTEXT("en_US",name);
            const UA_QualifiedName dateName_test = UA_QUALIFIEDNAME(1, name);
            UA_NodeId dataSourceId_test;
            UA_Server_addDataSourceVariableNode(server, UA_NODEID_NULL,
                                                UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER),
                                                UA_NODEID_NUMERIC(0, UA_NS0ID_ORGANIZES), dateName_test,
                                                UA_NODEID_NULL, v_attr_test, dateDataSource_test, &dataSourceId_test);
    }
```

If I add 4000 test variables to Data Access View in UaExpert, open62541 server crash : 

```
==11316== Memcheck, a memory error detector
==11316== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==11316== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
[...]

PR #804, opened 2016-09-20
Update ua_connection.c

StalderT:
fix https://github.com/open62541/open62541/issues/800


jpfr:
Great find!
Thank's for the PR.

Issue #810, opened 2016-09-21
Add features table to webpage


Pro:
What is your opinion about adding a table on the main webpage listing all the supported and non-supported features of open62541?

Additionally it may include information about the supported service sets, server profiles, multithreading support, ...

Currently it's not that easy for beginners to see what is supported and what not.

jpfr:
👍 

I can add a list of supported conformance units from the CTT.

We could take the list form node-opcua as the starting point:
https://github.com/node-opcua/node-opcua

Pro:
Uh, nice! node-opcua has exactly what I meant.
So probably it's also best to include the table into the README.md or a separate markup file to keep it consistent with the corresponding branch/code-base. Then on the webpage we may just include this file from current main branch.

I'll create a PR for the first draft.

Pro:
@jpfr I created the PR. You can directly push to my branch

PR #811, opened 2016-09-21
Features Table

Pro:
First draft for the features table. See https://github.com/open62541/open62541/issues/810

I took the supported services from `services.h`

For the lower table please check it carefully, since I just C&P it from node-opcua.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -1,0 +1,110 @@
+open62541 Supported Features
+============================
+
+
+## Supported Features
+
+| __**Service**__             |                                 |                      | Comment              |
+|:----------------------------|:--------------------------------|:--------------------:|:---------------------|
+| Discovery Service Set       |                                 |                      |                      |
+|                             | FindServers()                   |  :white_check_mark:  |                      |
+|                             | FindServersOnNetwork()          | :first_quarter_moon: | Branch: feature_mdns |
+|                             | GetEndpoints()                  |  :white_check_mark:  |                      |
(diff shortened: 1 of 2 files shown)


FEATURES.md
```diff
@@ -0,0 +1,110 @@
...
+## Supported Features
+
+| __**Service**__             |                                 |                      | Comment              |
+|:----------------------------|:--------------------------------|:--------------------:|:---------------------|
+| Discovery Service Set       |                                 |                      |                      |
+|                             | FindServers()                   |  :white_check_mark:  |                      |
+|                             | FindServersOnNetwork()          | :first_quarter_moon: | Branch: feature_mdns |
+|                             | GetEndpoints()                  |  :white_check_mark:  |                      |
+|                             | RegisterServer()                |  :white_check_mark:  |                      |
+|                             | RegisterServer2()               |     :full_moon:      | Branch: master       |
```
markusgraube: What is RegisterServer2()?
This isn't a defined service in the spec and needs further explanation

jpfr: It comes with OPC UA v1.03

> This Service allows a Server to register its DiscoveryUrls and capabilities with a Discovery Server.
It extends the registration information from RegisterServer with information necessary for
FindServersOnNetwork. This Service will be called by a Server or a separate configuration utility.
Clients will not use this Service.

jpfr:
Thanks for going ahead with this!
Merged.

FEATURES.md
markusgraube: Okay, thank you for the info

Issue #802, opened 2016-09-19
0.2 release is nigh! Last action items coordination


jpfr:
Hey there!

0.2 is feature complete. The last CTT tests of the implemented features are green.
Can we have a list of open points and who is working on them?

**The following issues need a fix / wontfix / wontfix now decision**
- [x] #666 Multiple subscriptions broken?
- [x] #660 Retransmission queue keeps growing when no acknowledgement is received 0.2 
- [x] #655 stress test causes loop (in client or server)
- [x] #632 The first session GUID is always the same
- [x] #509 Verify tutorials

**Verify / Update release packaging**
- Do we want to distribute the PDF documentation in every zip?
- Should we have a single-file release zip with the header + source + pdf docs?

**Update copyright header year**
- And update the AUTHORS file

**Write the release statement**

Please add your open points to the list. And add yourself to assignments. :+1:

ichrispa:
Don't quite agree with you here @jpfr. There are a lot more points on the todo list and I want to stress that Assignees are the ones who broke features, then made them experimental instead of fixing them.

You'll get a full list tomorrow.

jpfr:
Ok. Now is the time.

As said before, the best way to protect against regressions are unit tests.
CTT and unit tests are all green.
Let's take the opportunity and add some more tests where you experienced breakage.

ichrispa:
OK, mostly done.

Sry for being a bit pissed off @jpfr; the stricter typeDef handling managed to break a backend generator (in addition to the namespace compiler) that tied into several projects. That caused quite a bit of havoc and a lot of work.

We have the namespace compiler back. Before 0.2, I would like to get that one out of the experimental category. Generating complex object models based on NS0 and Di/FSI is quite a common use-case for me. 
[] Make namespace-compiler none-experimental
[] Re-Add Travis builds against full generated NS0 (check duplicate nodes in server, propper dataType tree)

I also expanded/fixed instantiation as current projects are heavily based on complex information models. Here's what is covered at the moment:

* We can instatiate object nodes from typedefs (obviously)
* We can instatiate superTypes (inheritance) along with objects:
```
Type:
+ MamalType
   v- Class = "mamalia"
   v- Species
   + Dog
      v Name

Instance of Dog:
+ Dog
  v- Class = "mamalia"
  v- Genus
  v= Name
```
* We can mask superType attributes
```
Type:
+ MamalType
   v- Class  = "mamalia"
   v- Species
   + Dog
      v- Species = "Canis"
[...]

PR #812, opened 2016-09-22
WIP: Add Test, Documentation and Example for instantiation and inheritence

markusgraube:
Following @ichrispa comments from #802, I would like to add  tests, documentation and an example for instantiation and inheritence in open62541


jpfr:
Nice!
Now that we have an example for inheritance, we could split the tutorial_noderelations into a part for inheritance and a part for XML nodeset generation.

Should be easy to do. And the example can be included in the docs with literalinclude as is done for the other tutorials.
So that the examples in the documentation are checked by travis to build.

markusgraube:
Yeah, that is the plan I am currently working on :)

PR #815, opened 2016-09-23
Add Test, Documentation and Example for instantiation and inheritence

markusgraube:
The rest of #812 

I tried to align examples, documentation and tests.

Is it now easier to understand the instantiation?


markusgraube:
Hmm, it seems, that we have to enable -DENABLE_GENREATE_NAMESPACE0 on Travis. What do you think?

jpfr:
It would be better, if the example worked with the precompiled dll/.so from our release.

When I last checked, it was possible to generate only new nodes.
And to use the ns0 bootstrapping in ua_server.c

Do you want to show an example where ns0 is bootstrapped by the nodeset compiler as well?

Pro:
@markusgraube if you want to create a specific unit test for the ns0 initialization, you could also use the following approach:

https://github.com/Pro/open62541/commit/950d078a0e903cd15f9e8567df08912f8979438e

It first runs all the other unit tests without ns0 and then the specific instantiation check with ns0.

Pro:
@markusgraube any update on this?
It would be really nice if we can add your tutorials!

markusgraube:
sorry, I have just forgotten that this task is still open. I will find some time for this in the next week

markusgraube:
closed in favor of #1169

PR #821, opened 2016-09-27
Fix return value

Lykurg:
UA_FooResponse_deleteMembers(&response) does not delete
response.responseHeader.serviceResult but it will initialize it. Thus
the return value is always UA_STATUSCODE_GOOD even if an error occurred.

Regression of 1cc01277d05e8caf0052534a505a7d6d95dd0740


jpfr:
Thanks!
Merged.

PR #831, opened 2016-09-30
Nodestore interface update

JGrothoff:


JGrothoff:
So, i reworked these changes in PR #832.

jpfr:
So when the nodestores are user-visible, then ua_nodes.h is becomes a public header?

JGrothoff:
Yes correct. So should i move ua_nodes to include and change the CMakeList?

PR #832, opened 2016-10-01
Added Nodestore switch

JGrothoff:
Added a new "layer" between the nodestore. It switches the nodestore by the namespace index.
So each namespace can have its own nodestore. The nodestore can be specified and changed via UA_Server_addNamespace_Nodestore(...) similar to UA_Server_addNamespace(...)

Compatibility with namespace parser is not  realy established yet: All nodes will be created in nodestore 0. (As default nodestore 0 an nodestore 1 are the same)


jpfr:
Hey @JGrothoff ,

We currently have multi-threading support in open62541, although marked as experimental.
And at least in Karlsruhe and Dresden, people rely on that feature. So we should have a plan to keep it alive even with your improvements.

The current multicore nodestore relies on RCU (a mechanism of the Linux kernel) to know when it is safe to free an old (deleted/replaced) node, i.e. to know when no other thread has a pointer to the node. The negative side of RCU is it had to be deeply integrated into the server implementation beyond the nodestore. That's where all those UA_RCU_LOCK macros come from.

With plugable nodestores, some users will want to use/write multithreading-capable versions. The simplest way is probably some checkout/release mechanism for nodes with reference counting.
With an additional `void (*UA_NodestoreInterface_release)(void *nsHandle, const UA_Node *node)` nodestore callback. That is more generic and does not rely on some architecture-specific mechanisms like RCU.

Adding a reference counter to the node-entries should be simple.
Just make it an unsigned int with some `0x01 << 16` liveness bit or similar.
So the node can be safely deleted when the counter + liveness bit drops to zero.

Counters can be made atomic if necessary, such as 
```c
#define UA_COUNTER_INCREASE(COUNTER) do { COUNTER++; } while(false)
#else
#define UA_COUNTER_INCREASE(COUNTER) __sync_fetch_and_add(&COUNTER, 1)
#endif
```
Switching pointers (replace) is can be made atomic with __sync_val_compare_and_swap.

If you have questions, do not hesitate to ask or call!

JGrothoff:
Hey @jpfr, thanks for the comment and ideas for multithreading support.
I am a bit busy this and the next week and will work on that after 16th of october.

jpfr:
Hey @JGrothoff,

I added support for atomic operations in this commit:
5467bae2dd7a087ba5fa18b4718b3553e4c9f0a9

As said before, even though OPC UA nodes are atomic and can only be replaced, memory reclamation can be tricky.

Our concurrent nodestore uses RCU (https://lwn.net/Articles/262464/) for memory reclamation and needs no locking. So every thread is as fast as possible and never waits. However, that does not make a good plugin, since RCU is intrusive in the rest of the library.

One possibility is to instead use reference counting in the nodes for memory reclamation.
To make things easy, we could have per-node locks in the first version.

If you are interested in lock-free memory reclamation, here are two papers with pointers to more literature:
http://www.cse.chalmers.se/~tsigas/papers/LFMemRec-TPDS.pdf
http://www.cs.technion.ac.il/~erez/Papers/cohen-petrank-oopsla15.pdf

JGrothoff:
Hej, thanks for your comments @jpfr.

I fixed the **memory leaks**.

The problem with the "**other pointers to the node**" was: During Service_AddNodes_existing the pointer to the node, which should be added, was read after an edit_node (and therefore a replace in the nodestore) operation.
This wasn't a problem before, because the "dirty cast" in the edit_node function didn't swap the pointer. I changed this to the multithreading aproach with getCopy and replace for full compatibilty with possible nodestores.

I will sum up the changes and put them into the nodestore_switch_final #900.

So now every getNode has a corresponding **releaseNode** function. The next step would be to rewrite the nodestore_concurrent with reference counting mechanisms. Afterwards the RCU Locks and unlocks can be removed.

Some weeks ago i wrote a small **mysql nodestore** prototype. If this is an interesting example, i can finish it and integrate it into the examples. I also wrote a small SQL script for the mysql nodestore that loads the basic information model from the XML file. (Needs only some seconds to load, the parser is not needed and can be done during runtime. It is also possible to dump the sql database to an xml file again, but i'm not sure if this is compliant with the 61499 nodeset xml description.) Probably there will be some issues to resolve, but i think we could handle that. Another interesting nodestore could be based on an relational database like mongodb.

jpfr:
Nice! 😄 
That's a really good progress.

The mysql nodestore sounds cool. I'd love to have sth like that in our examples.
But it's a bit difficult to integrate in our standard testsuite since it requires users to set up a dedicated mysql server, and so on.

If you are thinking about different backends anyway, you could try sqlite. https://sqlite.org/
Porting should be straightforward. Then we only need
- the installed sqlite library or
- download the single-file distribution of sqlite during the build process.
The rest can be handled "locally". And we can add a .sqlite file with a pre-allocated database to our repository.

jonasgreen88:
Hi,

What is the status of this PR? Is this nodestore plugin solution supposed to replace the UA_ENABLE_EXTERNAL_NAMESPACES that was removed from the master branch recently?

Br, Jonas

jpfr:
Merged via #1168

PR #822, opened 2016-09-28
Various fixes for pyUANamespace compiler

Pro:
Smaller bug fixes for namespace compiler. See commits for further details.

PR #837, opened 2016-10-07
NS Compiler: check if type exists

### LINKED ISSUE #835: Dependency ordering in namespace compiler
The pyUANamespace compiler does not yet order the nodes based on their dependency, i.e., it just prints out the order from the XML file.

This causes the resulting source code to fail, e.g., for Namespace0 the XML contains on line 1183:

```
<UAObject NodeId="i=2009" BrowseName="ServerCapabilities" ParentNodeId="i=2004">
    <DisplayName>ServerCapabilities</DisplayName>
    <Description>Describes capabilities supported by the server.</Description>
    <References>
      <Reference ReferenceType="HasProperty">i=3086</Reference>
      <Reference ReferenceType="HasProperty">i=3087</Reference>
      <Reference ReferenceType="HasProperty">i=3088</Reference>
      <Reference ReferenceType="HasProperty">i=3089</Reference>
      <Reference ReferenceType="HasProperty">i=3090</Reference>
      <Reference ReferenceType="HasProperty">i=3091</Reference>
      <Reference ReferenceType="HasProperty">i=3092</Reference>
      <Reference ReferenceType="HasComponent">i=3093</Reference>
      <Reference ReferenceType="HasComponent">i=3094</Reference>
      <Reference ReferenceType="HasTypeDefinition">i=2013</Reference>
      <Reference ReferenceType="HasComponent" IsForward="false">i=2004</Reference>
    </References>
  </UAObject>

```

Which has the referene `HasTypeDefinition=2013`, but the type definition for `i=2013` is on line 1794:
```
  <UAObjectType NodeId="i=2013" BrowseName="ServerCapabilitiesType">
    <DisplayName>ServerCapabilitiesType</DisplayName>
    <Description>Describes the capabilities supported by the server.</Description>
    <References>
      ...
    </References>
  </UAObjectType>
```

So the namespace initialization fails and creates an invalid information model.

@ichrispa, @Stasik0 
Is there already some code which builds the dependency graph and makes sure the order is correct?
If not, does it make sense to implement a reordering of the nodes based on the dependency graph.
The following fields indicate dependencies between nodes:
- ParentNodeId
- References

Pro:
The namespace compiler also needs to check if the (parent) ObjectTypeNode exists. Otherwise the instatiation of the node will fail and cause follow-up errors.

Fixes #835 at least for now


Pro:
Mostly integrated into https://github.com/open62541/open62541/pull/822

PR #838, opened 2016-10-07
Fix: Safe access to NodeId

Pro:
`servernode->nodeId` can't be reused because it may be deleted in addNodeInternalWithType if the node could not be added. Thus it may lead to an invalid memory access


jpfr:
Merging this. Good find!
In general, however, we should change the existing nodes to datasources inline.
Instead of removing and readding.

jpfr:
Hum. Readdding might be necessary to get the correct nodeids.

PR #739, opened 2016-08-03
WIP: Nodeset compiler for v0.3

jpfr:
This PR merges #725 and #729.
Furthermore, I started to refactor the open62541 code generation backend into a separate module.
So the compiler stages are represented in the code architecture.

PR #841, opened 2016-10-10
Improve variable handling

jmrahm:
The request improve datatype handling in your code. For data handling (write/read) in some clients, it is necessary to set the right Datatype in the Datatype-Node. Since some weeks ago there is no automatism to do that only BaseDataType is set. Especially for array's in UA Expert is the right datatype required. Hence I add some lines to set the right Datatype from variant type. Additionally the correct ValueRank is set, depending on the Variant value.

Up to now you have to set the variable datatype and valuerank in the code especially in UA_VariableAttributes. In case of your "server_nodeset" example it is more comfortable with this request, obviously there is no easy way to change the datatype of UA_VariableAttributes in generated code. As of now the request will do that for you.

### DIFF
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -525,18 +525,43 @@ copyCommonVariableAttributes(UA_Server *server, UA_VariableNode *node,
     /* Set the constraints */
     if(!UA_NodeId_isNull(&attr->dataType))
         retval  = UA_VariableNode_setDataType(server, node, vt, &attr->dataType);
-    else /* workaround common error where the datatype is left as NA_NODEID_NULL */
-        retval = UA_VariableNode_setDataType(server, node, vt, &vt->dataType);
+	else { // set Datatype, kown from variante datatype 
+		if(&attr->value.type->typeId != NULL) {
+    	    retval = UA_VariableNode_setDataType(server, node, vt, &attr->value.type->typeId);
+		}
+		else {
+			// set BaseDatatype
+			retval = UA_VariableNode_setDataType(server, node, vt, &vt->dataType);
+		}
+	}
         
     node->valueRank = -2; /* allow all dimensions first */
     retval |= UA_VariableNode_setArrayDimensions(server, node, vt,
                                                 attr->arrayDimensionsSize,
                                                 attr->arrayDimensions);
 
-    if(attr->valueRank != 0 || !UA_Variant_isScalar(&attr->value))
-        retval |= UA_VariableNode_setValueRank(server, node, vt, attr->valueRank);
-    else /* workaround common error where the valuerank is left as 0 */
-        retval |= UA_VariableNode_setValueRank(server, node, vt, vt->valueRank);
+	/* All ValueRank cases can be handle with this code, 
+	 * expect -3 (ScalarOrOneDimension) -> there is no special dynamic usecase for this
+	 */
+	UA_Int32 valueRank = -2; // If no rank fit
+    if(attr->valueRank == 0 && UA_Variant_isScalar(&attr->value)) {
+		// Scalar
+		valueRank = -1;
+	}
+	else {
+		if(attr->arrayDimensionsSize == 1) {
+			// OneDimension
+			valueRank = 1;
+		}
+		else {
+			if(attr->arrayDimensionsSize > 1) {
+				// OneOrMoreDimensions
+				valueRank = 0;
+			}
+		}
+	}
+	// Write ValueRank to Node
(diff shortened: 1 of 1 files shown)


jpfr:
Hey Julian,

Note that there is a difference between the DataType attribute and the actual type of the value.
That is because the DataType attribute can point to an abstract data type in the hierarchy. Such as `Number` or `BaseDataType`.

The current workarund is that we take the DataType attribute of the VariableType (with a hasTypeDefinition reference) if DataType == NODEID_NULL. If however a DataType is defined, it must be a subtype of what the VariableType requires.

I am quite satisfied with the way datatypes are currently handled.
Still, you say your code improves the server example. Can you show a snippet of these improvements?
Maybe there is a nice middle ground for our implementation.

jmrahm:
I see, DataType can be more generic. Ok, maybe it's more a workaround for myself. Let's describe my problem more in detail.
If you add a Node with an arry variable, let's say in the "server_variable" example like:

```C   
    UA_VariableAttributes attr;
    size_t arrayLength = 10;
	int32_t varArray[arrayLength];
	int32_t* newValue = varArray;
	UA_VariableAttributes_init(&attr);
	UA_Variant_init(&attr.value);
	UA_Variant_setArrayCopy(&attr.value, newValue, arrayLength, &UA_TYPES[UA_TYPES_INT32]);	
		
    attr.description = UA_LOCALIZEDTEXT("en_US","the question");
    attr.displayName = UA_LOCALIZEDTEXT("en_US","the question");

    /* 2) Add the variable node to the information model */
    UA_NodeId myIntegerNodeId2 = UA_NODEID_STRING(1, "the.question");
    UA_QualifiedName myIntegerName2 = UA_QUALIFIEDNAME(1, "the question");
    UA_NodeId parentNodeId2 = UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER);
    UA_NodeId parentReferenceNodeId2 = UA_NODEID_NUMERIC(0, UA_NS0ID_ORGANIZES);
    UA_Server_addVariableNode(server, myIntegerNodeId2, parentNodeId2,
                              parentReferenceNodeId2, myIntegerName2,
                              UA_NODEID_NULL, attr, NULL, NULL);
```
Some weeks ago the right type in Datatype was setted by the Variant type. All works well for me. But now, if you connect with the server via UA Expert there is no possibility to change any value in the array and the DataType is "BaseDataType". If you set the right Datatype by code like:
```C 
attr.dataType = UA_TYPES[UA_TYPES_INT32].typeId;
```
All works well again. The point is, I have to care about to sources for the Type (DataType and Variant Type). It's ok, but lousy.

And now let's go to the "server_nodeset" example. Currently there is no possibility to set the right DataType automatically. Because in the generated file is no line like obove. Hence the generic DataType "BaseDataType" is set. Maybe you should improve the NamespaceCompiler and add this line above to prevent this?! 

Is there  a more sensible usecase to set not the DataType to the type of the variant? I can't think of any. Is it useful to set the Dataype to the Variant, if it exist, and in other cases to an generic/reference or something else. This is what my code actually do (Only by creation of the node, it should may expand and should set also my changes). 
What do you think about? Is it a Issue for all or only for me? :-)

jpfr:
> ```c
attr.dataType = UA_TYPES[UA_TYPES_INT32].typeId;
```
Your solution seems adequate.

To give you a reference, part 3, §5.8.3 says
> Abstract DataTypes can be used in the AddressSpace, i.e. Variables and VariableTypes can point with their DataType Attribute to an abstract DataType. However, concrete values can never be of an abstract DataType and shall always be of a concrete subtype of the abstract DataType.

Concering the NodeSet XML.
Have you tried adding an attribute "DataType" to the NodeSet node description?
It seems the code for this is in place.
https://github.com/open62541/open62541/blob/0.2/tools/pyUANamespace/ua_node_types.py#L1033

jmrahm:
Hi Julius,

yes I tried.
Also in your example "server_nodeset.xml" is a "DataType" attribute set.
`<UAVariable DataType="Double" ParentNodeId="ns=1;i=1001" NodeId="ns=1;i=6001" BrowseName="1:Var1" UserAccessLevel="3" AccessLevel="3">`

But nevertheless the datattype is ignored.

The point is, should it be a feature in namespace compiler to set the datatype or in the open62541 code? The advantages to set it in open-code, you can be sure the right type is set.

jpfr:
bump @ichrispa ,
the nodeset compiler does not print the DataType attribute of nodes.
Can we fix the issue over there?

Pro:
The DataType attribute is now printed by the nodeset compiler. See #739

Issue #847, opened 2016-10-12
Status Code to error message


Pro:
Is there already something like `strerror(...)` for UA_StatusCode?

Currently I only output the number, e.g. `0x80010000` but it would be nice to print an error message like:
`BadUnexpectedError: An unexpected error occurred (0x80010000)`

For that we need a global method which converts error numbers to strings, take from https://opcfoundation.org/UA/schemas/Opc.Ua.StatusCodes.csv

If it is not already there I would suggest something like this
https://opensource.apple.com/source/gcc/gcc-926/libiberty/strerror.c (it is public domain)
and a compiler switch to disable/enable the error table.

Any suggestions for the method name?
- Same as stdlib: UA_Strerrno(statuscode) and UA_Strerror(statuscode)
- UA_StatusCode_Msg(statuscode) and UA_StatusCode_Name(statuscode)

Should I go on with it and implement it?

jpfr:
Sounds good.
I propose returning static char strings like
```c
const char * UA_StatusCode_msg(UA_StatusCode code);
const char * UA_StatusCode_name(UA_StatusCode code);
```
That makes it easy to use the names in logging.

Writing a small generator script for the csv file might save time in the long run.
See also [/tools/generate_nodeids.py](https://github.com/open62541/open62541/blob/0.2/tools/generate_nodeids.py).

Stasik0:
But please make it configurable via cmake since it will blast the binary size...

Pro:
Yep, that was my intention, because it will definitely increase the binary size and not everyone needs it

Pro:
@jpfr should there also be a fallback version.

E.g. if a developer writes in his high-level code:
```
printf("Client connect failed: %s - %s", UA_StatusCode_name(retval), UA_StatusCode_msg(retval));
```
and open62541 isn't compiled with `UA_ENABLE_STATUSCODE_MSG` this will fail. But maybe the developer wants to support both versions, i.e. he has to write:

```
#ifdef UA_ENABLE_STATUSCODE_MSG
	    printf("Client connect failed: %s - %s", UA_StatusCode_name(retval), UA_StatusCode_msg(retval));
#endif
```

Would it be better to define a method which just returns the error code if `UA_ENABLE_STATUSCODE_MSG=OFF`

I.e:

```
#ifndef UA_ENABLE_STATUSCODE_MSG
char *code[11];
const char *UA_StatusCode_name(UA_StatusCode s) {
    sprintf(code, "0x%08x", s);
[...]

PR #849, opened 2016-10-13
Add strerror like function for UA_StatusCode

Pro:
See #847


Pro:
@jpfr ready to merge

jpfr:
Good job!
I made some last changes that live in
https://github.com/open62541/open62541/tree/Pro-feature/statuscode_msg

- Never return NULL but a generic "unknown" description when the code is unknown.
- The UA_StatusCode_description functions always exists. And returns always "unknown" when support is disabled. Reducing complexity of the interface is worth a few bytes overhead, imho.
- Export a single function, do the rest with one-line inline functions
- Distribute functionality across less files. Add UA_StatusCodeDescription and function signatures to ua_types.h.
- The table is static and cannot be seen outside of ua_statuscode_descriptions.c

Please review/pull/improve and we can merge from this PR.

Pro:
Looks very good. You introduced some nice improvements!
No further comments, can be merged

PR #855, opened 2016-10-17
Fixed UA_Client_disconnect() if client is in error state

Root-Core:
If the client runs into an error state, we can't disconnect.
As all states, but UA_CLIENTSTATE_READY are connected, we should be able to disconnect (which is necessary to renew the connection - as required by the error client states).

### DIFF
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -682,7 +682,7 @@ UA_Client_connect(UA_Client *client, const char *endpointUrl) {
 }
 
 UA_StatusCode UA_Client_disconnect(UA_Client *client) {
-    if(client->state != UA_CLIENTSTATE_CONNECTED)
+    if(client->state == UA_CLIENTSTATE_READY)
         return UA_STATUSCODE_BADNOTCONNECTED;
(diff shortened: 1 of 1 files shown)


jpfr:
What should the client state be after the disconnect?
Imho, it should not stay UA_CLIENTSTATE_CONNECTED.

Root-Core:
Right, if an error occurred it is set to `UA_CLIENTSTATE_ERRORED` or `UA_CLIENTSTATE_FAULTED`.
As the comments in [ua_client.h#L60](https://github.com/open62541/open62541/blob/0.2/include/ua_client.h#L60) say, we should disconnect. This is just not possible atm.

The only lifetime state, not connected, should be `UA_CLIENTSTATE_READY`. All others might be connected.
It seems to be checked later in the `UA_Client_disconnect` function.

Atm it is possible, that the client leaves the connections open by reconnecting, until all handles of the OS are used.

jpfr:
Rebased and merged via #953.

jpfr:
Thanks for the PR!

Issue #844, opened 2016-10-11
Add userland ability to not instantiate nodes


ichrispa:
Hi all,

the latest couple of issues like #836, #835 lead to an interesting problem. The XML file describes a snapshot of an exitent node hierarchy. When we create information modells in this way, we do not want to instantiate nodes - they are already instantiated by the model. Modellers like UAModeller infact create an instance tree automatically when a type is selected.

Further more, when we try to instantiate a model component, we usually muck up: the nodeIds are fixed by the model, but we choose our own. 

In short, we might end up with duplicate DataType/ObjectType structures. Or worse, creating the model might fail all together.

I would very much like to have a way to keep the stack from instantiating types when informationmodels are created. To that end, I would propose adding a switch to AddNode_existing: A boolean value inidicating that no instantiation should take place. 

I would append two functions to the highlevel API: addObjectNode_noInstantiate() and addVariableNode_noInstantiate(). Those would then become the default creation functions for i.e. the namespace compiler.

@jpfr, @markusgraube, @Stasik0: Would that be a viable option? Any counter-proposals on how to solve this problem?

Best regards,

ichrispa

jpfr:
You are right. This requires a good solution that sets the case once and for all.

addObjectNode_noInstantiate is fine for me.
In addition, I propose we also expose its complement, addObjectNode_finalize, to finish the instantiation, call the constructor callback, ... at a later time.

With your code that matches children by their browseName, doing the instantiation at a later time (after some children have been added manually) should be unproblematic.

Stasik0:
+1 for noInstantiate

markusgraube:
+1 this seems to be a good solution

ichrispa:
I saw a branche pop up... is anyone actively working on this?

FlorianPalm:
To get my demo working with the current master, I have added flag to distinguish if an instance should be created not ( addObjectNode_noInstantiate(), addVariableNode_noInstantiate()). Seems to work so far. I am going to clean it and create a pull request.

Pro:
There are now the `_begin`and `_finish` methods. See referenced issues

PR #857, opened 2016-10-18
Add doc for _begin and _finish methods

FlorianPalm:
Added "UA_Server_addVariableNode_noInstantiation" and "UA_Server_addObjectNode_noInstantiation" to prevent duplication of child-Elements created by the instantiation procedure.

### DIFF
--- a/include/ua_server.h
+++ b/include/ua_server.h
@@ -695,7 +695,24 @@ __UA_Server_addNode(UA_Server *server, const UA_NodeClass nodeClass,
                     const UA_NodeAttributes *attr,
                     const UA_DataType *attributeType,
                     UA_InstantiationCallback *instantiationCallback,
-                    UA_NodeId *outNewNodeId);
+                    UA_NodeId *outNewNodeId,
+                    UA_Boolean instantiate);
+
+static UA_INLINE UA_StatusCode
+UA_Server_addVariableNode_noInstantiation(UA_Server *server, const UA_NodeId requestedNewNodeId,
+                          const UA_NodeId parentNodeId,
+                          const UA_NodeId referenceTypeId,
+                          const UA_QualifiedName browseName,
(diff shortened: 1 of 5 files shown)


HolgerJeromin:
refers to #844

jpfr:
Hey @FlorianPalm ,
good to have the noInstantiation code!

Imho, we are only doing this to let people add child nodes with known nodeids.
So the children do not get a random nodeid during instantiation.

The missing part is to run the instantiation eventually.
Can we make it an
`Service_addNodes_begin` and
`Service_addNodes_finish` pair?
That makes it clear that _finish is supposed to be run after the manual adjustments.

Furthermore, in /src/server/ua_server.c, a lot of function calls need the additional 'true'. But there is only one 'false' case. Imho the 'false' case could be handled without the convenience function. That keeps the signature nicer.

Pro:
@jpfr should we close this in favour of #882 ?

Pro:
@jpfr is this already in master with the `_begin` and  `_finish` methods?

jpfr:
Yes. Master has
```c
UA_StatusCode UA_EXPORT
UA_Server_addNode_begin(UA_Server *server, const UA_NodeClass nodeClass,
                        const UA_NodeId requestedNewNodeId,
                        const UA_QualifiedName browseName,
                        const UA_NodeId typeDefinition,
                        const void *attr, const UA_DataType *attributeType,
                        void *nodeContext, UA_NodeId *outNewNodeId);

UA_StatusCode UA_EXPORT
UA_Server_addNode_finish(UA_Server *server, const UA_NodeId nodeId,
                         const UA_NodeId parentNodeId,
                         const UA_NodeId referenceTypeId,
                         const UA_NodeId typeDefinitionId);
```

Let's keep this issue open to add better docs + an example for using these.

PR #858, opened 2016-10-18
fix LOG problem

TorbenD:
In function UA_Server_processSecureChannelMessage in ua_server_binary.c there was a problem with the UA_LOG_TRACE_CHANNEL. I fixed it in this commit!


HolgerJeromin:
Should we raise `UA_LOGLEVEL` on Travis to find such bugs?

jpfr:
We could raise the loglevel in a build that is not used to prepare release zips

PR #860, opened 2016-10-19
DataTypes may have empty members (e.g. DiscoveryConfiguration)

Pro:
Therefore the members variable has to be defined as NULL instead of empty array


jpfr:
```python
+        if len(self.members)==0:
+            return "#define %s_members NULL" % (self.name)
+        else:
```
After the return, we don't need an else case.
Can you remove the unnecessary indendation? The code will be more readable. And the diff will be much smaller / easy to understand.

Pro:
@jpfr done. ready to merge.

jpfr:
That is a really nice diff. :-)

Issue #869, opened 2016-10-29
source/server timestamp question/problem


StalderT:
Hello,

I have add to server.c :

```
static UA_StatusCode
readTest(void *handle, const UA_NodeId nodeId, UA_Boolean sourceTimeStamp,
             const UA_NumericRange *range, UA_DataValue *value) {
    if(range) {
        value->hasStatus = UA_TRUE;
        value->status = UA_STATUSCODE_BADINDEXRANGEINVALID;
        return UA_STATUSCODE_GOOD;
    }   
    UA_UInt32 val = (UA_UInt32)handle;
    UA_DateTime currentTime = UA_DateTime_now();
    UA_Variant_setScalarCopy(&value->value, &val, &UA_TYPES[UA_TYPES_UINT32]);
    value->hasValue = UA_TRUE;
    if(sourceTimeStamp) {
        value->hasSourceTimestamp = UA_TRUE;
        value->sourceTimestamp = currentTime;
    }   
    return UA_STATUSCODE_GOOD;
}
```

and

```
  for(UA_Uint32 i=0; i<10000; i++)
    {
            UA_DataSource dateDataSource_test = (UA_DataSource) {.handle = (void*)i, .read = readTest, .write = NULL};
            UA_VariableAttributes v_attr_test;
            UA_VariableAttributes_init(&v_attr_test);
            char name[100];
            snprintf(name, 100, "TEST%04d", i);
            v_attr_test.description = UA_LOCALIZEDTEXT("en_US",name);
            v_attr_test.displayName = UA_LOCALIZEDTEXT("en_US",name);
            const UA_QualifiedName dateName_test = UA_QUALIFIEDNAME(1, name);
            UA_NodeId dataSourceId_test;
            UA_Server_addDataSourceVariableNode(server, UA_NODEID_NULL,
                                                UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER),
                                                UA_NODEID_NUMERIC(0, UA_NS0ID_ORGANIZES), dateName_test,
                                                UA_NODEID_NULL, v_attr_test, dateDataSource_test, &dataSourceId_test);
    }
```

If I monitor TEST0001 with UaExpert, source timestamp and server timestamp always change.

In detectValueChange function, UA_String_equal(encoding, &mon->lastSampledValue) is always equal to 0, "changed" is set to true and value always resend to the client.

How can we change the code to only send the value if the value realy change?

StalderT:
[...]

PR #870, opened 2016-10-29
possible fix

StalderT:
Possible fix for https://github.com/open62541/open62541/issues/869


jpfr:
Hey @StalderT ,

thanks for the PR. It indeed does solve the issue.

Still, I was able to fix the problem at the source.
In the past, we allocated a string of the correct length.
Now, we try to stack-allocate if the encoding is small enough.

See the commit 2a127dc9e361a29cc76428af80544e547262f3ee.

Ok to close?

StalderT:
It works. Thanks

PR #877, opened 2016-11-04
Reuse previously fetched UA_DateTime_nowMonotonic()

Lykurg:


jpfr:
Thanks for the improvement!

PR #878, opened 2016-11-04
Be afraid of premature pessimization, use pre(in|de)crement  ;D

Lykurg:
There were some places where the post(in|de)crement could not be changed, e.g. myArray[i++]. I hope I spotted all cases.

### DIFF
--- a/deps/libc_time.c
+++ b/deps/libc_time.c
@@ -30,7 +30,7 @@ int __secs_to_tm(long long t, struct tm *tm)
     remsecs = (int)(secs % 86400);
     if (remsecs < 0) {
         remsecs += 86400;
-        days--;
+        --days;
     }
 
     wday = (int)((3+days)%7);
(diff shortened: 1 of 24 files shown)


jpfr:
Google's C++ style-guide says the difference between post/pre-decrement is not relevant for integers.
https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

There might be some old compilers that don't optimize the difference away.
But I doubt they are still in use.

On the other hand, post-increment is more readable (at least for me).

I'm undecided about this. Anybody with strong convictions in the core team?

Pro:
IMHO it would be better to use preincrement wherever possible, just to be consistent and you don't need to distinguish between integer or object. There may be some specialized compilers for devices which don't do the optimization.

Stasik0:
merging should not harm, anyways

PR #882, opened 2016-11-09
WIP: Simplify AddNodes

jpfr:


src/server/ua_server.c
```diff
@@ -131,78 +112,72 @@ UA_Server_forEachChildNodeCall(UA_Server *server, UA_NodeId parentNodeId,
     return retval;
 }
 
-static UA_StatusCode
-addReferenceInternal(UA_Server *server, const UA_NodeId sourceId, const UA_NodeId refTypeId,
-                     const UA_ExpandedNodeId targetId, UA_Boolean isForward) {
+static void
+addReferenceInternal(UA_Server *server, UA_UInt32 sourceId, UA_UInt32 refTypeId,
+                     UA_UInt32 targetId, UA_Boolean isForward) {
```
markusgraube: Do I understand it correctly that now this function can only handle namespace 0?
And why do we loose the StatusCodeResult?

The same is true for the other functions, right?

jpfr:
For this to be complete, we need the no-instantiation idea from #857 implemented.
And we need the nodeset compiler to use the no-instantiation code to avoid double members.

jpfr:
I propose to merge / integrate #857 into this PR as follows:

- Split AddNodes into `AddNodes_begin` and `AddNodes_finish`. The latter only does the instantiation, including the creation of missing children.
- Make the change visible to userland (but keep the original AddNodes that does both)
- Make the nodeset compiler use AddNodes_begin for all nodes. Then run AddNodes_finish at the end, once all "manually" defined nodes are in place.
- Use AddNodes_begin/finish in the normal ns0 bootstrap. Remove `deleteInstanceChildren` workaround.

Okay @open62541/open62541-core ?

ichrispa:
Hi @jpfr ,

when merging this branch with 0.2, the majority of our code does *not* run properly (type checks fail, parent is not hierarchical, etc.). This branch would cause havoc if merged with 0.2 in its current state. These errors are mainly caused by the attempt to instantiate objects during runtime. 

We will investigate where the problem comes from in the following days.

Best regards,

ichrispa

jpfr:
Okay. That's exactly why I gave you a heads-up before merging. 😄 
Please tell me if I can be of help in any way.

If there are serious problems on your end, an extreme measure would be to comment out some consistency checks and revive them for 0.3....

Your experience could also be help in making the log messages more expressive. We should assist users in tracking down any problems as good as we can.

ichrispa:
Hi @jpfr,

I took a look at the current implementation - the type checks fail for virtually the entire generated namespace 0. Something is very fishy about the type checks.

Most prerelevant is the fact that "Reference to the parent node is not hierarchical". If these occur for hasSubType and hasComponent (which are hierarchical), then of course nearly nothing can be added to NS0.

The error originates in isNodeInTree(): both hasComponent and hasSubtype are checked to see if they are in the HierarchicalReferences Tree. For both references, the function recurses (former to "aggregates", later to "hasChild"). For both parents the folliwing check in line 171 fails:

`if(!node->references[i].isInverse)
  continue;
`

The hierarchicalReference Parent can never be reached, hence the checks claim that none of the types in NS0 are valid. No types, no namespace, no game.

Even if all type checks are commented out, UA_Server_addNode_finish fails to locate the typeDefinition. I get a SEGFAULT when UA_Server_new() tries to finalize its capabilities node.


In conclusion: The type checks introduced in the pull request are generally valid for a userspace program attempting to initialize its instances. 
@jpfr: If you can identify the reason for the failing isInTree() checks, that would be really cool. I don't understand why the referenceNodes don't have the inverse reference... 

Namespace 0 must fully compile before we can merge it into 0.2...

Best regards,

ichrispa

jpfr:
Hey @ichrispa ,

thanks for pointing out the root cause.
As usual, the initial bootstrap is causing problems.
ns0-minimal is generated just fine. But it looks as if we were just lucky with the genreated ordering.

I propose we move the solution into the 0.3 cycle in order not to generate another blocking issue.
After all, these are just consistency checks and not fundamentally vital to operations.

ichrispa:
Hey @jpfr,

I would second the motion to move this directly to 0.3/master. Additionally, we should take a closer look at sorting-requirements in ua namespaces. This is becoming increasingly non-trivial; I'm going to forward a suggestion to you over a not-so-public-channel on how to tackle this problem.

Regards,

ichrispa

jpfr:
closing this and reopen as a pr to master

Pro:
@jpfr was this already reopened as a new PR to master?

@FlorianPalm 
Should we also delete the stale branches?
https://github.com/open62541/open62541/tree/no_instantiation
https://github.com/open62541/open62541/tree/no_instantiation_02

Issue #884, opened 2016-11-10
Windows Embedded Compact 2013 build


carljoerger:
Hello,
   I've modified the openc62541.c and openc62541.h files to compile as C++ for Windows Embedded Compact 2013 (under VS2015).  Is anyone else interested in this?  I haven't tested them yet.  Please give me any feedback...

[open62541.zip](https://github.com/open62541/open62541/files/584286/open62541.zip)

jpfr:
Hey @carljoerger,

Great work!
Sure we are interested. 😄 

Can you tell us the version you started with so we can make a diff for preparing a commit?

If you want, you can also make a pull request on github, so your name appears in the repository as the author of the commit.

Stasik0:
The files are marked as Git-Revision: v0.2.0-RC1, so probably the last release.

carljoerger:
I want to port it to WEC 7 (VS 2008).
The designated initializers are the main problem (or just porting it to C++).  Is there any chance that they could be changed?

HolgerJeromin:
@jpfr with git you can change the author of a commit yourself. No need to have an own pr to clean credit a different author.

carljoerger:
How do you make a pull request?  I'd sure like them to be committed, but I hacked up the code pretty well - any feedback is appreciated.

jpfr:
open62541.c diff: http://paste2.org/9HBn82yj
open62541.h diff: http://paste2.org/MOXCtJHy

You changed quite a bit there. Can you summarize the changes?

I see you changed to open62541.cpp. Normally you can just compile with a C-compiler and link against the object file from your C++ project.

I'm unsure what to do with this. Merging is a major effort.
And we support Visual Studio 2015 anyway. (And the VS 2013 C compiler should be supported as well, imho). More arguments towards merging would be helpful. 😄 

Github is very well documented:
https://help.github.com/articles/creating-a-pull-request/

carljoerger:
My main problem was that the Windows Embedded Compact 2013 C compiler in VS2015 didn't seem to support the bool type!  So rather than define bool as char or something like that, I compiled the code using  the C++ compiler.  C++ doesn't support designated initializers, so I had to rearrange some of the initializer lists and remove the designated initializers.  Some of the structures were defined after they were used, so I had to move those definitions.  The C++ compiler complained about assigning void pointers to structure pointers, so I added the appropriate casts.
[...]

PR #890, opened 2016-11-27
infrastructure: simplify appveyor; add visual studio 2008

jpfr:


Pro:
@jpfr FYI I tried to compile open62541 with Visual Studio 2010 and it failed because it does not support C99, and porting open62541 back would be quite an effort.
The same also holds for VS 2008.

jpfr:
The idea of #884 is to compile open62541 as C++.
That works on VS 2008 with some comparably small modifications.
Otherwise, we'd have to change *everything* to be C89 compliant.

Being C99 **and** C++ would not be a bad thing, imho.
But we have to see which resources are available to deal with #884.

PR #893, opened 2016-11-28
Fixed GCC warning suppression

Root-Core:


jpfr:
Thanks for the PR.
Can you add the same condition to the pragma pop 10 lines below?

Root-Core:
Thank you for the hint. ;)

jpfr:
Thank you 😄

PR #900, opened 2016-12-05
Added Nodestore switch

JGrothoff:
Added a new "layer" between the nodestore and the access to it. It switches the nodestore by the namespace index.
So each namespace can have its own nodestore. Multiple namespace can live in the same nodestore. A nodestore might even be an interface to a database or a filesystem etc.
The nodestore can be specified and changed via UA_Server_addNamespace_Full(...) similar to UA_Server_addNamespace(...). So nodestores become plugins. There is a nodestore standard, which is the default nodestore, used by the open 62541 stack.

A namespace concept is also added. This allows to easily handle namespaces and namespace specific stuff like nodestores, custom datatypes and even the numerical index. Moreover it is now possible to change the namespace array during runtime. 

Major changes to the public API are:
- include/ua_nodestore_interface.h:
  - definiton of nodestore interface get public
  - definition of nodestore functions signatures get public
  - functions for linking and unlinking namespace added
  - releaseNode function added
  - added handles to node visitor and iterator
- include/ua_nodes.h:
  - defintion of node struct gets public
  - node handling functions get public (deleteReferences, deleteMembersAnyNodeClass, copyAnyNodeClass)
- include/ua_namespace.h
  - definition of namespace struct
  - UA_NAMESPACE_UNDEFINED inserted
  - functions to handle a namespace (init, new, newFromChar)  
- include/ua_server.h
  - UA_ServerConfig: Namespaces replace namespace strings and custom data types
  - addNamespace_full / deleteNamespace / deleteNamespace_full added
- include/ua_types.h:
  - UA_findDataType gets namespace specific
- /plugins/ua_nodestore_standard.c/h
  - definitions for the standard nodestore as plugin (similar to network layers)
- /plugins/ua_config_standard.c/h
  - Changed server/client config standard to hold namespaces instead of stringarray


jpfr:
Hey @JGrothoff ,

I just took a quick glance at the code.
Good work!

Please don't be discouraged by the comments. As usual, the good parts are not commented, but the needs-work parts are 😄 
If it makes things easier, we can also just schedule a short call for next week.

Best regards,
Julius

jonasgreen88:
Hi,
I would like to implement my own nodestore to the open62541 stack and since the UA_ENABLE_EXTERNAL_NODESTORES feature has been removed recently I guess this is the replacement solution? What is the status of this PR and do you have any schedule for when it will be ready to merge?

Br, Jonas.

jpfr:
@jonasgreen88 

I saw your post on IRC. Can you elaborate what you intend to do?

Note that several namespaces can live in the same "nodestore".
There is `UA_Server_addNamespace` to add new namespaces.
https://github.com/open62541/open62541/blob/0.2/include/ua_server.h#L131

@FlorianPalm 
@JGrothoff 
Do we have a status for the PR?
It has not been syncrhonized with master for some time now.

JGrothoff:
Hey, due to high workload i had no time to syncronize it to the master again and again.
I think the last merge was in end of may, but i didn't push it, because some checks weren't compiling and i didn't get it fixed fast.

In general this doesn't replace the external_namespace feature, both interfaces could exists at the same time. But, if you make an external namespace you have to implement many things, that are already solved in open62541. Therefore, we added the nodestore plugin, which has a much smaller interface.  Moreover the standard nodestore becomes a plugable nodestore. This makes a continues testing of the interface easier. So there is no compiler switch necessary and the code won't get broken without noticing.

jonasgreen88:
@jpfr @JGrothoff 

Thanks for quick response! The use case is that a set of internal data points shall be translated to variable nodes on OPC UA. I don't want to load these parameters and cache its values and metadata (name, data type etc.) in the nodestore of the stack, but use an external interface (i.e. external nodestore or nodestore plugin) to load the data point and generate a variable node in runtime upon a request. The reason for this design is because it is a two processor solution where one processor handles the execution of the network interface with the OPC UA stack etc. while the other handles an application and publish its parameters on an interface.

I know that I could use the external_namespace feature, but as this interface was removed a couple of weeks ago (2d5aedd912ce6bd91c70a7447f4e019b6421e18c) I don't see a future in using it and would like to design my solution with an interface that will last for a while. That's why I'm interested in the nodestore plugin API.

pschoppe:
Coming from [this PR](https://github.com/open62541/open62541/pull/975), I am (still) highly interested in this, too. It would be awesome, if this was merged into the master branch :)

JGrothoff:
So, i didn't have much time, but i merged with the current master and squashed #832 together here.
It compiles, but some unit tests don't work properly. Maybe we can fix the bugs together and then merge this PR into master.

JGrothoff:
Hi, i added a simple reference counting mechanism to the standard nodestore. On my configuration all unit tests work. ( @jpfr I don't know why travis is not starting to check this PR? ). I checked valgrind only on the server example, but i think we can fix last memoryleaks in master.

I am on vacation now for 3 weeks. And i will be very busy afterwards. So should we merge the PR now? (@jpfr @FlorianPalm) Otherwise it will be quite hard to sync it again and again.

JGrothoff:
So, after some bugfixes everything should be fine for merging now, exept the const initalisation of the namespaces array. 

I also updated my server_nodestore example and updatet the first comment above. Nearly all changes to the API are mentioned there.

Maybe some one can help to fix this (@FlorianPalm?), because i am currently on vacation. How do i define a const struct array, so that app veyor in the second configuration will be happy: (https://github.com/open62541/open62541/pull/900/commits/7897a35c4d29ada45c92b65bda06069ab06088b0#diff-1460dbbb030b9af913afd8afa5e4092cR86) . I tried

jpfr:
Cool! Looks great!

In **ua_namespace.h**:
- `UA_NodestoreInterface* nodestore;` does not have to be a pointer. The nodestore carries a handle pointer anyway. So we can have less moving parts there.
- Do I have to set namespaces to a new server *by hand*? Or can I *statically* set them in the server config?
- I like that namespaces can come with their own datatypes. In addition, I sense a more general solution. We can add the ability to add datatypes at runtime separate from the ability to add namespaces. Users who need both can use both. Also, this decouples the namespaces from the encoding layer (which is used for client and server).

Do you think it is possible to split out adding datatypes at runtime? Then we can merge this PR in several smaller chunks.

jpfr:
@JGrothoff 
Are you okay if I take the NodeStore definition you made here and apply it to master?
This also means getting the node_get / node_release logic into master first.

This PR can then concentrate on the NodeStore switch logic only.
I would then mark you as the contributor for the node_get / node_release patch as it is mostly copy-paste from this PR.

JGrothoff:
Yes, i am totally fine with this. Feel free to copy, paste, slice, ... :-)
As i am off to Sweden from tomorrow, nothing will change here, so there will be no interference from my side. I can have a look at it again in three to four weeks from now. Again feel free to use, what you need from this PR.

Your bullet points:
- Yes, the nodestore interface could also live in the namespace struct.
- I think both is possible.
  - The default namespaces ns0 and ns1 are set statically, but can be edited or even extended. This can be seen in the server_nodestore.c example, where the nodestore of ns1 is changed in the config. During UA_Server_new(...) all namespaces in the config will be added in their array order.
  - Namespaces or parts from it (DataTypes, nodestore) can also be changed due to UA_Server_addNamespace_full(...). But we could also define an explicit UA_Server_changeNamespace_full().
  - I even added a compiler switch UA_ENABLE_LOAD_NAMESPACE0 to handle persistent nodestores. But this could also be modeled as a server config flag. (This was usefull for the mysql nodestore, as it already holded the ns0 independent from server start up.)
- I though it would be nice to only add the UA_Server_addNamespace_full to handle everything that is namespace specific and theirfore also the datatypes. But we could even add functions to change the Datatypes alone.

jonasgreen88:
@JGrothoff @jpfr 
Thanks for putting some effort into this PR to get it ready for the master branch. Really appreciate it!

What is the status right now to get it into the master branch? Is there some part I can help you with to get it ready?

JGrothoff:
Hi @jonasgreen88, the actual integration work is now performed by @jpfr. So he knows where help is needed now. You can watch the WIP in #1168.

jpfr:
Merged via #1168

PR #906, opened 2016-12-07
Generate properly versioned shared library links on Linux

MikeLooijmans:
Instruct CMake to create a properly versioned library.
Cross-compile environments like OpenEmbedded/Yocto are picky about adhering to these standards, this fixes QA errors resulting from building open62541 as a shared library for embedded targets.

Signed-off-by: Mike Looijmans <mike.looijmans@topic.nl>


jpfr:
Nice. Thanks for the PR.
I will add some definitions on top, so we don't have to hardcode the library version in multiple places.

PR #922, opened 2016-12-28
python wrapper

oroulet:
This is  python wrapper in a very very early stage. It only exposes a Server object with 2 methods: start and stop. But it is a clean start for someone needing more features. I will probably add a few ones like import_xml

I wrote it for opcua-modeler https://github.com/FreeOpcUa/opcua-modeler , but I am unsure where to put the code. I propose to add this to the open62541 repository in a directory called wrappers for example. Then you can put there  wrappers for other languages. I could also create a new repository under open62541. As you want. Ideally it would be great to integrate it to CMakefile build system, but it looked complicated so I did not try yet.


jpfr:
Hey @oroulet ,

looks good 😄 .
That provides a nice starting point for developing a generally usable python wrapper.
Thanks for the effort!

Having the wrapper in the main repository (in our next release) gives users the impression that the wrapper is finished. We already get loads of issue reports on features that are explicitly tagged as experimental or in-development.

Can we also put this into a separate github project open62541/open62541-python?
So it is endorsed by open62541 but with an independent release cycle (maybe in reference to the library major releases).

oroulet:
no problem. Can you create a repository then I will create a PR to it.

jpfr:
Hey @oroulet,

okay. I added a repository and gave you write access:
https://github.com/open62541/open62541-python

Some minor suggestions to the PR.
- Can we call this open62541 instead of openua?
- The IPython import is probably an artefact

Thanks for the good work!

jpfr:
Closing this PR. All follow up on https://github.com/open62541/open62541-python

PR #932, opened 2017-01-17
Don't copy elements for RegisterServer call

jpfr:
This simplifies `register_server_with_discovery_server` a bit.
Notably, members of the request are not copied but point into the server config.
Otherwise, we have to check the statuscode of every array allocation and every UA_copy.

Also, I'm under the impression that `request.server.discoveryUrlsSize` was always zero in the realloc.

@Pro Can you review the changes?


Pro:
Regarding your comment about discovery URLs always null, this is probably fixed with this commit:
https://github.com/open62541/open62541/commit/83012b5a2ee72a7a21593d0d969d5d6875cca676

PR #937, opened 2017-01-18
improve unit test to detect dangling reference to deleted node

bostjanv:
I modified unit test in a way to reproduce the dangling reference problem I've described on the mailing list.

https://groups.google.com/forum/#!topic/open62541/gwh0fjW0yJw

Bostjan


jpfr:
Thanks for the PR. 😄 
Your commit is added -- together with fix -- to master and 0.2
I didn't want to pull the PR directly without also having the fix.

PR #939, opened 2017-01-19
Added Solution folders and file filters for visual studio

Infinity95:
This should improve work with visual studio a bit due to the files being organized to reflect the underlying file structure.


jpfr:
Thanks a lot. 😄

PR #969, opened 2017-02-20
Modern cmake for a self configurable relocatable install patern

XB32Z:
Hi all, 
I am proposing this pull request to use modern cmake techniques in the open62541 project. This will enable people to use find_package on the it in a very simple way. 
Those changes generate a version and config file that allows to use the library as an IMPORTED target (no more variables :D ) 

hope you like it, 
Best

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,9 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.2)
 project(open62541)
 # set(CMAKE_VERBOSE_MAKEFILE ON)
 
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/tools/cmake")
-find_package(PythonInterp REQUIRED)
+find_package(PythonInterp 2.7 REQUIRED)
 find_package(Git)
 
 set(MDNSD_LOGLEVEL 300 CACHE STRING "Level at which logs shall be reported" FORCE)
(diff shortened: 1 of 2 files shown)


jpfr:
Hey @XB32Z,

I'm no expert in these "modern" cmake features. Can you tell us what the changes mean in practice?

From a quick glance at the PR, I'd request two changes:
- The latest cmake available in debian stable (jessie) is cmake 3.0. Requiring at least 3.2 would exclude potential users (https://packages.debian.org/de/jessie/cmake)
- We should not require Python2.7. When I checked last, we were compatible with both Python2.7 and Python3.

XB32Z:
Hey, 

Yes sure, users of the open62541 library are now doing something such as:
```
find_library (open62541 HINTS ... PATHS ....) 
find_directory (open62541.h ...)
```
Then have to include and link such as
```
include_directories (${open62541_DIRECTORIES})
target_link_libraries (myOPCUAtest ${open62541_LIBRARIES})
```
This works but is not so resistant to changes as the installed package is not relocatable and the users have to play with cmake variables to get this running. 
Whereas modern cmake treats external dependencies as a target. ie
`find_package (open62541 REQUIRED HITNS lib/cmake)`
then 
`target_link_libraries (myOPCUAtest open62541)`
This will handle passing the libraries to link, the directories to include, any defintions to add etc. it's much easier and resitant. 
I am changing now to version 3.0 as modern cmake is 3+ and for the python, it's just that for python 2, it has to be 2.7+ (at least on Windows)

Thx for you time :)

XB32Z:
Hi, 
Are you intrested by this? I just updated and saw that there is also a folder structure now, it's really cool.
Best,

jpfr:
Hey @XB32Z ,

I moved this to a branch that is local to that repository.
May we continue in #977 ?

PR #973, opened 2017-02-21
Add parsing of TCP Error Message 'ERR'

Pro:
The OPC Foundation LDS-ME Server may answer with an 'ERR' TCP message.

The error message is defined in OPC UA Part 6, Section 7.1, Table 34 and 37 

Are there any other places where we should check for that message type?

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -502,6 +502,19 @@ processMSG(UA_Server *server, UA_SecureChannel *channel,
     UA_deleteMembers(response, responseType);
 }
 
+/* ERR -> Error from the remote connection */
+static void processERR(UA_Server *server, UA_Connection *connection, const UA_ByteString *msg, size_t *offset) {
+    UA_TcpErrorMessage errorMessage;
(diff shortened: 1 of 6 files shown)


src/server/ua_server_binary.c
```diff
@@ -512,6 +512,12 @@ UA_Server_processSecureChannelMessage(UA_Server *server, UA_SecureChannel *chann
     UA_assert(channel);
     UA_assert(channel->connection);
     switch(messagetype) {
+    case UA_MESSAGETYPE_ERR: {
+        const UA_TcpErrorMessage *msg = (const UA_TcpErrorMessage *) message;
+        UA_LOG_ERROR_CHANNEL(server->config.logger, channel,
+                             "Server replied with an error message: %s %.*s", UA_StatusCode_name(msg->error), msg->reason.length, msg->reason.data);
```
jpfr: Should that be "client replied..."?

jpfr:
Very nice!
Imho the PR can be merged as-is.
As a next step, we can add the functionality to send ERR messages to the messaging layer.
At some points, we simply close the connection without being explicit about the reason.

src/server/ua_server_binary.c
Pro: Done in new squashed commit

PR #975, opened 2017-02-24
Extend external nodestore API to retrieve references in returnRelevan…

pschoppe:
As described on the group list (https://groups.google.com/forum/#!topic/open62541/1XLhfjERsM8), I think there is an issue in the external namespace API. This pull request is supposed to be an initial fix suggestion and to maybe get more attention than on the list. Additionally, enabling external namespaces on master leads to compilation errors, since there is no userwritemask member in UA_Node. The fix is probably not following the initial API design ideas, but I am not very familiar with the codebase, so this is a quick fix to demonstrate the issue. My feeling is that this should be somewhat integrated into the readNodes callback, but the way it is designed right now does not allow returning arbitrary sized result arrays (in this case an array of all the references associated to an external node).

### DIFF
--- a/include/ua_server_external_ns.h
+++ b/include/ua_server_external_ns.h
@@ -32,8 +32,11 @@ typedef UA_StatusCode (*UA_ExternalNodeStore_addReferences)
  UA_UInt32 *indices,UA_UInt32 indicesSize, UA_StatusCode *addReferencesResults,
  UA_DiagnosticInfo *diagnosticInfos);
  
- typedef UA_StatusCode (*UA_ExternalNodeStore_addOneWayReference)
+typedef UA_StatusCode (*UA_ExternalNodeStore_addOneWayReference)
 (void *ensHandle, const UA_AddReferencesItem *item);
+
+typedef UA_StatusCode (*UA_ExternalNodeStore_getOneWayReferences)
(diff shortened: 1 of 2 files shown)


FlorianPalm:
Hey Philipp,
as you have noticed, the current implementation of the external namespace interface is broken. We have decided to move the interface for external data sources from "service-level" to "node-level". That means, someone would not have to implement all the services by his self, but can rely on the implementation which is provided by the stack. Just the access functions to the nodestore have to be given. Have a look [here](https://github.com/open62541/open62541/pull/900). This will go into master very soon.
Florian

jpfr:
Hey @FlorianPalm,

I'm in favor of pulling in the PR.
1. We should not leave any functionality "broken". Users who download a release assume that implemented functionality works. Otherwise, there is no way to evaluate what our SDK is able to do before doing the actual implementation.
2. Even if the replacement is nearly done, it will not be included in the 0.2 release since it requires a change in the API and the 0.2 API is frozen.

Proposal for next steps:
- We merge this PR and add some checks to ensure the current external_namespaces are ready for the 0.3 release.
- Remove the current external_namespaces from master
- Merge the nodestore_plugin branch.

jpfr:
Merged to 0.2 and master via 3299b5c29a67e6459e31c8106929fa26963b77df.

jpfr:
Thanks @pschoppe for the PR!

PR #977, opened 2017-02-27
Modern cmake for a self configurable relocatable install patern

jpfr:


jpfr:
Hey @XB32Z ,
Hey @open62541/open62541-core ,

for the installation, I propose we always generate open62541.h and copy it to /usr/include.
So packages will only contain one header and the .so or .dll.
That saves us the complexity with /deps.

@XB32Z Is there some overlap between configure_package_config_file and the CPack functionality just below?

XB32Z:
Hey,

it shouldn't add much  complexity as its all hidden behind the package but it's maybe more simple. The CPack creates an installer which function is to copy binaries and file to new locations while the packaging creates cmake scripts to ease the configuration process (after the library as been install, to use it) (so far as I know x) ) 

Best

PR #980, opened 2017-03-06
simplify the connection settings and unify client/server handling of connection settings

jpfr:


Pro:
@jpfr is this still relevant after merging the recent connection config changes?

jpfr:
80% are no longer needed.
What remains interesting is not keeping a local and remote connection config.
This unifies e.g. the remote-maxSendSize and the local-maxReceiveSize to a single value.

jpfr:
Let's look at this again in the 0.4 cycle.
The PR is badly behind master. And its more a beauty-refactor than functionaltiy.

PR #981, opened 2017-03-07
Refactor timer to not depend on server-internals

jpfr:
This prepares the use of repeated jobs in the client.


jpfr:
This should be merged into master, not 0.2

PR #984, opened 2017-03-08
Enable travis OS X build

Pro:
This enables OSX builds on travis.
Currently 0.2 builds just fine there.

One drawback is that it may take up to a few hours until the OSX build starts because the OSX machines on travis are overloaded.


jpfr:
Nice one!

PR #989, opened 2017-03-15
Improve custom types test

bostjanv:
Julius,

if I am not mistaken custom arrays are not decoded properly.

Best regards,
Bostjan


Pro:
@bostjanv any reason you closed this?
IMHO there is a bug related to this, at least the test should also check all the values of the array, not only the first one

bostjanv:
I guess I misunderstood how to access decoded array elements. I will create another PR.

PR #990, opened 2017-03-16
Improve custom array test

bostjanv:


tests/check_types_custom.c
```diff
@@ -131,11 +131,16 @@ START_TEST(parseCustomArray) {
     ck_assert_ptr_eq(var2.type, &UA_TYPES[UA_TYPES_EXTENSIONOBJECT]);
     ck_assert_int_eq(var2.arrayLength, 10);
 
-    UA_ExtensionObject *eo = (UA_ExtensionObject*)var2.data;
-    ck_assert_int_eq(eo->encoding, UA_EXTENSIONOBJECT_DECODED);
-    ck_assert_ptr_eq(eo->content.decoded.type, &PointType);
-    Point *p2 = (Point*)eo->content.decoded.data;
-    ck_assert(p2->x == 0.0);
+    for (size_t i = 0; i < 10; i++) {
+        UA_ExtensionObject *eo = (UA_ExtensionObject*)var2.data + i;
```
bostjanv: Correct me if I am wrong but your suggestion seems incorrect. It could be something like
`UA_ExtensionObject *eo = &((UA_ExtensionObject*)var2.data)[i];`
but I am not sure if this would be easier to understand.

Pro:
Merged with https://github.com/open62541/open62541/commit/9452347a8c0b370b0aff10b09e0cae5fcb12316e

tests/check_types_custom.c
Pro: Yes, you are right, it's better to use `&((UA_ExtensionObject*)var2.data)[i];`

FYI:
https://stackoverflow.com/questions/4622461/difference-between-pointer-index-and-pointer

PR #991, opened 2017-03-16
Refactor periodic server register to use ua_timer.c

Pro:
@jpfr can you review, especially the ua_timer stuff

I added a DelayedJob (single shot) and refactored the periodic server register for better code readability.


jpfr:
Hey @Pro ,

I just pushed a change to master, where the changes to the repeated jobs are always processed in-order.

That might lead to some simplifications in this PR. For example, you would not need to remove jobs from the changes-list when changing an interval.

Further, I added some comments to ua_timer.h on which functions are thread-safe and can be used from a worker-thread outside of the main-loop.

Best, Julius

Pro:
@jpfr I now integrated your changes from master.

One thing which may still be a problem is the first part of the `UA_RepeatedJobsList_updateRepeatedJobInterval` method, where I need to check if the changes list still contains the job. I.e. it is not thread-safe: if updateRepeatedJobInterval gets the repeated job, but then is interrupted by the process method (which may delete the repeated job), then the update interval method accesses invalid memory.

I would like to avoid inserting mutexes here...

jpfr:
Hey @Pro ,

I like how this takes complexity out of the mdns code and provides a generic solution.
Cool!

> where I need to check if the changes list still contains the job. I.e. it is not thread-safe

For adding and removing jobs, we malloc a `UA_RepeatedJob` and enqueue it to the changes-queue. The main process then deques the change and applies. So only one thread can operate on the list of repeated jobs. You can do the same for updating the interval. Just invent a suitable format for "change-entries", e.g.

change->nextTime == UA_INT64_MAX -> remove
change->nextTime == UA_INT64_MAX-1 -> update interval
change->nextTime < UA_INT64_MAX-1 -> add

There's just one possibility not covered. You might return 'UA_StatusCode_Good` for changing the interval. But the job might be removed before the change is applied.
But then again, that's the possibility the current code doesn't cover either.

One last thing: Could you set your IDE to not add a space after if/while, etc.?
It's not that I prefer one or the other style. But we should try to be consisten throughout the code-base.

Pro:
Solved with https://github.com/open62541/open62541/commit/1969bad0666453b485cd2ac169268ac01353e792

PR #992, opened 2017-03-16
Pkg config

glima:


glima:
looking wrong on the base commit, doing it again

PR #993, opened 2017-03-16
Have a .pc.in file to install a pkg-config header

glima:
This way, people linking against installed open62541 on their systems
will be able to query its CFLAGS/LDFLAGS via pkg-config tool.

I also made the amalgamated header be installed only if the build flag
is set. The files that should be installed as headers otherwise are not
known to me. Anyway, all the examples point to the amalgamated header
nowadays.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,6 +199,7 @@ endif()
 
 file(MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/src_generated")
 configure_file("include/ua_config.h.in" "${PROJECT_BINARY_DIR}/src_generated/ua_config.h")
+configure_file("include/open62541.pc.in" "${PROJECT_BINARY_DIR}/src_generated/open62541.pc" @ONLY)
 
 if (UA_ENABLE_DISCOVERY_MULTICAST)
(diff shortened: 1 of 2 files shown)


Pro:
Looks good to me.

@open62541/open62541-core do we also want this in the 0.2 branch?

jpfr:
@glima Looks great!

@open62541/open62541-core 
Now that we have the ability to create installation packages, we should make a quite fundamental decision.
Do we want the `ua_` header files to end up in users `/include` directory or only the amalgamated `open62541.h`?

I prefer distributing the amalgamated header, like I prefer most of the design decisions of sqlite (https://github.com/LuaDist/libsqlite3/blob/master/sqlite3.h).
But I might be biased here.

Pro:
Same here, I would also vote for only distributing the amalgamated `open62541.h`

For advanced users/usage developers still have the possibility to not use amalgamation / the `/include` dir and select the required files on their own.

FlorianPalm:
Yep, I fully agree!

glima:
I like one header better as well, but if feels weird for the user, as it is now, the make install rule doing nothing WRT headers on the target dir...

glima:
ping?

jpfr:
Merged. We can have a separate discussion on distribution in packages for the 0.3 release.

glima:
Thanks a lot.

PR #995, opened 2017-03-17
Configuration for the compile under QNX6.5

MaxKravt:


CMakeLists.txt
```diff
@@ -147,9 +147,9 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
                     -Wno-unused-function -Wno-unused-label -Wpointer-arith -Wreturn-type -Wsign-compare
                     -Wmultichar -Wstrict-overflow -Wcast-qual -Wmissing-prototypes -Wstrict-prototypes
                     -Winit-self -Wuninitialized -Wformat-security -Wformat-nonliteral)
-    if(NOT WIN32 AND NOT CYGWIN)
-        add_definitions(-Wshadow -Wconversion -fvisibility=hidden -fPIC)
-    endif()
+     if(NOT WIN32 AND NOT CYGWIN AND NOT QNXNTO)
+        add_definitions(-Wshadow -Woconversion -fvisibility=hidden -fPIC)
```
Pro: Shouldn't this still be `-Wconversion`?

tools/amalgamate.py
```diff
@@ -61,7 +62,7 @@
 #endif\n''' % (outname.upper() + u"_H_", outname.upper() + u"_H_") )
 
 for fname in args.inputs:
-    with io.open(fname, encoding="utf8") as infile:
+    with io.open(fname, encoding=None) as infile:
```
Pro: Are you sure that the encoding hint in the first line also defines the encoding for the open command?

https://docs.python.org/2/library/io.html says that if encoding is not set, it is platform dependent.

I think on line 29 `file = io.open(args.outfile, 'w')` we should then also indicate the encoding?

jpfr:
Great!
Some of the CMake changes look as if we might pull them into our main build config eventually.
Is there a way to do continuous integration for QNX? Otherwise we might break the build at a later point without meaning to.

PR #997, opened 2017-03-24
Be paranoid and not return garbage on any struct fields

glima:
While debugging an issue I came accross that (does not fix the issue but
might help not to screw on other scenarious).

### DIFF
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -171,7 +171,7 @@ UA_EXPORT extern const UA_String UA_STRING_NULL;
  * of the char-array. */
 static UA_INLINE UA_String
 UA_STRING(char *chars) {
-    UA_String str; str.length = strlen(chars);
+    UA_String str; memset(&str, 0, sizeof(str)); str.length = strlen(chars);
     str.data = (UA_Byte*)chars; return str;
 }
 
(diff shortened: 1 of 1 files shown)


Stasik0:
The builds are failing. You may not use a single {0} if struct starts with a non-scalar.

glima:
That is weird, no warnings here, even with -Wall. OK, doing a memset version so that all compilers are happy.

Stasik0:
Proper initialization with 0 should be quicker ;) No idea why buikds/tests fail though. @jpfr, any idea?

jpfr:
Compound literals are generally not supported in C++.
They are compiler-specific extensions only, which we disable by using quite strict build flags.

I'm also not sure this needs fixing. Accessing uninitialized values are the symptom, not the cause.
Zero-initializing is additional work and our tools, like valgrind and clang-analyzer, can no longer point out the symptoms to us.

glima:
OK, fair. I guess what annoyed me when printing values inside GDB was the output for union fields, but now reading carefully I see we might be safe with what we had before, sorry for the noise. As the for the valgrind argument, I guess it would work just fine with or without the zeroing out initially.

PR #1008, opened 2017-04-08
Fix docker build error - " No CMAKE_CXX_COMPILER could be found."

### LINKED ISSUE #1007: Missing package in (Tiny)DockerFile
Description
===========
Missing package in the DockerFile apk command causes the build process to fail.
The problem files are DockerFile & TinyDockerFile

###### Command: 
```
$ docker build -t open62541.docker .
```
###### Error: 
```
CMake Error at CMakeLists.txt:2 (project):
  No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


See also "/tmp/open62541/build/CMakeFiles/CMakeOutput.log".
See also "/tmp/open62541/build/CMakeFiles/CMakeError.log".
The command '/bin/sh -c cmake -D UA_ENABLE_AMALGAMATION=true /tmp/open62541 && make' returned a non-zero code: 1
```
Fix: Add g++ to the apk command.

*Pull request will be issued shortly.*

AGIsmail:
Fixes #1007

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:3.3
-RUN apk add --no-cache cmake gcc musl-dev python make && rm -rf /var/cache/apk/*
+RUN apk add --no-cache cmake gcc g++ musl-dev python make && rm -rf /var/cache/apk/*
 ADD . /tmp/open62541
 WORKDIR /tmp/open62541/build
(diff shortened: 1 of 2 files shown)


jpfr:
Thanks for the fix!

PR #1014, opened 2017-04-20
Fix redefinition of SLIST_ENTRY on mingw winnt.h

TorbenD:
If the files are amalgameted there is a redefinition of SLIST_ENTRY. This was avoid earlier in the same way, but then moved to ua_util.h and finally deleted.


jpfr:
Thanks. :-)

PR #1019, opened 2017-04-27
Remove -Werror from CMakeLists

btrepp:
This should help with building on newer compilers
https://blog.flameeyes.eu/2009/02/future-proof-your-code-dont-use-werror/

This also fixes #1018 for compilation errors

PR #1020, opened 2017-04-27
Fixes docker builds

btrepp:
These 3 commits
1) Remove Werror which causes errors with FD_SET under MUSL c
2) Adds G++/and removes it in tiny docker file
    Updates the flags to build the shared lib and copy it. (Apparently the static lib isn't build under this instance. So we no longer copy it)
3) Updates alpine linux to the latest 3.5

Both TinyDockerfile and Dockerfile now build with sizes of 24.1MB and 236MB respectively.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,7 +111,7 @@ list(APPEND open62541_LIBRARIES "") # Collect libraries
 
 if(CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID}" STREQUAL "xClang")
     # Compiler
-    add_definitions(-std=c99 -pipe -Wall -Wextra -Werror -Wformat -Wno-unused-parameter
+    add_definitions(-std=c99 -pipe -Wall -Wextra -Wformat -Wno-unused-parameter
                     -Wno-unused-function -Wno-unused-label -Wpointer-arith -Wreturn-type -Wsign-compare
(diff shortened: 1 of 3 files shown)


Pro:
Can you rebase your PR onto 0.2 to include the changes from #1008 and resolve the conflicts.
Since there's already a PR for the "Remove Werror" #1019, please also exclude that commit from this pr. As already mentioned there, I vote against removing Werror.

jpfr:
Pro is right. Warnings should be fixed.

Maybe the problem is that docker / alpine uses the musl libc with different headers.
Can we make a docker build on Travis?

Also, why do we need a C++ compiler on docker?

jpfr:
Testing docker builds on travis...
http://bencane.com/2016/01/11/using-travis-ci-to-test-docker-builds/

Pro:
See https://github.com/open62541/open62541/tree/feature/ci_docker
g++ is required because CMake looks for a C++ compiler (did not yet check why exactly)

Pro:
Some of the changes of this PR are now in 0.2.

See:
https://github.com/open62541/open62541/commit/7b2577323cd22e34de94ca1c922bc9456af800d1
https://github.com/open62541/open62541/commit/643f6666444cd17c5b6d5773664b808d66a31b3b
https://github.com/open62541/open62541/commit/52b1d5485a1270221bbeceaea3450facdb3de810

The warnings in docker are also fixed, without removing the Werror flag

PR #1037, opened 2017-05-15
Try to fix random appveyor build fails

Pro:
Hopefully this fixes random fails on AppVeyor. Let's see what the result of this PR shows...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,7 @@ environment:
         CYG_CACHE: C:\cygwin\var\cache\setup
         CYG_BASH: C:/cygwin/bin/bash
 
+      
     matrix:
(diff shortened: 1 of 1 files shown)


jpfr:
Appveyor failed...
Funny.

Pro:
All green!
Looks like multithreaded compilation causes problems with msbuild. Maybe the project dependencies are not set correctly when cmake is used to create VS projects and thus two simultaneous compilations overwrite the same file.

jpfr:
Nice!
This was a nuissance for a long time.

Stasik0:
btw, could you include creation of a .dll into appveyor builds?

Pro:
@Stasik0 see #1039 for .dll builds on appveyor

PR #1039, opened 2017-05-16
enable shared libs build on travis and appveyor

Pro:
Now AppVeyor creates two artifacts:

- open62541-vs2013-static.zip
- open62541-vs2013-dynamic.zip

In travis an additional build is run with `BUILD_SHARED_LIBS=ON`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -66,10 +66,21 @@ build_script:
   - md build
   - cd build
   - echo "Testing %CC_NAME% with amalgamation"
-  - cmake -DUA_BUILD_EXAMPLES:BOOL=ON -DUA_ENABLE_AMALGAMATION:BOOL=ON -DUA_COMPILE_AS_CXX:BOOL=%FORCE_CXX% -G"%CC_NAME%" ..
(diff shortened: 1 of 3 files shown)


Stasik0:
maybe its worth it to add .a and .dll into travis build s.t. they apper on /releaseas/ page

Pro:
@Stasik0  You mean in this list:
image
And probably you meant .so instead of .a, correct? .a are already there.

Currently this list contains only static linked version.

Do we want (and need) static and dynamic versions for each OS?
Or should I just add it for the most common ones, i.e:

- open62541-linux64-dynamic.tar.gz
- open62541-win64-dynamic.zip

Additionally, on AppVeyor we zip up the complete build folder (including VS Projects and intermediate files). Should we there just pack together the same files as on travis (i.e., open62541 binaries, .c, .h, and two examples for server and client) @jpfr?

Suggestions welcome!

Stasik0:
i am not sure anymore to be honest :)

jpfr:
My modus operandi is to always drop the open62541.c/.h into a project and that's it.
But if people want a .so or .dll, why not?

I would reduce the content of the zips to a minimum.
Just to make people understand how little is needed to get going.
The .dll, the open62541.c/.h and the current documentation PDF.

If you think that entire IDE projects are helpful to get people started, we can discuss that as well.
But I'd rather not put them into official "release" packages.

Pro:
I also agree, that the .zip should be small. I'll change that for appveyor.

For travis, should I add the `dynamic` version for all the OSes or only linux64 and win64?

PR #1045, opened 2017-05-19
Clean up appveyor artifacts and .dll builds

Pro:
The appveyor artifacts now only contain the final output files and not the full build folder.
This also adds latex building on the Appveyor Builds to create the .pdf file


jpfr:
Cool! Merged.

PR #1049, opened 2017-05-23
WIP: remove public ua_job.h; call message processing from the network layer

jpfr:


jpfr:
@Pro 
This PR changes an internal API (jobs are just callbacks).
This required some adjustment of the discovery code.
Can you verify whether all features are still working as intended?
The tests are all green.

PR #1052, opened 2017-05-25
Fix #1051 encoding buffer exchange

### LINKED ISSUE #1051: Binary encoding/chunking bug
Description
===========

Possible bug in binary encoding/chunking code.

We are developing a specific application with soft-realtime constraints. We rely on open62541 stack for communication and data transfer on both server and client side. A couple of weeks ago while performing extensive long-running tests we started to see random and rare timeouts/disconnections. Since the application code is rather complex our initial guess was that we have some timing/synchronization issues in our code. But after some extensive debugging sessions it turned out that we (very likely) triggered a bug in the open62541 implementation of binary encoding/chunking.

After some further investigations I found out that the bug shows itself in the case where there is not enough space in the current chunk to encode the length of the array. The problematic function is Array_encodeBinary, explicitly this block of code:
```
/* Encode the array length */
UA_StatusCode retval = Int32_encodeBinary(&signed_length);
     if(retval != UA_STATUSCODE_GOOD || length == 0)
        return retval;
```

I am not excluding the possibility that a similar problem exists in some other part of the code as well. The bug could be also related to https://github.com/open62541/open62541/issues/1050

I tried to fix the issue by placing another exchangeBuffer call at the strategic place but this resulted in some random decoding errors on the client side. I suppose I am still missing something and my fix was not correct.

I've prepared two unit tests that trigger the bug. See https://github.com/bostjanv/open62541/commit/a7bee9ac884696f66b09f81401f4c0acf325532d

Disclaimer: Please note that since I'm not intimately familiar with the code my reasoning might be wrong.

Checklist
=========

 - [x] open62541 Version (release number or git tag): 0.2, master
 - [ ] Operating system:
 - [x] Critical issue

jpfr:


bostjanv:
Please see my last comment in https://github.com/open62541/open62541/issues/1051

PR #1073, opened 2017-06-09
removed all ifdefs related to UA_ENABLE_EXTERNAL_NAMESPACES

FlorianPalm:
first step to rework the interface to external nodestores


jonasgreen88:
Hi, is there any documentation of how the new interface for external nodestores will look like? When do you plan to have an implementation of this rework?

PR #1079, opened 2017-06-19
VariableTypeNode has no attribute accessLevel

janitza-thbe:
Since somit 645e4ef5 node management services fail to add VariableTypeNodes because it is tried to set _accessLevel_ attribute of _VariableTypeNode_ which has effectively no such attribute and further checks will fail properly with _UA_STATUSCODE_BADNODECLASSINVALID_.


jpfr:
Thanks! Merged.

PR #1081, opened 2017-06-19
Types which extend from Enumeration (ns=0;i=29) allow Int32 values only

janitza-thbe:


jpfr:
Great! So now we typecheck enums for the datatype (nodeid) of the node and the value.
https://github.com/janitza-thbe/open62541/blob/master/src/server/ua_services_attribute.c#L210

PR #1082, opened 2017-06-19
Many ValueRank and ArrayDimension fixes.

janitza-thbe:
See code documentation or commit messages for more information.
- Compatability of each other
- Illegal corrections of value ranks


src/server/ua_services_attribute.c
```diff
@@ -108,15 +108,18 @@ compatibleValueRankArrayDimensions(UA_Int32 valueRank, size_t arrayDimensionsSiz
             return UA_STATUSCODE_BADTYPEMISMATCH;
         break;
     case 0: /* the value is an array with one or more dimensions */
-        if(arrayDimensionsSize < 1)
+        if(arrayDimensionsSize > 3)
```
jpfr: I'm not sure about that change.
For the valueRank 0, scalars are not permitted. This check would allow them.
Also, is there a reason to not allow multi-arrays of more than three dimensions?

janitza-thbe: 1. You're right concerning the upper bound check. I checked and found nowhere a restriction about the number of dimensions. You could simply remove this check.
2. Here again it's problematic like at many other "flat" value checks. For Variables of Objects you're right. A value rank of zero and scalar value will collide. For the other cases Variables of ObjectTypes or VariableTypes a missing value indicates that the instance can choose the value rank by itself. Unfortunately we cannot differ between unset values and explicitly set values to "0" since the attributes mask of node attributes is not used on instantiation which gives information about the real and explicitly set attributes. Further this "flat" check isn't sufficient since the allowed value depends on the hierarchical structure / location of the affected Variable and VariableType (parents). So at last the whole check can and even has to be removed until no further information about the parent structure and type is provided.

src/server/ua_services_nodemanagement.c
```diff
@@ -191,9 +191,9 @@ typeCheckVariableNode(UA_Server *server, UA_Session *session,
     if(retval != UA_STATUSCODE_GOOD)
         return retval;
 
-    /* Fix the variable: Set a sane valueRank if required (the most permissive -2) */
-    if(node->valueRank == 0 &&
-       (!value.hasValue || !value.value.type || UA_Variant_isScalar(&value.value))) {
+    /* Fix the variable: Set a sane valueRank if required (the least permissive -2)
+     * A valueRank of 0 can mean an unset valueRank in case of a type, too! */
+    if(node->valueRank == 0 && !value.value.type && value.hasValue && UA_Variant_isScalar(&value.value)) {
```
Pro: @janitza-thbe is this still relevant with the new default value rank of `-2`?

PR #1084, opened 2017-06-26
Made network_tcp.c and network_udp.c IPv6 compatible. Uses IPv6 structures to initialise sockets. Still provides IPv4 and IPv4-only.

janitza-thbe:


jpfr:
Thanks, merged.

PR #1085, opened 2017-06-26
Added the helper procedure UA_String_equalNative(..) that compares given UA_String directly with native / zero-temrinated C string. This is often required on browsing a node's references and finding nodes in result list by browseName(s). The overhead (...

janitza-thbe:
...memory and CPU) to generate/alloc UA_String for equality comparison only is not acceptable.

### DIFF
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -162,6 +162,7 @@ typedef struct {
 UA_String UA_EXPORT UA_String_fromChars(char const src[]) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
 
 UA_Boolean UA_EXPORT UA_String_equal(const UA_String *s1, const UA_String *s2);
+UA_Boolean UA_EXPORT UA_String_equalNative(const UA_String *s1, const char *s2);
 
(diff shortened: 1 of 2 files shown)


jpfr:
What is the advantage over
```c
UA_String myString = ...;
char *myChars = "aaa";
UA_String_equal(&myString, UA_STRING(myChars));
```

UA_STRING does not use malloc, but runs strlen on the char pointer. The same as the proposed code.

janitza-thbe:
You're right, I didn't see it though it was in direct near within code. ;)
I don't know exactly how the compiler will optimise it. It's already declared as inline. But beyond this there is still an unnecessary variable declaration and assignment on stack while the PR variant uses the values directly which will speed up on slow embedded processors because of less instructions.
As a whole I think its unimportant so its your decision.

jpfr:
A small public API is an advantage.
So as a general rule, we try to have small, composable building blocks.

The difference should be negligible in code-size and speed.
If this is demonstrably not the case, we should provide a dedicated implementation.

Until then, let's keep the API with less complexity.

PR #1086, opened 2017-06-26
Fixed data type exclusion list in generate_types.py by adding two missing data structures that are already hard-coded present by open62541 C code. This collides on generation of full data type list that is defined by Opc.Ua.Types.bsd

janitza-thbe:


jpfr:
Looks good. Can you fix the tabs vs. spaces issue?
Python can be quite strict about this.

jpfr:
Thanks!

PR #1087, opened 2017-06-26
Added list of full data types and its automatic (re-)generation bash script. Searches Opc.Ua.Types.bsd for data types by grep and mixes with existing minimal data type list since not all required basic types are present in Opc.Ua.Types.bsd. Works well ...

janitza-thbe:
...full NodeSet2 to prevent ExtensionObjects and use defined data structures directly for holding (decoded values). Needs to be set as parameter for generate_datatypes.py in CMakeLists.txt


jpfr:
Thanks, merged. 😄

PR #1088, opened 2017-06-26
Very trivial status code description fixes taken from OPC Foundations UANodeSet (3146d80, Errata-1.03.3)

janitza-thbe:


jpfr:
Can the changes be made to https://github.com/open62541/open62541/blob/0.2/include/ua_constants.h#L98 as well?

The macro-definitions are easy to generate from csv with a regex or editor replacements.
Imho this was done manually.
There is also the script that generates descriptions that can be adjusted to generate macro definitions: https://github.com/open62541/open62541/blob/0.2/tools/generate_statuscode_descriptions.py

PR #1098, opened 2017-07-05
Update CMakeLists.txt

XB32Z:
The LIBS variable is not set - hence the example does not link


jpfr:
Great! Thanks for the fix.
Just acknowledge the CLA and we are good to go.

XB32Z:
Done

PR #1120, opened 2017-07-19
Add fuzzing for oss-fuzz

Pro:
https://github.com/google/oss-fuzz will run fuzzing once a day and create bug reports, if any bugs are found.

See also:
https://github.com/google/oss-fuzz/tree/master/projects/open62541

Note:
After merge wait on deleting the branch, until the files on oss-fuzz are updated.


tests/fuzz/fuzz_common.h
```diff
@@ -0,0 +1,62 @@
+//
+// Created by profanter on 18.07.17.
+// Copyright (c) 2017 fortiss GmbH. All rights reserved.
```
jpfr: Is this license header intentional?

Pro: Whoops, nope

tests/fuzz/fuzz_common.h
```diff
@@ -0,0 +1,62 @@
...
+//
+
+#ifndef OPEN62541_FUZZ_COMMON_H_H
```
jpfr: This file does the same as testing_networklayer.c/.h.
Is there a reason to duplicate?

tests/fuzz/fuzz_common.h
```diff
@@ -0,0 +1,62 @@
...
+#define OPEN62541_FUZZ_COMMON_H_H
+
+#include "ua_server.h"
+#include "ua_server_internal.h"
+#include "ua_config_standard.h"
+#include "ua_log_stdout.h"
+#include "ua_plugin_log.h"
+
+static UA_StatusCode
```
jpfr: Did you mean `static UA_INLINE ...`?

tests/fuzz/fuzz_common.h
Pro: I duplicated it, because the testing_networklayer is split into c and h files which makes the build for fuzz more complex.

Can we change the testing_networklayer to only have one header file?

jpfr:
I like this oss-fuzz thing a lot!

What are useful targets for fuzzing?
- Decoding in general (don't need to start a server for that)
- Message Chunking (don't need to start a server either)
- Decoding the payload for specific services (needs the server when the service gets executed)

At a later point, we might add fuzzing of timings.
On master, the tests have full control over a simulated system clock.
But I'm not sure how this can be integrated into the fuzzing-framework.

tests/fuzz/fuzz_common.h
jpfr: Is it not possible to simply replace ua_network_tcp.c with testing_networklayer.c in fuzzing_plugin_sources?

Pro: Yes, within open62541 this wouldn't be a problem, but there is the magic build script here:
https://github.com/google/oss-fuzz/blob/master/projects/open62541/build.sh

So for building the .cc file, we also need to separately build the testing_networklayer.c file. I had this in a test before, but it made the build script more blown up...

jpfr: I see.
Is it possible to build the fuzzing targets also with cmake?
Similar to how the examples are built.
That way, we have more freedom to compile several source files into one fuzzing target.

Pro:
In tests/fuzz/fuzz_common.h:

> @@ -0,0 +1,62 @@
+//
+// Created by profanter on 18.07.17.
+// Copyright (c) 2017 fortiss GmbH. All rights reserved.
+//
+
+#ifndef OPEN62541_FUZZ_COMMON_H_H
I see.
Is it possible to build the fuzzing targets also with cmake?
Similar to how the examples are built.
That way, we have more freedom to compile several source files into one fuzzing target.

-------

No, unfortunately not. OSS-Fuzz sets some flags which you need to use in the build script

Pro:
Yes, exactly there.
The CXXFLAGS should definitely passed to cmake. My concern is that ossfuzz may change them over time and then the script should still work.
Additionally the output files have to be stored in the $OUT directory.

I will have a look at it and see if I find a good solution. Using docker we can simply test everything

Pro:
It looks like I managed to get it working and keep the build.sh script very basic.

After merging this to master I'll create the PR on oss-fuzz so that oss-fuzz builds against master. And we can manage everything in our repo.

The next step is then to add more fuzzing targets

PR #1122, opened 2017-07-20
Default attributes

jpfr:


jpfr:
Once the default node attributes are in place, we can remove some of the 'workarounds' in the nodemanagement services.

PR #1123, opened 2017-07-21
Use proper defines for malloc, free, calloc, realloc

Pro:


Pro:
@jpfr I'm not sure if it is intentional to keep `malloc` instead of `UA_malloc` in the `ua_network_tcp.c`?

This PR replaces all usages of `malloc` to `UA_malloc` (and free, calloc, realloc).

jpfr:
The change is cool.
Some weeks ago, we define UA_malloc internally and the definition was not visible in the global API.
So that's probably why it as not used in the plugins until now.

PR #1127, opened 2017-07-24
Combine c runtime linkage with build type

mauricek:
When compiling as a shared lib, link against the dynamic c-runtime
version.
Additionally, clean up the compile flags to append /MT to /MD and rather
replace it. This fixes warnings from the compiler.


jpfr:
Thanks! Merged.

PR #1129, opened 2017-07-24
WIP: fixed network buffer (server only so far)

jpfr:


Stasik0:
More performance for single-thread?

jpfr:
The second-to-last PR of a feature that was long in the making.

After this, we can interleave the decoding and processing of services.
Example:  In a ReadRequest, decode the first ReadValueId, process it and encode the (partial) response. Then decode the next ReadValueId.
This gives us end-to-end processing of the read-service without a single malloc.
This makes use the fastest server in the west. And it could even simplify some code.

Pro:
@jpfr should this be included in the 0.3 release?

jpfr:
This is currentlly only half-baked.
Support for the client and multithreading is currently lacking.
It would be nice to have. But I cannot justify the complexity for my projects atm.

Pro:
@jpfr please note that I rebased your commit on current master (including force push)

StalderT:
any news about this PR?

Probably, change : 

```
# define FIXED_NETWORK_BUFFER true
UA_ByteString sendBuffer = {0, NULL};
UA_ByteString recvBuffer = {0, NULL};
#else
# define FIXED_NETWORK_BUFFER false
#endif
```

to 

```
# define FIXED_NETWORK_BUFFER
UA_ByteString sendBuffer = {0, NULL};
UA_ByteString recvBuffer = {0, NULL};
#else
# undef FIXED_NETWORK_BUFFER
#endif
```
can resolve the errors

jpfr:
outdated by the upcoming network layer

PR #1133, opened 2017-08-02
Fix fuzzing issue 2690

Pro:
Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2690

@jpfr can you please review.

This should fix at least the first symptoms of the current fuzzing bug.

### DIFF
--- a/src/ua_connection.c
+++ b/src/ua_connection.c
@@ -58,6 +58,12 @@ completeChunksUntil(UA_Connection *connection, UA_ByteString * UA_RESTRICT messa
             break;
         }
 
+        UA_Byte isFinal = message->data[complete_until+3];
+        if (isFinal != 'C' && isFinal != 'F' && isFinal != 'A') {
+            *garbage_end = true; /* the message type is not recognized */
+            break;
+        }
+
         /* Decoding failed or the message size is not allowed. The remaining
          * message is garbage. */
         UA_UInt32 chunk_length = 0;
@@ -125,7 +131,7 @@ UA_StatusCode
 UA_Connection_completeChunks(UA_Connection *connection,
                              UA_ByteString * UA_RESTRICT message,
                              UA_Boolean * UA_RESTRICT realloced) {
-    /* If we have a stored an incomplete chunk, prefix to the received message.
+    /* If we have stored an incomplete chunk, prefix to the received message.
      * After this block, connection->incompleteMessage is always empty. The
      * message and the buffer is released if allocating the memory fails. */
     if(connection->incompleteMessage.length > 0) {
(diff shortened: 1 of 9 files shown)


src/ua_connection.c
```diff
@@ -102,6 +108,8 @@ separateIncompleteChunk(UA_Connection *connection, UA_ByteString * UA_RESTRICT m
                 return retval;
             memcpy(connection->incompleteMessage.data, message->data, message->length);
             connection->releaseRecvBuffer(connection, message);
+            //The complete message is buffered, so no more content to be processed in message
+            message->length = 0;
```
Pro: I'm not sure if this is correct... If the message is copied into `connection->incompleteMessage` then `message` should be set to empty, because there is no more content to be processed by the following methods, especially because the content of `message` is not complete

src/ua_connection.c
```diff
@@ -58,6 +58,12 @@ completeChunksUntil(UA_Connection *connection, UA_ByteString * UA_RESTRICT messa
             break;
         }
 
+        UA_Byte isFinal = message->data[complete_until+3];
+        if (isFinal != 'C' && isFinal != 'F' && isFinal != 'A') {
+            *garbage_end = true; /* the message type is not recognized */
+            break;
+        }
```
Pro: We should ignore packages which do not have the corresponding `isFinal` Flag (see Spec Part 6, Table 26)

src/ua_connection.c
Infinity95: Isn't this redundant, since releaseRecvBuffer is called on the message? Internally this should free the message and set the length to 0, right? (ua_network_tcp just calls the deleteMembers function of UA_ByteString in its implementation)

Pro: Yes, you are right, it was just not implemented for the testing network layer `releaseRecvBuffer`. I'll remove this change

Pro:
My idea behind adding a build switch for the hex dump was to also use it not only for fuzz debugging, but also for other debugging cases where you need to see what data is received by the server.

I'm not sure if it is even possible to pass command line args to the fuzz binary, since there is no normal main method. I have to check that...

jpfr:
Yes, having this option for debugging is a good idea.
If we need the build flag, can we call it `UA_DEBUG_DUMP_PKGS` or something similar?
Just to emphasize that this is for debugging only and normal users needn't bother.

If the dumping is for general debugging, it could live in `/plugins`.
So the main library does not pull in dependencies outside of the C99 standard.
We go out of our way to not pull in `stdio.h` in the main lib. For embedded targets and such...

See also `plugins/ua_clock.c`.
It implements a method defined in the main lib with an implementation that is is OS-specific plugin for static linking.

Pro:
@jpf changed as requested. Please review, and if CI runs through, please merge.

jpfr:
👍

PR #1140, opened 2017-08-07
Fix memory leak on last array element

Pro:
Reference: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2883
Credit to OSS-Fuzz


jpfr:
👍

PR #1141, opened 2017-08-07
Add VxWorks Support

cabralfortiss:
Add vxworks option to CMakelists.txt, include the right files for "select" in networking, and solve uncompatible code with vxworks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,8 @@ option(UA_ENABLE_AMALGAMATION "Concatenate the library to a single file open6254
 option(UA_ENABLE_COVERAGE "Enable gcov coverage" OFF)
 option(BUILD_SHARED_LIBS "Enable building of shared libraries (dll/so)" OFF)
 
+set(UA_VXWORKS_WRS_KERNEL OFF CACHE BOOL "Enable if you want to compile for VxWorks as kernel Module")
+
 if(UA_ENABLE_COVERAGE)
   set(CMAKE_BUILD_TYPE DEBUG)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fprofile-arcs -ftest-coverage")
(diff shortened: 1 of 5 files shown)


Pro:
@open62541/open62541-core we should think about a general approach how to support such specific targets.

Here at fortiss we are currently also trying to compile open62541 on FreeRTOS which requires significant changes in the `ua_network_tcp` to support lwip. Of course we will make these changes available to the community when it is ready.

Can you think of a general way on how we can handle PRs with such platform specific changes?

One example is the approach which is used in 4diac FORTE:
https://git.eclipse.org/c/4diac/org.eclipse.4diac.forte.git/tree/

1. In the src/arch folder there is the source code for the specific architectures:
https://git.eclipse.org/c/4diac/org.eclipse.4diac.forte.git/tree/src/arch?h=develop
2. Each CMakeLists.txt file can add its own architecture by calling `forte_add_architecture` (https://git.eclipse.org/c/4diac/org.eclipse.4diac.forte.git/tree/src/arch/macos/CMakeLists.txt?h=develop#n14)
3. In CMake you can select the architecture you would like to use

Using this method, every arch/CMakeLists.txt can set its own CMake settings and custom plugin sources.

Could a similar approach also work for open62541?

cabralfortiss:
Travis checker gives the following error:

`/home/travis/build/open62541/open62541/build/src_generated/ua_namespaceinit_generated.c:51536:70: error: 
      string literal of length 254996 exceeds maximum length 4095 that ISO C99
      compilers are required to support [-Werror,-Woverlength-strings]
/home/travis/build/open62541/open62541/include/ua_types.h:164:52: note: 
      expanded from macro 'UA_STRING_ALLOC'
#define UA_STRING_ALLOC(CHARS) UA_String_fromChars(CHARS)`

What does it mean and how can I fix this from my introduced changes?

cabralfortiss:
I think creating an architecture folder will be the best for portability. The first step would be to define the interface from the core to the platform dependant code as a set of functions declarations in a header file, and then each platform defines these functions in their own c files. 

This is an interface to the newtorking features:
https://git.eclipse.org/c/4diac/org.eclipse.4diac.forte.git/tree/src/arch/bsdsocketinterf.h?h=develop

jpfr:
A folder structure
```
/plugins/freertos
/plugins/vxworks
```
is a good idea. However, we should be cautious about the complexity of the patches we allow.
In the past (2014is) we had a folder with architecture-patches. But it quickly went out of hand. Because some of the devs didn't have access to the build tools for all architectures. And the patches broke a lot during normal development.

Is it okay if we should put only
- plugin implementations
- CMakeLists defintitions

in the plugin folders?

With regards to this patch. The floating point code is used only for verrrry strange architectures that don't use IEE754 floats. Normally, you just have to set `UA_BINARY_OVERLAYABLE_FLOAT` to true. Then, the float is copied over from RAM according to the set endianness. That should be fine for 99% of cases.

cabralfortiss:
I don't much about the architecture of open62541. It looks like all the platform dependant code is in the plugin directory. The idea is that the code inside the plugins/vxworks won't even be compiled if it's being compiled for another platform, so it won't crash. 

Regarding the floating point part of the patch, I didn't quite understand your recommendation. Are you referring to this specific part:

https://github.com/open62541/open62541/pull/1141/commits/d5f17e702b9a0ff658b21caddff519dce481186a#diff-c3b90ce2fd11b5aac86ded3a6d7ac74cR370

If yes, then I'm not sure how `UA_BINARY_OVERLAYABLE_FLOAT` would help, since the problem is that I don't have the signbit function in my compiler and I didn't want to dig in too much since it looked like you said a verrrry strange case

jpfr:
With `UA_BINARY_OVERLAYABLE_FLOAT` enabled, you don't even need the sign bit.
The code where you make changes is not used in that case.

https://github.com/open62541/open62541/commit/d5f17e702b9a0ff658b21caddff519dce481186a#diff-c3b90ce2fd11b5aac86ded3a6d7ac74cR309

But I cannot know whether it works in vxWorks to just copy the 4 Byte (or 8 Byte for double) from RAM onto the network buffer.

cabralfortiss:
Ohhh, I didn't see that it was behind a big def. I'll check how the float is managed in vxworks and push again

Pro:
Closing in favour of #1260 (due to some travis issues)

PR #1143, opened 2017-08-08
Fuzz UA_binaryDecode

Pro:


jpfr:
So oss-fuzz checked that we never write into the buffer during decoding?
Nice!

Pro:
Yes, it complaints if you change the data which is fed into the fuzzing target. Really nice stuff!

PR #1144, opened 2017-08-08
Fix memory leak

Pro:
Credit to OSS-Fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2888

### DIFF
--- a/src/ua_types_encoding_binary.c
+++ b/src/ua_types_encoding_binary.c
@@ -1487,6 +1487,9 @@ UA_decodeBinaryInternal(void *dst, const UA_DataType *type) {
             ptr += sizeof(void*);
         }
     }
+    if (retval != UA_STATUSCODE_GOOD) {
+        UA_deleteMembers(dst, type);
+    }
     return retval;
 }
 


jpfr:
Hmm. A decoding method that finds an error should clean up after itself..
Can we fix the builtin-type that is responsible here?

Pro:
The problem in this case is the following:

1. `UA_decodeBinaryInternal` is called with type `Argument`
2. The loop iterates over the members of Argument (https://github.com/open62541/open62541/blob/master/src/ua_types_encoding_binary.c#L1473)
3. The first member is `name` of type `String`. This decoding is successful, and retval is still Good within the loop.
4. The second member is `dataType` where the decoding fails. This sets retval to some error and causes the loop to exit, which simply returns the error code from UA_decodeBinaryInternal
5. So the call from `decodeBinaryJumpTable` returns: https://github.com/open62541/open62541/blob/master/src/ua_types_encoding_binary.c#L1120
6. But then only UA_free is called on the `dst` object which does not free the memory of `Argument.name`

Pro:
Another way to fix this would be to add the following outside the for loop in `UA_decodeBinaryInternal`:

```
if (retval != UA_STATUSCODE_GOOD) {
    UA_deleteMembers(dst, type);
}
```

But IMHO that's the same?

jpfr:
I prefer the second way. It allows a nice conceptual abstraction:
We can add a comment that `decodeBinaryJumpTable` either succeeds or cleans up the data type that was passed to it.

PR #1147, opened 2017-08-09
Support fuzzing of concatenated packages

Pro:
processBinaryMessage is only able to handle `HEL` or `OPN` messages separately of `MSG`s. After a `HEL` it needs another call to processBinaryMessage to process `MSG` packages.

To know how many bytes are processed by processBinaryMessage, we need to get the number of processed bytes so we can remove them from the message and feed the remaining bytes into processing.

On the way to implement this, a few memory bugs and two critical DoS bugs (which cause the server to run in an endless loop) have been fixed.


jpfr:
Hey @Pro ,

good idea.

I will cherry-pick the following commits to a separate branch.
They can be committed before the main feature of this PR.
-  Align format with `hexdump -C`
- Avoid accessing address 0x01
- Fix processBinaryMessage freeing memory which it is not allowed 
- Properly initialize and delete send buffer 
- Fix DoS bug which leads to endless loop 
- Fix memory leak

With regards to the processing of several messages via fuzzing, I propose a refactor to go in before this PR: Call a callback on every complete message from within `UA_Connection_completeChunks`.
That means we don't call `UA_SecureChannel_processChunks` on *all* received chunks. Which is the problem this PR tries to solve.

Pro:
@jpfr did you already have time to work on your proposed callback?

Otherwise I may have time in the coming days. So if you could explain your callback approach a bit more in detail I can try to implement it.

jpfr:
Chunk handling was refactored as part of #1187 .
I can see whether that part can be merged to master ahead of the rest.

PR #1003, opened 2017-03-31
Feature: Security Policies as Plugins

Infinity95:
This PR implements a plugin architecture to support encryption with different security policies as plugins.
The architecture is mainly finished. Mainly error handling is still incomplete and im working on that.
Some encryption stuff still needs some work since im not quite sure about what the standard specifies.

I'm not quite done working on this, but thought it might be useful to already create a PR so others can see the current status. Also errors might get noticed earlier this way.

PR #1151, opened 2017-08-10
Endpoints in the configuration as a precursor to the encryption plugin

jpfr:
This PR collects features from PR #1003 that can be merged to master independently.
So #1003 becomes a bit more manageable in size.


examples/CMakeLists.txt
```diff
@@ -12,8 +12,9 @@ macro(add_example EXAMPLE_NAME EXAMPLE_SOURCE)
   add_executable(${EXAMPLE_NAME} ${STATIC_OBJECTS} ${EXAMPLE_SOURCE} ${ARGN})
   target_link_libraries(${EXAMPLE_NAME} open62541 ${open62541_LIBRARIES})
   assign_source_group(${EXAMPLE_SOURCE})
-  add_dependencies(${EXAMPLE_NAME} open62541_amalgamation_header open62541_amalgamation_source)
+  add_dependencies(${EXAMPLE_NAME} open62541_amalgamation)
```
Pro: This should stay `add_dependencies(${EXAMPLE_NAME} open62541_amalgamation_header open62541_amalgamation_source)`
I had to separate the target to avoid that `open62541.c` is created on every build even if I disable amalgamation. Additionally it solves some dependency issues for CI
Otherwise we can also add an intermediate target `open62541_amalgamation` which includes the `_header` and `_source`

Infinity95: I think this might have happened while i was merging something.

jpfr: Ok. Thanks.
Now I'm waiting for @Infinity95 to make a final merge of master into #1003 .
Then, the merge will happen in 2 or 3 waves.

src/server/ua_services_discovery.c
```diff
@@ -429,7 +429,7 @@ process_RegisterServer(UA_Server *server, UA_Session *session,
             UA_LOG_WARNING_SESSION(server->config.logger, session,
                                    "Could not unregister server %.*s. Not registered.",
                                    (int)requestServer->serverUri.length, requestServer->serverUri.data);
-            responseHeader->serviceResult = UA_STATUSCODE_BADNOTHINGTODO;
+            responseHeader->serviceResult = UA_STATUSCODE_BADNOTFOUND;
```
Infinity95: https://github.com/open62541/open62541/issues/1152#event-1201107739

jpfr:
This PR introduces **breaking changes to all server implementations!**
Can the @open62541/open62541-core team please sign it off?

The server configuration is now returned from a new/delete method pair.

The PR was cut out from the upcoming encryption implementation.
There, we define the server endpoints in the configuration.
So the configuration contains malloced members,

Here's a mock-up of the new handling of the server config.

```c
...

int main(int argc, char** argv)
{
    signal(SIGINT, signalHandler); /* catch ctrl-c */

    /* Create a server listening on port 4840 */
    UA_ServerConfig *config = UA_ServerConfig_new_default();

    /* Run the server loop */
    UA_StatusCode status = UA_Server_run(server, &running);
    UA_Server_delete(server);
    UA_ServerConfig_delete(config);
    return status;
}
```

I will send a short "breaking changes" mail to the mailing list when this is merged.

Issue #1132, opened 2017-08-01
Client connect timeout


Pro:
Currently we do not define any timeout value for when a client connects to a server.
I.e. for the connect call here:
https://github.com/open62541/open62541/blob/master/plugins/ua_network_tcp.c#L657

This means that the connect to an invalid IP will time out after the default OS setting which is on linux around 130 seconds.

Shouldn't we use the timeout from `UA_ConnectionConfig` instead?

This could be implemented using a combination of select and non-blocking socket:
https://stackoverflow.com/a/2597774/869402

StalderT:
+1

jpfr:
👍

PR #1154, opened 2017-08-11
use timeout for client connect

Pro:
Currently a connect to an invalid IP or an IP where no server is running has a timeout of around 120 seconds.

This PR implements the timeout also for this case. See #1132 

If the client is connecting to localhost on linux,
the timeout is not recognized and ECONNREFUSED wil be returned
if there's no server running. Therefore we need a loop.


Pro:
The main reason why I used the loop is to handle the linux specific case where you try to connect to a server on a local IP. There linux immediately returns with ECONNREFUSED. On Windows connect waits until the timeout.
To have a consistent behaviour I chose this approach. It also allows you to start the server within the timeout and the client is still able to connect.

I'll do the cosmetic cleanup 👍

PR #1158, opened 2017-08-12
Refactor duplicate code into a named function.

fmeerkoetter:
One of the comments even spelled out how this block of code should be named -- findSubscription.

### DIFF
--- a/src/client/ua_client_highlevel_subscriptions.c
+++ b/src/client/ua_client_highlevel_subscriptions.c
@@ -49,7 +49,7 @@ UA_Client_Subscriptions_new(UA_Client *client, UA_SubscriptionSettings settings,
     return UA_STATUSCODE_GOOD;
 }
 
-static UA_Client_Subscription *findSubscription(UA_Client *client, UA_UInt32 subscriptionId)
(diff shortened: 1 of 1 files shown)


src/client/ua_client_highlevel_subscriptions.c
```diff
@@ -49,14 +49,20 @@ UA_Client_Subscriptions_new(UA_Client *client, UA_SubscriptionSettings settings,
     return UA_STATUSCODE_GOOD;
 }
 
-/* remove the subscription remotely */
-UA_StatusCode
-UA_Client_Subscriptions_remove(UA_Client *client, UA_UInt32 subscriptionId) {
-    UA_Client_Subscription *sub;
+static UA_Client_Subscription *findSubscription(UA_Client *client, UA_UInt32 subscriptionId)
```
Pro: Nice one! One small change request:
Can you use const wherever possible, i.e., change this (if the compiler agrees) to 
`static UA_Client_Subscription *findSubscription(const UA_Client *client, const UA_UInt32 subscriptionId)`

Thanks!

PR #1162, opened 2017-08-16
Unbreak the build

fmeerkoetter:
I suspect that this is a fallout from
commit a8012d3040f903ffd127d06c70c73cbc87379602
Author: Infinity95 <mark.giraud@student.kit.edu>
Date:   Thu Aug 10 19:00:32 2017 +0200

    endpoints in the server config; new/delete functions for the server config


jpfr:
Yes. Good find.
By the way, the networklayer is already in the default config.
So you don't have to define one extra.

PR #1163, opened 2017-08-17
Generate pkgconfig files

mauricek:
pkgconfig is used to simplify finding and to configure usage of open62541
inside of external projects.


jpfr:
Have you seen this on the master branch?
https://github.com/open62541/open62541/blob/master/include/open62541.pc.in

Should we move this into the root directory?

mauricek:
Actually I was on the 0.2 branch and did not check that master might have a solution already.

Is there an ETA when we can expect master to become a release?

PR #1164, opened 2017-08-17
Free allocated client->reply always (on async_client_loop)

ccvca:
Hi,
the allocated memory from https://github.com/open62541/open62541/blob/async_client_loop/plugins/ua_network_tcp.c#L150 was not always freed. 

ccvca

### DIFF
--- a/src/ua_connection.c
+++ b/src/ua_connection.c
@@ -160,15 +160,25 @@ UA_Connection_receiveChunksNonBlocking(UA_Connection *connection, UA_ByteString 
 
     UA_StatusCode retval = UA_STATUSCODE_GOOD;
 
-        /* Listen for messages to arrive */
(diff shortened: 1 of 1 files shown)


ccvca:
I force-pushed my first commit, (I didn't get the "real" error)
The problem is caused by this chain:
- ua_network_tcp.c:149  malloc
- ua_network_tcp.c:167  if(resultsize == 0) return UA_STATUSCODE_GOOD; /* No result */
- Message with length 0
- ua_connection.c:168   retval |= UA_Connection_completeMessages(connection, chunks, realloced); //Nothong will be done here, because message->length == 0
- ua_client.c:851 if(retval != UA_STATUSCODE_GOOD || client->reply.length > 0) // does NOT match
- // memory is still allocated

ccvca

PR #1167, opened 2017-08-18
Fix typeCheckValue for Subtypes of BuiltIn types

janitza-thbe:
type check value fix: if target data type is a sub type of a built-in type then the value may have the type of the "parent" built-in type, too (e.g. UtcTime is sub-type of DateTime and has a DateTime value)

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -249,9 +249,17 @@ typeCheckValue(UA_Server *server, const UA_NodeId *targetDataTypeId,
         return UA_STATUSCODE_BADTYPEMISMATCH;
     }
 
-    /* Has the value a subtype of the required type? BaseDataType (Variant) can
-     * be anything... */
-    if(!compatibleDataType(server, &value->type->typeId, targetDataTypeId)) {
+    /* Has the value a subtype of the required type? BaseDataType (Variant)
+     * can be anything...
+     * If value is a built-in type:
+     * The target data type may be a sub type of the built-in type.
+     * (e.g. UtcTime is sub-type of DateTime and has a DateTime value) */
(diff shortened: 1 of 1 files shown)


Pro:
@janitza-thbe can you think of an easy unit test which covers this case?

Maybe you can add a test case somewhere here:
https://github.com/open62541/open62541/blob/master/tests/check_services_attributes.c#L555

jpfr:
Good one!

Can we move this into compatibleDataType? So all the tests are in one place.
Also (optionally for this PR) it would be great if we could cite the standard in a comment for this.

janitza-thbe:
@Pro Momentarily I've not very much time for setting up further tests, sorry for that. It's only a hint or suggestion that I backported from our private fork. You can add the tests to the pull request if you want.

@jpfr I tried to call compatibleDataType with switched types first but this leads to much more errors with the enumerations exception that is already implemented. Integrating this special check is not easily possible without changing the complete method signature of compatibleDataType. For my generated model based on FullNodeSet2 and another complex user space model in NS1 this works very well so far and the stack is running fine in multi-threading mode.

jpfr:
> I tried to call compatibleDataType with switched types first but this leads to much more errors with the enumerations exception that is already implemented. Integrating this special check is not easily possible without changing the complete method signature of compatibleDataType. 

We call compatibleDataType also for method calls to verify the arguments. It would be nice to have this fix applied everywhere we do a type-check.

You can just plug the `value->type` into compatibleDataType and keep the code logic exactly as-is.
The only information currently missing is `value->type->builtinType`.

jpfr:
@janitza-thbe 
Can you please sign the Contributor License Agreement as per the CI check marked pending?
I will then merge your changes via #1184.

Thanks for sending the fix!

janitza-thbe:
@jpfr 
Done, sorry, didn't realize that my CLA agreement was revoked, too.

PR #1168, opened 2017-08-18
WIP: Nodestore plugin interface

jpfr:
This is the first part of of #832 with only the plugin interface.
Adding the nodestore switch comes in a separate PR.
Authorship will be attributed to @JGrothoff for coming up with the bulk of the changes and the plugin API.

PR #1179, opened 2017-08-24
WIP: Nodeset compiler for v0.3 - Server bootstrapping

Pro:
Second part of #739.

Mainly contains the bootstrapping part on the server side.


jpfr:
This refactor of the namespace compiler splits out possible backends for generating code.

Some if the server internals, such as the nodestore, will get exposed via PR #1168.
However, since I regularly updated the ns0 generation in the main server, I know how much pain the initial bootstrap is. All kinds of circular dependencies that must not happen once some basic definitions are in place.

Since we now have "exchangeable" backends for the namespace compiler, can we make an open62541 backend that only uses the public API and assumes some basic nodes as pre-existent? By not relying on internal API, we can finally have a nodestore compiler that is future-proof and will not break with every change on master.

By the way, did you notice how the refactor has 4k lines less of code and compiles the full ns0 in 2 seconds? (Not bragging at all... 😏 )

Pro:
@jpfr, @ichrispa can you please review.

Finally this PR is at a stage where we can again create the complete NodeSet2 out of the .xml file.
Using a very basic pre-initialization of the server and the full NS0 XML file, there are no circular dependencies.

Since the pre-initialization is very basic, we rely on a minimal XML file which creates all the minimal required nodes. This also means that the nodeset compiler is now always used, not only if enabled explicitly. That should also lead to less code breaks of the NS compiler when changing the API.

It is not using the `_begin` and `_finish` methods, but since `_finish` should not be executed during bootstrapping, it is disabled when `bootstrapNS0` is set to true.

Still open TODOs:

- [ ] Fix CI
- [ ] The generated code uses `UA_encodeBinary` which is an internal function. Thus if amalgamation is enabled, that's not available. How do we handle this? Make `UA_encodeBinary` public?
- [ ] Modify tutorials to reflect the new nodeset compiler
- [ ] Test combination of multiple .xml files, e.g., combine `Opc.Ua.NodeSet2.xml` and `Opc.Ua.Di.NodeSet2.xml` (Devices Information Model)

ichrispa:
OK, so the code was changed quite significantly.

The compiler in its current state cannot handle more than one XML input file due to the argument parser.

```
/usr/bin/python ${PROJECT_DIR}/tools/nodeset_compiler/nodeset_compiler.py 
--generate-ns0 --ignore ${PROJECT_DIR}/tools/nodeset_compiler/NodeID_NS0_Base.txt 
${PROJECT_DIR}/tools/schema/namespace0/Opc.Ua.NodeSet2.xml 
${PROJECT_DIR}/tools/schema/namespace0/Opc.Ua.Di.NodeSet2.xml 
${PROJECT_DIR}/build/src_generated/ua_namespace0

nodeset_compiler.py: error: unrecognized arguments: /home/ichrispa/work/svn/working_copies/open62541/build/src_generated/ua_namespace0
```

In essence this means we cannot have users compile their own models. I have not precisely looked into the changes, but if there is anything of my original code left, the NS0 XML *must* be passed to the compiler in order for it to be able to link references to NS0 types... this appears to still be the case.

The old code for parsing more than one XML file appears to be still in place. This is an argument parsing issue because of the argsparser backend in python.


What's the concept here? We no longer compile userspace models per se, or is this just a side effect? In the later case, I would suggest we introduce --output or -o to pass the generation target files, then fumble around with the parser so it considers everything not parsed to be an input XML.

Pro:
Code wise there is more or less everything already prepared to handle multiple XML input files.
It is also still the case that it needs the NS0 XML as a basis to solve the references.
So this feature will definitely be supported. I will test this using the Devices Information Model (see my last TODO point in my previous comment)

I would also vote for an additional `-o` parameter.

ichrispa:
Hmmm... Kahn seems interesting, but it will run into similar problems as Djikstra with cyclical dependencies. 

As a more general note: I cannot compile the PR. Compiling Namespace 0 hits a good dozen of these errors:

```
/home/ichrispa/work/svn/working_copies/open62541/build/src_generated/ua_namespace0.c:17868:74: error: ‘UA_TYPES_RANGE’ undeclared (first use in this function)
 void *variablenode_ns_0_i_2370_variant_DataContents = UA_alloca(UA_TYPES[UA_TYPES_RANGE].memSize);
```

The compiler is right - these defines do indeed not exist.

I will continue reviewing the PR tomorrow to get behind the issue. If theres any time left over after my TODO list is chewed down, I'll try to contribute the -o option and multiple file compilation.

Pro:
Compiling should work, at least the Non-Multithreaded version. You probably need to run `make clean` to force a regeneration of the datatypes.
If you are trying to compile the full NS0, you also need to use the correct type definition file, i.e.:

`cmake -DUA_BUILD_EXAMPLES=ON -DUA_ENABLE_AMALGAMATION=OFF -DUA_NAMESPACE0_XML=../tools/schema/namespace0/Opc.Ua.NodeSet2.xml -DUA_DATATYPES_FILE=../tools/schema/datatypes_full.txt`

Unfortunately I had to add this additional compile switch, otherwise there are not all datatypes available.
Maybe there's a better way to handle this?

Pro:
@ichrispa note that I just pushed another change which now supports multiple XML input files. You can just repeat the `--xml` option.

I also had to change the way the `generate_datatypes.py` is parameterized:
If you want to create a specific nodeset you also need to have the corresponding types of that nodeset. This is due to the fact that for value initialization we need the UA_TYPES array to set a value so that the server is later on able to create the variable node. This is a major hurdle and we need to change that. Like this nobody will be able to generate a nodeset. Maybe we can avoid creating those dummy values (and thus relying on the UA_TYPES array) at all?

You can see a quick and dirty test with the DI here:
https://github.com/Pro/open62541/commit/232ed5c6008624c5722094321957a5855a1dc6b4
Of course this commit is not the preferred way to use additional compiled namespaces since it's fiddling around with the main CMakeLists. The current problem is that if you do not use the `--generate-ns0` switch (which delays the parent node references) adding nodes fails because the parent node is not yet there. This means that there is still a problem with the dependencies resolution.
See here:
https://github.com/Pro/open62541/commit/7273fe0b62ea8aaa6f0037173d459d0b77dc8e3a

So there's still work to do.

jpfr:
There is definitely support for multiple nodeset files in #739.
There is even automatic allignment of namespace indices if the same namespace (URI) appears twice in different files with different indices.

There are two types of nodeset files that are loaded via the command line arguments.
"Existing" files where the nodes are not generated and only used for verification.
"New" files where the nodes are generated in the backend. References to "existing" nodes are generated as well.

https://github.com/open62541/open62541/pull/739/files#diff-cadfbbc1c694b7d4daf4ae08e8323916R87

Pro:
Yayyy :boom: :raised_hands: :tada:

Now the nodeset compiler is able to fully create NS0 as library internal initialization.

And then the developer can compile his own nodeset file independent of the library and initialize it in his own code.
A full example on how this can be done is shown here https://github.com/open62541/open62541/blob/7273fe0b62ea8aaa6f0037173d459d0b77dc8e3a/examples/CMakeLists.txt#L73 (branch `testing/di_nodeset`) and as a full example here:
https://github.com/open62541/open62541/compare/nodeset_compiler_02_server...testing/di_nodeset?expand=1

Please try it out and also try to use your own .xml files and let me know if there are any issues.

@jpfr and @ichrispa let me know what you think about the general implementation and how that is handled. If the general setup is ok, we can go into a more detailed review.

jpfr:
Cool!!!
The changes I saw look good.
I am very much in favor of this.

This PR wants to go into #739. We can do that now and continue working there.
But I don't know what this means for squashing/rebasing on master, etc.

Feel free to propose a way forward.

Pro:
The easiest way to get this in a clean way into #739 is to create a diff patch and simply apply this patch as a new commit. I'd go with this. We can also split the diff into two commits. The first part under @jpfr name and the rest mine.

Otherwise if we should/want to keep the history here, we can cherry pick the commits onto #739.
But IMHO the history here is not so fancy to be worth keeping.

@jpfr Any preference?

jpfr:
👍 Sounds good.
Also +1 for a two-commit history.
One from me (which will not be CI-clean) and the latest fixes from you @Pro .

The changes to the datatype description generation are kinda independent and can be be merged to master ahead of time.

jpfr:
What about this "namespace interest group" telko?
Is there a date yet?

Pro:
No there's no date yet. I will first clean up the two PRs, add a unit test for the NC and make sure CI is satisfied. Then we have a good basis for our discussion.

Pro:
Merged into nodeset_compiler_02

PR #1187, opened 2017-08-31
Feature/encryption

jpfr:


jpfr:
@Infinity95 This would be the final structure for merging encryption.
I made some last changes to the securitypolicy API and how "temporary channels" are handled.
Can you verify?

@open62541/open62541-core Can the core developers have a look at this?

include/ua_plugin_securitypolicy.h
```diff
@@ -1,38 +1,286 @@
...
+struct UA_SecurityPolicy {
+    /* Additional data */
+    void *policyContext;
+
+    /* The policy uri that identifies the implemented algorithms */
+    UA_ByteString policyUri;
+
+    /* The local certificate is specific for each SecurityPolicy since it
+     * depends on the used key length. */
+    UA_ByteString localCertificate;
```
Infinity95: Shouldn't this be contained in the policyContext?

plugins/ua_securitypolicy_none.c
```diff
@@ -0,0 +1,304 @@
...
+/* Security Policy functions */
+/*****************************/
+
+static void
+policy_deletemembers_sp_none(UA_SecurityPolicy *policy) {
+    UA_ByteString_deleteMembers(&policy->localCertificate);
+}
+
+UA_StatusCode
+UA_SecurityPolicy_None(UA_SecurityPolicy *policy,
```
Infinity95: I like that we create the policy this way now, and that the context is inside of the policy.
Although the optional argument is now missing. If the policy needs some other user input besides certificates, trust lists or keys, there is now no way of passing this to the policy(context). For example if the random generator needs random sources which the user must supply. I propose we add a void pointer and make an implementation specific function (something like UA_SecurityPolicy_None_mbedtls). (There might be different implementations of the same policy)

src/server/ua_server_binary.c
```diff
@@ -570,27 +539,70 @@ processMSG(UA_Server *server, UA_SecureChannel *channel,
...
+/*     UA_LOG_ERROR(server->config.logger, UA_LOGCATEGORY_NETWORK, */
+/*                  "Connection %i | Client replied with an error message: %s %.*s", */
+/*                  connection->sockfd, UA_StatusCode_name(errorMessage.error), */
+/*                  (int)errorMessage.reason.length, errorMessage.reason.data); */
+
+/*     UA_TcpErrorMessage_deleteMembers(&errorMessage); */
+/* } */
+
+/* static void */
+/* sendOPNError(UA_Server *server, UA_Connection *connection, */
```
Infinity95: Why is this commented out?

src/server/ua_services_session.c
```diff
@@ -63,27 +138,81 @@ void Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
     response->authenticationToken = newSession->authenticationToken;
     response->responseHeader.serviceResult =
         UA_String_copy(&request->sessionName, &newSession->sessionName);
-    if(server->config.endpoints.count > 0)
-        response->responseHeader.serviceResult |=
-            UA_ByteString_copy(&server->config.endpoints.endpoints[0].endpointDescription.serverCertificate,
-                               &response->serverCertificate);
 
+    /* Todo: Copy from the session's endpoint */
```
Infinity95: Why is this commented out?

Infinity95:
I just tried running the server and throw the CTT on it and ran into two problems:
First, the server seems to be having some trouble with the network layer:

> [08/31/2017 15:00:50.189] warning/network       Error binding a server socket: 10013
> [08/31/2017 15:00:50.189] warning/network       Error binding a server socket: 10013
> [08/31/2017 15:00:50.189] info/network  TCP network layer listening on opc.tcp://ILT579:4840
> [08/31/2017 15:00:50.189] warning/network       Socket select failed with No error
> [08/31/2017 15:00:50.189] warning/network       Socket select failed with No error

Seconldy none of the CTT tests work because the opening! (not connecting) of securechannels fails

jpfr:
Right. I only checked with UAExpert and our client.

@Infinity95 Do we have ideas for testing encryption?
We should have some degree of coverage in our unit tests....

Infinity95:
For the plugin api a mocked plugin should suffice to write some tests. For an actual implementation i would suggest using test vectors. I found something on the nist website that could be useful for that.

jpfr:
This is now sub-1000 lines of added code. Including headers.
Some rough edges need cleaning up. Stay tuned.

src/ua_securechannel.c
```diff
@@ -222,10 +210,9 @@ hideBytesAsym(UA_SecureChannel *const channel, UA_Byte **const buf_start,
        channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT) {
         *buf_end -= securityPolicy->asymmetricModule.cryptoModule.
             getLocalSignatureSize(securityPolicy, channel->channelContext);
-        *buf_end -= 2; // padding byte and extraPadding byte
+        *buf_end -= 1; // padding byte and extraPadding byte
```
Infinity95: Why 1 byte? We need to hide one byte for the padding byte and one for a potential extra padding byte.

jpfr: fixed

Infinity95:
Where did the functionality of sendOPNerror go?

Infinity95:
The makeCertificateThumbprint function is currently never used. The Thumbprint needs to be generated so the reciever can verify it:
Part 6 page 52
> The receiver shall verify the ReceiverCertificateThumbprint and report a Bad_CertificateUnknown error if it does not recognize it.

src/server/ua_services_session.c
```diff
@@ -63,27 +138,81 @@ void Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
...
+                   &request->clientSignature.signature);
+        if(retval != UA_STATUSCODE_GOOD) {
+            response->responseHeader.serviceResult = retval;
+            UA_LOG_DEBUG_SESSION(server->config.logger, session,
+                                 "Failed to verify the client signature! %#10x", retval);
+            UA_ByteString_deleteMembers(&dataToVerify);
+            return;
+        }
+
+        retval  = UA_SecureChannel_generateNonce(channel, 32, &response->serverNonce);
```
Infinity95: I believe it might be necessary to move this into the activateSession function so that the nonce is always regenerated, regardless of the verification status of the signature.
For example if the client tries to activate a session and it fails because of an invalid signature, and then tries to activate again, the nonce should be different according to the standard.

src/ua_securechannel.c
```diff
@@ -108,99 +155,365 @@ UA_Session * UA_SecureChannel_getSession(UA_SecureChannel *channel, UA_NodeId *t
...
+    size_t potentialEncryptionMaxSize = (size_t)(*buf_end - *buf_start) + UA_SEQUENCE_HEADER_LENGTH;
+
+    /* Hide bytes for signature and padding */
+    if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGN ||
+       channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT) {
+        *buf_end -= securityPolicy->asymmetricModule.cryptoModule.
+            getLocalSignatureSize(securityPolicy, channel->channelContext);
+        *buf_end -= 2; // padding byte and extraPadding byte
+
+        /* Add some overhead length due to RSA implementations adding a signature themselves */
```
Infinity95: signature -> padding

include/ua_plugin_securitypolicy.h
```diff
@@ -88,20 +88,20 @@ typedef struct {
...
     size_t (*getLocalEncryptionKeyLength)(const UA_SecurityPolicy *securityPolicy,
                                           const void *channelContext);
 
-    /* Returns the length of the key used to encrypt messages from remote in bytes
+    /* Returns the length of the key used to encrypt messages from remote in bits
      *
      * @param securityPolicy the securityPolicy the function is invoked on.
      * @param channelContext the context to retrieve data from.
-     * @return the size of the remote asymmetric signature. Returns 0 if no
+     * @return the length of the remote asymmetric key. Returns 0 if no
```
Infinity95: note to self: remove asymmetric, since it is for both symmetric and asymmetric.

jpfr:
Can you put f0689253103c8b541b3a21c3cb6f9a958a054aca into a separate branch/PR?
I think it is a real improvement. But it has nothing to do with the encryption feature.

That makes things unnecessarily complicated.
For example, the commit is the reason why this PR is not mergeable atm.

Infinity95:
The reason i added this is because in the amalgamation the global pos and end variables are shadowed by other variables in the new functions that are added with this pr, which is one of the reasons why travis fails. I can of course put it in a separate PR but then we either have to merge that right away or i have to change the variable names in the new functions.

Infinity95:
I would also propose that file local functions should be prefixed with the file name to avoid naming conflicts in the amalgamation (this happend to me as well). I could do a separate PR after i'm finished with this one to enforce that, if you agree with that.

Infinity95:
Moved f068925 to #1223

Infinity95:
@jpfr Are you ok with that rebase? I would add all upcoming fixes to the last commit from now on, since it's a collection of fixes anyways. The features should all be there now.

jpfr:
👍

Infinity95:
Strange. I cannot reproduce the failing tests on my machine. Something with the message type doesn't seem to be ok. Is there a way i can get test logs or anything that would help from travis?
@Pro Maybe you can help me with this?

The tests all work on my machine and the CTT is ok as well.

jpfr:
If all works fine on your box, feel free to tackle the mbedtls plugin next.
Working out the build kinks takes time since travis is slooow atm.
So better work at something productive in parallel. 😄

Infinity95:
I'm working on integrating mbedtls via git submodules and am currently having issues with install(EXPORT) stuff in cmake. Was reading documentation while waiting for travis 😄

Pro:
@Infinity95 I had a quick look at the travis outputs.
The errors only occur on the unit tests using previously dumped binary TCP messages, i.e. `check_server_binary_messages`.

The error messages are:
```
18: Test command: /usr/bin/valgrind "--error-exitcode=1" "--suppressions=/home/travis/build/open62541/open62541/tools/valgrind_suppressions.supp" "--quiet" "--trace-children=yes" "--leak-check=full" "--run-libc-freeres=no" "--sim-hints=no-nptl-pthread-stackcache" "/home/travis/build/open62541/open62541/build/bin/tests/check_server_binary_messages" "/home/travis/build/open62541/open62541/build/tests/client_HELOPN.bin" "/home/travis/build/open62541/open62541/build/tests/client_CreateActivateSession.bin" "/home/travis/build/open62541/open62541/build/tests/client_Browse.bin" "/home/travis/build/open62541/open62541/build/tests/client_CLO.bin"
18: Test timeout computed to be: 9.99988e+06
18: Running suite(s): Test server with messages stored in text files
18: [01/01/1601 00:00:00.000] info/channel	Creating a new SecureChannel
18: [01/01/1601 00:00:00.000] info/channel	Connection 0 | SecureChannel 1 | Opened SecureChannel
18: [01/01/1601 00:00:00.000] info/session	Try to use Session with token 1cd777e3-9590-a4e2-6db6-4be40ee09eb9 but is not found
18: [01/01/1601 00:00:00.000] info/session	Try to use Session with token 1cd777e3-9590-a4e2-6db6-4be40ee09eb9 but is not found
18: [01/01/1601 00:00:00.000] info/channel	Connection 0 | SecureChannel 1 | Service request 527 without a valid session
18: [01/01/1601 00:00:00.000] info/channel	Connection 0 | SecureChannel 1 | CloseSecureChannel
18: [01/01/1601 00:00:00.000] info/network	Connection 0 | Processing the message failed with error BadTcpMessageTypeInvalid
18: 100%: Checks: 1, Failures: 0, Errors: 0
```

In a working master build it looks like this:

```
18: Test command: /usr/bin/valgrind "--error-exitcode=1" "--suppressions=/home/travis/build/open62541/open62541/tools/valgrind_suppressions.supp" "--quiet" "--trace-children=yes" "--leak-check=full" "/home/travis/build/open62541/open62541/build/bin/tests/check_server_binary_messages" "/home/travis/build/open62541/open62541/build/tests/client_HELOPN.bin" "/home/travis/build/open62541/open62541/build/tests/client_CreateActivateSession.bin" "/home/travis/build/open62541/open62541/build/tests/client_Browse.bin" "/home/travis/build/open62541/open62541/build/tests/client_CLO.bin"
18: Test timeout computed to be: 9.99988e+06
18: LLVM Profile Note: Set profile file path to "default.profraw" via default setting.
18: Running suite(s): Test server with messages stored in text files
18: [01/01/1601 00:00:00.000] info/session	Connection 0 | SecureChannel 0 | Session 00000001-0000-0000-0000-000000000000 | WriteRequest returned status code BadTypeMismatch
18: [01/01/1601 00:00:00.000] info/session	Connection 0 | SecureChannel 0 | Session 00000001-0000-0000-0000-000000000000 | AddNodes: Type-checking the variable node failed with error code BadTypeMismatch
18: 100%: Checks: 1, Failures: 0, Errors: 0
18/24 Test #18: check_server_binary_messages_browse ...   Passed    1.54 sec
```

So **I guess** that with encryption something changed that we now handle the session token in a different way. The session token is hardcoded in the binary files ( e.g. `/home/travis/build/open62541/open62541/build/tests/client_CreateActivateSession.bin` or the following .bin files) and thus all the calls in travis now fail.

Maybe we can just remove the `check_server_binary_messages` unit test since the same funcionality (`Browse`, `Read`, `Write`) is already handled by various other tests, e.g. `client_highlevel`. @jpfr?

Infinity95:
Wouldn't that mean that those errors have to show up locally as well? Or are the files somehow different on travis?

Pro:
The files are the same on travis, so it should also show up locally.

Travis uses the following cmd, maybe you can try it on your machine again (currently I'm on windows so I can't test it):

```
    cmake -DPYTHON_EXECUTABLE:FILEPATH=/usr/bin/python3 -DCMAKE_BUILD_TYPE=Debug -DUA_BUILD_EXAMPLES=ON -DUA_ENABLE_DISCOVERY=ON -DUA_ENABLE_DISCOVERY_MULTICAST=ON -DUA_BUILD_UNIT_TESTS=ON -DUA_ENABLE_COVERAGE=ON -DUA_ENABLE_VALGRIND_UNIT_TESTS=ON ..
    make -j && make test ARGS="-V"
```

Infinity95:
That seems to reproduce it. My first guess is that i didn't have the discovery enabled. I'll have to check what's causing this tomorrow.

Infinity95:
@Pro Can you help me out understanding the tests? As far as i understand it these are just binary messages that get played into UA_Server_processBinaryMessage, right? Then i don't understand how these tests worked before, since the Session token is randomized on creation and then sent back in the response, so the client can activate that session. That means that the played back activateSession won't work and the following messages won't work as well, since no session could be established. I'm also getting different error messages locally: Instead of BadTCPMessageType i get SecurityChecksFailed, which makes more sense to me.

Pro:
I think that those tests stopped working longer ago (or never really did what the should). The only issue now is that there is a memory leak and thus travis shows the error. If you look closely, the check has no error, but the memory leak causes the problem:

```
19: 100%: Checks: 1, Failures: 0, Errors: 0
19: ==26907== 217 (24 direct, 193 indirect) bytes in 1 blocks are definitely lost in loss record 3 of 3
19: ==26907==    at 0x4C2AB80: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
19: ==26907==    by 0x42B383: UA_Server_delayedCallback (ua_server_worker.c:164)
19: ==26907==    by 0x42CEA8: removeSecureChannel (ua_securechannel_manager.c:44)
19: ==26907==    by 0x42DC6C: UA_SecureChannelManager_close (ua_securechannel_manager.c:235)
19: ==26907==    by 0x448A34: Service_CloseSecureChannel (ua_services_securechannel.c:54)
19: ==26907==    by 0x429525: processSecureChannelMessage (ua_server_binary.c:513)
19: ==26907==    by 0x41C464: UA_SecureChannel_finalizeChunk (ua_securechannel.c:680)
19: ==26907==    by 0x41CFC4: UA_SecureChannel_processChunk (ua_securechannel.c:919)
19: ==26907==    by 0x429B00: processCompleteChunk (ua_server_binary.c:627)
19: ==26907==    by 0x418F3A: processChunk (ua_connection.c:140)
19: ==26907==    by 0x419029: UA_Connection_processChunks (ua_connection.c:165)
19: ==26907==    by 0x429B97: processBinaryMessage (ua_server_binary.c:641)
19: ==26907== 
```

The problem here is that a delayed callback is added but since the test never calls `UA_Server_run_iterate` this callback is never processed and thus leaks memory.

To solve this we should add the following to `UA_Server_delete` so that all callbacks in the queue are processed and deleted. This should then also fix the memleak here.

```
/* The server needs to be stopped before it can be deleted */
void UA_Server_delete(UA_Server *server) {
    /* Process delayed callbacks when all callbacks and
     * network events are done */
    processDelayedCallbacks(server);
#endif

    /* Delete all internal data */
    //....
```


Regarding the tests, which do not use the returned session token, we need to use some define which disables session checking (e.g. like it is done for fuzzing) or remove them completely:
https://github.com/open62541/open62541/pull/1147/files#diff-4e6990e0233b07366cc61b6a7cb8e928

Infinity95:
Althoug this change might not be necessary i believe. In normal operation we would call server_run and that would then call the appropriate shutdown function which processes remaining callbacks. It's just that the test doesn't use the normal functionality which then leads to that memory leak.
The tests don't give much value imo. The only thing they could uncover are potential memory leaks and segfaults, which should be covered by the other tests.

jpfr:
Usually, users call `UA_Server_run` to run the server. This does the clean-up internally.

Only when users want to drive the main-loop themelves, they have to call `UA_Server_run_shutdown` manually.
https://github.com/open62541/open62541/blob/master/src/server/ua_server_worker.c#L357

Can we keep the shutdown functionality in the shutdoyn method?
Maybe we can check in `UA_Server_delete` whether the server has been properly shut down. And do it ourselves if necessary.

jpfr:
Hey Mark,

whtat's the status? I see travis runs clean.
Only appveyor has some warnings..

Infinity95:
I fixed the tests by calling the server_shutdown_run. Everything else seems to be fine and Appveyor is failing on the doxygen stuff i think. Could you restart the Appveyor tests to see if that fixes it?

I still need to finish the last implementaion stuff for the securitypolicy#basic128rsa15 to test it with the CTT but i currently have very little time on hand. I will have time after the 10. of this month.

The branch could be merged, if you want, and any bugs and the corresponding fixes that would emerge from the CTT could be done in another separate PR, since those shouldn't be API breaking. Alternatively we can wait until next week until im done with the plugin and testing.

jpfr:
All green. Merged!

Thanks for the continued effort @Infinity95 !
The code is in a really good shape now. Tight and small.

In the /src directory we have only 450 lines of added code.
The remaining 500 lines are headers and the #none securitypolicies.

The remaining securitypolicies will be a bit heavier though.

**cloc of /src new:**
```
github.com/AlDanial/cloc v 1.70  T=0.15 s (310.4 files/s, 129163.3 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
C                               33           2045           1780          13338
C/C++ Header                    13            352            597           1030
-------------------------------------------------------------------------------
SUM:                            46           2397           2377          14368
-------------------------------------------------------------------------------
```

**cloc of /src old:**
```
github.com/AlDanial/cloc v 1.70  T=0.13 s (347.1 files/s, 142756.7 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
C                               32           1974           1726          12918
C/C++ Header                    13            336            561            991
-------------------------------------------------------------------------------
SUM:                            45           2310           2287          13909
-------------------------------------------------------------------------------
```

PR #1191, opened 2017-09-04
fixed default variable/variabletype attributes

janitza-thbe:
fixed default variable/variabletype attributes according to specified default values in OPC foundation's XSD (http://opcfoundation.org/UA/2011/03/UANodeSet.xsd)
No further tests or example builds done due to defaults changes.

### DIFF
--- a/src/server/ua_server_utils.c
+++ b/src/server/ua_server_utils.c
@@ -382,9 +382,10 @@ const UA_VariableAttributes UA_VariableAttributes_default = {
      0, NULL, 0, NULL},          /* value */
     {0, UA_NODEIDTYPE_NUMERIC,
      {UA_NS0ID_BASEDATATYPE}},   /* dataType */
-    -2,                          /* valueRank */
+    -1,                          /* valueRank */
     0, NULL,                     /* arrayDimensions */
-    UA_ACCESSLEVELMASK_READ, 0,  /* accessLevel (userAccessLevel) */
+    UA_ACCESSLEVELMASK_READ,     /* accessLevel */
+    UA_ACCESSLEVELMASK_READ,     /* userAccessLevel */
     0.0,                         /* minimumSamplingInterval */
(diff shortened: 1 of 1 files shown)


jpfr:
👍 
We can merge when Travis gives a green light.
Currently the return value -2 is expected in a unit test.

jpfr:
This PR sets the default valueRank to -1 == 'must be scalar'.
That means you cannot set an array value without explicitly changing the valuerank first.
This will confuse many first-time users of OPC UA. And break existing code.
For example some of our examples.

Would it be okay to leave _default at -2 == 'can be anything'?

janitza-thbe:
We have a fork anyway, so we don't mind at all.
But it's implicitly wrong to set it to other values since imported XML files with unset ValueRank attribute (like NodeSet2) expect it to be set to the XSD's specified default value of -1. So here the nodeset compilers have to implement a workaround for setting it explicitly to -1 if the attribute is missing. Otherwise the generated/loaded models are inherent wrong.

Stasik0:
What does the standard say? I mean having -1 per default in xsd still lets a possibility open for having c api default to -2?

jpfr:
basedatavariable

BaseVariableType, PropertyType and BaseDataVariableType all have a value rank of -2.

jpfr:
And they must have a value rank of -2.
If they didn't, none of their subtypes could have a value rank of -2.
Because we can only make the constraints stronger. And -2 is the least constrained.

Pro:
I ran into the same issue in the Nodeset Compiler. 
For the nodeset2.xml the default value rank should be -2 if it is not explicitly given. Otherwise creation of nodes will fail.

Stasik0:
Xsd says -1 https://github.com/OPCFoundation/UA-Nodeset/blob/master/Schema/UANodeSet.xsd

Should we raise an issue in opcf's github?

PR #1193, opened 2017-09-05
Make changes for a Fedora RPM

ctron:
This change adds an RPM file and makes some minor modifications to the build in order to create an RPM file which passed the Fedora review process.

For this spec file to work there has to be a tag named `open62541-<version>`, e.g. `open62541-0.2`. However this can be locally emulated by providing a sources file of the same name, e.g. `open62541-0.2.tar.gz`.

An easier option is to use `tito` for a local build and execute `tito --test --rpm` in order to build a local version of the latest git commit automatically.

Also see: https://bugzilla.redhat.com/show_bug.cgi?id=1487578


jpfr:
Hello @ctron, looks good!

Just a few questions remain.

- Is it possible to move the .tito directory to /tools/tito and give this information in the open62541.spec?
- When should the information in open62541.spec be updated? At release? When we split a new major branch from master? We should then keep a list of "steps to make a release" in the project wiki in github.
- Is the .tito/packages/.readme really needed? The .tito/packages folder is empty atm

jpfr:
Last question, is there a `tito` command line argument to build a local rpm that does need the .tito files at all?

I like the tool. But the config looks pretty generic and might not be needed to build nightly packages.

ctron:
I am currently testing this PR on the master branch and found a bunch of things that have changed. I will create an additional PR for the master branch.

> Is it possible to move the .tito directory to /tools/tito and give this information in the open62541.spec?

From what I know the `.tito` directory has to stay at the root of the repository.

> When should the information in open62541.spec be updated? At release? When we split a new major branch from master? We should then keep a list of "steps to make a release" in the project wiki in github.

Well there are two scenarios when the spec file needs an update. 1) When the build changes or 2) when you have a new version.

In the latter case I would suggest to increment the version as soon as a the next development cycle is started. So that when you branch off the release branch it already is up to date.

> Is the .tito/packages/.readme really needed? The .tito/packages folder is empty atm

I am not sure, and I will try to remove it.

ctron:
Yes, it looks like the readme can be removed.

jpfr:
Appveyor fails with the following message:
```
CMake Error at CMakeLists.txt:461 (install):
  install TARGETS given no ARCHIVE DESTINATION for static library target
  "open62541"
```

ctron:
Yes, I did fix that already for the PR for the master branch: #1195 I will amend this PR later that day.

ctron:
Sorry, it took me a bit longer. But here is the fix for this.

ctron:
Looks better now.

PR #1194, opened 2017-09-05
Fix interpreter locations

ctron:
This changes fixes the interpreter locations for a set of Python files.

### DIFF
--- a/tools/gdb-prettyprint.py
+++ b/tools/gdb-prettyprint.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 # WARNING: This is till work in progress
 #
 # Load into gdb with 'source ../tools/gdb-prettyprint.py'
--- a/tools/hex2bin.py
+++ b/tools/hex2bin.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 # This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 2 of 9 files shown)


Pro:
Thanks for the PR!
Note that most of the files in ` tools/pyUANamespace` will be overwritten/replaced by the refactoring PR #739. Thus we also need to integrate/change this in that PR branch.

I would change it to a more general shebang line, like:
`#!/usr/bin/env python` (note the difference, there's no slash after `env`). Does this also work for you @ctron?

IMHO your proposed shebang line will only work if python is in `/usr/bin/python`?

See also the discussion here:
https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take
and 
https://stackoverflow.com/a/2429517/869402

jpfr:
So `#!/usr/bin/env python`?

jpfr:
Imho this can be merged ahead of #739.

Pro:
Yes, I think `#!/usr/bin/env python` is the more general way.

Merging it before #739 is also ok. But still we need to port the correct shebangs to #739

ctron:
I will give it a test with `#!/usr/bin/env python`

ctron:
Yes, that works. I did update the PR.

jpfr:

PR #1199, opened 2017-09-07
Fix API inconsistency: Read arraydimensions

jpfr:


jpfr:
This addresses @Pro's comment in #1168 .

This PR proposes a small API change for the client.
Imho the change is low-risk. The compiler's type-checking will catch this immediately.

Comments by @open62541/open62541-core ?

Pro:
Is it possible to keep the old version and mark it as deprecated? We should try to keep the deprecated versions for all such changes until the next release.
That makes such changes more user friendly.

i.e. I propose the following (as we also discussed it in last telco):

- Currently we are post 0.2 release, thus all the API changes resulting in compilation errors of user code relying on 0.2 should be marked as deprecated but still kept.
- Version 0.3 will be released (at some point) which still includes the deprecated definitions.
- Right after the release of 0.3 we remove all deprecated functions in master and handle the API changes for 0.3 in the same way

This is already too late for 0.2 since there were many changes in the API, but at least we should try to handle it that way from now on. Does this sound like a plan?

So if someone had used 0.2, waits until the release of 0.3, and then uses 0.3 his code still compiles, but shows deprecated warnings.

jpfr:
I agree with the general approach for deprecating methods that shall be removed entirely.
Here we change the API. Similar to other API changes that have happened in the v0.3 release cycle.

If the current version shall be deprecated, how do we name the new version?
`UA_Client_readArrayDimensionsAttribute_v2`?

Pro:
Yep, since C does not support overloading, there is unfortunately no other way than to just change it without deprecation.
Since this is quite a small change I'm fine with that.

If there are bigger changes (like the one for node context) we should think of a way to at least provide a deprecated function name which can be used to use the old API (and document it soemwhere). Then a developer just has to change the function instead of the complete code around it.

Or we find a usable way with `#ifdef` to switch between the current 0.2 api and bleeding edge. Hmm...

I'm not sure what the best solution is. Breaking changes are always bad.

PR #1206, opened 2017-09-14
use integer shortnames in ua_types_encoding_binary.c

jpfr:


jpfr:
This commit is purely cosmetic.
It uses shortnames (u16, i32, ...) for integers. But the definitions are not exported, since users might have similar definitions in their projects already.

Comments @open62541/open62541-core ?

Pro:
:+1:

PR #1210, opened 2017-09-15
Fix #1180 ensure enums are 32bit integers

### LINKED ISSUE #1180: Ensure 32bit enums
Enums in C have at least the size of ``int``.
This can be 16bit on some architectures.
Currently, we just copy a 32bit value for the wire protocol.
This can eventually lead to serious bugs.

We can add more complications to the `UA_DataType` descriptions to handle enums in separate code-paths. Or we use int32_t instead of enums.

The current proposal is change the enum definitions from
```c
typedef enum {
    UA_NODECLASS_UNSPECIFIED = 0,
    UA_NODECLASS_OBJECT = 1,
    UA_NODECLASS_VARIABLE = 2,
    UA_NODECLASS_METHOD = 4,
    UA_NODECLASS_OBJECTTYPE = 8,
    UA_NODECLASS_VARIABLETYPE = 16,
    UA_NODECLASS_REFERENCETYPE = 32,
    UA_NODECLASS_DATATYPE = 64,
    UA_NODECLASS_VIEW = 128
} UA_NodeClass;
```
to
```c
#define UA_NODECLASS_UNSPECIFIED 0
#define UA_NODECLASS_OBJECT 1
#define UA_NODECLASS_VARIABLE 2
#define UA_NODECLASS_METHOD 4
#define UA_NODECLASS_OBJECTTYPE 8
#define UA_NODECLASS_VARIABLETYPE 16
#define UA_NODECLASS_REFERENCETYPE 32
#define UA_NODECLASS_DATATYPE 64
#define UA_NODECLASS_VIEW 128
typedef UA_Int32 UA_NodeClass
```

This has no impact on the code besides the python-based datatype generation.
Comments and ideas @open62541/open62541-core (and everybody else)?

Pro:
If that's not the case, the error during compile time will be similar to
`error: named bit-field 'static_assertion_failed_enum_must_be_32bit' has zero width`

### DIFF
--- a/include/ua_config.h.in
+++ b/include/ua_config.h.in
@@ -267,6 +267,20 @@ extern "C" {
 # define UA_BINARY_OVERLAYABLE_FLOAT 0
 #endif
 
+/**
+ * Static Assert
+ * ^^^^^^^^^^^^^
+ *
+ * Outputs an error message at compile time if the assert fails.
+ * Example usage:
+ * UA_STATIC_ASSERT(sizeof(long)==7, use_another_compiler_luke) *
+ * See: https://stackoverflow.com/a/4815532/869402 */
+#define UA_CTASTR2(pre,post) pre ## post
+#define UA_CTASTR(pre,post) UA_CTASTR2(pre,post)
+#define UA_STATIC_ASSERT(cond,msg) \
+    typedef struct { int UA_CTASTR(static_assertion_failed_,msg) : !!(cond); } \
+        UA_CTASTR(static_assertion_failed_,__COUNTER__)
+
 #ifdef __cplusplus
 } // extern "C"
 #endif
--- a/tools/generate_datatypes.py
+++ b/tools/generate_datatypes.py
@@ -191,7 +191,9 @@ class EnumerationType(Type):
         else:
             values = self.elements.items()
         return "typedef enum {\n    " + ",\n    ".join(map(lambda kv : "UA_" + self.name.upper() + "_" + kv[0].upper() + \
-                                                            " = " + kv[1], values)) + "\n} UA_%s;" % self.name
+                                                            " = " + kv[1], values)) + \
+               ",\n    __UA_{0}_FORCE32BIT = 0x7fffffff\n".format(self.name.upper()) + "} " + \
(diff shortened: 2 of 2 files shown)


jpfr:
Looks perfect. 👍 
Can we add the static assert as a one-liner to the generated enums in generate_datatypes.py?

Stasik0:
should we insert a 0xffffffff value to ensure it is at least 32 bit?https://github.com/open62541/open62541/issues/1180#issuecomment-324854968

Pro:
@jpfr do you mean, insert it so that every enum we generate will be checked with the static assert?
i.e. add `"UA_STATIC_ASSERT(sizeof(UA_{}) == sizeof(UA_Int32), enum_must_be_32bit);".format(self.name)` here:
https://github.com/open62541/open62541/blob/3ea21ea597ad67d28d8a84e14e559e2553c1851e/tools/generate_datatypes.py#L193

@Stasik0 is the 0xffffffff still necessary if we check with sizeof?

jpfr:
Yes. Adding the assertion to every generated enum covers UA_NodeClass as well.

The 0xffffffff forces a lower limit of the enum size.
See the GCC docs for the flag that enables enum-packing:

----

**-fshort-enums**

Allocate to an enum type only as many bytes as it needs for the declared range of possible values. Specifically, the enum type is equivalent to the smallest integer type that has enough room.

Warning: the -fshort-enums switch causes GCC to generate code that is not binary compatible with code generated without that switch. Use it to conform to a non-default application binary interface.

----

We could add a dummy as @Stasik0 suggested.
Eg with leading underscores to avoid a mess with code-completion.
`__UA_NODECLASS_FORCE32BIT = 0xffffffff`

We need the sizeof assertion in addition.
Otherwise, the compiler could use an enum with >32 bits. Which is equally bad.

Stasik0:
I mean with the 0xffffffff we will hopefully see the failing assertion less often

Pro:
Ah, ok, got it!
We have to use max signed int, instead of unsigned. So the value has to be `0x7fffffff`otherwise gcc complains with:
`ISO C restricts enumerator values to range of 'int'` due to the `-pedantic` flag,

PR #1214, opened 2017-09-16
Fix scripts to support python 3

Pro:
Also adds check on Linux for CI, to explicitly test python2 and python3

### DIFF
--- a/tests/fuzz/fuzz_binary_decode.cc
+++ b/tests/fuzz/fuzz_binary_decode.cc
@@ -21,7 +21,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 	size_t ptrSize = size;
 
 	// get some random type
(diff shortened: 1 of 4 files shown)


tests/fuzz/fuzz_binary_decode.cc
```diff
@@ -21,7 +21,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 	size_t ptrSize = size;
 
 	// get some random type
-	uint8_t typeIndex = ptr[0];
+	uint16_t typeIndex = ptr[0];
```
HolgerJeromin: Is this python 3 related?

Pro: No, but I found it because there were never more than 255 types before.
so not related to python3, that's why it is in a separate commit

PR #1215, opened 2017-09-17
Respect accessControl in nodemanagement service, Fix #1177

### LINKED ISSUE #1177: AccessControl Plugin: allowAddNode/Reference no effect
Currently you can set access rights in the Access Control Plugin (https://github.com/open62541/open62541/blob/394df3be8d6a19781927f0790d9c2a9142a2d6a8/include/ua_plugin_access_control.h#L54) but the settings for the following rights are not used:

* allowAddNode
* allowAddReference
* allowDeleteNode
* allowDeleteReference

This is an open issue and can be grabbed by an interested developer.

ccvca:
Fix Issue #1177

The extra functions (restricted*) prevent checking access rights when adding nodes/references by the server.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -73,6 +73,11 @@ add_example(server_inheritance server_inheritance.c)
 add_example(custom_datatype_client custom_datatype/client_types_custom.c)
 add_example(custom_datatype_server custom_datatype/server_types_custom.c)
 
+if(UA_ENABLE_NODEMANAGEMENT)
+    add_example(access_control_server access_control/server_access_control.c)
+    add_example(access_control_client access_control/client_access_control.c)
+endif()
+
 if(UA_BUILD_EXAMPLES_NODESET_COMPILER)
   if(BUILD_SHARED_LIBS)
     message(FATAL_ERROR "The nodeset compiler currently requires static linking to access internal API")
--- a/examples/access_control/client_access_control.c
+++ b/examples/access_control/client_access_control.c
@@ -1,0 +1,67 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+/**
+ * Using access_control_server
+ */
+
+#include <stdio.h>
+
+#include "open62541.h"
+
+int main(void) {
+    UA_Client *client = UA_Client_new(UA_ClientConfig_default);
+    UA_StatusCode retval = UA_Client_connect(client, "opc.tcp://localhost:4840");
+    if(retval != UA_STATUSCODE_GOOD) {
+        UA_Client_delete(client);
+        return (int)retval;
+    }
+
+    UA_NodeId newVariableIdRequest = UA_NODEID_NUMERIC(1, 1001);
+    UA_NodeId newVariableId = UA_NODEID_NULL;
+
+    UA_VariableAttributes newVariableAttributes = UA_VariableAttributes_default;
+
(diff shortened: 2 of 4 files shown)


examples/access_control/client_access_controll.c
```diff
@@ -0,0 +1,72 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
```
Pro: please rename to `client_access_control.c` (only one L at the end)

examples/access_control/server_access_controll.c
```diff
@@ -0,0 +1,71 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
```
Pro: please rename to `server_access_control.c` (only one L at the end)

examples/access_control/client_access_controll.c
```diff
@@ -0,0 +1,72 @@
...
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+/**
+ * Using access_control_server
+ */
+
+#include <stdio.h>
+
+#ifdef UA_NO_AMALGAMATION
```
Pro: only use the open62541.h in examples.

Refer to other examle files to see how it is used there.

examples/access_control/server_access_controll.c
```diff
@@ -0,0 +1,71 @@
...
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+#include <signal.h>
+#include <stdio.h>
+
+#ifdef UA_NO_AMALGAMATION
```
Pro: only use the amalgamated header here

ccvca:
The requested changes are forced pushed.

I've implement the changes at the original methods first and noticed that they are called way to often, because when the Server adds a node, the permission-Check is executed.

How can I determine if the node has been added by server or by a client? Does the server always uses the adminSession? If so, it could be integrated in the original methods.

ccvca:
```
/* Do not check access for server */
	if (session != &adminSession &&
		server->config.accessControl.allowDeleteNode &&
		!server->config.accessControl.allowDeleteNode(&session->sessionId, session->sessionHandle,
			item))
	{
....
```
When I integrate the check in the current methods the output of the example looks like this, because addNode also adds a 2 references (forward/backward)?
Is this the intended behaviour? What should happen if userAccess allows adding Nodes but do not allow to add References?

```
Called allowAddNode
Called allowAddReference
Called allowAddReference
Called allowAddReference
Called allowDeleteReference
Called allowDeleteNode
```

jpfr:
Only the `UA_Server_` methods use the adminSession.
For the attributes services, we allow all operations for the adminSession.
This is hard-coded as you suggested: https://github.com/open62541/open62541/blob/master/src/server/ua_services_attribute.c#L22

@Pro I can adjust #1168 after this PR is merged. Still waiting for feedback from Aachen.
In general, we should make the process for first-time contributors as easy as possible. 😄

ccvca:
Ok, so the question is, what is the desired behaviour?
Should UA_Client_addXXXNode result in checks of allowAddNode and allowAddReference (for the forward/backward references) or should it just result in a check of allowAddNode?

The first approach sounds more logical for advanced users,
the second is more straight forward _one Service_ <-> _one permission_

jpfr:
The first approach looks good to me.
Currently, when there is an error during instantiation, the new node is removed...

ccvca:
Ok, the first approach is (force) pushed to this request and the errors of Travis/AppVeyor are fixed.

jpfr:
From what I see, @Pro's comments were taken into account.
The PR looks perfect.

jpfr:
Thanks for the contribution @ccvca 😄

PR #1223, opened 2017-09-20
Prefix globals with g_ to avoid shadowing in amalgamation.

Infinity95:


jpfr:
Can you rebase on the (new) current master?
That should fix the build problem.

PR #1225, opened 2017-09-22
Add support for TinyCC

### LINKED ISSUE #1217: Unit tests don't run with TCC compiler (Float Encoding and Discovery)
TCC is a cool small C99 compiler. https://bellard.org/tcc/
Much faster than gcc, so good for development.
Our code compiles and the servers work out of the box.

But, some unit tests fail.
```
$ ./bin/tests/check_types_builtin
Running suite(s): Built-in Data Types 62541-6 Table 1
98%: Checks: 58, Failures: 1, Errors: 0
/home/jpfr/software/open62541/tests/check_types_builtin.c:825:F:encode:UA_Float_encodeShallWorkOnExample:0: Assertion 'dst.data[3] == result[i][3]' failed: dst.data[3] == 127, result[i][3] == 255
```

and

```
$ ./bin/tests/check_discovery
Running suite(s): Register Server and Client
[01/01/1601 00:00:00.000] info/network  TCP network layer listening on opc.tcp://virtualbox:4840
[01/01/1601 00:00:01.000] info/network  TCP network layer listening on opc.tcp://virtualbox:16664
[01/01/1601 00:00:01.000] info/network  Connection 9 | New connection over TCP from ::1
[01/01/1601 00:00:01.000] info/channel  Connection 9 | SecureChannel 1 | Opened SecureChannel
[01/01/1601 00:00:01.000] info/session  Try to use Session with token 00000000-0000-0000-0000-000000000000 but is not found
[01/01/1601 00:00:01.000] info/session  Connection 9 | SecureChannel 1 | Session 43be1b59-cda6-0265-38a3-9a19f70fa0ba | ActivateSession: Session activated
Speicherzugriffsfehler
```

Pro:
This PR adds unit tests and support for TinyCC (tcc) compiler

Fixes #1217

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,11 @@ matrix:
         - ANALYZE=false
         - PYTHON=python3
     - os: linux
+      compiler: tcc
+      env:
+        - ANALYZE=false
+        - PYTHON=python2
+    - os: linux
       compiler: clang
       env: ANALYZE=true
     - os: linux
@@ -73,6 +78,7 @@ addons:
       - mingw-w64
       - python-six
       - python3-six
+      - tcc
       - texlive-fonts-recommended
       - texlive-latex-extra
       - texlive-latex-recommended
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,6 @@ addons:
       - mingw-w64
       - python-six
       - python3-six
-      - tcc
       - texlive-fonts-recommended
       - texlive-latex-extra
       - texlive-latex-recommended
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -161,13 +161,12 @@ typeCheckVariableNode(UA_Server *server, UA_Session *session,
     /* If variable node is created below BaseObjectType and has its default valueRank of -2,
      * skip the test */
     const UA_NodeId objectTypes = UA_NODEID_NUMERIC(0, UA_NS0ID_BASEOBJECTTYPE);
-    const UA_NodeId refs[2] = {
-        UA_NODEID_NUMERIC(0, UA_NS0ID_HASSUBTYPE),
-        UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT)
-    };
+    UA_NodeId refs[2];
+    refs[0] = UA_NODEID_NUMERIC(0, UA_NS0ID_HASSUBTYPE);
+    refs[1] = UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT);
     if(node->valueRank != vt->valueRank &&
        node->valueRank != UA_VariableAttributes_default.valueRank &&
-       !isNodeInTree(&server->config.nodestore, parentNodeId, &objectTypes, refs , 2)) {
+       !isNodeInTree(&server->config.nodestore, parentNodeId, &objectTypes, refs, 2)) {
         /* Check valueRank against the vt */
         if(!compatibleValueRanks(node->valueRank, vt->valueRank))
             return UA_STATUSCODE_BADTYPEMISMATCH;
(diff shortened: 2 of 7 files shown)


jpfr:
Right. I see the code is getting a lot better with every compiler we add. :-)

Now the travis errors are getting seriously bizarre:
```
ld.so: rtld.c: 1338: dl_main: Assertion `_rtld_local._dl_rtld_map.l_libname'
```

Pro:
Yes, unfortunately I cannot reproduce this on my Ubuntu 16.04.

Googling around led me to the suspicion that it may be a problem with the stdlibc version, but unfortunately I won't have much time the coming days to check that out further. So @jpfr if you have time, you can see if you find a fix.

Stasik0:
Problems with gnu libs?

https://code.woboq.org/userspace/glibc/elf/rtld.c.html#1263

Stasik0:
Maybe we should also start testing with other C libs, eg https://www.uclibc.org/

jpfr:
Yeah!
Here's an (older) overview over common libc:

- http://www.etalabs.net/compare_libcs.html
- http://wiki.osdev.org/C_Library

Then, there are the libc of the various *BSD and embedded OS...

Pro:
@jpfr I just updated the branch and now it looks like tcc has problems with TLS (Thread Local Storage).
tcc does not support any kind of TLS, and when you execute the `check_client_async` it fails randomly with different types of errors. So my guess is that this is related to the TLS?

jpfr:
> @jpfr I just updated the branch and now it looks like tcc has problems with TLS (Thread Local Storage).

Without TLS, tests with parallel threads will fail.
They will be race conditions during encoding.
Most compilers without TLS target embedded systems where you don't have threads anyway.

TinyCC does not support TLS. This is the latest discussion I can find:
https://lists.nongnu.org/archive/html/tinycc-devel/2013-09/msg00104.html

Maybe we can disable tests with parallel threads for TCC.

Stasik0:
What about emulating tls with pthreads? Will also be useful for older compilers.

Pro:
@jpfr @Stasik0 using pthreads for TLS would look like this:
https://dulanja.blogspot.de/2011/09/how-to-use-thread-local-storage-tls-in.html

This means that we have to use `pthread_setspecific` and `pthread_getspecific` within the threaded function. That's not triavial and makes the code not so easy to read.

Anyway TLS should not be required if we do not compile with multithreading. So if multithreading is off, all the tests should work, even if TCC does not support TLS?

jpfr:
@Pro 
Rebased. This now hangs at a last unit test.
They all work on my box. Do you have an idea?

Pro:
Had exactly the same problem on my PC (Ubuntu 16.04 x64).
The problem is on this line, right?
https://github.com/open62541/open62541/blob/c34dcd0e993c12c250d5db3b794016f0004941ad/tests/check_types_builtin.c#L801

I.e. the encoding of `-0.0` fails.

Could it be an issue with the architecture used by travis, e.g. x86 vs x64

StalderT:
Just to see which version of tcc is used, can you fix https://travis-ci.org/open62541/open62541/jobs/332168988#L1005

```
tcc -v
```

Pro:
From the install log on travis you can also see the version:

> Unpacking tcc (0.9.26~git20120612.ad5f375-6) 

According to the github repo that's the up-to-date release, but not the up-to-date commit

jpfr:
0.9.26 is four years old.
tcc 0.9.27 was release five weeks ago.
There were a ton of fixes.

These are the release notes from the official repo.
http://repo.or.cz/tinycc.git/blob/HEAD:/Changelog

I'd run tests. But the problem does not show on my platform (amd64).

StalderT:
```
root@debian:/data/open62541# cat b.c
#include <math.h>
#include <stdio.h>

int main(void)
{
    float f;
    unsigned char *ptr = (unsigned char *)&f;

    f = 0.0f;
    printf("%02X %02X %02X %02X (0.0f)\n", ptr[0], ptr[1], ptr[2], ptr[3]);

    f = -0.0f;
    printf("%02X %02X %02X %02X (-0.0f)\n", ptr[0], ptr[1], ptr[2], ptr[3]);

    f = -NAN;
    printf("%02X %02X %02X %02X (-NAN)\n", ptr[0], ptr[1], ptr[2], ptr[3]);

    f = NAN;
    printf("%02X %02X %02X %02X (NAN)\n", ptr[0], ptr[1], ptr[2], ptr[3]);
}
```

```
root@debian:/data/open62541# tcc -v
tcc version 0.9.27 (prerelease) (x86-64 Linux)
root@debian:/data/open62541# tcc b.c && ./a.out 
00 00 00 00 (0.0f)
00 00 00 80 (-0.0f)
00 00 C0 7F (-NAN)
00 00 C0 7F (NAN)
```

```
root@ubuntu:/data/open62541# tcc -v
tcc version 0.9.25
root@ubuntu:/data/open62541# tcc b.c && ./a.out 
00 00 00 00 (0.0f)
00 00 00 00 (-0.0f)
00 00 C0 7F (-NAN)
00 00 C0 7F (NAN)
```

Probably a bug fixed in tcc (not included/fixed in tcc version included in ubuntu)

Pro:
Thanks for the tests. I will try to change travis so that it uses the newer tcc

Pro:
Looks better now, when all green we can finally merge

jpfr:
Building tcc from source is "brute force". 😄 
All is green now. So merging!

PR #1237, opened 2017-10-08
Fix decoding of zero float and double values.

smarsching:
Due to a missing else in front of an if branch in `Float_decodeBinary`
and `Double_decodeBinary`, special cases (like zero values) where not
handled correctly. Instead, `unpack754` was always used unless the value
was NaN.

This patch fixes this problem by adding the missing `else`.


jpfr:
Good find. Thanks for the fix!
We just have to shepherd this through our continuous integration suite..

jpfr:
Meged. Thanks for the contribution!

smarsching:
You are welcome. 😃 
Thanks for merging.

PR #1238, opened 2017-10-10
Rebase Nodeset compiler v2 on master

jpfr:
See #739


Pro:
@jpfr @ichrispa All green again.
Please review and then I would like to merge this.
Any smaller changes as discussed in the telko will then go into a new PR

jpfr:
Code-wise this looks ready to me. 👍 

What do the others think about moving the schema-files to an extra repository?
I'd prefer keeping them here (possibly moving them to /deps/schema as was previously done in this PR).
So users don't need git+internet once they have a copy of our source.
I assume there are people with no git on windows that just download the zip.

jpfr:
Maybe we cann include the license text for the schema definitions in the /deps/schema folder or wherever they end up..

Pro:
I added the schema files directly as a git submodule from the git repo here:
(Currently it is referenced to the fork https://github.com/Pro/UA-Nodeset because I needed to wait for a fix in the nodeset, I will change that before merging since the fix is now added to the OPCFoundation repo)

This makes it much easier to keep the schema definition in sync with the official reference. I remember that there were some issues in the past with the .xml and .bsd files.

As you already pointed out, it would mean that if you download the .zip you also need to download the .zip of ua-nodeset repo. In my opinion if someone wants to download the complete source and develop based on it, he should be able to check out the repo and all its submodules on another PC. Otherwise he downloads the amalgamated files which include the submodules anyway.

Btw we have the same problem with the `deps/mdnsd` stuff.

jpfr:
The position of using the official repository for the complete nodesets and also for the companion specs is clear. And it's a good idea!
How about adding cmake flags to include companion specs from the foundation repo in the build? Like UA_ENABLE_DI or UA_ENABLE_MTCONNECT.

Anyway, we will keep the Nodeset2.minimal.xml and some other schema files in our repo.
So it's just a little step to include the types.bsd as well and make the live of users who download the source zip a little less complicated.
There are many users who are not familiar with git.
Or who don't read the docs before starting. 😄

Pro:
Ok, got your point. It is now possible to compile the minimal NS0 without the submodule (I just copied the necessary files).

We can add additional CMake flags for the different node sets. Should we support all the ones which are in the UA-Nodeset repo?
- ADI
- DI
- FDI
- GDS
- MTConnect
- PLCopen
- Sercos

Additionally, if e.g. `UA_ENABLE_NS_DI` is ON, do we automatically add the Namespace to the server, i.e., add all the methods into `ua_server_ns0.c` with ifdef, or do we just generate the files and the developer has to handle the adding by calling the corresponding method?

jpfr:
Maybe we can make a "plugin" that contains the method signature for the different information models.

```c
UA_StatusCode UA_Server_addNodeset_DI(UA_Server *server);
UA_StatusCode UA_Server_addNodeset_PLCOpen(UA_Server *server);
```

That way, users who get a .dll or (RedHat/Debian) .so can opt into the companion specs without recompiling? Would that work for you?
We could then have a single CMAKE flag for building all of those.

To me, this PR can be merged as-is whenever you are ready. 👍 
So we can have the companion spec discussion independent from the nodeset compiler improvements.

HolgerJeromin:
@pro 
> This makes it much easier to keep the schema definition in sync with the official reference. 

Travis could check for difference in a local copy and the origin.

PR #1247, opened 2017-10-13
same as #1237 for master ?

StalderT:


jpfr:
Yep. Thanks.
We will try and `merge 0.2` into master over the coming days.
I'll merge this ahead of time when CI becomes all green.

PR #1248, opened 2017-10-13
Fix bugs preventing encryption from working correctly

Infinity95:
This PR fixes a few issues that are only present when using actual encryption.

### DIFF
--- a/plugins/ua_securitypolicy_none.c
+++ b/plugins/ua_securitypolicy_none.c
@@ -80,7 +80,7 @@ newContext_none(const UA_SecurityPolicy *securityPolicy,
 }
 
 static void
-deleteContext_none(void *const channelContext) {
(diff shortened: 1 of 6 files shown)


Infinity95:
Since this is a blocker for the first SecurityPolicy PR i'd like to merge this soon. I'm working on fixing the bugs that are blocking this PR (mainly #1269)

jpfr:
I added the fix in a separate PR #1280 so we get green integration tests for this ASAP.
@Infinity95 Let's talk later today on what is necessary to get this in...

jpfr:
The async client faults are gone for good!
Thanks for finding the issue with the thread-local variables. That was pure gold. 😄 
It was indeed fixed on 0.2 but not yet ported to master.

Infinity95:
Although it's strange, that the problem didn't appear earlier.

PR #1252, opened 2017-10-16
Fix: Ignore good timeout

ccvca:
Hi,
I just fix two bugs:

- UA_Client_runAsync return UA_GOOD even when connection->send failed
- __UA_Client_AsyncService close the connection when a timeout occure (in UA_Connection_receiveChunksBlocking)

ccvca


ccvca:
Failed Travis-Check seems like another error: #1269 
It occures also in other pull requests

ccvca:
One error was fixed by another commit, I force pushed the 2nd error so there are no merge conflicts.

- __UA_Client_AsyncService close the connection when a timeout occure (in UA_Connection_receiveChunksBlocking)

jpfr:
Looks good!
Thanks for the contribution.

Issue #1258, opened 2017-10-17
Including private plugin header file in sources


jonasgreen88:
Hi,

In ua_server_discovery.c a private header file of the plugin implementation is included. When I integrate the stack to my project I have developed my own config plugin and it is not named according to the example in this repository, and I don't think it has to  be.
https://github.com/open62541/open62541/blob/a894e31e415bf469f69ff2f3b26da32debbffd59/src/server/ua_server_discovery.c#L7

In my opinion the source files of the stack should only include private header files of the source and public header file of the include/ dir. My proposal is to add a callback function that is declared in a public header file of the include/ dir and implemented in the config plugin by the user. In ua_server_discovery.c this function is called to retrieve the client config when needed. What do you guys think?

Pro:
Thanks, this would definitely be an improvement.
Maybe you can create a PR and we can see how much it changes?

jonasgreen88:
Ok, I can do that. I just wanted to bring up the issue first to see if you agree to the change. I'll look into it during the coming days.

Pro:
Closing in favor of #1264

PR #1264, opened 2017-10-18
Added callback function to get client config from config plugin

jonasgreen88:
The server implementation needs to initialize a client when it
shall register itself to a discovery server. To avoid including a
private header file of the plugin, a callback function has been
declared in the server header file to be implemented by the plugin
implementation. This callback function are called by the server
when it shall initialize a client.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,6 +349,7 @@ set(exported_headers ${PROJECT_BINARY_DIR}/src_generated/ua_config.h
                      ${PROJECT_SOURCE_DIR}/include/ua_plugin_securitypolicy.h
                      ${PROJECT_SOURCE_DIR}/include/ua_plugin_nodestore.h
                      ${PROJECT_SOURCE_DIR}/include/ua_server_config.h
+                     ${PROJECT_SOURCE_DIR}/include/ua_client_config.h
                      ${PROJECT_SOURCE_DIR}/include/ua_client.h
                      ${PROJECT_SOURCE_DIR}/include/ua_client_highlevel.h
                      ${PROJECT_SOURCE_DIR}/include/ua_client_subscriptions.h)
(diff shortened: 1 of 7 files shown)


jonasgreen88:
Because ua_client.h is included in ua_server.h some include loop is created which causes the PR fail building. Any ideas how to solve that?

jonasgreen88:
#1258

jpfr:
The include loop should be solved by #ifdef include guards.
My guess is the problem stems from the amalgamation where we squash all files together.

Maybe we can have a file order for the amalgamation in CMakeLists.txt like that:
- ua_server.h
- ua_client.h
- ua_server_config.h

Otherwise, we might have to work with forward declarations of structures.

Best regards,
Jlius

Pro:
In this case I would go for a forward declaration of UA_ClientConfig in `ua_server.h` and remove `#include "ua_client.h"` as it is already done for UA_ServerConfig.

Like this we do not include all the client dependencies in the server.

@jonasgreen88 thanks for the PR, nice idea!

jonasgreen88:
@Pro @jpfr Before I uploaded the PR I tried to replace the include of ua_client.h with a forward declaration like this: 
```
struct UA_ClientConfig;
typedef struct UA_ClientConfig UA_ClientConfig;
```
But the problem I faced then was that the UA_ClientConfig was already type defined in ua_client.h. GCC does not allow a typedef to occur multiple times.

Pro:
Isn't it enough to just use `struct UA_ClientConfig;` without the typedef in the forward decl? Not sure though...

jonasgreen88:
If you don't typedef it I think you have to write `struct UA_ClientConfig UA_Server_getClientConfig(void)`. I'm not sure how the compiler would treat `UA_ClientConfig clientConfig = UA_Server_getClientConfig();` in that case.

jpfr:
Would it be better to invent a (samll) header file `ua_client_config.h`?
Then the client and server can depend on that.
And it would be symmetric to `ua_server_config.h`.

jonasgreen88:
> Would it be better to invent a (samll) header file ua_client_config.h?
Then the client and server can depend on that.
And it would be symmetric to ua_server_config.h.

I tried to add a ua_client_config.h and move the definition of UA_Struct to this file but kept the typedef in client.h. But I couldn't get rid of the include of ua_plugin_network.h which is needed as UA_Connection is defined there. Ua_plugin_network.h then includes ua_server.h and the include loop is created...

@jpfr Do you have an idea how to solve it?

Pro:
@jonasgreen88 I think I found a solution for this, let's see if CI is happy

PR #1277, opened 2017-10-26
Added more allowed status codes in Operation_CreateMonitoredItem

jonasgreen88:
According to UA Specification Part 4, chapter 5.12.2.1, the add
operation for the item shall succeed for the following statues:
* Denied read access (Bad_NotReadable)
* No access rights (Bad_UserAccessDenied)
* The index range does not deliver any data for the current value
  (Bad_IndexRangeNoData)


jonasgreen88:
This issue was found when executing the test case "Subscription Services - Subscription Basic - Err-030.js". The test case tries to setup a MonitoredItem with an IndexRange outside of the current value of the variable and expects to succeed doing that.

jpfr:
Merged. Thanks for finding and fixing.

PR #1279, opened 2017-10-27
WIP: Fix async client

jpfr:


Infinity95:
This fixes the unit test, but there's still problems present, when enabling multithreading. Those will still have to be fixed.

PR #1281, opened 2017-10-27
AccessLevel attribute check for reading/writing variable nodes

jonasgreen88:
When reading the description of the Bad_NotWritable status code
it says: "The access level does not allow writing to the Node" while
the description of the Bad_UserAccessDenied says:
"User does not have permission to perform the requested operation.".
Different status codes are used to explicitly tell if the node
doesn't accept the request or if the user is not allowed to
execute the request.


jpfr:

PR #1282, opened 2017-10-27
WIP: SecurityPolicy#Basic128Rsa15

Infinity95:
This PR implements the Basic128Rsa15 security policy using mbedtls as crypto library.
Currently there are two issues i'm having problems with, if someone would like to help out.
1. The cmake configuration isn't quite right yet and i'm having trouble setting up mbedtls as a submodule correctly (see the INSTALL target errors)
2. The CTT is mostly green (The security policy stuff is all green) ~~but i cannot connect with UAExpert, which gives me a BadConfigurationError. I haven't found out what's causing this yet~~.

There might be some noise in this PR that can be removed. ~~The commit fixing bugs can be basically removed with a rebase as soon as the fix/encryption branch is merged into master.~~

The server example currently contains specific hardcoded stuff for the CTT which could be moved into another example possibly.
Also, another example server could be created demonstrating the Encryption, so that the current examples without encryption still work and don't have to be modified.

Depends on #1248

After this PR is finished the following still needs to be done:

- Improve certificate handling (maybe plugin?)
- Unit tests for security policies and the plugin interface (#1336)
- Clean up and repair client side so clients can use encryption

It would also be interesting to fuzz the server with encryption enabled.

TODO before merging:
- Make CTT green for everything
- Modify CMake stuff to apply a patch to the CMakeLists.txt that fixes the install target problem
- Discuss the api of the plugin (see code review)


Pro:
Nice! Thanks!
Can we have an additional flag like `UA_ENABLE_BASIC128RSA15`. If set to off everything encryption related will be excluded from build?

Infinity95:
Sure. Do you know how i need to configure cmake so the install target stuff disappears? I've been sifting through docs and stackoverflow and nothing seems to work.

plugins/ua_securitypolicy_basic128rsa15.c
```diff
@@ -0,0 +1,1299 @@
...
+
+////////////////////////////////
+//// Begin AsymmetricModule ////
+////////////////////////////////
+
+static UA_StatusCode
+sha1Hash(const UA_ByteString *const message,
+         UA_ByteString *const hash) {
+    int mbedErr = 0;
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
```
jpfr: retval is not required. Just return statuscode-good...

plugins/ua_securitypolicy_basic128rsa15.c
```diff
@@ -0,0 +1,1299 @@
...
+    if(retval == MBEDTLS_ERR_MD_ALLOC_FAILED)
+        return UA_STATUSCODE_BADOUTOFMEMORY;
+
+    retval = mbedtls_md_hmac_starts(&context, key->data, key->length);
+    retval = mbedtls_md_hmac_update(&context, in->data, in->length);
+    retval = mbedtls_md_hmac_finish(&context, out->data);
+    mbedtls_md_free(&context);
+
+    if(retval != 0)
+        return UA_STATUSCODE_BADINTERNALERROR;
```
jpfr: badescuritycheksfailed?

plugins/ua_securitypolicy_basic128rsa15.c
```diff
@@ -0,0 +1,1299 @@
...
+        retval |= hmacSha1(secret, &A_and_seed, &outSegment);
+        retval |= hmacSha1(secret, &A, &ANext);
+
+        if(retval != UA_STATUSCODE_GOOD) {
+            if(bufferAllocated)
+                UA_ByteString_deleteMembers(&outSegment);
+            UA_ByteString_deleteMembers(&A_and_seed);
+            UA_ByteString_deleteMembers(&ANext_and_seed);
+            return retval;
+        } else {
```
jpfr: No else required.
The if branch always returns.

plugins/ua_securitypolicy_basic128rsa15.c
```diff
@@ -0,0 +1,1299 @@
...
+}
+
+static UA_StatusCode
+sym_generateNonce_sp_basic128rsa15(const UA_SecurityPolicy *securityPolicy,
+                                   UA_ByteString *out) {
+
+    if(securityPolicy == NULL || securityPolicy->policyContext == NULL || out == NULL)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    if(mbedtls_ctr_drbg_random(&((UA_SP_basic128rsa15_PolicyContextData *) securityPolicy->policyContext)->drbgContext,
```
jpfr: There's a lot going on here...
Difficult to read.

You could separate
- setting a pointer variable to the context
- fill the bytestring with randomness, set a retval
- check if retval == 0

plugins/ua_securitypolicy_basic128rsa15.c
```diff
@@ -0,0 +1,1299 @@
...
+    return retval;
+}
+
+static UA_StatusCode
+channelContext_setLocalSymEncryptingKey_sp_basic128rsa15(void *contextData,
+                                                         const UA_ByteString *key) {
+    if(key == NULL || contextData == NULL) {
+        fprintf(stderr,
+                "Error while calling channelContext_setLocalEncryptingKey_sp_basic128rsa15."
+                    "Null pointer passed.");
```
jpfr: Use logging?

plugins/ua_securitypolicy_basic128rsa15.c
```diff
@@ -0,0 +1,1299 @@
...
+
+    return UA_STATUSCODE_GOOD;
+}
+
+static UA_StatusCode
+policyContext_newContext_sp_basic128rsa15(UA_SecurityPolicy *const securityPolicy,
+                                          const UA_ByteString localCertificate,
+                                          const UA_ByteString localPrivateKey,
+                                          const UA_ByteString *const certificateTrustList,
+                                          const size_t certificateTrustListCount,
```
jpfr: Other parts of the sdk will want access to the trust list as well. Notably session handling.
Pull out into a dedicated plugin.
How would the security policy interact with a key infrastructure plugin?

plugins/ua_securitypolicy_basic128rsa15.h
```diff
@@ -0,0 +1,58 @@
...
+#include "ua_plugin_securitypolicy.h"
+#include "ua_plugin_log.h"
+
+typedef int(*UA_EntropySourceFunc_sp_basic128rsa15)(void *data, unsigned char *output,
+                                                    size_t len, size_t *olen);
+
+/** TODO: document
+ * For more information see mbedtls documentation on mbedtls entropy
+ */
+typedef struct {
```
jpfr: These seem like private definitions. Put on top of the .c file.
And leave just a skeleton API for initialization..

src/ua_connection.c
```diff
@@ -9,6 +9,7 @@
 #include "ua_types_generated_handling.h"
 #include "ua_transport_generated_encoding_binary.h"
 #include "ua_securechannel.h"
+#include <stdio.h>
```
jpfr: No.

plugins/ua_securitypolicy_basic128rsa15.h
Infinity95: I don't quite get what you mean. Could you elaborate?

plugins/ua_securitypolicy_basic128rsa15.c
Infinity95: Yeah. I forgot to change this. In an older version of the API the logger wasn't available here.

plugins/ua_securitypolicy_basic128rsa15.h
jpfr: Users of the plugin don't need to know the structures and the method signature.
So hide it from them.
The definitione can live in the .c file. No need to keep them in the public header.

Infinity95: This was intended to be supplied by the user. The user has the possibility of supplying entropy sources and functions to mbedtls. Since this differs from platform to platform it cannot be implemented in a plugin. The user can also choose to not supply any, in which case he just passes NULL and no entropy functions are initialized to mbedtls.

jpfr: Simplify, simplify, simplify.

I like the idea to empower users add entropy sources. I just fear the we open a huge can of worms where users have doubts about how secure the default settings are. I think we can cover 95% of use cases with our default implementation. And the remaining users are on special platforms where they fork+modify the plugins anyway.

I found the following in the mbedTLS docs: https://tls.mbed.org/kb/how-to/how-do-i-port-mbed-tls-to-a-new-environment-OS

- On Windows and different Unix platforms that provide /dev/urandom, a default OS-based source is registered. It can be disabled at compile-time without affecting the rest of the library.
- For obvious security reasons, the entropy module will refuse to output anything until a declared-strong source has been registered.

That sounds like we can have a fool-proof API.
- Use the default strong entropy sources on Windows/Unix-like.
- On all other platforms emit a compiler #warning + refuse to create entropy + state the missing entropy source in log output.
- Document the place where entropy sources shall be added inside the plugin, not in the plugin header. Then we can use the definitions from mbedtls and don't have to wrap everything.

This is a sensitive issue when it comes to security.
While I'm always on the crusade for small APIs, can you pull Christian Haas into the loop?
He knows this stuff a lot better.

plugins/ua_securitypolicy_basic128rsa15.c
Infinity95: What i had in mind was a plugin that can be queried for certificates and certificate validation.
The security policy can still choose to store a copy of the certificate in a parsed form or whatever, since it will only affect the implementation .c file. The plugin would be initialized once by the user with the certificates (or maybe their location, if we store stuff file-based) and then passed to the server and security policies.
Each security policy would then have a pointer to the plugin with which it can access the certificates and verify them via the plugin API.

Another thing to consider is where to store/load the certificates (and also the configuration, which is currently hard coded). This could maybe be handled by a plugin as well. (file system might not always be an option, right?)

Infinity95: Although i would like to put such a plugin into a separate PR. Either before or after this. It shouldn't be much of a problem to change the current implementation to work with a plugin for certificates.

Infinity95:
Ok i found a way to get rid of the install error but had to modify the mbedtls cmake for it to work:
`install(TARGETS ${mbedtls_static_target} ${mbedx509_static_target} ${mbedcrypto_static_target}
            DESTINATION ${LIB_INSTALL_DIR}
            PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)`
-->
`install(TARGETS ${mbedtls_static_target} ${mbedx509_static_target} ${mbedcrypto_static_target}
            DESTINATION ${LIB_INSTALL_DIR}
            PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
            EXPORT open62541Targets)`

I would like to avoid this. @Pro do you know a way around this?

jpfr:
Hey @Infinity95 ,

can you review / green-light the first three commits of #1293?
They concern the securechannel.

Thanks, Julius

jpfr:
Rebased on master.
Not building on my box due to manual steps required for mbedtls.

Can we include a CMake snippet to use the system-wide mbedtls library if available?
Here's somebody going that way:
https://github.com/hugsy/proxenet/blob/67fc6b552de36e13dee31cf17fa267ee08428b5c/CMakeLists.txt#L130

Infinity95:
Pleas don't be fooled if travis is green. I still need to enable the security policy for CI so that it is built as well.

jpfr:
Hey @Infinity95 

I made some changes where all certificates are accepted if no trust-list has been set.
Now I get this error: 
```
[2017-12-19 23:19:31.852 (UTC+0100)] warn/userland      No CA trust-list provided. Any remote certificate will be accepted.
[2017-12-19 23:19:31.854 (UTC+0100)] info/network       TCP network layer listening on opc.tcp://virtualbox:4840/
[2017-12-19 23:19:34.856 (UTC+0100)] info/network       Connection 5 | New connection over TCP from 192.168.56.1
[2017-12-19 23:19:34.857 (UTC+0100)] info/channel       Creating a new SecureChannel
[2017-12-19 23:19:34.858 (UTC+0100)] info/channel       Connection 5 | SecureChannel 1 | Opened SecureChannel
[2017-12-19 23:19:34.860 (UTC+0100)] info/channel       Connection 5 | SecureChannel 1 | CloseSecureChannel
[2017-12-19 23:19:34.860 (UTC+0100)] info/network       Connection 5 | Closed by the server
[2017-12-19 23:19:37.942 (UTC+0100)] info/network       Connection 5 | New connection over TCP from 192.168.56.1
[2017-12-19 23:19:37.975 (UTC+0100)] info/channel       Creating a new SecureChannel
[2017-12-19 23:19:37.977 (UTC+0100)] warn/securitypolicy        No CA trust-list defined. Accept remote certificate by default.
[2017-12-19 23:19:37.983 (UTC+0100)] warn/securitypolicy        Could not decrypt the message with error: RSA - Input data contains invalid padding and is rejected
[2017-12-19 23:19:37.984 (UTC+0100)] info/network       Connection 5 | Processing the message failed with error BadSecurityChecksFailed
[2017-12-19 23:19:37.985 (UTC+0100)] info/network       Connection 5 | Closed by the server
[2017-12-19 23:19:41.855 (UTC+0100)] info/channel       Connection 0 | SecureChannel 0 | SecureChannel has timed out
^C[2017-12-19 23:19:42.897 (UTC+0100)] info/userland    received ctrl-c
[2017-12-19 23:19:42.898 (UTC+0100)] warn/network       Socket select failed with Interrupted system call
[2017-12-19 23:19:42.899 (UTC+0100)] info/network       Shutting down the TCP network layer
```

I think my private key was generated with the wrong settings. Is this the right invokation?
`openssl genpkey -algorithm RSA -out private_key.pem -pkeyopt rsa_keygen_bits:2048`
Different key-lengths didn't help.

Infinity95:
Since usually the client message is the first one to be decrypted something on that side must be configured incorrectly. What client are you using?
Did you create a matching certificate to the private key and are you using that?

jpfr:
UAExpert.

Infinity95:
What keylength did you configure for the UAExpert certificate? Maybe this is set to something bigger than 2048 bits. If that's not the cause i'll try reproducing with your keygen command.

Infinity95:
Are you sure that you are using a matching private/public key pair? Because if i use a private key and a certificate that belongs to a different key, i get exactly this error.

Infinity95:
Why was the call to `mbedtls_ctr_drbg_seed` removed? Wee need to seed the rng.
I also find the name ua_accesscontrol_certificate misleading, since accesscontrol is possible with certificates but that's not what this plugin does. Wouldn't ua_certificate_manager or something be better? (just anything different from ua_accesscontrol_certificate)

jpfr:
> I also find the name ua_accesscontrol_certificate misleading, since accesscontrol is possible with certificates but that's not what this plugin does.

Point taken. We want to use the plugin for client and server.
For server-only, accepting a certificate would be part of access control.
Either for user credentials or for acting as a gatekeeper at the SecureChannel level.

> Why was the call to mbedtls_ctr_drbg_seed removed? Wee need to seed the rng.

Readded. We now use by default the platform-wide entropy source. Whereas no entropy source was added before. (One had to use a special API, but the example code didn't use that facility).

Infinity95:
The encryption stuff was in the server example on purpose so that it can be run with the CTT. The policy example doesn't suffice for that since the nodeset is empty.
I still haven't found out why the CTT crashes though, so I might make a new project that is dedicated to the CTT maybe. We could add a CTT specific server to the CTT-repository maybe?

jpfr:
Right. We could rename the current server to `server_ctt`.
Then we can stuff any complexity into it. And it won't be the first thing users look at.

jpfr:
I still have difficulties connecting with UAExpert.
The steps are as follows:

- Build with UA_ENABLE_ENCRYPTION and the examples enabled
- Create server certificate and private key with the tool in `/tools/certs/create_self-signed.py`.
- Run the server as `./bin/examples/server_basic128rsa15 ./certs/server_cert.der ./certs/server_private.der`
- Connect with UAExpert.

The certificate uses a 1024 byte key length. Changing this to 2048 did not help.
Can we adjust `create_self-signed.py` so that the generate certificate and key work out of the box?

Then we only have to add this description to the docs and we are good. 😄

Infinity95:
I can change the script next week or so.
What is the error message you get?
You can leave the key size at 2048 bit. Basic128Rsa15 supports keys between 1024 and 2048 bit inclusive.
As far as I remember I used the python script in the repository and just uncommented the same lines you uncommented. Use localhost.key (convert to der first) and the generated server_cert.der
I currently cannot confirm if this works. Can you try?
The command you added converts the certificate authority private key as far as i can tell, which is not what we need.

jpfr:
Ok, cool. I reverted the script to 2048bit keys.

Script:
```
 python ../tools/certs/create_self-signed.py ./certs/
```

Server and output:
```
jpfr@virtualbox:~/software/open62541/build$ ./bin/examples/server_basic128rsa15 ./certs/server_cert.der ./certs/server_private.der
[2018-01-01 17:46:53.321 (UTC+0100)] warn/userland      No CA trust-list provided. Any remote certificate will be accepted.
[2018-01-01 17:46:53.324 (UTC+0100)] info/network       TCP network layer listening on opc.tcp://virtualbox:4840/
[2018-01-01 17:46:58.947 (UTC+0100)] info/network       Connection 5 | New connection over TCP from 192.168.56.1
[2018-01-01 17:46:58.947 (UTC+0100)] info/channel       Creating a new SecureChannel
[2018-01-01 17:46:58.948 (UTC+0100)] info/channel       Connection 5 | SecureChannel 1 | Opened SecureChannel
[2018-01-01 17:46:58.949 (UTC+0100)] info/channel       Connection 5 | SecureChannel 1 | CloseSecureChannel
[2018-01-01 17:46:58.949 (UTC+0100)] info/network       Connection 5 | Closed by the server
[2018-01-01 17:47:01.777 (UTC+0100)] info/network       Connection 5 | New connection over TCP from 192.168.56.1
[2018-01-01 17:47:01.804 (UTC+0100)] info/channel       Creating a new SecureChannel
[2018-01-01 17:47:01.808 (UTC+0100)] warn/securitypolicy        mbedTLS returned an error: RSA - Input data contains invalid padding and is rejected
[2018-01-01 17:47:01.808 (UTC+0100)] info/network       Connection 5 | Processing the message failed with error BadSecurityChecksFailed
[2018-01-01 17:47:01.808 (UTC+0100)] info/network       Connection 5 | Closed by the server
[2018-01-01 17:47:03.325 (UTC+0100)] info/channel       Connection 0 | SecureChannel 0 | SecureChannel has timed out
```

Infinity95:
Are you using the localhost.key keyfile? (convert to der first)
I changed the script so no confusion should arise what key belongs to which certificate.

Infinity95:
@jpfr Can you check if the script works for you? It is also possible to supply the desired key size as a command line argument now.

jpfr:
These are mostly cosmetic changes.
They are welcome. But the issue remains.

I attach a Wireshark log. OpenSecureChannel with signing / no encryption.

[secure_conversation.zip](https://github.com/open62541/open62541/files/1634862/secure_conversation.zip)

```
[2018-01-16 03:14:19.790 (UTC+0100)] warn/userland      No CA trust-list provided. Any remote certificate will be accepted.
[2018-01-16 03:14:20.927 (UTC+0100)] info/network       TCP network layer listening on opc.tcp://virtualbox:4840/
[2018-01-16 03:14:21.413 (UTC+0100)] info/network       Connection 5 | New connection over TCP from 192.168.56.1
[2018-01-16 03:14:21.430 (UTC+0100)] info/channel       Creating a new SecureChannel
[2018-01-16 03:14:21.451 (UTC+0100)] info/channel       Connection 5 | SecureChannel 1 | Opened SecureChannel
[2018-01-16 03:14:21.485 (UTC+0100)] info/channel       Connection 5 | SecureChannel 1 | CloseSecureChannel
[2018-01-16 03:14:21.493 (UTC+0100)] info/network       Connection 5 | Closed by the server
[2018-01-16 03:14:23.900 (UTC+0100)] info/network       Connection 5 | New connection over TCP from 192.168.56.1
[2018-01-16 03:14:23.943 (UTC+0100)] info/channel       Creating a new SecureChannel
[2018-01-16 03:14:24.040 (UTC+0100)] warn/securitypolicy        mbedTLS returned an error: RSA - Input data contains invalid padding and is rejected
[2018-01-16 03:14:24.042 (UTC+0100)] info/network       Connection 5 | Processing the message failed with error BadSecurityChecksFailed
[2018-01-16 03:14:24.047 (UTC+0100)] info/network       Connection 5 | Closed by the server
```

Also, there was a memleak.

```
==4689== 512 bytes in 1 blocks are definitely lost in loss record 1 of 1
==4689==    at 0x4C2ABEF: malloc (vg_replace_malloc.c:299)
==4689==    by 0x113D48: UA_ByteString_allocBuffer (ua_types.c:187)
==4689==    by 0x15720D: asym_decrypt_sp_basic128rsa15 (ua_securitypolicy_basic128rsa15.c:209)
==4689==    by 0x160810: decryptChunk (ua_securechannel.c:799)
==4689==    by 0x160E63: UA_SecureChannel_processChunk (ua_securechannel.c:1019)
==4689==    by 0x174C42: processCompleteChunkWithoutChannel (ua_server_binary.c:688)
==4689==    by 0x174CCA: processCompleteChunk (ua_server_binary.c:714)
==4689==    by 0x15E274: processChunk (ua_connection.c:135)
==4689==    by 0x15E318: UA_Connection_processChunks (ua_connection.c:160)
==4689==    by 0x174D4F: processBinaryMessage (ua_server_binary.c:729)
==4689==    by 0x174E0A: UA_Server_processBinaryMessage (ua_server_binary.c:749)
==4689==    by 0x16FE1D: ServerNetworkLayerTCP_listen (ua_network_tcp.c:605)
```

Infinity95:
Well, you force pushed away my changes to the script. I'll fetch them from my laptop and repush them.
I'll check the logs as well. I tested it with the modified script with different key lengths on different machines and it works for me.

Infinity95:
Well the wireshark log alone doesn't help much, since OPN messages are always encrypted. It would really help if i had the keys you are using.

jpfr:
[keys.zip](https://github.com/open62541/open62541/files/1635047/keys.zip)

jpfr:
This example OPN message + key pair could also be transformed into a unit test to prevent regressions.

Infinity95:
The private key doesn't belong to the public key.
You can check with
`vimdiff <(openssl x509 -noout -modulus -inform DER -in server_cert.der | openssl md5) <(openssl rsa -noout -modulus -inform DER -in server_private.der | openssl md5)` for example

jpfr:
Oh. Didn't see that the last commit changed from `server_private.der` to `server_key.der`.
Got it to work!!
Now let's merge this beast in.

Infinity95:
Shouldn't we enable the encryption stuff on travis and appveyor (or at least travis, cause installing mbedtls dependency on windows might be a bit more complicated)?

jpfr:
There is now a package manager for Windows with official blessing.
https://blogs.msdn.microsoft.com/vcblog/2016/09/19/vcpkg-a-tool-to-acquire-and-build-c-open-source-libraries-on-windows/

jpfr:
Apparently based on cmake.
And they have a port for mbedtls.
https://github.com/Microsoft/vcpkg/blob/master/ports/mbedtls/portfile.cmake

Microsoft +5 points in my book. 😄

Infinity95:
Nice. I'm a bit disappointed that they don't seem to have vulkan in there though.

jpfr:
Bada Boom!

Infinity95:
Not waiting for travis thoug? :sweat_smile:

Stasik0:
Basic is only non-repudition, correct?

Infinity95:
Basic is encryption as well, but it is declared as deprecated since standard version 1.04. ("weak" algorithms)

vogsphar:
Great to see that Basic128Rsa15  is working.
What is the status about Basic256Sha256? (The feature table links to this WIP)

Infinity95:
Theres some old code i have that can be ported into a plugin, but i have other stuff to do right now so i can't tell when im going to get to work on this part again.

bobbthebuilder:
Impressive piece of work! 
@Infinity95 can you highlight rougly what needs to be done to get Basic256Sha256 working as well? Can one follow a similar way as you did for Basic128Rsa15 or would it even be possible to adapt major parts? I'm very interested what you would recommend.

Infinity95:
Some parts of Basic128Rsa15 can be reused i believe. But since the algorithms are different those have will have to be changed. It should be possible to implement the policy by just adding a plugin for it and maybe initializing an endpoint in the default config.

Infinity95:
@Stasik0 BTW i just realized symmetric signatures with MACs do not provide non-repuditation capability, only authenticity. I don't believe the standard provides a mechanism to ensrure non-repuditation.

Issue #1284, opened 2017-10-29
Segmentation fault in client in UA_Client_Subscriptions_manuallySendPublishRequest


StalderT:
Description
===========

client crash in UA_Client_Subscriptions_manuallySendPublishRequest

Background Information / Reproduction Steps
===========================================

Modify client.c : 

line 108 :

```
    UA_Client_Subscriptions_manuallySendPublishRequest(client);
```

to

```
    while(1){
        UA_Client_Subscriptions_manuallySendPublishRequest(client);
        sleep(1);
    }
```

start bin/examples/server in a shell
start bin/examples/client in a shell

in server shell press ctrl+z
wait 20 seconds
press fg in shell

client crash. Valgring result : 
```
==8535== Stack overflow in thread #1: can't grow stack to 0xbe710000
==8535== 
==8535== Process terminating with default action of signal 11 (SIGSEGV)
==8535==  Access not within mapped region at address 0xBE710FDC
==8535== Stack overflow in thread #1: can't grow stack to 0xbe710000
==8535==    at 0x804BB45: ??? (in /data/open62541-master/bin/examples/client)
==8535==  If you believe this happened as a result of a stack
==8535==  overflow in your program's main thread (unlikely but
==8535==  possible), you can try to increase the size of the
==8535==  main thread stack using the --main-stacksize= flag.
==8535==  The main thread stack size used in this run was 8388608.
==8535== Stack overflow in thread #1: can't grow stack to 0xbe710000
--8535-- VALGRIND INTERNAL ERROR: Valgrind received a signal 11 (SIGSEGV) - exiting
--8535-- si_code=1;  Faulting address: 0xBE710FDC;  sp: 0x62c15f30

valgrind: the 'impossible' happened:
   Killed by fatal signal

host stacktrace:
==8535==    at 0x58039BAE: ??? (in /usr/lib/valgrind/memcheck-x86-linux)

sched status:
  running_tid=1

Thread 1: status = VgTs_Runnable (lwpid 8535)
Segmentation fault
```

Checklist
=========
Please provide the following information:

 - [X] open62541 Version (release number or git tag): master https://github.com/open62541/open62541/commit/afd1e35a1664901e41f7ef4f624d986739fc27af
 - [X] Other OPC UA SDKs used (client or server): open62541
 - [X] Operating system: linux
 - [X] Critical issue

StalderT:
With UA_LOGLEVEL 100 : 

After fg, the server send in loop : 
```
[...]

PR #1288, opened 2017-10-30
possible fix for #1284

StalderT:


StalderT:
Must we do the same for UA_CLIENTSTATE_SECURECHANNEL?

like : https://github.com/open62541/open62541/compare/master...StalderT:patch-3 ???

PR #1290, opened 2017-10-31
Sample sever's start time in UA_Server_run_startup() 

jonasgreen88:
To allow a device who gets its time knowledge from the network to
initialize the OPC UA server and its address space during startup
but sample a proper Start time to the Server object the start time
is sampled in UA_Server_run_startup().
The device can now initialize the server during startup, go online
on the network and synchronize time and then start the OPC UA server
by calling UA_Server_run_startup() and get a proper sampled start
time in the Server object.


jpfr:
Merged. Thanks!

PR #1293, opened 2017-11-01
PVS issues

jpfr:


src/ua_securechannel.c
```diff
@@ -376,9 +376,8 @@ UA_SecureChannel_sendAsymmetricOPNMessage(UA_SecureChannel *channel, UA_UInt32 r
         }
     }
 
-    /* Encrypt message if mode not none */
-    if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGN ||
-       channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT) {
+    /* Encrypt message if mode is SignAndEncrypt */
+    if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT) {
```
Infinity95: Since symmetric keys for signing need to be exchanged while opening a securechannel, encryption has to be present as well, even if the mode is only sign. This change needs to be reverted.

Specification part 7 6.7.4
The OpenSecureChannel Messages are signed and encrypted if the SecurityMode is not None (even if the SecurityMode is SignOnly).

src/ua_securechannel.c
```diff
@@ -510,25 +507,22 @@ sendChunkSymmetric(UA_ChunkInfo *ci, UA_Byte **buf_pos, const UA_Byte **buf_end)
         signature.length = securityPolicy->symmetricModule.cryptoModule.
             getLocalSignatureSize(securityPolicy, channel->channelContext);
         signature.data = *buf_pos;
-        ci->errorCode = securityPolicy->symmetricModule.cryptoModule.
+        ci->errorCode |= securityPolicy->symmetricModule.cryptoModule.
             sign(securityPolicy, channel->channelContext, &dataToSign, &signature);
-        if(ci->errorCode != UA_STATUSCODE_GOOD) {
```
Infinity95: Why not check right after signing? Should we really try to encrypt a message if the signing process failed?

src/ua_securechannel.c
```diff
@@ -558,7 +552,7 @@ sendChunkSymmetric(UA_ChunkInfo *ci, UA_Byte **buf_pos, const UA_Byte **buf_end)
         if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT)
             *buf_end -= 2;
     }
-    return ci->errorCode;
+    return UA_STATUSCODE_GOOD;
```
Infinity95: Wouldn't returning res be more consistent here, even if it can only be GOOD at the moment? Someone might change something and expect it to return res.

src/ua_securechannel.c
```diff
@@ -610,11 +610,9 @@ UA_SecureChannel_sendSymmetricMessage(UA_SecureChannel *channel, UA_UInt32 reque
 
     /* Encoding failed, release the message */
     if(retval != UA_STATUSCODE_GOOD) {
-        if(!ci.final) {
-            /* the abort message was not sent */
-            ci.errorCode = retval;
+        /* the abort message was not sent */
+        if(!ci.final)
             sendChunkSymmetric(&ci, &buf_start, &buf_end);
```
Infinity95: Shouldn't we replace the message that didn't get encoded correctly with the abort message body before sending something? The status code from sendChunkSymmetric isn't checked here as well. Maybe log it if we don't want to override the other statuscode?

src/ua_securechannel.c
jpfr: Right. Then we can merge the if-condition with the previous one. (This was marked as an issue by PVS Studio.)
I'll include the reference to the spec in a comment.

src/ua_securechannel.c
jpfr: I don't expect signing to fails a lot (if ever).
So we can save a bit in the hot-loop of every received message.
The error code from signing will be caught down the line.

src/ua_securechannel.c
jpfr: I added a TODO with first ideas on error handling.

src/ua_securechannel.c
jpfr: done.

PR #1306, opened 2017-11-03
WIP : remove client-async-services

StalderT:
with https://github.com/open62541/open62541/pull/1299 nobody use client async services
Can we remove client-async-services to simplify client ?


jpfr:
The async client part will become quite significant.
It's the basis to do all the subscription handling the background.

Issue #1308, opened 2017-11-04
stress test causing that the server not sending all monitored items


StalderT:
Description
===========

stress test cause server not sending all monitored items

Background Information / Reproduction Steps
===========================================

When I try to monitor 10'000 items, who always change, in my case, only the ~2'000 last monitored items receive a state change.

Step to reproduce : use [server.txt](https://github.com/open62541/open62541/files/1443326/server.txt) and [client.txt](https://github.com/open62541/open62541/files/1443327/client.txt)

client log : 

```
networkFailure has changed 10000 27686!
networkFailure has changed 9999 27684!
networkFailure has changed 9998 27682!
networkFailure has changed 9997 27680!
networkFailure has changed 9996 27678!
networkFailure has changed 9995 27676!
networkFailure has changed 9994 27674!
networkFailure has changed 9993 27672!
networkFailure has changed 9992 27670!
networkFailure has changed 9991 27668!
networkFailure has changed 9990 27666!
networkFailure has changed 9989 27664!
.
.
.
networkFailure has changed 8200 24086!
networkFailure has changed 8199 24084!
networkFailure has changed 8198 24082!
networkFailure has changed 8197 24080!
networkFailure has changed 8196 24078!
networkFailure has changed 8195 24076!
networkFailure has changed 8194 24074!
networkFailure has changed 8193 24072!
networkFailure has changed 8192 24070!
networkFailure has changed 8191 24068!
networkFailure has changed 10000 27687!
networkFailure has changed 9999 27688!
networkFailure has changed 9998 27689!
networkFailure has changed 9997 27690!
networkFailure has changed 9996 27691!
networkFailure has changed 9995 27692!
networkFailure has changed 9994 27693!
networkFailure has changed 9993 27694!
networkFailure has changed 9992 27695!
networkFailure has changed 9991 27696!
networkFailure has changed 9990 27697!
networkFailure has changed 9989 27698!
networkFailure has changed 9988 27699!
networkFailure has changed 9987 27700!
networkFailure has changed 9986 27701!
networkFailure has changed 9985 27702!
networkFailure has changed 9984 27703!
```

after : 'networkFailure has changed 8191 24068!' we have : ''networkFailure has changed 10000 27687!'

monId < 8191 never arrive in MonitoredItem callback.

Checklist
=========
Please provide the following information:

 - [x] open62541 master Version https://github.com/open62541/open62541/commit/670006bfd43acf9e2d287f2785bf72ed2e73f356
[...]

PR #1311, opened 2017-11-05
save the position of the last monitored item sent

StalderT:
fix https://github.com/open62541/open62541/issues/1308

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -135,6 +135,44 @@ UA_Subscription_removeRetransmissionMessage(UA_Subscription *sub,
     return UA_STATUSCODE_GOOD;
 }
 
+static UA_MonitoredItem *
+selectFirstMonToIterate(UA_Subscription *sub) {
+    UA_MonitoredItem *mon = LIST_FIRST(&sub->monitoredItems);
+    if(sub->lastSendMonitoredItemId > 0) {
+        while(mon) {
+            if(mon->itemId == sub->lastSendMonitoredItemId)
+                break;
+        }
+        if(!mon)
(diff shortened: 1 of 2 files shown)


jpfr:
Good find. Thanks for the PR!

I made some small changes. http://paste.debian.net/994355/
1. Use the monitoreditem id and not a pointer. I'm not comfortable with pointers to monitoreditems that may be removed in between publishes. Even though we do an "innocent" comparison only. Another developer might try to access the "last mon" later on.
2. I split out some of the functionality into smaller methods.

Feel free to incorporate/modify the diff. As usual, you can introduce the changes from a diff file  with `git apply`.

StalderT:
Thanks for the new more readable patch and thanks for the improvement : 
```
+    if(!moreNotifications) {
+        /* All notifications were sent. The next time, just start at the first
+         * monitoreditem. */
+        sub->lastSendMonitoredItemId = 0;
```
Can you add this commit https://github.com/StalderT/open62541/commit/54a04deafe8db2048c84eae6a221b82be54b6b7d ?

StalderT:
I think "sub->lastSendMonitoredItemId = mon->itemId;" should be before TAILQ_FOREACH_SAFE

```
+    MonitoredItem_queuedValue *qv, *qv_tmp;
+    while(mon) {
+        sub->lastSendMonitoredItemId = mon->itemId;
+        TAILQ_FOREACH_SAFE(qv, &mon->queue, listEntry, qv_tmp) {
+            if(*pos >= minsSize)
+                return;
```

If we have a big revisedQueueSize we have a risk to always send values for this MonitoredItem

PR #1313, opened 2017-11-06
Various fuzzer fixes

Pro:
Fixes build error https://github.com/google/oss-fuzz/issues/953

and memory leak:  https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=4020

### DIFF
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -394,7 +394,7 @@ UA_SecureChannel_sendAsymmetricOPNMessage(UA_SecureChannel *channel, UA_UInt32 r
             return retval;
         }
 
-        /* Specification part 6, §6.7.4: The OpenSecureChannel Messages are
+        /* Specification part 6, 6.7.4: The OpenSecureChannel Messages are
          * signed and encrypted if the SecurityMode is not None (even if the
          * SecurityMode is SignOnly). */
         size_t unencrypted_length =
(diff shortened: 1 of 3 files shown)


src/ua_securechannel.c
```diff
@@ -394,7 +394,7 @@ UA_SecureChannel_sendAsymmetricOPNMessage(UA_SecureChannel *channel, UA_UInt32 r
             return retval;
         }
 
-        /* Specification part 6, �6.7.4: The OpenSecureChannel Messages are
+        /* Specification part 6, �6.7.4: The OpenSecureChannel Messages are
```
jpfr: Why replace ASCII with UTF8 here?

Pro: Wasn't on purpose, I'll revert it

src/ua_types_encoding_binary.c
```diff
@@ -885,7 +885,7 @@ UA_findDataTypeByBinary(const UA_NodeId *typeId) {
 /* ExtensionObject */
 static status
 ExtensionObject_encodeBinary(UA_ExtensionObject const *src, const UA_DataType *_) {
-    u8 encoding = src->encoding;
+    u8 encoding = (u8)src->encoding;
```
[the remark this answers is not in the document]
Pro: Yeah, it was a compilation error, looks like oss-fuzz uses even stricter compiler flags

Issue #1316, opened 2017-11-07
can't connect with siemens OPC UA .NET Client


StalderT:
Description
===========

can't connect with siemens OPC UA .NET Client

https://support.industry.siemens.com/cs/document/42014088/programming-an-opc-ua-net-client-with-c-for-the-simatic-net-opc-ua-server?dti=0&lc=en-WW

Background Information / Reproduction Steps
===========================================

compile open62541 with UA_BUILD_SELFSIGNED_CERTIFICATE

start sample server and try to connect with OPC UA .NET Client

![siemens](https://user-images.githubusercontent.com/13799456/32485426-ac8b1684-c3a3-11e7-8bb3-ee6e032c651a.PNG)

I have the error : "Server did not return an EndpointDescription that matched the one used to create the secure channel" (like https://github.com/open62541/open62541/issues/249 ?)

Wireshark log : [siemens.zip](https://github.com/open62541/open62541/files/1449301/siemens.zip)

Here someone arrive to connect with OPC UA .NET Client : https://groups.google.com/forum/#!topic/open62541/7LbGC68DV9k

I do not know if it's a track, but here we have : 

![10](https://user-images.githubusercontent.com/13799456/32485968-88950634-c3a5-11e7-9bf8-3eb523e5fd29.PNG)

and here we have :

![19](https://user-images.githubusercontent.com/13799456/32485972-8c7122d8-c3a5-11e7-820a-3ceea7682517.PNG)

For info I can connect with UaExpert but I have this warning : 

![uaexpert](https://user-images.githubusercontent.com/13799456/32485606-4f58f8d6-c3a4-11e7-8de7-9f068702fd05.PNG)

Checklist
=========
Please provide the following information:

 - [x] open62541 Version : Master
 - [x] Other OPC UA SDKs used (client or server): siemens OPC UA .NET Client
 - [x] Operating system: Linux/Windows
 - [x] Wireshark network dump attached
 - [x] Critical issue

Pro:
I would guess that the problem here is the Application URI and the certificate itself. Otherwise it could also be an issue with the self-signed certificate and the broken chain.

I had a similar problem using Eclipse Milo, where it complained that the Application URI of the Endpoint Description is not within the certificate.
You can try to set the Application URI of the server endpoint to the same which is used as `Subject Name` of the certificate. If that doesn't work (which didn't for me on Eclipse Milo) I had to create a self-signed certificate which includes the application description as a `Subject Alternative Name` (SAN).

[...]

PR #1318, opened 2017-11-07
fix connection problem with siemens OPC UA .NET Client

StalderT:
fix https://github.com/open62541/open62541/issues/1316


Pro:
Nice one, Thanks!
Very interesting that the slash causes siemens OPC UA .NET Client to fail.

jpfr:
Merged. Thanks!

Issue #1233, opened 2017-10-04
SegFault: UA_NodeReferenceKind differs from ReferenceNode DataType


Pro:
Regression of #1168 

`UA_Server_forEachChildNodeCall` uses the following code:

```
UA_Array_copy(parent->references, parent->referencesSize,
        (void**)&refs, &UA_TYPES[UA_TYPES_REFERENCENODE]);
```

https://github.com/open62541/open62541/blob/957beede9151469d4b427f7a387f4abfab910405/src/server/ua_server.c#L63

The ReferenceNode Type is defined in `Opc.Ua.Types.bsd` as:

```
  <opc:StructuredType Name="ReferenceNode" BaseType="ua:ExtensionObject">
    <opc:Documentation>Specifies a reference which belongs to a node.</opc:Documentation>
    <opc:Field Name="ReferenceTypeId" TypeName="ua:NodeId" />
    <opc:Field Name="IsInverse" TypeName="opc:Boolean" />
    <opc:Field Name="TargetId" TypeName="ua:ExpandedNodeId" />
  </opc:StructuredType>
```

but in `ua_plugin_nodestore.h` as:
```
typedef struct {
    UA_NodeId referenceTypeId;
    UA_Boolean isInverse;
    size_t targetIdsSize;
    UA_ExpandedNodeId *targetIds;
} UA_NodeReferenceKind;
```

Note the difference for targetID: for the latter case it is an array whereas the .bsd defines it as a single value.

This means that the UA_Array_copy uses a wrong type definition to copy the references and thus the call to `UA_Server_forEachChildNodeCall` results in a segfault.

@jpfr do we need an array for the targetIds? Can you have a look at this and push a fix?

We should also add a unit test which covers `UA_Server_forEachChildNodeCall`

Pro:
Fixed via #1323

PR #1323, opened 2017-11-07
Fix copy of references. 

Pro:
See #1233

### DIFF
--- a/include/ua_plugin_nodestore.h
+++ b/include/ua_plugin_nodestore.h
@@ -487,6 +487,10 @@ UA_Node_setAttributes(UA_Node *node, const void *attributes,
 UA_StatusCode UA_EXPORT
 UA_Node_copy(const UA_Node *src, UA_Node *dst);
 
+/* Allocate new node and copy the values from src */
+UA_EXPORT UA_Node *
+UA_Node_copy_alloc(const UA_Node *src);
+
 /* Add a single reference to the node */
 UA_StatusCode UA_EXPORT
 UA_Node_addReference(UA_Node *node, const UA_AddReferencesItem *item);
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -57,24 +57,27 @@ UA_Server_forEachChildNodeCall(UA_Server *server, UA_NodeId parentNodeId,
     /* TODO: We need to do an ugly copy of the references array since users may
      * delete references from within the callback. In single-threaded mode this
      * changes the same node we point at here. In multi-threaded mode, this
-     * creates a new copy as nodes are truly immutable. */
-    UA_ReferenceNode *refs = NULL;
-    size_t refssize = parent->referencesSize;
-    UA_StatusCode retval = UA_Array_copy(parent->references, parent->referencesSize,
-        (void**)&refs, &UA_TYPES[UA_TYPES_REFERENCENODE]);
-    if(retval != UA_STATUSCODE_GOOD) {
+     * creates a new copy as nodes are truly immutable.
+     * The callback could remove a node via the regular public API.
+     * This can remove a member of the nodes-array we iterate over...
+     * */
+    UA_Node *parentCopy = UA_Node_copy_alloc(parent);
+    if(!parentCopy) {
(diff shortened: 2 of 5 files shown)


Pro:
I now added a unit test and the corresponding fix.

Ready to be merged when all green.

src/server/ua_server.c
```diff
@@ -58,23 +58,22 @@ UA_Server_forEachChildNodeCall(UA_Server *server, UA_NodeId parentNodeId,
...
-        retval |= callback(ref->targetId.nodeId, ref->isInverse,
+        UA_NodeReferenceKind *ref = &refs[i - 1];
+        for (size_t j = 0; j<ref->targetIdsSize; j++)
+            retval |= callback(ref->targetIds[j].nodeId, ref->isInverse,
                            ref->referenceTypeId, handle);
     }
 
     server->config.nodestore.releaseNode(server->config.nodestore.context, parent);
-    UA_Array_delete(refs, refssize, &UA_TYPES[UA_TYPES_REFERENCENODE]);
+    UA_free(refs);
```
jpfr: You leak memory here.

src/server/ua_server.c
```diff
@@ -58,23 +58,22 @@ UA_Server_forEachChildNodeCall(UA_Server *server, UA_NodeId parentNodeId,
      * delete references from within the callback. In single-threaded mode this
      * changes the same node we point at here. In multi-threaded mode, this
      * creates a new copy as nodes are truly immutable. */
-    UA_ReferenceNode *refs = NULL;
-    size_t refssize = parent->referencesSize;
-    UA_StatusCode retval = UA_Array_copy(parent->references, parent->referencesSize,
-        (void**)&refs, &UA_TYPES[UA_TYPES_REFERENCENODE]);
+    UA_NodeReferenceKind *refs = NULL;
+    UA_StatusCode retval = UA_NodeReferenceKind_copy(parent->references, parent->referencesSize, &refs);
```
jpfr: So the problem happens when the iterate-callback removes references from the node itself?
The code would be simpler if we just made a copy of the entire node. :-)
And then continue to point to the nodeids in-place.

include/ua_plugin_nodestore.h
```diff
@@ -483,6 +483,10 @@ UA_StatusCode UA_EXPORT
 UA_Node_setAttributes(UA_Node *node, const void *attributes,
                       const UA_DataType *attributeType);
 
+/* Copy the references from one array to another */
+UA_StatusCode UA_EXPORT
+UA_NodeReferenceKind_copy(const UA_NodeReferenceKind *src, size_t srcSize, UA_NodeReferenceKind **dst);
```
jpfr: We do pretty much the same in UA_Node_copy.
Can we implement once and reuse?

jpfr: The method name is misleading.
It sounds as if we make a copy of one NodeReferenceKind structure.
But we deep-copy an array of NodeReferenceKind structures.

src/server/ua_server.c
Pro: I just looked again at the code, do we even need this ugly copy?
The callback in line 71 gets only node ids. And we should set the callback parameters to const

jpfr: The callback could remove a node via the regular public API.
This can remove a member of the nodes-array we iterate over...

Pro: Ah, right. Got it, so I will just create a copy of parent and use that copy for iteration

Pro:
@jpfr I change it now to use a copy of the parent node instead copying only the references.

include/ua_plugin_nodestore.h
```diff
@@ -487,6 +487,10 @@ UA_Node_setAttributes(UA_Node *node, const void *attributes,
 UA_StatusCode UA_EXPORT
 UA_Node_copy(const UA_Node *src, UA_Node *dst);
 
+/* Allocate new node and copy the values from src */
+UA_StatusCode UA_EXPORT
+UA_Node_copy_alloc(const UA_Node *src, UA_Node **dst);
```
jpfr: UA_Node_copy_alloc could return a pointer to the new node instead of going with double-pointers.

PR #1327, opened 2017-11-08
fix something missing for #1311

StalderT:
https://github.com/open62541/open62541/pull/1311


StalderT:
yes ```sub->lastSendMonitoredItemId = mon->itemId;``` must be after the loop else we fail one monitoredItem. Sorry !!!

jpfr:
Did you see the problem in practice?

I think the code is fine as-is.
We store the id of the monitoreditem where we had to stop.
Then we resume at this id (not after this id).

StalderT:
do you speak about the second commit of this PR or the two commits?

StalderT:
Because currently we have an infinite loop here:

https://github.com/open62541/open62541/blob/b9c9fc05108d071b9e848c57c6d9fc799d8f7d7c/src/server/ua_subscription.c#L142-L145

PR #1328, opened 2017-11-08
fix #1311

StalderT:
fix https://github.com/open62541/open62541/pull/1311
replace https://github.com/open62541/open62541/pull/1327


jpfr:
Merged. Thanks.

PR #1330, opened 2017-11-09
Better corpus generation for fuzzing

Pro:
This PR adds a c code file which can be used (manually) to generate the corpus for oss-fuzz.

The file `corpus_generator.c` is calling nearly all available client requests at the server. The server then uses `UA_debug_dumpCompleteChunk()` to dump the received package to disk.

After the corpus generator is executed, the files need to be copied from the build directory into the fuzz directory, to avoid any automatic/accidental change of files.

The goal is to reach a better coverage by the fuzzer.

Not yet ready to be merged. First we need to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=4097

@jpfr can you review?


Pro:
Fix before merge: #1331

Issue #1296, opened 2017-11-01
SecureChannel Unit Tests


jpfr:
We currently have no dedicated unit tests for the securechannel.
Without unit tests, we risk regressions and non-conformant behaviour in edge cases.
Especially since the securechannel failure modes interact with the connection, the server, the client, etc.

PR #1295 adds the possibility to manually trigger failure modes via special hooks that are only defined with a compile flag for non-production builds. Together with the custom network layer for testing and the custom clock for testing, we can now reproduce nearly all failure modes in unit tests.

@Infinity95 You talked about unit tests for the encryption layer. This issue is highly related to that.

Infinity95:
I could focus on first writing some unit tests for the SecureChannel before getting to work on the certificate plugin. I guess that would be better to avoid regressions like you said.

jpfr:
#1284 contains a candidate for a connection-related unit test....

Infinity95:
Unit tests for the SecureChannel are now WIP on this branch:https://github.com/Infinity95/open62541/tree/feature/unit-tests/securechannel

jpfr:
Nice!
If you want, you can open a PR marked as WIP in the title.

Infinity95:
Done: #1336

jpfr:
Closing this as the PR is ready.

Infinity95:
I would prefer if we leave this issue open, since that PR at the moment doesn't test all of the SecureChannel. I just want to merge the tests in bit by bit, so we can avoid regression quicker. Currently only the sendAsymmetricOPNMessage function and a few helpers are really tested.
The receiving function and the send Symmetric function haven't been tested yet.

jpfr:
ok.

PR #1336, opened 2017-11-09
Unit tests for some of the SecureChannel module functions

Infinity95:
This PR is aimed at creating unit tests for the SecureChannel module that specifically test just this module.
For that purpose i created a testing SecurityPolicy which can be found in the testing-plugins folder.
The testing policy also contains assertions, that generally check if the functions are called correctly.
Please leave comments with suggestions if you have any.

See also: #1296


Pro:
@Infinity95 you can now rebase on master, the other PR is merged

Infinity95:
Is it possible to disable the errors that currently occur in travis? I find them to be really annoying since i don't want to write an extra assert makro (or even cast every single pointer that is asserted) just so the error disappears. The nullpointer dereference analysis is also a false positive in my opinion, since the pointer that is dereferenced is checked beforehand to not be null.

jpfr:
You can just write
```c
ck_assert(securityPolicy != NULL);
```

Or, if you absolutely have to prevent a warning, add `-Wnoincompatible-pointer-types-discards-qualifiers` to the compiler settings in `/tests/CMakeLists.txt`. The settings here normally do not leak into the parent directory.

In general, we should use conersative settings everywhere. Otherwise we get lazy and the code quality decreases.

Infinity95:
Is there a way to avoid the analyzer warnings without having to add an if(xy != NULL) check to everything? I mean the dereference can never happen, since the assertion kills the program before that, or am i overlooking something?

Infinity95:
Shouldn't we exclude the tests and testing plugins from the coverage (if that's possible)?
They make the coverage seem better than it is i think.

jpfr:
> Shouldn't we exclude the tests and testing plugins from the coverage (if that's possible)?

👍 For more honesty in our numbers.

Infinity95:
How do i exclude them from coverage then? I saw that the examples were excluded, but where is it done?

Infinity95:
See #1375

Infinity95:
@Pro can you help me out with this? It seems something is going wrong with the amalgamation in travis but i can't figure out why or what is causing it. Locally on my machine everything is fine (tried multiple build configurations).

Infinity95:
Ok seems i fixed it. My ide added an include with <> brackets which caused the amalgamation to fail on travis and appveyor, but not locally. I don't know why lol

Infinity95:
@jpfr Can you merge this batch of unit tests after everything is green? I'd like to do these tests in smaller batches, so that we get some measures against regression without having to wait until every single test is done.

jpfr:
Hey @Infinity95 
Can you rebase on master? We can then merge.

jpfr:
Hey @Infinity95 ,

I rebase the branch on master. Can you give me push access to this branch in your repo?

Infinity95:
Done

jpfr:
Thx. Rebased.

I had to disable one unit test.
Some .hex files are missing.
E.g. ${CMAKE_CURRENT_SOURCE_DIR}/dumps/client_HELOPN.hex

jpfr:
@Pro Did you remove ${CMAKE_CURRENT_SOURCE_DIR}/dumps/client_HELOPN.hex?
That is, don't we need that anymore since we autogenerate the corpus?

tests/CMakeLists.txt
```diff
@@ -154,7 +160,56 @@ endif()
 # Readspeed server
 add_executable(check_server_readspeed server/check_server_readspeed.c $<TARGET_OBJECTS:open62541-object> $<TARGET_OBJECTS:open62541-testplugins>)
 target_link_libraries(check_server_readspeed ${LIBS})
-add_test_valgrind(server_readspeed ${TESTS_BINARY_DIR}/check_server_readspeed)
+add_test_valgrind(check_server_readspeed ${TESTS_BINARY_DIR}/check_server_readspeed)
+
+# Test server with network dumps from files
```
Pro: Remove these lines, up to line 212

PR #1337, opened 2017-11-09
Fix #1331 heap use after free

### LINKED ISSUE #1331: Heap-use-after-free in processDelayedCallbacks
This bug was found by oss-fuzz here https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=4097

The problem is the following:

1. Client sends a publishRequest
2. Then in the same TCP package a closeSessionRequest
(so the server won't process delayed callbacks in between)

The publishRequest will be added to the delayed callback by adding `UA_Subscription_answerPublishRequestsNoSubscription` then the closeSessionRequest adds a `removeSessionCallback` at the **head** of delayed callback list.

In `processDelayedCallbacks` the list is a FILO, thus it will then first call `removeSessionCallback` (and free the session memory), then `UA_Subscription_answerPublishRequestsNoSubscription` which still needs the session as a parameter and thus will fail with heap use after free.

I currently see two solutions to fix this, implementing one of these should fix it:
- When calling `UA_Server_delayedCallback` add the new entry at the end of the list
https://github.com/open62541/open62541/blob/302003d2448946fa4d8a02cc209b8931d31e975b/src/server/ua_server_worker.c#L171
- When processing delays start at the end of the list
https://github.com/open62541/open62541/blob/302003d2448946fa4d8a02cc209b8931d31e975b/src/server/ua_server_worker.c#L178

In both cases we would need to change the `server->delayedCallbacks` list to a double-linked-list to still be efficient.

@open62541/open62541-core any thoughts?

Pro:
See #1331

Immediately send UA_Subscription_answerPublishRequestsNoSubscription
before the session is deleted

### DIFF
--- a/src/server/ua_services_subscription.c
+++ b/src/server/ua_services_subscription.c
@@ -516,11 +516,8 @@ Service_DeleteSubscriptions(UA_Server *server, UA_Session *session,
     if(LIST_FIRST(&session->serverSubscriptions))
         return;
 
-    /* Send remaining publish responses in a delayed callback if the last
-     * subscription was removed */
-    UA_Server_delayedCallback(server, (UA_ServerCallback)
-                              UA_Subscription_answerPublishRequestsNoSubscription,
-                              session);
+    /* Send remaining publish responses if the last subscription was removed */
+    UA_Subscription_answerPublishRequestsNoSubscription(server, session);
 }
 
 static void


jpfr:
Perfect if that works.
Does it fly with the CTT?

PR #1350, opened 2017-11-12
fix documentation to provide ccmake

StalderT:


jpfr:
Good idea of making users aware of ccmake.

Using the GUI is optional. Can you add it to the list of additional packages below.
Maybe with a comment on how to invoke the GUI (That's not immediately clear from the install command).

StalderT:
You are absolutely right. Fixed

jpfr:
Merged. Thanks!

Issue #710, opened 2016-07-19
Ideas for API changes


Stasik0:
- [x] Remove liburcu from multithreaded versions
- [x] Attach sessions to method callbacks

- [ ] #705 (Use same struct for UA_Server_addDataSourceVariableNode as for UA_Server_setVariableNode_valueCallback)
-  [ ] UA_Server_addMethodNode needs a possibility to define node ids of "InputArguments" and "OutputArgument" nodes: https://github.com/open62541/open62541/commit/6d23db1acf0bd77416b2bfa68fe8c858f12be5af#diff-3e72cb3692072ede081976a22332e59aR882

HolgerJeromin:
You can change the list to a [task list](https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments). With this the progress would be visible as a progress bar:
`- [ ] `

jpfr:
Change `void * UA_Array_new(size_t size, const UA_DataType *type)` to
```c
UA_StatusCode UA_Array_new(void **outArray, size_t *outSize,
                           UA_Int32 requestedSize, const UA_DataType *type)
```
That lets us handle edge cases, such as NULL arrays, "inside".

jpfr:
The Server_addXXXNode function can be simplified/improved.
For example, addObjectTypeNode might take a `subTypeOf` argument instead of parent and parentReference.

uleon:
Not exactly an API change, but how about getting rid of urcu with 0.3 ?

Stasik0:
@uleon in favor of what?

uleon:
The last time I investigated this issue I had the impression that liburcu is used exclusively for the nodestore. I'd suggest to design a nodestore-plugin concept that made it possible to provide architecture/os/environment specific implementations. In that respect this topic would be in fact an API topic, wouldn't it?

jpfr:
+1 for nodestore plugins as a **new** public facing API

jpfr:
Add a server-wide node creation callback (suggested in https://groups.google.com/forum/#!topic/open62541/wugJCNEu_n8)

jpfr:
Add more limits to the server, e.g. maxsubscriptions, a.s.o.

jpfr:
Forward the session handle (userdata) with the method callback.

Pro:
Related: #1168

ichrispa:
So to summarize:
- [ ] Remove liburcu from multithreaded versions
[...]

PR #1351, opened 2017-11-12
adding samplingInterval to UA_Client_Subscriptions_addMonitoredItem

StalderT:
If you agree, I can modify examples/includes/tests files.


jpfr:
Yes! This is a good addition!
Please align the tests and examples to use the new API.
We can merge then.

@open62541/open62541-core 
We will have a breaking API change for the client monitoreditems.
Is that acceptable?

StalderT:
Ok. What do you want as default value in examples ? 250 like UaExpert (I think) ?

StalderT:
https://github.com/open62541/open62541/issues/710

jpfr:
This looks good!

@open62541/open62541-core 
Okay to squash&merge?

FlorianPalm:
I am good with it!

Pro:
:+1:

PR #1359, opened 2017-11-14
save last seen immediate subscription

### LINKED ISSUE #1356: question about immediate call of UA_Subscription_publishCallback
Description
===========

https://github.com/open62541/open62541/blob/4ff3993542d66e3050beeb7351f339433239944e/src/server/ua_services_subscription.c#L475-L484

In the case where the first subscription in the list is very loaded and the flag UA_SUBSCRIPTIONSTATE_LATE is often/always set to true, is there not a risk that the second subscription of the list is not processed? For "the sake of fairness" between the different subscriptions, should not we save the position of the last subscription processed? like : https://github.com/open62541/open62541/pull/1311/files. Or put at the end of the list the subscription processed ?

What do you think?

Checklist
=========

 - [X] open62541 Version master

StalderT:
fix #1356

### DIFF
--- a/src/server/ua_services_subscription.c
+++ b/src/server/ua_services_subscription.c
@@ -474,14 +474,33 @@ Service_Publish(UA_Server *server, UA_Session *session,
 
     /* Answer immediately to a late subscription */
     UA_Subscription *immediate;
-    LIST_FOREACH(immediate, &session->serverSubscriptions, listEntry) {
-        if(immediate->state == UA_SUBSCRIPTIONSTATE_LATE) {
-            UA_LOG_DEBUG_SESSION(server->config.logger, session, "Subscription %u | "
-                                 "Response on a late subscription", immediate->subscriptionID);
-            UA_Subscription_publishCallback(server, immediate);
-            break;
-        }
+    UA_Boolean found = true; 
+    UA_UInt32 more = 1;
+
+    if (session->lastSeenSubscriptionID > 0){ 
+        found = false;
+        more = 2;
     }
+
+    while (more){
+       LIST_FOREACH(immediate, &session->serverSubscriptions, listEntry) {
+            if (!found){
+                if (session->lastSeenSubscriptionID == immediate->subscriptionID){
+                    found = true; 
+                }     
+            }else{
+                if(immediate->state == UA_SUBSCRIPTIONSTATE_LATE) {
+                    session->lastSeenSubscriptionID = immediate->subscriptionID;
+                    UA_LOG_DEBUG_SESSION(server->config.logger, session, "Subscription %u | "
+                                         "Response on a late subscription", immediate->subscriptionID);
+                    UA_Subscription_publishCallback(server, immediate);
+                    return;
+                }     
+            }     
+        }     
+        more--;
+    }
+    session->lastSeenSubscriptionID = 0;
 }
 
 static void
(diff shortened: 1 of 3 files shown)


StalderT:
Now all the subscriptions are treated fairly.

```
test has changed 9999!
test has changed 4000!
test has changed 9000!
test has changed 3000!
test has changed 8000!
test has changed 2000!
test has changed 7000!
test has changed 1000!
test has changed 6000!
test has changed 0!
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131551315302948140 after 131551315325507020 diff 22558880
test has changed 5000!
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131551315325507050 after 131551315325512400 diff 5350
test has changed 9999!
test has changed 4000!
test has changed 9000!
test has changed 3000!
test has changed 8000!
test has changed 2000!
test has changed 7000!
test has changed 1000!
test has changed 6000!
test has changed 0!
```

jpfr:
This is a welcome addition. We should merge it soon.

Is it possible to not have the `while (more)` loop?
It is very difficult to understand. And it should be possible to find the relevant late subscription in one pass over the linked list.

You can store a pointer to the "best candidate so far". Start with the first subscription and replace it whenever a "better" candidate is found.
- Late subscription
- Late subscription after the last subscription that was sent.

StalderT:
We can remove the ```while (more)``` but we need to make 2 pass, something like this :

```
    UA_Subscription *immediate;

    if (session->lastSeenSubscriptionID > 0){ 
       UA_Boolean found = false;
       LIST_FOREACH(immediate, &session->serverSubscriptions, listEntry) {
            if (!found){
                if (session->lastSeenSubscriptionID == immediate->subscriptionID){
                    found = true;
                }   
            }else{
                session->lastSeenSubscriptionID = immediate->subscriptionID;
                if(immediate->state == UA_SUBSCRIPTIONSTATE_LATE) {
                    UA_LOG_DEBUG_SESSION(server->config.logger, session, "Subscription %u | "
                                         "Response on a late subscription", immediate->subscriptionID);
                    UA_Subscription_publishCallback(server, immediate);
                    return;
                }   
            }   
        }   
    }   

    LIST_FOREACH(immediate, &session->serverSubscriptions, listEntry) {
        session->lastSeenSubscriptionID = immediate->subscriptionID;
        if(immediate->state == UA_SUBSCRIPTIONSTATE_LATE) {
            UA_LOG_DEBUG_SESSION(server->config.logger, session, "Subscription %u | "
                                 "Response on a late subscription", immediate->subscriptionID);
            UA_Subscription_publishCallback(server, immediate);
            return;
        }   
    }   
    session->lastSeenSubscriptionID = 0;
```
If we found anything or if there are LATE in the list before lastSeenSubscriptionID and not LATE after lastSeenSubscriptionID we need a second pass.

see fix https://github.com/StalderT/open62541/commit/cc7e20527db46e26e91790898e60956f973f899c

StalderT:
ok for merging ?

Pro:
IMHO it would be cleaner if you replace the variable `more` with `loopCount` and then use

`for(int i=0; i<loopCount; i++)` instead of `while(more)`

The general concept is:
If you know how often you want to execute a loop (as it is for this case) use for. If you do not know that, use while.

Pro:
And it would also help if there's a two sentence comment explaining why this loop is required. If someone looks at the code in 5 years, he/she will probably not get the idea behind it

StalderT:
Thank you for the constructive review. I will update the PR.

jpfr:
Looking good.
We can squash/merge once Travis is green.

Issue #1304, opened 2017-11-03
Cache service responses if connection is interrupted and resend them on reconnect


mlgiraud:
Description
===========
The server for example should hold on to service responses for a while if the connection was interrupted  for some reason, so the client has the opportunity to reconnect and still recieve them. Currently this doesn't seem to be implemented.
See Spec Part 6 §7.1.6 for more.

teccheck:
With recent releases this feature should be implemented. Feel free to reopen if needed

PR #1370, opened 2017-11-16
fix #1273

### LINKED ISSUE #1273: V0.2 -> V0.3 BadTcpMessageType
Description
===========
Upgraded my client from V0.2 to V0.3, both amalgamated versions. Changed UA_ClientConfig_Standard to UA_ClientConfig_Default.
Getting BadTcpMessageTypeInvalid, varied with succesfull reads of the server time. 
Changing back to open62541 .h/.c 0.2 files and _standard the connection is fine. 

https://pastebin.com/ZBgdQgtS

[v0.3 bad tcp type.pcapng.zip](https://github.com/open62541/open62541/files/1411190/v0.3.bad.tcp.type.pcapng.zip)

Checklist
=========

 - [x] open62541 Version (release number or git tag): 0.2 -> 0.3
 - [x] Operating system: Mac 
 - [ ] Critical issue

StalderT:


jpfr:
Nice one!
This is marked as WIP.

All tests are green. Can we merge?

StalderT:
I need help !!!

Actual PR fix https://github.com/open62541/open62541/issues/1273

But there is another problem! I try to explain !

Without the PR :

--------

Pass 1) The client connect to server and read current time

image

result as expected

------

Pass 2) The client tries to read the current time but can not.

image

result as expected

-------

Pass 3) The client try to reconnect but can not

image

The problem is here !!!

The server assign the old session to the client here : 

https://github.com/open62541/open62541/blob/d7adcf85504f6f15b5cd0e8fea4de9c8dcfe2dd6/src/server/ua_server_binary.c#L431-L432

after the server send a sendServiceFault here : 

https://github.com/open62541/open62541/blob/d7adcf85504f6f15b5cd0e8fea4de9c8dcfe2dd6/src/server/ua_server_binary.c#L485

because : 

https://github.com/open62541/open62541/blob/d7adcf85504f6f15b5cd0e8fea4de9c8dcfe2dd6/src/server/ua_server_binary.c#L481

here the session->channel is NULL

How can we correct the problem ? Should the client reset the securityToken when closing the client? Should the server assign the session to the secure channel?

image

jpfr:
There are two options.

1. The client can recover the old session on a new securechannel. For that, the client does not call `CreateSession` but only `ActivateSession` with the old `UserIdentityToken`.
2. The client can open a new SecureChannel. Then the old `UserIdentityToken` should not be reused.

StalderT:
Actually, when the client send OpenSecureChannelRequest, server respond : 
image

Then the client call GetEndPointRequest but fail because session are not activate on the channel

StalderT:
What do you think of this : https://github.com/StalderT/open62541/commit/75ee6af522d42eb0503e509b453517143c388067 ?

Infinity95:
Take a look at the Specification Part 6 §7.1.6
It is written there, that a connection can be re-established without creating a new SecureChannel.
This is currently not implemented i believe and should solve the problem (i think?)
So if we can re-establish the connection, the old securechannel is used and we don't have any problems. But if this fails, a new securechannel is created and we can try to reattach the old session to the new securechannel.

StalderT:
You are absolutely right. I modified the PR to unblock the situation while waiting for a correct implementation. What do you think ?

#1304

StalderT:
ok to merge?

StalderT:
fix via #1386

PR #1373, opened 2017-11-16
Added support for OperationLimits

heno-hms:
Added support for OperationLimits object, corresponding additions to server config and responding with BadTooManyOperations for relevant services.

Also added support for setting MaxBrowseContinuationPoints and MinSupportedSampleRate of the ServerCapabilities object.

Also added support for limiting the number of references per nodes in the BrowseRequest response.

### DIFF
--- a/include/ua_server_config.h
+++ b/include/ua_server_config.h
@@ -86,7 +86,6 @@ struct UA_ServerConfig {
 
     /* Limits for Requests */
     UA_UInt32 maxReferencesPerNode;
-    UA_UInt16 maxBrowseContinuationPoints;
 
     /* Limits for Subscriptions */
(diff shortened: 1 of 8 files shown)


Pro:
@heno-hms thanks for the addition! I reordered and updated the Nodeset XML. Our goal is to keep the Minimal NS0 in the same order as Full NS0 NodeSet2.xml

Can you add some quick unit tests which cover the limits? This will ensure that the functionality you added will not be broken in the future.

jpfr:
@Pro If the nodeset xml is ready, ok we merge now?

It seems @heno-hms is out of time.
And the PR might bitrot quickly.

We can open an issue for the missing unit tests.

Issue #1332, opened 2017-11-09
UA_Client_Subscriptions_manuallySendPublishRequest fail if revisedPublishingInterval > config.timeout


StalderT:
Description
===========

UA_Client_Subscriptions_manuallySendPublishRequest fail if revisedPublishingInterval > config.timeout

Background Information / Reproduction Steps
===========================================

add ```config->publishingIntervalLimits.min  = 10000;``` to sample examples/server

start examples/client

the second call to UA_Client_Subscriptions_manuallySendPublishRequest https://github.com/open62541/open62541/blob/302003d2448946fa4d8a02cc209b8931d31e975b/examples/client.c#L151 fail with : 

```
[11/09/2017 09:23:41.961] info/client	Error receiving the response with status code BadRequestHeaderInvalid
```

Should the client use "config.timeout + revisedPublishingInterval" instead of config.timeout or should the server sometimes send a keepalive?

Checklist
=========
Please provide the following information:

 - [X] open62541 Version : master

basyskom-jvoe:
According to Part 4, 5.13.1.1 f, the server is required to send a keepalive after the maximum keep alive count has been reached. Depending on the settings, this could be minutes or even hours after the timeout in config.timeout.

Values for PublishingInterval and MaxKeepAliveCount are only limited by the data types used (double and UInt32). If the server permits large values, using revisedPublishingInterval*revisedMaxKeepAliveCount + config.timeout could lead to manuallySendPublishRequest blocking for a very long time in which the client can't be used to read/write/call methods.

StalderT:
closed in favor of https://github.com/open62541/open62541/issues/1485

PR #1374, opened 2017-11-16
fix #1230

### LINKED ISSUE #1230: reconnecting timeout
Description
===========
By removing cable between OPC-UA server and client the data won't be sent. The reconnecting function in the example runs client_connect but stays in this method for about 70 seconds before starting a new connection. 

Using the pause and resume button the connection will be restored immediately after the resume button has been pressed.

[connection timeout.pcapng.zip](https://github.com/open62541/open62541/files/1336315/connection.timeout.pcapng.zip)

Windows 10 virtual machine with B&R simulator which runs the OPC-UA server, also tried with proses simulation server. Same problem.
Mac with open62541 client, example code: client loop connect.

Checklist
=========

 - [x] open62541 Version (release number or git tag): 0.3
 - [x] Other OPC UA SDKs used (client or server): Prosys simulation server, B&R simulator
 - [x] Operating system: Windows 10 and Os X
 - [ ] Critical issue

StalderT:


jpfr:
You want to handle timeouts of up to 1 ms differently?

StalderT:
With this commit https://github.com/open62541/open62541/commit/e5f33bb7b436f748d5891b18688e6c41179b3cec, client never detect connection lost  with server and always return UA_STATUSCODE_GOODNONCRITICALTIMEOUT

StalderT:
timeouts <1 ms = expired ?

StalderT:
Do you agree ? What must we make for UA_Client_Service_publish ? See https://github.com/open62541/open62541/issues/1332

PR #1375, opened 2017-11-17
Ignore test and examples directory for coveralls

Infinity95:
See also: #1336

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 *.pyc
 *.gcno
 *.gcda
+*.swp
 .dirstamp
 /bin
 m4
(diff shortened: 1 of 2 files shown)


Infinity95:
There seems to be a bug with 100% covered files where the relevant lines are shown as 0.
[This](https://stackoverflow.com/questions/37373398/why-does-coveralls-show-0-coverage-when-every-source-file-is-100-covered) Stackoverflow post might be a possible solution?

Infinity95:
I've removed everything now i think. @jpfr / @Pro  can you take a look and merge if all is fine?

PR #1387, opened 2017-11-20
Unit Tests for Windows/Appveyor

Pro:
This PR adds unit tests for appveyor and includes some additional fixes.


Pro:
@jpfr one thing I'm not sure about here is the encoding of `-NAN` on windows.
It is different between linux and windows.
See:
https://github.com/open62541/open62541/pull/1387/files#diff-3fe8a60c370c9b03b1388aaf5d8b9363R815

Is there anything specific in the spec? If windows does it wrong then we need to somehow change the float binary encoding on windows.

PR #1392, opened 2017-11-23
fix possible infinity loops and add a unit test unit

StalderT:


StalderT:
Can maybe fix SecureChannel_timeout_fail test. I will try to reactivate the test if all green.

StalderT:
I think we must add ```connection->close(connection);``` here

https://github.com/open62541/open62541/blob/dfc2543739e19e2ae86d875e4d3c7ae4adaaf49c/plugins/ua_network_tcp.c#L200-L203

@jpfr @Pro do you agree ?

Pro:
@StalderT  thanks for the hint about `--track-fds=yes`.
I added a new PR which enables this. See #1395
That PR also includes all the necessary close calls so that no connection remains open

StalderT:
What is the best way to force Travis/AppVeyor to rebuild?

Pro:
@StalderT only we (admins) can force Travis/Appveyor to requild. If you really really need to force a new build you could e.g. rebase your branch on master, or if it is already up to date then push a new commit to your branch.

The better way would be if you enable Travis and Appveyor for your own fork and then you develop your features first only in your fork where you can restart Travis/Appveyor on your own, and when all is green you can create a PR.

StalderT:
@Pro Ok thanks. I understand. Do you know where can I activate Appveyor / Travis in my fork.

Pro:
It's the same as adding a normal project:

https://stackoverflow.com/questions/26343492/test-a-forked-github-project-by-travis

https://www.appveyor.com/docs/

StalderT:
for me you can merge

PR #1395, opened 2017-11-23
Enable valgrind open FD check and fix open FDs after shutdown

Pro:


StalderT:
👍 thank you

Pro:
Yeah right... I thought about 0 is stdout and thus it will not be used by our app, but if you are on an embedded device, it could also be 0. Thanks, will fix that

StalderT:
@Pro I made : this https://github.com/StalderT/open62541/commit/f8e7160d1af8d00b4f62f896ddefea0c5e7814d3

If you want I can make a PR

Pro:
Thanks for the idea! `INVALID_SOCKET` on Windows is just a define for -1, and SOCKET is a define for int.
Therefore I would say that we should keep the code simple and only use ifdef where necessary. Additionally we do not use these defines within the `ua_network_tcp.c`
The only really minor change I would make is to use `UA_Int32` instead of `int` to be consistent with `UA_Connection`.

Stasik0:
In my oppinion, one should not rely on the fact that INVALID_SOCKET is -1 in a particular header. Things may break later.

StalderT:
Actual ```ua_network_tcp.c```
https://github.com/open62541/open62541/blob/0048d5aa5e43dee4e95edc107d13cbfb63f2606c/plugins/ua_network_tcp.c#L348-L351

Pro:
Ok, agee 👍 
@StalderT would be nice if you can create a PR for that

HolgerJeromin:
> In my oppinion, one should not rely on the fact that INVALID_SOCKET is -1 in a particular header. Things may break later.

We could have a central ua_invalid_socket if the code should be nicer

PR #1403, opened 2017-11-27
fix testing clock and simplify receiveServiceResponse

StalderT:
testing_clock must take into account the real elapsed time (for example while waiting on a select).

With testing_clock fixed, receiveServiceResponse can now be simplified.

@Pro can you see why discovery test fails ?

https://travis-ci.org/StalderT/open62541/jobs/307870584#L6383

https://github.com/open62541/open62541/blob/5ff6e4b3fc29b53e580f68acafbd6a3e6ff25617/tests/server/check_discovery.c#L178-L187

### DIFF
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -281,10 +281,7 @@ receiveServiceResponse(UA_Client *client, void *response, const UA_DataType *res
         retval = UA_Connection_receiveChunksBlocking(&client->connection, &rd,
                                                      client_processChunk, timeout);
 
-        if (retval == UA_STATUSCODE_GOODNONCRITICALTIMEOUT)
-            break;
-
-        if(retval != UA_STATUSCODE_GOOD) {
+        if(retval != UA_STATUSCODE_GOOD && retval != UA_STATUSCODE_GOODNONCRITICALTIMEOUT) {
             if(retval == UA_STATUSCODE_BADCONNECTIONCLOSED)
                 client->state = UA_CLIENTSTATE_DISCONNECTED;
             else
--- a/tests/client/check_client_securechannel.c
+++ b/tests/client/check_client_securechannel.c
@@ -56,7 +56,8 @@ START_TEST(SecureChannel_timeout_max) {
     UA_StatusCode retval = UA_Client_connect(client, "opc.tcp://localhost:4840");
     ck_assert_uint_eq(retval, UA_STATUSCODE_GOOD);
 
-    UA_fakeSleep(UA_ClientConfig_default.secureChannelLifeTime);
+    UA_fakeSleep((UA_UInt32)(UA_ClientConfig_default.secureChannelLifeTime*0.95));
+                             // Less than UA_ClientConfig_default.secureChannelLifeTime
 
     UA_Variant val;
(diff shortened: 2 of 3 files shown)


Pro:
The new clock implementation introduces a timing issue:

The registration of the server times out too early:
https://travis-ci.org/StalderT/open62541/jobs/307870584#L6223

And therefore the unregister call fails with "BADNOTHINGTODO" here
https://travis-ci.org/StalderT/open62541/jobs/307870584#L6229

On every server iteration the registered servers are checked. If they don't update the registration within X Minutes, it is timed out. It looks like your implementation of the simulated clock causes the timeout to happen too early.
See
https://github.com/open62541/open62541/blob/master/src/server/ua_services_discovery.c#L519

For the unit test we set the timeout to 1 second
https://github.com/open62541/open62541/blob/master/tests/server/check_discovery.c#L33

Not sure why this happens though, maybe you need to debug into it.

StalderT:
@Pro  thanks for your quick reply.

If we wait here 2 seconds (more than the timeout) : 

https://github.com/open62541/open62541/blob/5ff6e4b3fc29b53e580f68acafbd6a3e6ff25617/tests/server/check_discovery.c#L180-L181

It's normal that we have a timeout and that UA_Server_unregister_discovery return BADNOTHINGTODO ?

Pro:
Yes, if we wait more than 1 second, the registration of the server should time out.
Since we set the timeout to 1 second, we expect that the server re-registers within 1 second. If not, we remove it.

With the call to unregister in the unit test we check if this timeout functionality is working correctly. We expect that the server is still registered when we call unregister, and therefore it is a unit test fail if the server is already kicked out

StalderT:
I believe that I have understood. In this case we should not use UA_fakeSleep over 1s

jpfr:
> testing_clock must take into account the real elapsed time (for example while waiting on a select).

I don't understand. Why is that required?

The advantage of our current appraoch is that can determinisitically simuate timing conditions. If we add the real system clock, additional delays for context switches etc. might make it impossible to simulate some edge cases.

StalderT:
I found the problem in ```receiveServiceResponse```

```
UA_DateTime maxDate = UA_DateTime_nowMonotonic() + 5000;

while(1){
    UA_DateTime now = UA_DateTime_nowMonotonic();
    if(now >= maxDate)
        return;
    
    make somthing like select with timout but select return nothing

    if (select return something)
        return;
}
```

With actual implementation ```maxDate``` = 5000 and ```now``` always equal to 0 and we never exit the loop. We have a infinit loop. In real life we exit the loop after 5s.

Works with the PR.

What do you think?

jpfr:
Ok. Understood.
Thanks for the explanation.

Can we have both?
Deterministic tests and not loop indefinitely here?

We could overload the "select" method in the network layer for the testing plugin...
Just overwrite the function pointer in the server-config. With a "man in the middle" that forwards time if he needs to.

StalderT:
@jpfr Ok I agree. Somethink like that should works : https://github.com/StalderT/open62541/commit/79ea8cdefaee223cb3b68a88f26c50b75b0a8947

If you agree, can you try to integrate this?

jpfr:
Looks good. Can you try to integrate yourself?

I can do it. But I would need some days to find a timeslot...

StalderT:
I can try but currently I don't know how I can overlap UA_select.

PR #1407, opened 2017-11-28
fix select test timing

StalderT:
https://github.com/open62541/open62541/pull/1403


plugins/ua_network_tcp.h
```diff
@@ -11,12 +11,24 @@ extern "C" {
 #include "ua_server.h"
 #include "ua_client.h"
 
+#ifdef _WIN32
+# include <winsock2.h>
+#else
+# include <sys/time.h>
+# include <sys/types.h>
+# include <unistd.h>
+#endif
```
Pro: Can you put these includes into the corresponding source file to keep the header free of any non necessary includes?
That avoids that these files will also be included in any user program.

StalderT: It's difficult because we need to define fd_set and timeval for UA_select

jpfr:
Hey @StalderT ,

I tried the approach where the client-side recv method is extended.
This makes only minimal changes. And only in the testing plugin.
We don't want to complicate the public headers with "internals".

Can you review my last commit?
The overriden recv method is used in one of the client-securechannel tests.

The server-side is more complicated. Is it enough to start with the client only?

StalderT:
@jpfr

👍 👍 👍 good idea ! But I did not understand why to use UA_Client_recvSleepDuration

It will change the server part when we have a test that does not work. Or adding a #define in the code like:

```
    ret = select(....);
#if TEST
    if (ret==0)
        UA_fakeSleep(timeout);
#endif
```

StalderT:
@jpfr 

I had a test with blockServer. This test simulated the non-response of a server (like plugging out the network cable). This test has been removed. Is it possible to add one or do you think this case is tested elsewhere? Actual test only check UA_STATUSCODE_BADSECURECHANNELCLOSED.

```
-    while(*running){
+    while(*running)
-        if (*blockServer)
+        UA_Server_run_iterate(server, true);
-            UA_realSleep(100);
-        else
-            UA_Server_run_iterate(server, true);
-    }
```

jpfr:
> It will change the server part when we have a test that does not work. Or adding a #define in the code like:

The define is not necessary. The client will only use the "man in the middle" recv method when we explicitly set it. See here:
https://github.com/StalderT/open62541/blob/920005bed944643a7738905b7edf99d95927fc6e/tests/client/check_client_securechannel.c#L104

> I had a test with blockServer. This test simulated the non-response of a server (like plugging out the network cable).

Maybe we can extend `UA_Client_recvSleep` to `UA_Client_recvTesting` where we can externally set the statuscode of the next recv call.

```c
UA_StatusCode
UA_Client_recvTesting(UA_Connection *connection, UA_ByteString *response,
                    UA_UInt32 timeout) {

    if(UA_Client_recvTesting_result != UA_STATUSCODE_GOOD) {
        UA_StatusCode temp  =UA_Client_recvTesting_result;
        UA_Client_recvTesting_result = UA_STATUSCODE_GOOD;
        return temp;
    }

    UA_StatusCode res = UA_Client_recv(connection, response, timeout);
    if(res == UA_STATUSCODE_GOODNONCRITICALTIMEOUT)
        UA_fakeSleep(timeout);
    else
        UA_fakeSleep(UA_Client_recvSleepDuration);
    UA_Client_recvSleepDuration = 0;
    return res;
}
```

StalderT:
OK for me 👍 but we need to make ```UA_fakeSleep(timeout);``` before return temp;

jpfr:
@StalderT 
Feel free to continue in that direction.
You're now the owner of this. 😄

StalderT:
for me you can merge

Pro:
I would see the `UA_Client_recvTesting` method better fitting into `testing_networklayers.c` instead of `testing_clock.c` since it is mainly a network related method.

Can you change that, then we can merge?

StalderT:
@Pro done and rebased

PR #1117, opened 2017-07-17
made client non blocking during connecting stage;

sherylll:
only for review, code that use the modified function has to be changed

PR #1410, opened 2017-11-29
WIP: Async client API

ccvca:
Hi,
This pull request add UA_Client_AsyncService_read, UA_Client_AsyncService_write, UA_Client_AsyncService_call. (Like [include/ua_client.h]( https://github.com/open62541/open62541/blob/8b685b6540231813b61df8f6cac9b4561b894fd3/include/ua_client.h))

- Add field useBlockingSocket to UA_ConnectionConfig
- Add UA_Client_AsyncService_* functions and example/client_async.c

If there are no proposals for a diffrerent way of implementing , I would add the other UA_Client_Service_* functions before merging this request.

ccvca

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -298,6 +298,7 @@ set(exported_headers ${PROJECT_BINARY_DIR}/src_generated/ua_config.h
                      ${PROJECT_SOURCE_DIR}/include/ua_plugin_nodestore.h
                      ${PROJECT_SOURCE_DIR}/include/ua_server_config.h
                      ${PROJECT_SOURCE_DIR}/include/ua_client.h
+                     ${PROJECT_SOURCE_DIR}/include/ua_client_async.h
                      ${PROJECT_SOURCE_DIR}/include/ua_client_highlevel.h
 )
 set(internal_headers ${PROJECT_SOURCE_DIR}/deps/queue.h
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -64,6 +64,8 @@ add_example(client_connect_loop client_connect_loop.c)
 
 add_example(server_mainloop server_mainloop.c)
 
+add_example(client_async client_async.c)
+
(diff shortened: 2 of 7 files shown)


jpfr:
Nice one!
Do you also want to overload the callback methods with the exact request and response message types?

ccvca:
>  Do you also want to overload the callback methods with the exact request and response message types?
I do not see a possibility to pass the Userdata and UserCallback-Pointer to a callback function at the moment. Otherwise  UA_Client_AsyncService* could register a static callback wich do the conversions and call the custom callback. But this would require some rework in ua_client.

@jpfr Do you know an easy solution for a typed callback?

FlorianPalm:
There is ongoing work in another pull-request
https://github.com/open62541/open62541/pull/1117
Looking at your code, it seems your work is based on this pull request. We are going to add typed callbacks for user-land aswell.

jpfr:
> @jpfr Do you know an easy solution for a typed callback?

Easy. Just cast the method pointer. Only the response argument changes.
And the pointer can be cast from void* to UA_ReadResponse* or similar.
```c
void valueRead(UA_Client *client, void *userdata,
              UA_UInt32 requestId, const UA_ReadResponse *response);
// Then cast the method pointer to (UA_ClientAsyncServiceCallback*)
```

@ccvca Can you coordinate with FlorianPalm so that we don't duplicate work?
Merging stuff you already have ahead of time is okay. But the efforts should point into the same direction.

ccvca:
>Looking at your code, it seems your work is based on this pull request.

The example is based on the former branch async_client_loop (Which was a previous stage of your pull-request), the rest just uses __UA_Client_AsyncService which is already part of the master-branch. 

@jpfr @FlorianPalm I didn't recognize that this Pull-Request heads towards the same direction. I'll have a look at it.

ccvca:
After a short overview, #1117 contains all features of this Pull Request, so I will close this one.

Issue #1411, opened 2017-11-29
[BadSequenceNumberUnknown] after PublishResponse without notification


StalderT:
Description
===========
foud with https://github.com/open62541/open62541/issues/1409

After a PublishResponse without notification, next PublishResponse return BadSequenceNumberUnknown

![image](https://user-images.githubusercontent.com/13799456/33376726-190781ae-d50f-11e7-9813-da9d4dd8ab68.png)

Wireshark log : [seq.zip](https://github.com/open62541/open62541/files/1513890/seq.zip)

Background Information / Reproduction Steps
===========================================
 use https://github.com/open62541/open62541/issues/1409

Should we not return after 

https://github.com/open62541/open62541/blob/3db5c9bde1c22d7d9de3c2284a89401f6978d084/src/client/ua_client_highlevel_subscriptions.c#L342

if ```notificationDataSize == 0```

```
if (msg->notificationDataSize==0)
   return;
```

in order to not add current sequence number to the list of pending acks ?

Checklist
=========
Please provide the following information:

 - [x] open62541 Version master
 - [x] Wireshark network dump attached

StalderT:
fixed in https://github.com/open62541/open62541/pull/1497

StalderT:
For the moment #1497 is not merged. The problem is still here.

PR #1412, opened 2017-11-29
fix for #1411

StalderT:


StalderT:
if you agree, for me you can merge

Pro:
Fix for #1411

jpfr:
I have to admit, I don't fully understand the change.

Every PublishResponse contains exactly one NotificationMessage.
And every NotificationMessage has a sequence number.
In the existing code we add the new number to the list of pending acknowledgements.

In the patch you add the pending acknowledgements in the list a second time.
I'm not sure about that. Can you add comments that explain what is happening?
Otherwise the next developer will be seriously stuck looking at the code.

StalderT:
Ok, I missed something. Sorry!!!!

Thank you for the explanations.

I try to explain actual problem.

With current code : 

--------------

1) First publish response (with a notification message)

image

AvailableSequenceNumbers 1
Sequence Number 1

--------------

2) Next publish request ack sequenceNumber 1

image

-----------------

3) Next publish response (without notification message)

image

AvailableSequenceNumbers : none
Sequence Number 2

--------------

4) Next publish request ack sequenceNumber 2

image

--------------

5) Next publish response (without notification message) return BadSequenceNumberUnknow
because last publish request try to ack sequenceNumber 2.

Because SequenceNumber 2 is not present in retransmission message queue  : 

```
    if(notifications == 0) {
        /* Send sequence number for the next notification */
        message->sequenceNumber = sub->sequenceNumber + 1;
    } else {
        /* Increase the sequence number */
        message->sequenceNumber = ++sub->sequenceNumber;

        /* Put the notification message into the retransmission queue. This
         * needs to be done here, so that the message itself is included in the
         * available sequence numbers for acknowledgement. */
        retransmission->message = response->notificationMessage;
        UA_Subscription_addRetransmissionMessage(server, sub, retransmission);
    }
```


image

-----------------------------

I propose a new fix.

StalderT:
Ok to merge or do you need more information ?

StalderT:
Can we merge? Do you want changes or explanations?

Issue #1224, opened 2017-09-20
open62541 vs Indusoft Web Studio interoperability issue


adliner:
Hi,
I tried using open62541 sample server codes with Indusoft Web studio by Wonderware. Unfortunately Web studio gets frozen when trying to connect to sample server. Is it any known issue already seen with any other OPC-UA clients?

This can be easily reproduced by downloading a trial version of Indusoft Web Studio 7 or 8: http://www.indusoft.com/Products-Downloads/HMI-Software/InduSoft-Web-Studio

Then simple clicking:
![image](https://user-images.githubusercontent.com/8227441/30666413-885ba586-9e54-11e7-8c3b-c111f8643c4e.png)

I tried with latest amalgamated open62541 lib in demo project: https://github.com/adliner/OPCServerDemo but this issue also occurs with any other sample server codes.

This basically looks more like Indusoft Web Studio bug but on the other hand it works fine with other servers based on different OPC-UA stacks, so perhaps it's some open62541 standard compatibility issue.

PS: I also captured WireShark traffic dump, available at: https://github.com/adliner/OPCServerDemo/blob/master/OPC/indusoft_wireshark_dump.pcapng

adliner:
Hi,
I've been trying to investigate this issue on my own. Hence I did some network traffic analysis using Wireshark.

I compared connection between Indusoft client and:
- open62541 sample server
- OPC-UA demo server available at opc.tcp://opcua.demo-this.com:51210/UA/SampleServer


I noticed some thing. open62541 generated packet contain 'ArraySize' set to '-1'(0xffffffff) when describing null string while demo server sends packet containing just zeros (0x00000000) Just see this one below:

![image](https://user-images.githubusercontent.com/8227441/30826350-9dfb7aa2-a236-11e7-97d3-59ed4d94b970.png)


Is this expected behavior in regard of OPC-UA specification?

jpfr:
Hey @adliner ,

both are valid values in OPC UA.

-1 is an undefined string.
0 is a string of length zero.

The application should be able to deal with both.

Stasik0:
Can't we change to 0 to be "more interoperable"?

Stasik0:
@adliner: you can try to tweak the answer yourself:

https://github.com/open62541/open62541/blob/b3ad5a6c89ebb14654ea5239111ee34bffcec14f/src/server/ua_services_securechannel.c#L9

[...]

PR #1417, opened 2017-11-30
possible fix for #1224

StalderT:
see #1224


Pro:
@StalderT can you add the corresponding issue to your PRs description additionally to the commit message?
This will make it easier to track it from the issue itself since it will show up in the issue list.

I edited the description for this PR.

PR #1419, opened 2017-11-30
Fix missing code for VxWorks

cabralfortiss:
Architecture dependant code that changed in the last few months are fixed to compile in vxworks


jpfr:
Merged.
Thanks for the contribution!

Issue #1255, opened 2017-10-16
Split huge generated files of nodeset compiler into smaller methods/files


Pro:
This is an issue which can be grabbed by first time contributors and is fairly easy to implement.

The nodeset compiler just generates one huge file with one huge function which initialized the address space.

To avoid issues with stack overflow on Windows (e.g. see #1251) this file should be split into multiple subfiles and sub functions.

jpfr:
Hey @Pro 
Do we have to split files? Our amalgamation is huge and compiles reasonably fast.
Maybe it's enough to split the namespace generation into smaller methods.

Pro:
I'm not sure if the stack overflow in Visual Studio (as mentioned in #1251) is caused by the huge method or huge file.
Probably it's the huge method, so first try should be to split the method into sub-methods (which also makes the whole process easier to handle compared to multiple files)

Pro:
This was fixed with #1424. The huge method is now split into one method per node

PR #1421, opened 2017-11-30
Make small functions to create the nodeset instead of a big one

cabralfortiss:
#1255


Pro:
Nice, thanks!
This will create for every node a separate function, right?

@jpfr should we try to reduce the number of functions for performance improvement, e.g. only create a new function for every 10th node?

PR #1429, opened 2017-12-02
Fix heap buffer overflow

Pro:
Fix heap buffer overflow if exchangeBuffer did not help

see https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=4496

Credit to oss-fuzz


jpfr:
Woot! oss-fuzz is getting a lotta credit these days.

Pro:
Yes, don't know why there are now so many hits, but very interesting :+1: Maybe because we now have a more complete corpus.

This one was a very tough one, it only showed up if the fuzzer was run at least 200 times. Some very interesting integer overflow when subtracting `g_end - g_pos` while `g_end` is smaller than `g_pos`. That is only achieved if `exchangeBuffer()` is called two times without any additional encoding. The integer overflow was only caused if the `g_end` is on a specific address.

Issue #968, opened 2017-02-19
UA_Client_manuallyRenewSecureChannel() successfully, if connection is closed


Root-Core:
Description
===========

An call of `UA_Client_manuallyRenewSecureChannel()` returns `UA_STATUSCODE_GOOD` if the connection is closed.

This leads to errors in the `__UA_Client_Service()`, as the following `UA_SecureChannel_sendBinaryMessage()` call will fail with status `UA_STATUSCODE_BADCONNECTIONCLOSED`. This again will set the clients state to `UA_CLIENTSTATE_FAULTED`, but as the connection is dead it should be `UA_CLIENTSTATE_ERRORED`.

It would be set to `UA_CLIENTSTATE_ERRORED`, if the `UA_Client_manuallyRenewSecureChannel()` would return an faulted status.

Background Information / Reproduction Steps
===========================================

- Connect an client to an server
- Use eg. `UA_Client_Service_write()` to set values on the server
- Let the client loop
- If you now shutdown the server, the client will try to renew the secure channel in an loop 

I hope that's all, as the actual client is way more dynamic.
But this should be enough to reproduce the issue.

Checklist
=========

 - open62541 Version: 2.0 trunk (server and client)
 - Operating system: ELDK on ARM7
 - Critical issue: It might be, as the state machine is broken.

I added some additional logging to `__UA_Client_Service()` (just search "ROOT_CORE"):
https://gist.github.com/Root-Core/2e0924a175deca05297d7836824389f0

Output:
> ROOT_CORE: Renew success
> Sending a request of type 671
> ROOT_CORE: Binary message result: 80ae0000
> ROOT_CORE: Renew success
> Sending a request of type 671
> ROOT_CORE: Binary message result: 80ae0000
> [...]

Thank you all for your good work, keep it up.  :thumbsup:

jpfr:
We need to make this behaviour configurable for 0.3.
Maybe with boolean options in the client config
- autoReconnectSession
- autoReconnectSecureChannel

Then, an appropriate error is returned when autoreconnect is disabled *or* if autoreconnect fails.

For 0.2, returning the correct status code is fine.
@Root-Core Can you make a PR from the changes?
Concerning the logging: This is a good idea. In addition, you can use `UA_StatusCode_name` to get a "speaking" name of the statuscode instead of logging hex codes.

Best regards,
Julius

Root-Core:
I will look into it. Should I add this kind of logging there?
[...]

PR #1436, opened 2017-12-06
add test for #968

StalderT:
#968


jpfr:
Merged. Thanks again!

PR #1441, opened 2017-12-07
fix #1439 (UTC / LocalTime)

### LINKED ISSUE #1439: Logging time differs from system time
The time displayed in the logs is one hour behind my actual time. This happens on linux and windows.
Maybe something with summer/winter time?

revision: 926724d9256fcab391ddf5db526053c24fa2c9fe

StalderT:
fix https://github.com/open62541/open62541/issues/1439

### DIFF
--- a/examples/client_connect_loop.c
+++ b/examples/client_connect_loop.c
@@ -81,9 +81,9 @@ int main(void) {
         if (retval == UA_STATUSCODE_GOOD &&
             UA_Variant_hasScalarType(&value, &UA_TYPES[UA_TYPES_DATETIME])) {
             UA_DateTime raw_date = *(UA_DateTime *) value.data;
-            UA_String string_date = UA_DateTime_toString(raw_date);
-            UA_LOG_INFO(logger, UA_LOGCATEGORY_CLIENT, "string date is: %.*s", (int) string_date.length, string_date.data);
-            UA_String_deleteMembers(&string_date);
+            UA_DateTimeStruct dts = UA_DateTime_toStruct(raw_date);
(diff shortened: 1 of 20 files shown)


jpfr:
Very nice!

For symmetry, I propose we rename the current `UA_DateTime_now` to `UA_DateTime_nowUtcTime` and add a compatibility macro
```c
#define UA_DateTime_now() UA_DateTime_nowUtcTime()
```

so we have
- UA_DateTime_nowLocalTime
- UA_DateTime_nowUtcTime
- UA_DateTime_nowMonotonic

StalderT:
@jpfr Good idea. I modify the PR accordingly.

StalderT:
@jpfr Do you want that UA_DateTime_now is replaced everywhere by UA_DateTime_nowUtcTime? Or is it better to use compatibility macro?

Infinity95:
Maybe it would be a good idea to add an indicator for the timezone to the logs?
Something like `%time% (UTC+%timediff%)`?

jpfr:
Hello there,

I had a look at the standard.
> A DateTime value shall be encoded as a 64-bit signed integer (see Clause 5.2.2.2) which
represents the number of 100 nanosecond intervals since January 1, 1601 (UTC).

So DateTime should normally be UTC only. I think we are good if we just provide  UA_DateTime_localTimeUtcOffset,

The PR now contains some additional simplifications to time conversion.
We had lots of different macros defined. And they started to get confusing.

jpfr:
Found the error. There was a utf8 character in a comment. Latex didn't like that.

PR #1449, opened 2017-12-12
Added free function pointer to the UA_Variant struct

jonasgreen88:
This free function pointer allows the user to get notified by the
stack when the data buffer attached to the variant is free to reuse
if storage type NODELETE is chosen.

### DIFF
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -512,6 +512,7 @@ typedef struct {
     void *data;                   /* Points to the scalar or array data */
     size_t arrayDimensionsSize;   /* The number of dimensions */
     UA_UInt32 *arrayDimensions;   /* The length of each dimension */
+    void (*free)(void *data, size_t arrayLength, const UA_DataType *type); /* Function to free *data if storageType equals NODELETE. Can be set to NULL. */
 } UA_Variant;
 
 /* Returns true if the variant has no value defined (contains neither an array
--- a/src/server/ua_server_utils.c
+++ b/src/server/ua_server_utils.c
@@ -377,7 +377,7 @@ const UA_VariableAttributes UA_VariableAttributes_default = {
     {{0, NULL}, {0, NULL}},      /* description */
     0, 0,                        /* writeMask (userWriteMask) */
     {NULL, UA_VARIANT_DATA,
-     0, NULL, 0, NULL},          /* value */
+     0, NULL, 0, NULL, NULL},    /* value */
     {0, UA_NODEIDTYPE_NUMERIC,
      {UA_NS0ID_BASEDATATYPE}},   /* dataType */
     -2,                          /* valueRank */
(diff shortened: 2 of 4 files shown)


jpfr:
Hello @jonasgreen88,
I'm sure you have some use case for this.
Can you elaborate why this is necessary?

jonasgreen88:
Sure. We have a separate buffer reserved for loading variable values. When the value of a variable node is requested by the stack via the data source callback the actual variable value is loaded into the buffer and a pointer is attached to the variant. If a request contains several read operations it is not possible to reuse the memory buffer immediately when next variable value is requested. The buffers attached to variants are needed by the stack until it encodes the complete response. Instead the next requested value is added to the buffer after the previous one. So all variable values loaded for the read operations in a single request from a network client is packed into this separate buffer we have for variable values. To reset this buffer when a response has been encoded to allow a new request to utilize the complete size of it, some notification from the stack is needed. This free function pointer gives our implementation a notification when the buffers attached to the variant are not needed by the stack anymore.

jpfr:
Do you use a datasource callback to access the value?
You could deep-copy the value during the read.

The server uses malloc during normal operations. Are you sure you need the pointer magic here?

I understand your use case. On the other hand, we mix server internals with the data type definitions.
Imho, the data types should be "clean".

jonasgreen88:
Yes, we use datasource callback to be able to forward error codes related to accessing the value.

Even if we use the heap of the stack to store loaded values we need a way to limit how much of the heap variable values can allocate together for read operations of a single response. Else we could run into the case where the complete heap is allocated for values and when the stack needs temporary memory to encode the response there is no more memory available. Therefore we have a separate buffer for variable values.

It doesn't have to be part of the UA_Variant type. What we are looking for is a notification from the stack when the memory we attach to the variant is free to use again. Do you have any suggestion where it could be implemented?

jpfr:
We could add an additional callback `serviceFinished` to the server.
Then you can add a memory buffer that is reset when the service is done.

But this would only apply to services called over the network. Not 'local' calls to the read service, etc.

jonasgreen88:
That would definitely work as well. Mixing server internals with data types definitions are avoided and one single buffer could be used for datasource callbacks. The buffer is always free to use at the next call to the datasource callback, independent on how many read operations the request contains.

Unfortunately I don't think I will have time to look into what changes that have to be made really soon though.

jpfr:
Is it okay if you maintain your current "hack" internally until we have a long-term solution?

jonasgreen88:
Yes, no problem. We will use this hack until a better solution for open62541 is developed.

jpfr:
Zero-copy read-operations are implemented in #1471.
There you can reuse the same memory for consecutive calls to datasource-read callbacks.

jonasgreen88:
Thanks! Then I close this PR.

PR #1457, opened 2017-12-13
Removes unneeded initializations

StalderT:
Initializations are performed in the __UA_Client_Service function

https://github.com/open62541/open62541/blob/983f956003269732b023e24a68a32bedbedb8f18/src/client/ua_client.c#L306


jpfr:
👍

Issue #1462, opened 2017-12-19
intention of low maxNotificationPerPublish


WoutvdBerg:
What is the intention of the standard 10 maxNotificationPerPublish in UA_SubscriptionSettings_Default?

I'm running a server on a PLC, which aren't fast but with the low maxNotificationPerPublish it makes the communication even worse.

jpfr:
This default seems low indeed.
Feel free to propose a PR where the default is increased.
Or, if possible, set it to zero to allow any number of notifications.

WoutvdBerg:
@jpfr, like this?

jpfr:
Cool, issue resolved.

PR #1464, opened 2017-12-19
use maximum notification per publish

WoutvdBerg:
#1462 
Always use the maximum notification per publish.
This increases speed on slow devices like PLC's, because PLC's can add more notifications in a publish response.


jpfr:
Perfect. 😄
Merged.

Issue #1283, opened 2017-10-28
When monitorring several variables,how can I know which variable has been chanced?


Croton-D:
I tried client.c to monitory some variables.
1:  add monitoredItem like this:
UA_NodeId monitorThis = UA_NODEID_STRING(4, "A");
UA_Client_Subscriptions_addMonitoredItem(client, subId, monitorThis, UA_ATTRIBUTEID_VALUE,
                                             &handler_TheAnswerChanged, NULL, &monId);
UA_NodeId monitorThis = UA_NODEID_STRING(4, "B");
UA_Client_Subscriptions_addMonitoredItem(client, subId, monitorThis, UA_ATTRIBUTEID_VALUE,
                                             &handler_TheAnswerChanged, NULL, &monId);
2:change the handler_TheAnswerChanged function as:
handler_TheAnswerChanged(UA_UInt32 monId, UA_DataValue *value, void *context) {
    printf("monid is %d  and value is %f\n",monId,*(UA_Float*)value->value.data);//the variables are all floats
}
3: call the function to get whether varibles has been changed:
UA_Client_Subscriptions_manuallySendPublishRequest(client);
sleep(200);
UA_Client_Subscriptions_manuallySendPublishRequest(client);

then ,I got the message ：
monid is 1  and value is 1.500000
monid is 1  and value is 2.000000
monid is 1  and value is 2.500000

using UA_Expert,I found A was 1.5 and B was 2.0 changing to 2.5,but the messages reported the same monid.
My question is how can I know which variable has been changed when then has the same monid?

StalderT:
You can use context, like : 

```
UA_NodeId monitorThis = UA_NODEID_STRING(4, "A");
UA_Client_Subscriptions_addMonitoredItem(client, subId, monitorThis, UA_ATTRIBUTEID_VALUE,
&handler_TheAnswerChanged, (void*)1, &monId);
UA_NodeId monitorThis = UA_NODEID_STRING(4, "B");
UA_Client_Subscriptions_addMonitoredItem(client, subId, monitorThis, UA_ATTRIBUTEID_VALUE,
&handler_TheAnswerChanged, (void*)2, &monId);
```

and

```
handler_TheAnswerChanged(UA_UInt32 monId, UA_DataValue value, void* context) {
printf("monid is %d and value is %f context %d\n",monId,(UA_Float)value->value.data, (int) context);//the variables are all floats
}
```

Code should be tested!

jpfr:
[...]

PR #1466, opened 2017-12-19
remove monitoringitem item id from handler

WoutvdBerg:
adds parameter to the datachange handler to identify a monitored item when multiple subscriptions are used

### DIFF
--- a/tests/fuzz/corpus_generator.c
+++ b/tests/fuzz/corpus_generator.c
@@ -360,7 +360,7 @@ translateBrowsePathsToNodeIdsRequest(UA_Client *client) {
 
 
 static void
-monitoredItemHandler(UA_UInt32 subId, UA_UInt32 monId, UA_DataValue *value, void *context) {
+monitoredItemHandler(UA_DataValue *value, void *context) {
 
 }
 


StalderT:
@WoutvdBerg Just to know, why are you using monId and subId in the callback? Why do not you use the variable "context"? Should not we remove monId from the callback function and only use the context?

WoutvdBerg:
@StalderT, because I didn't know where to use the context for.
Can the context be used to assign some custom data that will be transferred later to the handler?

StalderT:
see https://github.com/open62541/open62541/issues/1283

WoutvdBerg:
I understand, I will remove the subId and monId from the handler functions tomorrow. leaving only datavalue and context.

jpfr:
Hey @WoutvdBerg ,

I'm unconvinced about using the context for everything and removing the MonitoredItem identifier.
It simplifies. And simplification is always good. But ...

- Having the identifer in the callback has practically zero cost.
- Maybe a user does not want to allocate custom memory for a monitoreditem and uses a NULL-context. Still, the identifier can be useful. For example to look up more information in a third system.
- We would break the API for many existing applications.

Can have a fresh look at this after the 0.3 release when we make big API changes?
Actually, I quite liked to first approach where the subscription identifier was added to the callback. 😄

WoutvdBerg:
I agree with your points but I didn't get the concept with only a monitoreditemid returned. This only makes sense in singlesubscription applications. Both solutions are in the pull request.

> Can have a fresh look at this after the 0.3 release when we make big API changes?
Agreed

I will let the pull request as it is.

jpfr:
#1497 has landed and the API was completely revamped.
The old API is still available, but marked as deprecated.
Please reopen if you want to make these changes to the (now) deprecated API.

PR #1467, opened 2017-12-19
client subscriptions improvement

StalderT:
The client must know when a new session has been created in order to recreate the subscriptions. The current implementation does not allow to do it. The PR adds a callback that is called when the client changes state. An example has been added. In addition, when a new session is created, open62541 automatically releases the subscriptpion and the monitereditems.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -12,6 +12,20 @@
 
 #define UA_MINMESSAGESIZE 8192
 
+
+ /********************/
+ /* Set client state */
+ /********************/
+static void
+setClientState(UA_Client *client, UA_ClientState state)
+{
+    client->state = state;
+    if (client->state != client->lastSendState){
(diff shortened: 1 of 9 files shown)


src/client/ua_client_connect.c
```diff
@@ -12,6 +12,20 @@
...
 #define UA_MINMESSAGESIZE 8192
 
+
+ /********************/
+ /* Set client state */
+ /********************/
+static void
+setClientState(UA_Client *client, UA_ClientState state)
+{
+    client->state = state;
```
jpfr: Can we simply check if state == client-state at the beginning?
This avoids having a second lastSendState variable.

StalderT: Yes ! 👍

jpfr:
Nice one! Rebased.
Pull is okay from my side.

jpfr:
Merged. Thanks!

PR #1446, opened 2017-12-11
fix #1443

StalderT:
fix #1443

Issue #1443, opened 2017-12-11
client delete/disconnect, secure channel related?


WoutvdBerg:
Description
===========
Total edit:

Using the connect example and breaking the connection results in malloc error.
Also not breaking cable with current master results in malloc error.

Program breaks on UA_Client_delete and _disconnect. I think because of the secure channel changes made since 0.3

Reverted back to 0.2 stable and it worked there.

Wireshark:
[readrequest in closesessions requests.pcapng.zip](https://github.com/open62541/open62541/files/1547229/readrequest.in.closesessions.requests.pcapng.zip)
lldb: 
![schermafbeelding 2017-12-11 om 09 57 19](https://user-images.githubusercontent.com/31468040/33824055-651da3b4-de5d-11e7-9bc6-a655e2bfeed6.png)
Background Information / Reproduction Steps
===========================================
reconnecting example
Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): current master
 - [x] Other OPC UA SDKs used (client or server):  Siemens PLC/uaCppServer
 - [x] Operating system: OS X
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [x] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

StalderT:
Can you provide the line 17752 of your actual open62541.c file ?

Can you try with https://github.com/open62541/open62541/pull/1444 ?

WoutvdBerg:
It has to do with the remoteNonce.

UA_ByteString_deleteMembers(&channel->remoteNonce);

If I comment this out and try to reconnect it also fails on the remoteNonce but in 
processDecodedOPNResponse - >
UA_ByteString_deleteMembers(&client->channel.remoteNonce);

With #1444 it just goes by disconnectWhenFaulted but still crashes.

StalderT:
And if you comment ```UA_ByteString_deleteMembers(&response->serverNonce);
``` in ```processDecodedOPNResponse``` ?

WoutvdBerg:
With 
`UA_ByteString_deleteMembers(&client->channel.remoteNonce);` Commented out it works (with comment in UA_SecureChannel_deleteMembersCleanup)

` UA_ByteString_deleteMembers(&response->serverNonce);` commented only results in malloc error

StalderT:
Can you try to only replace 

```client->channel.remoteNonce = response->serverNonce;```

 by

[...]

PR #1468, opened 2017-12-20
fix #1458

### LINKED ISSUE #1458: closeSession between packets reassemble
Description
===========
When browsing a node with a lot of references to nodes there is a session close request between the network packets.

The problem occurs in __UA_Client_Service and fails because UA_Client_disconnect is called because UA_STATUSCODE_GOODNONCRITICALTIMEOUT is returned from receiveServiceResponse.
Disabling the: `if(now >= maxDate) ` return results in a good read/browse.

browse request with 4 times a node with 1500 nodes.
Packet reassembling is good but there is a random close session request.

Program: 
https://pastebin.com/raw/Lbt9g2vh
loglevel 100: 
https://pastebin.com/raw/wAHNc7Aj
wireshark:
[closesessionrequest.pcapng.zip](https://github.com/open62541/open62541/files/1558774/closesessionrequest.pcapng.zip)

Checklist
=========

 - [x] open62541 Version (release number or git tag): current master
 - [x] Other OPC UA SDKs used (client or server): Siemens
 - [x] Operating system: OS X
 - [ ] Critical issue

StalderT:
fix #1458 
see #1446 and #1443 

For info, the server part does not check if UA_ByteString_copy failed : 
https://github.com/open62541/open62541/blob/2b3c72dc48f65dd5210dc3ecbdc39db9a753959b/src/server/ua_securechannel_manager.c#L156
https://github.com/open62541/open62541/blob/2b3c72dc48f65dd5210dc3ecbdc39db9a753959b/src/server/ua_securechannel_manager.c#L201-L203

@WoutvdBerg can you test with this PR?

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -111,13 +111,13 @@ static void
 processDecodedOPNResponse(UA_Client *client, UA_OpenSecureChannelResponse *response) {
     /* Replace the token */
     UA_ChannelSecurityToken_deleteMembers(&client->channel.securityToken);
-    client->channel.securityToken = response->securityToken;
-    UA_ChannelSecurityToken_init(&response->securityToken);
+    UA_ChannelSecurityToken_copy(&response->securityToken, &client->channel.securityToken);
+    UA_ChannelSecurityToken_deleteMembers(&response->securityToken);
 
     /* Replace the nonce */
     UA_ByteString_deleteMembers(&client->channel.remoteNonce);
-    client->channel.remoteNonce = response->serverNonce;
-    UA_ByteString_init(&response->serverNonce);
+    UA_ByteString_copy(&response->serverNonce, &client->channel.remoteNonce);
+    UA_ByteString_deleteMembers(&response->serverNonce);
 
     if(client->channel.state == UA_SECURECHANNELSTATE_OPEN)
         UA_LOG_DEBUG(client->config.logger, UA_LOGCATEGORY_SECURECHANNEL,


jpfr:
I don't understand the change.
The end-result of "moving" the values should be the same as copying.
With the added advantage that "moving" cannot fail.

Can you explain or recreate a situation where the changes of this PR prevent a bug?

WoutvdBerg:
@StalderT confirmed, problem is solved.

@jpfr, problem occurred when timeout was smaller than the reply from server to a read request. 
Also using a small securechannelrenewal (replacing 75% with 0.01%) made the client crash. Now it doesn't.

StalderT:
Maybe the ByteString data memory is allocated in the response header and freed here https://github.com/StalderT/open62541/blob/294c4da78a3391c8b1434e61737e4c94d01b371d/src/client/ua_client_connect.c#L194 ???

UA_ByteString_copy prevent this

WoutvdBerg:
Tested a further 2 weeks, problem never occured again with the fix.

@jpfr, will this fix be part of 0.3?

jpfr:
Hello @WoutvdBerg ,

I would have merged this long ago. But I actually don't understand how this can fix anything!
The code is exactly the same. Only that it "moves" the string instead of copying it.

@StalderT suggested that the string may be deleted somewhere else.
But this is impossible once we _init it in the response.

To make a long story short, if these changes actually do fix anything, then we have a bigger problem that we don't know yet enough about.

Do we have the chance of adding a unit test that reproduces the issue you see @WoutvdBerg ?

WoutvdBerg:
Hey @jpfr,
Thanks for the answer.

Adding a unit test would be a great.
I will dig deeper into this issue.

StalderT:
@jpfr Where and how is generate response->securityToken ?

WoutvdBerg:
@jpfr, @StalderT,
This one can be closed without merging. #1446 fixed the problem already.
Two different commits were mixed during testing somehow. I tested the current master, works. Reverted commit c01989ee94274941e18c0f4bf2c1c484516b189a and problem occurred again.

Tested with the client reconnect example so no need for another unit test.

StalderT:
Thanks for your feedback.

PR #1481, opened 2017-12-24
fix cppcheck error

StalderT:
[server_certificate.c:31]: (error) Resource leak: fp
[server.c:34]: (error) Resource leak: fp


jpfr:
Nice, thanks!

PR #1487, opened 2018-01-03
WIP: Don't require thread-local support on all platforms

jpfr:


Pro:
When this is merged we can probably also fix the TinyCC PR #1225

PR #1496, opened 2018-01-08
ua_network_tcp select check error

StalderT:
For example, pressing ctrl+c in client_subscription_loop cause "select" returning -1. After, the client infinity wait in the next recv. We must check when select return -1.


jpfr:
Looks good!
Just one little thing.
The second comment `/* No result */` does not look right.

StalderT:
Thanks. I will check tomorrow.

StalderT:
fixed / rebased

jpfr:
Top!

Issue #511, opened 2016-01-15
Implement an event-loop in the client


Stasik0:
We got the following:

Hello, 

I compiled and implemented the server and client from the tutorials and almost everythings works fine.
1. However when I change the value by writing a new Value, the value gets changed but when I call out UA_Client_Subscriptions_manuallySendPublishRequest(), the handler ("The Answer has changed") isn't call out. 
   
   What could be the reason for this ?
2. Also I realized that in order to find out whether the value on my MonitoredItem  has changed, I continuosly need to call UA_Client_Subscriptions_manuallySendPublishRequest().

Is there a way to set  the subscription, so that it automatically calls the PublishRequest whenever the value changes?

@ichrispa could you please look into it, do you monitor the mailing list?

ichrispa:
Hi @Stasik0,

we are getting a lot of tutorial related stuff lately... please forward to the mailing list:

(1) Please provide a test case client/server and the version of the stack to test against.
(2) No; there where attempts by @FlorianPalm was working on a clientside job/multithreading case? It can be attempted to actually multithread this by hand, but I last used that 9 months ago and I have no clue if that still works... it did back then.

Best regards,

ichrispa

NetoLe:
for no 2. If I get this right you mean kind of a push-notification? When I read over the spec I think this is not meant to be implemented.
I looked up in the spec and it says: "c) NotificationMessages are sent to the Client in response to Publish requests." (Part 4 p.69)
and "d) At the beginning of a cycle, if there are Notifications to send but there are no Publish requests queued, the Server enters a wait state for a Publish request to be received."
So there is always a need fo a request-call from the client side, where you can't know if the value changed. 
Especially: "requestHandle IntegerId The requestHandle given by the Client to the request."
which is located in the responseHeader, so the client needs to call the request. 

Or did I get this wrong and you mean that you send a request and the response will be hold until a real value changed?

jpfr:
> Or did I get this wrong and you mean that you send a request and the response will be hold until a real value changed?

Something like that.
There is an internal timer that samples values with a recurring interval -> MonitoredItem
A second timer sends out all the samples received until then for several variables -> Subscription

OPC UA strictly adheres to the request/response pattern. But responses can be sent asynchronously and do not need to respect the ordering of the requests. Therefore, we can just "freeze" publish requests for the subscription and reply only when the timer runs out.

ichrispa:
[...]

Issue #1485, opened 2018-01-02
Connection is closed by subscriptions with publishingInterval*maxKeepAliveCount > client timeout


basyskom-jvoe:
Description
===========
While adding support for subscription parameters to QtOpcUa, I have run into a problem with subscriptions in the open62541 client.

Currently, PublishRequest and PublishResponse are treated as synchronous messages (The timeout value in UA_ClientConfig_default is 5 seconds):
- A PublishRequest is sent
- The client waits up to timeout in UA_Client_Subscriptions_manuallySendPublishRequest
- If a PublishResponse is received before the timeout, everything works as expected

If a publishingInterval > timeout or a combination of publishingInterval*maxKeepAliveCount > timeout is used, there are two cases:
- The value changes before the timeout, a PublishResponse is received, everything works as expected
- There is no value change, the client runs into the timeout

For the second case, the check in ua_client.c line 324 applies and the connection to the server is closed.
According to the specification, the server can delay sending the PublishResponse until maxKeepAliveCount is reached, so this timeout should not lead to a closed connection.

I see no practicable workaround for this issue. Increasing the client timeout to a value publishingInterval*maxKeepAliveCount would block the client even longer.

IMO, the only possible solution is to use asynchronous messages for PublishRequest and PublishResponse.
Sending the publish requests could be enabled after the first successful call to UA_Client_Subscriptions_new, the next one is sent after the response for the previous request has been received.

Any ideas?

Background Information / Reproduction Steps
===========================================
Create a subscription with requestedPublishingInterval of 6000
Add a monitored item for a value that does not change often
Send a PublishRequest using manuallySendPublishRequest

Checklist
=========
Please provide the following information:

 - [X] open62541 Version (release number or git tag): 386a2f4c808c77853a8057133208554644d0047d
 - [X] Other OPC UA SDKs used (client or server): open62541 server and other servers
 - [X] Operating system: Ubuntu 16.04

StalderT:
https://github.com/open62541/open62541/issues/1332

StalderT:
Can be closed (#1497) ?

PR #1497, opened 2018-01-08
client async publish and new subscription API

StalderT:
According the open62541 readme : 

image

Asynchronous service in the client will be in 0.4 release. 

Currently, the client use synchronous UA_Client_Subscriptions_manuallySendPublishRequest who can cause problems (see https://github.com/open62541/open62541/issues/1485).

Like https://github.com/basyskom-jvoe, IMHO, the only possible solution is to use asynchronous service for PublishRequest.

@jpfr Can we add a minimal asynchronous service for PublishRequest in 0.3 ? Could you give your opinion and comment the PR?

Could not we also add an asynchronous minimum service, like #1410 in version 0.3 until #1117 is complete?

Thanks in advance.


jpfr:
Hey @StalderT,

if we can get an implementation and testing in before 0.3, that would be great.
If you are motivated, you have all the support.

I would make `UA_Client_AsyncService_backgoundPublish` an internal method that is not visible to the end user.

So the user just calls `UA_Client_runAsync` and does not have to do anything else.
Inside of `UA_Client_runAsync` we then also ensure that the server has enough publish requests.

StalderT:
@jpfr Thanks for your reply. I will work on it quickly!

WoutvdBerg:
Are there function to add/delete monitoreditems aysnc, or will they be part of this PR?

Because a async publish loop without async adding of monitoring items means that monitored items have to be added before running the publish loop.

Correct me where I'm wrong.

StalderT:
In the example the monitored items are added before the publish loop. But you can add/remove monitored items between two calls of runAsync. Do you need another functionnality?

jpfr:
This is getting good!

Two comments:

- I understand that `backgroundPublishResponseTimeout` defines a timeout between calls to `publish`.
Other implementations try to keep a fixed number of publish requests `in transit` (say 10). So when they receive two publish responses, two new requests are sent out. Note that some servers send events as soon as possible. So you cannot always rely on a fixed publication interval.
Can we replace `backgroundPublishResponseTimeout` with `publishRequestsInTransit`. I'm not sure whether this is a good wording. The standard should have some terminology for that.

- Can we directly use `UA_Client_processPublishResponse` for the async service callback? That removes one level of indirection.

StalderT:
1) ```backgroundPublishResponseTimeout``` is used to generate error when we don't receive PublishResponse.  A new PublishRequest is send when we receive the PublishResponse (when ```backgroundWaitingPublishResponse``` is FALSE).

2) No because we must also call : 

```
    UA_PublishRequest_deleteMembers(&client->backgroundPublishRequest);
    client->backgroundWaitingPublishResponse = UA_FALSE;
```

Do you understand/agree ?

StalderT:
for example we can replace backgroundWaitingPublishResponse with publishRequestsInTransit

WoutvdBerg:
> In the example the monitored items are added before the publish loop. But you can add/remove monitored items between two calls of runAsync. Do you need another functionnality?

Async adding and deleting of monitoring items would be nice because when a lot of monitored items are added and the response is taking a while all other requests can continue.

jpfr:
The code is fine for what it does. But now is the time to make the subscription model correct.
We are *nearly* there.

You have to count how many publish requests are currently "sleeping" in the server.
So you can replenish when you get a response.
Many scenarios have many (around 10) requests "sleeping" in the server.

> backgroundPublishResponseTimeout is used to generate error when we don't receive PublishResponse. 

What this timeout should be is already determined by the subscription configuration.
This is negotiated during createSubscription.

Imho we only need an additional counter `UA_UInt16 outStandingPublishRequests` in the client-config.
And an additional counter `UA_UInt16 currentlyOutStandingPublishRequests` in the client.
**Nothing else is required to be stored explicitly in the client.**

>     No because we must also call :
```
    UA_PublishRequest_deleteMembers(&client->backgroundPublishRequest);
    client->backgroundWaitingPublishResponse = UA_FALSE;
```

See above.
Can we move everything to the general case in `UA_Client_processPublishResponse`?
Just call `UA_Client_processPublishResponse` with the async result.
(Maybe that needs a slight adjustment of the `UA_Client_processPublishResponse` signature. To use it as an async callback.)
If it's not possible, okay.

This stuff is complicated enough. We should try to have all moving parts in one place.

StalderT:
Ok I understand. 

Can you explain the advantage to have many PublishRequest sleeping in the server?

The timeout is needed to detect problem like network link failure. The negotiated timeout during createsubcription works well when all goes fine, but if we have a network link failure we never detect the problem. We need to keep the timeout.

Actually, async callback don't transfert request to the callback function. Many services need to have this to treat the reponse. If we want to have many simultaneous PublishRequest, we need to store the request to pass them to the callback function.

What do you think?

jpfr:
> Can you explain the advantage to have many PublishRequest sleeping in the server?

The server can send publish responses right away after the publication interval.
And if there are more than `maxNotificationsPerPublish` notifications, several responses are sent right away.

> The timeout is needed to detect problem like network link failure. The negotiated timeout during createsubcription works well when all goes fine, but if we have a network link failure we never detect the problem. We need to keep the timeout.

Please note that if `maxKeepAliveCount` is high and there are no notifications, it can take a long time (many publication intervals) until a response comes.

> Actually, async callback don't transfert request to the callback function. Many services need to have this to treat the reponse. If we want to have many simultaneous PublishRequest, we need to store the request to pass them to the callback function.

Right. You can just attach it as a pointer to the `AsyncServiceCall` structure.

StalderT:
Changes done !

Now I have two questions/problem : 

----------

1) I don't know how to deal with subscription Acknowledgements with multiple publishrequest. The current implementation generate BadSequenceNumberUnknow.

image

Have you an idea?

----------

2) After a network link failure, the client create a new securechannel and reactivate the session (no new publishrequest are send to the server). 

After that, with open62541 server all works fine, the server send a publishresponse for a pending publishrequest.

If I try with opc.tcp://opcua.demo-this.com:51210/UA/SampleServer, the server don't send a publishresponse.

The OPC specification say : 

image

If I good understand, open62541 server should not automatically send a publishresponse after reactivating a session, because the client should call republish service.

Currently, client part don't implement republish service.

We need : 
1) Fix the server part and fix the client
or
2) Fix the server and disable the ability in the client part to reactivate a session.

To fix the server we can add a field in the UA_PublishResponseEntry structure, like UA_Boolean waitRepublish and set this field to TRUE in Service_ActivateSession (for all entry in the list) and modify UA_Session_getPublishReq to only return request with waitRepublish set to FALSE.

--------

What do you think?

StalderT:
I found the way to deal with Acknowledgements.

Reactivate a session is very no trival. I think is better to desactivate this possibilty in order to have most robust client but with less functionnality.

Do you agree?

jpfr:
> Reactivate a session is very no trival. I think is better to desactivate this possibilty in order to have most robust client but with less functionnality.

👍 for robustness and clean architecture in the first step.
Can we document the current state of session recovery in the docs and matching TODOs in the code?

StalderT:
@jpfr modifications done! I have added TODOs in the code. If the PR is ok for you I will add some more unit tests.

Thanks for your review.

StalderT:
@WoutvdBerg 
> Async adding and deleting of monitoring items would be nice because when a lot of monitored items are added and the response is taking a while all other requests can continue.

According to the README, it is not planned to add the asynchronous functions in version 0.3 but in version 0.4. You can use the UA_Client_Subscriptions_addMonitoredItems function to add several items at once. 

In order to delete several items at once, we can also create the function UA_Client_Subscriptions_removeMonitoredItems. I will works on it.

basyskom-jvoe:
Many thanks for working on the issue!
I would be nice if the async publish could be part of the 0.3 release as we are planning to use 0.3 for the qtopcua open62541 plugin.

What would be the best way to use async publish in combination with intermittent synchronous service calls? My first idea would be to call UA_Client_runAsync with a short timeout and run queued synchronous operations before the next call to UA_Client_runAsync.

StalderT:
>What would be the best way to use async publish in combination with intermittent synchronous service calls? My first idea would be to call UA_Client_runAsync with a short timeout and run queued synchronous operations before the next call to UA_Client_runAsync.

Currently I think so. A nice feature would be to be able to "wake up" UA_Client_runAsync. @jpfr Do you have an idea / proposal?

basyskom-jvoe:
I have just built your branch and modified our plugin to use async publish with outStandingPublishRequests = 5 and a timeout of 1 ms for UA_Client_runAsync. UA_Client_runAsync is triggered by a QTimer whenever there are no pending synchronous operations.

Everything works as expected, my test application has become much more responsive as there is no more waiting in UA_Client_manuallySendPublishRequest before e. g. a write or read request is sent.

StalderT:
> Everything works as expected, my test application has become much more responsive as there is no more waiting in UA_Client_manuallySendPublishRequest before e. g. a write or read request is sent.

Thank you for testing and feedback.

src/client/ua_client_highlevel_subscriptions.c
```diff
@@ -510,4 +568,51 @@ UA_Client_Subscriptions_clean(UA_Client *client){
         UA_Client_Subscriptions_forceDelete(client, sub); /* force local removal */
 }
 
+UA_StatusCode
+UA_Client_Subscriptions_backgroundPublish(UA_Client *client) {
+    if (client->state < UA_CLIENTSTATE_SESSION)
+        return UA_STATUSCODE_BADSERVERNOTCONNECTED;
+
+    if (client->config.outStandingPublishRequests == 0)
```
basyskom-jvoe: There should also be a check if the client has any subscriptions. If there are no subscriptions, the server gets spammed with publish requests when UA_Client_runAsync() is called.

StalderT: Thanks. Done.

StalderT:
All green. For me ready to merge.

jpfr:
> What would be the best way to use async publish in combination with intermittent synchronous service calls? My first idea would be to call UA_Client_runAsync with a short timeout and run queued synchronous operations before the next call to UA_Client_runAsync.

The synchronous calls work as follows:
- The synchronous call has a timeout and a request id
- When the response with the correct request id arrives, return the response
- If the timeout is exceeded, return as well
- If another response comes in, see if we have an asynchronous callback stored for that. If yes execute, otherwise log an error and discard the response

I don't understand what "queued synchronous operations" are in that context.
Most things can probably fit into the existing process. For example, when we can decide whether to send a new publish request inside the internal (asynchronous) processing callback for the publish response, 
If we want to have timed events in the client, we can add the timer mechanism (`/src/ua_timer.h`) that was already made independent from the server code.

On a second note, the standard defines two error codes that the publish service can return:
- Bad_TooManyPublishRequests  The Server has reached the maximum number of queued Publish requests.
- Bad_NoSubscription  There is no Subscription available for this session.

We should react to them, so that we don't generate empty overhead.
E.g. by reducing the queue size we try to fill up.

jpfr:
The addition of removeMonitoredItems was moved to a separate PR: #1500.

jpfr:
The server-side changes are moved to a dedicated commit.

jpfr:
I will take the liberty and roll back the addition of the request to the `__UA_Client_AsyncService`.
This has two reasons.

- We want to be able to stack-allocate the request. It is even dangerous to keep a pointer the request then!
- If the user wants to access the request (and knows he has heap-allocated it), he can simply put a pointer into the userdata.

jpfr:
The API is now reverted to not forward the request explicitly into the callback.
That was surprisingly easy to make. And I haven't found drawbacks or impossible use cases.
All the unit tests were untouched and are green.

Can you verify the changes?
We are ready then. 👍 

Thank you for the huge effort @StalderT ! This is a huge improvement!

basyskom-jvoe:
> I don't understand what "queued synchronous operations" are in that context.
Most things can probably fit into the existing process. For example, we can decide whether to send a new publish request inside the internal (asynchronous) processing callback for the publish response,
If we want to have timed events in the client, we can add the timer mechanism (/src/ua_timer.h) that was already made independent from the server code.

We use UA_Client as part of a Qt object running in a separate thread. Read/write/call/browse operations are submitted by the user from the main thread. They are stored in a queue. As soon as an operation using UA_Client has finished, the next one from the queue is executed. QOpcUaClient is a generic client, so we don't know in advance what the user is going to do. We can't assume that all monitored items are added after connect and the rest of the runtime, there's just publish requests and publish responses. An user could want to read a variable, call a method or add another monitored item anytime.

If there are any subscriptions, the publishing mechanism must be allowed to run by calling UA_Client_runAsync(). At the same time, we want to keep the delay between the submission of an operation and its execution as short as possible. Our current solution is running UA_Client_runAsync() for a short time whenever our queue is empty. If there are other operations in our queue they are executed before the next call to UA_Client_runAsync().
I assume that this approach won't be a problem as long as the next call to UA_Client_runAsync() (which sends new publish requests by calling UA_Client_Subscriptions_backgroundPublish()) is executed before the lifetime of our subscriptions runs out.

Let's assume a synchronous operation (e. g. browse for a large folder) which blocks UA_Client_Service() for nearly the whole timeout. While UA_Client_Service() is waiting for the BrowseResponse, publish responses are received and processed, but no new publish requests are sent. If there is a subscription configured with a publishing interval and lifetime count low enough to run out while UA_Client_Service is waiting for the BrowseResponse, the subscriptions will be deleted by the server.
Could this be avoided by calling UA_Client_Subscriptions_backgroundPublish() inside the function which processes the incoming publish request?

jpfr:
The integration of OPC UA into a GUI application with the "main" loop outside of our control is the litmus test our API has to pass. Very good you are looking into this!

> I assume that this approach won't be a problem as long as the next call to UA_Client_runAsync() (which sends new publish requests by calling UA_Client_Subscriptions_backgroundPublish()) is executed before the lifetime of our subscriptions runs out.

Yes.

> Could this be avoided by calling UA_Client_Subscriptions_backgroundPublish() inside the function which processes the incoming publish request?

Already implemented. Once we process a publish response, we reduce the number of outstanding requests and "fill up" if necessary.

StalderT:
>>  Could this be avoided by calling UA_Client_Subscriptions_backgroundPublish() inside the function which processes the incoming publish request?

> Already implemented. Once we process a publish response, we reduce the number of outstanding requests and "fill up" if necessary.

I think it's not a good idea, I the server always respond (stress, many items, etc..), we can recursively loop and generate stack error. Must be removed.

With the last commit, response are never freed. Implementation is incorrect.

Here : https://github.com/StalderT/open62541/blob/9182ede318cd4defebd14290965aee4850fb571f/src/client/ua_client_highlevel_subscriptions.c#L474, currentlyOutStandingPublishRequests must only be decrease if PublishResponse are received by tha asyncronous task (not by UA_Client_Subscriptions_manuallySendPublishRequest)

jpfr:
> I think it's not a good idea, I the server always respond (stress, many items, etc..), we can recursively loop and generate stack error. Must be removed.

We don't blow up the stack. We only send out requests to fill up.
But we won't recurse into processing new publish responses.

I agree that we should react to server errors. See my comment on Bad_TooManyPublishRequests and Bad_NoSubscription.

> With the last commit, request are never freed.

Right. This needs to be changed to a `UA_PublishRequest_delete`.
https://github.com/StalderT/open62541/blob/9182ede318cd4defebd14290965aee4850fb571f/src/client/ua_client_highlevel_subscriptions.c#L517

> currentlyOutStandingPublishRequests must only be decrease if PublishResponse are received by tha asyncronous task

Right. We should then increase currentlyOutStandingPublishRequests inside UA_Client_Subscriptions_manuallySendPublishRequest. This also prevents us from sending one request too many when we wait for the response and process other publish responses.

StalderT:
The last commits seems good! Thank you !!!

> I agree that we should react to server errors. See my comment on Bad_TooManyPublishRequests and Bad_NoSubscription.

Bad_TooManyPublishRequests : we must decrease client->config.outStandingPublishRequests and block new publishrequest until we receive a publishresponse

```
OPC Unified Architecture, Part 4
5.13.5.1
Bad_TooManyPublishRequests. If a Client receives this Service result for a Publish request it shall not issue another Publish request before one of its outstanding Publish requests is returned from the Server.
```

Bad_NoSubscription : fatal error ? closing the connection ?

jpfr:
> we must decrease client->config.outStandingPublishRequests and block new publishrequest until we receive a publishresponse

Just doing the first could be enough.

Okay. I pushed some last changes. I will now stop developing on this branch so that we don't duplicate efforts. Feel free to continue improving and just tell me when you are happy with the branch.

This is already pretty solid. And way better then what we had.

jpfr:
> Bad_NoSubscription : fatal error ? closing the connection ?

Closing seems a bit extreme. Maybe remove all remaining subscriptions from the session?
And log a warning of course.

StalderT:
@jpfr Do you know how to fix this error? : https://travis-ci.org/open62541/open62541/jobs/328087157#L1586

jpfr:
Sorry, didn't test with fuzzing activated.
Just copy this line into the CMakeLists.txt that runs the fuzzer.
https://github.com/StalderT/open62541/blob/f67662c027fa9d6ebbc9cb8746731732a3cacecd/tests/CMakeLists.txt#L51

StalderT:
now sample client crash

```
1 endpoints found
URL of endpoint 0 is opc.tcp://localhost:4840
Browsing nodes in objects folder:
NAMESPACE NODEID           BROWSE NAME      DISPLAY NAME    
0         61               FolderType       FolderType      
0         2253             Server           Server          
1         the.answer       the answer       the answer      
0         50189            current time     current time    
1         50000            Demo             Demo            
1         62541            hello_world      Hello World     
0, 85 --- 40 ---> NodeId 0, 61
0, 85 --- 35 ---> NodeId 0, 2253
0, 85 --- 35 ---> NodeId 1, 10
0, 85 --- 35 ---> NodeId 0, 50189
0, 85 --- 35 ---> NodeId 1, 50000
0, 85 --- 47 ---> NodeId 1, 62541
Create subscription succeeded, id 1
Monitoring 'the.answer', id 1
The Answer has changed!

Reading the value of node (1, "the.answer"):
the value is: 46

Writing a value of node (1, "the.answer"):
the new value is: 47
[2018-01-12 15:12:02.741 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.749 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.750 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.750 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.750 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.751 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.751 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.751 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.752 (UTC+0100)] error/client	No subscription we remove all subscriptions
[2018-01-12 15:12:02.752 (UTC+0100)] error/client	No subscription we remove all subscriptions
==34705== Invalid read of size 8
==34705==    at 0x11B5A1: UA_Client_Subscriptions_forceDelete (ua_client_highlevel_subscriptions.c:105)
==34705==    by 0x11B585: UA_Client_Subscriptions_remove (ua_client_highlevel_subscriptions.c:97)
==34705==    by 0x112F40: main (client.c:153)
==34705==  Address 0x5a62478 is 56 bytes inside a block of size 64 free'd
==34705==    at 0x4C2CDDB: free (vg_replace_malloc.c:530)
==34705==    by 0x11B650: UA_Client_Subscriptions_forceDelete (ua_client_highlevel_subscriptions.c:111)
==34705==    by 0x11CBE5: UA_Client_Subscriptions_clean (ua_client_highlevel_subscriptions.c:606)
==34705==    by 0x11C716: processPublishResponse (ua_client_highlevel_subscriptions.c:490)
==34705==    by 0x11C971: processPublishResponseAsync (ua_client_highlevel_subscriptions.c:544)
==34705==    by 0x1162FC: processAsyncResponse (ua_client.c:144)
==34705==    by 0x11650B: processServiceResponse (ua_client.c:197)
==34705==    by 0x124A99: UA_SecureChannel_finalizeChunk (ua_securechannel.c:772)
==34705==    by 0x125258: UA_SecureChannel_processChunk (ua_securechannel.c:1040)
==34705==    by 0x11672E: client_processChunk (ua_client.c:250)
==34705==    by 0x1226F6: completeChunkTrampoline (ua_connection.c:187)
==34705==    by 0x1225E7: processChunk (ua_connection.c:139)
==34705==  Block was alloc'd at
==34705==    at 0x4C2BBAF: malloc (vg_replace_malloc.c:299)
==34705==    by 0x11B252: UA_Client_Subscriptions_new (ua_client_highlevel_subscriptions.c:30)
==34705==    by 0x112B48: main (client.c:97)
==34705== 
```

Because when we remove a subscription, we call UA_Client_Subscriptions_remove. UA_Client_Subscriptions_remove call UA_Client_Service_deleteSubscriptions and receive the publishresponse with UA_STATUSCODE_BADNOSUBSCRIPTION who remove subscriptions. After the call of UA_Client_Service_deleteSubscriptions, the subscription is invalid and generate the crash.

Do you have any idea how to properly handle UA_STATUSCODE_BADNOSUBSCRIPTION in publishresponse? Log and do nothing?

StalderT:
It was not easy to put everything in green but now it's ok. Sorry to have a little "pollute" the PR with many commits but it was for the good cause :-)

One last question : 

According to the specification : 

removesub

MonitoredItems are deleted on the server when a subscription is removed.

Currently, the client manually remove the MonitoredItems (in the server) before remove a subscription here : 

https://github.com/open62541/open62541/blob/d86f981317b401e05e97803bda60dc25e7261f68/src/client/ua_client_highlevel_subscriptions.c#L70-L77

To optimize, can we remove theses lines? The MonitoredItems are removed form the client when we receive the DeleteSubscritpions response (already done).

jpfr:
> To optimize, can we remove theses lines?

👍

StalderT:
@jpfr can you merge https://github.com/open62541/open62541/pull/1500, then I could merge master to this branch?  Thanks in advance.

jpfr:
#1500 won't complete on travis. But the logs are looking good.
If nothing helps, we can directly merge this PR, as all commits are included.

StalderT:
fix https://github.com/open62541/open62541/issues/511

StalderT:
Ok to merge?

jpfr:
I'm on it. This is getting into a full-blown API refactor.
The old API is kept for compatibility...

basyskom-jvoe:
This is exactly what we need, will this feature be available in v0.3?

jpfr:
> This is exactly what we need, will this feature be available in v0.3?

Yes! I have very little capacity atm. Can have look in a few days.
Anyone of @open62541/open62541-core wants to champion the PR to get it in sooner?

StalderT:
Maybe we can put the new API in a separate PR and merge this PR?

jpfr:
Yes. Let's do that.

StalderT:
Ready to merge ?

StalderT:
Apparently not

jpfr:
Rebased on master there were a ton of conflicts.
The new keepalive test hangs with async..

StalderT:
> The new keepalive test hangs with async..

Found why and fixed!

jpfr:
> Found why and fixed!

Great! I wouldn't have though of that.

The last commit adds "minimal but complete" API with respect to the Subscription and MonitoredItem service sets. `ua_client_subscriptions.c` was nearly halved in size.
We can recreate the old API (as deprecated) and convenience methods (e.g. to store and retrieve the current parameters) transparently on top.

src/client/ua_client_subscriptions.c
```diff
@@ -0,0 +1,687 @@
...
+    /* Find the internal representation */
+    UA_Client_Subscription *sub = findSubscription(client, request.subscriptionId);
+    if(!sub) {
+        response.responseHeader.serviceResult = UA_STATUSCODE_BADSUBSCRIPTIONIDINVALID;
+        return response;
+    }
+    
+    /* Call the service */
+    __UA_Client_Service(client,
+                        &request, &UA_TYPES[UA_TYPES_MODIFYSUBSCRIPTIONRESPONSE],
```
basyskom-jvoe: &UA_TYPES[UA_TYPES_MODIFYSUBSCRIPTIONREQUEST] must come first. Using UA_TYPES_MODIFYSUBSCRIPTIONRESPONSE as type of the request kills the session.

StalderT: Thanks! Fixed.

basyskom-jvoe: Thanks for the fix, my unit tests are all green now.

basyskom-jvoe:
Are you planning to add single versions of UA_Client_Subscriptions_setPublishingMode() and UA_Client_MonitoredItems_setMonitoringMode() for convenience?

StalderT:
@jpfr I can not solve the latest clang errors. Could you take a look at it?

StalderT:
All lights are green.

@jpfr can you change examples/tutorial_client_events.c (currently disabled by a #if 0) in ordre to use the new API?

StalderT:
> Are you planning to add single versions of UA_Client_Subscriptions_setPublishingMode(), UA_Client_MonitoredItems_setMonitoringMode() and UA_Client_MonitoredItems_modify() for convenience?

For my part, not for the moment. You can create an issue or a PR for this.

jpfr:
> Are you planning to add single versions of UA_Client_Subscriptions_setPublishingMode(), UA_Client_MonitoredItems_setMonitoringMode() and UA_Client_MonitoredItems_modify() for convenience?

That should be easy to do. For example as static inline methods in the header.
Let's merge this PR first.

StalderT:
Currently sample client displays : 

```
1 endpoints found
URL of endpoint 0 is opc.tcp://localhost:4840
Browsing nodes in objects folder:
NAMESPACE NODEID           BROWSE NAME      DISPLAY NAME    
0         61               FolderType       FolderType      
0         2253             Server           Server          
1         the.answer       the answer       the answer      
0         50189            current time     current time    
1         50000            Demo             Demo            
1         62541            hello_world      Hello World     
0, 85 --- 40 ---> NodeId 0, 61
0, 85 --- 35 ---> NodeId 0, 2253
0, 85 --- 35 ---> NodeId 1, 10
0, 85 --- 35 ---> NodeId 0, 50189
0, 85 --- 35 ---> NodeId 1, 50000
0, 85 --- 47 ---> NodeId 1, 62541
Create subscription succeeded, id 1
Monitoring 'the.answer', id 1
The Answer has changed!

Reading the value of node (1, "the.answer"):
the value is: 46

Writing a value of node (1, "the.answer"):
the new value is: 47
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
Subscription removed
Method call was successful, and 1 returned values available.
```

When we remove a subscription we must remove the subscriptions from the client list before sending the request to the server to avoid : 

```
[2018-02-17 12:59:21.191 (UTC+0100)] warn/client	Received Publish Response with code BadNoSubscription
```

StalderT:
@jpfr Hello, I finished working on the PR. I let you finalize it. Thank you.

jpfr:
I cleaned up the history a bit.
The final files are identical to before the rebase.

You are Co-Author of the "new API" commit to give credit for the incredible bugfixing. 👍 

We can merge when the checks have run.

StalderT:
#if 0 was removed for examples/tutorial_client_events.c

Just need to change ```UA_Client_Subscriptions_addMonitoredEvent``` to the new API

jpfr:
The events tutorial gives me valgrind issues.
Somehow related to shutting down the client network layer after ctrl-c.

I'm on it...

StalderT:
> The events tutorial gives me valgrind issues.
Somehow related to shutting down the client network layer after ctrl-c.

Same result with master :-(

StalderT:
> The events tutorial gives me valgrind issues.
Somehow related to shutting down the client network layer after ctrl-c.

I have the solution

StalderT:
But that does not explain why connection_write is called while the connection is closed.

jpfr:
Found it.
The SIGINT signal (from ctrl-c) caused the `select` call to abort.
Which was interpreted as a closed connection.

Now let's wait for green.

StalderT:
@jpfr 

Maybe we can also add : 

```
    if(channel->connection) {
        if (channel->connection->state == UA_CONNECTION_CLOSED)
            return UA_STATUSCODE_BADCONNECTIONCLOSED;
    }
```

at the at the beginning of ```UA_SecureChannel_sendSymmetricMessage``` 

What do you think?

jpfr:
> What do you think?

👍

StalderT:
@jpfr 

with https://github.com/open62541/open62541/pull/1497/commits/ea10cf8a68a3865b4bac59eb193088a75160bafd, how can the client detect an session inactivity (connection lost / network link failure)?

StalderT:
> with ea10cf8, how can the client detect an session inactivity (connection lost / network link failure)?

Can we add a callback for that ?

jpfr:
There are different timeouts on the session and securechannel level.

jpfr:
Nicely done. Thanks for the huge contribution!

StalderT:
Thank you too.

basyskom-jvoe:
Thank you very much for that great improvement!

PR #1508, opened 2018-01-15
Client timer

jpfr:
Follow up for #1497.
Implements timed callbacks for the client.
Securechannel and session keepalive is intended to be moved over here.


StalderT:
👍

StalderT:
Can we add something like «getfd» to the connection? Then the client can use select to wait an event on the socket or on a external event. If you agree I can work on it.

jpfr:
The connection struct already contains that.
https://github.com/open62541/open62541/blob/master/include/ua_plugin_network.h#L63

So this is not the preferred way to drive clients. But on systems with mutliple clients, it is advantageous to `select` on multiple sockets at once. Hence the connection is exposed to power-users.

StalderT:
Thank you for your reply. I did not pay attention to the fact that the content of the structure was publicly available. Sorry!

src/client/ua_client.c
```diff
@@ -354,11 +421,78 @@ __UA_Client_AsyncService(UA_Client *client, const void *request,
...
+                *nextTimeout = UA_MAXTIMEOUT;
+            else
+                *nextTimeout = (UA_UInt16)nextCall;
+        }
+    }
+
     return retval;
 }
+
+/* A worker thread could be spliced into here... */
```
sherylll: Work thread already implemented in my fork. Will merge that

Pro:
As discussed in the telco today, this code is completely outdated and partially included in master.

PR #1512, opened 2018-01-17
Async highlevel functions with client worker

sherylll:
A working example is given by client_async.c
The UA_Client_connect() function (from the upstream master) works in this example, however my implementation of the connect function is different in that I treated it similarly as with other high-level functions, which looks like this:
```
UA_Client_connect_async(client, "opc.tcp://localhost:4840", onConnect,
			&connected); 
//connected is set to true once the client is connected, onConnect is a user callbck
```
In my implementation openSecureChannel is triggered by an ACK from the hello message, and Session is triggered by the completion of openSecureChannel, and so forth. I will see if I can combine the two.


jpfr:
Can you rebase on the current master and remove dead files?
For example all that eclipse stuff.

We should strip the PR to the bare minimum.
We can also merge some changes that are not directly related to async connect ahead of time.

jpfr:
By the way, after such a long development, it makes often sense to clean up the history with `git rebase -i`.

sherylll:
Oh I overlooked the eclipse files, will fix it later today

Infinity95:
The eclipse files are always put in the same directory, right? If so you could add that to the .gitignore as well to avoid adding any such stuff in the future.

Infinity95:
You can reduce a lot of noise in this PR by changing back the tabs to spaces. A lot of changes are simply because of auto reformatting i believe. Since we use 4 spaces as indentation in the rest of the library it would probably be best if you changed that in your ide settings and the modified files.

Root-Core:
As mentioned in #1117 it would be nice to have at least an asynchronous version of `UA_Client_connect()` in 0.3. 

There are some conflicts and as @sherylll said, there might be a problem with building his code. What is missing in the PR right now?

sherylll:
replaced by #1598

PR #1506, opened 2018-01-15
add InfoBits Overflow

StalderT:
fix https://github.com/open62541/open62541/issues/1505

PR #1517, opened 2018-01-18
add shrinkMonitoredItemQueue and a test

StalderT:
fix https://groups.google.com/forum/#!msg/open62541/5LylGYaew_Y/xaPxHtn8AQAJ

see https://github.com/open62541/open62541/pull/1506

### DIFF
--- a/src/server/ua_services_subscription.c
+++ b/src/server/ua_services_subscription.c
@@ -330,6 +330,9 @@ Operation_ModifyMonitoredItem(UA_Server *server, UA_Session *session, UA_Subscri
     setMonitoredItemSettings(server, mon, mon->monitoringMode, &request->requestedParameters);
     result->revisedSamplingInterval = mon->samplingInterval;
     result->revisedQueueSize = mon->maxQueueSize;
+
+    /* Remove some notifications if the queue is now too small */
+    MonitoredItem_ensureQueueSpace(mon);
 }
 
 void Service_ModifyMonitoredItems(UA_Server *server, UA_Session *session,
--- a/src/server/ua_subscription.h
+++ b/src/server/ua_subscription.h
@@ -63,6 +63,10 @@ void UA_MoniteredItem_SampleCallback(UA_Server *server, UA_MonitoredItem *monito
 UA_StatusCode MonitoredItem_registerSampleCallback(UA_Server *server, UA_MonitoredItem *mon);
 UA_StatusCode MonitoredItem_unregisterSampleCallback(UA_Server *server, UA_MonitoredItem *mon);
 
+/* Remove entries until mon->maxQueueSize is reached. Sets infobits for lost
+ * data if required. */
+void MonitoredItem_ensureQueueSpace(UA_MonitoredItem *mon);
+
 /****************/
 /* Subscription */
 /****************/
--- a/src/server/ua_subscription_datachange.c
+++ b/src/server/ua_subscription_datachange.c
@@ -47,30 +47,48 @@ MonitoredItem_delete(UA_Server *server, UA_MonitoredItem *monitoredItem) {
     UA_free(monitoredItem); // TODO: Use a delayed free
 }
 
-static void
-ensureSpaceInMonitoredItemQueue(UA_MonitoredItem *mon, MonitoredItem_queuedValue *newQueueItem) {
-    /* Enough space, nothing to do here */
-    if(mon->currentQueueSize < mon->maxQueueSize)
+void
+MonitoredItem_ensureQueueSpace(UA_MonitoredItem *mon) {
+    UA_Boolean valueDiscarded = false;
+    MonitoredItem_queuedValue *queueItem;
+#ifndef __clang_analyzer__
+    while(mon->currentQueueSize > mon->maxQueueSize) {
+        /* maxQueuesize is at least 1 */
+        UA_assert(mon->currentQueueSize >= 2);
+
+        /* Get the item to remove */
+        if(mon->discardOldest) {
+            /* Remove the last */
+            queueItem = TAILQ_LAST(&mon->queue, QueuedValueQueue);
+        } else {
+            /* Keep the first, remove the second */
+            queueItem = TAILQ_FIRST(&mon->queue);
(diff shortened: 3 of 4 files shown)


StalderT:
Someone has an idea why we have these warnings : https://travis-ci.org/open62541/open62541/jobs/330310162#L1673 ?

Pro:
To further investigate clang static analyzer errors/warnings the best solution is to run the clang scan build locally. It has a very nice HTML view of the reported error.

i.e. install clang-3.9 on linux
then:

```
cd open62541
mkdir build-scan
cd build-scan
scan-build-3.9 cmake -DUA_BUILD_EXAMPLES=ON -DUA_BUILD_UNIT_TESTS=ON ..
scan-build-3.9 -enable-checker security.FloatLoopCounter -enable-checker security.insecureAPI.UncheckedReturn --status-bugs -v  make -j
```

It then tells you something like:
```
scan-build: Run 'scan-view /tmp/profanter/scan-build-2018-01-18-134415-6911-1' to examine bug reports.
```
(I had to use `scan-view-3.9` as command)

For your convenience I add the output as a screenshot here:

image


As you can see, in line 341 (Point 7a) the memory of `mon->queue` is released, then the while loop stopped, and then on line 350 you get the deleted item, which will fail with a SegFault on line 352

StalderT:
@Pro  Wow, thanks for the explanations 👍 

But I still do not understand why there is this warning. I call ```TAILQ_REMOVE(&mon->queue, queueItem, listEntry);``` then at line 350 I get a new item, not the deleted ?

Pro:
It could also be that the warning is a false positive, i.e., the static analyzer does not understand the TAILQ_REMOVE macro entirely. The probability for that is although not so high... But before you tell clang that it is fine what you are doing and you want to ignore the warning, you should make 1000% sure that there is no case at all where you access the deleted item.

StalderT:
https://www.mail-archive.com/misc@opensmtpd.org/msg00529.html : 

image

Pro:
Ok, as I thought...
Then you need to tell clang that it is fine what you are doing, i.e. mark this specific line as false positive and add a corresponding comment why it is false positive

StalderT:
for me can be merged

jpfr:
Hey @StalderT ,

I unified your changes with similar logic that existed in `ua_subscription_datachange.c`.
Can you confirm? We are then ready to merge.

StalderT:
Hey @jpfr ,

The new item is inserted at the end of the queue : 
```
    /* Add the sample to the queue for publication */
    TAILQ_INSERT_TAIL(&monitoredItem->queue, newQueueItem, listEntry);
```

```
        /* Get the item to remove */
        if(mon->discardOldest) {
            /* Remove the last */
            queueItem = TAILQ_LAST(&mon->queue, QueuedValueQueue);
        } else {
            /* Keep the first, remove the second */
            queueItem = TAILQ_FIRST(&mon->queue);
            queueItem = TAILQ_NEXT(queueItem, listEntry);
}
```
----------------

1) When ```MonitoredItem_ensureQueueSpace``` is called by ```sampleCallbackWithValue```

When ```mon->discardOldest==true``` we need to remove the first item

When ```mon->discardOldest==false``` we need to remove the last but one item

----------------

2) When ```MonitoredItem_ensureQueueSpace``` is called by ```Operation_ModifyMonitoredItem```

When ```mon->discardOldest==true``` we need to remove the first item

When ```mon->discardOldest==false``` we need to remove the last item

If the queue size is reduce to 1, we must also make sure to remove the eventual status of the remaining item.

jpfr:
> The new item is inserted at the end of the queue :

Hrmpf. My bad.

> If the queue size is reduce to 1, we must also make sure to remove the eventual status of the remaining item.

The standard says: If a Notification is discarded for a DataValue and the size of the queue is larger than one, then the Overflow bit (flag) in the InfoBits portion of the DataValue statusCode is set.

Does the standard say anything about removing infobits? Currently we won't add new infobits when the queue size is 1. So remaining infobits indicate lost information before the queue was resized to 1.

StalderT:
> Does the standard say anything about removing infobits? Currently we won't add new infobits when the queue size is 1. So remaining infobits indicate lost information before the queue was resized to 1.

I don't know but : https://groups.google.com/forum/#!topic/open62541/5LylGYaew_Y

jpfr:
Better?

StalderT:
> Better?

Yes. Thank you!

One last problem: ```MonitoredItem_ensureQueueSpace``` should remove the newest and not remove the second-newest when called by ```Operation_ModifyMonitoredItem```.

Maybe we can add a parameter to ```MonitoredItem_ensureQueueSpace``` :

```
void
MonitoredItem_ensureQueueSpace(UA_MonitoredItem *mon) {
                               UA_Boolean recentlyItemAdded) {
```

```
            queueItem = TAILQ_LAST(&mon->queue, QueuedValueQueue);
            if(recentlyItemAdded) {
                /* Keep the newest, remove the second-newest */
                queueItem = TAILQ_PREV(queueItem, QueuedValueQueue, listEntry);
            }
```

```sampleCallbackWithValue``` can call ```MonitoredItem_ensureQueueSpace``` with ```recentlyItemAdded``` to ```true``` and ```Operation_ModifyMonitoredItem``` can call ```MonitoredItem_ensureQueueSpace``` with ```recentlyItemAdded``` to ```false```.

jpfr:
> One last problem: MonitoredItem_ensureQueueSpace should remove the newest and not remove the second-newest when called by Operation_ModifyMonitoredItem.

There is a really good explanation required here.
Sadly, ModifyMonitoredItems is not described in much detail.

I understand the standard in the way that the user is always interested in the newest value.
The newest value is always kept. Independent of `discardOldest`, an entry is deleted that was already in the queue before.

This principle seems to be logical also when reducing the queue size.
Maybe we have to ask for clarification on the bugtracker of the OPC Foundation.

StalderT:
> I understand the standard in the way that the user is always interested in the newest value.
The new value is always kept when a new notification is added to the queue.
An entry is deleted that was already in the queue before. Independent of discardOldest.

Explained like that I agree 👍 

Thank you for all your work.

PR #1519, opened 2018-01-18
Various bugfixes detected by fuzzer

Pro:


jpfr:
The circular reference checking is cool!
But it does not yet catch every case.
It could be that the circle does not start at the first iteration.
Say we are 5 references deep. There could be a circular reference to the third one.

I propose we keep a pointer to the previous stack-frame.
```c
struct ref_history {
    struct ref_history *parent; // points to the node_ref in
    const UA_NodeId *id; // the id of the node at this depth
}
```
The we can put ref_history on the stack (use it as a normal variable) and forward the pointer to the recursive call. For checking circles we can then go back in history until we hit a `NULL` parent.

Pro:
Good point!
So for every recursive call to `isNodeInTree` we iterate over the ref history and check if we visited the node already?!
I'll change this.

PR #1521, opened 2018-01-18
Fixed description of some status codes

Root-Core:
The OPC Foundation provides an invalid (non escaped) CSV for status codes. This destroyed some generated status code descriptions.


Pro:
Thanks!
We probably also need to fix it in the script here (merge column 3 and 4, even if it is deprecated:
https://github.com/open62541/open62541/blob/master/tools/generate_statuscode_descriptions.py

Can you also create an issue on OPC Foundation repo to get it fixed, otherwise we may ran into the same issue if we regenerate the status codes in the future.
https://github.com/OPCFoundation/UA-.NET/blob/master/Stack/Core/Schema/Opc.Ua.StatusCodes.csv

Even github complains, when opening that file:
grafik

Root-Core:
Sure: https://github.com/OPCFoundation/UA-.NET/pull/166

PR #1525, opened 2018-01-19
fix loadFile when alloc error

StalderT:


jpfr:
Good find!
The same applies to server_basic128...c

PR #1526, opened 2018-01-19
change ID to Id (in server part)

StalderT:


HolgerJeromin:
I hope this is no public API?

StalderT:
> I hope this is no public API?

No, the public API is not affected. As you can see, all the modified files are located in the ```src/server``` directory. None are in the ```include``` or ```examples``` directories.

PR #1531, opened 2018-01-21
add connectivity check background task 

StalderT:
When #1497 will be merged.

According to the specification, periodically reading the State Variable

image


StalderT:
Can not be a part of 0.3?

Pro:
Yes, it can be, if we manage to merge #1497 within the next two weeks, this one should be easy to merge.

StalderT:
> Yes, it can be, if we manage to merge #1497 within the next two weeks, this one should be easy to merge.

Thanks !

StalderT:
replaced by #1598

PR #1532, opened 2018-01-22
Add clientContext to the state change callback

basyskom-jvoe:
The current version of the client state change callback does not provide a way to pass a context to the callback. This makes access to members of a c++ class containing the callback method difficult.

This patch adds void *clientContext to UA_ClientConfig. The value of clientContext is used as a third parameter in UA_ClientStateCallback.


jpfr:
Cool! 👍 for the addition of a context to the client.
One small change request. You already get the the pointer to the client.
Can we add a method `UA_Client_getContext(UA_Client *client);` to return the context?
Then you don't have to change the API.

This has the added value that all other calbacks for the client also get access to the context.

basyskom-jvoe:
Done.

StalderT:
@jpfr  Just one question : currently open62541 don't check parameters passed to a function.

Eg : 

https://github.com/open62541/open62541/blob/c54a3e26439600a075c5ac427da1aae0f0781655/src/client/ua_client.c#L78-L81

Here open62541 check if client is non NULL : 

https://github.com/open62541/open62541/blob/c54a3e26439600a075c5ac427da1aae0f0781655/src/client/ua_client.c#L85-L86

What do we have to do ?
1) Remove this test?
2) Add this test in all publics functions ?

PR #1536, opened 2018-01-23
Enable CI encryption, temporarily disabled OSX and appveyor latex builds

Pro:
Set `UA_ENABLE_ENCRYPTION=ON` for continuous integration
Enable temporarily disabled CI checks


jpfr:
I tried this.
The Ubuntu version used by Travis does not have `libmbedtls-dev`.
It was still called PolarSSL at the time.
We can build from source...

PR #1537, opened 2018-01-23
Various fuzzer bug fixes

Pro:
See the corresponding commit messages

### DIFF
--- a/src/server/ua_nodes.c
+++ b/src/server/ua_nodes.c
@@ -297,6 +297,12 @@ copyCommonVariableAttributes(UA_VariableNode *node,
      * we need to decode it and set the decoded value instead of the encoded object */
     UA_Boolean valueSet = false;
     if(attr->value.type != NULL && UA_NodeId_equal(&attr->value.type->typeId, &extensionObject)) {
+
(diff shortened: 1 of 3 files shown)


Pro:
@Infinity95 can you please review the first commit?
If an `ActivateSession` call is made two times immediately following each other, the same session will be added two times to the channel:
https://github.com/open62541/open62541/pull/1537/commits/b51b9e8d7e9f769bb2c65b12bb6802d27140da51#diff-43b83dcbe0ed74d0819aa0e6096cd436R277

Which leads to an infinite loop during cleanup:
https://github.com/open62541/open62541/blob/b51b9e8d7e9f769bb2c65b12bb6802d27140da51/src/ua_securechannel.c#L84

Infinity95:
@Pro Isn't that only possible if we use multithreading?

Pro:
> @Pro Isn't that only possible if we use multithreading?
No, a client can simply call the ActivateSession service two times with the exact same data/or the second time with the already activated session.

(found by oss-fuzz)

Infinity95:
Right. This can only happen when using SecurityPolicy#None. Should be fine like this now.

PR #1538, opened 2018-01-23
Add code style editorconfig and clang format hook

Pro:
First draft/idea to add cpplint check to travis

It only can check against google's cpp style guide.

There are currently over 6000 issues, mainly spaces around if, and lines longer 80:

https://justpaste.it/1g3c1

The question is, do we strictly want to be conform to the google cpp style guide:
https://google.github.io/styleguide/cppguide.html

(except some warnings due to C style casts and similar)


HolgerJeromin:
http://editorconfig.org could be interesting for keeping the style

Pro:
> http://editorconfig.org could be interesting for keeping the style
yes, also a good idea! the problem I see here is that there can still be commits which do not adhere the style. And we need to define the editorconfig config file.

jpfr:
tcc 0.9.27 was release five weeks ago.
0.9.26 is four years old.

These are the release notes from the official repo.
http://repo.or.cz/tinycc.git/blob/HEAD:/Changelog
There were a ton of fixes.

I'd run tests. But the problem does not show on my platform (amd64).

Pro:
I just rebased on master to include #1552 

This should significantly reduce the output of cpplint

StalderT:
@Pro 

My contribution to this PR : https://github.com/StalderT/open62541/pull/3

There is only one error left (https://travis-ci.org/StalderT/open62541/jobs/333957899#L1266). But I do not know where this line is generated. If you have an idea? I removed errors for lines longer than 80.

If you want I can make a new PR or else can you integrate the commits to this PR?

tools/travis/travis_linux_script.sh
```diff
@@ -18,6 +18,17 @@ if ! [ -z ${LINT+x} ]; then
...
+        cpplint include/*.h
+	if [ $? -ne 0 ] ; then exit 1 ; fi
+        cpplint plugins/*.c
+	if [ $? -ne 0 ] ; then exit 1 ; fi
+        cpplint plugins/*.h
+	if [ $? -ne 0 ] ; then exit 1 ; fi
+        cpplint examples/*.c
+	if [ $? -ne 0 ] ; then exit 1 ; fi
+        cpplint examples/*.h
+	if [ $? -ne 0 ] ; then exit 1 ; fi
```
Pro: Is there a specific reason why you added the cpp lint calls here?

It would be better to change it here https://github.com/open62541/open62541/pull/1538/commits/4a19333e4d2d6517b7546674f505c1b87eead427#diff-af3b638bc2a3e6c650974192a53c7291R722
so that others can simply run `make cpplint` on their system and it is the same call as on travis.

Pro:
Merging this into master will cause merge conflicts for other PRs, so we should make sure that any bigger PRs are already merged.

tools/travis/travis_linux_script.sh
StalderT: > Is there a specific reason why you added the cpp lint calls here?

No

> It would be better to change it here 4a19333#diff-af3b638bc2a3e6c650974192a53c7291R722
so that others can simply run make cpplint on their system and it is the same call as on travis.

Yes 👍

StalderT: Something like that : https://github.com/StalderT/open62541/pull/3/commits/f1c5e70dd4d58716dfb4ee51ca98d92bdd08aefc ?

The example directory is missing.

jpfr:
The changes are pretty invasive.
@Pro is right that we would have trouble rebasing a lot of work.

There is the possibility to check only the diffs of new commits if they are conformant.
With clang-format, the diff (and only the diff) can be automatically formatted according to style rules.

https://github.com/arduino/ArduinoCore-arc32/wiki/Creating-a-pre-commit-hook-for-clang-format

Pro:
I would also go for the following option:

- Close this PR without merging
- Create an http://editorconfig.org file to ensure the code style for all new commits
- Motivate developers to use the clang pre commit hook (since it is only locally) and add the corresponding docs

Which code style do we want to use? The one checked by cpplint (i.e. Google Code Style)?

jpfr:
Based on this configurator/preview, I think we are closest to the LLVM style.
https://zed0.co.uk/clang-format-configurator/

Pro:
To install the clang format hook use
`ln -s ../../tools/clang-format_precommit_hook .git/hooks/pre-commit`

(note that the path has to be relative to the hooks dir. See https://stackoverflow.com/questions/4592838/symbolic-link-to-a-hook-in-git)

PR #1160, opened 2017-08-16
WIP: initial support for freertos

vatmi:
Draft version supporting freertos with open 62541 using LwiP socket interface.  
Since most of the deafult  socket interfaces were not supported with the used LwiP version 1.4.1 most of them are commented out. The other changes are related to the memory settings in freertos and timing aspects.

This PR is not ready to be merged into master since it changes the default behavior of `tcp_network_tcp.c` in a way that it doesn't work on linux/windows anymore. We need to find a good way to integrate this as a plugin or patch.

Issue #1149, opened 2017-08-10
Example for Embedded System


Gfast2:
Hello open62541,

After firing up the [OPC UA Server examples](https://github.com/open62541/open62541/tree/0.2/examples)  on Beaglebone Black (Raspberry Pi like) and viewing them by [UaExpert](https://www.unified-automation.com/de/produkte/entwicklerwerkzeuge/uaexpert.html), I was very impressed by open62541 project and far more enough motivated  to port this library to my [ESP32](http://esp32.net/) based project. 
Because of its toolchain (freeRTOS + LWIP) and my skill set, I got stucked on implementating open62541 in this enviroment. What I found is only [this issue](https://github.com/open62541/open62541/issues/103) 🏷 in Github, which not tell the story to the end. 🐇 
Because of the trend - _The embedded stuff is getting much performanter then before_. It make more sense to port good libraries like open62541 not only for Linux in its example, but also for typical ones like _ARM M4_, _Arduino_, _ESP32_ and so on. 😃 
Any plan❔

Pro:
Hi @Gfast2,
nice to see that you are also currently working on porting open62541 to freeRTOS + LWIP.

Interestingly, we are currently also trying the same and managed change to code in a way that open62541 compiles, at least in a quick and dirty approach. I will see if we can push it to a repository so that you can also test it and improve it.

Additionally, there is also this PR: #1141

Gfast2:
@Pro ,
*cool*! Compare with MQTT, OPC UA ist still not adopted widely on big IoT Platform like [AWS](https://aws.amazon.com/iot/). I think one of the big problem is its licensing nature, but not its core idea/design. For the much more serious IIoT (Industrial IoT) applications, OPC UA has to be the way of the further.

---

_(Right now I was just drunk by figuring out how many RTOSes out there by [this post](https://www.quora.com/Is-anyone-still-using-a-proprietary-embedded-OS-like-VxWorks-QNX-ThreadX-etc-on-production-systems-or-has-the-world-moved-onto-embedding-Linux-onto-all-platforms))_

---

OK, I'll give my best to tracking the working progress of [#1141](https://github.com/open62541/open62541/pull/1141), hopefully I will do some dedication for it. If I not, I will still leave a eye on this. May be somehow later. 

_ps:_
[...]

PR #1546, opened 2018-01-24
First Changes for freeRTOS

cabralfortiss:
An adaptation of the code to freeRTOS OS using LwIP stack. It was tested with LwIP 2.0.3 and worked partially (the opcua server became unstable after many nodes were added, but the problem was on the heap management, and wans't solved yet). 
Related to pull request #1160 and issue #1149

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,6 +114,9 @@ mark_as_advanced(UA_ENABLE_UNIT_TEST_FAILURE_HOOKS)
 set(UA_VXWORKS_WRS_KERNEL OFF CACHE BOOL "Enable if you want to compile for VxWorks as kernel Module")
 mark_as_advanced(UA_VXWORKS_WRS_KERNEL)
 
+set(UA_FREERTOS OFF CACHE BOOL "Enable if you want to compile for freeRTOS")
+mark_as_advanced(UA_FREERTOS)
+
 # Build options for debugging
 option(UA_DEBUG "Enable assertions and additional functionality that should not be included in release builds" OFF)
 mark_as_advanced(UA_DEBUG)
@@ -234,6 +237,16 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
         remove_definitions(-Werror -Wpedantic -Wno-static-in-inline -fPIC)
         add_definitions(-D_WRS_KERNEL)
     endif()
+    
+    if(UA_FREERTOS)
+       SET(UA_FREERTOS_INCLUDES "" CACHE STRING "Folders to include from the freeRTOS OS")
+       include_directories(${UA_FREERTOS_INCLUDES})
+        # Disable flags for freeRTOS
+        remove_definitions(-fPIC -Wconversion )
+        add_definitions(-DUA_FREERTOS -DLWIP_TIMEVAL_PRIVATE=0 -DLWIP_COMPAT_MUTEX=0 -DLWIP_POSIX_SOCKETS_IO_NAMES=0 -mcpu=cortex-m3 -mthumb -g -Wall -O0 -specs=nano.specs 
+                        -ffunction-sections -fdata-sections  -fno-exceptions -fstack-usage -Wno-unused-variable -Wno-format -Wno-format-security -Wno-format-nonliteral)
+        list(APPEND open62541_LIBRARIES c m stdc++ supc++)
+    endif(UA_FREERTOS)
 
     # Linker
     set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "") # cmake sets -rdynamic by default
--- a/include/ua_config.h.in
+++ b/include/ua_config.h.in
@@ -110,10 +110,23 @@ extern "C" {
 # include <malloc.h>
 #endif
 
-#define UA_free(ptr) free(ptr)
-#define UA_malloc(size) malloc(size)
-#define UA_calloc(num, size) calloc(num, size)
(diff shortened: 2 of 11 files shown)


jpfr:
Nice! This looks good!

We now support Windows/Linux/MacOS/BSDs/vxWorks/QNX/FreeRTOS.

On the other hand, it becomes difficult to understand the nested ifdefs.
The plugins seem to have reached a point where they "break under their own weight".
The API to the server is pretty stable.
Imho we should start implementing architecture ports in separate files.

@cabralfortiss Can you just copy the three plugins you ported into separate files?
These can then be lwip only and remove a lot of the macro-magic.
`/plugins/ua_networklayer_tcp_lwip.c`
`/plugins/ua_networklayer_udp_lwip.c`
`/plugins/ua_clock_freertos.c`

Then cmake can select the correct source file based on the architecture.
@open62541/open62541-core Are you okay if we adapt the source structure in this way?

@cabralfortiss 
Is it possibe to compile for Linux against FreeRTOS as a library or simulator?
Then we can add continuous integration for the FreeRTOS port.

Stasik0:
Any ideas how to get it into ci?

cabralfortiss:
Hi. regarding the new files for the freeRTOS in the plugin folder, I also agree that the defines are getting uglier and uglier. But I noticed that the worst part is in the includes and defines at the top of them. Would be a good idea to have a header file with all \#incluedes and \#defines for each architecture with the same same but in different folders? For example, a plugins/general/ua_arch.h and plugins/freertos/ua_arch.h files with all necessary includes and defines for each architecture. Then, in CMake, include folder plugins/general OR plugins/freertos. And where needed, just included <ua_arch.h>.

plugins/general could also be then subdivided in WIN32 and posix separate folders.

HolgerJeromin:
In the past (2015ish) we had code like this
```
#ifndef UA_free
#define UA_free(ptr) free(ptr)
#endif
```
That way a user can overwrite the memory management without changing our code...
https://github.com/open62541/open62541/commit/e62e5a409fbf62472c76bdb6731a7b887120e718

This looks much cleaner and flexible.

cabralfortiss:
I'm not sure if I understand exaclty what the relation of my comment to yours. I just meant to put all defines and include in a separate ua_arch.h file and include that file. If a new architecture doesn't have a proper ua_arch.h file (for example a missing include, or doesn't have a  ioctl function), the compilation will fail.

HolgerJeromin:
> I'm not sure if I understand exaclty what the relation of my comment to yours.

an architecture does not need to define everything, but only special stuff. This way an architecture could be easier to keep up to date.

cabralfortiss:
Exactly. I meant to put in this new file only what now is already behind the architecture definition (UA_FREERTOS in this case)

Stasik0:
http://www.ovpworld.org/about_continuous_integration

Here a ci proposal

jpfr:
👍 for an ua_arch_freertos.h file.
Maybe we can resurrect the `/ports` folder.

For networking, if there is still a big overlap, is it maybe sensible to split the implementation?
`ua_networking_generic.c`
`ua_networking_posix.c`
`ua_networking_lwip.c`

The clock port is a 100% replacement. So it cold live in a separate file.

cabralfortiss:
I like the idea and the direction where this is going. The cleanest way of doing platform-independent I think is to have all hardware related functions in their own folder. Each folder can have its own CMakefile that adds the files to the list to compile when a variable in CMake selected the platform. So in the main CMakefile you just select the platform on a variable. I'm not sure about copying a file to ${BUILD_DIR}/ua_arch.h since I never did that in CMake, but I find #include <ua_arch.h> easier. 

Since the port change I think will be a big one, maybe it would be better to have in another branch (I think), so what should I do with this pull request? I just want to add that I have another commit for cleaning before this request is merged.

jpfr:
>  I'm not sure about copying a file to ${BUILD_DIR}/ua_arch.h since I never did that in CMake, but I find #include <ua_arch.h> easier.

It would be cool if users would no need to modify their code once cmake did its thing.
We can include ua_arch.h from ua_config.h which is also generated/copied by cmake.

> Since the port change I think will be a big one, maybe it would be better to have in another branch

Ok. Tell us when you are ready and we merge this first.

StalderT:
Just a question. With this method, will we still have a single amalgated file (open62541.c/h) containing all the platforms or will we have one file per platform?

cabralfortiss:
This is now ready to be merged

jpfr:
Cool. Just two minor things.
- Can `UA_freeRTOS` in the cmake script be renamed to all-uppercase? For consistency.
- Is it necessary to have `../deps/queue.h`? Normally, cmake should have added /deps to the include directories.

cabralfortiss:
I'll change the UA_freeRTOS to all uppercase.

The problem with queue.h is that in freeRTOS there's also a file called queue.h so there are problems in including and a really mess. By including directly the deps/queue.h the queue.h in freeRTOS is properly included when needed.

jpfr:
Merged. 😄

Pro:
Regarding the architecture structure discussion:
In FORTE (https://www.eclipse.org/4diac/en_rte.php) we are already using different folders for architecture dependent code. 
The arch folder contains independent code, whereas all the subfolders implement it for the corresponding platform. See here:
https://git.eclipse.org/c/4diac/org.eclipse.4diac.forte.git/tree/src/arch?h=develop

Adding a new architecture will automatically extend the options for CMake, and in CMake you can then select which architecture you are using. This will then also work with Amalgamation.

I'm currently doing some experiments with open62541 and Arduino, and started to propose a similar structure for open62541 here, more on that later
https://github.com/Pro/open62541/commit/2212f0c6e7ac16cc4ecf58dd4104a43407099344

PR #1549, opened 2018-01-25
Fix build for iOS

basyskom-jvoe:
When building open62541 for iOS, thread-local storage is not supported for arm and the i386 simulator but UA_THREAD_LOCAL gets expanded to _Thread_local and the build fails with "error: thread-local storage is not supported for the current target".

This patch fixes the issue by checking for these two platforms before the existing checks are done.

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -84,6 +84,7 @@ Building on OS X
 
 .. code-block:: bash
 
+   xcode-select --install
    brew install cmake
    pip install six # python 2/3 compatibility workarounds
    pip install sphinx # for documentation generation
--- a/src/ua_util.h
+++ b/src/ua_util.h
@@ -37,6 +37,11 @@ extern "C" {
     #if (TARGET_IPHONE_SIMULATOR == 1 && TARGET_CPU_X86 == 1) || \
         (TARGET_OS_IPHONE == 1 && TARGET_CPU_ARM == 1)
         #define UA_THREAD_LOCAL
+    #else
+        #include <AvailabliltyMacros.h>
+        #if (MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_7)
+            #define UA_THREAD_LOCAL
+        #endif
     #endif
 #endif
 
--- a/src/ua_util.h
+++ b/src/ua_util.h
@@ -32,16 +32,26 @@ extern "C" {
  * only used for some testing strategies. ``UA_THREAD_LOCAL`` is empty if the
  * feature is not available. */
 
-#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
-# define UA_THREAD_LOCAL _Thread_local /* C11 */
-#elif defined(__cplusplus) && __cplusplus > 199711L
-# define UA_THREAD_LOCAL thread_local /* C++11 */
-#elif defined(__GNUC__)
-# define UA_THREAD_LOCAL __thread /* GNU extension */
-#elif defined(_MSC_VER)
-# define UA_THREAD_LOCAL __declspec(thread) /* MSVC extension */
-#else
-# define UA_THREAD_LOCAL
+#if defined(__APPLE__) && defined(__MACH__)
+    #include <TargetConditionals.h>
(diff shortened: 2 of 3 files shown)


jpfr:
Thanks for the fix.
I would wait with merging until the discussion around ports handling in #1546 has settled.

Is that okay?

basyskom-jvoe:
No problem.
If any changes to this patch are required as a result of the discussion in #1546, please let me know.

WoutvdBerg:
As this is already apple related: OS X 10.6 doesn't support thread local storage either.
Possible to add this also?

basyskom-jvoe:
> As this is already apple related: OS X 10.6 doesn't support thread local storage either.
Possible to add this also?

I have no machine running that version, so I can't test a possible fix. Could you provide a patch?

WoutvdBerg:
This is my last day with access to a mac. If you have a mac it's possible to set a minimum requirement for the compiler. 

> -mmacosx-version-min=10.6

basyskom-jvoe:
> This is my last day with access to a mac. If you have a mac it's possible to set a minimum requirement for the compiler.

Could you please test my last commit on your machine? I only have 10.12.4 available.

WoutvdBerg:
Can't test it on the mac anymore but as the code is the same, it will work.

When you specify a project on your 10.12.4 and give -mmacosx-version-min=10.6 to the compiler, it won't compile with current master. If you do with your commit it should work.

Pro:
Currently OSX build is disabled on Travis. Can you update your PR to reenable it again?
https://github.com/open62541/open62541/blob/master/.travis.yml#L53

When that runs through we can merge. IMHO this does not change that much that we have to extract it to a separate arch file.

StalderT:
@Pro Maybe you can add a second macos build in travis using -mmacosx-version-min=10.6

Pro:
Thanks for the update.
It looks like there is a file missing:

```
/Users/travis/build/open62541/open62541/src/ua_util.h:41:18: fatal error: 'AvailabliltyMacros.h' file not found
        #include <AvailabliltyMacros.h>
```
https://travis-ci.org/open62541/open62541/jobs/334600129#L416


Googling suggests to install `xcode-select --install` which we would like to avoid. Otherwise everyone compiling on OS X would need to install those.

Can you try to avoid including the missing header or is there another way to achieve this functionality?

> @Pro Maybe you can add a second macos build in travis using -mmacosx-version-min=10.6

Adding an additional osx build `-mmacosx-version-min=10.6` can be easily done in the same OS X build instance by adding an additional cmake build block here:
https://github.com/open62541/open62541/blob/master/tools/travis/travis_osx_script.sh#L36

@basyskom-jvoe would you also have the time to quickly add that and check that all the travis builds are green? Your help is very much appreciated!

basyskom-jvoe:
> Can you try to avoid including the missing header or is there another way to achieve this functionality?

I am not aware of an alternative way to detect the OSX version.
If somebody who is more familiar with OSX knows a way, i'll update my commit.

Pro:
@open62541/open62541-core should we expect the user to install `xcode-select --install` to fix this issue?

basyskom-jvoe:
Any decisions on this yet?

Pro:
I would simply go for the additional requirement.

@basyskom-jvoe can you create a PR which adds this additional requirement to the list here:
https://github.com/open62541/open62541/blob/master/doc/building.rst
And add the command to the travis install for OS X so that CI is happy:
https://github.com/open62541/open62541/blob/master/tools/travis/travis_osx_before_install.sh

And then rebase onto current master (I merged the reenabling of OS X builds already).

basyskom-jvoe:
Looks like the tools are already installed... Any ideas?

Pro:
Ok, then it could be that the version of travis' OS X is too old. You can try setting a specific xcode version:

https://docs.travis-ci.com/user/reference/osx/#Xcode-version

If you want to try out some travis builds it's best if you push to your forked repository in a separate branch and enable travis. When that one runs through you can update the PR. This will keep the noise here down a bit.

basyskom-jvoe:
I have just successfully built the current master without any changes from this pull request, so the initial pull request is obsolete now.

As I don't have access to an OS X 10.6 machine to debug the other issue and the oldest OS X available in travis is 10.10, I am closing this pull request.

PR #1551, opened 2018-01-25
Fix incorrect padding calculation

Infinity95:
This PR fixes incorrect padding calculation while decrypting.
The paddingSize byte wasn't removed resulting in decoding errors, when dealing with more than one chunk.

### DIFF
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -759,16 +759,20 @@ decryptChunk(UA_SecureChannel *channel, const UA_SecurityPolicyCryptoModule *cry
 
         if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT ||
            (messageType == UA_MESSAGETYPE_OPN &&
-            channel->securityMode != UA_MESSAGESECURITYMODE_NONE)) {
-            // we need to add one to the padding size since the paddingSize byte itself need to be removed as well.
-            // TODO: write unit test for correct padding calculation
-            paddingSize = (size_t) chunk->data[chunkSizeAfterDecryption - sigsize - 1] + 1;
+            channel->securityMode > UA_MESSAGESECURITYMODE_NONE)) {
+            paddingSize = (size_t) chunk->data[chunkSizeAfterDecryption - sigsize - 1];
 
             size_t keyLength = cryptoModule->getRemoteEncryptionKeyLength(securityPolicy, channel->channelContext);
             if(keyLength > 2048) {
                 paddingSize <<= 8; /* Extra padding size */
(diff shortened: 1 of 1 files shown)


src/ua_securechannel.c
```diff
@@ -755,15 +755,16 @@ decryptChunk(UA_SecureChannel *channel, const UA_SecurityPolicyCryptoModule *cry
...
-        sigsize = cryptoModule-> getRemoteSignatureSize(securityPolicy, channel->channelContext);
+        sigsize = cryptoModule->getRemoteSignatureSize(securityPolicy, channel->channelContext);
 
         if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT ||
            (messageType == UA_MESSAGETYPE_OPN &&
             channel->securityMode != UA_MESSAGESECURITYMODE_NONE)) {
-            paddingSize = chunk->data[chunkSizeAfterDecryption - sigsize - 1];
+            // we need to add one to the padding size since the paddingSize byte itself need to be removed as well.
+            // TODO: write unit test for correct padding calculation
+            paddingSize = (size_t) chunk->data[chunkSizeAfterDecryption - sigsize - 1] + 1;
```
jpfr: Why would this work for a single chunk but not multiple chunks?
Is the extraPaddingSize considered?

👍 for unit tests

Infinity95: Because if we have a single chunk then the padding at the end of the chunk doesn't do anything.
If we have two chunks with AAAA as data and paading size of 0, then the reassembled chunk would look like this:
AAAA0AAAA0
This obviously gives a decoding error. if only AAAA0 needs to be decoded, the decoder just ignores the 0

And yes, you're right about the extra padding size. I forgot that.

Infinity95:
We just never received any chunked messages. Since the decoder only decodes as much as it needs the extra byte wasn't relevant and just ignored.

jpfr:
> We just never received any chunked messages. Since the decoder only decodes as much as it needs the extra byte wasn't relevant and just ignored.

Didn't we check the signatures? They come after the padding.

Infinity95:
The structure of the message is just fine. We just need the padding size in order to remove the padding from the message, since we don't want it during decoding.
The signature is taken from the back of the message buffer. We also know its size, since it is defined by the policy.

jpfr:
> The signature is taken from the back of the message buffer. We also know its size, since it is defined by the policy.

The padding is also signed. So you say that we correctly used padding for checking the signature.
But then we incorrectly used it to define the message end?

So we computed the padding size twice for decoding/signature checking?

PR #1339, opened 2017-11-10
stop setting UA_SUBSCRIPTIONSTATE_LATE at the end of monitoredItem list

StalderT:

PR #1556, opened 2018-01-29
fix LifetimeCount and call UA_Subscription_publishCallback if a publi…

StalderT:
…shResponse is present

Currently, the LifeTimeCount is not checked if the securechannel is invalid. LifeTimeCount is only checked when maxKeepAliveCount occurs.

This PR fix it.

Other change : UA_Subscription_publishCallback only call UA_Subscription_publishCallback if publishResponse is available (small performance improvement).

Part of https://github.com/open62541/open62541/pull/1339

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -264,12 +264,31 @@ UA_Subscription_publishCallback(UA_Server *server, UA_Subscription *sub) {
     UA_LOG_DEBUG_SESSION(server->config.logger, sub->session,
                          "Subscription %u | Publish Callback",
                          sub->subscriptionId);
+    /* Dequeue a response */
+    UA_PublishResponseEntry *pre = UA_Session_getPublishReq(sub->session);
+    if(pre) {
+        sub->currentLifetimeCount = 0; /* Reset the lifetimecounter */
+    } else {
+        UA_LOG_DEBUG_SESSION(server->config.logger, sub->session,
+                             "Subscription %u | The publish queue is empty",
+                             sub->subscriptionId);
+        ++sub->currentLifetimeCount;
+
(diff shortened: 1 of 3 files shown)


Pro:
@StalderT if there is no notification available, shouldn't the server still send back an empty Publish Responses. By quickly looking at your code this behaviour is now changed?

We probably also need a unit test for that case?

StalderT:
> if there is no notification available, shouldn't the server still send back an empty Publish Responses.

If we have no notification and if maxKeepAliveCount is reached and if we have a pending publish request we must send an empty response. This is the case now and with the PR.

> We probably also need a unit test for that case?

Yes ! A new unit test was added to the PR : https://github.com/open62541/open62541/pull/1556/commits/043e1dd0920e5874ed9eaf1fe9c9de65a369b814

jpfr:
Hey @StalderT ,

hooray for adding the unit test!
I just pushed some simplifications.
We are now really close to the original code and the unit tests are green.

Ready to push from my point of view. Okay?

Thanks for pushing the issue for such a long time.

StalderT:
Here : https://github.com/StalderT/open62541/blob/3c5fbc7ff2a3e7d52be52f04cb0494aa097bf313/src/server/ua_subscription.c#L276 we return if maxKeepAliveCount is not reached. In this case we don't increase currentLifetimeCount.

StalderT:
Unit test pass because maxKeepAliveCount is set to 1

StalderT:
This time sounds good. Thanks !!!!

I just modified the test to use a keepalive set to 2.

For me ready to merge when all the tests will be green.

jpfr:
Perfect!
Just one additional line of code in the main lib.

StalderT:
👍 Thanks !

PR #1559, opened 2018-01-30
fixed encoding of overlayable types likes enums (these are int32!)

janitza-thbe:
Currently overlayable types like Enumerations are encoded as ExtensionObjects which is errorneous.

You can see an example in the attached screenshot from UAExpert. There the node class attribute of each node (which is an as Int32 overlayable enum type indeed) is marked as faulty.

ua_expert_nodeclass_invalid_datatype

This pull request changes the encoded type ID to Int32 for all overlayable types. After that UAExpert does not complain any more.

### DIFF
--- a/src/ua_types_encoding_binary.c
+++ b/src/ua_types_encoding_binary.c
@@ -1006,6 +1006,8 @@ enum UA_VARIANT_ENCODINGMASKTYPE {
     UA_VARIANT_ENCODINGMASKTYPE_ARRAY       = (0x01 << 7)  /* bit 7 */
 };
 
+
+
 ENCODE_BINARY(Variant) {
     /* Quit early for the empty variant */
     u8 encoding = 0;
@@ -1014,11 +1016,12 @@ ENCODE_BINARY(Variant) {
 
     /* Set the content type in the encoding mask */
     const bool isBuiltin = src->type->builtin;
-    const bool isOverlayable = src->type->overlayable;
+    const bool isAlias = src->type->membersSize == 1
+                         && UA_TYPES[src->type->members[0].memberTypeIndex].builtin;
     if(isBuiltin)
         encoding |= UA_VARIANT_ENCODINGMASKTYPE_TYPEID_MASK & (u8)(src->type->typeIndex + 1);
-    else if(isOverlayable)
-        encoding |= UA_VARIANT_ENCODINGMASKTYPE_TYPEID_MASK & (u8)(UA_TYPES_INT32 + 1);
+    else if(isAlias)
+        encoding |= UA_VARIANT_ENCODINGMASKTYPE_TYPEID_MASK & (u8)(src->type->members[0].memberTypeIndex + 1);
     else
         encoding |= UA_VARIANT_ENCODINGMASKTYPE_TYPEID_MASK & (u8)(UA_TYPES_EXTENSIONOBJECT + 1);
 
@@ -1037,7 +1040,7 @@ ENCODE_BINARY(Variant) {
         return ret;
 
     /* Encode the content */
-    if(!isBuiltin && !isOverlayable)
+    if(!isBuiltin && !isAlias)
         ret = Variant_encodeBinaryWrapExtensionObject(src, isArray, ctx);
     else if(!isArray)
         ret = encodeBinaryInternal(src->data, src->type, ctx);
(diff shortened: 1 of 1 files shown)


jpfr:
Hello @janitza-thbe,

Good call! We should fix this.

Int32 is not the only overlayable type. `Overlayble` is a flag in the type description to indicate whether we can `memcpy` from the network message to the C representation.
On a little endian platform, all numbers (int, float) and all structures without padding or pointers are overlayable.

In the attribute services, the following snippet is used to detect enums:
```c
static enum type_equivalence
typeEquivalence(const UA_DataType *t) {
    if(t->membersSize != 1 || !t->members[0].namespaceZero)
        return TYPE_EQUIVALENCE_NONE;
    if(t->members[0].memberTypeIndex == UA_TYPES_INT32)
        return TYPE_EQUIVALENCE_ENUM;
    if(t->members[0].memberTypeIndex == UA_TYPES_BYTE && t->members[0].isArray)
        return TYPE_EQUIVALENCE_OPAQUE;
    return TYPE_EQUIVALENCE_NONE;
}
```

janitza-thbe:
Thanks for the clarification and hint to the possible solution!

So do you plan promptly an own fix on this with help of the posted snippet as part of the public or at least internal API?

In this case feel free to close/neglect this PR as your solution will certainly more compliant to your naming and code ordering schema. I'll implement a private quick fix by copying&pasting the code for myself.

jpfr:
Thanks. The generic fix looks about perfect.
Will merge asap.

jpfr:
Thank you. Merged.

janitza-thbe:
Sorry for the inconvinience but I have to "open" this issue again.

I agonised a bit about the handling and enum/structure detection (in both method `typeEquivalence()` and the fix above that is derived from it) and think that it's **not acepptably** and produces **positive falses** in case of:

1. structures with one element only:
   - in `typeEquivalence()` these are detected as enums if the member type is `Int32`
   - in the fix these are detected as enums for each member type
2. own generated data types that different from NS0
   - these are *never* detected validly neither as enums nor as structures

jpfr:
> not acepptably and produces positive falses

Full ACK.
It was a different situation in the write service where we could assume the user *wants the type to match*. This is not good during encoding.

> In my opnion the best solution would be to extend the UA_DataType structure by an extra enum field type or kind with the enum literals Builtin, Enum, Structure/Opaque and the special structure type OptionSet which is not supported by open62541 SDK so far.

That sounds good.
The initial fix with the enum can go in whenever ready.
I'd merge OptionSet in a second step. This is a rather big feature.

Can we leave the rest of the encoding as-is and only replace the last changes?
Afterwards, we can then gradually improve things, e.g. by not even defining members for enum types.

`Kind` sounds good as the `type` name is already widely used. So we can avoid confusion.
Is there some preferred naming in the standard?

janitza-thbe:
> I'd merge OptionSet in a second step. This is a rather big feature.

Yep, you're right and I agree. I only wanted to give you a hint, that `OptionSet` is a big open point that needs special attention like optional members of structures, too. ;-)

> Can we leave the rest of the encoding as-is and only replace the last changes?
Afterwards, we can then gradually improve things, e.g. by not even defining members for enum types.

It's your decision. The current solution has to be replaced before a possible release. The question is if you, @jpfr , will implement the described solution (w/o `OptionSet`) in upcoming days/weeks?

> `Kind` sounds good as the `type` name is already widely used. So we avoid confusion.
Is there some preferred naming in the standard?

I didn't evaluated this. The `kind` naming schema came up during setup of our user model since there is everything full of `type`, too, and seems to be a good alternative.

jpfr:
Hello @janitza-thbe ,

I am trying to resolve the topics you mentioned.

On my box, the enums are already correctly transmitted without this PR!
The Datatype of enums has the `builtin` flag set to true and the type-index of int32.
So they are treated exactly as int32 in the encoding layer.

What are the steps to reproduce the error in your initial screenshot?

janitza-thbe:
I've an adapted `generate_datatypes.py` that accepts multiple model `*.xml` to builtin our data types, with configurable namespace indexes, too. These, because they are different from `ns=0` must not get the flag `builtin` set due to the OPC UA terminology.

fabian727:
do I understand correct that encoding of enumerations as Int32 is the correct way? 
At the moment I implement a client and want to rewrite the enumeration-int32 with their string. What is the considered way to do this? If I read the value I get the DataType as int32 and not as the defined enumeration type.
At the moment I would implement a lookup-table for all Int32 variables, which states which enumeration type it is.

fabian727:
Yep, I got it right.
For better detection there are "propertyType"  and "HasTypeDefinition". And you have to save it client-side or request it every time

PR #1560, opened 2018-01-30
Feature/event preparation

aDogCalledSpot:
Preliminary changes to the subscription system which will allow easier implementation of events. Changes include but are not limited to moving non-datachange-specific functions away from ua_subscription_datachange.c and adding a union to the UA_MonitoredItem_queuedValue struct to specify whether a datachange or event-based service is being used.


jpfr:
Looks good!
Can you remove non-functional changes, such as whitespace changes?
This makes the diff smaller and easier to review.
Thanks.

aDogCalledSpot:
All whitespaces changes have been removed, only "proper" changes are in the pull request.

aDogCalledSpot:
I've edited the tests under the assumption that they are designed for datachange.

src/server/ua_subscription.c
```diff
@@ -19,6 +19,106 @@
 
 #ifdef UA_ENABLE_SUBSCRIPTIONS /* conditional compilation */
 
+UA_MonitoredItem *
```
jpfr: Can we keep these functions in `ua_subscription_datachange.c`? They nearly did not change.

We should keep subscriptions separate from monitoreditems where possible.
The file can be renamed to `ua_subscription_monitoreditems.c` when events drop.

aDogCalledSpot: The idea was to have an additional file `ua_subscription_events.c` in which the functions for the events would be. The functions were moved to `ua_subscription.c` since monitoredItems are also defined in `ua_subscription.h`. Can of course still put them back into `ua_subscription_datachange.c`.

src/server/ua_subscription.c
```diff
@@ -17,8 +17,6 @@
 #include "ua_subscription.h"
 #include "ua_server_internal.h"
 
-#ifdef UA_ENABLE_SUBSCRIPTIONS /* conditional compilation */
```
jpfr: Please keep this.
There's the matching #endif at the end of the file.

PR #1561, opened 2018-01-31
add client to dataChangeHandler (API change !!!!)

StalderT:
Sometimes we need to have the client context in the monitored item handling function. Eg: to get the ```clientContext``` (https://github.com/open62541/open62541/pull/1532). But ```client``` is not present. We can't use ```context``` because it can be used to identify the monitoredItem.

That change the API but it is really needed !

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,19 @@
 This changelog reports changes to the public API. Internal refactorings and bug
 fixes are not reported here.
 
+2018-02-05 pro <profanter at fortiss.org>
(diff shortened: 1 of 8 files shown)


Pro:
Can be merged when all green. The API change was approved in our last telco to be more compliant with the rest of the api.

This affects the callbacks UA_MonitoredItemHandlingFunction and UA_MonitoredEventHandlingFunction

StalderT:
Thanks

Issue #1547, opened 2018-01-24
Duplicated Nodes after Modeling


lehned:
Description
===========
In a test, i made with _UA Modeller_ a object type _batype_ , containing variable _at_ with TypeDefinition _AnalogItemType_ (_EURange_ property). This object type is used by an object instance _myb_ as _type definition_. This test model is exported as XML File, describing the new nodes for the _address space_. The nodeset compiler converts this to files _testmodel.c/.h_ which is included in test program _test.c_.

Problem:
In _UA Expert_ browsing the object _myb_ contains duplicated entities for each component , once the one described by XML (6009) and the same with unknown node (ID's 53xxx). Also in object type browsing all components are duplicated, once from XML and once from unknown node (ID 53xxx).

It seems the open62541 stack generates additional nodes according to the _type definition_, which is in conflict to the xml model which explicitly describes this additional nodes for an object related to the their _type definition_.

What is the right way to handle this? 

Thanks for your great work !

Background Information / Reproduction Steps
===========================================

1. compile model with `python ../tools/nodeset_compiler/nodeset_compiler.py -e ../deps/Schema/Opc.Ua.NodeSet2.xml -x ../module/issue_dup_entries.xml ../module/testmodel` required FULL_NS0
```
<?xml version="1.0" encoding="utf-8"?>
<UANodeSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:uax="http://opcfoundation.org/UA/2008/02/Types.xsd" xmlns="http://opcfoundation.org/UA/2011/03/UANodeSet.xsd" xmlns:s1="http://yourorganisation.org/issue_dup_entries/Types.xsd" xmlns:ua="http://unifiedautomation.com/Configuration/NodeSet.xsd" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <NamespaceUris>
        <Uri>http://yourorganisation.org/issue_dup_entries/</Uri>
    </NamespaceUris>
    <Aliases>
        <Alias Alias="UInt16">i=5</Alias>
        <Alias Alias="Organizes">i=35</Alias>
        <Alias Alias="HasModellingRule">i=37</Alias>
        <Alias Alias="HasTypeDefinition">i=40</Alias>
        <Alias Alias="HasSubtype">i=45</Alias>
        <Alias Alias="HasProperty">i=46</Alias>
        <Alias Alias="HasComponent">i=47</Alias>
        <Alias Alias="Range">i=884</Alias>
[...]

PR #1565, opened 2018-02-01
WIP: Forbid duplicated references

janitza-thbe:
Implemented detection of duplicated references.
Prevents to add equal references more than once and to break the model (on deletion only the first matched reference is deleted since each reference must be unique) and returns UA_STATUSCODE_BADDUPLICATEREFERENCENOTALLOWED

### DIFF
--- a/include/ua_server.h
+++ b/include/ua_server.h
@@ -990,7 +990,11 @@ UA_Server_addMethodNode(UA_Server *server, const UA_NodeId requestedNewNodeId,
  * _finish method adds the references to the parent (and the TypeDefinition if
  * applicable), copies mandatory children, performs type-checking of variables
  * and calls the node constructor(s) at the end. The _finish method may remove
- * the node if it encounters an error. */
+ * the node if it encounters an error.
+ * The _finishSkipParent is a variant of the _finish method which additionally
+ * allows to indicate if adding the parent reference should be skipped if it is
+ * already added before manually.
+ * */
 
 /* The ``attr`` argument must have a type according to the NodeClass.
  * ``VariableAttributes`` for variables, ``ObjectAttributes`` for objects, and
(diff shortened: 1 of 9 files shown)


janitza-thbe:
Travis indicates many "Adding reference to parent failed" errors.
@Pro, this indicates problems inside the nodeset compiler which produces code that leads to duplicated references.

Pro:
Thanks for the PR and the hint, I will have a look at it in the coming days, definitely before 0.3 release.

Pro:
@jpfr this adds an additional `_finishSkipParent` method as public API where you can indicate if the parent reference should be added in the `_finish` step or not.
This is needed since right after adding the node we need to add the parent references, otherwise some consistency checks will fail when adding child nodes, but then `_finish` would fail, since it tries to add the parent reference again.
If you are OK with that you can squash and merge.

jpfr:
Why do we need `_finishSkipParent` if we prevent the addition of duplicate references anyway?

There are two options:
- Instantiate the children, then instantiate the parent
- Instantiate the parent, then instantiate the children

**We should choose one of them, document it and stick with it.**
I don't care which of it we choose. But it should be 100% on either side.

This PR puts use somewhere in the middle. Which is dangerous.

jpfr:
This PR is a bundle of changes.

- UA_ExpandedNodeId_equal
- UA_QualifiedName_equal
- Prevent duplicate references

Can be merge right away.
@Pro @janitza-thbe In a separate PR?

Pro:
That was also my first thought, but there are some problems with the consistency checks:

**Case 2** Instantiate the parent, then instantiate the children:
If we first instantiate the parent, the _finish method will add all the mandatory children before they are added from the nodeset, thus we have again the problem from #1547 and #1568

**Case 1** Instantiate the children, then instantiate the parent:
This will also not work due to some consistency checks, e.g. this one:
https://github.com/open62541/open62541/blob/70819bf5cb0a6b5071663e750c88f1984f154588/src/server/ua_services_nodemanagement.c#L845
It checks if the parent node, where the abstract variable is added as a child, is below `BaseDataVariableType` or `ObjectTypes` (i.e. the parent must be a type and not an instance). If `_finish` is not yet called for the parent node, its parent reference is missing and thus `isNodeInTree` will always return false and the consistency check fails.
There are a few more of this kind.

That's why we need the following steps:
1. create the parent node
2. add all parent references
3. add the child nodes from Nodeset
4. create missing mandatory children of the parent

Another solution would be to change the `_begin` that it also creates the parent reference, but this again will change quite some behaviour...

jpfr:
> That's why we need the following steps:

    create the parent node
    add all parent references
    add the child nodes from Nodeset
    create missing mandatory children of the parent

Thanks for unpacking the issue. Full ACK to do it this way.
But then we **only** do it this way.

Then let's change _begin to include the parent and reference nodeids + do the checks.
And change _finish to not include parent and reference nodeids.

Pro:
Closing in favour of #1603

Issue #1118, opened 2017-07-18
After timeout of session, get aborted


JinGuoGe:
Description
===========
Hi, @jpfr 
Thanks for fixing issue #1071 .
I tested with updated version 3702ffe. Crashing issue was solved but instead of it, get aborted.
This is gdb log.
https://paste2.org/Ua8cx4jF
This is my testing environment screenshot.
https://user-images.githubusercontent.com/15504060/28234883-6662c24a-6937-11e7-9fcf-062735af565e.png

Background Information / Reproduction Steps
===========================================
About reproduction steps, it is exactly same with reproduction test programs which i provided in #1071 

Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): 3702ffe
 - [x] Other OPC UA SDKs used (client or server):  eclipse milo
 - [x] Operating system: ubuntu 
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jpfr:
Again, how to reproduce?
This happens with the same client and pulling out the network cable as for #1071?

JinGuoGe:
Yes, but reproduction process is exactly same. But issue is different. I think crashing issue was solved. Last issue was crash issue, but this issue is abort issue, not crashed.

JinGuoGe:
@jpfr 
How are you? Could you tell me reason and way to solve this issue?
Thanks.

StalderT:
duplicate of https://github.com/open62541/open62541/issues/1284 ?

StalderT:
@JinGuoGe Is the problem still present with the last master commit?

Pro:
Closing as it is probably fixed by latest master commits from @StalderT

StalderT:
@Pro No my commits doesn't fix this issue. I found a way to reproduce the issue. Can you reopen ?

StalderT:
@jpfr I need your help : 

Step to reproduce : 
1) Modifiy ```src/server/ua_subscription.c``` : 
```
static int test=0;
void
UA_Subscription_publishCallback(UA_Server *server, UA_Subscription *sub) {
    UA_LOG_DEBUG_SESSION(server->config.logger, sub->session,
                         "Subscription %u | Publish Callback",
                         sub->subscriptionId);
test++;
if (test>5){
                UA_Session_deleteSubscription(server, sub->session,
[...]

PR #1570, opened 2018-02-02
workaround for #1118

StalderT:
#1118

### DIFF
--- a/src/ua_timer.c
+++ b/src/ua_timer.c
@@ -35,6 +35,7 @@ struct UA_TimerCallbackEntry {
 
     UA_TimerCallback callback;
     void *data;
+    UA_Boolean removed;
 };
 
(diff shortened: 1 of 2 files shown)


jpfr:
This is a very difficult issue.
The timer currently supports multithreaded use.
We should not loose that.

You loop over the items in `UA_Timer_removeRepeatedCallback`.
This was avoided before.
`UA_Timer_removeRepeatedCallback` can be called from another thread while we reorder the the items in the "main loop".

StalderT:
closed in favour of https://github.com/open62541/open62541/pull/1577

StalderT:
> This is a very difficult issue.

I understand. But we should still try to fix this issue. Otherwise we will have some similar problems afterwards.

PR #1573, opened 2018-02-02
Important workarounds for attribute checking (really BIG thing still to do!)

janitza-thbe:
compatible value rank fixes:
- needed to weaken check system since checks are wrong in common cases
- adding nodes by regular API (w/o using begin/finish directly) does not work!
- didn't attract attention since @Pro 's nodeset compiler turns off ALL checks on adding nodes / uses begin/finish methods
- added many ToDo's that describe the general insufficiencies like:
** requirement of "specifiedAttributes" implementation to differentiate between nodes with unset attributes and nodes with attributes that explicitly should match the applied default attributes
** requirement of more structural checks

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -573,7 +573,13 @@ compatibleDataType(UA_Server *server, const UA_NodeId *dataType,
 }
 
 /* Test whether a valurank and the given arraydimensions are compatible. zero
- * array dimensions indicate a scalar */
+ * array dimensions indicate a scalar
+ *
+ * TODO Big thing: Implement "specifiedAttributes" for nodes
+ * to allow the differentiation between unset values which implicitly get errorneously the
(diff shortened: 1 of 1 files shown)


jpfr:
> adding nodes by regular API (w/o using begin/finish directly) does not work!

The "normal" addnodes just calls _begin and _finish in a row.
There should not be any differences.

https://github.com/open62541/open62541/blob/master/src/server/ua_services_nodemanagement.c#L951

Pro:
The checks in the nodeset compiler are only disabled for NS0 initialization. There are some unit tests for PLCopen and ADI/DI which go through the full compatibility checking.

@janitza-thbe do you maybe have a short example by hand, which shows a use case where the value rank check fails?

jpfr:
This has no impact on the public API.
Let's continue after the first 0.3 release candidate.

PR #1576, opened 2018-02-02
TEST PR ONLY RELATED TO #1570

StalderT:
Just to see if the test include in https://github.com/open62541/open62541/pull/1570 failed with master


StalderT:
fail as expected

PR #1581, opened 2018-02-05
Check recursion limit for DiagnosticInfo containing DiagnosticInfo

Pro:
See https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=5835

Credit to oss-fuzz


jpfr:
You missed on return path.

Can we put the check and increase/decrease directly around the recursive call?
`ret |= DECODE_DIRECT(dst->innerDiagnosticInfo, DiagnosticInfo);`

Now we are down to zero issues on oss-fuzz! Champagne!

PR #1583, opened 2018-02-05
Actually verify certificates when creating a securechannel

Infinity95:


Pro:
@Infinity95 looks like this can be merged right away into 0.3?

Since it also comes with an API change we should target at least 0.3 for this.

Infinity95:
Should be mergeable right away. And it's preferable to have this in 0.3 since having no certificate verification at all is undesirable. You can merge as soon as everything is green imho.

PR #1585, opened 2018-02-06
Fix server side indexRange support for monitored items

basyskom-jvoe:
IndexRange must be copied to the monitored item before the first call to UA_MonitoredItem_SampleCallback().
Not doing so produces a data change notification containing the entire array.


jpfr:
Good find. Thanks!

PR #1587, opened 2018-02-07
Added Basic256Sha256 plugin according to Basic128Rsa15, added endpoint in the default config.

bobbthebuilder:
This is a first try to start with Basic256Sha256 encryption.
The specs i followed can be found on:

| Encryption | Security Policy Specification |
| ----------- |:----------|
| Basic128Rsa15  | http://opcfoundation.org/UA-Profile/UA/SecurityPolicy%23Basic128Rsa15 |
| Basic256Sha256 | http://opcfoundation.org/UA-Profile/UA/SecurityPolicy%23Basic256Sha256 |

According to the advice of @Infinity95 i basically added a new endpoint and modified the Basic128Rsa15 plugin by (re)defining
```c
#define UA_SECURITYPOLICY_BASIC256SHA256_RSAPADDING_LEN 66
#define UA_SHA256_LENGTH 32
#define UA_SECURITYPOLICY_BASIC256SHA256_SYM_KEY_LENGTH 32
#define UA_SECURITYPOLICY_BASIC256SHA256_MINASYMKEYLENGTH 256
#define UA_SECURITYPOLICY_BASIC256SHA256_MAXASYMKEYLENGTH 512
```
and replacing the mbedtls function calls from Sha1 to Sha256.

When i use UA_ServerConfig_new_basic256sha256 in `/examples/encryption/server_basic128rsa15.c/`
the server starts successfully but when i connect with UaExpert (setting Basic256Sha256 and SignAndEncrypt) the error is: `BadSecurityPolicyRejected`.
However, the problem is in:
```c 
static UA_StatusCode
asymmetricModule_compareCertificateThumbprint_sp_basic256sha256(const UA_SecurityPolicy *securityPolicy,
                                                               const UA_ByteString *certificateThumbprint) {
    if(securityPolicy == NULL || certificateThumbprint == NULL)

    Basic256Sha256_PolicyContext *pc = (Basic256Sha256_PolicyContext*)securityPolicy->policyContext;
    if(!UA_ByteString_equal(certificateThumbprint, &pc->localCertThumbprint))
        return UA_STATUSCODE_BADCERTIFICATEINVALID;

    return UA_STATUSCODE_GOOD;
}
```
where certificateThumbprint->length = 20 but pc->localCertThumbprint.length = 32.
Thats probably an issue related to Sha1 (160 Bits/20Bytes) vs. Sha256 (256 Bits/32Bytes).

I can see that the `UA_SecureChannel `has a fixed size remoteCerificateThumbprint buffer that has size 20
and is afterwards initialized in `UA_SecureChannel_init` but increasing the buffer size to 32 did not help.
Admittedly, i'm a bit confused about that because i don't expect `src/ua_securechannel.c` to exclusively work with Basic128Rsa15 only.

To try this out, replace UA_ServerConfig_new_basic128rsa15 with UA_ServerConfig_new_basic256sha256 in examples/encryption/server_basic128rsa15.c.

Its clear that both `ua_securitypolicy_basic128rsa15.c` and `ua_securitypolicy_basic256sha256.c` are almost equal. I'd like to do the refactoring and put them together. Not to mention unit tests and an example.

Most probably i have a general misunderstanding in the procedure or oversaw an important point. Maybe @Infinity95 or any of the experienced contributors/maintainers here can guide me a bit or give further advices how to get this working?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -438,6 +438,10 @@ if(UA_ENABLE_ENCRYPTION)
             ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic128rsa15.h)
     set(default_plugin_sources ${default_plugin_sources}
             ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic128rsa15.c)
+    set(default_plugin_headers ${default_plugin_headers}
+            ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic256sha256.h)
+    set(default_plugin_sources ${default_plugin_sources}
+            ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic256sha256.c)
 endif()
 
 if(UA_DEBUG_DUMP_PKGS)
--- a/examples/server.c
+++ b/examples/server.c
@@ -34,7 +34,7 @@ loadFile(const char *const path) {
     /* Get the file length, allocate the data and read */
     fseek(fp, 0, SEEK_END);
     fileContents.length = (size_t)ftell(fp);
-    fileContents.data = (UA_Byte*)UA_malloc(fileContents.length * sizeof(UA_Byte));
+    fileContents.data = (UA_Byte *)UA_malloc(fileContents.length * sizeof(UA_Byte));
     if(fileContents.data) {
         fseek(fp, 0, SEEK_SET);
         size_t read = fread(fileContents.data, sizeof(UA_Byte), fileContents.length, fp);
@@ -87,11 +87,11 @@ helloWorld(UA_Server *server,
            size_t inputSize, const UA_Variant *input,
            size_t outputSize, UA_Variant *output) {
     /* input is a scalar string (checked by the server) */
-    UA_String *name = (UA_String *) input[0].data;
+    UA_String *name = (UA_String *)input[0].data;
     UA_String hello = UA_STRING("Hello ");
     UA_String greet;
     greet.length = hello.length + name->length;
-    greet.data = (UA_Byte *) UA_malloc(greet.length);
+    greet.data = (UA_Byte *)UA_malloc(greet.length);
     memcpy(greet.data, hello.data, hello.length);
     memcpy(greet.data + hello.length, name->data, name->length);
     UA_Variant_setScalarCopy(output, &greet, &UA_TYPES[UA_TYPES_STRING]);
@@ -132,8 +132,8 @@ main(int argc, char **argv) {
     if(argc < 3) {
         UA_LOG_FATAL(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND,
                      "Missing arguments for encryption support. "
-                     "Arguments are <server-certificate.der> "
-                     "<private-key.der> [<trustlist1.crl>, ...]");
+                         "Arguments are <server-certificate.der> "
+                         "<private-key.der> [<trustlist1.crl>, ...]");
         return 1;
     }
 
@@ -145,11 +145,11 @@ main(int argc, char **argv) {
     UA_ByteString *trustList = NULL;
     size_t trustListSize = 0;
     if(argc > 3) {
-        trustListSize = (size_t)argc-3;
-        trustList = (UA_ByteString*)
+        trustListSize = (size_t)argc - 3;
+        trustList = (UA_ByteString *)
             UA_alloca(sizeof(UA_ByteString) * trustListSize);
         for(size_t i = 0; i < trustListSize; i++)
-            trustList[i] = loadFile(argv[i+3]);
+            trustList[i] = loadFile(argv[i + 3]);
     }
 
     /* Loading of a revocation list currentlu unsupported */
@@ -157,9 +157,12 @@ main(int argc, char **argv) {
     size_t revocationListSize = 0;
 
     UA_ServerConfig *config =
-        UA_ServerConfig_new_basic128rsa15(4840, &certificate, &privateKey,
+        /*UA_ServerConfig_new_basic128rsa15(4840, &certificate, &privateKey,
                                           trustList, trustListSize,
-                                          revocationList, revocationListSize);
+                                          revocationList, revocationListSize);*/
+        UA_ServerConfig_new_basic256sha256(4840, &certificate, &privateKey,
+                                           trustList, trustListSize,
+                                           revocationList, revocationListSize);
     UA_ByteString_deleteMembers(&certificate);
     UA_ByteString_deleteMembers(&privateKey);
     for(size_t i = 0; i < trustListSize; i++)
@@ -329,7 +332,7 @@ main(int argc, char **argv) {
 
         /* add an matrix node for every built-in type */
         void *myMultiArray = UA_Array_new(9, &UA_TYPES[type]);
-        attr.value.arrayDimensions = (UA_UInt32 *) UA_Array_new(2, &UA_TYPES[UA_TYPES_INT32]);
+        attr.value.arrayDimensions = (UA_UInt32 *)UA_Array_new(2, &UA_TYPES[UA_TYPES_INT32]);
         attr.value.arrayDimensions[0] = 3;
         attr.value.arrayDimensions[1] = 3;
         attr.value.arrayDimensionsSize = 2;
@@ -458,5 +461,5 @@ main(int argc, char **argv) {
     UA_StatusCode retval = UA_Server_run(server, &running);
     UA_Server_delete(server);
     UA_ServerConfig_delete(config);
-    return (int) retval;
+    return (int)retval;
 }
--- a/include/ua_plugin_securitypolicy.h
+++ b/include/ua_plugin_securitypolicy.h
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
  *
- *    Copyright 2017 (c) Mark Giraud, Fraunhofer IOSB
+ *    Copyright 2017-2018 (c) Mark Giraud, Fraunhofer IOSB
  *    Copyright 2017 (c) Julius Pfrommer, Fraunhofer IOSB
  *    Copyright 2017 (c) Stefan Profanter, fortiss GmbH
  */
@@ -25,7 +25,7 @@ struct UA_SecurityPolicy;
 typedef struct UA_SecurityPolicy UA_SecurityPolicy;
 
 typedef struct {
-    UA_String signatureAlgorithmUri;
+    UA_String uri;
 
     /* Verifies the signature of the message using the provided keys in the context.
      *
@@ -78,8 +78,8 @@ typedef struct {
      * @param channelContext the context to retrieve data from.
      * @return the length of the signing key in bytes. Returns 0 if no length can be found.
      */
-    size_t (*getLocalSigningKeyLength)(const UA_SecurityPolicy *securityPolicy,
-                                       const void *channelContext);
+    size_t (*getLocalKeyLength)(const UA_SecurityPolicy *securityPolicy,
+                                const void *channelContext);
 
     /* Gets the local signing key length.
      *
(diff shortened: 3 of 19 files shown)


bobbthebuilder:
To try this out, replace `UA_ServerConfig_new_basic128rsa15` with `UA_ServerConfig_new_basic256sha256` in `examples/encryption/server_basic128rsa15.c`.

Infinity95:
I believe the certificate thumbprint is always a 20 bit sha1 hash. Can you check with the specification? I don't have acces to it right now.

bobbthebuilder:
Thank you for your response! I searched the the specification and made a find in Part 4 "Services" of the Specifiation, p. 157: "_EncryptingCertificate | ByteString | The SHA1 thumbprint of the DER form of the encrypting Certificate._" - seems you were right! :-)

I changed my code and got rid of the reported error, now getting the next from from mbedTLS:
"mbedTLS returned an error: RSA - Input data contains invalid padding and is rejected"

Infinity95:
You have to change the encryption as well, since this uses the oaep padding scheme.
Take a look at mbedtls_rsa_rsaes_oaep_encrypt

Pro:
thanks for the PR! You should set up your git environment correctly, i.e. your config.email is not set to the email address you use on github, thus the strange CLA message.
See https://help.github.com/articles/setting-your-commit-email-address-in-git/

After doing that, you need to amend your commit and change the author, then force push to your branch:
https://stackoverflow.com/questions/750172/change-the-author-and-committer-name-and-e-mail-of-multiple-commits-in-git

jpfr:
Nice! This is a very welcome addition.

I sense an opportunity to reduce code duplication.
The code is 85% identical to the basic128rsa15 policy.

The existing implementation has a struct `Basic128Rsa15_PolicyContext`.
We could rename that to `MBedTLS_PolicyContext` and specialize in the structure.
So the main code remains the same.

- Put keylength etc. in the context struct
- Put function pointers to specific encryption algorithms in the context struct
- If that is too complicated, use an enum in the context struct to differentiate between policies and put `if/else` in the code

bobbthebuilder:
@Infinity95 You are right, the RSAES-OAEP padding scheme is used for asym_encrypt.
The error "mbedTLS returned an error: RSA - Input data contains invalid padding and is rejected" i however got comes out of the asym_decrypt function. Using mbedtls_rsa_rsaes_oaep_decrypt() instead of 
mbedtls_pk_decrypt() didn't solve it, no matter what i tried. Do you understand why?

Infinity95:
Before calling the function you have to set the correct padding scheme in the rsa context. Did you consider that?

bobbthebuilder:
Yes, i call mbedtls_rsa_set_padding before:
```c
mbedtls_rsa_context *rsaContext = mbedtls_pk_rsa(cc->policyContext->localPrivateKey);
mbedtls_rsa_set_padding(rsaContext, MBEDTLS_RSA_PKCS_V21, MBEDTLS_MD_SHA1); // also tried other parameter combinations
//...
int mbedErr = mbedtls_rsa_rsaes_oaep_decrypt(rsaContext,                 // ctx
                                             mbedtls_ctr_drbg_random,    // f_rng
                                             &pc->drbgContext,           // p_rng
                                             MBEDTLS_RSA_PRIVATE,        // mode
                                             label,                      // label
                                             0,                          // label_len
                                             &outLength,                 // olen
                                             data->data + inOffset,      // input
                                             decrypted.data + offset,    // output
                                             decrypted.length - offset); // output_max_len
```

bobbthebuilder:
@Infinity95 i got rid of the aforementioned mbedtls errors by my last commit.
When i start the server and try to connect with UaExpert, the error message i'm getting is "Error 'Bad' was returned during OpenSecureChannel". The server log however shows no errors and despite many debug output i was not yet able to locate the faulty function.

@Pro I followed your description yesterday and afterwards CLA was displayed as signed here. Unfortunately, the next commit after a pull from another box produced another not CLA signed commit, that was not intended by me of cause. I try to fix it.

@jpfr Full agree, my plan is to get the feature basically working, then refactor both Basic128Rsa15 and Basic256Sha256 following the steps you outlined.

Infinity95:
Since the encryption algorithm uses sha1 as rsa padding you only have an overhead of 42 bytes. You specified the padding length for sha256. Try changing that.

bobbthebuilder:
@Infinity95 changed UA_SECURITYPOLICY_BASIC256SHA256_RSAPADDING_LEN to 44, that didn't help.

bobbthebuilder:
I took a closer look and observed:

- on the asymmetric side each and every function is invoked at least once and none of them raises a bad status code.

- on the symmetric side the following function are invoked:
1. sym_getEncryptionKeyLength_sp_basic256sha256
2. sym_generateNonce_sp_basic256sha256
3. sym_generateKey_sp_basic256sha256
4. swapBuffers_Basic256Sha256
- on the symmetric side the following functions are NOT invoked:
1. sym_verify_sp_basic256sha256
2. sym_sign_sp_basic256sha256
3. sym_getSignatureSize_sp_basic256sha256
4. sym_encrypt_sp_basic256sha256
5. sym_decrypt_sp_basic256sha256
- channelContext's functions are all invoked except
1. channelContext_compareCertificate_sp_basic256sha256

Thus i blame the symmetric part to be erroneous.

Here is the call stack on the server side when a client connects:
```
info/network       Connection 5 | New connection over TCP from 127.0.0.1
info/userland     after continue 1
info/channel       Creating a new SecureChannel
info/channel       Connection 5 | SecureChannel 1 | Opened SecureChannel
info/channel       Connection 5 | SecureChannel 1 | CloseSecureChannel
info/network       Connection 5 | Closed by the server
info/network       Connection 5 | New connection over TCP from 127.0.0.1
info/userland     continue 1
info/userland     after continue 1
info/userland     asymmetricModule_compareCertificateThumbprint_sp_basic256sha256
info/channel       Creating a new SecureChannel
info/userland     channelContext_newContext_sp_basic256sha256
info/userland     parseRemoteCertificate_sp_basic256sha256
info/userland     asym_makeThumbprint_sp_basic256sha256
info/userland     asymmetricModule_compareCertificateThumbprint_sp_basic256sha256
info/userland     asym_decrypt_sp_basic256sha256
info/userland     asym_getRemoteSignatureSize_sp_basic256sha256
info/userland     asym_verify_sp_basic256sha256
info/userland     sym_getEncryptionKeyLength_sp_basic256sha256
info/userland     sym_generateNonce_sp_basic256sha256
info/userland     sym_getEncryptionKeyLength_sp_basic256sha256
info/userland     sym_generateKey_sp_basic256sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     sym_8_last
info/userland     channelContext_setRemoteSymSigningKey_sp_basic256sha256
info/userland     channelContext_setRemoteSymEncryptingKey_sp_basic256sha256
info/userland     channelContext_setRemoteSymIv_sp_basic256sha256
info/userland     sym_generateKey_sp_basic256sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     swapBuffers_Basic256Sha256
info/userland     sym_8_last
info/userland     channelContext_setLocalSymSigningKey_sp_basic256sha256
info/userland     channelContext_setLocalSymEncryptingKey_sp_basic256sha256
info/userland     channelContext_setLocalSymIv_sp_basic256sha256
info/channel       Connection 5 | SecureChannel 2 | Opened SecureChannel
info/userland     asym_getLocalSignatureSize_sp_basic256sha256
info/userland     channelContext_getRemoteAsymEncryptionBufferLengthOverhead_sp_basic256sha256
info/userland     channelContext_getRemoteAsymPlainTextBlockSize_sp_basic256sha256
info/userland     channelContext_getRemoteAsymPlainTextBlockSize_sp_basic256sha256
info/userland     asym_getLocalSignatureSize_sp_basic256sha256
info/userland     asym_decrypt_sp_basic256sha256
info/userland     asym_decrypt_sp_basic256sha256
info/userland     asym_getLocalSignatureSize_sp_basic256sha256
info/userland     channelContext_getRemoteAsymEncryptionBufferLengthOverhead_sp_basic256sha256
info/userland     channelContext_getRemoteAsymPlainTextBlockSize_sp_basic256sha256
info/userland     asym_getLocalSignatureSize_sp_basic256sha256
info/userland     asym_sign_sp_basic256sha256
info/userland     asym_encrypt_sp_basic256sha256
info/userland     channelContext_getRemoteAsymPlainTextBlockSize_sp_basic256sha256
info/userland     channelContext_getRemoteAsymEncryptionBufferLengthOverhead_sp_basic256sha256
info/userland     channelContext_getRemoteAsymPlainTextBlockSize_sp_basic256sha256
info/network       Connection 5 | Closed by the server
info/channel       Connection 0 | SecureChannel 2 | SecureChannel has timed out
info/userland     channelContext_deleteContext_sp_basic256sha256
```
Does anybody see what goes wrong here?

jpfr:
How do you generate the server certificate/key pair?
Note that you will need a longer key for 256 bit.

Infinity95:
The minimum required size is 2048 bit, which is the default length configured in the script.

bobbthebuilder:
@jpfr i generated the server certificate the key pair using `open62541/tools/certs/create_self-signed.py`.
I also tried setting the default length to 4096 bit, but that didn't change anything.

Infinity95:
Here are my changes we made today:
https://github.com/Infinity95/open62541/tree/feist/feature/SecurityPolicy%23Basic256Sha256_new

bobbthebuilder:
@Infinity95 thank you for pushing the changes! Great that you also changed md_hmac to md_hmac_Basic256Sha256 and UA_SHA256_LENGTH vs. UA_SHA1_Length mixup i forgot to commit.

According to the spec v. 1.0.4., chapter 7. page 19 the SymmetricEncryptionAlgorithm_AES256-CBC has a key size of 256 bits and a block size of 16 bytes. The counterpart in sym_decrypt_sp_basic256sha256 makes use auf a keylength that has size 128, do you agree it has to be 256 here instead?
(tried using 256 bits but it had no effect).

Infinity95:
Since symmetric encryption is just what the name implies, i.e. symmetric, the key should be the same for both directions. In this case 32 bytes (256 bits /8).
So it should be block size = 16 byte and key length = 32 byte for both sides.

I'll try to make the nececarry API changes we discussed as soon as possible. Stay tuned for that.

Infinity95:
I think i might have found a possible error we have. Currently we return the block size as key length but that's wrong. The key length is 32 byte whereas the block size is 16 byte. This is possible since aes uses the actual 32 byte key to derive 16 byte round keys which are then xored onto the actual 16 byte block (simplified of course)

jpfr:
Can you just add the whole shebang of
- cleartextblocksize
- encryptedblocksize

directly as integers into the UA_SecurityPolicyCryptoModule structure?

And since we are already on it, do we really need methods to get getLocalEncryptionKeyLength and getLocalEncryptionKeyLength or isn't this rather predefined by the policy?
If not, a sentence in the comments about varying keylengths would help.

Infinity95:
@jpfr That was the plan.
In the asymmetric case the sizes have to be extracted from the certificates/keys. In the symmetric case the sizes are always constant for one policy. It would be best though to always use the getLocalEncryptionKeyLength etc. functions to be consistent i believe. I don't really have time today but i'll try to make the necessary changes tomorrow after my exam and create a separate PR for that. Shouldn't be too many changes required but we should try to include it in 0.3 i think.

Infinity95:
I fixed it. Please rebase on my branch that i linked in a previous comment.
Also add your name to the header.

bobbthebuilder:
@Infinity95 sorry for the late response, i had the opportunity to test your changes on friday afternoon and can confirm that you succeeded - great, thank you very much for your effort! I waited for #1613  to be reviewed by @jpfr and now did the rebase.

One side note: The endpoints in `plugins/ua_config_default.c` do not support logins with username and  password as they are still commented out in the endpoints of Basic128Rsa15 as well as Basic256Sha256. Did we oversaw putting them in or is this intentional?

Infinity95:
I'll have to check. Don't know off the top of my head what's going on there.
As soon as the other PR is merged you should probably rebase on master to get rid of the redundant commits.

bobbthebuilder:
@Infinity95 no problem, just to let you know, i enabled the username/password policy in the Basic256Sha256 endpoint and grabbed open62541/examples/encryption/server_basic128rsa15.c (using UA_ServerConfig_new_basic256sha256()) and hoped the best but the error i got is BadIdentityTokenInvalid because of 
```c
if(token->encryptionAlgorithm.length > 0)
    return UA_STATUSCODE_BADIDENTITYTOKENINVALID;
```
in open62541/plugins/ua_accesscontrol_default.c

(when outcommenting this check, UA_String_equal(&token->password, pw) is false later).

Infinity95:
@bobbthebuilder Please rebase on master before continuing to work on this. It will probably save you a lot of work later on, since some of the api changed.

The user identity tokens use their own policy which can be different from the one in the securechannel. I don't know exactly how this works right now but if the securechannel is already encrypted i believe it is not necessary to encrypt anything in the user token.
You'll have to check with the standard here.

Pro:
Replaced by #1634

PR #1589, opened 2018-02-08
Request both timestamps in addMonitoredItems

basyskom-jvoe:
In the current code, the call to UA_CreateMonitoredItemsRequest_init()
sets timestampsToReturn to 0 which means source timestamp only.
There is currently no way to override the requested timestamps, so
UA_TIMESTAMPSTORETURN_BOTH is a better default.


jpfr:
The client subscription API will become a lot more complete in #1497.
We keep the old API for compatibility throughout 0.3.
Let's merge this once #1497 is in.

jpfr:
#1497 has landed.
You can now select the timestampstoreturn.

If you want changes in the (deprecated) old API, please reopen.

PR #1593, opened 2018-02-13
Add flag for interactive valgind dumps

Infinity95:
The PR adds support for enabling valgrind to dump leak checks every iteration of the server loop.
This makes it possible to do interactive checking and using the valgrind dumps to monitor the server while fuzzing for example.


Pro:
Can you mark the option as advanced in CMake?

Pro:
Appveyor will be fixed with #1594

Pro:
Thanks, appveyor fix landed. Please rebase onto master, then all should be green

PR #1598, opened 2018-02-18
WIP client : timed async service & async highlevel functions with client worker for 0.4

StalderT:
replace #1531


include/ua_client.h
```diff
@@ -65,12 +65,18 @@ typedef void (*UA_ClientStateCallback)(UA_Client *client, UA_ClientState clientS
...
  * Subscription Inactivity callback
- * ------------------------- */
+ * -------------------------------- */
 
 #ifdef UA_ENABLE_SUBSCRIPTIONS
 typedef void (*UA_SubscriptionInactivityCallback)(UA_Client *client, UA_UInt32 subscriptionId, void *subContext);
 #endif
 
+/**
+ * Inactivity callback
```
Pro: @StalderT can you add a few sentences to explain the two inactivity callback. I.e. when are the callbacks called, for what can they be used

StalderT: @Pro done in the definition of UA_ClientConfig

StalderT:
Ready to merge?

src/client/ua_client.c
```diff
@@ -419,6 +419,38 @@ __UA_Client_AsyncService(UA_Client *client, const void *request,
     return UA_STATUSCODE_GOOD;
 }
 
+static UA_StatusCode
+UA_Client_backgroundConnectivityCheck(UA_Client *client) {
+    if(client->config.connectivityCheckInterval) {
```
jpfr:
```c
if(!client->config.connectivityCheckInterval)
    return UA_STATUSCODE_GOOD;
```
This reduces the indendation of the remaining method and makes it more readable.

include/ua_client.h
```diff
@@ -90,14 +96,34 @@ typedef struct UA_ClientConfig {
     /* Callback function */
     UA_ClientStateCallback stateCallback;
 #ifdef UA_ENABLE_SUBSCRIPTIONS
+    /* When outStandingPublishRequests is greater than 0,        */
```
jpfr: This comment style is rather unusual for our code base.

src/client/ua_client.c
```diff
@@ -427,8 +459,12 @@ UA_Client_runAsync(UA_Client *client, UA_UInt16 timeout) {
     if (retvalPublish != UA_STATUSCODE_GOOD)
         return retvalPublish;
 #endif
+    UA_StatusCode retval = UA_Client_backgroundConnectivityCheck(client);
```
jpfr: BackgroundConnectivityCheck does not respect the `timeout` variable.
Instead it uses the *global* timeout settings of the client.

Since this is a synchronous service call, we might block for a long time.
Can we make this an asynchronous call?

In the long term, once we have an event loop in the client, we can time-trigger the connectivity check.

StalderT:
@jpfr In the last commit I added a new API function named : ```__UA_Client_TimedAsyncService```. Now we can use asynchronous service with timeout feature. Ok for you ? Else I can remove the last commit. Thanks in advance for your reply.

jpfr:
> In the last commit I added a new API function named : __UA_Client_TimedAsyncService. Now we can use asynchronous service with timeout feature. Ok for you ? Else I can remove the last commit. Thanks in advance for your reply.

Does that mean we can use `UA_Client_AsyncService` to use the global timeout (in the client config) and `UA_Client_TimedAsyncService` for a specific timeout?
Yes! That is a very good feature.

StalderT:
> Does that mean we can use UA_Client_AsyncService to use the global timeout (in the client config) and UA_Client_TimedAsyncService for a specific timeout?

Yes, good idea. Fixed in the previous commit.

StalderT:
For me, ready to merge

StalderT:
Have you questions/comments about this PR?

examples/client_connectivitycheck_loop.c
```diff
@@ -0,0 +1,73 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+/* Enable POSIX features */
+#ifndef _XOPEN_SOURCE
```
Root-Core: We should keep this consistent, see #1629

jpfr:
The PR only adds API but does not break any.
Let's merge after the first release candidate.

StalderT:
@jpfr  With the latest commits both ```__UA_Client_AsyncService``` and ```__UA_Client_TimedAsyncService are ```timed```.

Would not it be better to rename ```__UA_Client_TimedAsyncService``` to something like ```__UA_Client_AsyncServiceEx```?

StalderT:
RC-1 is out. Can we merge?

examples/client_connectivitycheck_loop.c
```diff
@@ -0,0 +1,73 @@
...
+
+#ifdef _WIN32
+# include <windows.h>
+# define UA_sleep_ms(X) Sleep(X)
+#else
+# include <unistd.h>
+# define UA_sleep_ms(X) usleep(X * 1000)
+#endif
+
+UA_Boolean running = true;
```
fmeerkoetter: these two could also be made static. would limit its visibility to this translation unit and would be consistent with stopHandler/inactivityCallback

examples/client_connectivitycheck_loop.c
```diff
@@ -0,0 +1,73 @@
...
+#ifndef _BSD_SOURCE
+# define _BSD_SOURCE
+#endif
+
+#include "open62541.h"
+#include <signal.h>
+
+#ifdef _WIN32
+# include <windows.h>
+# define UA_sleep_ms(X) Sleep(X)
```
fmeerkoetter: the UA_ prefix is normaly used for API elements. does it make sense for mere example to define macros which use the UA_ prefix?

examples/client_connectivitycheck_loop.c
StalderT: same as client_connect_loop.c, client_subscription_loop.c, server.c, ....

You can do a PR to change that if you want.

StalderT:
Can we merge?

FlorianPalm:
If I am not mistaken, then the connect function (*UA_ClientConnectionTCP*) within *ua_network_tcp.c* is still blocking.
We have put some effort into a more or less non-blocking connect-version (have a look  at #1512, extension of *ua_network_tcp.c* with *UA_ClientConnectionTCP_poll*). Although the used function *getAddrInfo* is still blocking, there could be a solution with [c-ares](https://github.com/c-ares/c-ares)

Your PR extends open62541 with async-service support with timeout-feature, which I really appreciate. I would like to see both of the features together. 
At first glance, I would say it would be easier to integrate the non-blocking-connect-code from #1512  into your approach. 
What do you think?

StalderT:
> If I am not mistaken, then the connect function (UA_ClientConnectionTCP) within ua_network_tcp.c is still blocking.

Yes, connect function is blocking.

Normally (I hope so) this PR (https://github.com/open62541/open62541/pull/1598) will be merge soon (in 0.3 too). The you can use async-service with timeout-feature.

If I good understand https://github.com/open62541/open62541/pull/1512 brings several new features : 

- non blocking connection
- repeated callbacks
- high level async

For high level async you can use async-service with timeout-feature. But for non blocking connection, I thinks an another mechanism must be put in place. For repeated callbacks you use ```UA_Client_run_iterate```. Currently,we use ```UA_Client_runAsync``` to treat background task. I think these two functions need to converge (then duplicate code can be removed). Probably ```UA_Client_runAsync``` is not the good wording.

Have I answered your question?

I will look at these next days in more detail https://github.com/open62541/open62541/pull/1512

sherylll:
Hi,

I'm working on merging this PR and #1512. I removed some duplicate function declarations from both PRs. It should also be possible to put together the two function `UA_Client_runAsync` and `UA_Client_run_iterate`. 

However I wonder why in `check_client_async` the program should be timed exactly (there you use timeout to control asyncCounter). I would simply run `UA_Client_run_iterate` so long until all the requests are processed. But with regards to timing, in unit tests `UA_DateTime_nowMonotonic` always returns 0, is the `UA_fakeSleep` the only way to solve this? Since `UA_Client_run_iterate` uses `UA_Timer` and the timer always returns some garbage values in the tests, it's a bit tricky to get the whole thing right... Any help/suggestions on that?

I would make another PR combining this PR and #1512 this week (probably missing some features from this PR) then we can add one thing at a time. What do you say?

jpfr:
> It should also be possible to put together the two function UA_Client_runAsync and UA_Client_run_iterate.

👍 

> But with regards to timing, in unit tests UA_DateTime_nowMonotonic always returns 0, is the UA_fakeSleep the only way to solve this? Since UA_Client_run_iterate uses UA_Timer and the timer always returns some garbage values

Garbage values should never be returned. Please report this in an issue.

The unit tests indeed use a "manual" clock.
- Some of the unit tests concern timeouts. But we don't want to wait many minutes for the unit tests to complete
- Some edge cases require precise timing that is difficult to replicate deterministically due to preemption of processes by the OS scheduler.

StalderT:
> I would simply run ```UA_Client_run_iterate```` so long until all the requests are processed.

In this case : https://github.com/sherylll/open62541/blob/92df5bacd7c5c0e554b97b8a2389e6a54aff1422/tests/client/check_client_async_connect.c#L33-L34 can eat 100% of the CPU. 

With https://github.com/sherylll/open62541/blob/92df5bacd7c5c0e554b97b8a2389e6a54aff1422/examples/client_async.c#L156-L157, we need to wait 100ms to call ```UA_Client_run_iterate```` again and check if we have receive a response.

Currently with ```UA Client runAsync``` we specify a timeout and we treat immediately an incoming response.

Calling :
```
while(running)
    UA Client runAsync( ... );
```

uses almost no CPU time.

sherylll:
Now the reworked version of PR #1512 is here: [](https://github.com/StalderT/open62541/pull/4)

> we need to wait 100ms to call `UA_Client_run_iterate` again and check if we have receive a response.
You can also modify the function to specify a timeout value. For example by putting a wrapper around it. However for most cases I think it's okay to let the user decide how often this function is called.

plugins/ua_network_tcp.c
```diff
@@ -1168,12 +1168,8 @@ UA_ClientConnectionTCP(UA_ConnectionConfig conf,
             connected = true;
             break;
         }
-        /*fix double free (clang)*/
```
sherylll: If these lines were replaced by `ClientNetworkLayerTCP_close` clang compiler will complain. This is because in `ClientNetworkLayerTCP_close` there is this condition check `if(tcpConnection->server)` before we really free `tcpConnection->server`. However I think it's only a problem of the compiler...

StalderT: Ok. Thank you for your explanation. But we need to fix it without changing UA_ClientConnectionTCP

clang is often right.

Maybe setting tcpConnection->server and connection->handle at NULL after the free can resolve the issue.

tests/client/check_client_async_connect.c
```diff
@@ -99,16 +98,15 @@ START_TEST(Client_connect_async){
                                               &asyncCounter, &reqId);
         }
         /*manual clock for unit tests*/
-        UA_fakeSleep(20);
+        UA_realSleep(20);
```
sherylll: change this back to `fakeSleep` then this test should work again

StalderT:
Closed so as not to pollute the history. We will open later.

PR #1603, opened 2018-02-19
forbid duplicate references and change _begin to also add parent references

Pro:
This is a replacement of #1565

It changes the behavior of _begin to also add the parent and hasTypeDef reference. The _finish method then only adds the remaining children.

See the corresponding discussion in #1565

### DIFF
--- a/include/ua_server.h
+++ b/include/ua_server.h
@@ -985,12 +985,26 @@ UA_Server_addMethodNode(UA_Server *server, const UA_NodeId requestedNewNodeId,
  * pseudo-random unique NodeIds. Existing children are detected during the
  * _finish part via their matching BrowseName.
  *
- * The _begin method prepares the node and adds it to the nodestore. It may copy
- * some unassigned attributes from the TypeDefinition node internally. The
- * _finish method adds the references to the parent (and the TypeDefinition if
- * applicable), copies mandatory children, performs type-checking of variables
- * and calls the node constructor(s) at the end. The _finish method may remove
- * the node if it encounters an error. */
+ * The _begin method:
+ *  - prepares the node and adds it to the nodestore
+ *  - copies some unassigned attributes from the TypeDefinition node internally
+ *  - adds the references to the parent (and the TypeDefinition if applicable)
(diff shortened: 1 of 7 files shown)


jpfr:
`Operation_addNode_finish` still requires the typeDefinition.
Because `Operation_addNode_begin` does not have it.
Otherwise this looks good.

janitza-thbe:
@Pro:
The latest changes can lead to NULL pointers due to invalid nodestore usage.
In `Operation_addNode_begin()` and `copyChildNode()` the `node` pointer is used further after adding the node to the nodestore (e.g. passed to `addParentAndTypeRef()` or dereferenced the get its ID).
As far as I analyzed adding a node to the nodestore does **not** increment the node's reference counter.
After a possible concurrent delete operation onto the node (`UA_Nodestore_get()` -> `UA_Nodestore_delete()`) the data region to which the original node pointer points could already be freed!
So either a fresh node pointer has to be retrieved directly after adding the node to the nodestore or you have to work further with just its ID and the further functions that require it can selectively fetch their own pointer to the node pointers if _really_ required.

Pro:
Thanks for digging into it.

The concurrent delete operation can lead to the same problem anywhere in the code, where `UA_Nodestore_get()` is used, i.e. `UA_Nodestore_get()` can be called somewhere in a method and the node is returned. That code is interrupted and another one calls `UA_Nodestore_delete()` while the first thread still holds the node. Looking at the current nodestore implementation, there is no check in _delete() if refcount is 0 at runtime. It only uses a UA_assert which is disabled on release builds therefore not checked.

@jpfr IMHO deleting a node from the nodestore should only be allowed if refcount is 0, and checked with an if. Of course this is only relevant on Multithreading. Should we add a TODO or somthing like
```c
UA_NodeMap_delete() {
//...
#ifdef UA_ENABLE_MULTITHREADING
if (node->refCount > 0)
  return SOME_ERROR_CODE:
#endif
//...
}
```

Regarding the `node` passing to `addParentAndTypeRef()` I'll create a new PR which uses the `_get` method.

PR #1609, opened 2018-02-20
Nodeset Compiler: Print custom datatype definition instead of using binaryEncode/Decode hack

Pro:
See:
https://groups.google.com/d/msgid/open62541/99749a8f-85c6-4835-adcd-30adf0738f07%40googlegroups.com

Up to now the datatype was not correctly handled if an alias was used. Now that there are the corresponding variables included in the nodeset, `UA_encodeBinary` needs to be used to encode the variable values.

**WARNING:**
This moves `UA_encodeBinary` and `UA_decodeBinary` to the public API. We need to find a better solution for this.
Maybe add a wrapper for `UA_encodeBinary` for the public API?


jpfr:
Aliases are the symptoms. They are not the cause.
If I replace all the aliases with the NodeIds in the nodeset xml, the problem persists.
Otherwise do a simple search/replace in the nodeset compiler.

The ideal solution would be to not manually encode values in the generated code at all.
We have all the facilities in place to do that now.

In the generated code:
1. Print the type for a new custom structure
2. Print the type description (with macros referring to the custom structure to get the padding right)
3. Add the custom type description to the server
4. Add the variable node with our standard API

Pro:
Closed in favour of #1914

PR #1613, opened 2018-02-22
SecurityPolicy Interface change

Infinity95:
This changes the SecurityPolicy interface a bit to unify asymmetric and symmetric module usage.
It also enables missing functionality that i overlooked ~(not commited yet)~, which is needed for Basic256Sha256 for example.

~I will try to finish this today so we can quickly merge.~


Infinity95:
@jpfr Can you review the changes i made? I think i'm done now and we can merge if everything is green and you approve.

include/ua_plugin_securitypolicy.h
```diff
@@ -81,9 +101,9 @@ typedef struct {
      *                       the keys to encrypt data.
      * @param data the data that is encrypted. The encrypted data will overwrite
      *             the data that was supplied. */
-    UA_StatusCode(*encrypt)(const UA_SecurityPolicy *securityPolicy,
-                            void *channelContext,
-                            UA_ByteString *data) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
+    UA_StatusCode (*encrypt)(const UA_SecurityPolicy *securityPolicy,
```
jpfr: change is not required

include/ua_plugin_securitypolicy.h
```diff
@@ -92,17 +112,17 @@ typedef struct {
      * @param channelContext the channelContext which contains information about
      *                       the keys needed to decrypt the message.
      * @param data the data to decrypt. The decryption is done in place. */
-    UA_StatusCode(*decrypt)(const UA_SecurityPolicy *securityPolicy,
-                            void *channelContext,
-                            UA_ByteString *data) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
+    UA_StatusCode (*decrypt)(const UA_SecurityPolicy *securityPolicy,
```
jpfr: additional space not required

include/ua_plugin_securitypolicy.h
```diff
@@ -92,17 +112,17 @@ typedef struct {
...
 
     /* Returns the length of the key used locally to encrypt messages in bits
      *
      * @param securityPolicy the securityPolicy the function is invoked on.
      * @param channelContext the context to retrieve data from.
      * @return the length of the local key. Returns 0 if no
      *         key length is known. */
-    size_t (*getLocalEncryptionKeyLength)(const UA_SecurityPolicy *securityPolicy,
+    size_t (*getLocalKeyLength)(const UA_SecurityPolicy *securityPolicy,
                                           const void *channelContext);
```
jpfr: indendation of the second line of arguments

include/ua_plugin_securitypolicy.h
```diff
@@ -111,8 +131,57 @@ typedef struct {
      * @param channelContext the context to retrieve data from.
      * @return the length of the remote key. Returns 0 if no
      *         key length is known. */
-    size_t (*getRemoteEncryptionKeyLength)(const UA_SecurityPolicy *securityPolicy,
+    size_t (*getRemoteKeyLength)(const UA_SecurityPolicy *securityPolicy,
+                                           const void *channelContext);
```
jpfr: indendation of the second line of arguments

plugins/ua_securitypolicy_basic128rsa15.c
```diff
@@ -13,11 +13,13 @@
...
 #include <mbedtls/entropy_poll.h>
 #include <mbedtls/error.h>
-#include <ua_plugin_pki.h>
 
+#include "ua_plugin_pki.h"
+#include "ua_plugin_securitypolicy.h"
 #include "ua_securitypolicy_basic128rsa15.h"
 #include "ua_types.h"
 #include "ua_types_generated_handling.h"
+#include "../src/ua_securechannel.h"
```
jpfr: This is an internal API. You can't use it from here.
What do you need that is part of the SecureChannel?

src/server/ua_securechannel_manager.c
```diff
@@ -81,13 +83,14 @@ UA_SecureChannelManager_cleanupTimedOut(UA_SecureChannelManager *cm, UA_DateTime
...
-    channel_list_entry* entry;
+static UA_Boolean
+purgeFirstChannelWithoutSession(UA_SecureChannelManager *cm) {
+    channel_list_entry *entry;
     LIST_FOREACH(entry, &cm->channels, pointers) {
         if(LIST_EMPTY(&(entry->channel.sessions))) {
             UA_LOG_DEBUG_CHANNEL(cm->server->config.logger, &entry->channel,
                                  "Channel was purged since maxSecureChannels was "
-                                 "reached and channel had no session attached");
+                                     "reached and channel had no session attached");
```
jpfr: indendation

src/server/ua_services_session.c
```diff
@@ -182,8 +182,10 @@ void Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
     }
 
     UA_LOG_DEBUG_CHANNEL(server->config.logger, channel,
-           "Session " UA_PRINTF_GUID_FORMAT " created",
-           UA_PRINTF_GUID_DATA(newSession->sessionId.identifier.guid));
+                         "Session "
```
jpfr: ???

include/ua_plugin_securitypolicy.h
```diff
@@ -111,8 +131,57 @@ typedef struct {
...
+     *
+     * @param securityPolicy the securityPolicy the function is invoked on.
+     * @param channelContext the context to retrieve data from.
+     * @return the size of plaintext blocks in bytes. Returns 0 if no key length is known.
+     */
+    size_t (*getRemotePlainTextBlockSize)(const UA_SecurityPolicy *securityPolicy,
+                                          const void *channelContext);
+} UA_SecurityPolicyEncryptionAlgorithm;
+
+typedef struct {
```
jpfr: The CryptoModule is a very shallow container now.
Imho we can get rid of the indirection and have

- symmetricSignatureAlgorithm
- symmetricEncryptionAlgorithm
- asymmetricSignatureAlgorithm
- ssymmetricEncryptionAlgorithm

side by side.

Infinity95: I kept it because it's easier passing both algorithms around in the securechannel that way.
I can remove it though.

plugins/ua_securitypolicy_basic128rsa15.c
Infinity95: I refactored the getEncryptionOverhead function out of the securitypolicy into the securechannel since it is the same for all policies. We could make it part of a ua_plugin_securitypolicy.c file i guess?

include/ua_plugin_securitypolicy.h
Infinity95: This was inconsistent with the rest of the function pointer declarations and imho looks better with the space.

jpfr: Right, the look is better.
Still, can you move purely cosmetic changes into a different commit from the functional changes?

include/ua_plugin_securitypolicy.h
jpfr: Okay, that sounds reasonable.

include/ua_plugin_securitypolicy.h
```diff
@@ -309,6 +363,21 @@ typedef struct {
     UA_EndpointDescription endpointDescription;
 } UA_Endpoint;
 
+/* Gets the number of bytes that are needed by the encryption function in
+     * addition to the length of the plaintext message. This is needed, since
```
jpfr: Indendation issues..

src/server/ua_securechannel_manager.c
```diff
@@ -210,10 +211,9 @@ UA_SecureChannelManager_renew(UA_SecureChannelManager* cm, UA_SecureChannel *cha
     UA_ByteString_deleteMembers(&channel->remoteNonce);
     UA_ByteString_copy(&request->clientNonce, &channel->remoteNonce);
 
-    const size_t keyLength = channel->securityPolicy->symmetricModule.cryptoModule.
-        getLocalEncryptionKeyLength(channel->securityPolicy, channel->channelContext);
     UA_ByteString_deleteMembers(&channel->localNonce);
-    UA_SecureChannel_generateNonce(channel, keyLength, &channel->localNonce);
```
jpfr: This can be simplified to `UA_SecureChannel_generateNonce(channel, &channel->localNonce)`.
The SecureChannel has access to the policy internally.

Then we don't have to include `ua_plugin_securitypolicy.h` here. And the code remains decoupled.

src/server/ua_services_session.c
```diff
@@ -245,7 +245,7 @@ Service_ActivateSession(UA_Server *server, UA_SecureChannel *channel,
     if(session->validTill < UA_DateTime_nowMonotonic()) {
         UA_LOG_INFO_SESSION(server->config.logger, session,
                             "ActivateSession: SecureChannel %i wants "
-                            "to activate, but the session has timed out",
+                                "to activate, but the session has timed out",
```
jpfr: ?

include/ua_plugin_securitypolicy.h
Infinity95: A different commit or a different PR alltogether?

plugins/ua_plugin_securitypolicy.c
```diff
@@ -0,0 +1,25 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ *    Copyright 2018 (c) Mark Giraud, Fraunhofer IOSB
+ */
+
+#include "ua_plugin_securitypolicy.h"
+
+size_t
+UA_SecurityPolicy_getRemoteAsymEncryptionBufferLengthOverhead(const UA_SecurityPolicy *securityPolicy,
```
jpfr: Hmm. Now we have a hard binding between a plugin and the main library.
This implementation is not intended to be exchanged.

I think it is better to stay with a per-policy implementation as before.

tests/testing-plugins/testing_policy.c
```diff
@@ -114,9 +139,27 @@ sym_getLocalEncryptionKeyLength_testing(const UA_SecurityPolicy *securityPolicy,
...
+    ck_assert(channelContext != NULL);
+    return keySizes->sym_enc_blockSize;
+}
+
+static size_t
+sym_getRemoteEncryptionBlockSize_testing(const UA_SecurityPolicy *securityPolicy,
+                                         const void *channelContext) {
+    ck_assert(securityPolicy != NULL);
+    ck_assert(channelContext != NULL);
+    return keySizes->sym_enc_blockSize; // TODO: Different size for remote
```
jpfr: Is there an example where th remote block size is different from the local block size?
If not, we can unify the two methods.

Infinity95: In the symmetric case, no. In the asymmetric case yes, if the remote cert for example has 2048 bits and the local one 1028.

plugins/ua_plugin_securitypolicy.c
Infinity95: Then every policy has to implement the same function. I don't think that's a good idea.
The implementation is in a plugin source which can be used by all other policies.

src/server/ua_services_session.c
Infinity95: This is a formatter bug in my IDE. I try to avoid commiting them but they slip through rather often.

include/ua_plugin_securitypolicy.h
jpfr: Commit is enough.

plugins/ua_plugin_securitypolicy.c
jpfr: > Now we have a hard binding between a plugin and the main library.

I assumed this method is also called from the main library.
Since you put it into a public header where it can be seen by the plugins and by the main lib.
But it is only used by the plugins.

How about we add a file `/plugins/ua_securitypolicy_helper.h` (or similar) that collects common functionality.
So we don't increase the API surface of the library but still reuse code between plugins.

Infinity95: That's also possible. But the function is used by the main library and the plugins (in securechannel.c and in basic128rsa15.c)

jpfr: Right. I found out why I was confused.
I have to click "load" on securechannel.c to see the full diff on github.
Otherwise ctrl-c ctrl-f doesn't find the usage.

I don't care much about which way this goes.
But .. there should be a clear separation between the plugins and the main lib.

The plugins are meant to be exchangeable with no hard dependencies from the main lib.
Any interaction between code in /src and /plugins must go through a public plugin interface.

Infinity95: Isn't that the case right now? The function is part of the public plugin api but has only one implementation in a general plugin source file.
Another approach would be to put it into the src part and make it available as part of the public api.

Imo the current solution should be sufficient and avoids having extra files that only contain a few lines of code. I think we could maybe refactor one or two functions that currently reside in the securechannel into the security policy just like the getOverhead function. I'll have to check though.

jpfr: > I think we could maybe refactor one or two functions that currently reside in the securechannel into the security policy just like the getOverhead function. I'll have to check though.

👍

Infinity95:
@jpfr Can you check once more if the commit structure is fine now?
If everything is green you can merge if you approve.

jpfr:
Some repeated code, e.g. for nonce generation is moved to reusable functions.
@Infinity95 Can you have a look at the latest changes?

Other than that, your API and PR are ready to merge.

src/server/ua_services_session.c
```diff
@@ -186,56 +166,27 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
...
+       channel->securityMode != UA_MESSAGESECURITYMODE_SIGNANDENCRYPT)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    if(!channel->securityPolicy)
+        return UA_STATUSCODE_BADINTERNALERROR;
+    const UA_SecurityPolicy *securityPolicy = channel->securityPolicy;
+    const UA_ByteString *localCertificate = &securityPolicy->localCertificate;
+
+    size_t dataToVerifySize = localCertificate->length + session->serverNonce.length;
+    UA_ByteString dataToVerify = {dataToVerifySize, (UA_Byte*)UA_alloca(dataToVerifySize)};
```
Infinity95: Baaaad idea. Certificates can be 500-1000 bytes size. Allocating that on the stack isn't a good idea imho. Furthermore it's user supplied data and the size can be set to any desired length.

Infinity95: Okay i just saw the check you added. I don't think its a good idea to limit the size to 4096.
Why are we not using malloc here? I see no reason not to.

tests/testing-plugins/testing_policy.c
```diff
@@ -243,8 +243,6 @@ generateNonce_testing(const UA_SecurityPolicy *securityPolicy,
                       UA_ByteString *out) {
     ck_assert(securityPolicy != NULL);
     ck_assert(out != NULL);
-    if(out->length != 0)
```
Infinity95: Why remove this check?

jpfr: You're right. It could be left in.

The nonce is generated with UA_ByteString_alloc.
If that invariant fails, we have loads of other problems as well.
It should be detected basically everywhere.

src/server/ua_services_session.c
jpfr: - I hate branches (if/else)
- Removing branches makes me happy
- Alloca requires less branches

I added a TODO that refers to MaxSenderCertificateSize from part 6.
If you want to add that or a solution with malloc, feel free to do so.

PR #1522, opened 2018-01-19
Additional logic of setting monitoring mode

ChaJiWon:
fix #1520  issue.

This commit is related with service specification (Part4)

```
7.18 Monitoring Mode

DISABLED_0
The item being monitored is not sampled or evaluated, and Notifications are not generated or queued. Notification reporting is disabled.

REPORTING_2
The item being monitored is sampled and evaluated, and Notifications are generated and queued. Notification reporting is enabled.
```

PR #1614, opened 2018-02-23
fix #1522

StalderT:
fix #1522 compilation
https://travis-ci.org/open62541/open62541/jobs/345152138#L1849

and 

```
If Operation_SetMonitoringMode is called with

    UA_MONITORINGMODE_REPORTING
    UA_MONITORINGMODE_SAMPLING
    UA_MONITORINGMODE_REPORTING

So the callback is registered twice.
```

### DIFF
--- a/src/server/ua_services_subscription.c
+++ b/src/server/ua_services_subscription.c
@@ -394,9 +394,15 @@ Operation_SetMonitoringMode(UA_Server *server, UA_Session *session,
         *result = UA_STATUSCODE_BADMONITOREDITEMIDINVALID;
         return;
     }
+
+    if(mon->monitoredItemType != UA_MONITOREDITEMTYPE_CHANGENOTIFY) {
+        *result = UA_STATUSCODE_BADNOTIMPLEMENTED;
+        return;
+    }
   
     /* check monitoringMode is valid or not */
     if(smc->monitoringMode > UA_MONITORINGMODE_REPORTING) {
+        *result = UA_STATUSCODE_BADMONITORINGMODEINVALID;
         return;
     }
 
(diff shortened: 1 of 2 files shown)


src/server/ua_services_subscription.c
```diff
@@ -388,6 +388,7 @@ static void
 Operation_SetMonitoringMode(UA_Server *server, UA_Session *session,
                             struct setMonitoringContext *smc,
                             UA_UInt32 *monitoredItemId, UA_StatusCode *result) {
+    // TODO check monitoredItemType
```
Pro: What do you mean by this comment?

StalderT: I do not know if we are concerned here, but do something like this:

https://github.com/open62541/open62541/blob/c3885fb0cebd5502df38cdf0d29fd56e098562d6/src/server/ua_subscription_datachange.c#L35-L52

to see with @aDogCalledSpot https://github.com/open62541/open62541/pull/1560

Pro: Ah, sorry, now I got what you mean. For now I would check for the itemType and if it isn't `CHANGENOTIFY` then set the status code to BadUnsupportedService or BadNotImplemented or something similar

src/server/ua_services_subscription.c
```diff
@@ -403,22 +404,23 @@ Operation_SetMonitoringMode(UA_Server *server, UA_Session *session,
     if(mon->monitoringMode == smc->monitoringMode)
         return;
 
-    mon->monitoringMode = smc->monitoringMode;
     if(mon->monitoringMode == UA_MONITORINGMODE_REPORTING) {
```
Pro: Shouldn't this be 'if(smc->monitoringMode == ...)`? Same for the next if?
smc contains the new monitoring mode, so if the mode is changed, e.g. set to disabled, the queue should be deleted, and not if the monitored item is already deleted? Therefore the previous code was correct (except the val->data stuff), or am I wrong?

PR #1618, opened 2018-02-26
Use node ID to make sure refcount is updated

Pro:
This addresses the comment here:
https://github.com/open62541/open62541/pull/1603#issuecomment-368455022

Additionally, @janitza-thbe requested an additional method which allows to get the parent of a specific node.
@jpfr Should we add it as a public API?
```c
static UA_StatusCode
UA_Server_getNodeParent(UA_Server *server, const UA_Node *node, UA_NodeId **parentNodeId, UA_NodeId **parentReferenceId) {
    for (size_t i=0; i< node->referencesSize; i++) {
        if (node->references[i].isInverse == UA_TRUE) {
            for (uint8_t r =0; r<UA_PARENT_REFERENCES_COUNT; r++) {
                // TODO handle subtypes of parent reference types
                if (UA_NodeId_equal(&node->references[i].referenceTypeId, &parentReferences[r])) {
                    *parentNodeId = &node->references[i].targetIds[0].nodeId;
                    *parentReferenceId = &parentReferences[r];
                    return UA_STATUSCODE_GOOD;
                }
            }
        }
    }

    return UA_STATUSCODE_BADNOTFOUND;
}
```

### DIFF
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -513,6 +513,13 @@ copyChildNode(UA_Server *server, UA_Session *session,
             return retval;
         }
 
+        /* Add all the children of this child to the new child node to make sure we take
+        * the values from the nearest inherited object first.
+        * The call to addNode_finish will then only add the children from the type and
+        * thus skip the direct children of rd->nodeId.nodeId
+        */
+        copyChildNodes(server, session, &rd->nodeId.nodeId, &newNodeId);
+
         /* Add the parent reference */
         /* we pass the nodeId instead of node to make sure the refcount
          * is increased and other calls can not delete the node in the meantime */
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -55,6 +55,19 @@ UA_Server_setNodeContext(UA_Server *server, UA_NodeId nodeId,
 /* Consistency Checks */
 /**********************/
 
+
+#define UA_PARENT_REFERENCES_COUNT 2
+
+UA_NodeId parentReferences[UA_PARENT_REFERENCES_COUNT] = {
+    {
+        0, UA_NODEIDTYPE_NUMERIC, {UA_NS0ID_HASSUBTYPE}
+    },
+    {
+        0, UA_NODEIDTYPE_NUMERIC, {UA_NS0ID_HASCOMPONENT}
+    }
+};
+
+
 /* Check if the requested parent node exists, has the right node class and is
  * referenced with an allowed (hierarchical) reference type. For "type" nodes,
  * only hasSubType references are allowed. */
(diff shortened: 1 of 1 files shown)


janitza-thbe:
Thanks, @Pro!
I don't mind if it keeps internal API.

Just a few last hints to the changes:
- In `copyChildNode()->addParentAndTypeRef()` you're still using the `node` (`&node->nodeId`). It should be `&newNodeId`
- In `Operation_addNode_begin()->addParentAndTypeRef()` you're still using the `node` (`&node->nodeId`). The node ID should be copied before.
- `parentReferences` should be `const`
- `UA_PARENT_REFERENCES_COUNT` should be automatically calculated by the preprocessor by using `(sizeof(parentReferences) / sizeof(UA_NodeId))`

Pro:
For your second point I'm not sure if it really makes sense here to copy the node ID to cover that specific multithreading race condition. When should we copy it? Right after the call to _insert we still have the same problem that node may be deleted by some other method. So the only way to handle it would be to pass a local nodeId to the _insert and then copy it to `outNewNodeId` if not NULL. But is this really necessary?`There are probably many other places where we have a similar problem...

I used a define for `UA_PARENT_REFERENCES_COUNT` to keep the code easier to read and probably more efficient.

Regarding the additional `copyChildNodes` call in `copyChildNode` (as @janitza-thbe wrote via mail):
It is required to inherit direct mandatory children of a child itself. The following call to _finish will then only add the mandatory children of the node's type. See also the unit test `check_node_inheritance`.

janitza-thbe:
```
Regarding the additional copyChildNodes call in copyChildNode:
It is required to inherit direct mandatory children of a child itself. The following call to _finish will then only add the mandatory children of the node's type. See also the unit test check_node_inheritance.
```
Yepp, I realised it in meanwhile, too. I excluded the function call to `copyChildNodes()` now in my special case during adding nodes from my generated node code.

Indeed it's a serious problem that the code isn't almost Thread-Safe (a difference to _Multithreading_ which means both multiple worker threads and implicated thread-safeness). There is still a lot of work to invest. But I think it's better to fix possible race condition problems now, where found, instead of really expensive debugging work later.

janitza-thbe:
By debugging I realised that at least `UA_NS0ID_HASPROPERTY` is missing as possible parent reference type.
It seems not to be trivial to reverse lookup the parent reference. Isn't in fact _every_ hierarchical reference allowed as parent reference? I'm a bit confused and don't have the standard but my application isn't working any more after the changes and complains about missing parent references that couldn't be look up.

Pro:
I will for now add a TODO to remind us later when we harden the stack for multithreading. Then we probably need to make sure that you can not call the delete service, before the add service finished, which then resolves this problem.

Regarding the parent references it's then probably the best solution to search for any subtype of hierarchical reference. I would suggest that you add this method on your fork, since currently in the main stack such a method is not required. It could be that we will add it in the future.

janitza-thbe:
Full agreement. I already implemented it and added even ´HAS_PROPERTY´ to `parentReferences`.
I didn't investigated which effect has this missing reference type on usage of `parentReferences` in `typeCheckVariableNode` and `addParentAndTypeRef`.

PR #1621, opened 2018-02-28
NodesetCompiler use Byte array instead of string

### LINKED ISSUE #1254: Use statically initialized char array instead of huge string for nodeset generation
This is an issue which can be grabbed by first time contributors and is fairly easy to implement.

Currently the nodeset compiler creates a huge string for ByteArray initialization which may cause issues on compilers which have a string limit of 65535 bytes.

To circumvent this, we should use static char arrays.

E.g. currently we use:

`UA_ByteString variable = UA_BYTESTRING_ALLOC("A very huge string");`

This should then be something like:

```
char stringArr[230] = [ 102, 30, 35, 130, ...... ];
UA_ByteString variable;
variable.length = 230;
variable.data = UA_malloc(variable.length);
memcpy(&stringArr, &variable.data, variable.length);
```

Or even better if the stringArray is initialized on the global stack and the byte array just references it (which would avoid malloc and memcopy).

nizamogluyekta:
Due to issue #1254 : The huge strings are changed as arrays of ASCII values of each element inside for ByteArrays. And also some implementations are done due to task.

Fixes #1254

### DIFF
--- a/tools/nodeset_compiler/backend_open62541_datatypes.py
+++ b/tools/nodeset_compiler/backend_open62541_datatypes.py
@@ -11,11 +11,6 @@ def generateBooleanCode(value):
         return "true"
     return "false"
 
-def stringtoByteArray(value):
-    value = value.strip()
-    byteform = [ord(c) for c in value] #transform every char to ASCII
-    return byteform
-
 def makeCLiteral(value):
     return value.replace('\\', r'\\\\').replace('\n', r'\\n').replace('\r', r'')
 
@@ -44,14 +39,15 @@ def generateXmlElementCode(value, alloc=False):
     value = makeCLiteral(value)
     return u"UA_XMLELEMENT{}({})".format("_ALLOC" if alloc else "", splitStringLiterals(value))
 
-def generateByteStringCode(value, valueName):
-    asciiarray = stringtoByteArray(value)
+def generateByteStringCode(value, valueName, global_var_code):
+    asciiarray = [ord(c) for c in value.strip()]
     asciiarraystr = str(asciiarray).rstrip(']').lstrip('[')
-    return "UA_Byte stringArr[{len}] = {{{data}}};\n{instance}->length = {len};\n{instance}->data = stringArr;"\
-                                                .format(len=len(asciiarray), data=asciiarraystr, instance=valueName)
-
-    #here the info is returned in the form we want the code to take. after this line the code will appear just as we wished it to be.
-
+    global_var_code.append("static const UA_Byte {instance}_byteArray[{len}] = {{{data}}};".format(
+        len=len(asciiarray), data=asciiarraystr, instance=valueName
+    ))
+    # Cast away const with '(UA_Byte *)(void*)(uintptr_t)' since we know that UA_Server_addNode_begin will copy the content
+    return "{instance}->length = {len};\n{instance}->data = (UA_Byte *)(void*)(uintptr_t){instance}_byteArray;"\
+                                                .format(len=len(asciiarray), instance=valueName)
 
 def generateLocalizedTextCode(value, alloc=False):
     vt = makeCLiteral(value.text)
@@ -86,7 +82,7 @@ def generateDateTimeCode(value):
     mSecsSinceEpoch = int((value - epoch).total_seconds() * 1000.0)
     return "( (UA_DateTime)(" + str(mSecsSinceEpoch) + " * UA_DATETIME_MSEC) + UA_DATETIME_UNIX_EPOCH)"
 
-def generateNodeValueCode(prepend , node, instanceName, valueName, asIndirect=False):
+def generateNodeValueCode(prepend , node, instanceName, valueName, global_var_code, asIndirect=False):
     if type(node) in [Boolean, Byte, SByte, Int16, UInt16, Int32, UInt32, Int64, UInt64, Float, Double]:
         return prepend + "(UA_" + node.__class__.__name__ + ") " + str(node.value) + ";"
     elif type(node) == String:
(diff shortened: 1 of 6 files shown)


tools/nodeset_compiler/backend_open62541_datatypes.py
```diff
@@ -40,7 +46,15 @@ def generateXmlElementCode(value, alloc=False, max_string_length=0):
     return "UA_XMLELEMENT{}({})".format("_ALLOC" if alloc else "", splitStringLiterals(value, max_string_length=max_string_length))
 
 def generateByteStringCode(value, alloc=False, max_string_length=0):
-    return "UA_BYTESTRING{}({})".format("_ALLOC" if alloc else "", splitStringLiterals(value, max_string_length=max_string_length))
+   return "char stringArr?{}! = {};\n\
```
Pro: Why do you use this cryptic `?` and `!` notation and replace it later? Can't you directly use `[]`? Adding the byte array then has to use the correct brackets.

tools/nodeset_compiler/backend_open62541_datatypes.py
```diff
@@ -40,7 +46,15 @@ def generateXmlElementCode(value, alloc=False, max_string_length=0):
...
+    UA_ByteString variable;\n\
+    variable.length = {};\n\
+    variable.data = (UA_Byte *)UA_malloc(variable.length);\n\
+    memcpy(&stringArr, &variable.data, variable.length)".format(len(stringtoByteArray(value, max_string_length=max_string_length)) \
+                                                                ,stringtoByteArray(value, max_string_length=max_string_length),\
+                                                                len(stringtoByteArray(value, max_string_length=max_string_length)) )
```
Pro: Try to avoid calling stringtoByteArray three times. Just store the result before and then use the resulting value here.

tools/nodeset_compiler/backend_open62541_datatypes.py
```diff
@@ -71,26 +85,30 @@ def generateDateTimeCode(value):
...
+        return prepend + generateStringCode(node.value, alloc=asIndirect, max_string_length=max_string_length)
     elif type(node) == XmlElement:
-        return generateXmlElementCode(node.value, alloc=asIndirect, max_string_length=max_string_length)
+        return prepend + generateXmlElementCode(node.value, alloc=asIndirect, max_string_length=max_string_length)
     elif type(node) == ByteString:
         # replace whitespaces between tags and remove newlines
-        return "UA_BYTESTRING_NULL" if not node.value else generateByteStringCode(re.sub(r">\s*<", "><", re.sub(r"[\r\n]+", "", node.value)), alloc=asIndirect, max_string_length=max_string_length)
+        return prepend + "UA_BYTESTRING_NULL" if not node.value else generateByteStringCode(re.sub(r">\s*<", "><", re.sub(r"[\r\n]+", "", node.value)),
+                                      alloc=asIndirect, max_string_length=max_string_length).replace("[","{").replace("]","}")\
+                                                                                            .replace("?","[").replace("!","]")
```
Pro: See previous comment, then remove the last two replace operations.

tools/nodeset_compiler/backend_open62541_datatypes.py
Pro: Use 4 spaces indentation

Pro: The alloc parameter is not used anymore, remove it from the parameters

tools/nodeset_compiler/backend_open62541_nodes.py
```diff
@@ -521,7 +521,7 @@ def generateNodeCode_begin(node, nodeset, max_string_length, generate_ns0, paren
         code.append("UA_NODEID_NULL,")
     code.append("(const UA_NodeAttributes*)&attr, &UA_TYPES[UA_TYPES_{}ATTRIBUTES],NULL, NULL);".format(node.__class__.__name__.upper().replace("NODE" ,"")))
     code.extend(codeCleanup)
-    
+
```
Pro: Try to avoid changes like this

tools/nodeset_compiler/backend_open62541_datatypes.py
```diff
@@ -40,7 +46,15 @@ def generateXmlElementCode(value, alloc=False, max_string_length=0):
     return "UA_XMLELEMENT{}({})".format("_ALLOC" if alloc else "", splitStringLiterals(value, max_string_length=max_string_length))
 
 def generateByteStringCode(value, alloc=False, max_string_length=0):
-    return "UA_BYTESTRING{}({})".format("_ALLOC" if alloc else "", splitStringLiterals(value, max_string_length=max_string_length))
+   return "char stringArr?{}! = {};\n\
+    UA_ByteString variable;\n\
+    variable.length = {};\n\
+    variable.data = (UA_Byte *)UA_malloc(variable.length);\n\
```
Pro: Add a check if `variable.data` is NULL and return `UA_STATUSCODE_BADOUTOFMEMORY`.

jpfr:
> I thought about adding an additional UA_VariantStorageType like UA_VARIANT_DATA_NOCOPY where UA_Variant_copy takes the pointer instead of copying it.

Is the additional complexity worth it?

If we need absolute minimum resource consumption, we could implement the binary nodeset file format. We could have a special nodestore that uses zero initialization and generates node only when they are used.

http://documentation.unified-automation.com/uasdkhp/1.1.1/html/md_opcua_binary_fileformat.html

Rumor has the binary nodeset format will play an important role in the standard.

Pro:
Ok, then I will try to get CI green again with the current status, and skip the issue of double memory usage for the (huge) nodes.

PR #1622, opened 2018-02-28
Fix memoryleak in certificateVerification_verify

Infinity95:
Stumbled across this while implementing valgrind monitoring for fuzzing.

### DIFF
--- a/plugins/ua_pki_certificate.c
+++ b/plugins/ua_pki_certificate.c
@@ -77,6 +77,7 @@ certificateVerification_verify(void *verificationContext,
         /*              UA_LOGCATEGORY_SECURITYPOLICY, */
         /*              "Verifying the certificate failed with error: %s", buff); */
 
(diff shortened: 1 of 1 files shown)


Pro:
Ah cool, thanks.

Just for clarification:
What exactly do you mean by 'valgrind monitoring for fuzzing'? Note that oss-fuzz already uses its own memory leak detector from clang (Address Sanitizer, Memory Sanitizer) so there shouldn't be any need for additional valgrind checks?

OSS-fuzz already compiles with encryption enabled. Here we just need to add some corpus entries for valid certificates and then oss-fuzz will also test the whole certificate/encryption code.

Infinity95:
It's just a little side experiment in my bachelor's thesis. The main focus is blackbox fuzzing over the network of opcua devices.
Although i'm testing my implementation with open62541 since i'm familiar with the code and can better identify if something is wrong with the fuzzing component or with the stack.

PR #1624, opened 2018-03-01
Added highlevel API for historical access

Root-Core:
At the moment, it allows to read a single node historically.
It could be altered to support multiple nodes if desired, but this could make the API slightly more complex.

The `readAtTime` function is "missing", as I havn't found a server supporting it to test against.

An example is included, which relies on `Unified Automations OPC UA ANSI C Demo Server` atm, but as I'm working on server side historical access this could change.

Regards.

### DIFF
--- a/examples/client_historical.c
+++ b/examples/client_historical.c
@@ -128,11 +128,11 @@ int main(int argc, char *argv[]) {
     /* Read historical values */
     printf("\nStart historical read (4, \"Demo.History.ByteWithHistory\"):\n");
     retval = UA_Client_readHistorical(client, UA_NODEID_STRING(4, "Demo.History.ByteWithHistory"), readHist,
-        UA_DateTime_fromUnixTime(0), UA_DateTime_now(), 10, false, UA_TIMESTAMPSTORETURN_BOTH, (void *)UA_FALSE);
+        UA_DateTime_fromUnixTime(0), UA_DateTime_now(), false, 10, false, UA_TIMESTAMPSTORETURN_BOTH, (void *)UA_FALSE);
 
     printf("\nStart historical read (4, \"Demo.History.DoubleWithHistory\"):\n");
     retval = UA_Client_readHistorical(client, UA_NODEID_STRING(4, "Demo.History.DoubleWithHistory"), readHist,
-        UA_DateTime_fromUnixTime(0), UA_DateTime_now(), 10, false, UA_TIMESTAMPSTORETURN_BOTH, (void *)UA_TRUE);
+        UA_DateTime_fromUnixTime(0), UA_DateTime_now(), false, 10, false, UA_TIMESTAMPSTORETURN_BOTH, (void *)UA_TRUE);
 
 cleanup:
     UA_Client_disconnect(client);
--- a/examples/client_historical.c
+++ b/examples/client_historical.c
@@ -1,0 +1,141 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+#include <stdio.h>
+#include "open62541.h"
+
+#ifdef _WIN32
+# include <windows.h>
+# define UA_sleep_ms(X) Sleep(X)
+#else
+# include <unistd.h>
+# define UA_sleep_ms(X) usleep(X * 1000)
+#endif
+
+static UA_Boolean
+readHist(const UA_NodeId nodeId, const UA_Boolean isInverse,
+         const UA_Boolean moreDataAvailable,
+         const UA_HistoryData *data, void *isDouble) {
+
+    printf("\nRead historical callback:\n");
+    printf("\tValue count:\t%d\n", data->dataValuesSize);
+    printf("\tIs inverse:\t%d\n", isInverse);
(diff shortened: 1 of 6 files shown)


include/ua_client.h
```diff
@@ -258,6 +258,17 @@ UA_Client_Service_write(UA_Client *client, const UA_WriteRequest request) {
     return response;
 }
 
+/**
+* Historical Access Service Set
+* ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ */
+static UA_INLINE UA_HistoryReadResponse
+UA_Client_Service_history_read(UA_Client *client, const UA_HistoryReadRequest request) {
```
jpfr: CamlCase `UA_Client_Service_historyRead`

include/ua_client_highlevel.h
```diff
@@ -212,6 +212,67 @@ UA_Client_readUserExecutableAttribute(UA_Client *client, const UA_NodeId nodeId,
...
 }
 
+/**
+ * Historical Access
+ * ^^^^^^^^^^^^^^^^^
+ * The following functions can be used to read a single node historically.
+ * Use the regular service to read several nodes at once.
+ */
+typedef UA_Boolean
+(*UA_HistoricalIteratorCallback)(const UA_NodeId nodeId, const UA_Boolean isInverse,
```
jpfr: - Add the client pointer as the first argument
- The nodeId can be a const pointer
- Integers don't have to be const
- rename "handle" to "callbackContext"

What is the meaning of moreDataAvailable?
There is a continuation in the server?

What if the client fails after signaling moreDataAvailable?
We could model the "continuation" as a persistent context also on the client side.
And close the "continuation" with a status code in the callback if an error occurs.

include/ua_client_highlevel.h
```diff
@@ -212,6 +212,67 @@ UA_Client_readUserExecutableAttribute(UA_Client *client, const UA_NodeId nodeId,
...
+                                 const UA_Boolean moreDataAvailable,
+                                 const UA_HistoryData *data, void *handle);
+
+/* Don't call this function, use the typed versions */
+UA_HistoryReadResponse UA_EXPORT
+__UA_Client_readHistorical(UA_Client *client, const UA_NodeId nodeId,
```
jpfr: I don't see any code using this "hidden" function. E.g. in a static inline method.
Do we need the __ at the beginning?

Rename to `...historyRead`

include/ua_client_highlevel.h
```diff
@@ -212,6 +212,67 @@ UA_Client_readUserExecutableAttribute(UA_Client *client, const UA_NodeId nodeId,
...
+                           void* details, const UA_TimestampsToReturn timestampsToReturn,
+                           const UA_ByteString continuationPoint, const UA_Boolean releaseConti);
+
+/* Don't call this function, use the typed versions */
+UA_StatusCode UA_EXPORT
```
jpfr: What does the details pointer do?

jpfr:
> The readAtTime function is "missing", as I havn't found a server supporting it to test against.

Do we have to explicitly have to support anything?
The parameters of historyRead are an ExtensionObject of the following type:

grafik

We can make a version of the service that takes a user-defined ExtensionObject.
The "normal" methods are then special cases.

examples/client_historical.c
```diff
@@ -0,0 +1,160 @@
...
+    }
+    else {
+        printf("Failed to read data logging status.\n");
+        UA_Variant_delete(val);
+        goto cleanup;
+    }
+    UA_Variant_delete(val);
+
+#ifdef UA_ENABLE_METHODCALLS
+    /* Active server side data logging via remote method call */
```
fmeerkoetter: should this read "Activate server side data..."

examples/client_historical.c
```diff
@@ -0,0 +1,160 @@
...
+            goto cleanup;
+        }
+
+        /* The server logs a value every 50ms by default */
+        printf("Successfully started data logging. Let the server collect data for 2000ms..\n");
+        UA_sleep_ms(2000);
+    }
+#else
+    if (!active) {
+        printf("Method calling is not allowed, you have to active the data logging manually.\n");
```
fmeerkoetter: "allowed" is slightly misleading. the feature is not available in this build

ccvca:
Is this Pull-Request replaced by #1740?

Root-Core:
@ccvca No, as this PR implements a client functionality. And #1740 does implement the server functionality.

@jpfr Ofc, we don't need to support all the functions. It was just a hint for completeness sake.
I think the high level approach here is making things easy, so I would like to stay true to it.

I will update the PR in the next days / next week with the requested changes.
I had some health issues the last weeks, so I was not able to work on anything.

jpfr:
Hey @Root-Core ,

it would be great to merge this for the client!

We recently merged #1915.
The datatype defintiions could be reused.

And we can write against the server-side implementations.

basyskom-meerkoetter:
@Root-Core would you mind some help getting this PR on master? @peter-rustler-basyskom (who did the server side) has some free time at his hand. It would be nice to "complete" the HDA story in open62541 by having the client and the server side.

peter-rustler-basyskom:
This PR is obsolete.
All what is here is already merged with #2282.
Please close it.

PR #1627, opened 2018-03-01
Typo in cmake option: Trailing S missing in BUILD_SHARED_LIBS.

bobbthebuilder:
This typo introduced the following bug:

When building open62541 against mbedtls, the compiler options
for the C Runtime-Libraries mbedtls is built with do not match
the ones for open62541 anymore resulting in linker errors.
This bug is MSVC-related only.


jpfr:
Thanks for the fix!

PR #1630, opened 2018-03-02
Fix spec file of RPM build

ctron:
This change does add the additional build time dependency on
python-six and also corrects the directory name of the cmake
files.


jpfr:
Thx. 😄

Issue #1334, opened 2017-11-09
"stress" client and server question


StalderT:
Description
===========

When using [s.txt](https://github.com/open62541/open62541/files/1457481/s.txt) [c.txt](https://github.com/open62541/open62541/files/1457482/c.txt), client produce : 

```
test has changed 10000 42684!
test has changed 9000 40684!
test has changed 8000 38684!
test has changed 7000 30119!
test has changed 6000 31119!
test has changed 5000 36651!
test has changed 4000 46935!
test has changed 3000 47935!
test has changed 2000 48935!
test has changed 1 50934!
test has changed 10000 42685!
test has changed 9000 43685!
test has changed 8000 44685!
test has changed 7000 52467!
test has changed 6000 51467!
test has changed 5000 45935!
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131547055686963030 after 131547055691876790 diff 4913760
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131547055691878600 after 131547055693162530 diff 1283930
test has changed 10000 62684!
test has changed 9000 63685!
test has changed 8000 64685!
test has changed 7000 72467!
test has changed 6000 71467!
test has changed 5000 65935!
test has changed 4000 66935!
test has changed 3000 67935!
test has changed 2000 68935!
test has changed 1 70934!
test has changed 10000 62685!
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131547055693166770 after 131547055702124200 diff 8957430
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131547055702125710 after 131547055703172880 diff 1047170
test has changed 10000 82684!
test has changed 9000 83685!
test has changed 8000 84685!
test has changed 7000 92467!
test has changed 6000 91467!
test has changed 5000 85935!
test has changed 4000 86935!
test has changed 3000 87935!
test has changed 2000 88935!
test has changed 1 90934!
test has changed 10000 82685!
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131547055703177850 after 131547055712333420 diff 9155570
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131547055712334830 after 131547055713166660 diff 831830
```

before
```
UA_Client_Subscriptions_manuallySendPublishRequest retval 0 0 before 131547055686963030 after 131547055691876790 diff 4913760
```
we have ```test has changed 5000 45935!```

[...]

PR #1632, opened 2018-03-03
modify notification queues in the server fix #1631

### LINKED ISSUE #1631: Notification Queues in the Server
The latest changes to the subscriptions handling added two more DateTime fields that have to be considered for the publish. This is getting more and more complicated.

The following could remove all of these special cases.
**Each `MonitoredItem_queuedValue` is put into two TAILQ: One for the MonitoredItem and one global TAILQ for the entire subscription**. The global TAILQ is always ordered according to the creation date of the notification. Indepent of which MonitoredItem is the origin.

Then, for publishing, we simply iterate over the global (subscription) TAILQ in order.
No funky skipping of MonitoredItems and/or Notifications based on timestamps.

@StalderT You know this code very well. Anything I missed?

StalderT:
#1631


src/server/ua_subscription.h
```diff
@@ -76,15 +66,31 @@ typedef struct UA_MonitoredItem {
     QueuedValueQueue queue;
 } UA_MonitoredItem;
 
+typedef struct MonitoredItem_queuedValue {
+    TAILQ_ENTRY(MonitoredItem_queuedValue) listEntry;
+    QueuedValueQueue *listHead;
```
jpfr: Getting the listhead is implicit in the `mon` pointer.

src/server/ua_subscription.h
```diff
@@ -130,14 +136,13 @@ struct UA_Subscription {
...
 
     /* MonitoredItems */
     LIST_HEAD(UA_ListOfUAMonitoredItems, UA_MonitoredItem) monitoredItems;
-    /* When the last publish response could not hold all available
-     * notifications, in the next iteration, start at the monitoreditem with
-     * this id. If zero, start at the first monitoreditem. */
-    UA_UInt32 nextMonitoredItemIdToBrowse;
+
+    QueuedValueQueue notificationQueue;
+    UA_UInt32 readyNotifications;
```
jpfr: Can you add a sentence on the difference on "ready" and "pending"?

src/server/ua_subscription.h
```diff
@@ -76,15 +66,31 @@ typedef struct UA_MonitoredItem {
     QueuedValueQueue queue;
 } UA_MonitoredItem;
 
+typedef struct MonitoredItem_queuedValue {
```
jpfr: Can we rename this struct to simply "notification"?

src/server/ua_subscription.h
StalderT: Thanks ! Fixed.

StalderT:
Using this sample server : https://github.com/open62541/open62541/files/1457481/s.txt and monitoring 10'000 items with UaExpert (publishingInterval 500, max notifications per publish 0, sampling interval 250).

With current master, the server use 26% of CPU
With the PR, the server use 18% of CPU

jpfr:
> We have a gain of 30%.

Wow!!

StalderT:
For me, done.

StalderT:
https://github.com/open62541/open62541/pull/1632/commits/f60e72845d52049890690458da97953d0169950c reintroduce https://github.com/open62541/open62541/issues/1334

jpfr:
Did you see the error or is this theoretical?

Since new notifications are added to the end of the global TAILQ and published out from the beginning of the TAILQ, how is it possible that the error remains?

I quote your explanation:
> Because, if we have, many monitored items (more than notificationsPerPublish) who always changing, if a monitored items (already sended) is sampled bettween two UA_SUBSCRIPTIONSTATE_LATE publish request/response, then the sample monitored item is send in the next UA_SUBSCRIPTIONSTATE_LATE publish/request and cause that the server continuously send publish response (cause bad performance issue in the client and in the server).

A subscription is only late if notifications could not be sent.
If the server can send notifications faster than they are produced, the state will be UA_SUBSCRIPTIONSTATE_NORMAL eventually.

Otherwise, the behaviour is totally fine. Making the queue of the MonitoredItems smaller might help.

StalderT:
> Did you see the error or is this theoretical?

I see the error in real : 

In the same second, monitored item is sended multiple times : 

image

image

With a publishInterval of 10s : 
image

StalderT:
Wireshark log : [bug1.zip](https://github.com/open62541/open62541/files/1777946/bug1.zip)

jpfr:
Your client is lying to you.

You screenshot says MaxNotificationsPerPublish = 0 (infinite).
But Wireshark tells us that the client requested max 100 notifications per publish.

Our server is doing exactly the right thing.

maxnotifications

StalderT:
How can I convince you? Can we schedule an IRC chat?

jpfr:
> How can I convince you?

Show an example where we break the spec.

Did the server do something wrong in the Wireshark log?
Imho we **had to send so many responses**. Since there we many more than 100 notifications in the publish interval. But only 100 can be sent in each publish response.

> Can we schedule an IRC chat?

Sure. IRC works. In our channel some time tomorrow?
Skype is possible as well.

jpfr:
15:00 would work.
19:00 otherwise.

StalderT:
It works perfectly !!!! 👍 

Thanks for all.

jpfr:
Thank you for shepherding.

It was good to unpack the problem.
Seeing that it consisted of two (more or less) separate subproblems helped.

Let's merge this after some history cleaning.
And after the first 0.3 release candidate.

src/server/ua_subscription_datachange.c
```diff
@@ -35,27 +35,31 @@ MonitoredItem_delete(UA_Server *server, UA_MonitoredItem *monitoredItem) {
...
-            TAILQ_REMOVE(&monitoredItem->queue, val, listEntry);
-            UA_DataValue_deleteMembers(&val->data.value);
-            UA_free(val);
+        /* Clear the queued notifications */
+        UA_Notification *notification, *notification_tmp;
+        TAILQ_FOREACH_SAFE(notification, &monitoredItem->queue, listEntry, notification_tmp) {
+            /* Remove the item from the queues */
+            TAILQ_REMOVE(&monitoredItem->queue, notification, listEntry);
+            TAILQ_REMOVE(&sub->notificationQueue, notification, globalEntry);
+            --sub->notificationQueueSize;
```
StalderT: ```notificationQueueSize``` may be reduced, and in some cases, ```readyNotifications``` may be greater than ```notificationQueueSize```. I added a test in https://github.com/open62541/open62541/pull/1632/commits/d7202d14b8b350875435a37bb5c3c8afed452c14

StalderT:
Just to understand/know : Will be merged in master, 0.3 or master+0.3 ? Same for https://github.com/open62541/open62541/pull/1598. Thanks in advance.

jpfr:
Both. This PR is internal.

#1598 only adds API. So I am for merging. After good testing. Since we are already in the freeze period.

We had to be fast with the rc to make the window for Fedora.

StalderT:
@jpfr thanks for the clarifications

jpfr:
@StalderT 

There are a few fixups and iterations in the commit log.
Can you rebase and clean up the history to something nice and linear?
We can then merge to master and 0.3.

StalderT:
>There are a few fixups and iterations in the commit log.
Can you rebase and clean up the history to something nice and linear?
We can then merge to master and 0.3.

Done. Difficult to do better because of the chronology of commits. Otherwise I can do a single commit.

StalderT:
@jpfr All green. Do you want to clean up the history even more?

jpfr:
Merged. Thanks for seeing this through @StalderT !

StalderT:
thanks !

Can you align 0.3 ? Thanks in advance.

Issue #823, opened 2016-09-28
multiple .xml files with pyUANamespace compiler and Amalgamation


Pro:
I have a few issues/questions about the pyUANamespaceCompiler especially to @Stasik0 and @ichrispa, but of course also others ;)

For your convenience I created a simple repo for this issue which contains all the described files and which you can use to test:
https://github.com/Pro/open62541_namespaces

Assume the following setup:
* I have a project which uses open62541 as a dependency with static linking
* `UA_ENABLE_GENERATE_NAMESPACE0=ON`
* In my project I want to use [Opc.Ua.Di.NodeSet2.xml](https://opcfoundation.org/UADevices/1.1/Opc.Ua.Di.NodeSet2.xml) as an additional base namespace
* The `Opc.Ua.Di.NodeSet2` namespace is extended with a custom ObjectType `PlcDeviceType` using UaModeler which generates `project1.NodeSet2.xml` by Company1
* The `Opc.Ua.Di.NodeSet2` is extended again with a custom type `MachineDeviceType` by Company2  which generates `project2.NodeSet2.xml`

So finally I have 4 .xml files:
- `Opc.Ua.NodeSet2.xml`
- `Opc.Ua.Di.NodeSet2.xml`
- `project1.NodeSet2.xml` depending on `Opc.Ua.Di.NodeSet2.xml` (see below for content)
- `project2.NodeSet2.xml` depending on `Opc.Ua.Di.NodeSet2.xml` (see below for content)

Now I want to compile everything together, i.e., I need to create .c/.h files for the .xml models. This can be done with the `generate_open62541CCode.py` script. In the tutorial it is mentioned that one should add additional .xml files to the corresponding line in the open62541 CMakeLists.txt file. I don't want to change the base files of open62541, thus I create a custom call to `generate_open62541CCode.py` like it is done for the examples.

Here is now my question:
How does the call to `generate_open62541CCode.py` have to look like, to generate the model which includes `Opc.Ua.Di.NodeSet2.xml`,  `project1.NodeSet2.xml`, `project2.NodeSet2.xml`

I tried it with the following, but (after fixing #822) it fails during gcc make. This error only occurs if **amalgamation is enabled**. If it is disabled, it works fine.

CMakeLists.txt:
```
# generated namespace for custom .xml models
add_custom_command(OUTPUT ${PROJECT_BINARY_DIR}/src_generated/test_generated.c
        ${PROJECT_BINARY_DIR}/src_generated/test_generated.h
[...]

PR #1637, opened 2018-03-06
Add missing ifdef for compiling namespace externally

cabralfortiss:
I added some namespaces and followed #823 until some point to generate my namespaces files. I wanted to compile these files in my own project in C++ wich uses the open62541 static library. I added then in this PR what's needed for C++ and the big change is that I want to include only the folder where open62541.h file is, so the only thing missing was the ua_types_encoding_binary.h which is not exported so it is only present in the open62541.c file. I added the declarations of it for the namespaces and included a ["hack"](https://github.com/open62541/open62541/compare/master...cabralfortiss:master#diff-36dccac8f1a25167289a84f42ba74460R199) to avoid compiling errors for namespace0 in amalgamation

### DIFF
--- a/tools/generate_datatypes.py
+++ b/tools/generate_datatypes.py
@@ -491,8 +491,14 @@ printh('''/* Generated from ''' + inname + ''' with script ''' + sys.argv[0] + '
 extern "C" {
 #endif
 
+#ifdef UA_NO_AMALGAMATION
 #include "ua_types.h"
-''' + ('#include "ua_types_generated.h"\n' if outname != "ua_types" else ''))
+''' + ('#include "ua_types_generated.h"\n' if outname != "ua_types" else '') + '''
+#else
+#include "open62541.h"
+#endif
+
+''')
 
 filtered_types = iter_types(types, False)
 filtered_opaque_types = iter_types(types, True)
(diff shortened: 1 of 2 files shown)


Pro:
This is more or less the same as #1609. The UA_encodeBinary should not be visible in the public API

Pro:
Maybe we can use this PR as a quick fix and then work on the other issue later on.

jpfr:
@Pro 
Perfect. I merge to master. So the "hack" won't be in 0.3.

@cabralfortiss 
You can always compile the library in C without the amalgamation enabled to a .dll or .lib (.so or .a).
And then link your C++ project against that.

cabralfortiss:
Good that's merged. 

What's the problem with UA_encodeBinary being public?

@jpfr I normally do that. Now I wanted to have the open62541 untouched (without new namespaces) so I took my namespaces' files and compiled with my C++ project. 

Also, another question regarding this:

Python generates many files for a namespace. But only ua_namespace_MYNAME.c/h and ua_types_MYNAME_generated.c/h are used. Two other files ua_types_MYNAME_generated_encoding_binary.h and ua_types_MYNAME_generated_handling.h are generated but not used. 

Why is this? Would it be possible to have all the files in one like an amalgamated? Or is my approach wrong? Or should the namespaces be exported also as libraries?

jpfr:
> What's the problem with UA_encodeBinary being public?

The definition of the encoding method has evolved over time.
Internal APIs are easier to change.
Now that this is pretty stable we can make a version (without all those callbacks) of the encoding methods public.

The generated files are split since we generate documentation from the documented files.
So the datatype definitions were kept separate.

Improvements in this regards are welcome.
I just added an issue that is also related to the datatype generation: #1638
Any help is welcome.

PR #1640, opened 2018-03-07
Pragma message should be a string with quotation marks

ctron:


jpfr:
Thanks. The patch was applied to 0.3 as well.

PR #1641, opened 2018-03-08
fix ua_client_subscriptions logger

StalderT:


StalderT:
must also be merged in 0.3

Issue #1639, opened 2018-03-07
Reason for forced static CRT


the-om:
Description
===========
Currently, when building open62541 as a static library with MSVC, the compiler flag /MD (link with dynamic CRT) gets replaced with /MT (static CRT) in CMakeLists.txt. This causes issues during linking if conflicting settings are used in other modules. I just would like to know if there is a good reason for forcing this setting and overriding the user's choice.

Background Information / Reproduction Steps
===========================================
Build mbedtls with dynamic crt
Build open62541 with msvc with encryption enabled

Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): 0.3
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Windows
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

Stasik0:
https://github.com/open62541/open62541/commit/b86938501a48e5c6c003b66df9d11e9d5d261582

Stasik0:
Maybe you could propose some override logic in the cmake file?

the-om:
Thank you for the link. I had seen that commit but somehow missed the comment.
What would be the best way to do this? Introduce a switch UA_FORCE_STATIC_CRT that's on by default?

Stasik0:
Yes, like UA_MSVS_FORCE_STATIC_CRT. Advanced and on by default. And then wrap the foreach loop in an if? @jpfr?

PR #1642, opened 2018-03-08
Add cmake option UA_MSVC_FORCE_STATIC_CRT

the-om:
#1639 
Provide users an option to disable the automatic override of /MD compiler switch with /MT.
Only relevant for MSVC.


jpfr:
Thanks! Merged.

Issue #1643, opened 2018-03-09
Can not build with encryption and amalgamation


the-om:
Description
===========
#1634 breaks my build.
The reason are 3 violations of the one definition rule when using amalgamation.
The functions sha1, asym_getRemotePlainTextBlockSize and asym_getRemoteEncryptionKeyLength are declared static which works for non-amalgamated builds, but when amalgamated they are defined twice in the same translation unit (open62541.c) which is not legal.

Background Information / Reproduction Steps
===========================================
Build with UA_ENABLE_AMALGAMATION and UA_ENABLE_ENCRYPTION

Compiler: Microsoft Visual Studio 2017

ccvca:
Possible solutions might be:
- Use an "include Guard" around the static function definitions,
- Use normal header/source and use only one implementation

jpfr:
- pull out the common code in common .c file that is used for both SecurityPolicy implementations
- rename the duplicate methods to something more specific

PR #1645, opened 2018-03-09
fix security policy amalgamation odr violations

the-om:
Adds two additional files in the plugins directory: ua_securitypolicy_common.h/.c
Pulls out sha1 function into these common files and renames other conflicting functions.
These new files could also be used in the future to reduce code duplication between security policy basic128rsa15 and basic256sha256.
#1643


jpfr:
Thanks :-)

PR #1649, opened 2018-03-12
Feature/remove const qualifier from data variable in edit node

aDogCalledSpot:
The function `UA_Server_editNode` takes a `const void *data`, which does not make a lot of sense because in some cases said data needs to be changed. For example: an element needs to be added to a doubly-linked list which belongs to a node. No matter where you insert it, the reference to the previous and/or next item will have to be edited.
To achieve this while not changing too much all attributes which lose their const qualifier by simply changing the `editNode` function have been edited so that the const is cast away when calling the function but the callback from `editNode` will cast the data back to const.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,9 @@ mark_as_advanced(UA_ENABLE_DETERMINISTIC_RNG)
 option(UA_ENABLE_VALGRIND_INTERACTIVE "Enable dumping valgrind every iteration. CAUTION! SLOWDOWN!" OFF)
 mark_as_advanced(UA_ENABLE_VALGRIND_INTERACTIVE)
 
+option(UA_MSVC_FORCE_STATIC_CRT "Force linking with the static C-runtime library when compiling to static library with MSVC" ON)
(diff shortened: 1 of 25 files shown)


jpfr:
The effort to make the library more consistent is of course appreciated.

This PR only changes the callback API. But none of the users of it.
Which of the services can be improved or simplified by changing to non-const input arguments?

aDogCalledSpot:
I stumbled upon this problem while implementing the events. Both events and monitored items are saved in lists belonging to object nodes. Creating events or monitored items leads to the problems I described above where the data (a list entry) needs to have its references updated when being inserted.
After consulting @Pro we came to the conclusion that there will likely be more situations where the data should be mutable and that removing the const qualifier makes sense.

jpfr:
Understood.
Can you rebase so that merging is possible?

aDogCalledSpot:
Having some problems with Git, will close this PR and open a newer cleaner one. Sorry for the inconvenience.

jpfr:
Most of the time, `rebase -i master`, followed by a `push -f` is enough to clean up the history.

PR #1652, opened 2018-03-13
removed const qualifier from editNode function

aDogCalledSpot:
Cleaned up branch from earlier pull request.

Comment from older PR:
The function UA_Server_editNode takes a const void *data, which does not make a lot of sense because in some cases said data needs to be changed. For example: an element needs to be added to a doubly-linked list which belongs to a node. No matter where you insert it, the reference to the previous and/or next item will have to be edited.
To achieve this while not changing too much all attributes which lose their const qualifier by simply changing the editNode function have been edited so that the const is cast away when calling the function but the callback from editNode will cast the data back to const.


jpfr:
👍

aDogCalledSpot:
I have an update on this pull request here https://github.com/open62541/open62541/pull/1669

PR #1656, opened 2018-03-15
Fix/deconstruct nodes

jpfr:


jpfr:
outdated / fixed on master

PR #1657, opened 2018-03-15
Client performance issue on large responses

anzeskerjanc:
## Problem:
Large responses on server are divided into 64k chunks.  
Client for every received chunk increases (UA_realloc) complete message size for one chunk size(64k). On large responses this means a lot of memory allocating for increasing buffer size just for 64k.

## Solution:
Increase memory size for more than just newly arrived chunk.


jpfr:
Hello @anzeskerjanc,

were the performance issues solved?

jpfr:
I see you opened #1668.

Sometimes it is easier to rebase and force-push to the old PR.

PR #1663, opened 2018-03-16
Enable auto generation for types required by EventFilter

basyskom-jvoe:
There is a working API for event monitoring with event filters, so the operator types for the where clause should be available by default.


basyskom-jvoe:
Will this commit also be merged into the 0.3 branch?

Pro:
yes

Issue #1584, opened 2018-02-06
[Server][Feature Request] Callback to application when client tries to read non-existing node


JoJoBond:
Description
===========
I would like to ask for a feature where the server triggers a application callback whenever a client tries to read/subscribe to a non-existing node. The application can react to that request, e.g. create the node ad hoc, alter/map the nodeid to a different one, etc.

Background Information / Reproduction Steps
===========================================
Often, especially when you are working with PLCs, you don't know what data is available on your source side. So a client must 'blindly' read a non-browseable value and the server will try to provide that value based on predefined node syntax. When using OPC DA servers for accessing Siemens S7 PLC data you are often able to directly address the data that you need, e.g.
"DB10.Real16" reads a 4 byte real on byte position 16 in data block 10.
The server will read that data from PLC and relay it to the client. After reading the node once, it becomes browseable and can be used as usual afterwards.

I don't have working cmake environment but did a very basic implementation on the single file v0.2 sources:

open62541.c: https://paste2.org/nLf7J3m7
open62541.h: https://paste2.org/AemjW7ax
patch file: https://paste2.org/L1yk7BKb

Checklist
=========
Please provide the following information:

 - [ ] open62541 Version (release number or git tag): Does not apply
 - [ ] Other OPC UA SDKs used (client or server): Does not apply
 - [ ] Operating system: Does not apply
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached: Does not apply
 - [ ] Wireshark network dump attached: Does not apply
 - [ ] Self-contained code example attached: Does not apply
 - [ ] Critical issue: No

StalderT:
Good idea. I will propose a PR.

jpfr:
There are two possibilities:

- Add the callback to the server that is used in the read service
- Add the callback to the nodestore plugin (where nodestore_get fails to find the node)

I tend to minimize the API "surface" of the main library.
The nodestore is a "plugin" and free for modification.
What are the optinions from @open62541/open62541-core ?

By the way, **if you modify the nodestore plugin, you can return the created node already for the first read.** So you don't have to "read twice" to get data out.

StalderT:
If you want : https://github.com/StalderT/open62541/commit/b3c015ce7295c6e29ade6b0de182dd3592a72f8b

> By the way, if you modify the nodestore plugin, you can create the missing node already for the first read. So you don't have to "read twice" to get data out.

With this commit, the node is created at the first read

[...]

PR #1665, opened 2018-03-17
Added callback to default nodestore for when a requested node is not …

JoJoBond:
Addresses issue #1584 .
Added callback to default nodestore for when a requested node is not found. This way the application can dynamically create a node on request.

### DIFF
--- a/plugins/ua_nodestore_default.c
+++ b/plugins/ua_nodestore_default.c
@@ -42,6 +42,8 @@ typedef struct {
     UA_UInt32 size;
     UA_UInt32 count;
     UA_UInt32 sizePrimeIndex;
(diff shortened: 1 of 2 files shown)


StalderT:
👍 

@jpfr can be merged in 0.3?

jpfr:
The plugins are meant to be forked and adjusted.
This is why they have the CC0 license.

The use case can be covered without changing the plugin API.
Just add the node transparently in the background when it is not found inside the nodestore plugin.

You can even go further. With the _get and _release mechanism, some nodes can be created and deleted on the fly without every putting them into the hashmap.

Pro:
@jpfr should we close this as wontfix?

Pro:
Closing as wontfix

PR #1669, opened 2018-03-19
UPDATE: remove const qualifier from edit node

aDogCalledSpot:
I missed the instance of the const qualifier in the declaration of `UA_EditNodeCallback`.

Update from: https://github.com/open62541/open62541/pull/1652


Pro:
replaced by #1677

PR #1673, opened 2018-03-19
First Batch for PubSub Support

jpfr:


CMakeLists.txt
```diff
@@ -440,14 +443,15 @@ set(default_plugin_sources ${PROJECT_SOURCE_DIR}/plugins/ua_network_tcp.c
...
 if(UA_ENABLE_ENCRYPTION)
-    set(default_plugin_headers ${default_plugin_headers}
-            ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic128rsa15.h)
-    set(default_plugin_headers ${default_plugin_headers}
-            ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic256sha256.h)
-    set(default_plugin_sources ${default_plugin_sources}
-            ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic128rsa15.c)
-    set(default_plugin_sources ${default_plugin_sources}
-            ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic256sha256.c)
+    list(APPEND default_plugin_headers ${PROJECT_SOURCE_DIR}/plugins/ua_securitypolicy_basic128rsa15.h
```
fmeerkoetter: I might be wrong but this looks unrelated to the pub/sub topic per se

jpfr: Um yeah. The change is functionally equivalent to the previous version.
It is an aligntment with the style of the pubsub configuration just below.

If you take offense in this, we can pull this out into a separate commit for the same PR.

fmeerkoetter: No offense taken :-)

PR #1688, opened 2018-03-26
Second Batch for PubSub Support

andreasebner:
1. PubSub connection interface and handling.
2. UDP-Plugin for PubSub.

This introduces the following API changes:

NEW **UA_Server_addPubSubConnection**: create a new PubSub connection. The transport profile URL specifies the requestet connection type. The server config contains PubSub transport layers as the factory for new pubsub connections at runtime. **UA_Server_removePubSubConnection**: remove a connection from the current PubSub config. **UA_PubSubConnection_getConfig**: get the config of a created connection.

More details about connection handling can be found in tutorials_pubsub_connection


Pro:
Nice!
IMHO we can now remove the `ua_network_udp.c/h` since these files do not compile anyways and are not used at all. @jpfr should we include the deletion in this PR?

examples/pubsub/tutorial_pubsub_connection.c
```diff
@@ -0,0 +1,87 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Pro: Shouldn't this be CC0 (in general we have all the examples with CC0 license)?

src/pubsub/ua_pubsub_manager.c
```diff
@@ -0,0 +1,170 @@
...
+            newConnection->config = tmpConnectionConfig;
+            newConnection->channel = server->config.pubsubTransportLayers[i].createPubSubChannel(newConnection->config);
+            if(!newConnection->channel){
+                UA_PubSubConnection_delete(newConnection);
+                if(server->pubSubManager.connectionsSize > 0){
+                    newConnectionsField = (UA_PubSubConnection *)
+                            UA_realloc(server->pubSubManager.connections,
+                                       sizeof(UA_PubSubConnection) * (server->pubSubManager.connectionsSize));
+                    if(!newConnectionsField) {
+                        return UA_STATUSCODE_BADINTERNALERROR;
```
Infinity95: The memory allocated for tmpConnectionConfig  will be leaked here I believe?

include/ua_server_pubsub.h
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Take a look on the PubSub Tutorials for mor details about the API usage.
+ *
+ * Connections
+ * -----------
+ *
+ *
+ */
+
+/**********************************************/
```
jpfr: Very minor last comments.
Can we use rst section headers *or* fat-comment headers?
But not both.

include/ua_server_pubsub.h
```diff
@@ -0,0 +1,85 @@
...
+    size_t connectionPropertiesSize;
+    UA_KeyValuePair *connectionProperties;
+    UA_Variant connectionTransportSettings;
+} UA_PubSubConnectionConfig;
+
+UA_StatusCode
+UA_PubSubConnection_getConfig(UA_Server *server, UA_NodeId connectionIdentifier,
+                              UA_PubSubConnectionConfig *config);
+
+/***************************************************************************/
```
jpfr: That's a big fat comment for something that is rather self-explanatory.
I understand that the structure is helpful in long source files.
But it is a bit much here.

plugins/ua_network_pubsub_udp.c
```diff
@@ -0,0 +1,462 @@
...
+
+    UA_NetworkAddressUrlDataType address;
+    if(UA_Variant_hasScalarType(&connectionConfig->address, &UA_TYPES[UA_TYPES_NETWORKADDRESSURLDATATYPE])){
+        address = *(UA_NetworkAddressUrlDataType *)connectionConfig->address.data;
+    } else {
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "PubSub Connection creation failed. Invalid Address.");
+        return NULL;
+    }
+    //allocate and init memory for the UDP multicast specific internal data
+    UA_PubSubChannelDataUDPMC * channelDataUDPMC = (UA_PubSubChannelDataUDPMC *) UA_calloc(1, (sizeof(UA_PubSubChannelDataUDPMC)));
```
jpfr: Can we make an effort and break very long lines?
I don't want to impose a strict length limit.

But many people like to split their window horizontally.

plugins/ua_network_pubsub_udp.c
```diff
@@ -0,0 +1,462 @@
...
+    newChannel->state = UA_PUBSUB_CHANNEL_PUB;
+    return newChannel;
+}
+
+/**
+ * Subscribe to a given address.
+ *
+ * @return UA_STATUSCODE_GOOD on success
+ */
+static UA_StatusCode UA_PubSubChannelUDPMC_regist(UA_PubSubChannel *channel, UA_ExtensionObject *transportSettings){
```
jpfr: All other source files have the following schema:

```c
static UA_StatusCode
UA_PubSubChannelUDPMC_regist(UA_PubSubChannel *channel, UA_ExtensionObject *transportSettings) {
```

Just for consistency.

plugins/ua_network_pubsub_udp.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef _WIN32
+#include <sys/socket.h>
+#endif
+#include "ua_plugin_pubsub.h"
+
+//UDP multicast network layer specific internal data
+typedef struct {
+    int ai_family;                        //Protocol family for socket.  IPv4/IPv6
+    struct sockaddr_storage *ai_addr;     //https://msdn.microsoft.com/de-de/library/windows/desktop/ms740496(v=vs.85).aspx
```
jpfr: The plugins are exchangeable so that we allow different implementations.
Here we require a specific networking API in the plugin header already.
Can we move `UA_PubSubChannelDataUDPMC` inside the plugin so that the interface is agnostic?

include/ua_plugin_pubsub.h
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * The PubSub Connection API is the interface between concrete network implementations and the internal pubsub
+ * code.
+ *
+ * The PubSub specification enables the creation of new connections on runtime.
+ * Wording: 'Connection' -> OPC UA standard 'highlevel' perspective,
+ * Channel -> open62541 implementation 'lowlevel' perspective
+ * A channel can be assigned with different network implementations like UDP, MQTT, AMQP.
+ * The channel provides basis services like send, regist, unregist, receive, close.
+ *
```
jpfr: blank line

include/ua_server_pubsub.h
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * 2. Call the add[element] function and pass in the configuration.
+ * 
+ * 3. The add[element] function returns the unique nodeId of the internally created element.
+ *
+ * Take a look on the PubSub Tutorials for mor details about the API usage.
+ *
+ * Connections
+ * -----------
+ *
```
jpfr: blank lines

plugins/ua_network_pubsub_udp.c
```diff
@@ -0,0 +1,462 @@
...
+        UA_free(newChannel);
+        return NULL;
+    }
+
+    //check if the ip address is a multicast address
+    if(requestResult->ai_family == PF_INET){
+        struct in_addr imr_interface;
+        inet_pton(AF_INET, addressAsChar, &imr_interface);
+        if((ntohl(imr_interface.s_addr) & 0xF0000000) != 0xE0000000){
+            UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
```
jpfr: You leak the addrinfo in hints.
freeaddrinfo needs to be called in every return path after line 164.

plugins/ua_network_pubsub_udp.c
```diff
@@ -0,0 +1,462 @@
...
+            UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "PubSub Connection sending failed.");
+            return UA_STATUSCODE_BADINTERNALERROR;
+        }
+        nWritten += n;
+    }
+    return UA_STATUSCODE_GOOD;
+}
+
+/**
+ *
```
jpfr: Blank line

plugins/ua_network_pubsub_udp.h
jpfr: Let me rephrase the comment:
This is **inside** the plugin. So we can actually use system-specific API.
But since the structure is used only inside ` plugins/ua_network_pubsub_udp.c`, let's move the definition there.

Just to separate concerns and keep the API surface small.

src/pubsub/ua_pubsub_manager.c
andreasebner: The configuration copy is asigned in line 50 `newConnection->config = tmpConnectionConfig; ` and (in case of error) freed correctly in line 56 `UA_PubSubConnection_delete(newConnection); `.

jpfr:
Nice! Can you squash the PR to a single commit and rebase on master?
I think you don't yet have the new sonarcube travis config.
So I don't see this PR on travis at all.

Pro:
Note that sonarcube still fails on PRs. This will (hopefully) be fixed in c7655458b4d96a843f5eb194044e200960a14ec4 which lands with #1691

jpfr:
@Pro Can we disable sonarcube until then?

jpfr:
@Pro I would remove the old udp layer in a later commit.
Let's get this in first. Otherwise we have diverging ongoing internal dev.

Pro:
> @Pro Can we disable sonarcube until then?
It looks like it's fine now, so we just have to wait for CI, and then merge. So approx 1 hour and then it's fixed.

Pro:
You can now rebase onto master and the sonarcloud issue should be resolved

jpfr:
Congrats @andreasebner !

PR #1695, opened 2018-03-29
Variant binary encoding size optimization

anzeskerjanc:
If variant contains array of overlayable elements, then we don't need to iterate over all elements. The same optimisation is used in array encoding (Array_calcSizeBinary).


jpfr:
Great! Thanks for the contribution.

Issue #1698, opened 2018-03-29
Compiling error on 0.3 and large binary


eliasnaslund:
Description
===========
I'm compiling tutorial_server_datasource.c exactly as it is with the single source distribution and get error when compiling with all of the features disabled.

If I manually alter the code to make it compile the binary is over 100K larger then the same example with v0.2. Is it possible to decrease the overhead? both constant data and code size have increased a lot even though no options is enabled.

Background Information / Reproduction Steps
===========================================
```
/**
 * Feature Options
 * ---------------
 * Changing the feature options has no effect on a pre-compiled library. */
#define UA_LOGLEVEL 800
//#define UA_ENABLE_METHODCALLS
//#define UA_ENABLE_NODEMANAGEMENT
//#define UA_ENABLE_SUBSCRIPTIONS
/* #undef UA_ENABLE_MULTITHREADING */
/* #undef UA_ENABLE_ENCRYPTION */

/* Advanced Options */
//#define UA_ENABLE_STATUSCODE_DESCRIPTIONS
//#define UA_ENABLE_TYPENAMES
/* #undef UA_ENABLE_DETERMINISTIC_RNG */
/* #undef UA_ENABLE_GENERATE_NAMESPACE0 */
/* #undef UA_ENABLE_NONSTANDARD_UDP */
//#define UA_ENABLE_DISCOVERY
/* #undef UA_ENABLE_DISCOVERY_MULTICAST */
//#define UA_ENABLE_DISCOVERY_SEMAPHORE
/* #undef UA_ENABLE_UNIT_TEST_FAILURE_HOOKS */
/* #undef UA_ENABLE_VALGRIND_INTERACTIVE */
//#define UA_VALGRIND_INTERACTIVE_INTERVAL 1000

/* Options for Debugging */
/* #undef UA_DEBUG */
/* #undef UA_DEBUG_DUMP_PKGS */
```

I have tried both v0.3rc1 and 0.3-rc1-62-ge2dde65. With both versions I get:
```
gcc -std=c99 -Os tutorial_server_datasource.c open62541.c -o tutorial_server_datasource -s
open62541.c:41065:5: warning: initialization makes integer from pointer without a cast [-Wint-conversion]
     NULL,  /*.clientContext */
     ^
open62541.c:41065:5: note: (near initialization for ‘UA_ClientConfig_default.outStandingPublishRequests’)
open62541.c:41067:5: warning: excess elements in struct initializer
     10 /* .outStandingPublishRequests */
[...]

PR #1699, opened 2018-03-29
partial fix for #1698

StalderT:


Pro:
@StalderT can you rebase your commits onto current 0.3 and then change this PR to merge into 0.3

For fixes which should go into 0.3, we decided to use that approach. From time to time we then simply merge 0.3 into master and thus have all the fixes from 0.3 also in master.

PR #1703, opened 2018-03-31
encryption fix key length size fix #1702

### LINKED ISSUE #1702: client_secure_new problem
0) https://github.com/open62541/open62541/commit/da64add5c0d5eb0fda2ea7f4f5e04cc20e237ab9
1) We use an unmodified examples/encryption/client_basic256sha256.c
2) We utilize a DER certificate and a DER key that correctly work in an freeopcua client
3) We use no trusted list

The problem occurs in file src/client/ua_client.c method UA_Client_secure_init:

retval = client->securityPolicy.channelModule.newContext(securityPolicy, remoteCertificate,
                                                             &client->channel.channelContext);

retval has the value of 0x80180000 (UA_STATUSCODE_BADCERTIFICATEUSENOTALLOWED).

Any hints what the problem could be?

StalderT:
fix #1702 

according to https://msdn.microsoft.com/en-us/library/windows/desktop/dd323379(v=vs.85).aspx

For basic128rsa15: The minimum asymmetric key length is 1024; the maximum asymmetric key length is 4096. 

For basic256sha256: The minimum asymmetric key length is 1024; the maximum asymmetric key length is 4096. 

mbedtls_pk_rsa return a length in bytes : https://tls.mbed.org/api/structmbedtls__rsa__context.html#a06a537fe6045d6c08d92b2a91461b5d8

@Infinity95 can you check and confirm ?


Infinity95:
For Basic128Rsa15 the minimum key length is 1024 and the maximum key length is 2048.
For Basic256Sha256 the minimum key length is 2048 and the maximum key length is 4096.
(see http://opcfoundation.org/UA/SecurityPolicy/#Basic128Rsa15 and http://opcfoundation.org/UA/SecurityPolicy/#Basic256Sha256)

StalderT:
Ok. Thanks for the clarification.

PR #1705, opened 2018-03-31
Add common.h to cmake project

Infinity95:
This adds the common.h file to the cmake project which is necessary for some IDEs to recognize it for analysis.


examples/common.h
```diff
@@ -1,3 +1,4 @@
+#include "open62541.h"
```
jpfr: Includes are normally below the copyright notice.

jpfr:
Thanks!

PR #1718, opened 2018-04-04
Nodeset compiler: Support aliases for all uses of NodeIds; Improve support for utf8 strings

jpfr:


Pro:
Related/Replaces #1609

Probably you also want to include the commit 743e56fb8a9bdfe3d5fa4159c7ea48e696bb7654

jpfr:
@Pro Done. Thanks for pointing this out.
This PR is now a mixed bag of improvements. I had to deal with a difficult xml file where also Aliases were used for the data types.

Shall we keep #1609 to track the handling of custom data types in the nodeset compiler for 0.4?

Pro:
Can you also cherry pick 218b1c01c6d15d4d7da0b613f4cf29e61a75cbc8 to set the default log level to error

Interestingly you now do not run into the problem I had when replacing the datatype alias (the commit I mentioned in previous message). The current master version skips some variable nodes, because the datatype is using an alias, which is not resolved during printing the code and thus the whole variable is skipped. That's why I enabled the loglevel to error so that you can at least see that something bad is happening (otherwise it does not output anything at all).
Then when I replaced the datatype alias, the missing vairables were printed. These also include some custom data types, which require `encodeBinary`. So I'm a bit confused why in this PR this is not required. Looks like there are still some variables not printed, so #1609 is still open.

PR #1719, opened 2018-04-04
Fix null read and duplicate node IDs

Pro:
These issues were found by oss-fuzz.

@jpfr please check my second commit in the nodestore. That bug was really hard to find.
Summarized the following was the case:
1. Some nodes were created and stored in the hash map `ns->entries` (for simplicity let's assume it's an array with node IDs, and the `idx += hash2` is just an iterator `i++`): `[27, 100, 50074]`
2. The node `100` is deleted, so the hashmap will then be `[27, thombstone, 50074]`
3. `findFreeSlot(ns, 50074)` iterates over the array, finds the thombstone on second position and returns that slot since it's the first free slot for that node id. 
4. `UA_NodeMap_insertNode` then assumes that the "random" node id is ok, since there is a free slot.

The problem here is that `findFreeSlot`stops at the first thombstone, but the corresponding node ID could still be somewhere in the nodeset. I added an additional `findOccupiedSlot` which checks if the node ID is really not used.

### DIFF
--- a/plugins/ua_nodestore_default.c
+++ b/plugins/ua_nodestore_default.c
@@ -342,15 +342,24 @@ UA_NodeMap_insertNode(void *context, UA_Node *node,
         while(true) {
             node->nodeId.identifier.numeric = identifier;
             slot = findFreeSlot(ns, &node->nodeId);
-            if(slot)
+            /* findFreeSlot will return a free slot, but will not check, if the node ID already
+             * exists in the nodestore. There may be a free slot before the already existing node
+             * is found, e.g. if another node occupied the slot and is now freed.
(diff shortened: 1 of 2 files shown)


Pro:
@jpfr is this ok to be merged into 0.3 (and then master)?

jpfr:
You have included this comment.

>  findFreeSlot will return a free slot, but will not check, if the node ID already exists in the nodestore.

But findFreeSlot clearly does check for that. Only that sometimes the existing NodeId appears at a later slot. Can you update the comments to explain the issue as clearly as your issue report?

Pro:
on hold until #1727 is clarified

Pro:
Closing in favour of #1727

PR #1726, opened 2018-04-06
Add filter support for numeric absolute filters

peter-rustler-basyskom:
The following changes are made for server support:
 * Creating an DEADBANDTYPE_ABSOLUTE filter for non numeric data type from client fail with BADFILTERNOTALLOWED
 * Creating a DEADBANDTYPE_PERCENT filter will fail with BADNOTIMPLEMENTED (impementing it is not possible as long we do not have the EURange)
 * Updates to clients will be filtered for DEADBANDTYPE_ABSOLUTE filters for numeric scalar and array types

If this is also code which should go into 0.3 release, I can also do a pull request for 0.3 branch. In my opinion the first 2 bullet points fixes bugs, because we do always report GOOD when applying filters by clients, but do not handle them right.

### DIFF
--- a/src/server/ua_subscription.h
+++ b/src/server/ua_subscription.h
@@ -81,7 +81,8 @@ struct UA_MonitoredItem {
     UA_UInt32 maxQueueSize;
     UA_Boolean discardOldest;
     // TODO: dataEncoding is hardcoded to UA binary
-    UA_DataChangeTrigger trigger;
+    UA_DataChangeFilter filter;
+    UA_Variant lastValue;
 
     /* Sample Callback */
     UA_UInt64 sampleCallbackId;
--- a/src/server/ua_subscription_datachange.c
+++ b/src/server/ua_subscription_datachange.c
@@ -64,6 +64,7 @@ MonitoredItem_delete(UA_Server *server, UA_MonitoredItem *monitoredItem) {
     LIST_REMOVE(monitoredItem, listEntry);
     UA_String_deleteMembers(&monitoredItem->indexRange);
     UA_ByteString_deleteMembers(&monitoredItem->lastSampledValue);
+    UA_Variant_deleteMembers(&monitoredItem->lastValue);
     UA_NodeId_deleteMembers(&monitoredItem->monitoredNodeId);
     UA_Server_delayedFree(server, monitoredItem);
 }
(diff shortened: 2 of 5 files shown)


src/ua_types.c
```diff
@@ -1114,3 +1114,12 @@ UA_Array_delete(void *p, size_t size, const UA_DataType *type) {
     }
     UA_free((void*)((uintptr_t)p & ~(uintptr_t)UA_EMPTY_ARRAY_SENTINEL));
 }
+
+UA_Boolean
+isDataTypeNumeric(const UA_DataType *type) {
```
Pro: There could also be custom subtypes which derive from the base datatypes. These should also be handled.

src/server/ua_subscription_datachange.c
```diff
@@ -10,6 +10,7 @@
 #include "ua_subscription.h"
 #include "ua_server_internal.h"
 #include "ua_types_encoding_binary.h"
+#include <math.h>
```
Pro: Do we have math library on every architecture, or is it dependent on the architecture? Not sure about that. Anyways global includes should be before local includes. (I.e. this should be the first include)

src/server/ua_subscription_datachange.c
```diff
@@ -146,10 +148,108 @@ void MonitoredItem_ensureQueueSpace(UA_MonitoredItem *mon) {
...
+    }
+    return false;
+}
+
 /* Errors are returned as no change detected */
 static UA_Boolean
 detectValueChangeWithFilter(UA_MonitoredItem *mon, UA_DataValue *value,
                             UA_ByteString *encoding) {
+    if (isDataTypeNumeric(value->value.type)
+            && (mon->filter.trigger == UA_DATACHANGETRIGGER_STATUSVALUE
```
Pro: `mon->filter` can also be NULL? So there should be a NULL check?

src/server/ua_subscription.h
```diff
@@ -81,7 +81,8 @@ struct UA_MonitoredItem {
     UA_UInt32 maxQueueSize;
     UA_Boolean discardOldest;
     // TODO: dataEncoding is hardcoded to UA binary
-    UA_DataChangeTrigger trigger;
+    UA_DataChangeFilter filter;
+    UA_Variant lastValue;
```
Pro: Can't we reuse `lastSampledValue`?

peter-rustler-basyskom: This is not possible. lastSampleValue is a Byterepresentation of the value and will be string compared. If it is not equal, it will be reported to the client. For me this is basiclly a hack which helps to compare values type independant. Here for filters you have to check which type is in the variant and handle it properly. This could be done for other not numeric types, too, but it isn't done yet. If someone implement it, it is possible to remove lastSampledValue and use the variant. Until then I would let it like it is in this PR.

src/server/ua_subscription_datachange.c
peter-rustler-basyskom: filter is a value based member and can't be null. It is no pointer in the monitored item.

src/ua_types.c
peter-rustler-basyskom: Hi, can you please shortly explain how i can detect a type not already existing?

in the filtering code i do:
if (UA_Variant->type == &UA_TYPES[UA_TYPES_SBYTE])
((const UA_SByte*)UA_Variant->data)[index] to get the right value for array type variant.
Then I do: (a>b?a-b:b-a)
to compare with dead band value.

See outOfDeadBand in ua_subscription_datachange.c:154

How can I do that for custom types ?

Then I do
if(type == &UA_TYPES[UA_TYPES_SBYTE])
to detect if a type is a numeric type.

See isDataTypeNumeric in ua_types.c:1119

How can I do that for custom types ?

jpfr:
Okay to merge. CI is all green.
Do you want to rebase-interactive to squash the the fix-commits?

I can do it. But don't have write-access to your repo.

src/ua_types.c
jpfr: @Pro Can we do that in an additional PR if the need comes up?
We then need an additional mechanism that ensures that the encoding is identical to that of the base datatype. Because we cast to double/float/int_x for the comparison.

src/server/ua_subscription.h
jpfr: Let's explore the performance impact.
If it is acceptable I would also prefer to have a "single source of truth" for the last value.
In a next PR if we decide to do this.

jpfr:
Thanks for the contribution @peter-rustler-basyskom !

Issue #1711, opened 2018-04-03
Random NodeId generation may overflow or result in endless loop


Pro:
Currently we are creating a "random" node ID for new nodes, if none is specified:
https://github.com/open62541/open62541/blob/8066c68aa254366dbed2b9cc5e49e89e7fc3ea9c/plugins/ua_nodestore_default.c#L340

I just saw that there may be two issues:

- Overflow of `identifier` is not handled. Even though identifier is 32bit, a long running server which creates many events may result in an overflow (every event gets its own node ID temporarily).

- The `findFreeSlot` loop may result in an endless loop if there are no more free slots available (i.e., no more available node IDs).

jpfr:
Solved in #1727.

PR #1727, opened 2018-04-06
nodestore: avoid potential endless loops and overflows

bbara:
Hey,

potential fix for #1711.
Instead of converting to UA_UInt64 we could also check against a boundary  (< UA_UINT32_MAX - hash2) before the addition happens so we could keep entirely on 32 bit arithmetic here.

Also influences #1719 - this findFreeSlot checks the complete used slots until it returns.
Should be a little more performant (worst case factor 2) than running again through the whole set.

Cheers


Pro:
I closed #1719 in favour of this PR since this includes a more complete bugfix.

When all green we can merge into 0.3, and then merge 0.3 back to master

Issue #1680, opened 2018-03-22
Fortify compiler flags


jpfr:
Now that we have encryption, we have to think about exposing our system to external attacks.
With fuzzing by @Pro we are already pretty good in that regard.
@Infinity95 is uncovering potential attack vectors as well.

In case of an eventual exploit, the right compiler flags would give us some more mitigation "for free".
Here's a list of recommended compiler flags. But they depend on the compiler...

https://developers.redhat.com/blog/2018/03/21/compiler-and-linker-flags-gcc/

Adding sensible fortification for gcc/Linux would be a nice first contribution for interested new developers.

jpfr:
Fortification was added in #1732.

Issue #1700, opened 2018-03-29
Write doc: how to create a minimal binary and check how to minimize binary file


Pro:
There's no documentation on how to create a minimum size binary.

We should add a document on that, especially for resource limited devices.
It cloud include notes about:
- Use / customize the minimal nodeset NS0
- Decrease loglevel
- Disable typenames
- Set compiler optimization options (e.g. `-Os`)

Additionally we should analyze the generated binary and see where most of the size is used up (e.g., static objects, nodeset compiler output, ...)

More ideas are welcome!

jpfr:
Even with all the optimizations, the binary size more than doubled since we added nodeset definitions that are generated from xml.

It was ~80kb. Now ~200kb.

Pro:
We should try to minimize our minimal nodeset XML and remove some (unnecessary?) nodes. When doing that, one should make sure that the file is NOT reordered, to be able to see the removed items in a diff, i.e. lines should only be deleted, but not moved.

jpfr:
#1697 is adding nodes to the minimal nodeset. ☹️ 
They are required to pass the CTT.

👍 for minimizing the minimal xml
We can also add a "nano" option that does not pass the CTT but makes most clients happy.

jpfr:
By the way, there are docs for minimizing the binary size.
https://open62541.org/doc/current/building.html#minimizing-the-binary-size

In place since a year or so...

jpfr:
See #1638 for a way to further reduce the binary size...

Stasik0:
@jpfr do you try nano or 2017 nano? As I told yesterday. It it a ctt bug?

The support of Diagnostic Objects and Variables is optional for this Profile despite it being defined as “mandatory” in UA Part 5. Support of Diagnostic Objects and Variables is mandatory in some higher level Profiles. Exposing types in the AddressSpace is optional for this Profile except if custom types (i.e. types that are derived from well-known ObjectTypes, VariableTypes, ReferenceType or DataTypes) are used. Exposing all supported types in the AddressSpace is mandatory in some higher level Profiles. This profile supersedes the "Nano Embedded Device Server Profile".
 URI	http://opcfoundation.org/UA-Profile/Server/NanoEmbeddedDevice2017

Stasik0:
Can we ommit more strings in the nodeset? Like not setting locale, or using the same locale string or ommitng Browsename/displayname for ns0? Will ctt eat that?

jpfr:
[...]

PR #1732, opened 2018-04-11
CMake: enable IPO and additional hardening

bbara:
Hey,

looked into #1680. Checked provided info ([1]), but also additional sources.

- Added functionality to check if flag is supported by compiler.
- Activated stack-protector-strong (should be default anyway) for the normal build. Not sure why Release was built without.
- Added additional run-time buffer overflow detection (which needs at least -O1).
- (Therefore) Activated -O2 as default Release optimization.
- Added Link-Time-Optimization (-flto)/Inter-Procedural-Optimization (IPO).
- Added two future compiler checks, which currently signalize their activity with a cmake warning.

- Added Release unit testing on Travis.

Any further ideas?

[1] https://developers.redhat.com/blog/2018/03/21/compiler-and-linker-flags-gcc/
[2] https://blog.regehr.org/archives/1307
[3] https://wiki.debian.org/Hardening
[4] https://blog.quarkslab.com/clang-hardening-cheat-sheet.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
-cmake_minimum_required(VERSION 3.0)
+ cmake_minimum_required(VERSION 3.0...3.12)
 project(open62541)
 # set(CMAKE_VERBOSE_MAKEFILE ON)
+if(${CMAKE_VERSION} VERSION_LESS 3.12)
+    cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
+endif()
 
 string(TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_LOWER_CASE)
 
@@ -273,6 +276,9 @@ endif()
 option(UA_DEBUG_DUMP_PKGS "Dump every package received by the server as hexdump format" OFF)
 mark_as_advanced(UA_DEBUG_DUMP_PKGS)
 
+option(UA_ENABLE_HARDENING "Enable Hardening measures (e.g. Stack-Protectors and Fortify)" ON)
+mark_as_advanced(UA_ENABLE_HARDENING)
+
 # Build Targets
 option(UA_BUILD_EXAMPLES "Build example servers and clients" OFF)
 option(UA_BUILD_TOOLS "Build OPC UA shell tools" OFF)
@@ -370,6 +376,7 @@ endif()
 # Compiler Settings #
 #####################
 
+include(CompilerFlags)
 if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID}" STREQUAL "xClang"))
     # Compiler
     add_definitions(-std=c99 -pipe
@@ -385,8 +392,33 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
                     -Wnested-externs
                     -Wmultichar
                     -Wundef
-                    -Wc++-compat)
+                    -Wc++-compat
+                    -fno-strict-aliasing # fewer compiler assumptions about pointer types
+                    -fexceptions # recommended for multi-threaded C code, also in combination with C++ code
+                    )
 
+    if (NOT MINGW)
+        if(UA_ENABLE_HARDENING)
+            check_cc_flag("-fstack-protector-strong") # more performant stack protector, available since gcc 4.9
+            check_cc_flag("-fstack-clash-protection") # increased reliability of stack overflow detection, available since gcc 8    
+            check_cc_flag_untested("-mcet -fcf-protection") # future use (control flow integrity protection)
+        endif()
+
+        # IPO requires too much memory for unit tests
+        # GCC docu recommends to compile all files with the same options, therefore ignore it completely 
+        if(NOT UA_BUILD_UNIT_TESTS)
+            # needed to check if IPO is supported (check needs cmake > 3.9)
+            if("${CMAKE_VERSION}" VERSION_GREATER 3.9)
+                cmake_policy(SET CMP0069 NEW) # needed as long as required cmake < 3.9
+                include(CheckIPOSupported)
+                check_ipo_supported(RESULT CC_HAS_IPO) # Inter Procedural Optimization / Link Time Optimization (should be same as -flto)
+                if(CC_HAS_IPO)
+                    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+                endif()
+            endif()
+        endif()
+    endif()
+	
     if(UA_ENABLE_AMALGAMATION)
         add_definitions(-Wno-unused-function)
     endif()
@@ -410,10 +442,18 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
         endif()
     endif()
 
+    if(UA_ENABLE_HARDENING AND CMAKE_BUILD_TYPE STREQUAL "Release")
+        check_cc_flag("-D_FORTIFY_SOURCE=2") # run-time buffer overflow detection (needs at least -O1)    
+    endif()
+
     # Strip release builds
     if(CMAKE_BUILD_TYPE STREQUAL "MinSizeRel" OR CMAKE_BUILD_TYPE STREQUAL "Release")
-        add_definitions(-ffunction-sections -fdata-sections -fno-stack-protector -fno-unwind-tables
+        add_definitions(-ffunction-sections -fdata-sections -fno-unwind-tables
                         -fno-asynchronous-unwind-tables -fno-math-errno -fno-ident)
+        # remove stack-protector with MinSizeRel
+        if (CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
+            add_definitions(-fno-stack-protector)
+        endif()
         if(NOT OS9)
             set(CMAKE_C_LINK_FLAGS "${CMAKE_C_LINK_FLAGS} -s")
             set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -s")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -432,7 +432,7 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
 
     # Debug
     if(BUILD_TYPE_LOWER_CASE STREQUAL "debug")
-        if ("x${CMAKE_C_COMPILER_ID}" STREQUAL "xClang" AND NOT UA_ENABLE_UNIT_TESTS_MEMCHECK)
+        if("x${CMAKE_C_COMPILER_ID}" STREQUAL "xClang" AND NOT UA_ENABLE_UNIT_TESTS_MEMCHECK)
             # Add default sanitizer settings when using clang and Debug build.
             # This allows e.g. CLion to find memory locations for SegFaults
             message("Sanitizer enabled")
@@ -442,7 +442,7 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
         endif()
     endif()
 
-    if(UA_ENABLE_HARDENING AND CMAKE_BUILD_TYPE STREQUAL "Release")
+    if(UA_ENABLE_HARDENING AND (CMAKE_BUILD_TYPE STREQUAL "Release") OR (CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo"))
         check_cc_flag("-D_FORTIFY_SOURCE=2") # run-time buffer overflow detection (needs at least -O1)    
     endif()
 
@@ -451,7 +451,7 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
         add_definitions(-ffunction-sections -fdata-sections -fno-unwind-tables
                         -fno-asynchronous-unwind-tables -fno-math-errno -fno-ident)
         # remove stack-protector with MinSizeRel
-        if (CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
+        if(CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
             add_definitions(-fno-stack-protector)
         endif()
         if(NOT OS9)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
- cmake_minimum_required(VERSION 3.0...3.12)
+cmake_minimum_required(VERSION 3.0...3.12)
 project(open62541)
 # set(CMAKE_VERBOSE_MAKEFILE ON)
 if(${CMAKE_VERSION} VERSION_LESS 3.12)
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,6 +1,8 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.0...3.12)
 project(open62541-examples)
-
+if(${CMAKE_VERSION} VERSION_LESS 3.12)
+    cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
+endif()
 # This examples folder can also be built standalone.
 # First install open62541 using `make install` then
 # copy this folder to any other location and call CMake directly:
--- a/tools/cmake/CompilerFlags.cmake
+++ b/tools/cmake/CompilerFlags.cmake
@@ -1,0 +1,17 @@
+# check if a C compiler flag is possible
+include(CheckCCompilerFlag)
+function(check_cc_flag CC_FLAG)
+    check_c_compiler_flag("${CC_FLAG}" CC_HAS_${CC_FLAG})
+    if(CC_HAS_${CC_FLAG})
+        add_definitions("${CC_FLAG}")
+    endif()
+endfunction()
+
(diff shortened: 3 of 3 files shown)


CMakeLists.txt
```diff
@@ -229,6 +229,27 @@ endif()
 # Compiler Settings #
 #####################
 
+# needed to check if a C compiler flag is possible
+include(CheckCCompilerFlag)
+function(check_cc_flag CC_FLAG)
```
Pro: Can you put these functions into a different file, e.g. somewhere here:
https://github.com/open62541/open62541/tree/master/tools/cmake

tools/travis/travis_linux_script.sh
```diff
@@ -233,28 +233,50 @@ else
         echo -en 'travis_fold:end:script.build.multithread_discovery\\r'
     fi
 
-    echo -e "\r\n== Unit tests (full NS0) ==" && echo -en 'travis_fold:start:script.build.unit_test_ns0_full\\r'
+    #echo -e "\r\n== Unit tests (full NS0; Debug) ==" && echo -en 'travis_fold:start:script.build.unit_test_ns0_full_debug\\r'
```
Pro: Is there a specific reason why you remove the full NS0 Debug build?

CMakeLists.txt
```diff
@@ -244,7 +265,11 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
                     -Wnested-externs
                     -Wmultichar
                     -Wundef
-                    -Wc++-compat)
+                    -Wc++-compat
+                    -fno-strict-aliasing # fewer compiler assumptions about pointer types
+                    -fexceptions # recommended for multi-threaded C code, also in combination with C++ code
+                    -fstack-protector-strong # stack protector - should be default anyway
```
Pro: I remember that there was some issue with the stack protector and using the nodeset compiler.
See:
https://github.com/open62541/open62541/blob/master/CMakeLists.txt#L653

When using the stack protector and the big NS0 file, gcc tries to optimize the file and takes more than 30 seconds compared to currently 1-2 seconds.
Since it is disabled explicitly for the ua_namespace0.c file, it should be fine to set it here.

Pro:
I guess that the timeout in travis is due to some additional optimization performed due to the `O2` flag for the amalgamated full NS0 `open62541.c`. That file is quite huge (10.2**MB**) on my system.

Just ran a test locally with full NS0 and amalgamation:
- With `-O0` build takes around 10 seconds
- With `-O2` after 10 **Minutes** the build is still running

So the problem with O2 is that we can not use it with the amalgamated file. Maybe we should add this option/info to the readme and related to #1700

tools/travis/travis_linux_script.sh
bbara: I thought that a Debug build is basically a subset of the Release build.
So if the unit tests run with the Release, they almost sure run with the Debug build and therefore we can save time.

CMakeLists.txt
bbara: Ah ok nice! I recognized that the strong stackprotector is available only since gcc 4.9 but the travis-ci version is 4.8.4. I might add a check if it is available and activate the "default" protector otherwise (-fstack-protector).

tools/travis/travis_linux_script.sh
Pro: Nope :)
Within the stack we use in some places an `UA_assert`, e.g. here:
https://github.com/open62541/open62541/blob/74b7578cd7e9569beb5bb64098a0dd8ead5cf846/plugins/ua_nodestore_default.c#L280

This assert for example checks, if all the node references are correctly released. `UA_assert` is only enabled on debug builds, to keep the performance impact low, but still show an error during developing.
Therefore it would make sense to build open62541 with all the features enabled and run all the unit tests, to also cover these asserts.

bbara: Thanks, didn't know that! :)
Ok so for now it might be a good idea to run the full NS0 just in debug mode in the CI build routine.

I'm currently thinking about a couple of flags pre-sets for different build and optimization types. Additionally, they should not be set globally but as target properties.

bbara:
Sorry, totally forgot about this one...
What do you think about using toolchain files [1] for each target?
This would allow specifying flags depending on the compiler and target
and maintaining them would be easier.

I added a CMake version handling and made use of the in-build static analyzer options.
cpplint is deactivated for now, as it breaks the build due to some linting errors.

[1] https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html

jpfr:
Good idea to have arch specific compiler settings.
Maybe we can have ‘Hardened’ as a build target. Similar to Debug and MinSizeRel.

It is easy to use the ‘include ‘ cmake directive to execute architecture specific cmake scripts in the current cmake scope.

Pro:
Yep, goot idea. I would also vote for some toolchain files. Our CMakeLists becomes bigger and bigger, so it doesn't hurt to split it up a bit.

@bbara would you have time to update this PR?

Additionally what I also thought is that we should have some CMake options which we can use in CI to enable/disable a set of features, i.e. something like `UA_CI_FULL` or so.

If you look into the CI build scripts, all the CMake options `UA_ENABLE_...` are used multiple times and if you are adding a new feature you need to add it on many places so that it is included in the CI build (travis_linux, travis_osx, appveyor). Not sure though what the best solution for this would be.

bbara:
I will also take a look on #1758 , as this seems to be similar.

Pro:
@bbara should we merge this first to avoid bit rot, and then you can start, based on that, with the better cmake structure?

bbara:
I'm currently thinking about it, but I think it makes sense to rework it a little bit.
IMO, it doesn't make sense to add the protectors to the Release Build.
The optimizers and the additional flags do make sense,
but in general the protectors should just be added to the test runs,
as they should in general not be used in productive code but detect such flaws
and prevent them before reaching Release status.

So my current idea is to add the protectors to the library build process 
when the tests are active or something like that.

bbara:
Hey @Pro ,

finally I found some time... Still much to do but this is one possible way to go.

Some ideas I tried to add:
- set encryption ON as default if the library is found
- new build types. I added a few that make sense imo.
- updated the MinSizeRel build type with minimal default options and add the contrary "experimental".
- added static analyzer during the build process. I think for a working cpplint someone has to clang-format the whole stuff 😁 I added a new target for formatting but it might break the build due to some include resorting and so on. So definitely some work until all 4 (clang-tidy, cppcheck, cpplint and iwyu) can be set active per default.

If you want to have a small build, use MinSize and add your features, if you want to have a feature-rich build, use Experimental and remove the needless features.
Release and Debug use the default options.

Additionally, I'm not sure why ninja was that slow on Travis... 😁 

Let me hear what you think about this approach :)

Pro:
@bbara is there any update on this? It would be a schame if we loose all your changes!

bbara:
@Pro hey. finally some progress again...
Added static analyzer to build process, might be a little more intuitive than executing them as standalone.
Four possible stages: off, minimal (just linting), reduced (basic analyzing too), full (additional cppcheck that checks all the dependencies too - takes LONG)
Did some code style changes to allow a clang-format process that fixes the other stuff to be compliant to cpplint. Maybe it is a good idea to build with reduced by default (forces style checks).
Also activated IPO and enabled hardening by default (for now, just gcc and clang).
nicer cmake structure for better build type/option/compiler handling is still missing...
I will squash the existing commits soon, let me hear what you think.

CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
-cmake_minimum_required(VERSION 3.0)
+ cmake_minimum_required(VERSION 3.0...3.12)
```
Pro: what does this exactly mean? We should try to keep the minimum required version as low as possible.

CMakeLists.txt
```diff
@@ -351,6 +362,112 @@ if(UA_ENABLE_MULTITHREADING)
     MESSAGE(WARNING "UA_ENABLE_MULTITHREADING is enabled. The feature is under development and marked as EXPERIMENTAL")
 endif()
 
+############################
```
Pro: Can we put this whole linting section into a subfolder and just include it here. e.g. into the tools/CMakeLists.txt

This file is already quite long.

CMakeLists.txt
```diff
@@ -351,6 +362,112 @@ if(UA_ENABLE_MULTITHREADING)
...
+    # clang-tidy has certain warnings as errors
+    find_program(CLANG_TIDY_EXE NAMES "clang-tidy")
+    if(CLANG_TIDY_EXE)
+        set(CMAKE_C_CLANG_TIDY "${CLANG_TIDY_EXE};-p=compile_commands.json;-checks=cert-*,performance-*,readability-*,-readability-braces-around-statements,-clang-diagnostic-ignored-optimization-argument;\
+-warnings-as-errors=cert-*,performance-*,readability-*,-readability-braces-around-statements")
+        set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_EXE};-p=compile_commands.json;-checks=cert-*,performance-*,readability-*,-readability-braces-around-statements,-clang-diagnostic-ignored-optimization-argument;\
+-warnings-as-errors=cert-*,performance-*,readability-*,-readability-braces-around-statements")
+    endif()
+endif()
+if(UA_ENABLE_STATIC_ANALYZER STREQUAL FULL)
```
Pro: else if

deps/ziptree.h
```diff
@@ -186,11 +186,11 @@ __##name##_ZIP_FIND(struct type *root, const keytype *key) {            \
     enum ZIP_CMP eq = (cmp)(key, &(root)->keyfield);                    \
     if(eq == ZIP_CMP_EQ) {                                              \
         return root;                                                    \
-    } else if(eq == ZIP_CMP_LESS) {                                     \
+    }                                                                   \
+    if(eq == ZIP_CMP_LESS) {                                            \
```
Pro: Why is this change necessary? It's better to extract this to a new PR.

include/ua_client_subscriptions.h
```diff
@@ -151,7 +151,7 @@ UA_CreateMonitoredItemsResponse UA_EXPORT
 UA_Client_MonitoredItems_createEvents(UA_Client *client,
             const UA_CreateMonitoredItemsRequest request, void **contexts,
             UA_Client_EventNotificationCallback *callbacks,
-            UA_Client_DeleteMonitoredItemCallback *deleteCallback);
+            UA_Client_DeleteMonitoredItemCallback *deleteCallbacks);
```
Pro: Changes like these should not belong to this PR. Better move them to a separate PR and explain why it is necessary to change this.

src/client/ua_client_internal.h
```diff
@@ -206,7 +206,7 @@ openSecureChannel(UA_Client *client, UA_Boolean renew);
 UA_StatusCode
 receiveServiceResponse(UA_Client *client, void *response,
                        const UA_DataType *responseType, UA_DateTime maxDate,
-                       UA_UInt32 *synchronousRequestId);
+                       const UA_UInt32 *synchronousRequestId);
```
Pro: Extract these changes to a separate PR, which we can merge separately.

src/client/ua_client_subscriptions.c
```diff
@@ -199,9 +198,8 @@ UA_Client_Subscriptions_deleteSingle(UA_Client *client, UA_UInt32 subscriptionId
 /* MonitoredItems */
 /******************/
 
-void
-UA_Client_MonitoredItem_remove(UA_Client *client, UA_Client_Subscription *sub,
-                               UA_Client_MonitoredItem *mon) {
+void UA_Client_MonitoredItem_remove(UA_Client *client, UA_Client_Subscription *sub, UA_Client_MonitoredItem *mon) {
```
Pro: Why this reformatting?

src/server/ua_services_discovery_multicast.c
```diff
@@ -62,10 +62,11 @@ static UA_StatusCode
...
     // wake up select
-    if(write(server->discoveryManager.mdnsSocket, "\0", 1)){}; //TODO: move to arch?
-    if(pthread_join(server->discoveryManager.mdnsThread, NULL)) {
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "Multicast error: Can not stop thread.");
+    if (write(server->discoveryManager.mdnsSocket, "\0", 1)) {
+        // TODO: if makes no sense here?
+    }  // TODO: move to arch?
+    if (pthread_join(server->discoveryManager.mdnsThread, NULL)) {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER, "Multicast error: Can not stop thread.");
```
Pro: Try to avoid reformatting or if necessary move to separate PR

src/server/ua_subscription.h
```diff
@@ -18,6 +18,8 @@
 #include "ua_types.h"
 #include "ua_types_generated.h"
 #include "ua_session.h"
+#include "ua_workqueue.h"
+#include "ua_plugin_nodestore.h"
```
Pro: Are these includes really necessary?

tools/travis/travis_linux_script.sh
```diff
@@ -395,6 +395,16 @@ if [ $? -ne 0 ] ; then exit 1 ; fi
 cd .. && rm build -rf
 echo -en 'travis_fold:end:script.build.unit_test_ns0_full\\r'
 
+echo -e "\r\n== clang-format & Static Code Analyzer ==" && echo -en 'travis_fold:start:script.build.clang_format_static_analyzer\\r'
```
Pro: I would put this into the analyze section inside CLang.
This should only be run once for every commit.

https://github.com/open62541/open62541/blob/0b5b11e2f3003ce8e62e9b6237ebb0e444606e7f/tools/travis/travis_linux_script.sh#L102

CMakeLists.txt
bbara: The cmake_minimum_required function internally sets the cmake_policy to the given version,
as described in [1].
This leads to the enforcement of the given cmake_policy, even if a newer cmake version is installed/available.
The range still enforces the given minimum but internally uses a newer version if available
(in this case up to 3.12).
This allows us to use newer features while keeping the low required version.

[1] https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

CMakeLists.txt
bbara: Yes, but I think the division into separate include files should be done in a different PR,
because it includes several other things too.

src/server/ua_subscription.h
bbara: The problem comes up after running clang-format, since the includes get rearranged (sorted by alphabet) and then the build breaks because these two includes were already done by a previous include which is now subsequent and therefore the include is not done yet and the symbols cannot be found.

So if i remember correctly, something like this is the case:
#include "zzz.h" // includes ua_workqueue.h and ua_plugin_nodestore.h
#include "ua_subscription.h"

After rearranging, zzz is after ua_subscription and therefore the workqueue references, which are used in the ua_subscription.h, cannot be found.
(sorry for the weird description, i hope you understand what I mean 😁)

tools/travis/travis_linux_script.sh
bbara: In fact, we can remove this section after a clang-format is done on the whole code base and the fix commits are merged.
I just included it as quality criteria because clang-format or one of the SCAs broke the build.
So this section guarantees, that the project is still building after performing a clang-format on the whole code base, which is not included within this PR.

deps/ziptree.h
bbara: This and several other small styling issues are required to satisfy the static analyzer/linter.
This is required because it is considered as "bad" to have an else if after a previous return.

tools/travis/travis_linux_script.sh
Pro: clang-format will never be completely done for the whole code base. The only thing we could do is to gradually enforce some specific format for the changed lines of a specific PR. Otherwise it leads to many merge conflicts.
E.g. something like this:
https://github.com/material-foundation/clang-format-ci


So express it in other words:
After merging this PR, the whole code base conforms to the REDUCED setting, correct? And `make clangformat` checks that for every new commit?
Shouldn't this be checked with clang-tidy? I'm a bit confused here...

Pro:
Thanks for the additional clarification!

Still I would at least create two separate PRs, one for lint (and the required changes) and then one for hardened cmake settings.

bbara:
Yep, good idea. Will do that.

Pro:
Regarding the automatic clang-format check: That should also be in a third, separate PR where we can discuss that.

E.g. it would be possible to use `git clang-format` to only check the diff of a PR against our format.
See:
https://github.com/ray-project/common/blob/master/.travis/git-clang-format
and
https://github.com/ray-project/common/blob/master/.travis/check-git-clang-format-output.sh
and even improved:
https://github.com/material-foundation/clang-format-ci (which suggests code-changes inline)

bbara:
Cool references, thanks for them!

Regarding the third PR: the automatic clang-format check is part of the linters work.
At least if we enforce our current rules (CPPLINT.cfg; not done in the current CI process), 
this is done implicitly.
Using the linters without a previous clang-format do not make sense, 
because every style violation is a warning and breaks the build during the CI.

From my point of view it is worth a discussion to do a clang-format on the whole code base and
then activate the REDUCED linters as the default cmake setting (ignored if not installed).
Additionally add the clang-format-ci tool as CI "analyze step".

Pro:
We had the discussion about clang-format already almost a year ago. See https://github.com/open62541/open62541/pull/1538#issuecomment-360973092

Reformatting the whole source code is definitely no option. It will create too many merge conflicts for forks and other PRs. The only viable way is to enforce a specific style on the diff of new commits. Like this the code gradually moves to a specific code style.
Probably the last link above is the best way to do it, where you directly get change suggestions.
If you are motivated to create a third PR for the clang-format of the git diff, we would really appreciate it!

CMakeLists.txt
```diff
@@ -410,10 +442,18 @@ if(NOT UA_COMPILE_AS_CXX AND (CMAKE_COMPILER_IS_GNUCC OR "x${CMAKE_C_COMPILER_ID
         endif()
     endif()
 
+    if(UA_ENABLE_HARDENING AND CMAKE_BUILD_TYPE STREQUAL "Release")
```
Pro: This should probably also added for `RelWithDebInfo`?

Issue #1735, opened 2018-04-12
Discussion on Historical Access


peter-rustler-basyskom:
Open62541 currently does not support historical access.

We consider this an important feature and would like to discuss how to implement historical access in open62541.

peter-rustler-basyskom:
This is the first approach suggested by @jpfr in an E-Mail:

> It would be good if we had a plugin API that allows different storage backends for Historical Data Access (HDA).
There exist a wide range of time-series databases for different use cases. SQL databases with a datetime primary key, column-oriented databases, key-value stores, etc.
A specialized time-series database can split the payload for internal analytics. But the most flexible / general approach would be to just store a bytestring with the binary encoded value payload for every timestamp. 

> A not-too-complicated storage backend would be RocksDB. RocksDB is an ordered key-value store. http://rocksdb.org/
If we use UA_DataTime for the key, we have a barebones in-memory timeseries database with on-disk persistence.
Since the keys are ordered, we can also iterate over the elements in a range (time1, time2).
For this, two methods are defined:

>- Key -> Hash
>- (Key1, Key2) -> Equal / Greater / Smaller

> With a HDA storage backend in place, we need a mechanism to trigger the storage of new entries. We can reuse the existing MonitoredItem implementation for this.
If we allow user-defined callbacks for MonitoredItem notifications, we can notify the local user about changes. And the same mechanism can be used to trigger HDA storage.

> The following are proposed steps towards HDA for discussion.

>- Propose a plugin C-API for exchangeable HDA storage backends
>- Implement an example HDA storage backend (e.g. based on RocksDB)
>- Implement an extension of the MonitoredItem implementation to allow callbacks for new notifications. This includes a new public API for "local" notification callbacks.
>- Propose a C-API for adding local HDA "listeners"/subscribers that write into a HDA backend.
>- Implement the HDA "listeners" based on MonitoredItem Notification callbacks
>- Implement the HDA services (readHistorical / updateHistorical)

basyskom-jvoe:
> With a HDA storage backend in place, we need a mechanism to trigger the storage of new entries. We can reuse the existing MonitoredItem implementation for this.
If we allow user-defined callbacks for MonitoredItem notifications, we can notify the local user about changes. And the same mechanism can be used to trigger HDA storage.

If we use the MonitoredItem implementation to trigger the storage of new entries, we would have to sample the node's value. I see two drawbacks of this approach:
- Overhead from periodic sampling, even if the value doesn't change (often)
[...]

PR #1740, opened 2018-04-13
Historical read  server implementation

peter-rustler-basyskom:
This is the historical read server implementation.
It was used a discussion base in Issue #1735 and is in production stage, now.

Works:
 * Historical read of one or multiple notes (ReadRawModifiedDetails with details->isReadModified == false, known as "read raw")
 * plugin based memory store backend
 * continuous read
 * range based read 
 * testing for backend and server
 * data gathering with polling, value set and/or user defined way

Missing:
 * Historical Modify (ReadRawModifiedDetails with details->isReadModified == true)
 * Other Historical data features (ReadEventDetails, ReadProcessedDetails, ReadAtTimeDetails)
 * Database based storage (can be added with plugin interface)


jpfr:
Hey @peter-rustler-basyskom ,

looking good!
This is already a high quality PR.

I have only one major comment.
The PR introduces a local "polling" feature.
In order to avoid duplicate features, I would like to see MonitoredItems used for this "polling".
We then also get all the filtering mechanisms and events handling for free.

I added a feature for local MonitoredItems in this PR: #1793
With this, the user can define a local MonitoredItem and add new entries to the database for historical values with custom code. If no sampling interval shall be used, the user can add a "value callback" instead. That is called for every write operation.

This should reduce the complexity of adding historical data quite a bit.
What are your thoughts?
We can also have a short phone call over the next days.

basyskom-jvoe:
Hi @jpfr,

thank you for having a look at the pull request.

The server-side monitored items look perfect for replacing the periodic callback based approach. The database won't be flooded with unnecessary entries if the trigger is set to STATUS_VALUE and the user can decide if he wants every change reported or just changed above a certain deadband value.

Would you suggest to remove the polling API entirely or just replace its current implementation with a server-side monitored item?

include/ua_plugin_history_data_backend.h
```diff
@@ -0,0 +1,52 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "ua_types.h"
+
+typedef struct {
+    void *context;
+    UA_StatusCode (*getHistoryData)(void * context,
```
jpfr: The plugin API should stay as close as possible to the OPC UA API.
Part 11 of the standard defines HistoryReadDetails, which can be one of 4 subtypes:
- ReadEventDetails
- ReadRawModifiedDetails
- ReadProcessedDetails
- ReadAtTimeDetails

Can we rename `getHistoryData` to `historyRead_rawModifiedDetails` and forward the ReadRawModifiedDetails structure we get from the service call? That way, we can add the other HistoryRead variations as additional callbacks later on without breaking the API.

- It is good to let the server handle continuation points and to forward a "skip" parameter to the plugin.
- What is the difference between maxSize and numValuesPerNode? Both seem to bound the max number of returned values. Since numValuesPerNode is part of the ReadRawModifiedDetails structure, maxSize seems to be redundant.
- The service defines an argument `indexRange` for the HistoryRead service. This can be parsed to `UA_NumericRange` for the internal API. Forward the `UA_NumericRange` to `historyRead_rawModifiedDetails` as well.
- The arguments should be commented to avoid misunderstandings.

include/ua_plugin_history_data_backend.h
```diff
@@ -0,0 +1,52 @@
...
+                                    UA_NodeId * nodeId,
+                                    size_t skip,
+                                    size_t maxSize,
+                                    UA_UInt32 numValuesPerNode,
+                                    UA_Boolean returnBounds,
+                                    UA_DataValue ** result,
+                                    size_t * resultSize,
+                                    UA_Boolean * hasMoreData);
+
+    UA_StatusCode (*addHistoryData)(void * context,
```
jpfr: Again, the HistoryUpdate service has variations defined in Part 11.

For future extensibility, rename the callback to `historyUpdate_updateDataDetails`.
The `UpdateDataDetails` structure can be reused for the plugin API.

include/ua_plugin_nodestore.h
```diff
@@ -177,12 +180,23 @@ typedef enum {
...
+} UA_HistorizingInternal;
+#endif
+
 typedef struct {
     UA_NODE_BASEATTRIBUTES
     UA_NODE_VARIABLEATTRIBUTES
     UA_Byte accessLevel;
     UA_Double minimumSamplingInterval;
-    UA_Boolean historizing; /* currently unsupported */
+    UA_Boolean historizing;
```
jpfr: Can we have a callback in the global HDA plugin to query whether a node is historizing?

include/ua_plugin_nodestore.h
```diff
@@ -177,12 +180,23 @@ typedef enum {
...
+#ifdef UA_ENABLE_HISTORIZING
```
jpfr: I would make the HDA plugin global for the entire server.
Storing the settings in the node itself can be problematic.

When we make a copy of the node, we don't know whether this is to make a new node or to replace the existing node with a modified copy.

src/server/ua_services_attribute.c
```diff
@@ -1342,3 +1360,128 @@ __UA_Server_write(UA_Server *server, const UA_NodeId *nodeId,
...
+void
+Service_HistoryRead(UA_Server *server, UA_Session *session,
+                          const UA_HistoryReadRequest *request,
+                          UA_HistoryReadResponse *response) {
+    UA_HistoryReadResponse_init(response);
+    if (request->historyReadDetails.encoding != UA_EXTENSIONOBJECT_DECODED) {
+        response->responseHeader.serviceResult = UA_STATUSCODE_BADNOTSUPPORTED;
+        return;
+    }
+    if (request->historyReadDetails.content.decoded.type == &UA_TYPES[UA_TYPES_READRAWMODIFIEDDETAILS]) {
```
jpfr: Exit early with `UA_STATUSCODE_BADHISTORYOPERATIONUNSUPPORTED` if the type is unsupported.
This reduces the nesting level of the remaining function.

tests/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ set(test_plugin_sources ${PROJECT_SOURCE_DIR}/plugins/ua_network_tcp.c
                         ${PROJECT_SOURCE_DIR}/plugins/ua_accesscontrol_default.c
                         ${PROJECT_SOURCE_DIR}/plugins/ua_pki_certificate.c
                         ${PROJECT_SOURCE_DIR}/plugins/ua_nodestore_default.c
-                        ${PROJECT_SOURCE_DIR}/tests/testing-plugins/testing_clock.c
```
jpfr: `testing_clock.c` was removed. Why?

include/ua_plugin_history_data_backend.h
```diff
@@ -0,0 +1,52 @@
...
+                                    UA_DataValue * value,
+                                    UA_NodeId * nodeId);
+} UA_HistoryDataBackend;
+
+typedef enum {
```
jpfr: There is now only one option: User monitors the value and sets historical data.
Can be removed.

include/ua_server.h
```diff
@@ -985,9 +987,21 @@ UA_Server_addMethodNode(UA_Server *server, const UA_NodeId requestedNewNodeId,
                                      outputArgumentsSize, outputArguments, UA_NODEID_NULL, NULL,
                                      nodeContext, outNewNodeId);
 }
-
 #endif
 
+#ifdef UA_ENABLE_HISTORIZING
```
jpfr: These can be all removed with the use of local MonitoredItems to update the historical data.

basyskom-jvoe:
@jpfr 

Many thanks for the code review.
@peter-rustler-basyskom will be back in three weeks, let's continue the discussion then.

tests/CMakeLists.txt
peter-rustler-basyskom: It is a double entry.

include/ua_plugin_history_data_backend.h
peter-rustler-basyskom: * Renaming getHistoryData to historyRead_rawModifiedDetails: I am not sure if that is a good idea. I would expect functions more low level for the plugin like insert, replace, update, delete of data. For that we should rename addHistoryData to insertHistoricalData. I would also prefer to have the heavy lifting of the answer of read raw outside of the plugin, but I have to think how to do that. For now there is no good solution. We could make getMatch part of the plugin API and move all the logic into the server outside the plugin, but I am not sure if all databases will support that way of finding values. For ReadProcessedDetails and ReadAtTimeDetails I would expect the server doing the job using getHistoryData. For ReadEventDetails I would expect an own plugin definition (not sure about that).

* Difference between maxSize and numValuesPerNode: numValuesPerNode is the count of requested values by the client. maxSize is the maximum serveable amount of values for one response by the server. So, the server will response with a continuiation point if it can not serve all by one response.

* Forward the UA_NumericRange: At the one hand this sounds good for optimizing database access, on the other hand this is only a good idea for databases that store the real types of data and do not store a bytestring of the serialized data. But if they appear in the future it will be a good idea. Agreed

* arguments commented: agreed

Root-Core:
Is this PR (partially) superseded by #1852 and #1849?
It seems, that some parts of this PR should be implemented as / in a plugin.

Splitting the historical access into multiple pull requests is difficult to comprehend, if you are late to the party.

For my part, there is no need for storing the data within the stack. The device / my software which the server is part of already does that for me and this should be valid for a lot of use cases. So it shouldn't be necessary, nor part of `ua_server.c`.
However, a example / default plugin would be a good place for a logic like this.

IMHO the continuation point should be handled by the server. This would fit the setting of `ServerCapabilities`, where `MaxHistoryContinuationPoints` is defined.

Just my two cents, good work so far. :)

Root-Core:
True, had a similar insight later that day, but no time to edit the comment. It would be good to have some kind of "ContinuationPoint"-Plugin, as CPs aren't  only used by historical access.

peter-rustler-basyskom:
Closing for successor: #1852

PR #1744, opened 2018-04-17
Workaround a commercial SDK messing up the subscription sequence numbers

jpfr:
One of the commercial SDKs (not naming names) is sending wrong subscription sequence numbers from time to time.

See the attached Wireshark dump.

This PR demotes the problem to a warning. So the connection is no longer immediately closed.

Everybody okay with that?

[publish_sequencenumber.zip](https://github.com/open62541/open62541/files/1921915/publish_sequencenumber.zip)

### DIFF
--- a/src/client/ua_client_subscriptions.c
+++ b/src/client/ua_client_subscriptions.c
@@ -638,12 +638,16 @@ UA_Client_Subscriptions_processPublishResponse(UA_Client *client, UA_PublishRequ
     sub->lastActivity = UA_DateTime_nowMonotonic();
 
     /* Detect missing message - OPC Unified Architecture, Part 4 5.13.1.1 e) */
-    if((sub->sequenceNumber != msg->sequenceNumber) && (msg->sequenceNumber != 0) &&
-        (UA_Client_Subscriptions_nextSequenceNumber(sub->sequenceNumber) != msg->sequenceNumber)) {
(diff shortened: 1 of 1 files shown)


ccvca:
Maybe add a comment, that this code is a compatible fix and not the OPC UA Sepcification. This may be changed back to the "correct" way when this is no longer an Issue in the other SDK.

https://github.com/open62541/open62541/pull/1744/files#diff-f4e6b7466f3a54fd8576287d6e1a7b61R646
Should this be deleted or kept commented out?

src/client/ua_client_subscriptions.c
```diff
@@ -638,12 +638,13 @@ UA_Client_Subscriptions_processPublishResponse(UA_Client *client, UA_PublishRequ
...
-                     "Invalid subscritpion sequenceNumber");
-        UA_Client_close(client);
-        return;
+    if(UA_Client_Subscriptions_nextSequenceNumber(sub->sequenceNumber) != msg->sequenceNumber) {
+        UA_LOG_WARNING(client->config.logger, UA_LOGCATEGORY_CLIENT,
+                     "Invalid subscription sequence number: expected %u but got %u",
+                     UA_Client_Subscriptions_nextSequenceNumber(sub->sequenceNumber),
+                     msg->sequenceNumber);
+        /* UA_Client_close(client); */
+        /* return; */
```
ccvca: Should this be deleted or kept commented out

jpfr:
Jep, done @ccvca .

PR #1746, opened 2018-04-19
Replace lastSampledValue with UA_DataValue in Monitored Item

peter-rustler-basyskom:
This is a optimazion followup patch for absolute numeric filter handling.
The following advantages are included:
 * lastSampledValue removed and lastValue renamed to lastSampledValue (size improvement)
 * lastSampledValue is compared to last one without hacky encoding and string compare (easier code without hack)
 * response encoding is done when changes are there, only (performance improvement)
 * UA_DataValue has a isEqual function, so does UA_Variant. Both do not encode or copy data. (update) UA_Variant do encode for complex types. For all usual types there are special equal handling.

See also #1726

### DIFF
--- a/include/ua_types.h
+++ b/include/ua_types.h
@@ -564,6 +564,9 @@ UA_Variant_isEmpty(const UA_Variant *v) {
     return v->type == NULL;
 }
 
+UA_Boolean UA_EXPORT
+UA_Variant_isEqual(const UA_Variant *v1, const UA_Variant *v2);
+
 /* Returns true if the variant contains a scalar value. Note that empty variants
  * contain an array of length -1 (undefined).
  *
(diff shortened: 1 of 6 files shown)


jpfr:
Hello @peter-rustler-basyskom ,

I'm not sure about the value comparison.
Values inside a variant can be complex structures that contain arrays, and so on.
To do a proper comparison, you either need to do the binary encoding or recursively walk over all data type members. And you have custom comparison methods for every "builtin" data type.

The comparison approach in this PR works only for simple data types that contain no pointers.
Such as integers.

peter-rustler-basyskom:
Hello @jpfr,
Thank you for reviewing my patches.
I agree to your comment. I made a commit for that issue.
The commit provide for most usual data type a special handling for the equality check.
The fall back for complex data types is the encoding again. But the encoding is done for the UA_Variant, only. The UA_DataValue is not encoded.
Please review again.

Thanks

jpfr:
PRs are evaluated based on whether
- missing features are added
- the code is simplified (better organized or less lines overall)
- experiments show a significant performance improvement

This PR does not claim to add missing features. Still, more code is added overall.
Significant performance improvements have not been demonstrated.

I'm not sure we can do a lot of improvement in this area of the stack.
Closing for now. Please reopen if I misread the changes and/or the situation has changed.

Some more detail.

Right now, we have the "hack" of using the binary encoding to test for changes. On the upside, this is a trivial and small implementation for a problem that can become quite thorny otherwise. For example if the value is a data structure containing arrays, etc. 

The "filter" code (not the deadband type) removed in this PR is required by the standard.
The user can define in the MonitoredItem whether he wants to trigger a datachange upon a change of the value, the timestamp, the statuscode or a combination thereof. So the PR actually removes functionality.

jpfr:
Edit: I had a second look at the change. I see now that the "ignore filter" is handled properly in the comparison method. Still, I am unsure whether the additional complexity is worth the performance improvements if there are any.

I am taking a hard stance now to save you the trouble of more back and forth.
If you are still passionate about this, the criteria have been made clear.
Meet these and we can be convinced.

PR #1754, opened 2018-04-25
Fix connection close after secure channel renew with background publish

basyskom-jvoe:
The client currently sets the new security token as soon as it receives a OpenSecureChannelResponse. If a publish response with the old tokenId is received after the OpenSecureChannelResponse, the client closes the connection.

This patch saves the new security token to the nextSecurityToken member of UA_SecureChannel. If a message with the new tokenId arrives, the old security token is replaced with the new security token.
If the client is the first to send a message, the token is also replaced.

OPC UA part 4, 5.5.2.1 states that a client should request a new security token after 75% of the old token's  lifetime has elapsed. It also states that a client should accept messages secured with an expired security token for up to 25% of the token lifetime. This is not covered by this patch, so there could still be cases where a message with the old tokenId arrives after the client has switched to the new security token. This should be addressed in a follow up patch.


jpfr:
Thanks for the contribution! Looking forward to the follow-up patch.

PR #1756, opened 2018-04-26
Fix the missing message detection in the client subscription

basyskom-jvoe:
More than one consecutive keepalive message and the first notification message after a keepalive message share the same sequence number.
Don't print a warning for these cases.

This is related to https://github.com/open62541/open62541/commit/792cade2c6d9b05c2d1c69af7ad61a9131ab26c7


jpfr:
We might not even need an additional bool flag for the special case.
We can detect if the subscription message is empty.
And then skip the `sub->sequenceNumber = msg->sequenceNumber;` step.

basyskom-jvoe:
Good idea, I have updated the commit.

jpfr:
Perfect, thanks!

PR #1757, opened 2018-04-26
Generate datatypes only when the feature is used

jpfr:
This PR adds a file datatypes_method.c.
We can use the mechanism for all optional features to reduce the binary footprint.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -503,7 +503,7 @@ if (UA_ENABLE_FULL_NS0)
     if(NOT UA_FILE_NS0)
         set(UA_FILE_NS0 ${PROJECT_SOURCE_DIR}/deps/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml)
     endif()
-    set(UA_FILE_DATATYPES "")
+
     set(UA_FILE_NODEIDS ${PROJECT_SOURCE_DIR}/deps/ua-nodeset/Schema/NodeIds.csv)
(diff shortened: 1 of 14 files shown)


CMakeLists.txt
```diff
@@ -499,33 +499,33 @@ endif()
 # Generate source files #
 #########################
 
-if (UA_ENABLE_FULL_NS0)
+set(UA_FILE_NODEIDS ${PROJECT_SOURCE_DIR}/deps/ua-nodeset/Schema/NodeIds.csv)
+set(UA_FILE_STATUSCODES ${PROJECT_SOURCE_DIR}/deps/ua-nodeset/Schema/StatusCode.csv)
+set(UA_FILE_TYPES_BSD ${PROJECT_SOURCE_DIR}/deps/ua-nodeset/Schema/Opc.Ua.Types.bsd)
```
Pro: Note that we had a discussion that it should be possible to build the minimal nodeset without checking out the ua-nodeset submodule.
See also:
https://github.com/open62541/open62541/commit/c205c4d5455a66a27dab32f9a303db4908139475

I'm fine if we always depend on the submodule :+1:

Pro:
For the long run it would also be nice to be able to select different profiles (https://opcfoundation-onlineapplications.org/profilereporting/) in CMake which enable/disable specific features. That's already a nice start for that

CMakeLists.txt
jpfr: Reversed.

PR #1759, opened 2018-04-27
Merge master into 0.3 (partial)

StalderT:


Pro:
Thanks!
Why only partial?

git checkout master
git merge 0.3
coit as new PR

should work

jpfr:
> Why only partial?

We now have quite a few features on master that are not intended for 0.3.
Also removing deprecated API is not intended for master.

Thanks @StalderT. Looking all good.

Pro:
> We now have quite a few features on master that are not intended for 0.3.
Oh yeah sorry, mistook it for master into 0.3, not the other way around.

Thanks!

PR #1766, opened 2018-05-01
Add option to use immutable nodes

jpfr:


jpfr:
@Pro 
Understood.
This PR is not a good place to introduce arch changes. It was however easiest to expose the atomic operations in public definitions for the plugins to use.

This is now solved with a (dirty) include of a private header. At least until the main arch PR lands.
I'd like to merge this one ahead.

PR #1774, opened 2018-05-07
Fix PubSub compile Error PubSub + Amalgamation; memfix delta message

andreasebner:
Fixed:
1. Enable Amalgamation + PubSub results currently in a compile-error.
2. PubSub Publish test extended, occured mem-error on delta messages fixed.


andreasebner:
@jpfr Thanks for your hint. I changed the cmake configuration now.

jpfr:
#1778 pulls in just the fix ahead of time.

PR #1781, opened 2018-05-15
Fix PubSub compile Error PubSub + Amalgamation

andreasebner:
1. Enable Amalgamation + PubSub results currently in a compile-error. (conditional compilation)
2. Extended PubSub publish test (generate delta messages)


jpfr:
Hey @andreasebner 

I see the following error: https://travis-ci.org/open62541/open62541/jobs/379245838
```
open62541.c:5980:19: fatal error: queue.h: No such file or directory
 #include <queue.h>
```

Use `#include "../deps/queue.h"` instead.

PR #1785, opened 2018-05-17
Allow addition of duplicate references

jpfr:
This speeds up the addition of many nodes considerably. Before, we
looped over many references to see whether the same reference to the
parent already existed.

### DIFF
--- a/src/server/ua_nodes.c
+++ b/src/server/ua_nodes.c
@@ -527,23 +527,11 @@ addReferenceKind(UA_Node *node, const UA_AddReferencesItem *item) {
 
 UA_StatusCode
 UA_Node_addReference(UA_Node *node, const UA_AddReferencesItem *item) {
-    UA_NodeReferenceKind *existingRefs = NULL;
     for(size_t i = 0; i < node->referencesSize; ++i) {
         UA_NodeReferenceKind *refs = &node->references[i];
-        if(refs->isInverse != item->isForward
-                && UA_NodeId_equal(&refs->referenceTypeId, &item->referenceTypeId)) {
-            existingRefs = refs;
-            break;
-        }
-    }
-    if(existingRefs != NULL) {
-        for(size_t i = 0; i < existingRefs->targetIdsSize; i++) {
-            if(UA_ExpandedNodeId_equal(&existingRefs->targetIds[i],
-                                       &item->targetNodeId)) {
-                return UA_STATUSCODE_BADDUPLICATEREFERENCENOTALLOWED;
-            }
-        }
-        return addReferenceTarget(existingRefs, &item->targetNodeId);
+        if(refs->isInverse != item->isForward &&
+           UA_NodeId_equal(&refs->referenceTypeId, &item->referenceTypeId))
+            return addReferenceTarget(refs, &item->targetNodeId);
     }
     return addReferenceKind(node, item);
 }


janitza-thbe:
I introduced detection of duplicate references in #1565 and it was further picked up by @Pro with #1603 by following reasons:

- to detect model / API usage errors by adding references more than once (like in node set compiler that was fixed with #1603)
- removal of reference in loop was assumed to be finished when reference was found (but there could be and definitively were in most cases more than once because of node set compiler bugs)
- needs more memory if there are are identical entries in the references lists
- needs more runtime CPU load to loop over bigger reference lists on certain operations

If you want to reintroduce the errors and ignore the ``BADDUPLICATEREFERENCESNOTALLOWED`` status code that is defined by specification, you have to review/fix the removal of references to, to loop over _ALL_ references from/to a node and not only remove the _First_ matching once since there can be more than once.

Keep in mind that there could come up or already exist certification test cases that check this behaviour.

jpfr:
You are right. That was my fear.
So we keep duplicate checking.
But it will get complicated...

The current performance is not acceptable.

We could use hashing of nodeids (from the nodestore) to order the references in the array. This gives us O(log n) lookup times with binary search. But the insertion time degrades to O(n). And we win nothing.

Maybe we have to put a hashmap into each node after all.

jpfr:
Or we put references in a sorted binary tree...

Stasik0:
Can not we check for duplicates less frequently? For example just at server start for nodes that are inserted in the offline mode? This would remedy the issue of adding many nodes upfront.

zanettea:
In my project I just used a define to disable the check since I know it won't never occur to add duplicate references because:

- I verified once, in debug mode,  the model was correct (no dups)
- as I guess in most of the projects, I don't allow clients using nodemanager services (UA_ENABLE_NODEMANAGEMENT = OFF)
- all dynamically added / removed nodes are strictly under my control, so I can avoid dups by design

I bet many other users can met the above three conditions, but of course the general case requires dups validation. So, why don't you let user enable / disable by define (and for instance force to true if UA_ENABLE_NODEMANAGEMENT is ON or Debug build is ON).

Proposal:

`UA_ENABLE_REFERENCE_DUPLICATES_CHECK=ON|OFF|ASSERT`

- ON : current implementation
- OFF : disable dups checking
- ASSERT : dups checking enabled and abort execution when trying to add dups (can be used in debug mode to easily validate the model)

jpfr:
@zanettea We could target a short-term solution and a long-term solution.

Good idea of letting the user choose.
Maybe we can replace the compiler flag with a runtime configuration.

The short-term solution could be to add a boolean flag `relaxRedundantReferenceConstraint` to the server config. See also here: https://github.com/open62541/open62541/blob/master/include/ua_server_config.h#L122

In the long-term, we ideally do not have to choose.
For example by putting references in an ordered red-black tree instead of an array.
This makes looking up references very fast.
For this, we can use tree.h, which is similar to the queue.h we already use.
http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/tree.h?annotate=1.29

Contributions for both the short-term and the long-term solution are very welcome.

jpfr:
Use of a binary tree for fast reference lookup has already been implemented.
Also, duplicate references (same source, target, reference type) are prevented but don't generate errors.

PR #1788, opened 2018-05-17
If array length is set to 0 it should also be set in the encoding

Pro:
We found this issue with the following sample code (i.e., calling a method with no input args)

```
    UA_CallRequest request2;
    UA_CallRequest_init(&request2);
    UA_CallMethodRequest item;
    UA_CallMethodRequest_init(&item);
    item.methodId = UA_NODEID_NUMERIC(0,100);
    item.objectId = UA_NODEID_NUMERIC(0,1000);
    request2.methodsToCall = &item;
    request2.methodsToCallSize = 1;
    UA_Client_Service_call(client, request2);
```

Note that `UA_CallMethodRequest_init(&item)` sets `inputArgumentsSize` to 0.

If the array length is set to 0, the encoding function will use `i32 signed_length = -1;` instead, since none of the ifs match. So the method call above results in the following wireshark:
image

Whereas it should be 0 array size, as it is done with UaExpert:

image

The `-1` causes a `BadOutOfMemory` result when called on a B&R OPC UA Server.

### DIFF
--- a/src/ua_types_encoding_binary.c
+++ b/src/ua_types_encoding_binary.c
@@ -468,7 +468,7 @@ Array_encodeBinary(const void *src, size_t length, const UA_DataType *type, Ctx 
         return UA_STATUSCODE_BADINTERNALERROR;
     if(length > 0)
         signed_length = (i32)length;
-    else if(src == UA_EMPTY_ARRAY_SENTINEL)
+    else if(src == UA_EMPTY_ARRAY_SENTINEL || length == 0)
         signed_length = 0;
 
     /* Encode the array length */


jpfr:
Ha! They took -1 for an unsigned integer?

With this PR you remove any ability to send null arrays (in contrast to arrays of length 0).

See UA_Array_new for the use of the sentinel pointer.

The docs of variant may also be helpful.

Maybe we can improve the documentation to make people more aware of the different empty array types and how to work with them.

Pro:
> Ha! They took -1 for an unsigned integer?

Looks like it. Not sure though which OPC UA stack they are using. I would guess it's a commercial one. Maybe I can find out more on this...

> With this PR you remove any ability to send null arrays (in contrast to arrays of length 0).

Oh, right... Not good

It is also a bit misleading if I explicitly set the length of the array to 0, but the encoding changes it to -1 and sends that over the wire. We could also initialize an array per default to size -1, instead of 0, and then send 0 if the size is set to 0.

Should we then change the initialization of `UA_CallMethodRequest_init(&item);` so that `inputArguments` is set to the empty sentinel so that we have the same data as UaExpert? Or do we want to ignore this issue (since it is a fault on the B&R side)?

Pro:
It looks like the problematic server is from Unified Automation (we just had another similar problem with a camera).

image

I will try to reproduce it with a trial version.

Pro:
I tried to reproduce it with the 1.5.6 version of the C++ SDK, but there it works. So it looks like the users of the SDK did some bad casting. I need to investigate that further. Until I do not have anything new I will close this PR since our stack is working correctly.

Pro:
A bugfix was now added to the B&R side. Our server is behaving correctly, so closing as wontfix

PR #1793, opened 2018-05-21
Add local MonitoredItems for the server

jpfr:
With this, the server can have local MonitoredItems with a callback method and context pointer each.


Pro:
Nooooo, this causes conflicts with the events PR :(

@jpfr Can we try to avoid making such parallel PRs in the future? It's not so much fun to resolve merge conflicts if they could be avoided easily...

jpfr:
Ouch. Sorry.

This was to support ongoing development of #1740.
We always have a couple of overlapping long-running PR at the same time.

I had a look at the conflicts. Most are pretty benign.
Only `ua_subscription_datachange.c` requires some effort.

Please tell me if I can chip in somehow.
I'll make sure to synchronise in the future.

Pro:
No problem, I delegated it to Ari, he already has unit tests almost ready and can then test the resolved version immediately

PR #1668, opened 2018-03-19
Client performance issue on large responses

anzeskerjanc:
## Problem:
Large responses on server are divided into 64k chunks.
Client for every received chunk increases (UA_realloc) complete message size for one chunk size(64k). On large responses this means a lot of memory allocating for increasing buffer size just for 64k.

## Solution:
Increase memory size for more than just newly arrived chunk.

PR #1797, opened 2018-05-23
Client performance issue on large responses

anzeskerjanc:
New PR from https://github.com/open62541/open62541/pull/1668


jpfr:
Hey @anzeskerjanc ,

Thanks for the contribution.

This adds a linked list (chunk part) inside a linked list (chunks).
I would like to avoid this complexity.

Can we maybe directly allocate "max-chunk-size" for the chunk-entries?
Then we don't have to reallocate at all.

anzeskerjanc:
I don't think this is always a good idea. Especially when sometimes you can expect really large chunks.

jpfr:
> I don't think this is always a good idea. Especially when sometimes you can expect really large chunks.

It is a good idea especially if you expect large chunks.
So the allocated buffer will nearly always be full.
We then don't have to allocate double the memory just to assemble the chunk.

If memory is an issue, the max chunk size can be reduced.
Note that the max message size can still be large.

anzeskerjanc:
Then you put me before hard decision about setting max chunk size :) 
+ we adding more library settings. 

For example, when you want to transfer data at 1Gbps, then you should set max chunk size at least 50MB. Do you really want to allocate 50MB for every chunk? 

Less complex solution is my first with exponent growing buffer.

jpfr:
> Then you put me before hard decision about setting max chunk size :)
> we adding more library settings.

We already configure a maximum receiveBufferSize.
No chunk can be bigger than that.

https://github.com/open62541/open62541/blob/b0403147390d546e060bb64146739b4e7f6a4990/include/ua_plugin_network.h#L45

> For example, when you want to transfer data at 1Gbps, then you should set max chunk size at least 50MB. Do you really want to allocate 50MB for every chunk?

The absolute limitation on TCP packet size is 64K (65535 bytes), but in practicality this is far larger than the size of any packet you will see, because the lower layers (e.g. ethernet) have lower packet sizes.
https://stackoverflow.com/questions/2613734/maximum-packet-size-for-a-tcp-connection

> Less complex solution is my first with exponent growing buffer.

Behind the scenes, realloc also usually grows memory to some power of two.

There is no clear argument.
At this point, it would be good to have a way to make measurements. So we can trade off complexity against performance.

What would be a good setup to test performance?

anzeskerjanc:
[chunk.txt](https://github.com/open62541/open62541/files/2050618/chunk.txt)

This is my simple performance test code, which isolate time of UA_SecureChannel_appendChunk and UA_SecureChannel_finalizeChunk. 

Test time with different implementations (Windows 10, Intel i7, 16GB):
linked list: 480ms
exponent growing buffer: 842ms
original implementation(realloc): 92635ms

jpfr:
Wow! This are huge improvements.

The standard says in part 6:
> ReceiveBufferSize  The largest MessageChunk that the sender can receive.

What about immediately allocating connection->localConf.recvBufferSize for the chunk?
Then we can append without a realloc / growing+copying the buffer.
And it would save the malloc and memcpy in UA_SecureChannel_finalizeChunk.

Can this bring us down to ~300ms?

jpfr:
By the way. You are making a 25MB chunk.
The current default recvBufferSize is 65kb.

Maybe we can measure meaking a 25MB message out of 65kb chunks.

jpfr:
Aaah. Please ignore my last comments.
The PR is perfect for what it does.

This is in effect assembling the message and not assembling the chunk.
My mistake. Chunks itself can also be split/joined on the TCP/IP level.
I was on the wrong level.

We however should rename some structures to make the code more clear.
Putting a `ChunkData` inside a `ChunkEntry` is not exactly telling.
What about having a `ChunkPayload` inside a `MessageEntry`?

anzeskerjanc:
I agree with you. I will change the names.

jpfr:

PR #1807, opened 2018-05-25
Sixth batch for PubSub support

andreasebner:
This PubSub update includes the first part of the information model representation for the PubSub enteties. The representation can be enabled/disabled with the new cmake flag: UA_ENABLE_PUBSUB_INFORMATIONMODEL.

Currently are the objects for WriterGroups, PubSubConnections, DataSetWriter and PublishedDataSets created automatically (without all propertys).

The WriterGroup PublishingInterval is read- and writeable and can be changed under the usage of the standard OPC UA metods. The updateded value is used imediatly. The upcoming PubSub updates will use the same method for all propertys.

This introduces the following API changes:
NEW
UA_WriterGroup_updateConfig: Update the config of a exising WriterGroup. The update function is also used from the information model implemenation for update propertys.


src/pubsub/ua_pubsub_ns0.c
```diff
@@ -0,0 +1,353 @@
...
+    UA_BrowsePathResult_deleteMembers(&bpr);
+    return resultNodeId;
+}
+
+static void
+onRead(UA_Server *server, const UA_NodeId *sessionId, void *sessionContext,
+       const UA_NodeId *nodeid, void *nodeContext,
+       const UA_NumericRange *range, const UA_DataValue *data) {
+    UA_Variant value;
+    memset(&value, 0, sizeof(UA_Variant));
```
jpfr: UA_Variant_init does the same and is more readable.

jpfr:
Perfect! Merged.

Issue #1730, opened 2018-04-10
Check if methods are correctly linked/deleted on object inheritance


Pro:
This transforms the mail by @janitza-thbe into an issue report so that we do not forget about it

-------------

`nodemanagement.c:deleteNodeOperation()` has two open ToDos regarding model checking:

* `nodemanagement.c:copyChildNode()` only links methods of objects by simply adding a reference, instead of copying the method node of the object type. For Mandatory nodes that's correct.
For later created optional objects (flag `forceCreateOptional`) which are then deleted, this may cause the deletion of the method node itself, and not only the link. (This still needs to be verified, if it behaves like this). Therefore other object type instances may refer to an invalid method node.

* Additionally it is not yet clear if simply linking the object type method is allowed. UaModeler explicitly creates a copy of instantiated objects, and not only links them. Using the attribute `methodDeclarationId` it refers to the method of the object type. In this way open62541 and UaModeler behave differently.

Pro:
We now also ran into a related problem with linking the method nodes:

As already mentioned above, we are currently simply adding a reference from the Object to the Method in the ObjectType if the method is set to mandatory (as already pointed out by @janitza-thbe):

https://github.com/open62541/open62541/blob/5acb762c1430c9e0de0e19efcf262664b8c06c13/src/server/ua_services_nodemanagement.c#L475

This means that the method id for the ObjectType and Object are both the same.

The problem now arises on how one can set different method callbacks for different instances of the ObjectType. E.g. lets say we have a `MamalType` with a method `Greet`. Now I create two instances, a dog and a cat. And using `UA_Server_setMethodNode_callback` I can only pass the method ID, which is the same for both instances, so there is no way to set two different callbacks for the inherited method. (The corresponding example code is attached at the end).

So there are two solutions for this problem:
1. Do not simply link, but copy the method and create a new ID
2. Or extend the `UA_Server_setMethodNode_callback` to use the method ID and the object ID/owner. But this would mean quite some changes in the code since we cannot simply change a node property anymore.

@jpfr I would go for solution 1 and simply change the `copyChildNodes` method to also copy method nodes. Do you agree?

-------------

```c
#include <signal.h>
#include "open62541.h"

UA_Boolean running = true;
static void stopHandler(int sig) {
    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND, "received ctrl-c");
    running = false;
}

static UA_StatusCode animalGreeter(UA_Server *server,
[...]

PR #1812, opened 2018-05-28
Copy method node instead of referencing it

cabralfortiss:
This is a test solution for #1730. I tested locally and it worked. This pull request will help trigger a CI we have.

### DIFF
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -470,24 +470,11 @@ copyChildNode(UA_Server *server, UA_Session *session,
     if(!isMandatoryChild(server, session, &rd->nodeId.nodeId))
         return UA_STATUSCODE_GOOD;
 
-    /* No existing child with that browsename. Create it. */
-    if(rd->nodeClass == UA_NODECLASS_METHOD) {
-        /* Add a reference to the method in the objecttype */
-        UA_AddReferencesItem newItem;
-        UA_AddReferencesItem_init(&newItem);
-        newItem.sourceNodeId = *destinationNodeId;
-        newItem.referenceTypeId = rd->referenceTypeId;
-        newItem.isForward = true;
-        newItem.targetNodeId = rd->nodeId;
-        newItem.targetNodeClass = UA_NODECLASS_METHOD;
-        Operation_addReference(server, session, NULL, &newItem, &retval);
-        return retval;
-    }
-
-    /* Node exists and is a variable or object. Instantiate missing mandatory
+    /* Node exists and is a variable, object or method. Instantiate missing mandatory
      * children */
     if(rd->nodeClass == UA_NODECLASS_VARIABLE ||
-       rd->nodeClass == UA_NODECLASS_OBJECT) {
+       rd->nodeClass == UA_NODECLASS_OBJECT ||
+       rd->nodeClass == UA_NODECLASS_METHOD) {
         /* Get the node */
         UA_Node *node;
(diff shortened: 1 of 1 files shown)


jpfr:
This solves one special case where methods are children to several parents.

The general case is that objects, variables and methods can all have several parents.
How could we solve the general case that nodes are only auto-deleted when their last parent is deleted?

Pro:
@jpfr could we use the following approach to automatically delete nodes when the last parent is deleted:

When deleting node `n1`, we check all its children ( e.g. `c1` and `c2`). A children is only a node which is referenced through a `HierarchicalReferences` i=33 or any subtype of that, which also includes `HasComponent`, `HasSubtype` and `Organizes`.

In `c1` we check if there is any other node except `n1` with `c1` as the reference target and hierarchical reference type (including subtypes). If yes, then do not delete `c1`. If there aren't any other nodes we can delete `c1` and continue recursively.

Does that solve the issue?

jpfr:
We can do that for some nodes. But not for all.
Objects do not need a parent. And some objects do not have a parent.
There are even a few examples in the (full) ns0.

What are the cases when we need to trigger a check?

sqadasi:
> 
> 
> We can do that for some nodes. But not for all.
> Objects do not need a parent. And some objects do not have a parent.
> There are even a few examples in the (full) ns0.
> 
> What are the cases when we need to trigger a check?

Objects without parents should be deleted without a problem in my opinion, because no other node have them as a child. So deleting them won't affect other nodes. However, parentless objects may have children with the hierarchical references "HasChild or its Subtypes (Aggregates or HasSubType)" to other objects, and deleting those children will remove them also from their other parents tree. For example, if we delete an object instance which has method nodes (not copied, only referenced), the method nodes will be removed from the ObjectType Node.

In the special case of methods, which are only referenced when creating object instances, we could exclude them when deleting their object instance parents. So the current implementation of referencing methods doesn't necessarily have to be changed to create copies of methods.

Pro:
> We can do that for some nodes. But not for all.
> Objects do not need a parent. And some objects do not have a parent.
> There are even a few examples in the (full) ns0.

Can we add an additional internal boolean, which indicates if the node should be deleted when it has no children?
By default this is set to true, and only for some special cases (e.g. the one from NS0) we set it to false.

And we add a User API which allows to set this boolean. Like this we delete the nodes by default, but give the user the possibility to disable auto deletion.

> So the current implementation of referencing methods doesn't necessarily have to be changed to create copies of methods.

We have to change that, otherwise it's not possible to change the callback for a specific instance of a type. See https://github.com/open62541/open62541/issues/1730#issuecomment-391395119 for further explanation

jpfr:
> Objects without parents should be deleted without a problem in my opinion,

Nope. There are objects without parents. For example ModellingRule_Mandatory.

> Can we add an additional internal boolean, which indicates if the node should be deleted when it has no children?

Can we do it the other way around?
Delete the node when the only parent is deleted.
We already do it similarly. *But*: We currently don't check if there are other parents.
That's not good.

https://github.com/open62541/open62541/blob/master/src/server/ua_services_nodemanagement.c#L1160

> We have to change that, otherwise it's not possible to change the callback for a specific instance of a type. See #1730 (comment) for further explanation

I consider that bad practice. You also don't switch method implementations for instances of a Java class.
The sane way would be to test internally which object calls the method.
An if required, one can always manually replace the method node with a custom one.

Pro:
See https://github.com/open62541/open62541/issues/1730#issuecomment-527492175

PR #1814, opened 2018-05-29
SEGV fix for subscription data change

janitza-thbe:
Fixed illegal access to possible ``nullptr`` list element.

### DIFF
--- a/src/server/ua_subscription_datachange.c
+++ b/src/server/ua_subscription_datachange.c
@@ -61,7 +61,8 @@ MonitoredItem_delete(UA_Server *server, UA_MonitoredItem *monitoredItem) {
     }
 
     /* Remove the monitored item */
-    LIST_REMOVE(monitoredItem, listEntry);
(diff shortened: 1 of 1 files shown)


jpfr:
Hello @janitza-thbe ,

thanks for the fix.
What are the conditions to reproduce the error?

Maybe we can fix the error at the source instead of adding a check here...

janitza-thbe:
The error happens on connecting with an OPC UA client.
Under special circumstances in ``ua_services_subscription.c::Operation_CreateMonitoredItem()`` either the call to ``UA_NodeId_copy()`` or the once to ``setMonitoredItemSettings()`` could fail and the monitored item is deleted by help of the affected method ``MonitoredItem_delete()`` directly afterwards.
Inside ``MonitoredItem_delete()`` the macro ``LIST_REMOVE()`` is executed conditionless but the monitored item has never been added to the list (from previous caller) before.

jpfr:
Thanks for looking into it. Merged. 😄

PR #1816, opened 2018-05-29
Event improvements

jpfr:
@Pro
@aDogCalledSpot 

Can you review?

We still need to look into the handling of the event fields.
The severity field should be transmitted with every EventNotification.
But it isn't.

### DIFF
--- a/src/server/ua_subscription_datachange.c
+++ b/src/server/ua_subscription_datachange.c
@@ -65,13 +65,9 @@ void UA_MonitoredItem_delete(UA_Server *server, UA_MonitoredItem *monitoredItem)
         UA_Notification *notification, *notification_tmp;
         TAILQ_FOREACH_SAFE(notification, &monitoredItem->queue,
                            listEntry, notification_tmp) {
-            /* Remove the item from the queues */
-            TAILQ_REMOVE(&monitoredItem->queue, notification, listEntry);
-            TAILQ_REMOVE(&sub->notificationQueue, notification, globalEntry);
-            --sub->notificationQueueSize;
-            UA_Notification_delete(notification);
+            /* Remove the item from the queues and free the memory */
+            UA_Notification_delete(sub, monitoredItem, notification);
         }
-        monitoredItem->queueSize = 0;
     }
 
 #ifdef UA_ENABLE_SUBSCRIPTIONS_EVENTS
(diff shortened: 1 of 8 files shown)


src/server/ua_subscription_datachange.c
```diff
@@ -171,6 +162,9 @@ MonitoredItem_ensureQueueSpace(UA_Server *server, UA_MonitoredItem *mon) {
                 TAILQ_INSERT_TAIL(&mon->queue, overflowNotification, listEntry);
                 TAILQ_INSERT_TAIL(&mon->subscription->notificationQueue, overflowNotification, globalEntry);
             }
+            ++mon->queueSize;
+            ++sub->notificationQueueSize;
+            ++sub->eventNotifications;
```
aDogCalledSpot: Why is the queue size increased here? How is the while loop supposed to terminate?

jpfr: That is a good observation,

The queue size is increase because it represents the number of elements in the queue.
And since we add an element...

Without the increase, we get an unsigned underflow.
With the increase, we ever exit the loop.

Where is the handling of overflow events described in the standard?
So we can reference that.

src/server/ua_subscription_events.c
```diff
@@ -256,40 +256,55 @@ UA_Server_filterEvent(UA_Server *server, const UA_NodeId *eventNode, UA_EventFil
...
+        UA_ReadValueId_init(&rvi);
+        rvi.nodeId = bpr.targets[0].targetId.nodeId;
+        rvi.attributeId = clause->attributeId;
+
+        UA_DataValue v = UA_Server_read(server, &rvi, UA_TIMESTAMPSTORETURN_NEITHER);
+        notification->fields.eventFields[i] = v.value;
         UA_BrowsePathResult_deleteMembers(&bpr);
+
+        UA_Boolean whereClauseResult = UA_TRUE;
+        return whereClausesApply(server, filter->whereClause, &notification->fields, &whereClauseResult);
```
aDogCalledSpot: The `return` should be outside of the loop. Doing this ensures all data is copied correctly.

src/server/ua_subscription_datachange.c
aDogCalledSpot: Part 4 5.12.1.5 (page 61).

"If Events are lost, an Event of the type
_EventQueueOverflowEventType_ is placed in the queue. [...] It is put into the Queue of
the _MonitoredItem_ in addition to the size of the Queue defined for this _MonitoredItem_"

Sounds like it would be best to keep count of the overflows separately.

jpfr:
@Pro @aDogCalledSpot 
Some features were deliverately commented out until we have a good solution.

Now it is again up to you.

PR #1829, opened 2018-06-01
Add support for EventQueueOverflowEvents

aDogCalledSpot:
Since the specification only states that the MonitoredItems' queue sizes are unaffected by EventQueueOverflowEvents, an additional counter has been added to MonitoredItems to keep track of the "actual" amount of items in the queue, even if the queueSize only reflects events which aren't overflows.
The subscription queues do not handle OverflowEvents in a special way.

@jpfr Please review.

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -303,14 +303,6 @@ prepareNotificationMessage(UA_Server *server, UA_Subscription *sub,
             UA_EventFieldList_init(&notification->data.event.fields);
             efl->clientHandle = mon->clientHandle;
             enlPos++;
-
-            /* increase the monitoredItem queueSize if an eventOverflow was moved,
-             * so it remains the same when the size is reduced in UA_Notification_delete */
-            UA_NodeId overflowId = UA_NODEID_NUMERIC(0, UA_NS0ID_SIMPLEOVERFLOWEVENTTYPE);
-            if (efl->eventFieldsSize == 1 && efl->eventFields[0].type == &UA_TYPES[UA_TYPES_NODEID]
-                    && UA_NodeId_equal((UA_NodeId *)efl->eventFields[0].data, &overflowId)) {
-                mon->queueSize++;
-            }
         }
 #endif
         else {
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -297,24 +297,20 @@ prepareNotificationMessage(UA_Server *server, UA_Subscription *sub,
         } else if(mon->monitoredItemType == UA_MONITOREDITEMTYPE_EVENTNOTIFY && enl) {
             UA_assert(enl != NULL); /* Have at least one event notification */
 
-            /* TODO: The following lead to crashes when we assumed notifications to be ready... */
-            /* /\* removing an overflowEvent should not reduce the queueSize *\/ */
-            /* UA_NodeId overflowId = UA_NODEID_NUMERIC(0, UA_NS0ID_SIMPLEOVERFLOWEVENTTYPE); */
-            /* if (!(notification->data.event.fields.eventFieldsSize == 1 */
-            /*       && notification->data.event.fields.eventFields->type == &UA_TYPES[UA_TYPES_NODEID] */
-            /*       && UA_NodeId_equal((UA_NodeId *)notification->data.event.fields.eventFields->data, &overflowId))) { */
-            /*     --mon->queueSize; */
-            /*     --sub->notificationQueueSize; */
-            /* } */
-
             /* Move the content to the response */
             UA_EventFieldList *efl = &enl->events[enlPos];
             *efl = notification->data.event.fields;
             UA_EventFieldList_init(&notification->data.event.fields);
             efl->clientHandle = mon->clientHandle;
-            /* EventFilterResult currently isn't being used
-               UA_EventFilterResult_deleteMembers(&notification->data.event.result); */
             enlPos++;
+
+            /* increase the monitoredItem queueSize if an eventOverflow was moved,
+             * so it remains the same when the size is reduced in UA_Notification_delete */
+            UA_NodeId overflowId = UA_NODEID_NUMERIC(0, UA_NS0ID_SIMPLEOVERFLOWEVENTTYPE);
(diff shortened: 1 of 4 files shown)


src/server/ua_subscription_datachange.c
```diff
@@ -126,46 +126,61 @@ MonitoredItem_ensureQueueSpace(UA_Server *server, UA_MonitoredItem *mon) {
...
+
+             UA_EventFieldList_init(&overflowNotification->data.event.fields);
+             overflowNotification->data.event.fields.eventFields = UA_Variant_new();
+             if(!overflowNotification->data.event.fields.eventFields) {
+                 UA_EventFieldList_deleteMembers(&overflowNotification->data.event.fields);
+                 UA_free(overflowNotification);
+                 return UA_STATUSCODE_BADOUTOFMEMORY;
+             }
+
+             UA_Variant_init(overflowNotification->data.event.fields.eventFields);
```
jpfr: _init not required after _new

src/server/ua_subscription_datachange.c
```diff
@@ -126,46 +126,61 @@ MonitoredItem_ensureQueueSpace(UA_Server *server, UA_MonitoredItem *mon) {
...
+             overflowNotification->data.event.fields.eventFieldsSize = 1;
+             UA_Variant_setScalarCopy(overflowNotification->data.event.fields.eventFields,
```
jpfr: Check the return value of _copy

src/server/ua_subscription_datachange.c
```diff
@@ -126,46 +126,61 @@ MonitoredItem_ensureQueueSpace(UA_Server *server, UA_MonitoredItem *mon) {
...
+                 TAILQ_INSERT_HEAD(&mon->queue, overflowNotification, listEntry);
+                 TAILQ_INSERT_HEAD(&mon->subscription->notificationQueue, overflowNotification, globalEntry);
+             } else {
+                 TAILQ_INSERT_TAIL(&mon->queue, overflowNotification, listEntry);
+                 TAILQ_INSERT_TAIL(&mon->subscription->notificationQueue, overflowNotification, globalEntry);
+             }
+
+             /* removing an OverflowEvent and adding an OverflowEvent causes the amount
+              * of overflows to remain the same */
+             if (!deletingOverflowEvent) {
```
jpfr: For deletingOverflowEvent, why do we even bother removing the old notification?
We add a new overflowEvent with exactly the same content....

src/server/ua_subscription_datachange.c
```diff
@@ -126,46 +126,61 @@ MonitoredItem_ensureQueueSpace(UA_Server *server, UA_MonitoredItem *mon) {
...
+                 ++mon->eventOverflows;
```
jpfr: mon->eventOverflows is only written into.
But it is never read.

src/server/ua_subscription.c
```diff
@@ -295,24 +295,20 @@ prepareNotificationMessage(UA_Server *server, UA_Subscription *sub,
...
             /* Move the content to the response */
             UA_EventFieldList *efl = &enl->events[enlPos];
             *efl = notification->data.event.fields;
             UA_EventFieldList_init(&notification->data.event.fields);
             efl->clientHandle = mon->clientHandle;
-            /* EventFilterResult currently isn't being used
-               UA_EventFilterResult_deleteMembers(&notification->data.event.result); */
             enlPos++;
+
+            /* increase the monitoredItem queueSize if an eventOverflow was moved,
```
jpfr: Can we keep mon->queueSize the same as the actual length of the linked list?
Adding these special cases is just a magnet for future bugs.

src/server/ua_subscription_datachange.c
```diff
@@ -128,46 +128,64 @@ MonitoredItem_ensureQueueSpace(UA_Server *server, UA_MonitoredItem *mon) {
...
-        /* } */
+         if(mon->monitoredItemType == UA_MONITOREDITEMTYPE_EVENTNOTIFY) {
+             /* check if an overflowEvent is being deleted
+              * TODO: make sure overflowEvents are never deleted */
+             UA_NodeId overflowId = UA_NODEID_NUMERIC(0, UA_NS0ID_SIMPLEOVERFLOWEVENTTYPE);
+
+             /* Check if an OverflowEvent is being deleted */
+             if (del->data.event.fields.eventFieldsSize == 1
+                   &&  del->data.event.fields.eventFields[0].type == &UA_TYPES[UA_TYPES_NODEID]
+                   &&  UA_NodeId_equal((UA_NodeId *)del->data.event.fields.eventFields[0].data, &overflowId)) {
```
Pro: Here we also need to check for any subtypes of overflow events, not only for `overflowId`. I.e. use `isNodeInTree` to check if `(UA_NodeId *)del->data.event.fields.eventFields[0].data` is a child of OverflowEventType. Or do we only have the one `SimpleOverflowEventType` type? Maybe in the future we support other overflow types...

src/server/ua_subscription_datachange.c
```diff
@@ -128,46 +128,64 @@ MonitoredItem_ensureQueueSpace(UA_Server *server, UA_MonitoredItem *mon) {
...
-        /*         TAILQ_INSERT_TAIL(&mon->queue, overflowNotification, listEntry); */
-        /*         TAILQ_INSERT_TAIL(&mon->subscription->notificationQueue, overflowNotification, globalEntry); */
-        /*     } */
-        /*     ++mon->queueSize; */
-        /*     ++sub->notificationQueueSize; */
-        /*     ++sub->eventNotifications; */
-        /* } */
+         if(mon->monitoredItemType == UA_MONITOREDITEMTYPE_EVENTNOTIFY) {
+             /* check if an overflowEvent is being deleted
+              * TODO: make sure overflowEvents are never deleted */
```
Pro: Is this TODO still relevant?

aDogCalledSpot: Yes, it is in accordance with the specification Part 4: 5.12.1.5

> " If discardOldest is set to TRUE [the EventQueueOverflowEvent] is put at the beginning of the queue and is never discarded"

PR #1833, opened 2018-06-03
Enable PubSub tests and static code analysis

andreasebner:
Enable PubSub unit tests and static code analysis for tranvis.


Pro:
Should we also enable it on appveyor + drMemory?

jpfr:
Delta frames off by default?

I have no opinion here. A bit of documentation would be good though.

jpfr:
> Should we also enable it on appveyor + drMemory?

It can only help..

andreasebner:
@Pro @jpfr I will fix the current travis/test problems and enable the PubSub port for appveyor + drMemory.

andreasebner:
@jpfr The PubSub option is disabled by default. The UA_ENABLE_PUBSUB_DELTAFRAMES make only sense while PubSub is enabled. Because of this i changed the default to OFF. If you like, we can change this back. I will update the PR and add some documentation for the flags.

jpfr:
> The PubSub option is disabled by default. The UA_ENABLE_PUBSUB_DELTAFRAMES make only sense while PubSub is enabled. 

Cool.

PR #1841, opened 2018-06-11
Add server helper function "getNamespaceByName"

mlang-de:
Request namespace index by namespace uri from server.

### DIFF
--- a/include/ua_server.h
+++ b/include/ua_server.h
@@ -1071,7 +1071,9 @@ UA_Server_deleteReference(UA_Server *server, const UA_NodeId sourceNodeId,
 UA_UInt16 UA_EXPORT UA_Server_addNamespace(UA_Server *server, const char* name);
 
 /* Get namespace by name from the server. */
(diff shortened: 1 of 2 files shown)


jpfr:
Thanks for the contribution!

Just two minor points:

- Travis returns the following log message:
```
/tmp/open62541/build/open62541.c:22540:84: error: implicit declaration of function 'MAX' [-Werror=implicit-function-declaration]
     if(0 == strncmp((char*)server->namespaces[idx].data, (char*)namespaceUri.data, MAX(server->namespaces->length, namespaceUri.length)))
```
- Can we reduce the line-length for the public header file?
We generate PDF documentation from the source-file. And long lines result in ugly breaks.
There are a lot of other examples you can take as orientation.

mlang-de:
I guess the changes are good for now.
What is the max line length in general?

jpfr:
Perfect! Thanks for the contribution.

PR #1846, opened 2018-06-12
Move number of endpoints check to UA_Server_run_startup

mlang-de:
Some systems (for instance embedded systems) could get there IP-configuration late. This time could be used to create the namespace as well as spend some time to parsing files (e.g. XML for address space).
Other case is that the OPC UA Server is part of a whole system will be started first on request (e.g. after 3 hours runtime), but the address space creation should be done each time at system Startup phase.
All code changes do not change any already available behaviour and do not break any API, since the variable lateEndpointCreate in struct UA_ServerConfig will be initialiazed with 0==false.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -244,13 +244,6 @@ UA_Server_new(const UA_ServerConfig *config) {
     if(!config)
         return NULL;
 
-    /* At least one endpoint has to be configured */
-    if(config->endpointsSize == 0) {
-        UA_LOG_FATAL(config->logger, UA_LOGCATEGORY_SERVER,
-                     "There has to be at least one endpoint.");
-        return NULL;
-    }
-
     /* Allocate the server */
     UA_Server *server = (UA_Server *)UA_calloc(1, sizeof(UA_Server));
     if(!server)
--- a/src/server/ua_server_worker.c
+++ b/src/server/ua_server_worker.c
@@ -305,6 +305,12 @@ UA_StatusCode
 UA_Server_run_startup(UA_Server *server) {
     UA_Variant var;
     UA_StatusCode result = UA_STATUSCODE_GOOD;
+	
+	/* At least one endpoint has to be configured */
+    if(server->config.endpointsSize == 0) {
+        UA_LOG_WARNING(server->config.logger, UA_LOGCATEGORY_SERVER,
+                       "There has to be at least one endpoint.");
+    }
 
     /* Sample the start time and set it to the Server object */
     server->startTime = UA_DateTime_now();


mlang-de:
I am not sure why Travis fails.
Some hints?

Error 1 on clang C:
  File "/home/travis/build/open62541/open62541/tools/c2rst.py", line 73, in <module>
    c = f.readlines()
  File "/usr/lib/python3.4/codecs.py", line 319, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfc in position 417: invalid start byte
make[3]: *** [doc_src/server_config.rst] Error 1
make[2]: *** [doc/CMakeFiles/doc.dir/all] Error 2
make[1]: *** [doc/CMakeFiles/doc.dir/rule] Error 2
make: *** [doc] Error 2

Error 2 on tcc C:
In file included from /home/travis/build/open62541/open62541/plugins/ua_accesscontrol_default.c:8:
In file included from /home/travis/build/open62541/open62541/plugins/ua_accesscontrol_default.h:11:
In file included from /home/travis/build/open62541/open62541/include/ua_server.h:24:
/home/travis/build/open62541/open62541/build/src_generated/ua_nodeids.h:8879: error: missing #endif
make[2]: *** [CMakeFiles/open62541-plugins.dir/plugins/ua_accesscontrol_default.c.o] Error 1
make[1]: *** [CMakeFiles/open62541-plugins.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[  4%] Built target open62541-generator-statuscode
INFO:__main__:Generating Code
INFO:__main__:NodeSet generation code successfully printed
[  4%] Built target open62541-generator-namespace
make: *** [all] Error 2

jpfr:
This looks like a very useful feature.
We have to discuss some of the ramifications, though.

**Server Config**
- Currently we assume that the server config does not change once the server has started.
- Currently, during server instantiation, we copy the config struct to the server struct (instead of just setting a pointer). So any changes to the server config at runtime will be lost to the eventual `UA_ServerConfig_delete` as the changes will not "leave the server".

Therefore, if we decide to make the configuration editable at runtime, then we have to make the server point to the external configuration. The indirection needs to be changed everywhere the config is accessed.

For consistency with the user API functions, please rename to `UA_Server_getConfig`.

**LateEndpointsCreate**
I'm not sure we need `lateEndpointsCreate`. We could simply remove the check for endpoints entirely.
This gives us the most flexibility and the default configuration factories will create endpoints anyway.

Can we split topical changes into a nice commit history?
Removing the endpoints check is largely separate from the `getConfig` stuff.
`git add -p` and `git rebase -i` are great helpers in keeping or restoring commit hygiene.

mlang-de:
**Server config**
Ok, I get you. Would be a bigger modification!

What's about adding a proper user function like:
The function is not reviewed only to show how it could work.
``` C  
UA_StatusCode
UA_Server_addNetworkLayer(UA_Server* server, UA_ServerNetworkLayer* networkLayers, 
                          UA_UInt16* portNumber, size_t networkLayersSize)
{
     UA_ServerConfig* conf = &server->config;
     conf->networkLayers = (UA_ServerNetworkLayer*)
         UA_realloc(conf->networkLayers, sizeof(UA_ServerNetworkLayer));
     if(!conf->networkLayers)

     for(size_t i = conf->networkLayersSize; i < (conf->networkLayersSize + networkLayersSize); i++)
     {
          conf->networkLayers[i] = UA_ServerNetworkLayerTCP(UA_ConnectionConfig_default,
                                                         portNumber[i - conf->networkLayersSize], 
                                                         conf->logger);
     }

     conf->networkLayersSize += networkLayersSize;

     return UA_STATUSCODE_GOOD;
}
```

I am not "happy" to change the closed config idea based on this one feature. Maybe your have further feature which requires an open config?!
One more option could be to enable/disable the bootstrapNS0 variable for adding nodes from an external source. But anyway this could be handled over an user function to and protect the server config.

> Can we split topical changes into a nice commit history?
Removing the endpoints check is largely separate from the getConfig stuff.
git add -p and git rebase -i are great helpers in keeping or restoring commit hygiene.

If we have decided how we will handle this feature I will try my best. I am a GitHub newbie :)

mlang-de:
**LateEndpointsCreate**

> I'm not sure we need lateEndpointsCreate. We could simply remove the check for endpoints entirely. This gives us the most flexibility and the default configuration factories will create endpoints anyway.

I'm against removing completly. Maybe change it to a logger warning or info without a bad statuscode in function UA_Server_run_startup( )?

``` C
if(config->endpointsSize == 0) {
    UA_LOG_WARNING/INFO(config->logger, UA_LOGCATEGORY_SERVER,
                 "There has to be at least one endpoint."); // "There is no endpoint configure. No communication possible!"
}
```

jpfr:
> I'm against removing completly. Maybe change it to a logger warning or info without a bad statuscode in function UA_Server_run_startup( )?

👍

mlang-de:
@jpfr What is the preferred way?

- move the endpoint check to UA_Server_run_startup without throwing an error. (already confirmed)
- add a user function to add entpoints similar to the proposal, probably only single endpoint adding. (Or the GetConfig-solution) (pending)

If not the GetConfig-solution, both in this pull request, because UA_Server_GetConfig function will be obsolete?!

jpfr:
> Move the endpoint check to UA_Server_run_startup without throwing an error. (already confirmed)

Yes. Let's only throw a log warning.

> Add a user function to add entpoints similar to the proposal, probably only single endpoint adding. (Or the GetConfig-solution) (pending)

Late addition of endpoints is a valid use case. We should support it.
We can add `UA_Server_getConfig` right now as well. (Note the caml-casing).

Can we make the two changes in two separate commits?

After the release of 0.3 we can have the server point to an external config instead of copying the struct into the server. I would the latter after the 0.3 release.
Otherwise it will get difficult to backport fixes to the 0.3 branch and vice versa.

mlang-de:
I tried the "git rebase" Feature, but obviously not correct...
Should I delete this PL and create a new clean one?

mlang-de:
In [7c67f6f](https://github.com/open62541/open62541/pull/1846/commits/7c67f6f388ba97bc10cf7a21a360241b01d6d5e0) are only the required changes for this PullRequest in two files. 
@jpfr Sorry for the big commit chaos!

mlang-de:
Rebase branch/Pull request onto master.

mlang-de:
@jpfr appveyor is pending since several days. Is it possible to retrigger it?

jpfr:
Already did.

jpfr:
Merged. :+1:

PR #1849, opened 2018-06-14
WIP: Historical Access Plugin definition

jpfr:
This PR proposes a "complete" plugin definition for Historical Access.
Note that historical data is not directly tied to nodes.
This keeps the node definition simple and the implementation complexity depends on the plugin provider.

@peter-rustler-basyskom 
For discussion in relation with #1740.
The implementation from #1740 would implement a "middle layer" that provides a minimal Historical Access implementation with an exchangeable "low level" API where people can plug in different storage backends.

### DIFF
--- a/include/ua_plugin_historical_access.h
+++ b/include/ua_plugin_historical_access.h
@@ -1,0 +1,149 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ *
+ *    Copyright 2018 (c) basysKom GmbH <opensource@basyskom.com> (Author: Peter Rustler)
+ *    Copyright 2018 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
+ */
+
+#ifndef UA_PLUGIN_HISTORICAL_ACCESS_H_
+#define UA_PLUGIN_HISTORICAL_ACCESS_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "ua_server.h"
+
+/**
+ * .. _historical-access:
+ *
+ * Historical Access Plugin API
+ * ============================
+ *
+ * Storage and access to historical data is implemented in a user-defined
+ * plugin. The plugin API does not mandate a specific kind of storage backend or
+ * handling of continuation points.
+ *
+ * The plugin is set globally in the server and used by the HistoryRead and
+ * HistoryUpdate services.
+ *
+ * The function pointers in the HistoryAccess plugin can be set to NULL. The
+ * server returns ``UA_STATUSCODE_BADNOTIMPLEMENTED`` if a requested method is
+ * not available.
+ */
+
+typedef struct {
+    void *context;
+    void (*deleteMembers)(void *context);
+
+    /* Is the node historizing? The callback is used by the normal Read service
+     * for the historizing attribute. */
+    UA_Boolean (*isHistorizing)(UA_Server *server, void *haContext,
+                                const UA_NodeId *nodeId);
+
+    /***************/
+    /* HistoryRead */
+    /***************/
+
+    UA_StatusCode
+    (*historyRead_raw)(UA_Server *server, void *haContext,
+                       const UA_NodeId *sessionId, void *sessionContext,
+                       const UA_ReadRawModifiedDetails *details,
+                       const UA_HistoryReadValueId *id,
+                       UA_TimestampsToReturn timestampsToReturn,
+                       UA_Boolean releaseContinuationPoints,
+                       UA_ByteString *outContinuationPoint,
+                       UA_HistoryData *outHistoryData);
+
+    /* UA_StatusCode */
+    /* (*historyRead_modified)(UA_Server *server, void *haContext, */
+    /*                         const UA_NodeId *sessionId, void *sessionContext, */
+    /*                         const UA_ReadRawModifiedDetails *details, */
+    /*                         const UA_HistoryReadValueId *id, */
+    /*                         UA_TimestampsToReturn timestampsToReturn, */
+    /*                         UA_Boolean releaseContinuationPoints, */
+    /*                         UA_ByteString *outContinuationPoint, */
+    /*                         UA_HistoryModifiedData *outHistoryData); */
+
+    /* UA_StatusCode */
+    /* (*historyRead_processed)(UA_Server *server, void *haContext, */
+    /*                          const UA_NodeId *sessionId, void *sessionContext, */
+    /*                          const UA_ReadProcessedDetails *details, */
+    /*                          const UA_HistoryReadValueId *id, */
+    /*                          UA_TimestampsToReturn timestampsToReturn, */
+    /*                          UA_Boolean releaseContinuationPoints, */
+    /*                          UA_ByteString *outContinuationPoint, */
+    /*                          UA_HistoryData *outHistoryData); */
+
+    /* UA_StatusCode */
+    /* (*historyRead_atTime)(UA_Server *server, void *haContext, */
+    /*                       const UA_NodeId *sessionId, void *sessionContext, */
+    /*                       const UA_ReadAtTimeDetails *details, */
+    /*                       const UA_HistoryReadValueId *id, */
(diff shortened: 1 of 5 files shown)


include/ua_plugin_historical_access.h
```diff
@@ -0,0 +1,148 @@
...
+    /* Is the node historizing? */
+    UA_Boolean (*isHistorizing)(UA_Server *server, void *haContext,
+                                const UA_NodeId *nodeId);
+
+    /***************/
+    /* HistoryRead */
+    /***************/
+
+    UA_StatusCode
+    (*historyRead_raw)(UA_Server *server, void *haContext,
```
peter-rustler-basyskom: haContext is not needed, because it is accessable from the server->config object.

include/ua_plugin_historical_access.h
```diff
@@ -0,0 +1,148 @@
...
+ * HistoryUpdate services.
+ *
+ * The function pointers in the HistoryAccess plugin can be set to NULL. The
+ * server returns ``UA_STATUSCODE_BADNOTIMPLEMENTED`` if a requested method is
+ * not available.
+ */
+
+typedef struct {
+    void *context;
+    void (*deleteMembers)(void *context);
```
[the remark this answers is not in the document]
jpfr: The other plugins follow the same schema.
The plugin is then destroyed as part of the server configuration.

include/ua_plugin_historical_access.h
```diff
@@ -0,0 +1,148 @@
...
+
+    /* Is the node historizing? */
+    UA_Boolean (*isHistorizing)(UA_Server *server, void *haContext,
+                                const UA_NodeId *nodeId);
```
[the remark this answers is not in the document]
jpfr: Variable nodes have a `historizing` attribute.
The normal read service would query the attribute value in the HistoricalAccess plugin.

include/ua_plugin_historical_access.h
```diff
@@ -0,0 +1,148 @@
...
+
+    /***************/
+    /* HistoryRead */
+    /***************/
+
+    UA_StatusCode
+    (*historyRead_raw)(UA_Server *server, void *haContext,
+                       const UA_NodeId *sessionId, void *sessionContext,
```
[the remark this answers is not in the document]
jpfr: The session is opaque for the end user.
But he may attach a context value to the session during authentication.

include/ua_plugin_historical_access.h
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * Storage and access to historical data is implemented in a user-defined
+ * plugin. The plugin API does not mandate a specific kind of storage backend or
+ * handling of continuation points.
+ *
+ * The plugin is set globally in the server and used by the HistoryRead and
+ * HistoryUpdate services.
+ *
+ * The function pointers in the HistoryAccess plugin can be set to NULL. The
+ * server returns ``UA_STATUSCODE_BADNOTIMPLEMENTED`` if a requested method is
```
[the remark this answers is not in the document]
jpfr: Perfect.

include/ua_plugin_historical_access.h
```diff
@@ -0,0 +1,148 @@
...
+
+    UA_StatusCode
+    (*historyRead_raw)(UA_Server *server, void *haContext,
+                       const UA_NodeId *sessionId, void *sessionContext,
+                       const UA_ReadRawModifiedDetails *details,
+                       const UA_HistoryReadValueId *id,
+                       UA_TimestampsToReturn timestampsToReturn,
+                       UA_Boolean releaseContinuationPoints,
+                       UA_ByteString *outContinuationPoint,
+                       UA_HistoryData *outHistoryData);
```
peter-rustler-basyskom: What if we need to change other data from the response.
Or if we do not need to provide UA_HistoryData to the client.
In error condition for example.

include/ua_plugin_historical_access.h
jpfr: The server object is intentionally opaque for the end user.
The above is possible only when using internal headers.

include/ua_plugin_historical_access.h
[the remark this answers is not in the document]
peter-rustler-basyskom: Where do the HistoricalAccess plugin have that information?
And who is calling this function pointer? There is no user from this outside the plugin itself ?

include/ua_plugin_historical_access.h
jpfr: True. We could simply give a HistoryReadResult into the callback.
Then the plugin would allocate a member inside the `historyData` ExtensionObject.

The advantage of spelling this out is a type-stable API where the plugin knows whether a `UA_HistoryData` or a `UA_HistoryModifiedData` is expected.

include/ua_plugin_historical_access.h
```diff
@@ -0,0 +1,149 @@
...
+
+    /***************/
+    /* HistoryRead */
+    /***************/
+
+    UA_StatusCode
+    (*historyRead_raw)(UA_Server *server, void *haContext,
+                       const UA_NodeId *sessionId, void *sessionContext,
+                       const UA_ReadRawModifiedDetails *details,
+                       const UA_HistoryReadValueId *id,
```
Root-Core: The request can consist of several nodes, as "this parameter contains the list of items upon which the historical retrieval is to be performed" (OPC UA Spec Part 4). How to determine, how many items are requested here?

jpfr: The demuxing would be done by the server before dropping into the plugin.

See the definition of the data types coming in.

Root-Core: This could cause problems with some databases. Time critical, high performance or fail proof hardware often writes data as soon as possible to the storage, to have recent data in place after a fail occurred.

As they might write the data of multiple channels to the same files, you would need to iterate over the same files / time ranges for every channel, which destroys performance. You could collect the data of all channels in one go, if you know what to look for.

Root-Core:
We have to consider some things: 

1. The plugin call might be sync and accessing the data might take a while
1.a. The spec tells us to send an answer in the given time, regarding the clients timeout hint
1.b. This could be an answer without values, but with given continuation points
1.c. The plugin will likely not be able to handle this itself, as it is not its purpose

2. If we split up the raw and modified functions - which totally makes sense - we should not give the full ReadRawModifiedDetails
2.a. When a continuation point is set, we furthermore shall ignore the details (and TimeStampsToReturn)
2.b. This means, that the important parts of the details need to be stored along the CP
2.c. (Sure, the client should not change the details between requests in first place.. but we can not control the clients behavior)

3. The plugin read methods should not handle the release of CPs
3.a. Decoupling the release of CPs from the rest of the API might help to keep the API clean
3.b. This is also valid for creating CPs in the first place
3.c. This claim supports the idea to implement a CP plugin

4. It could be a good idea to have a quiet open API, that gets stripped down in the example plugin
4.a. For example, we might implement demuxing in the plugin, rather than stack sided
4.b. The final "read" methods (after demuxing) could and should be as specific as possible

Just a couple of thoughts. However, not all points might be valid.

jpfr:
Hmm. Ok.
Those are the right kind of questions.

Would such a plugin be reusable on its own?
I fear that we always require a hard coupling between the cp mechanism and the historical data implementation.
So that we may keep them together just as well.

Root-Core:
The historical access plugin could instantiate the CP plugin.
The CP plugin could hold a pointer per CP to a struct, defined by the HA plugin.
This struct could contain the information, that belongs to the CP. In this case, options that should be ignored while using a CP. Also it could hold a handle to timer for handling possible client timeouts. Not sure about that.

We could reuse the CP plugin for browsing, as browsing does support CPs as well. I've not looked further into details about this though. I suppose we also need to implement some kind of struct to hold information there.

------------------------

The most difficult thing to me is, how to handle the client timeout hint.
I will try to explain, what's in my head right now.. i'm not sure to accomplish this.

One solution could be to send responses without values (but including a CP) if the plugin takes its time. We may collect / calculate data in the background until the client requests with the CP and send empty responses (again if not finished with the first value) / partial responses (if data is buffered yet) with a CP / full responses (if everything is done) without a CP.
This could be coupled with a timer that sends empty / partial responses, if we need to as the timer runs out.

It could get difficult to have the plugin as "async" as possible if the server isn't, but i think it is doable.

jpfr:
Superseded by #1915

PR #1851, opened 2018-06-14
Fix gcc compiler warning: no previous prototype for 'UA_dump_hex_pkg'

mlang-de:
Add #ifdef condition


jpfr:
Thanks for the contribution!

PR #1860, opened 2018-06-20
Bad max check when creating a new subscription

jonasgreen88:
During the creation of a new subscription the number of active
subscriptions are compared to the maximum number of subscriptions
allowed to be created. However, this check validated against
number of received publish requests instead of number of
active subscriptions.


jonasgreen88:
This fix should probably be added to the 0.3 rc-2 as well.

jpfr:
Thanks!

We will not make changes to a release candidate. But this will be cherry-picked to the 0.3?branch.

PR #1862, opened 2018-06-20
Added HistoryServerCapabilities to ServerCapabilities

Root-Core:
OPC Spec Part 11 - 5.4.2: 
> The ServerCapabilitiesType Objects for any OPC UA Server supporting Historical Access shall contain a Reference to a HistoryServerCapabilitiesType Object.

I made the capabilities configurable via the server config file. This could also help the regarding implementation, as we could quickly check, if the implementation supports the requested feature.
It only gets applied, if `UA_ENABLE_HISTORIZING` is enabled.

The problem: The `HistoryServerCapabilities` are added to `ServerCapabilities` regardless if `UA_ENABLE_HISTORIZING` is set. I don't know how one would get the node compiler to do this dynamically, without creating a new NodeSet containing Minimal and Part 11.

Also adding `HistoricalRawData` Profile, if `UA_ENABLE_HISTORIZING` is enabled.
This could get a bit messy in the future, as the Historical Profiles are split into features.
It could be better to get some dynamic into the "profile system", regarding to the servers configuration.
This could easily done by checking the server config and setting the profile array late in `UA_Server_initNS0()`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,7 @@ endif()
 
 # Options
 set(UA_LOGLEVEL 300 CACHE STRING "Level at which logs shall be reported")
+option(UA_ENABLE_HISTORIZING "Enable server to provide historical access." ON)
 option(UA_ENABLE_METHODCALLS "Enable the Method service set" ON)
 option(UA_ENABLE_NODEMANAGEMENT "Enable dynamic addition and removal of nodes at runtime" ON)
 option(UA_ENABLE_SUBSCRIPTIONS "Enable subscriptions support." ON)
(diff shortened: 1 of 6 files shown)


jpfr:
Thanks. Looks good. 😄 

@peter-rustler-basyskom This PR is pretty much independent of the discussion for the historizing plugin.
It is merged ahead of time to reduce the "surface" of the actual historizing implementation.

Root-Core:
@jpfr I digged a bit around and it seems, that @Pro wants to keep the order of `Opc.Ua.NodeSet2.Minimal.xml` regarding `Opc.Ua.NodeSet2.xml`. Should I reorder the additional items?

@jpfr Yes, the goal of the PR was to decrease the complexity of the "real" implementation. It will make things easier, if such "features" are split into smaller independent chunks. :+1:

Pro:
> @jpfr I digged a bit around and it seems, that @Pro wants to keep the order of Opc.Ua.NodeSet2.Minimal.xml regarding Opc.Ua.NodeSet2.xml. Should I reorder the additional items?

Correct. The `Opc.Ua.NodeSet2.Minimal.xml`  should only differ by deletions from the full NodeSet2.xml. There shouldn't be any other changes like additions or reorderings. This will make it easier to include new features/releases in the future.

PR #1865, opened 2018-06-21
Create self-signed ApplicationInstanceCertificate

ckmk14:
At the moment the "UA_BUILD_SELFSIGNED_CERTIFICATE" option is not creating a self-signed certificate. Therefore there are trusting issues with some tools like UAExpert.


jpfr:
Thanks.
Can you remove the commenter sections?

If required we have that in the repository.

PR #1877, opened 2018-06-26
Remove all usages of UA_THREAD_LOCAL

### LINKED ISSUE #1769: UA_THREAD_LOCAL still used in variable
On branch 0.3, in commit d783f983c7a7db9bbd7ef52a48732d3aa40ebaf0 all uses of UA_THREAD_LOCAL were removed,  but there's still one in https://github.com/open62541/open62541/blob/0.3/src/ua_types.c#L74.

I noticed this because in commit 432cc0c85e0e5cee48aa0936de4e56967a12a3c0, the configuration of VxWorks (WRS_KERNEL) was changed related to UA_THREAD_LOCAL .

If UA_THREAD_LOCAL is meant to stay, then I'd like to add again the DEFINE that made VxWorks define UA_THREAD_LOCAL  as empty as it was before.

Pro:
@jpfr do you see a better way to handle `static pcg32_random_t UA_rng = PCG32_INITIALIZER;` not being threadsafe?

Fixes #1769

### DIFF
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -33,9 +33,9 @@ const UA_ByteString
     UA_SECURITY_POLICY_NONE_URI = {47, (UA_Byte *)"http://opcfoundation.org/UA/SecurityPolicy#None"};
 
 #ifdef UA_ENABLE_UNIT_TEST_FAILURE_HOOKS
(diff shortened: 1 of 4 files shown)


jpfr:
I'm okay with that.
Even race conditions don't change the non-security-grade randomness nature of the thing.

PR #1878, opened 2018-06-26
Formatting

Infinity95:
Some source files are formatted differently than others and about, i guess, 50% of the code uses the same style.
I would suggest changing the style file by file so we have a uniform coding style.
As an example i formatted the client_async file according to the style that is used in most other source files.

What do you think of this?

### DIFF
--- a/examples/client_async.c
+++ b/examples/client_async.c
@@ -25,35 +25,36 @@
 #endif
 
 #define NODES_EXIST
+
 /* async connection callback, it only gets called after the completion of the whole
  * connection process*/
 static void
-onConnect (UA_Client *client, void *userdata, UA_UInt32 requestId,
-           void *status) {
-    printf ("Async connect returned with status code %s\n",
-            UA_StatusCode_name (*(UA_StatusCode *) status));
+onConnect(UA_Client *client, void *userdata, UA_UInt32 requestId,
+          void *status) {
+    printf("Async connect returned with status code %s\n",
+           UA_StatusCode_name(*(UA_StatusCode *)status));
 }
 
 static
 void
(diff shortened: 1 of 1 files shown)


jpfr:
I would not rewrite 50% of the code to adjust styling.
A uniform style is a good thing. And it is easy enough to be changed.
But it prevents us from backporting patches from master to 0.3.

Going forward, we could try to formalize our coding style and enforce it for future commits.

Infinity95:
I will write a style guide (or update the current existing one) so we can use that for future pull requests.
Concerning the back-porting of patches: Wouldn't it be possible to apply the formatting to both master and 0.3? That way any patches that come after the reformat shouldn't cause any major issues, right?
This also doesn't have to be done in one huge pull request but my idea was rather to do it file by file to gradually introduce the formatting changes.

Pro:
I strongly vote against commits which only cover code style changes, especially of a huge code part. This always breaks a lot of other open PRs and it also may introduce merge issues with. Eg. the architecture PR (on which I will work the next days) is one of these which will cause many merge conflicts if we reformat code somewhere else. Also  forks (like the one from @janitza-thbe) will have major issues with that. We had that issue quite often in the past.

Code style reformatting is definitely a good thing, but please only on code which is changed anyways in a commit.

Infinity95:
You're right.
I will like i said try to make a style guide proposal that we should use for future commits.

HolgerJeromin:
In theory this should be no problem as we have an .editorconfig file in the repo:
https://github.com/open62541/open62541/blob/e4e3f1e6ebcfdb18637324cd331d261a29733a74/.editorconfig

PR #1879, opened 2018-06-26
Breaking API: Forward endpoint information and the remote certificate to AccessControl

jpfr:
The PR is a reaction to https://groups.google.com/forum/#!topic/open62541/8qujwG8fTC4

The authentication plugin needs to know some information about the SecureChannel.
For example to allow additional access only over an encrypted channel.

With this change we forward the endpoint information and the remote certificate used for the SecureChannel to the AccessControl plugin.


jpfr:
This is changing a public API.
Albeit only for the AccessControl plugin.

Still, is this okay for @open62541/open62541-core ?

PR #1884, opened 2018-06-29
Add '#ifdef __cplusplus extern "C" {...' in ua_client_highlevel_async.h

ccvca:
This fixes  `undefined reference to` when using async functions from C++.


jpfr:
Thanks!

Issue #1861, opened 2018-06-20
DataSubscription issue when the client doesn't provide enough publish requests before publishing


jonasgreen88:
Description
===========
I have found what I belive is an issue in the DataSubscription implementation. I would like to let you evaluate the logs to see what you think. 

I did a setup where a client sets up one subscription with 2 monitored items monitoring different variable nodes. The data source of the variables changes frequently, faster than the configured publishing/sampling intervals.  The queue size of the monitored items are 1. The client does only have 1 outstanding publish request and allows only 1 notification entry per publish request. The publishing interval of the subscription equals the sampling interval of the monitored items (5 seconds in the logs).

When the monitored items detect value changes (in every sample in this case) the server can only produce 1 of the changes and then have to wait for a second publish request to publish the next one. This is where the issue shows up. When the server have responded with the notification of the first monitored item and then receives a new publish request from the client, it doesn't always publish the notification of the second monitored item. See wireshark log "log-monItemQueueSize_1-publishRequests_1-maxNotifPerPubRsp_1.pcapng". The notification of the second monitored items is only published every second publish interval.

The following configuration changes solve/hide the issue:
- Set the monitored item queue size to 2 (log-monItemQueueSize_2-publishRequests_1-maxNotifPerPubRsp_1.pcapng).
- Configure the client to initially provide 2 publish requests (log-monItemQueueSize_1-publishRequests_2-maxNotifPerPubRsp_1.pcapng).
- Allow 2 notifications in one publish request (log-monItemQueueSize_1-publishRequests_2-maxNotifPerPubRsp_1.pcapng).
What is common with all these changes that solve/hide the issue are that the server have enough resource to publish all changes immediately when they occurs, it doesn't have to wait for a new publish requests from the client.

What do you guys think? I guess it is a valid configuration that should work?

Attached logs: [wireshark-logs.zip](https://github.com/open62541/open62541/files/2119138/wireshark-logs.zip)

Background Information / Reproduction Steps
===========================================

Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): d88df4281dde5744db06d6c738d0fcc6b73523e8
 - [x] Other OPC UA SDKs used (client or server): Client lib: https://github.com/FreeOpcUa/python-opcua
 - [x] Operating system: Embedded OS / Windows 7
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [x] Wireshark network dump attached
 - [ ] Self-contained code example attached
[...]

PR #1885, opened 2018-06-29
Update timer implementation to not change order of callbacks

jonasgreen88:
If there are a set of timer callbacks in the timer list that are
executed in the same timer process cycle the order of them were
inversed every time after they had been executed, except the first
callback. Also, if a set of timer callbacks with same timer interval
are added within 1 second of each other, they are synchronized to
expire at the same time and added in front of each other, except the
first one which will keep the first position. If four timer callbacks
with the same time interval are added after each other they are put
in the timer list in the following order: #1, #4, #3, #2.

This change fixes so callback timers that will be executed with
the same time interval and that are put together in a "block" in the
timer list are not inversed within this block after each execution of
them. It also fixes so all timers within a "block" are added in a
deterministic order. As described above they where added in reversed
order to the timer list except the first one. Now all timers that can
be put together in a "block" are added in reversed order within this
block. One advantage of putting the callback timers in reversed order
is if a subscription with a set of monitored items are created with
the same publish/sampling interval. In that case the monitored items
that are created after the subscription has been created will be
executed by the timer process before the subscription's publish
function is called. This execution order will cause minimum delay
of data notifications to be published.

Solves issue #1861.

### DIFF
--- a/src/ua_timer.c
+++ b/src/ua_timer.c
@@ -128,7 +128,6 @@ addTimerCallbackEntry(UA_Timer *t, UA_TimerCallbackEntry * UA_RESTRICT tc) {
     SLIST_FOREACH(tmpTc, &t->repeatedCallbacks, next) {
         if(tmpTc->nextTime >= tc->nextTime)
             break;
-        afterTc = tmpTc;
 
         /* The goal is to have many repeated callbacks with the same repetition
          * interval in a "block" in order to reduce linear search for re-entry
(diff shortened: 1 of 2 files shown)


jpfr:
This is great!
The changes are subtle. But they improve the default behavior a lot.

A few cosmetic points.

- spurious space in line 138
- the else condition around line 142 is not needed
- two sentences on the ordering behavior in ua_timer.h would be good.
To document for future users and contributors.

The change itself is perfect and ready to merge.

jonasgreen88:
Thanks for the feedback. I have updated the PR according to those.

jpfr:
Perfect. 👍 
Thanks for the contribution!

PR #1889, opened 2018-07-02
Small fixes for minimal nodeset and NS0 init

Pro:
@jpfr do we need the ByteString for the Types.xsd (second commit below)?

Otherwise I would remove it from the minimal nodeset


jpfr:
I have a couple requests with respect to commit 020eca3dada9c84b2b84584fabb93ffb68f8a2f3.

1. The standard explicitly says in Part 3, §8.7 BaseDataType:

> This abstract DataType defines a value that can have any valid DataType.
> It defines a special value null indicating that a value is not present.

So we can only support empty values of the DataType of the Variable is BaseDataType.
Otherwise the model is faulty.

2. We can "separate concerns" for the change of `isNodeInTree`.
I agree that it is important to consider derived ReferenceTypes.

But `isNodeInTree` already supports arrays of considered ReferenceTypes.
I suggest that we rather modify `getTypeHierarchy` to return the hierarchy either upwards or downwards.
Then we can plug the hierarchy into the existing `isNodeInTree`.

3. Improving logging to include the NodeId is a good idea.
But please put that into a separate commit. Or even a separate PR.
Also the UA_LOG_NODEID_WRAP macro can be put into a place where the entire SDK can see it (internally).

It is really to split an existing commit into separate features.

`git reset --soft HEAD^`  (go back and edit the last commit)
`git reset` (take all changed files out of the staging area)
`git add -p <file>` (interactively stage only the parts of the file that is required for the feature-commit)
`git commit`

Pro:
Just a general note:
Somehow the return value of the NS0 init method was removed in previous commits and even if there was an error, CI did not complain. So many issues fixed by this PR were there before. E.g. also the following:

> So we can only support empty values for variables with the DataType BaseDataType (Variant).
> Otherwise the model is faulty.

Then we may have a bigger issue. The 1.04 NodeSet has some variables which are of subtypes of the BaseDataType. E.g. adding node `OldValues - ns=0;i=3024` the old code complains with:
```
Only Variables with data type BaseDataType may contain a null (empty) value
```

If you look at the XML:

```xml
  <UAVariable NodeId="i=3024" BrowseName="OldValues" ParentNodeId="i=3022" DataType="i=920">
    <DisplayName>OldValues</DisplayName>
    <References>
      <Reference ReferenceType="HasTypeDefinition">i=68</Reference>
      <Reference ReferenceType="HasProperty" IsForward="false">i=3022</Reference>
    </References>
  </UAVariable>

  <UADataType NodeId="i=920" BrowseName="HistoryEventFieldList">
    <DisplayName>HistoryEventFieldList</DisplayName>
    <References>
      <Reference ReferenceType="HasSubtype" IsForward="false">i=22</Reference>
    </References>
    <Definition Name="HistoryEventFieldList">
      <Field Name="EventFields" ValueRank="1" />
    </Definition>
  </UADataType>
  <UADataType NodeId="i=22" BrowseName="Structure" IsAbstract="true">
    <DisplayName>Structure</DisplayName>
    <Description>Describes a value that is any type of structure that can be described with a data encoding.</Description>
    <References>
      <Reference ReferenceType="HasSubtype" IsForward="false">i=24</Reference>
    </References>
  </UADataType>
  <UADataType NodeId="i=24" BrowseName="BaseDataType" IsAbstract="true">
    <DisplayName>BaseDataType</DisplayName>
    <Description>Describes a value that can have any valid DataType.</Description>
    <References />
  </UADataType>

  <UAVariableType NodeId="i=68" BrowseName="PropertyType" ValueRank="-2">
    <DisplayName>PropertyType</DisplayName>
    <Description>The type for variable that represents a property of another node.</Description>
    <References>
      <Reference ReferenceType="HasSubtype" IsForward="false">i=62</Reference>
    </References>
  </UAVariableType>
```

This shows that `OldValue` is of DataType `HistoryEventFieldList` and not directly a `BaseDataType`.
Are null values allowed for subtypes of BaseDataType? Because than we can simply do a `isNodeInThree` check.

Regarding 2. and 3. I will split the commits. This PR grew bit by bit that's why there is a bit of mess in here.

jpfr:
> Then we may have a bigger issue. The 1.04 NodeSet has some variables which are of subtypes of the BaseDataType.

It seems like a lot of existing information models have trouble with the rule from Part 3, §8.7 BaseDataType.

How about we allow null values everywhere but print a WARNING log message if the type is not a BaseDataType?

PR #1890, opened 2018-07-02
Add UA_NodeId_toString method

Pro:
This method will be used for improving logging, e.g. if specific node id's are not found, the node ID is printed to log.


jpfr:
Cool! Just some minor comments:

- Name should be `UA_NodeId_toString`
- Can we return a statuscode if the malloc for the string fails?
- The XML-files use a specific format for NodeIds. For example `ns=0;i=1`. Can we use the same format?
- `check_types_custom.c` is not the place for the tests imho. We use it to test user-defined datatypes.
Maybe the normal types-checks or logging-checks are a better place.

include/ua_plugin_log.h
```diff
@@ -118,6 +121,90 @@ UA_LOG_FATAL(UA_Logger logger, UA_LogCategory category, const char *msg, ...) {
...
+            UA_snprintf((char*)nodeIdStr->data, nodeIdStr->length, "ns=%d;i=%lu",
+                        nodeId->namespaceIndex, (unsigned long )nodeId->identifier.numeric);
+            break;
+        case UA_NODEIDTYPE_STRING:
+            /* ns (16bit) = 5 chars, strlen + nullbyte */
+            nodeIdStr->length = 11 + nodeId->identifier.string.length + 1;
+            nodeIdStr->data = (UA_Byte*)UA_malloc(nodeIdStr->length);
+            if (nodeIdStr->data == NULL)
+                return UA_STATUSCODE_BADOUTOFMEMORY;
+            UA_snprintf((char*)nodeIdStr->data, nodeIdStr->length, "ns=%d;i=%.*s", nodeId->namespaceIndex,
```
basyskom-jvoe: The identifier type is specified as i= for numeric, s= for string, g= for GUID and b= for bytestring  (spec, part 6 page 22).

kjeldflarup:
A'm I right that this is not a part of 0.3?

Pro:
Yes, you are right. At the time this PR was created we already had a feature freeze on the 0.3 branch. This will be included in the 0.4 release

PR #1892, opened 2018-07-03
Bugfix memory leak in local monitored item

peter-rustler-basyskom:
If you run the test without the fix and enabled memory sanitizer, you will see the memory leak in output.
With the fix it is gone.

Note: Enable the sanitizer with the following lines in the CmakeList.txt file.
```
set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-omit-frame-pointer -fsanitize=address")
set (CMAKE_C_LINKER_FLAGS "${CMAKE_C_LINKER_FLAGS} -fno-omit-frame-pointer -fsanitize=address")

```

### DIFF
--- a/src/server/ua_subscription_datachange.c
+++ b/src/server/ua_subscription_datachange.c
@@ -431,6 +431,25 @@ sampleCallbackWithValue(UA_Server *server, UA_MonitoredItem *monitoredItem,
                                               value);
     }
 
+    // If someone called UA_Server_deleteMonitoredItem in the user callback,
+    // then the monitored item will be deleted soon. So, there is no need to
+    // add the lastValue or lastSampledValue to it.
+    //
+    // If we do so, we will leak
+    // the memory of that values, because UA_Server_deleteMonitoredItem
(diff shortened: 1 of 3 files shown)


jpfr:
Hello @peter-rustler-basyskom ,

thanks for the contribution. 👍 
This is a very good and complete PR.
It's good to have a unit test in place to prevent future regressions.

I would like to commit the PR.
But even more so, I would like to understand why this fix is actually working. 😄 

I am rather surprised to see this solution working.
Without the fix, `binaryEncoding` is moved to `the monitoredItem->lastSampledValue`.
From there it will be freed up in the next sample callback or in `UA_MonitoredItem_delete`.

So how can we even lose the data? Do we forget a `UA_MonitoredItem_delete` somewhere?

The valgrind tool is similar to the sanitizers prints the stack-trace of the data loss origin.
`valgrind --leak-check=full --track-origins=yes ./server`

What does this output without the fix in place?

jpfr:
👍

PR #1896, opened 2018-07-05
Compile for OS-9 PowerPC and ARMv7 with CLANG.

Kei-Thomsen:
I have modified 3 files to be able to compile for OS-9 for PowerPC and ARMv7. OS-9 is a real-time-operating system. The project has been started with the new CLANG compiler.
Therefore I compiled the project for Linux with the CLANG first and found out, that a definition for __FLOAT_WORD_ORDER is missing in ua_config.h(.in) . This has been fixed in ua_config.h.in , together with an additional define UA_BINARY_OVERLAYABLE_FLOAT for OS-9 (_OS9000).

CMakeLists.txt contains the needed Libraries for OS-9 and it must not link with -s

plugins/ua_network_tcp.c is doing a nonblocking connect. As the OS-9 select can't check for sockets to be writable, it has to poll the connection with connect() and check the error codes for success or fail.

I don't think that anyone (except me) will be able to compile this project for OS-9 at the moment, as the CLANG is still in alpha test. I'm not getting any Warning during compilation of my changes in ua_network_tcp.c

It will be great, if my OS-9 work will make it into the official version.

Kei Thomsen


jpfr:
Thanks for the contribution!
Good to see another os in our portfolio. :smile:

We have users targeting strange hardware.
Big-endian, mixed-endian, non-IEEE754 floating point, and so on.

The clang definitions in config.h.in look like they are valid for a range of target architectures. Some of them might require special handling.
Why don’t you adjust your clang to define the discovery macros?
Or put everything within an os9 macro...

If we could not detect little endian IEEE754 fp, a slower default encoding routine is used. That works on every architecture.

jpfr:
>  -- But both are not strange! Best Hardware ever ;-)

Full Ack.

The `UA_BINARY_OVERLAYABLE_FLOAT` defines whethe we can memcpy floats and doubles on the network buffer. Otherwise we need to manually convert to IEEE754 first.
The endianness is handled by `UA_BINARY_OVERLAYABLE_INTEGER`.

The following addition could lead to IEEE754 fp being assumed when this is not the case.
Because __BYTE_ORDER refers to the integer endianness, not IEEE754.

> +# if defined(__clang__) && !defined(__FLOAT_WORD_ORDER)
> +#  define __FLOAT_WORD_ORDER __BYTE_ORDER
> +# endif

Kei-Thomsen:
I did this, because the CLANG (for Linux, not for OS-9) is not defining the __FLOAT_WORD_ORDER by default. To be able to compile, my idea is, that if it is compiled with CLANG and the __FLOAT_WORD_ORDER is not set, we can assume that the float order is the same like the byte order. 

Otherwise the following 
# elif __FLOAT_WORD_ORDER == __LITTLE_ENDIAN
#  define UA_BINARY_OVERLAYABLE_FLOAT 1
# endif
will not compile, as __FLOAT_WORD_ORDER is nowhere defined within CLANG with musl.

It might be changed to e.g.
# elif defined(__FLOAT_WORD_ORDER) && (__FLOAT_WORD_ORDER == __LITTLE_ENDIAN)
#  define UA_BINARY_OVERLAYABLE_FLOAT 1
# endif

Then it will compile without UA_BINARY_OVERLAYABLE_FLOAT , but at least it will compile.

Kei-Thomsen:
Now we are clear with the float endian stuff. Thank you!
It has been tested on PowerPC (Big-Endian CPU) and ARMv7 (Little-Endian CPU).
The actual PR contains the OS-9 modifications based on the master branch.

Kei-Thomsen:
The AppVeyor failed:

2>downloading http://mirrors.rit.edu/CTAN/systems/win32/miktex/tm/packages/needspace.tar.lzma...
2>Unfortunately, the package needspace could not be installed.Please check the log file:
2>C:/miktex/texmfs/data/miktex/log/pdflatex.log
2>======================================================================
2>Project : error PRJ0019: A tool returned an error code from "Generating PDF documentation from LaTeX sources"
2>Build log was saved at "file://c:\projects\open62541\build\doc\doc_pdf.dir\Debug\BuildLog.htm"
2>doc_pdf - 1 error(s), 1 warning(s)
========== Build: 0 succeeded, 1 failed, 9 up-to-date, 1 skipped ==========
*** Make doc_pdf. Exiting ... ***
Command exited with code 1

Don't think it has to do with my PR ?!?

jpfr:
Great! Merged.

Kei-Thomsen:
Great, many thank for your help!

PR #1904, opened 2018-07-06
Fix response order for client-/serverhandle in GetMonitoredItems method

mlang-de:
CTT script "Address Space Model/Address Space Method/001.js" fails.
The open62541 Server returns in the response the Output arguments switched.
OutputArgument[0] must contain serverHandle and OutputArgument[1] must contain clientHandle.


jpfr:
Thanks for the fix!

PR #1908, opened 2018-07-09
Fix method get monitored item error handling for CTT Method Call script 004

mlang-de:
CTT Method Call script 004 throws error for GetMonitoredItem method:
- Call.Response.Results[0].OutputArguments[0] is empty.
- Call.Response.Results[0].OutputArguments[1] is empty.


jpfr:
Thanks!

PR #1915, opened 2018-07-11
Feature/service plugin api historical read interface only

peter-rustler-basyskom:


include/ua_plugin_history_data_service.h
```diff
@@ -0,0 +1,79 @@
...
+     */
+    void
+    (*readRaw)(void *context,
+               UA_Server *server,
+               const UA_NodeId *sessionId,
+               void *sessionContext,
+               const UA_HistoryReadRequest *request,
+               const UA_ReadRawModifiedDetails *details,
+               UA_HistoryReadResponse *response,
+               UA_HistoryReadResult* results);
```
Root-Core: The `results` parameter is a member of `response` (as you correctly state) and therefore could be omitted as a additional parameter. It might confuse while implementing the plugin function.

src/server/ua_services_attribute.c
```diff
@@ -1402,6 +1415,51 @@ __UA_Server_write(UA_Server *server, const UA_NodeId *nodeId,
...
+                response->results = (UA_HistoryReadResult*)UA_Array_new(response->resultsSize, &UA_TYPES[UA_TYPES_HISTORYREADRESULT]);
+                server->config.historyDataService.readRaw(server->config.historyDataService.context,
+                                                           server,
+                                                           session ? &session->sessionId : NULL,
+                                                           session ? session->sessionHandle : NULL,
+                                                           request,
+                                                           details,
+                                                           response,
+                                                           response->results);
+                return;
```
Root-Core: As the plugin functions kind of implement the service - if you want - the functions could return a operation level status code, which is used as service result. I'm not sure how important / useful this is.

Well, you already force it by accessing the `response` object, but it might be more convenient.

peter-rustler-basyskom: The return statuscode is in the response. I do not see the requirement for an extra statuscode. What to do with that statuscode ? how should we react if this statuscode is bad and in the response is a bad statuscode and the other way around. If we have a bad statuscode we still need to answer the client. I like a singe source of true...

include/ua_plugin_history_data_service.h
peter-rustler-basyskom: Yes, the results object is already in the response. But to make t easier for the programmer it is here, also. In the response it is an extrension object and do not have the type. To show the programmer that we expect a UA_HistoryReadResult array to be filled, it is here.

Root-Core: Fair enough. Totally forgot, that the response actually is an extension object. :+1:

src/server/ua_services_attribute.c
Root-Core: The idea was to take the returned value as `serviceResult`. Sure you can set it in the response itself. Just wanted to discuss this, as it came into my mind on the first time digging through the PR. It's not exactly a requirement. :)

include/ua_plugin_history_data_service.h
jpfr: Hmm. I see the point.

On the other hand, having the same "pointer" twice in the arguments can be very confusing for users.

peter-rustler-basyskom: It is not the same pointer twice.
You have one untyped (void*) pointer as a member in the response.
And you have the results pointer which has the proper type and allocated memory you just have to fill.

That makes it much harder for a programmer to return the wrong type in the response.

src/server/ua_services_attribute.c
```diff
@@ -1402,6 +1415,51 @@ __UA_Server_write(UA_Server *server, const UA_NodeId *nodeId,
...
+#ifdef UA_ENABLE_HISTORIZING
+
+#include "ua_plugin_history_data_service.h"
+
+void
+Service_HistoryRead(UA_Server *server,
+                    UA_Session *session,
+                    const UA_HistoryReadRequest *request,
+                    UA_HistoryReadResponse *response) {
+    UA_HistoryReadResponse_init(response);
```
Root-Core: This overrides the `requestHandle` and `timestamp` of the response, which areset earlier in `processMSG()`. This does not work for async clients, like Milo. Furthermore, it's already initialized in `processMSG()` via `UA_init()`.

peter-rustler-basyskom: Removed

include/ua_plugin_history_data_service.h
jpfr: I see three points with this API for readRaw

- The plugins becomes much more complicated. Because it has to cover memory allocations for the response message internally.
- We loose helpful type information in the API. The input always contains a UA_ReadRawModifiedDetails. But we need to cast that from an ExtensionObject. The output always is a UA_HistoryData. But we need to manually set that into an ExtensionObject. The API should prevent all this casting business inside the plugin, which is always a source of confusion and errors.
- `details` and `results` is contained in the request and response. I would prefer we do not "repeat ourselves"  in the API.

Why don't we take the API for historyRead_raw suggested in #1849?
https://github.com/open62541/open62541/blob/eaaac51e6523aef6af618b23c63c1410aceaada7/include/ua_plugin_historical_access.h#L49

That was a very serious proposal.
Please come forward with technical arguments if this is not possible.
If not, I assume we just take historyRead_raw from #1849 and be done with it.

include/ua_plugin_nodestore.h
```diff
@@ -188,7 +188,7 @@ typedef struct {
     UA_NODE_VARIABLEATTRIBUTES
     UA_Byte accessLevel;
     UA_Double minimumSamplingInterval;
-    UA_Boolean historizing; /* currently unsupported */
+    UA_Boolean historizing;
```
jpfr: The `read` and `write` services need to be extended so that users can actually interact with the historizing attribute.

src/server/ua_server_binary.c
```diff
@@ -221,6 +221,14 @@ getServicePointers(UA_UInt32 requestTypeId, const UA_DataType **requestType,
         *responseType = &UA_TYPES[UA_TYPES_SETMONITORINGMODERESPONSE];
         break;
 #endif
+#ifdef UA_ENABLE_HISTORIZING
+        // For History read
```
jpfr: Use C-style comments (/* */) instead of C++ (//)

include/ua_plugin_history_data_service.h
```diff
@@ -0,0 +1,79 @@
...
+     * Set it to NULL if you do not need it.
+     *
+     * context is the context of the UA_HistoryDataService. UA_HistoryDataService.context
+     * server is the server this node lives in.
+     * sessionId and sessionContext identify the session which set this value.
+     * nodeId the node id for which data was set.
+     * value the new value.
+     */
+    void
+    (*setValue)(void *context,
```
jpfr: The argument order should be
`UA_Server *server, void *haContext, const UA_NodeId *sessionId, void *sessionContext, const UA_NodeId *nodeId, void nodeContext, const UA_DataValue *value`

src/server/ua_services_attribute.c
```diff
@@ -1097,16 +1097,29 @@ writeValueAttribute(UA_Server *server, UA_Session *session,
...
+                                                       session ? session->sessionHandle : NULL,
+                                                       &node->nodeId,
+                                                       &adjustedValue);
+#endif
         /* Callback after writing */
         if(retval == UA_STATUSCODE_GOOD && node->value.data.callback.onWrite)
-            node->value.data.callback.onWrite(server, &session->sessionId,
-                                              session->sessionHandle, &node->nodeId,
+            node->value.data.callback.onWrite(server,
+                                              session ? &session->sessionId : NULL,
```
jpfr: There is always a session. We the plugin should not have to check for NULL here.
This can be state in the API docs.

We could add an assertion that session != NULL. to the service.

src/server/ua_services_attribute.c
```diff
@@ -1097,16 +1097,29 @@ writeValueAttribute(UA_Server *server, UA_Session *session,
         else
             retval = writeValueAttributeWithRange(node, &adjustedValue, rangeptr);
 
+#ifdef UA_ENABLE_HISTORIZING
+        if(retval == UA_STATUSCODE_GOOD && server->config.historyDataService.setValue)
```
jpfr: Check if the historizing attribute is true.
Only call `setValue` if that is the case.

What is the interaction between calling `setValue` here and a variable that is monitored with a local MonitoredItem for the HistoricalData? I presume for the latter case, the `historizing` attribute would be true as well?

include/ua_plugin_nodestore.h
peter-rustler-basyskom: Done

src/server/ua_server_binary.c
peter-rustler-basyskom: Done

src/server/ua_services_attribute.c
peter-rustler-basyskom: It should be possible to add values to the database even when the historizing attribute is not set.
But it is a good idea to provide the historizing attribute to the setValue call.
This is done.

jpfr:
:+1:

PR #1916, opened 2018-07-12
Changes in the OS-9 part of ua_network_tcp.c after the move to arch/

Kei-Thomsen:
After the ua_network_tcp.c has been moved to the arch/ directory, there are a few small modifications needed in the ifdef _OS9000 part. Before, it was errno__ which is not longer existing here and therefore errno again. And a remove of WIN32_INT as a casting. That's all to get the project compiling for OS-9 again.

Signed-off-by: Kei-Thomsen <thomsen@microsys.de>


jpfr:
Merged. Thanks. :smile:

Kei-Thomsen:
Thank you merging my OS-9 changes. I'm so happy about it.
Kei

PR #1918, opened 2018-07-12
Fix stack-use-after-scope in ua_network_tcp.c

basyskom-jvoe:
open62541 applications built with clang with ASAN enabled fail with
>==25154==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7ffd3812dac0 at pc 0x0000004ddc8d bp 0x7ffd3812d380 sp 0x7ffd3812cb30

This patch fixes the issue. For more details, see https://github.com/google/sanitizers/wiki/AddressSanitizerUseAfterScope


jpfr:
Thanks!

PR #1919, opened 2018-07-12
Remove manual adding of coverity certificate since we do not have sudo rights

Pro:


jpfr:
We used to have sudo rights.
Also because we need some custom packages.

Did we switch the travis environment?

Pro:
> Did we switch the travis environment?

Yes, 3 Weeks ago I switched to the dockerized builds:
https://github.com/open62541/open62541/pull/1876/commits/1ed15e59a3ce7d151ef85e5e49a1bee556c181aa

Sudo rights should not be necessary for cleanly defined build scripts, and dockerized build has the advantage that we could create an additional build script to run all the travis scripts locally in a (travis) docker before pushing.

This coverity certificate looks like it was a hack and if really necessary should be in the `before_install` script. To finally test it we need to merge this PR and see if it still can push to coverity (after the rest of CI is green).

jpfr:
Ok, cool.

Pro:
Looks like I was too optimistic... Travis somehow changed something in their build setup and that's why now all the travis builds fail. See:
https://github.com/travis-ci/travis-ci/issues/9876
https://github.com/travis-ci/travis-ci/issues/9875

PR #1922, opened 2018-07-16
Fix propagation of UA_ARCHITECTURE_* to install

ccvca:
The installed version does not have UA_ARCHITECTURE_* set as compiler definition. This sets the definition as compiled.

If not set, this error is shown (and many more):
>error: ‘UA_SOCKET’ does not name a type; did you mean ‘UA_FORMAT’?


ccvca:
Force pushed a commit to fix build

PR #1934, opened 2018-07-20
Select default architecture if it is not set through compiler define. Fixes #1925

### LINKED ISSUE #1925: error: ‘UA_SOCKET’ does not name a type;
--------------------------------------------------------------------------------

Description
===========
I have a compilation error when I compile open62541-master with C++ compilator :+1: 
[ 33%] Building CXX object CMakeFiles/openLoki.dir/fabrique.cpp.o
/usr/bin/c++    -I/home/idevice/OpenLoki/../open62541-master   -o CMakeFiles/openLoki.dir/fabrique.cpp.o -c /home/idevice/OpenLoki/fabrique.cpp
In file included from /home/idevice/OpenLoki/fabrique.h:6:0,
                 from /home/idevice/OpenLoki/fabrique.cpp:9:
/home/idevice/OpenLoki/../open62541-master/open62541.h:21300:5: error: ‘UA_SOCKET’ does not name a type
     UA_SOCKET sockfd;                 /* Most connectivity solutions run on
     ^
/home/idevice/OpenLoki/../open62541-master/open62541.h:25644:17: error: ‘UA_SOCKET’ was not declared in this scope
 ssize_t UA_send(UA_SOCKET sockfd, const void *buf, size_t len, int flags); //equivalent to posix send implementation
                 ^

Checklist
=========

 - [ ] open62541 Version (release number or git tag): Master
 - [ ] Other OPC UA SDKs used (client or server): Client
 - [ ] Operating system: Linux Debian 8
 - [ ] Critical issue

Pro:


cabralfortiss:
The problem is that when amalgamated, the ua_architecture.h is added first, and then the ua_config.h, so this won't work unless you put the ua_config.h first, but from what I remember, this breaks CI even worst, and also I think that's not right.

I'll go for the split ua_config.h file. Take the configuration variables specific and only for the open62541 stack above, and after the architecutre.h, put the rest of the config.h file, which are mostly more architecture dependant code, but more in a general way (I think this second part should be moved to the architecture.h file anyhow, but maybe not for now until this new architecture platform sink in to all users)

jpfr:
> The problem is that when amalgamated, the ua_architecture.h is added first, and then the ua_config.h, so this won't work unless you put the ua_config.h first, but from what I remember, this breaks CI even worst, and also I think that's not right.

The ua_architecture.h depends on ua_config.h. Because we add all the cmake definitions to ua_config.h.

The problem is that we are including ua_architecture.h in the middle of the config.h.
That's not going to fly with the amalgamation.
https://github.com/open62541/open62541/blob/master/include/ua_config.h.in#L63

Imho we should continue to include the config.h first.
And make it work for all architectures. As it was until a week ago.

If I have an amalgamated release and want to enable an additional feature, the top of the .h is where I expect the definitions.

cabralfortiss:
#1925 

Take it from my fork if you like it. I cannot push, I think, to this PR. 

https://github.com/cabralfortiss/open62541/commit/cb11d37bb598eeca631505ebc1a0240908bf3c67

jpfr:
Looks good! Ready to merge?

PR #1940, opened 2018-07-24
WIP: Fix #1930

### LINKED ISSUE #1930: BadSecureChannelTokenUnknown occur after 7m30 on MonitoredItems
Description
===========
A client initiate subcription/monitoredItems on 2 variables.
After 7mn30 of good working (this use case is reproductible), The client is disconnected : 
2018-07-19 11:13:01,018 ERROR [0x000055f7735b1e40] TSVC::OPCUABackend - runIterate(): fail, reason is  -2138636288 ( 'BadSecureChannelTokenUnknown' )

Checklist
=========

 - [ ] open62541 Version (release number or git tag): Master
 - [ ] Other OPC UA SDKs used (client or server): Client
 - [ ] Operating system: Debian 8
 - [ ]
[SubscriptionAborted.pcapng.gz](https://github.com/open62541/open62541/files/2209237/SubscriptionAborted.pcapng.gz)

 Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Critical issue

Infinity95:
#1930

### DIFF
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -330,51 +330,20 @@ hideBytesAsym(const UA_SecureChannel *channel, UA_Byte **buf_start, const UA_Byt
     }
 }
 
-/* Sends an OPN message using asymmetric encryption if defined */
-UA_StatusCode
-UA_SecureChannel_sendAsymmetricOPNMessage(UA_SecureChannel *channel, UA_UInt32 requestId,
-                                          const void *content, const UA_DataType *contentType) {
-    if(channel->securityMode == UA_MESSAGESECURITYMODE_INVALID)
-        return UA_STATUSCODE_BADSECURITYMODEREJECTED;
-
+static UA_StatusCode
+padChunkAsym(UA_SecureChannel *channel, const UA_ByteString *const buf, size_t securityHeaderLength,
+             UA_Byte **buf_pos) {
     const UA_SecurityPolicy *const securityPolicy = channel->securityPolicy;
-    UA_Connection *connection = channel->connection;
-    if(!connection)
-        return UA_STATUSCODE_BADINTERNALERROR;
 
-    /* Allocate the message buffer */
-    UA_ByteString buf = UA_BYTESTRING_NULL;
-    UA_StatusCode retval =
-        connection->getSendBuffer(connection, connection->config.sendBufferSize, &buf);
-    if(retval != UA_STATUSCODE_GOOD)
-        return retval;
-
-    /* Restrict buffer to the available space for the payload */
-    UA_Byte *buf_pos = buf.data;
-    const UA_Byte *buf_end = &buf.data[buf.length];
-    hideBytesAsym(channel, &buf_pos, &buf_end);
-
-    /* Encode the message type and content */
-    UA_NodeId typeId = UA_NODEID_NUMERIC(0, contentType->binaryEncodingId);
-    retval = UA_encodeBinary(&typeId, &UA_TYPES[UA_TYPES_NODEID], &buf_pos, &buf_end, NULL, NULL);
-    retval |= UA_encodeBinary(content, contentType, &buf_pos, &buf_end, NULL, NULL);
-    if(retval != UA_STATUSCODE_GOOD) {
-        connection->releaseSendBuffer(connection, &buf);
-        return retval;
-    }
-
-    /* Compute the length of the asym header */
-    const size_t securityHeaderLength = calculateAsymAlgSecurityHeaderLength(channel);
-
-    /* Pad the message. Also if securitymode is only sign, since we are using
+    /* Also pad if the securityMode is SIGN_ONLY, since we are using
      * asymmetric communication to exchange keys and thus need to encrypt. */
     if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGN ||
        channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT) {
         const UA_Byte *buf_body_start =
-            &buf.data[UA_SECURE_CONVERSATION_MESSAGE_HEADER_LENGTH +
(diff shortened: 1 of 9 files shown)


Infinity95:
We basically have two options: have a current and next token for both directions, or a last, current and next token.
We need this because it is possible that we are already sending with the next token, but the current token is still used by the server. So when we revolve the tokens the current token is discarded at the moment.

jpfr:
Exactly. Imho we can live with just

- latestToken
- previousToken

Holding two tokens is only required for the transition between tokens.
Then we would allow either `previous` or `latest`.
Once we received a message for the `lastetToken`, we know that the switchover is complete and we can delete the content of `previousToken`.

Infinity95:
Ah, you meant deleting the token, but still having the variable?

Infinity95:
We have two problems i believe. First, we cannot use two variables like we currently have, because we currently call revolveTokens in the client to switch the tokens for sending. We have to do this, because the securechannel needs to use the new keys. But then we cannot keep the old token in the other variable because other checks assume that the other variable (nextToken) contains a fresh token after a renew has been called. Either we use two separate token pairs for sending and receiving or we use three tokens (prev, curr, next). This makes everything easier and i don't see any disadvantages to this approach. (I mean we could save like 20 bytes, but is that really a concern here?)
Also, we have to keep the old keys for receiving while already using newly generated keys for sending.
So we would have to split the key generation into first generating local keys for sending, and then generate the remote keys, as soon as we receive a message with the new token. (I believe this should be possible without many changes)

@jpfr I think it would be best if i incorporate the changes you made to the chunk processing into this PR. Otherwise i can't thoroughly test if my fixes are working. I'll just rebase my branch onto yours and continue from there?

jpfr:
Yes, you can rebase on top of the other branch.

The functionality is ready.

There remains a CI issue I did not have the time to look at.

Ok. Let’s talk Monday.

Infinity95:
Ok the temporary fix seems to be working so far.
I'm still struggling with CI though.

@Pro could you have a look at what's going wrong with the fuzzing tests?
I don't really know how to interpret the output of the log.

Pro:
Sometimes the log in Travis becomes too long and you have to open the raw log:
https://api.travis-ci.org/v3/job/421077858/log.txt

That shows the error message (with some ASCII Encoded color output, but just ignore these):

```
[ 77%] [32mBuilding C object CMakeFiles/open62541-object.dir/src/server/ua_subscription_datachange.c.o[0m
[1m/home/travis/build/open62541/open62541/src/ua_securechannel.c:1335:32: [0m[0;1;31merror: [0m[1m
      expected ';' after expression[0m
    retval = UA_STATUSCODE_GOOD
[0;1;32m                               ^
[0m[0;32m                               ;
[0m1 error generated.
make[2]: *** [CMakeFiles/open62541-object.dir/src/ua_securechannel.c.o] Error 1
```

If you look into the code the error is obvious (as the output already says):
https://github.com/open62541/open62541/blob/50442fa5853bd8c7715e4210beef72f8010ed56d/src/ua_securechannel.c#L1335

Missing ';'

Infinity95:
Thanks for the quick reply. Classical error on my part :sweat_smile:
Should run through now. Any idea on how to fix the mingw CI? It keeps complaining about unsupported format specifiers. I've tried using lu, zu, PRIu64 and PRIuPTR without any success.

Infinity95:
@Pro The last failing step is due to visual studio 2008 not having inttypes.h
Is it possible that we package the file with our implementation to avoid this error? Or should i look for another solution for my formatting? (maybe cast to int and print with int format specifiers??)

tests/fuzz/ua_debug_dump_pkgs_file.c
```diff
@@ -157,9 +157,8 @@ UA_debug_dumpCompleteChunk(UA_Server *const server, UA_Connection *const connect
         TAILQ_INIT(&dummy.messages);
         UA_ByteString messageBufferCopy;
         UA_ByteString_copy(messageBuffer, &messageBufferCopy);
-        UA_SecureChannel_decryptAddChunk(&dummy, &messageBufferCopy);
+        UA_SecureChannel_decryptAddChunk(&dummy, &messageBufferCopy, UA_TRUE);
         UA_SecureChannel_processCompleteMessages(&dummy, &dump_filename, UA_debug_dump_setName_withChannel);
-        UA_SecureChannel_deleteMessages(&dummy);
```
jpfr: There could be chunks in the dummy SecureChannel that have not been complete.
Better keep `UA_SecureChannel_deleteMessages(&dummy);`

src/ua_securechannel.h
```diff
@@ -77,6 +77,7 @@ struct UA_SecureChannel {
     UA_MessageSecurityMode  securityMode;
     UA_ChannelSecurityToken securityToken; /* the channelId is contained in the securityToken */
```
jpfr: Add a few comments that explain the securityTokens and how they are rotated.
This will save the next developer a few hours of tracking things down.

src/ua_securechannel.c
```diff
@@ -1007,21 +1166,73 @@ checkAsymHeader(UA_SecureChannel *const channel,
...
 
+static UA_StatusCode
+putPayload(UA_SecureChannel *const channel, UA_UInt32 const requestId, UA_MessageType const messageType,
+           UA_ChunkType const chunkType, UA_ByteString *chunkPayload) {
+    switch(chunkType) {
+    case UA_CHUNKTYPE_INTERMEDIATE:
+    case UA_CHUNKTYPE_FINAL:
+        return addChunkPayload(channel, requestId, messageType,
+                               chunkPayload, chunkType == UA_CHUNKTYPE_FINAL);
+    case UA_CHUNKTYPE_ABORT:deleteLatestMessage(channel, requestId);
```
jpfr: That indendation is pretty hard to read.

jpfr:
Is this still WIP or can we kick off the next review?

Issue #1946, opened 2018-07-30
Compile Error due [help wanted]


fabian727:
Hi guys,
I get compiler errors. I can not state which effect creates to start this error. All I can state is:

1. clone repo
2. init submodules
3. create dirs for building
4. initialisation for building: cmake ..; ccmake . (enable FULL_Namespace,Encryption)
5. make -j4
6. Everything works fine. Some time goes by
7. ccmake . (for new configuration, like Encryption or Events)
8. make clean && make -j4
9. see following logging-output:

```
...
Integrating file '/home/user/open62541/arch/ua_architecture_functions.h'...done.
The size of /home/user/open62541/debug_x86/open62541.h is 1676345 Bytes.
[ 14%] Building C object CMakeFiles/open62541-plugins.dir/plugins/ua_nodestore_default.c.o
[ 18%] Built target open62541-amalgamation-header
[ 18%] Building C object CMakeFiles/open62541-plugins.dir/plugins/ua_config_default.c.o
[ 20%] Building C object CMakeFiles/open62541-plugins.dir/plugins/ua_securitypolicy_none.c.o
[ 22%] Building C object CMakeFiles/open62541-plugins.dir/plugins/ua_securitypolicy_basic128rsa15.c.o
Scanning dependencies of target open62541-object
[ 22%] Building C object CMakeFiles/open62541-plugins.dir/plugins/ua_securitypolicy_basic256sha256.c.o
[ 24%] Building C object CMakeFiles/open62541-plugins.dir/arch/ua_network_tcp.c.o
[ 26%] Building C object CMakeFiles/open62541-object.dir/src/ua_types.c.o
[ 28%] Building C object CMakeFiles/open62541-object.dir/src/ua_types_encoding_binary.c.o
[ 36%] Built target open62541-plugins
[ 36%] Building C object CMakeFiles/open62541-object.dir/src_generated/ua_types_generated.c.o
[ 38%] Building C object CMakeFiles/open62541-object.dir/src_generated/ua_transport_generated.c.o
[ 38%] Building C object CMakeFiles/open62541-object.dir/src/ua_util.c.o
[ 40%] Building C object CMakeFiles/open62541-object.dir/src/ua_connection.c.o
[ 40%] Building C object CMakeFiles/open62541-object.dir/src/ua_securechannel.c.o
[ 42%] Building C object CMakeFiles/open62541-object.dir/src/server/ua_session.c.o
[...]

PR #1948, opened 2018-07-30
same call header for UA_parseEndpointUrl

fabian727:


Pro:
Closing since issue #1946 was resolved without this cosmetic change.

fabian727:
it' s a bit of a cosmetic and a bit of correct coding style:
[https://github.com/open62541/open62541/blob/master/src/ua_util.c#L35](c-code)
```UA_StatusCode
UA_parseEndpointUrl(const UA_String *endpointUrl, UA_String *outHostname,
u16 *outPort, UA_String *outPath) {
```
[https://github.com/open62541/open62541/blob/master/include/ua_util.h#L40](header)
```UA_StatusCode UA_EXPORT
UA_parseEndpointUrl(const UA_String *endpointUrl, UA_String *outHostname,
UA_UInt16 *outPort, UA_String *outPath);
```

Yes the UA_UInt16 is defined as u16, but then you could just spare it. Else I would recommend to use it ;)

PR #1951, opened 2018-07-31
[WIP] JSON Encoding

lukeMlng:


src/ua_types_encoding_json.c
```diff
@@ -0,0 +1,5101 @@
...
+                }
+                ret |= calcWriteKey(ctx, "Body", commaNeeded);
+                ret |= CALC_DIRECT(&src->content.encoded.body, String);
+                break;
+            }
+            default:
+                ret = UA_STATUSCODE_BADINTERNALERROR;
+        }
+
+        ret |= ADDJSONCHAR(ObjEnd);
```
jpfr: There is a huge if/else in this function. Its difficult to read for an outsider.
After the if/else block we return directly. So this can be simplified.

1. If the `if` is not visible from the `else` without scrolling, use comments to spell out what the else means
```c
} else { /* Now begins the case where... */
```
2. If the method returns directly after the if/else block, return from the function at the end of the if-block. And don't use an else at all.
3. If argument 2. is the case, put the less likely code path in the if-block. And the hot-path directly.
4. If large blocks of code remain, refactor into smaller methods called from a "master" function.
Use UA_assert to document what we know for sure about the input.

jpfr: This general hint can/should be applied in several similar places in this file.

src/ua_types_encoding_json.c
```diff
@@ -0,0 +1,5101 @@
...
+            ret |= calcWriteKey(ctx, "Body", UA_FALSE);
+            ret |= calcJsonInternal(src->content.decoded.data, contentType, ctx, useReversible);
+            ret |= ADDJSONCHAR(ObjEnd);
+        }
+    }
+    return ret;
+}
+
+static status
+Variant_calcJsonWrapExtensionObject(const UA_Variant *src, const bool isArray, CtxJson *ctx, UA_Boolean useReversible) {
```
jpfr: That line is too long. I'm in the habit of splitting the editor window vertically.
We have no fixed rule. But maybe we can keep to <120 chars per line.

src/ua_types_encoding_json.c
```diff
@@ -0,0 +1,5101 @@
...
+    }
+    
+    if(length > 1){
+        ADDJSONCHAR(ArrayEnd);
+    }
+    return ret;
+}
+
+static status
+calcAddMatrixContentJSON(CtxJson *ctx, void* array, const UA_DataType *type, size_t *index, UA_UInt32 *arrayDimensions, size_t dimensionIndex, size_t dimensionSize, UA_Boolean useReversible) {
```
jpfr: Why do we specifically have to have code for matrices?
The binary encoding does not differentiate between vector/matrix/tensor.

src/ua_types_encoding_json.c
```diff
@@ -0,0 +1,5101 @@
...
+        UA_Boolean commaNeeded = UA_FALSE;
+
+        ret |= ADDJSONCHAR(ArrayStart);
+
+        for (size_t i = 0; i < arrayDimensions[dimensionIndex]; i++) {
+            if (commaNeeded) {
+                ret |= ADDJSONCHAR(Comma);
+            }
+
+            ret |= calcJsonInternal(((u8*)array) + (type->memSize * *index), type, ctx, useReversible);
```
jpfr: Please put braces around `*index`. The `* *` looks pretty scary.

src/ua_types_encoding_json.c
```diff
@@ -0,0 +1,5101 @@
...
+    ctx->depth++;
+    
+    if (dimensionIndex == (dimensionSize - 1)) {
+        /* Stop recursion: The inner Arrays are written */
+        UA_Boolean commaNeeded = UA_FALSE;
+
+        ret |= ADDJSONCHAR(ArrayStart);
+
+        for (size_t i = 0; i < arrayDimensions[dimensionIndex]; i++) {
+            if (commaNeeded) {
```
jpfr: We don't need `commaNeeded` here.
The test can just be `if(i > 0)`.

src/ua_types_encoding_json.c
jpfr: For me, a matrix always has 2 dimensions.
If you want to handle any number of dimensions, then call this thing multi-array.

jpfr:
I appended a few fast commits with simplifications to the control flow.
Please review, take as inspiration and squash into the main commit when the changes are okay.

jpfr:
This `commaNeeded` business prevents additional simplifications of the code...
Essentially, we need to track whether we are in an object or array and there is an element before ours.

- Add an array `UA_Boolean commaNeeded[UA_ENCODING_MAX_RECURSION` to the CtxJson structure.
- The current `depth` is an Index into `commaNeeded`.
- Set `commaNeeded[++ctx->depth] = false` when we enter into a new object or array
- The API becomes as following:

```c
startObject(ctx); // write "{", increase the depth, set commaNeeded at depth to false
endObject(ctx); // write "}", decrease the depth
writeKey(ctx, name, value, type); // write comma if needed, write the key-value pair

// analog API for arrays
startArray(ctx);
endArray(ctx);
writeElement(ctx, value, type);
```

In many places, we have an if/else just for the commaNeeded stuff. This code can then be unified.

jpfr:
Rebased on master.
Let's merge this soon.

jpfr:
Merged! Congrats @lukeMlng for this great PR. 😄

PR #1956, opened 2018-08-01
[WIP] Always process chunks in-order

jpfr:
Before, the client could perform a RenewSecureChannel between two
chunks.

Still WIP. To be merged after #1940.


Infinity95:
I believe this should be merged before #1940 because otherwise the fix won't work and i cannot test if it will work (since the incorrect order prevents the other checks from passing)

PR #1965, opened 2018-08-07
fix crash for client connect async / fix #1912

### LINKED ISSUE #1912: Asynchronous connect with inaccessible server
I am encountering an issue with the asynchronous client connect in cases where the server is not accessible at the time of the call to UA_Client_connect_async(), but becomes accessible a short while later. My aim is to keep attempting to connect until the server appears and the connection is established.

I have tried the following scenario:

1) call UA_Client_connect_async(...)
2) call UA_Client_run_iterate(client, 0) periodically
...
3) server becomes accessible
...
4) still calling UA_Client_run_iterate(client, 0) periodically
5) but the client never connects

The first few calls to `UA_Client_run_iterate()` returned `UA_STATUSCODE_GOOD`, then the status changed to `UA_STATUSCODE_BADCONNECTIONCLOSED` and I started seeing the output

```
error/network      No connection to server.
```

on the console. The client never connected.

Since this was not working as expected, I tried calling the pair

```
UA_Client_connect_async(...)
UA_Client_run_iterate(client, 0)
```

in a loop. This actually works if the server becomes accessible after several such calls; but if the loop with the pair of functions executes more than a few dozen times, the process eventually crashes with SIGSEGV in `UA_Connection_receiveChunksNonBlocking()`.

I can provide more data, including backtraces, valgrind reports, and possibly a minimal example, but I would first like to know whether I am missing something obvious.

Context:

open62541 0.4 development version, master from about two weeks ago
CentOS 7.2
Server that I'm trying to connect to is written in Java (Eclipse Milo)

StalderT:
fix crash in #1912

### DIFF
--- a/arch/ua_network_tcp.c
+++ b/arch/ua_network_tcp.c
@@ -551,8 +551,11 @@ static void
 ClientNetworkLayerTCP_close(UA_Connection *connection) {
     if (connection->state == UA_CONNECTION_CLOSED)
         return;
-    UA_shutdown(connection->sockfd, 2);
-    UA_close(connection->sockfd);
+
+    if(connection->sockfd != UA_INVALID_SOCKET) {
+        UA_shutdown(connection->sockfd, 2);
+        UA_close(connection->sockfd);
+    }
     connection->state = UA_CONNECTION_CLOSED;
 }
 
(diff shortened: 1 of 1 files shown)


jpfr:
This will probably break again with the next refactor.
Can we have a unit test for the scenario from #1912?

StalderT:
> Can we have a unit test for the scenario from #1912?

Hard to do! I will try next week.

PR #1974, opened 2018-08-12
NodesetGenerator: CharArray is now considered as String

zanettea:
Trying to run the script generate_datatypes.py on data type definition (*.bsd) like the following:

```xml
    <opc:StructuredType BaseType="ua:ExtensionObject" Name="ExtendedKeyValuePair">
        <opc:Field TypeName="opc:CharArray" Name="key"/>
        <opc:Field TypeName="ua:Variant" Name="value"/>
        <opc:Field TypeName="ua:LocalizedText" Name="description"/>
    </opc:StructuredType>
```

the script enters an infinite loop, basically because the type `opc:CharArray ` is not recognized as builtin type String. The patch ensures that `opc:CharArray ` is recognized and transformed to String.

Btw, the standard says:

```xml
<opc:StructuredType Name="String">
  <opc:Documentation>A UTF-8 null terminated string value.</opc:Documentation>
  <opc:Field Name="Value" TypeName="Char" Terminator="00" />
</opc:StructuredType>
<opc:StructuredType Name="CharArray">
  <opc:Documentation>A UTF-8 string prefixed by its length in characters.</opc:Documentation>
  <opc:Field Name="Length" TypeName="Int32" />
  <opc:Field Name="Value" TypeName="Char" LengthField="Length" />
</opc:StructuredType>
```

but apparently the library only deals with CharArray calling it String.

### DIFF
--- a/tools/generate_datatypes.py
+++ b/tools/generate_datatypes.py
@@ -271,6 +271,16 @@ def parseTypeDefinitions(outname, xmlDescription, namespace):
                     return False
         return True
 
+    def unknownTypes(element):
+        "Return all unknown types"
+        unknowns = []
+        for child in element:
+            if child.tag == "{http://opcfoundation.org/BinarySchema/}Field":
+                childname = child.get("TypeName")
+                if childname[childname.find(":")+1:] not in types:
+                    unknowns.append(childname)
+        return unknowns
+
     def skipType(name):
         if name in excluded_types:
             return True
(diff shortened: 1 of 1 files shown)


jpfr:
> String: A UTF-8 null terminated string value.

That is surprising. Part 6 of the standard says explicitly
"All String values are encoded as a sequence of UTF-8 characters without a null terminator and preceded by the length in bytes."

The 1.04 type definitions (https://opcfoundation.org/UA/schemas/1.04/Opc.Ua.Types.bsd) don't actually define CharArray. It is only used for structure members.

Maybe we can solve this without adding CharArray to the list of builtins.
And instead just rewrite the type name when parsing structure members.

jpfr:

PR #1985, opened 2018-08-16
Fix read attribute EventNotifier from Object or View since Events support

mlang-de:
The View and Object node structures are not equal anymore if UA Events support is enabled.


jpfr:
Thanks for the fix. 👍 
Just a cosmetic request: Can we have the usual 4-space indendation?

Then we can merge right away.

PR #1989, opened 2018-08-17
Move referencesSize assignement ahead of the if-statement

mlang-de:
If the src-referenecesSize is <=0 the dst-referencesSize will not be initialized since in function UA_Node_copy_alloc no memset is done.

### DIFF
--- a/src/server/ua_nodes.c
+++ b/src/server/ua_nodes.c
@@ -162,6 +162,7 @@ UA_Node_copy(const UA_Node *src, UA_Node *dst) {
 
     /* Copy the references */
     dst->references = NULL;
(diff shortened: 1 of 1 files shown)


jpfr:
We need to be careful here.
With the change, if the calloc fails, `UA_Node_deleteMembers(dst)` is called with new dst->referencesSize > 0 but dst->references == NULL.

Usually, we assign the array size only after allocation has succeeded.

mlang-de:
I can close the PullRequest in the current master:head will be calloc used. I have an older implementation of function UA_Node_copy_alloc where malloc was used!

PR #1990, opened 2018-08-17
Add function UA_Server_addNode_raw

mlang-de:
This function is required if the application will add nodes from a XML-Nodeset.


Pro:
Can you explain in a bit more detail why we would need this method in the User space? Up to now it is not necessary to have this method in the user space to add a XML Nodeset. We have the _begin method (which compared to the _raw also does the typecheck and adds references to parent)

Since the method would then be in the public API, there should also be a documentation for it in the header.

mlang-de:
At the moment we add the binary converted-XML unordered to address space. So we need a rework our XML-to-binary parser to get an ordered binary file. Then we can use the _begin and _end-method. We put this task on the roadmap. Based on this I will close this PR.

PR #1852, opened 2018-06-14
High level plugin implementation for history data plugin. (read_raw)

peter-rustler-basyskom:

PR #1994, opened 2018-08-20
Historical data plugin

jpfr:
This PR builds on #1852 in the two following ways.

- The plugin struct, UA_HistoricalDataService, is renamed to UA_HistoricalDatabase.
A service has a specific meaning in OPC UA. And the plugin provides several of these services.
- The new public API from #1852 is moved to /plugins/historydata.
We don't need this as public API at the moment. And this gives us the opportunity to performance-test and get experience before we freeze the additional API.

The actual functionality of #1852 is 100% left unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,7 +257,7 @@ endif()
 
 if(UA_ENABLE_HISTORIZING)
     set(historizing_exported_headers
-        ${PROJECT_SOURCE_DIR}/include/ua_plugin_history_data_service.h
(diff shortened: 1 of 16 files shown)


jpfr:
@Pro This is a rather big PR. We should have two pairs of eyes on it.
Do you approve?

Pro:
If it's OK I will have a look at it end of next week, this week is completely packed.

jpfr:
👍

jpfr:
Great! Good job @peter-rustler-basyskom 👍

Issue #1839, opened 2018-06-11
Custom event returns empty variant


NIPE-SYSTEMS:
I'm trying to send a custom event (event type `ns=2;i=1005`, for nodeset see below) from an open62541-server to an open62541-client. It seems that when I'm sending a populated event to the client, the client gets some empty variants instead of the populated data. It seems that this only effects fields of the custom event type. Fields of e.g. the `BaseEventType` seem to work fine.

I might have done it wrong. After hours of debugging I haven't found the issue.

The detailed informations below consist of (hopefully) the whole code that is needed for triggering and receiving the custom event:

1. open62541 client: subscribing to the event
2. open62541 client: subscription callback
3. open62541 server: triggering the event
4. and: the nodeset of the custom event type

### 1. open62541 client: subscribing to the event

```cpp
#define UA_CLIENT_SELECT_CLAUSES_AMOUNT 3

uint16_t m_namespace = 2; // via UA_Client_NamespaceGetIndex

UA_MonitoredItemCreateRequest request = UA_MonitoredItemCreateRequest_default(UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER));
request.itemToMonitor.attributeId = UA_ATTRIBUTEID_EVENTNOTIFIER;
request.requestedParameters.samplingInterval = 100;
UA_EventFilter_init(&m_filter);
m_filter.selectClauses = (UA_SimpleAttributeOperand*)UA_Array_new(UA_CLIENT_SELECT_CLAUSES_AMOUNT, &UA_TYPES[UA_TYPES_SIMPLEATTRIBUTEOPERAND]);
if(!m_filter.selectClauses) {
    m_logger.error("Failed to allocate select clauses for subscription.");
    return false;
}
for(size_t i = 0; i < UA_CLIENT_SELECT_CLAUSES_AMOUNT; ++i) {
    UA_SimpleAttributeOperand_init(&m_filter.selectClauses[i]);
}

// ---------------------------------------------------------------------
// SELECT CLAUSE 1 ("2:IsAddition" from ns=2;i=1005)

m_filter.selectClauses[0].typeDefinitionId = UA_NODEID_NUMERIC(m_namespace, 1005);
m_filter.selectClauses[0].browsePathSize = 1;
m_filter.selectClauses[0].browsePath = (UA_QualifiedName*)UA_Array_new(m_filter.selectClauses[0].browsePathSize, &UA_TYPES[UA_TYPES_QUALIFIEDNAME]);
[...]

PR #1996, opened 2018-08-21
add EventFields 'Time' and 'SourceName' in server events tutorial

### LINKED ISSUE #1995: UAExpert does't show events of the server events tutorial
Hello everyone,

I've tried to get the demo from **examples/tutorial_server_events.c** running, however it was not as trivial as expected...

I tested the demo first with UAExpert, which didn't show the event triggered by the UAMethod "Generate Event" in its event view. Testing with other UA Clients (Softing and OPC Foundation Sample Client) I got the triggert event, but only after a RepublishRequest (in about 18 seconds after the method call). 
What I've noticed by the "working clients", is that when I change the filter settings of the Clause Fields, so I only have the EventFields checked which are getting set by the demo program (SourceNode, ReceiveTime, EventId, EventType) by the function **eventSetStandardFields()** from "ua_subscription_events.c" and (Message, Severity) by the function **setUpEvent** from "tutorial_server_events.c", I get the events directly after calling the triggering method.

Unfortunately setting up the filter in UAExpert is restricted, so I added the missing EventField "Time" (property of the BaseEventType) to the function **setUpEvent** and the triggered event is now appearing in UAExpert!

Wouldn't it be better to add all the mandatory EventFields and not only (Message and Severity) to the demo program, so it can be tested directly without the need to modify the demo program or the Client filters?

That could be the same problem as already mentioned in issue #1839.

Best regards

sqadasi:
I extended the example to add the rest mandatory EventFields 'Time' and 'SourceName'.
UaExpert Client needs the EventField 'Time' to show the events on Event View.

That fixes #1995 and maybe relevant to issue #1839.

@Pro and @aDogCalledSpot please check it:)

### DIFF
--- a/examples/tutorial_server_events.c
+++ b/examples/tutorial_server_events.c
@@ -41,7 +41,8 @@ addNewEventType(UA_Server *server) {
  * All we need for this is our `EventType`. Once we have our event node, which is saved internally as an `ObjectNode`,
  * we can define the attributes the event has the same way we would define the attributes of an object node. It is not
  * necessary to define the attributes `EventId`, `ReceiveTime`, `SourceNode` or `EventType` since these are set
- * automatically by the server. In this example, we will only be setting `Severity` and `Message`.
+ * automatically by the server. In this example, we will be setting all the rest mandatory EventFields of the BaseEventType: 
+ * `Time` (to make the example UaExpert compliant), `Severity`, `Message` and `SourceName`.
  */
 static UA_StatusCode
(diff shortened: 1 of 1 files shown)


sqadasi:
> That's not the right place there, since the user should be able to set his own Time and SourceName value. If we would set it in triggerEvent, it would overwrite the user value.

@Pro you are right, 'Time' is the time at which the event occured and should be set by the user at the closest place to the event generator.

PR #1998, opened 2018-08-21
Store retransmission message in right order

mlang-de:
CTT subscription Basic script passed for me.


StalderT:
Does the CTT pass by applying this patch on version 0.3?

mlang-de:
Unfortunately I can not test version 0.3, but it is only 1 line changed and it looks like it could fit into the branch 0.3.

Pro:
Nice! Thanks! I will backport it to 0.3

PR #1999, opened 2018-08-21
 Fix CTT errors for subscription sequence number handling

mlang-de:
- rename subscription variable sequence number to nextSequenceNumber
- adjust logic


mlang-de:
@StalderT it seems to me that this PR could also be applied to branch 0.3.

PR #2013, opened 2018-08-24
Move CreateSession in front of the EndpointDescription creation

mlang-de:
Fixes CTT error for "Session Minimum 2 Parallel/002.js": CreateSession.Response.ServerEndpoints are not needed when a ServiceFault is returned. Expected <0> but got <1>.


Pro:
This also needs to be picked onto 0.3

Pro:
Thanks @mlang-de!!

jpfr:
Great!

PR #2014, opened 2018-08-24
[WIP] Remove OR'ing statuscodes

Infinity95:
By OR'ing statuscodes we lose information in some cases and in those cases where they don't overlap it complicates getting the actual statuscodes.
Also in some cases we only check at the end of a function if we encountered an error. If the error happens at the beginning of the function the rest is still executed which might lead to other problems.

### DIFF
--- a/plugins/ua_securitypolicy_basic128rsa15.c
+++ b/plugins/ua_securitypolicy_basic128rsa15.c
@@ -520,7 +520,7 @@ sym_generateKey_sp_basic128rsa15(const UA_SecurityPolicy *securityPolicy,
         if(offset + hashLen > out->length) {
             outSegment.data = NULL;
             outSegment.length = 0;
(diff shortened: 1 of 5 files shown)


jpfr:
This looks good!
But CI will probably fail in some places.

Some compilers, e.g. old VisualStudio, do not want to `goto` over variable declarations.
You might have to move some variable declarations to the front.

Or avoid goto entirely by moving code into a new function that is wrapped by the cleanup /error part.

jpfr:
Oh. He ate it. Cool...

Infinity95:
I'm still working on some other files, so i'll just create a new PR for those as soon as they are done.

jpfr:
You can prefix the PR title with [WIP] if it is not done.

Infinity95:
As was the case with this one 😅

PR #2018, opened 2018-08-27
fix Service_CreateSession after #2013

StalderT:
Hello,

With #2013, ```UA_SessionManager_createSession``` was moved up.

In some cases of errors the session is not removed. This PR fix this problem.

In some other places the value of ```response->responseHeader.serviceResult``` is overwritten. This PR fix this problem.

@pro @jpfr  can you review the code?

@mlang-de can you test this PR with CTT?

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
  *
  *    Copyright 2014-2018 (c) Julius Pfrommer, Fraunhofer IOSB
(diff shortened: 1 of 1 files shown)


mlang-de:
I agree with the current changes. Maybe it make sense to return the result from line 142 (UA_String_copy) to response->responseHeader.serviceResult |= too?

StalderT:
> Maybe it make sense to return the result from line 142 (UA_String_copy) to response->responseHeader.serviceResult |= too?

Ok thanks! Tomorrow I modify the PR.

jpfr:
Thanks!

PR #2021, opened 2018-08-28
Update URIs in SecurityPolicy

ckmk14:
The URIs for the "SecurityPolicy [B] - Basic256Sha256" are not set correctly.

Regarding [1] the RSA signtaure algorithm has the URI "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"

[1] https://opcfoundation-onlineapplications.org/profilereporting/index.htm?ModifyProfile.aspx?ProfileID=57c0f86d-400a-4146-a919-cb65679aea97


Pro:
@Infinity95

Infinity95:
This change is correct. We had already discussed this in person.

jpfr:
Great :+1:

PR #2022, opened 2018-08-28
Remove invalid free in ua_subscription_events

aDogCalledSpot:
An invalid free was causing the server to crash when either creating multiple monitored items on a single node or shutting down the client before closing the subscription.

### DIFF
--- a/src/server/ua_subscription_events.c
+++ b/src/server/ua_subscription_events.c
@@ -28,7 +28,8 @@ UA_MonitoredItem_removeNodeEventCallback(UA_Server *server, UA_Session *session,
             UA_MonitoredItem *iter = ((UA_ObjectNode *) node)->monitoredItemQueue;
             for (; iter->next != entry; iter=iter->next) {}
             iter->next = entry->next;
-            UA_free(entry);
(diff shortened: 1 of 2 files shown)


jpfr:
Won’t this lead to leaks?
Note that UA_Server_delayedFree was added exactly for this use case.

Also, we should cover the scenario you explained with a unit test. Otherwise this will regress over time.

aDogCalledSpot:
The monitored item already is being freed with a delayedFree at the end of `UA_MonitoredItem_delete`, so memory leakts aren't happening. I'll make sure to add a unit test.

jpfr:
Hmm. CLA-Assistant seems to hang on this issue.
@FlorianPalm, did you register CLA-Assistant for us?

It should be possible that you re-check the open PR on their site to unblock this one.

Pro:
I regsitered CLAassistant and just did a recheck.
Anyways can you please check that you also have access to cla @jpfr.
If I remember correctly, every member of the orga should be able to control it

jpfr:
Ah ok.
No I don't have access from my account.

PR #2026, opened 2018-08-29
added arch for windows embedded compact 7

crafti5:
There are some questions in code what shall we do with particular snippets, like the missing time functions or the "errorno == EINTR".

If someone has an idea how we can fix this then a will do this and also test this for WEC7 :)

### DIFF
--- a/arch/ua_architecture_definitions.h
+++ b/arch/ua_architecture_definitions.h
@@ -17,6 +17,9 @@
 #include <stddef.h>
 
 /* Include stdint.h and stdbool.h or workaround for older Visual Studios */
+#ifdef UNDER_CE
+# include "stdint.h"
+#endif
 #if !defined(_MSC_VER) || _MSC_VER >= 1600
 # include <stdint.h>
 # include <stdbool.h> /* C99 Boolean */
--- a/arch/ua_network_tcp.c
+++ b/arch/ua_network_tcp.c
@@ -608,7 +608,7 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
 
     if(clientsockfd == UA_INVALID_SOCKET) {
             UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Could not create client socket: %s", UA_ERRNO);
+                            "Could not create client socket: %s", strerror(UA_ERRNO));
             ClientNetworkLayerTCP_close(connection);
             return UA_STATUSCODE_BADDISCONNECT;
     }
@@ -628,7 +628,7 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
     if ((error == -1) && (UA_ERRNO != UA_ERR_CONNECTION_PROGRESS)) {
             ClientNetworkLayerTCP_close(connection);
             UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Connection to  failed with error: %s", UA_ERRNO);
+                            "Connection to  failed with error: %s", strerror(UA_ERRNO));
             return UA_STATUSCODE_BADDISCONNECT;
     }
 
--- a/arch/ua_network_tcp.c
+++ b/arch/ua_network_tcp.c
@@ -608,7 +608,7 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
 
     if(clientsockfd == UA_INVALID_SOCKET) {
             UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Could not create client socket: %s", strerror(UA_ERRNO));
+                            "Could not create client socket: %s", UA_ERRNO);
             ClientNetworkLayerTCP_close(connection);
             return UA_STATUSCODE_BADDISCONNECT;
     }
@@ -628,7 +628,7 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
     if ((error == -1) && (UA_ERRNO != UA_ERR_CONNECTION_PROGRESS)) {
             ClientNetworkLayerTCP_close(connection);
             UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Connection to  failed with error: %s", strerror(UA_ERRNO));
+                            "Connection to  failed with error: %s", UA_ERRNO);
             return UA_STATUSCODE_BADDISCONNECT;
     }
 
--- a/arch/wec7/ua_architecture.h
+++ b/arch/wec7/ua_architecture.h
@@ -130,6 +130,7 @@ void UA_sleep_ms(size_t ms);
 }
 #define UA_LOG_SOCKET_ERRNO_GAI_WRAP UA_LOG_SOCKET_ERRNO_WRAP
 
+#define UA_fileExists
 #include "../ua_architecture_functions.h"
 
 /* Fix redefinition of SLIST_ENTRY on mingw winnt.h */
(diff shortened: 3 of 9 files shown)


jpfr:
Great! Is there a way to test this in our Continuous Integration infrastructure?

Otherwise the build could break when we make changes to the library.

crafti5:
That's a good question, but i think it won't be easy because with WEC systems you depend on a platform SDK which will be delivered from a systembuilder. 
Furthermore It is necessary to have Visual Studio 2008 for WEC7 :-/

arch/wec7/ua_architecture.h
```diff
@@ -0,0 +1,130 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information.
+ *
+ *    Copyright 2018 (c) Stephan
```
HolgerJeromin: Perhaps it would be best to describe which Stephan this wrote :)

HolgerJeromin:
> Furthermore It is necessary to have Visual Studio 2008 for WEC7 :-/

VS2008 is [available in CI](https://www.appveyor.com/docs/build-environment/).

arch/ua_architecture_definitions.h
```diff
@@ -17,6 +17,9 @@
 #include <stddef.h>
 
 /* Include stdint.h and stdbool.h or workaround for older Visual Studios */
+#ifdef UNDER_CE
+# include "stdint.h"
```
Pro: You can move this include into `arch/wec7/ua_architecture.h`

arch/ua_network_tcp.c
```diff
@@ -101,7 +101,9 @@ connection_recv(UA_Connection *connection, UA_ByteString *response,
         if(resultsize == -1) {
             /* The call to select was interrupted manually. Act as if it timed
              * out */
+#ifndef UNDER_CE /* What shall we do here? */
             if(errno == EINTR)
```
Pro: This should be `UA_ERRNO` instead of `errno`. I will change that in another PR (#2042). Then you can change `UA_ERRNO` in your architecture definition and remove this change (rebase after the other PR is merged)

arch/wec7/ua_architecture.h
Pro: Jep, please include your full name, or your github username in parenthesis.

arch/wec7/ua_architecture.h
```diff
@@ -0,0 +1,130 @@
...
+#define UA_AGAIN WSAEWOULDBLOCK
+#define UA_EAGAIN EAGAIN
+#define UA_WOULDBLOCK WSAEWOULDBLOCK
+#define UA_ERR_CONNECTION_PROGRESS WSAEWOULDBLOCK
+
+#define UA_fd_set(fd, fds) FD_SET((UA_SOCKET)fd, fds)
+#define UA_fd_isset(fd, fds) FD_ISSET((UA_SOCKET)fd, fds)
+
+#ifdef UNDER_CE
+# define errno
```
Pro: This should then be `#define UA_ERRNO`. I guess it should be `#define UA_ERRNO WSAGetLastError()`
You would need to test that...

arch/wec7/ua_clock.c
```diff
@@ -0,0 +1,63 @@
...
+    GetSystemTime(&st);
+    SystemTimeToFileTime(&st, &ft);
+    ULARGE_INTEGER ul;
+    ul.LowPart = ft.dwLowDateTime;
+    ul.HighPart = ft.dwHighDateTime;
+    return (UA_DateTime)ul.QuadPart;
+}
+#ifndef UNDER_CE
+// The problem here is that not all time functions are implemented in Windows CE.
+UA_Int64 UA_DateTime_localTimeUtcOffset(void) {
```
Pro: As a workaround you could just return 0 for the offset and mention it in a comment that the local time offset is always 0 here.

plugins/ua_log_stdout.c
```diff
@@ -50,16 +50,24 @@ __attribute__((__format__(__printf__, 3 , 0)))
...
+#endif
 
 #ifdef UA_ENABLE_MULTITHREADING
     pthread_mutex_lock(&printf_mutex);
 #endif
 
     printf("[%04u-%02u-%02u %02u:%02u:%02u.%03u (UTC%+05d)] %s/%s" ANSI_COLOR_RESET "\t",
            dts.year, dts.month, dts.day, dts.hour, dts.min, dts.sec, dts.milliSec,
+#ifdef UNDER_CE
+           logCategoryNames[category]);
```
Pro: This will not be correct since you would also need to change the format string. Is there anything against leaving the original code here and just ignore the fact that in WCE we do not have offset (i.e. it will be 0)

src/server/ua_services_discovery.c
```diff
@@ -379,8 +379,13 @@ process_RegisterServer(UA_Server *server, UA_Session *session,
             return;
         }
         memcpy(filePath, requestServer->semaphoreFilePath.data, requestServer->semaphoreFilePath.length );
+#ifdef UNDER_CE
```
Pro: In the PR #2042 I also added a new define for `UA_fileExists`.

You can redefine that one in your `ua_architecture.h` file with something like:

```
#define UA_fileExists(X) ({ \
                FILE *fp = fopen(X,"rb"); \
                bool exists = (fp==NULL); \
                if(fp) \
                    fclose(fp); \
                exists; \
})
```

Pro:
Oh, and you also need to sign our CLA by logging in here and following the steps:
https://cla-assistant.io/open62541/open62541?pullRequest=2026

src/server/ua_services_discovery.c
crafti5: @Pro unfortunately this won't compile with my VS2017. It always says "undeclared identifier"

crafti5: I have also tested this but this wont work either. :-/ The statement expression seems to be a GCC feature which isn't supported in Visual Studio compiler.

Pro: then as an alternative you can use your own function to your main architecture header:

```
UA_Boolean UA_fileExists(const char* path) {
                FILE *fp = fopen(X,"rb");
                UA_Boolean exists = (fp==NULL);
                if(fp)
                    fclose(fp);
                return exists;
}
```
That should work too and it will automatically disable the other define since that one is checking if `UA_fileExists` is already defined or not

crafti5: okay i will try this :)

arch/ua_network_tcp.c
```diff
@@ -608,7 +608,7 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
 
     if(clientsockfd == UA_INVALID_SOCKET) {
             UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Could not create client socket: %s", strerror(UA_ERRNO));
+                            "Could not create client socket: %s", UA_ERRNO);
```
Pro: This should still be strerror. If that's not supported in WEC, then redefine strerror in your architecture file

arch/ua_network_tcp.c
```diff
@@ -628,7 +628,7 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
     if ((error == -1) && (UA_ERRNO != UA_ERR_CONNECTION_PROGRESS)) {
             ClientNetworkLayerTCP_close(connection);
             UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Connection to  failed with error: %s", strerror(UA_ERRNO));
```
Pro: Same here

crafti5:
@Pro I have updated this review ;)

Pro:
Cool! Looks very clean now!

The only thing is the failing Travis CI builds. This looks like it has a problem with some unrelated external script.

crafti5:
Without this script we can not merge this pull request, is it right?

Pro:
Yes, we first need to find out, why travis is failing. Only when all is green we can merge.

If you want to have a look at it, that would be nice! It's not related to your changes, but some package which was updated and now probably all builds fail.

Pro:
Probably travis is fixed in my new PR. Closing this in favour of #2059.

PR #2037, opened 2018-09-02
Move template back

Pro:
Github does not see the issue template in the tools folder


Pro:
The advantage of using the `.github` folder is that it is at least hidden under linux. And it is immediately clear that the files in that folder are for github. If we put the `issue_template.md` file into the root folder, it may be misleading.

Pro:
Ok, I moved it back to the root folder to be consistent with the rest

PR #2047, opened 2018-09-07
UpdateSecurityPolicyCertificate

ckmk14:
This enables the stack to change/update the ApplicationInstanceCertificate (plus private key) at runtime.


jpfr:
Great! Merged.

It is a bit awkward that the old certificate has to be known to make the update.
The user could directly point to the SecurityPolicy he wants to update.

But let's first see how the actual usage patterns in the GDS are.

PR #2055, opened 2018-09-13
[Review] add UADP layer 2 plugin

hthiery:
Version of PubSub Ethernet/Layer2 transport layer.


jpfr:
Thanks for the contribution!
Looking good. 👍 

Since the transport layer is Linux-only, Windows builds will break.
Is it possible to move the functionality to the /arch folder where we keep platform-specific code?
And enable compilation only with PubSub enabled and on Linux.

hthiery:
what do you mean we should move? the whole plugin or only parts inside the plugin that are platform specific?

hthiery:
The reason for two licenses is just copy/past from "plugins/ua_network_pubsub_udp.c/h".

hthiery:
I will copy the two files "plugins/ua_network_pubsub_ethernet.c/h" to "arch/posix/ua_network_pubsub_ethernet.c/h" ... right?

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+ */
+
+static UA_StatusCode
+UA_parseEndpointUrlEthernet(const UA_String *endpointUrl, UA_PubSubChannelDataEth *channelDataEth) {
+    /* related to OPC-UA specification the URL for ethernet should be:
+     *    opc.eth://<host>[:<vid>[.<pcp>]]
+     * with <host> .. MAC address, IP address or name
+     *      <vid>  .. VLAN ID
+     *      <pcp>  .. priority
+     */
```
mwalle: Describe what the <host> looks like. According to OPC-UA Part 14:

"The host is a MAC address, an IP address or a registered name like a hostname. The format 
of a MAC address is six groups of hexadecimal digits, separated by hyphens (e.g. 01-23-45-
67-89-ab). A system may also accept hostnames and/or IP addresses if it provides means to 
resolve it to a MAC address (e.g. DNS and Reverse-ARP)."

I guess resolving any hostname is out of scope for now..

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+    char urlStr[64];
+    char *ptr, *ptrHost;
+    char *ptrVid = NULL;
+    char *ptrPrio = NULL;
+    int i;
+
+    if (endpointUrl->length > sizeof(urlStr)) {
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
+    sprintf(urlStr, "%.*s", (int) endpointUrl->length, endpointUrl->data);
```
mwalle: I would avoid copying strings. See below.

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+            ptrPrio = ptr + 1;
+        }
+    }
+    snprintf(channelDataEth->destinationMac, MAX_MAC_ADDR, "%s", ptrHost);
+    for (i = 0; channelDataEth->destinationMac[i] != 0; i++) {
+        if (channelDataEth->destinationMac[i] == ' ')
+            channelDataEth->destinationMac[i] = ':';
+    }
+    channelDataEth->vid  = (UA_UInt16) ((ptrVid == NULL) ? 0 : atoi(ptrVid));
+    channelDataEth->prio = (UA_Byte) ((ptrPrio == NULL) ? 0 : atoi(ptrPrio));
```
mwalle: use strtoul() here, first because it supports endptr and second it supports error codes

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+        /* check for VLAN priority */
+        ptr = strchr(ptrVid, '.');
+        if (ptr != NULL) {
+            *ptr = 0;
+            ptrPrio = ptr + 1;
+        }
+    }
+    snprintf(channelDataEth->destinationMac, MAX_MAC_ADDR, "%s", ptrHost);
+    for (i = 0; channelDataEth->destinationMac[i] != 0; i++) {
+        if (channelDataEth->destinationMac[i] == ' ')
```
mwalle: shouild be '-' not ' ', see <host> description above

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+            channelDataEth->destinationMac[i] = ':';
+    }
```
mwalle: to avoid copying, maybe we can use sscanf()? or a for loop with strtoul().

instead of storing the string representation use UA_Byte[ETH_ALEN]

jpfr:
Hmm. The current PubSub UDP layer is also in /plugins as well.

Let's simplify as follows:
Just make sure the continuous integration (travis, appveyor) is all green.
We then merge as-is.

Moving into /arch becomes relevant afterwards if and when a second architecture besides Linux is added.

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+    struct ether_header *ethHdr;
+    ssize_t ethHdrLen;
+    UA_UInt16 etherType;
+
+    ethHdr = (struct ether_header*) buffer;
+    ethHdrLen = sizeof(*ethHdr);
+    etherType = ntohs(ethHdr->ether_type);
+    if (etherType == ETHERTYPE_VLAN) {
+        ethHdrLen += 4;
+        etherType = *((UA_UInt16 *) (ethHdr + ethHdrLen + 2));
```
mwalle: ntohs() missing?

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+    newChannel = (UA_PubSubChannel*)UA_calloc(1, sizeof(UA_PubSubChannel));
+    if (!newChannel) {
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "PubSub Connection creation failed. Out of memory.");
+        UA_free(channelDataEth);
+        return NULL;
+    }
+
+    /* open a RAW socket */
+    int sockFd;
+    if ((sockFd = UA_socket(PF_PACKET, SOCK_RAW, htons(ETH_P_ALL))) < 0) {
```
mwalle: by using ETH_P_ALL all packets will be received regardless of their ethertype.. i guess this is not the intended behaviour.

I see that there are two ethertypes needed ETHERTYPE_UADP and ETHERTYPE_VLAN. Anybody has any idea how to solve this? Is it possible to bind one raw socket to two ethertypes? use two sockets? use ETH_P_ALL with a custom BPF? All of this is really linux dependant.

Btw i don't know if receiving raw packets with etherype=VLAN is working with network cards which supports hardware vlan filtering, so this should be tested.

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+        UA_close(sockFd);
+        UA_free(channelDataEth);
+        UA_free(newChannel);
+        return NULL;
+    }
+
+    struct sockaddr_ll sll;
+    sll.sll_family = AF_PACKET;
+    sll.sll_ifindex = ifopts.ifr_ifindex;
+    sll.sll_protocol = htons(ETH_P_ALL);
```
mwalle: same here

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+    ethHdrLen = sizeof(*ethHdr);
+    etherType = ntohs(ethHdr->ether_type);
+    if (etherType == ETHERTYPE_VLAN) {
+        ethHdrLen += 4;
+        etherType = *((UA_UInt16 *) (ethHdr + ethHdrLen + 2));
+    }
+    if (etherType != ETHERTYPE_UADP) {
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
+            "PubSub connection receive failed. Not UADP format.");
+        return UA_STATUSCODE_BADINTERNALERROR;
```
mwalle: With ETH_P_ALL any messages could be received here, without being an error. Statuscode should be something like SKIP or GOOD_NO_MSG.

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,422 @@
...
+/**
+ * Close channel and free the channel data.
+ *
+ * @return UA_STATUSCODE_GOOD if success
+ */
+static UA_StatusCode
+UA_PubSubChannelEthernet_close(UA_PubSubChannel *channel) {
+    UA_free(channel->handle);
+    UA_free(channel);
+    return UA_STATUSCODE_GOOD;
```
mwalle: UA_close(sock) missing?

plugins/ua_network_pubsub_ethernet.c
jpfr: OPC UA PubSub uses a custom ethertype.

hthiery:
> I will copy the two files "plugins/ua_network_pubsub_ethernet.c/h" to "arch/posix/ua_network_pubsub_ethernet.c/h" ... right?
> 
> or do we need another folder "linux"

@jpfr: do you have any suggestions how to handle the linux specific code?

 - what CMakeLists.txt I have to add the files?
 - how to check about architecture?
 - where to copy the ua_networkd_pubsub_ethernet.* specific files?

Any hint where I have to take a look on and orientate?

hthiery:
@jpfr  any idea why this travis builds fail?
`
Exception occurred:
  File "/home/travis/.local/lib/python2.7/site-packages/sphinx/highlighting.py", line 26, in <module>
    from sphinx.ext import doctest
SyntaxError: unqualified exec is not allowed in function 'run' it contains a nested function with free variables (doctest.py, line 97)
The full traceback has been saved in /tmp/sphinx-err-EkkdML.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make[3]: *** [doc/CMakeFiles/doc] Error 2
make[2]: *** [doc/CMakeFiles/doc.dir/all] Error 2
make[1]: *** [doc/CMakeFiles/doc.dir/rule] Error 2
make: *** [doc] Error 2
`

hthiery:
@jpfr any Idea why the windows build fails?

plugins/ua_network_pubsub_ethernet.c
```diff
@@ -0,0 +1,437 @@
...
+    ssize_t ethHdrLen;
+    UA_UInt16 etherType;
+
+    ethHdr = (struct ether_header*) buffer;
+    ethHdrLen = sizeof(*ethHdr);
+    etherType = ntohs(ethHdr->ether_type);
+    if (etherType == ETHERTYPE_VLAN) {
+        ethHdrLen += 4;
+        etherType = ntohs(*((UA_UInt16 *) (ethHdr + ethHdrLen + 2)));
+    }
```
mwalle: so UDAP packets with any vlan id are accepted.. vlan id should be checked too. But as pointed out above.. this receive handler will then be called for **any** vlan packet.

Where does the receiving interface comes from? Shouldn't it be within the opc-eth:// uri? Where does one know behind which interface the target mac address is reachable?

mwalle:
@jpfr please don't merge yet. there will be a major rework in the receiving code to resolve the issues above.

jpfr:
Ok. We will hold until the PR prefix is changed to [Review].

mwalle:
> @jpfr: do you have any suggestions how to handle the linux specific code?
> 
> * what CMakeLists.txt I have to add the files?
> * how to check about architecture?
> * where to copy the ua_networkd_pubsub_ethernet.* specific files?
> 
> Any hint where I have to take a look on and orientate?

FYI. I've introduced another `UA_ENABLE_PUBSUB_ETH_UADP` option, which enabled the plugin and the tutorial example code. If it is enabled without CMAKE_SYSTEM being Linux, there will be an error message.

jpfr:
> I've introduced another UA_ENABLE_PUBSUB_ETH_UADP option

👍

mwalle:
@jpfr, one question which came up during discussions: does one PubSubConnection receives exactly the frames/packets for its own multicast group. Eg. if the sender registers to `opc.eth://01-00-00-00-00-01`, the receiver will only receive frames for this particular (multicast) hardware address. Or is it supposed to receive frames destined for other mutlicast group (or its own unicast address)?

hthiery:
@jpfr, any hints regarding @mwalle's question?

I would propose to merge th changes and do further changes in another pull-request.

hthiery:
Is there more required than enabling editing for maintainers?

jpfr:
@mwalle 

The packets destined for your unicast address can also be received.
If you filter them out internally is not defined by the spec. At least I did not see it mentioned.

jpfr:
The PR was split as described above and the ARGV build flag was removed.

jpfr:
Merged. Thanks @hthiery for the contribution!

mwalle:
> The packets destined for your unicast address can also be received.
> If you filter them out internally is not defined by the spec. At least I did not see it mentioned.

@jpfr ok, but i'm not sure, if we should do that. There might be frames which are not supposed to be received by that channel. Think of two processes running on the same machine, each one registers one multicast group (same with its unicast address). Without filtering, each process would receive all multicast frames.
If one process would like to register for multiple multicast groups, multiple channels would have to be created.

PR #2062, opened 2018-09-17
Correct Event validity check

### LINKED ISSUE #2027: Event validity is not checked properly when applying event filter
Hello everyone!

I have noticed that UaExpert is specifing the BaseEventType in the SimpleAttributeOperand of all Clauses (properties/ Fields of the EventType) except for the Clause "CondtionId", which has the NodeId of the ConditionTypeNode and is always present in the selectClauses array set by UaExpert (from part 9, which is not implemented yet in open62541. Note: "CondtionId" is not explicitly modelled as component of the ConditionType, but should be provided by the Server. Please refer to Part 9, 5.5.2, Page 20, Table 10). UaExpert also specifies the BaseEventType for Clauses of the ConditionType and its Subtypes as well as for the Clauses of the AuditEventType and its Subtypes. That behaviour of UaExpert is complaint to the part 4 specs. (Please refer to Part 4, 7.17.3, Page 133)

So, when using UaExpert, it is true for all Clauses except "CondtionId", that the condition in the following code snippet will alway be evaluated as False and thus the SimpleAttributeOperand will be resolved. (no problem until now)
https://github.com/open62541/open62541/blob/3b4f84f248f5e14999da6afd0179979b51898fb0/src/server/ua_subscription_events.c#L250

However, for the "CondtionId" Clause, or when using other UA Clients (OPC Foundation Demo Client and Softing), where the real EventTypes are specified in the SimpleAttributeOperand, the previous code snippet will be evaluated as True and thus the other check will be evaluated:
https://github.com/open62541/open62541/blob/3b4f84f248f5e14999da6afd0179979b51898fb0/src/server/ua_subscription_events.c#L251

Problems:
1) Here, I think that the actual selectClauses[ i ] should be passed instead the first selectClauses[ 0 ]. The typeId of selectClauses[ 0 ] depends on the first Clause set by the client filter (However, when using UaExpert, the typeId of the first Clause is always BaseEventType).

2) The function
https://github.com/open62541/open62541/blob/3b4f84f248f5e14999da6afd0179979b51898fb0/src/server/ua_subscription_events.c#L138
checks whether the EventType Property is present, but then checks whether the **EventType Property NodeId** is a SubType of the passed "validEventParent" (which is a BaseEventType or a subtype of it), and this will be always evaluated as False and thus, the SimpleAttributeOperand won't be resolved. With this manner, the values of the fields of the subTypes of BaseEventType won't show up in the Event View of UA Clients other than UaExpert! 
I think here we have to get the **value of the EventType Property** (which is the NodeId of the EventType) and check whether it is a subType of the BaseEventType. However, we should check first that the EventType is not a subType of ConditionType (not supported yet) so "CondtionId" Clause won't be resolved, and thus a "CondtionId" Field won't show up in the normal events.

@aDogCalledSpot, what do you think? should I make a PR to change/correct that behaviour? 

Best regards

### LINKED ISSUE #2038: One event triggers two MonitoredItems
My client creates two MonitoredItems (different callbacks) targetting the same object (`i=42`). My server then triggers one of the monitored events (`i=1337`). open62541 executes both callbacks on the client-side indicating that both MonitoredItems received an event. I had expected that open62541 only executes the callback of the "correct" MonitoredItem.

1. Are my ids and select clauses correct? (I've not used `BaseEventType` because I want to distinguish the events.)
2. Am I missing something?
3. If everything works as expected, how is it possible to distinguish between two received events on the client-side?

Thanks in advance!

See below for code examples (omitted error handling, no errors occuring). This may be related to #2003. I'm using Visual Studio 2015 on Windows 10 with open62541 at af95b006494a23eff648bc40ff1cceaf3c15598a. /cc @aDogCalledSpot @Pro 

# Triggering on the server-side

```cpp
// assuming EventType with i=1337 and ObjectType with i=42
UA_NodeId nodeIdEvent;
UA_Server_createEvent(server, UA_NODEID_NUMERIC(ns, 1337), &nodeIdEvent);
UA_DateTime now = UA_DateTime_now();
UA_Server_writeObjectProperty_scalar(server, nodeIdEvent, UA_QUALIFIEDNAME_STATIC(0, "Time"), &now, &UA_TYPES[UA_TYPES_DATETIME]);
UA_String string = UA_STRING_STATIC("Hello World!");
UA_Server_writeObjectProperty_scalar(server, nodeIdEvent, UA_QUALIFIEDNAME_STATIC(ns, "Text"), &string, &UA_TYPES[UA_TYPES_STRING]);
UA_Server_triggerEvent(server, nodeIdEvent, UA_NODEID_NUMERIC(ns, 42), NULL, UA_TRUE);
```

# Create MonitoredItem that should trigger

```cpp
// assuming EventType with i=1337 and ObjectType with i=42
UA_MonitoredItemCreateRequest request = UA_MonitoredItemCreateRequest_default(UA_NODEID_NUMERIC(ns, 42));
request.itemToMonitor.attributeId = UA_ATTRIBUTEID_EVENTNOTIFIER;
UA_EventFilter filter;
UA_EventFilter_init(&filter);
filter.selectClauses = (UA_SimpleAttributeOperand *)UA_Array_new(1, &UA_TYPES[UA_TYPES_SIMPLEATTRIBUTEOPERAND]);

UA_QualifiedName Text = UA_QUALIFIEDNAME_ALLOC(ns, "Text");
UA_SimpleAttributeOperand_init(&filter.selectClauses[0]);
filter.selectClauses[0].typeDefinitionId = UA_NODEID_NUMERIC(ns, 1337); // NOTICE: triggered id
filter.selectClauses[0].browsePathSize = 1;
filter.selectClauses[0].browsePath = &Text;
filter.selectClauses[0].attributeId = UA_ATTRIBUTEID_VALUE;

// ... 3 more UA_SimpleAttributeOperand

filter.selectClausesSize = 1;
request.requestedParameters.filter.encoding = UA_EXTENSIONOBJECT_DECODED;
request.requestedParameters.filter.content.decoded.data = &filter;
request.requestedParameters.filter.content.decoded.type = &UA_TYPES[UA_TYPES_EVENTFILTER];
UA_Client_MonitoredItems_createEvent(client, subId, UA_TIMESTAMPSTORETURN_SOURCE, request, NULL, &callbackShouldTrigger, NULL);
```

# Create MonitoredItem that should NOT trigger

```cpp
// assuming other EventType with i=2337 and same ObjectType with i=42
UA_MonitoredItemCreateRequest request = UA_MonitoredItemCreateRequest_default(UA_NODEID_NUMERIC(ns, 42));
request.itemToMonitor.attributeId = UA_ATTRIBUTEID_EVENTNOTIFIER;
UA_EventFilter filter;
UA_EventFilter_init(&filter);
filter.selectClauses = (UA_SimpleAttributeOperand *)UA_Array_new(4, &UA_TYPES[UA_TYPES_SIMPLEATTRIBUTEOPERAND]);

UA_QualifiedName Status1 = UA_QUALIFIEDNAME_ALLOC(ns, "Status1");
UA_SimpleAttributeOperand_init(&filter.selectClauses[0]);
filter.selectClauses[0].typeDefinitionId = UA_NODEID_NUMERIC(ns, 2337); // NOTICE: different id (not triggered)
filter.selectClauses[0].browsePathSize = 1;
filter.selectClauses[0].browsePath = &Status1;
filter.selectClauses[0].attributeId = UA_ATTRIBUTEID_VALUE;

// ... 3 more UA_SimpleAttributeOperand

filter.selectClausesSize = 4;
request.requestedParameters.filter.encoding = UA_EXTENSIONOBJECT_DECODED;
request.requestedParameters.filter.content.decoded.data = &filter;
request.requestedParameters.filter.content.decoded.type = &UA_TYPES[UA_TYPES_EVENTFILTER];
UA_Client_MonitoredItems_createEvent(client, subId, UA_TIMESTAMPSTORETURN_SOURCE, request, NULL, &callbackShouldNotTrigger, NULL);
```

Pro:
Make sure the EventType is not a Subtype of CondtionType (Part 9 not supported yet).
1) The check should ignore the "ConditionId" Clause set by UaExpert and ignore all Events which are Conditions. (However, if a Condition Event was created and UaExpert is used, then all the Fields of that Event except the "ConditionId" will be presenet in the Event View) -> The User should not use Events which are Conditions in that case!
2) The second check is to make sure the Event has a TypeDefinition of an BaseEventType SubType.

Fixes #2027 and fixes #2038 

Cherry-picked from @sqadasi

### DIFF
--- a/src/server/ua_subscription_events.c
+++ b/src/server/ua_subscription_events.c
@@ -143,9 +143,36 @@ isValidEvent(UA_Server *server, const UA_NodeId *validEventParent, const UA_Node
         UA_BrowsePathResult_deleteMembers(&bpr);
         return UA_FALSE;
     }
+    
+	/* get the EventType Property Node */
+    const UA_Node* tLeafNode = UA_Nodestore_get(server, &bpr.targets[0].targetId.nodeId);
+    UA_Variant tOutVariant;
+    /* read the Value of EventType Property Node (the Value should be a NodeId) */
+    UA_Server_readValue(server, tLeafNode->nodeId, &tOutVariant);
+    UA_NodeId tEventType = *((UA_NodeId*)tOutVariant.data);
+
+    /* Make sure the EventType is not a Subtype of CondtionType
+     * First check for filter set using UaExpert
+     * (ConditionId Clause won't be present in Events, which are not Conditions)
+     * Second check for Events which are Conditions or Alarms (Part 9 not supported yet) */
+    UA_NodeId conditionTypeId = UA_NODEID_NUMERIC(0, UA_NS0ID_CONDITIONTYPE);
     UA_NodeId hasSubtypeId = UA_NODEID_NUMERIC(0, UA_NS0ID_HASSUBTYPE);
-    UA_Boolean tmp = isNodeInTree(&server->config.nodestore, &bpr.targets[0].targetId.nodeId,
-                                  validEventParent, &hasSubtypeId, 1);
+    if(UA_NodeId_equal(validEventParent, &conditionTypeId) ||
+       isNodeInTree(&server->config.nodestore, &tEventType, 
+    		         &conditionTypeId, &hasSubtypeId, 1)){
+        UA_LOG_ERROR(server->config.logger, UA_LOGCATEGORY_USERLAND,
+    	         "Alarms and Conditions are not supported yet!");
+        UA_Nodestore_release(server, (const UA_Node *) tLeafNode);
+        UA_BrowsePathResult_deleteMembers(&bpr);
+        return UA_FALSE;
+    }
+
+    /* check whether Valid Event other than Conditions */
+    UA_NodeId baseEventTypeId = UA_NODEID_NUMERIC(0, UA_NS0ID_BASEEVENTTYPE);
+    UA_Boolean tmp = isNodeInTree(&server->config.nodestore, &tEventType,
+                                  &baseEventTypeId, &hasSubtypeId, 1);
+
+    UA_Nodestore_release(server, (const UA_Node *) tLeafNode);
     UA_BrowsePathResult_deleteMembers(&bpr);
     return tmp;
 }
@@ -248,7 +275,7 @@ UA_Server_filterEvent(UA_Server *server, UA_Session *session,
     /* iterate over the selectClauses */
     for(size_t i = 0; i < filter->selectClausesSize; i++) {
         if(!UA_NodeId_equal(&filter->selectClauses[i].typeDefinitionId, &baseEventTypeId) &&
-           !isValidEvent(server, &filter->selectClauses[0].typeDefinitionId, eventNode)) {
+           !isValidEvent(server, &filter->selectClauses[i].typeDefinitionId, eventNode)) {
             UA_Variant_init(&notification->fields.eventFields[i]);
             /* EventFilterResult currently isn't being used
             notification->result.selectClauseResults[i] = UA_STATUSCODE_BADTYPEDEFINITIONINVALID; */
--- a/src/server/ua_subscription_events.c
+++ b/src/server/ua_subscription_events.c
@@ -141,14 +141,21 @@ isValidEvent(UA_Server *server, const UA_NodeId *validEventParent, const UA_Node
     UA_BrowsePathResult bpr = UA_Server_browseSimplifiedBrowsePath(server, *eventId, 1, &findName);
     if(bpr.statusCode != UA_STATUSCODE_GOOD || bpr.targetsSize < 1) {
         UA_BrowsePathResult_deleteMembers(&bpr);
-        return UA_FALSE;
+        return false;
     }
     
-	/* get the EventType Property Node */
+	/* Get the EventType Property Node */
     UA_Variant tOutVariant;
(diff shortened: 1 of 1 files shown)


jpfr:
The fix itself is good!
Some things about the code should be fixed though.

- `UA_Server_readValue` does not require that the node is "checked out" first.
- `tOutvariant` is not cleaned up. Leaks memory.
- The content of `tOutVariant` should be checked to contain a NodeId scalar. Otherwise the code can be crashed by writing a different type of value.
- `tEventType` should be a const pointer to NodeId.
- The variable `tmp` could have a more "speaking" name

Did you review internally @Pro?

Should be all easy to address. Let's merge this soon!

Pro:
I only cherry picked but did not review yet. Maybe @aDogCalledSpot can review, otherwise @jpfr feel free to update the PR.
Since I'm offline for the next 2 weeks I won't be able to update the PR.

aDogCalledSpot:
I've fixed the things mentioned here but do not have push access to this branch. Should I create a new pull request or can my branch be merged otherwise?

jpfr:
@aDogCalledSpot I added you as a collaborator to the repo.
You should be able to (force) push to the branch.

PR #2064, opened 2018-09-18
[Review] Fix/cleanup unsafe c names

mtenberge:
Apparently, OPC-UA identifiers may contain a wide range of special characters, including quotation marks, dots and spaces.
For example, this is a fragment from an UANodeSet XML exported from Siemens TIA Portal V15:
```
<UAVariable NodeId="ns=3;s=&quot;dbCounters_748&quot;.&quot;AllData&quot;.&quot;Element&quot;[11].&quot;CountData&quot;.&quot;Product&quot;[0]" BrowseName="3:0" ParentNodeId="ns=3;s=&quot;dbCounters_748&quot;.&quot;AllData&quot;.&quot;Element&quot;[11].&quot;CountData&quot;.&quot;Product&quot;" DataType="ns=3;s=DT_&quot;udtCounter_627&quot;" AccessLevel="3">
    <DisplayName>0</DisplayName>
    <References>
        <Reference ReferenceType="HasTypeDefinition">ns=3;s=VT_"udtCounter_627"</Reference>
        <Reference ReferenceType="HasComponent" IsForward="false">ns=3;s="dbCounters_748"."AllData"."Element"[11]."CountData"."Product"</Reference>
    </References>
</UAVariable>
```
When these identifiers end up in the generated C files, the result is no longer valid C code.

This PR tries to address these problems.
The code in tools/nodeset_compiler only requires a few fixes, most of the code was already handling this issue correctly.
The script tools/generate_datatypes.py requires more work. I have chosen to keep the `.name` properties intact, and create separate 'C-safe' name properties on the various classes. I'm not sure if this is your preferred way to solve this issue.

Note: unfortunately, this PR alone is not sufficient to compile the above mentioned UANodeSet file error-free, but it is a start...

### DIFF
--- a/tools/generate_datatypes.py
+++ b/tools/generate_datatypes.py
@@ -67,14 +67,16 @@ def getTypeName(xmlTypeName):
 class StructMember(object):
     def __init__(self, name, memberType, isArray):
         self.name = name
+        self.cSafeName = re.sub(r'[^\w]', '', self.name)
         self.memberType = memberType
         self.isArray = isArray
 
 class Type(object):
     def __init__(self, outname, xml, namespace):
         self.name = xml.get("Name")
+        self.cSafeName = re.sub(r'[^\w]', '', self.name)
         self.ns0 = ("true" if namespace == 0 else "false")
-        self.typeIndex = outname.upper() + "_" + self.name.upper()
+        self.typeIndex = outname.upper() + "_" + self.cSafeName.upper()
         self.outname = outname
         self.description = ""
         self.pointerfree = "false"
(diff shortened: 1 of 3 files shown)


Pro:
Thanks for the contribution!

Some comments:
IMHO it would make more sense to use the similar method in `generate_datatypes.py` as we are already using in the namespace compiler: `makeCLiteral`. Then use this makeCLiteral method wherever you are currently using the cSafeName.
This avoids duplication of the string and using two variables may lead to inconsistencies.

For the namespace compiler, you can also just change the `makeCLiteral` method to include your additional escaping regex? And then also use it in `backend_open62541_nodes.py`

mtenberge:
Using makeCLiteral instead of cSafeName seems fine, I'll modify that.

As for adding the regex to the existing makeCLiteral: I actually started exactly that way ;-) However, it lead to a number of new problems, because that function is used in many places, whereas only a string in an (Extended)NodeID needs to have its quotes escaped. I'll take a look into that again, because it doesn't sound logical to me...

mtenberge:
It turns out the function `splitStringLiterals` also escapes any existing quotes it finds, whether they are already escaped or not.
In 4 cases, `splitStringLiterals` is used after `makeCLiteral`. If the quotes would be escaped in `makeCLiteral`, they will erroneously be double-escaped by `splitStringLiterals`.

In the remaining 2 cases (in `generateNodeIdCode` and in `generateExpandedNodeIdCode`), `makeCLiteral` is used without `splitStringLiterals`.

I think I can safely add the same the regex to both `makeCLiteral` and `splitStringLiterals`, as it only matches unescaped quotes. Then all uses of either or both functions are properly escaped.

jpfr:
Looks good to me.
`makeCLiteral` and `makeCIdentifier` are reused between the nodeset compiler and the datatype generation.
But that is okay, as we don't want to integrate the two too much.

Any comments @Pro ?
Other I'd squash-merge this as-is.

PR #2065, opened 2018-09-18
Get server certificate of create session rsp from endpoint

jonasgreen88:
The server certificate included in the Create session response have
been retrieved from the secure channel that the session is created 
on. This works if the security policy of the secure channel != security
policy none.

If the secure channel is not secured (security policy none) but the
security policy of the UserIdentityToken != security policy none 
(e.g. encrypted password when using UserIdentityToken Username_1),
the certificate of the endpoint must be present in the Create session 
response. In this case it is not possible to retrieve it from the channel,
as the channel is not secured and has no reference to the certificate of 
the device.

A common solution that works in both cases is to retrieve the
certificate from the endpoint of the server that the client has
connected to.

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -162,10 +162,21 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
     response->responseHeader.serviceResult |=
         UA_String_copy(&request->sessionName, &newSession->sessionName);
 
+    UA_ByteString_init(&response->serverCertificate);
+
     if(server->config.endpointsSize > 0)
-        response->responseHeader.serviceResult |=
-            UA_ByteString_copy(&channel->securityPolicy->localCertificate,
-                               &response->serverCertificate);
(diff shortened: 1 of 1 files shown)


jonasgreen88:
We are working on adding the possibility to use security policies of UserIdentityTokens when authenticating a session. The CTT failed our device as it didn't provide the server certificate in the create session response. This solution made the CTT to continue further in our testing and all regression tests of this repository still pass (without knowing if the unit tests verify this field). Would be nice if someone can run CTT with security policy of the secure channel != security policy none to verify the fix for that case.

We are unsure if this is the best way to find the currently used endpoint (matching security mode, security policy uri and endpoint url). The solution is inspired from how CTT does it. If there is a better way to do it in this stack we can change method.

jpfr:
Thanks for the contribution!

Eventually we might have a pointer from SecureChannel/Session to an endpoint.
For now lets keep this solution.

PR #2066, opened 2018-09-18
fix issue with freeing transport layer memory

hthiery:
- Added helper function to reallocate memory.
 - Added unit test

Signed-off-by: Heiko Thiery <heiko.thiery@kontron.com>

### DIFF
--- a/include/ua_plugin_pubsub.h
+++ b/include/ua_plugin_pubsub.h
@@ -85,6 +85,11 @@ typedef struct {
     UA_PubSubChannel * (*createPubSubChannel)(UA_PubSubConnectionConfig *connectionConfig);
 } UA_PubSubTransportLayer;
 
+
+UA_StatusCode
+UA_Server_addPubSubTransportLayer(UA_ServerConfig *config,
+        UA_PubSubTransportLayer *pubsubTransportLayer);
+
 #endif /* UA_ENABLE_PUBSUB */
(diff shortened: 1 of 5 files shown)


hthiery:
> We could rename this to `UA_ServerConfig_addPubSubTransportLayer`.
Done

> And make sure to mention in the docs that this is not meant to be done when the server has been started already.
Where should I add this?

hthiery:
> What about the `transportProfileUri` contained in the `UA_PubSubTransportLayer`?
> The current code assumes that the profile string is stack-allocated. This should be added to the documentation as well.

Here I do not fully understand what you mean.

Do you mean that the structue `UA_PubSubTransportLayer` holds an element  `transportProfileUri` that is just a pointer. And it is not clear where the data is allocated? In the current code (`plugins/ua_network_pubsub_udp.c`) that returns a `UA_PubSubTransportLayer` element the used string is const and should be allocated on the heap.

jpfr:
No. In the current implementation, the profile string is statically allocated (in ROM).
https://github.com/open62541/open62541/blob/master/plugins/ua_network_pubsub_udp.c#L454

Otherwise it would have to be freed in the cleanup code that is modified by this PR.
I'm fine either way. But we should document whether the profile string is freed during cleanup.

jpfr:

PR #2071, opened 2018-09-18
Process normal services before initalizing the message context

jonasgreen88:
To support network layers that only provides 1 send buffer only one
message context can be initialized at once. Some services, e.g.
Delete Subscriptions, also initializes message contexts when they
are processed. To only have 1 message context active concurrently,
the normal services are now processed before the message context are
initialized.

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
  *    Copyright 2014-2017 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
  *    Copyright 2014-2016 (c) Sten Grüner
@@ -559,9 +559,7 @@ processMSG(UA_Server *server, UA_SecureChannel *channel,
      * layers only providing one send buffer, only one message context can be
      * initialized concurrently. */
     if(serviceType == UA_SERVICETYPE_NORMAL)
-    {
         service(server, session, request, response);
-    }
 
     /* Start the message */
     UA_NodeId typeId = UA_NODEID_NUMERIC(0, responseType->binaryEncodingId);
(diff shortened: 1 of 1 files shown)


jonasgreen88:
Found this issue today when running CTT. We have developed our network layer to only provide 1 send buffer to the stack. Don't know if that has been a goal to achive that in this stack or if it is just an unintentionally feature so far. Anyway, I think it is a nice feature for memory constrained devices to only have to provide 1 buffer if the stack is executed in a single thread and it would be good if the stack continue to support that.

Feel free to comment the solution, if you have any idea of a different way to acheive the same result, please share.

Infinity95:
I'm currently working on a rework for the network layer. Currently (imho) the network layer is a mess and the server in some cases uses completely different functionality than the client.
I'm working on cleaning up the code and trying to establish a clear boundary between the networking code and the rest, which includes having the same interface for both server and client.
I guess this could be merged as a temporary fix, and I'll check, if it can be solved in an easier way during my reworking process.
@jpfr What do you think?

jonasgreen88:
> I'm currently working on a rework for the network layer. Currently (imho) the network layer is a mess and the server in some cases uses completely different functionality than the client.
> I'm working on cleaning up the code and trying to establish a clear boundary between the networking code and the rest, which includes having the same interface for both server and client.
> I guess this could be merged as a temporary fix, and I'll check, if it can be solved in an easier way during my reworking process.
> @jpfr What do you think?

That sounds good. Would be nice to include this fix in some form to also get it merged to the 0.3 branch. Right now, with our network layer configuration (with only 1 send buffer) we get an issue in CTT when it sends the Delete Subscription service and the OPC UA server shall respond to the publish requests it already has received before it returns the response of the Delete Subscription service.

An alternative for us if you don't like to add this solution is to extend our network layer with 2 send buffers to handle this case.

jpfr:
> Right now, with our network layer configuration (with only 1 send buffer) we get an issue in CTT when it sends the Delete Subscription service and the OPC UA server shall respond to the publish requests it already has received before it returns the response of the Delete Subscription service.

Does the proposed PR solve this issue?
This looks easy enough to backport on 0.3.

jonasgreen88:
> > Right now, with our network layer configuration (with only 1 send buffer) we get an issue in CTT when it sends the Delete Subscription service and the OPC UA server shall respond to the publish requests it already has received before it returns the response of the Delete Subscription service.
> 
> Does the proposed PR solve this issue?
> This looks easy enough to backport on 0.3.

Yes, it solves the issue. We have also run a complete Micro embedded profile test with CTT and have not seen that this change have caused any other issues neither.

jpfr:
Thanks! Merged.

Issue #2072, opened 2018-09-20
Server may crash if event buffer overflows


coder4803:
Description
===========
Server may crash if event buffer overflows (and someone is monitoring those events). This might be duplicate of #2020 .

Reproduction Steps
===========================================
1. Start server event tutorial (tutorial_server_events.exe)
2. Start client event tutorial (tutorial_client_events.exe <address>)
3. Start uaExpert and connect to the server
4. Call generateEvent method multiple times quickly (5-10 times per second)
   --> server crashes

Crash happens in "UA_MonitoredItem_ensureQueueSpace" function on line "TAILQ_INSERT_AFTER(&sub->notificationQueue, del, after_del, globalEntry);"

For some reason queueSize can go to "negative" if overflow happens.
UA_assert(mon->queueSize >= 2); doesn't work catch it here because queueSize is unsigned integer :)

Background Information
===========================================

**open62541 Version:**  a43604a28a9c0c48bc82b90b938a3fb38d97d83a (Cosmetic changes; Replace "extern C" with _UA_BEGIN_DECLS, Sep 9, 2018)

**Other OPC UA SDKs used:** uaExpert

**Operating system:** Windows 10

**Critical issue:** yes

**Logs from server:**
[New Thread 17216.0x2d88]
[New Thread 17216.0x2548]
[New Thread 17216.0xefc]
[New Thread 17216.0x910]
[2018-09-20 11:48:34.268 (UTC+0300)] info/network       TCP network layer listening on opc.tcp://peost-lt:4840/
[2018-09-20 11:48:37.999 (UTC+0300)] info/network       Connection 468 | New connection over TCP from 127.0.0.1
[2018-09-20 11:48:38.001 (UTC+0300)] info/channel       Creating a new SecureChannel
[2018-09-20 11:48:38.001 (UTC+0300)] warn/securitypolicy        No PKI plugin set. Accepting all certificates
[2018-09-20 11:48:38.002 (UTC+0300)] info/channel       Connection 468 | SecureChannel 1 | Opened SecureChannel
[2018-09-20 11:48:38.002 (UTC+0300)] info/session       Try to use Session with token ns=1;g=9dce7b1f-5f76-3f0c-c40c-c05ca19a8968 but is not found
[2018-09-20 11:48:38.173 (UTC+0300)] info/session       Connection 468 | SecureChannel 1 | Session ns=1;g=0b67bae3-d153-6e26-3523-12f15e25b2bb | ActivateSession: Session activated
[...]

PR #2076, opened 2018-09-21
Fix OverflowEvent counting

aDogCalledSpot:
Fixes https://github.com/open62541/open62541/issues/2072.

The problem wasn't an overflow, but an underflow. This was causing the notification queue size to be reduced below 0. Usages of eventOverflows should be correctly found now.

Also, there are generally less overflows being generated since the generation was moved outside of the while loop in UA_MonitoredItem_ensureQueueSpace. The wording in the specification isn't entirely clear here. In Part 4 5.12.1.5 it says that when "events are discarded" (plural), an overflow is generated but it only discards the first event "without discarding any other events". I  believe having only overflowEvent indicate the removal of numerous events sounds more reasonable than generating one for each them - opinions?

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -306,6 +306,16 @@ prepareNotificationMessage(UA_Server *server, UA_Subscription *sub,
             *efl = notification->data.event.fields;
             UA_EventFieldList_init(&notification->data.event.fields);
             efl->clientHandle = mon->clientHandle;
+
+            /* Check if this is an overflowEvent */
+            UA_NodeId overflowBaseId = UA_NODEID_NUMERIC(0, UA_NS0ID_EVENTQUEUEOVERFLOWEVENTTYPE);
+            if(efl->eventFieldsSize == 1 &&
+               efl->eventFields[0].type == &UA_TYPES[UA_TYPES_NODEID] &&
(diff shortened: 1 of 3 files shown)


jpfr:
Great! Looks good.

- I think we can reduce the indendation of the event-code in this change.
- Can you add a unit test for the queue overflow behaviour?

A proper review will then follow.

aDogCalledSpot:
Unfortunately, the bug wasn't deterministic in any way and would only occur randomly under heavy load. I've added a small "stress test" which will generally cause a segfault after 35-40 loops if run on the current master branch but I can't entirely rule out that false positive outcomes may occur.

jpfr:
> Unfortunately, the bug wasn't deterministic in any way and would only occur randomly under heavy load. 

In our unit test framework, we can reproduce everything deterministically.
Because we can manually "fake" the clock.

I will have a look at this. "Mostly works" doesn't cut it at this stage of the project.
The changes are looking good though!

jpfr:
@aDogCalledSpot Your fix is correct. 👍 

I made a few changes to have the functionality more "in one place".
But the functional core of your fix remains.

Can you review the last commit in https://github.com/jpfr/open62541/tree/hotfix/event_queue_crash ?

PR #2077, opened 2018-09-21
Asymetric decrypt func of Basic128Rsa15 did not set rsa padding

jonasgreen88:
The function asym_decrypt_sp_basic128rsa15() did not set any rsa
padding. If some other function of the security policy
was executed, setting a different padding, it made the decrypt
function to not work anymore.
The RSA padding is now set to PKCS-1 v.1.5 with hash ID 0. It equals
the padding the encrypt function of the security policy sets.


jpfr:
Thanks!

PR #2080, opened 2018-09-23
Fix issues uncovered by oss-fuzz

jpfr:


StalderT:
@jpfr is 0.3 also concerned?

jpfr:
No. This is only for master.
Maybe we can add a second branch to oss-fuzz.

Issue #2041, opened 2018-09-04
OPC Alarm 


jhj0955:
Dear whom it may concern,
I want to test the alarm function. Is it implemented?

If so, how can I test it?

Pro:
Duplicate of #1947

We do not yet support Alarms and Conditions, but hope to have it soon

PR #2083, opened 2018-09-24
[Review] - First Implementation of Alarms and Conditions

### LINKED ISSUE #1947: Support for Alarms & Conditions (Part 9)
Description
===========
OPC UA Part 9, 5.5.7 specifies the ConditionRefresh method of the ConditionType object.
This method takes a subscription ID belonging to the client's session as parameter and issues event notifications for retained conditions using the subscription passed as parameter.

Implementing the ConditionRefresh method handler requires the following capabilities:
- Find out if a subscription ID belongs to a session
- Get the monitored items for a subscription and check for matching retained conditions using the event filter
- Insert event notifications into the queue of a subscription

Is there a way to do this using the public server API or is this a feature like the GetMonitoredItems method that should be implemented inside the library?

### LINKED ISSUE #2738: Fire complex event with ObjectTypes
Hello!

I am currently experimenting with Pro's euromap test branch and the server events tutorial:
https://github.com/Pro/open62541/tree/test/euromap/examples/nodeset/euromap

I can successfuly build and run the server_nodeset_euromap project on Windows.
cmake .. -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_AMALGAMATION=NO -DUA_BUILD_EXAMPLES=YES -G "Visual Studio 15 2017"

And I could trigger some simple events. For example:
Create Euromap event "CycleParametersEventType" and set field "BoxGoodPartsCounter": 
image

I used the function UA_Server_writeObjectProperty_scalar as described in the events tutorial with the qualified name - UA_QUALIFIEDNAME(3, "BoxGoodPartsCounter")

But I don't know how I can trigger a complex event, which contains several ObjectTypes. The Euromap specification requests to subtype the CycleParametersEventType and add some ObjectTypes. 

image

How can I set the BackPressure event field?
I would have expected to find a function which takes a browsepath, but I couldn't find any. And UA_Server_writeObjectProperty_scalar only takes one qualified name. 

Many thanks in advance for your support!

sqadasi:
Facets to be implemented:
1- A&C Base Condition Server Facet (only mandatory)
2- A&C Enable Server Facet (only mandatory)
3- A&C Alarm Server Facet (only mandatory)
4- A&C Acknowledgeable Alarm Server Facet (complete)

First Commit is a test to add the ConditionId Field to Events which are Conditions when using UaExpert. The Commit is to create a PR indicating the work on the first implementation of Part 9.

#2041, fixes #1947, fixes #2738


src/server/ua_subscription_events.c
```diff
@@ -425,17 +459,22 @@ UA_Event_addEventToMonitoredItem(UA_Server *server, const UA_NodeId *event,
 }
 
 static const UA_NodeId objectsFolderId = {0, UA_NODEIDTYPE_NUMERIC, {UA_NS0ID_OBJECTSFOLDER}};
-static const UA_NodeId parentReferences_events[2] =
+static const UA_NodeId parentReferences_events[4] =
```
basyskom-meerkoetter: why not let the compiler deduce the array size? if the UE_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS is off, there will be two unused and uninitialized entries

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2390 @@
...
+                      UA_Variant *output) {
+    UA_QualifiedName fieldConfirmededState = UA_QUALIFIEDNAME(0, CONDITION_FIELD_CONFIRMEDSTATE);
+    UA_QualifiedName fieldConfirmedStateId = UA_QUALIFIEDNAME(0, CONDITION_FIELD_TWOSTATEVARIABLE_ID);
+    UA_QualifiedName fieldComment = UA_QUALIFIEDNAME(0, CONDITION_FIELD_COMMENT);
+    UA_NodeId conditionNode;
+
+    UA_NodeId conditionTypeNodeId = UA_NODEID_NUMERIC(0, UA_NS0ID_CONDITIONTYPE);
+    if(UA_NodeId_equal(objectId, &conditionTypeNodeId)) {
+        UA_LOG_WARNING(server->config.logger, UA_LOGCATEGORY_USERLAND,
+                       "Cannot call method of ConditionType Node. StatusCode %s", UA_StatusCode_name(UA_STATUSCODE_BADNODEI
```
basyskom-meerkoetter: there are 13 cases where "condition" is misspelled "conditoin" in this file.

src/server/ua_subscription_events.c
```diff
@@ -425,17 +459,22 @@ UA_Event_addEventToMonitoredItem(UA_Server *server, const UA_NodeId *event,
...
+#endif /*UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS*/
+    };
 UA_StatusCode
 UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_NodeId origin,
                        UA_ByteString *outEventId, const UA_Boolean deleteEventNode) {
     /* Make sure the origin is in the ObjectsFolder (TODO: or in the ViewsFolder) */
     UA_NodeId *parentTypeHierachy = NULL;
     size_t parentTypeHierachySize = 0;
-    getTypesHierarchy(&server->config.nodestore, parentReferences_events, 2,
+    getTypesHierarchy(&server->config.nodestore, parentReferences_events, 4,
```
basyskom-meerkoetter: this shouldn't be hardcoded.
please calc the size of the array.

examples/tutorial_server_alarms_conditions.c
```diff
@@ -0,0 +1,508 @@
...
+    UA_NodeId parentNodeId = UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER);
+    UA_NodeId parentReferenceNodeId = UA_NODEID_NUMERIC(0, UA_NS0ID_ORGANIZES);
+    UA_NodeId variableTypeNodeId = UA_NODEID_NUMERIC(0, UA_NS0ID_BASEDATAVARIABLETYPE);
+    UA_Server_addVariableNode(server, UA_NODEID_NULL, parentNodeId,
+                              parentReferenceNodeId, CallbackTestVariableName,
+                              variableTypeNodeId, attr, NULL, outNodeId);
+}
+
+static void
+addVariable_2_changeSeverityOfCondtion_2(UA_Server *server, UA_NodeId* outNodeId) {
```
basyskom-meerkoetter: misspelling

examples/tutorial_server_alarms_conditions.c
```diff
@@ -0,0 +1,508 @@
...
+}
+
+/**
+ * this is user specific function which will be called upon acknowledging an alarm notification.
+ * In this example we will set the Alarm to Inactive state. The server is responsible of setting
+ * standard fields related to Acknowledge Method and triggering the alarm notification.
+ */
+static UA_StatusCode
+enteringAckedStateCallback(UA_Server *server, const UA_NodeId *condition) {
+    /* set ActiveState/Id to false*/
```
basyskom-meerkoetter: this comment is "not carrying its own weight".

please compare "activeStateId = false;" with "set ActiveState/Id to false"

examples/tutorial_server_alarms_conditions.c
```diff
@@ -0,0 +1,508 @@
...
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND,
+                     "Setting ActiveState/Id Field failed. StatusCode %s", UA_StatusCode_name(retval));
+    }
+
+    return retval;
+}
+
+static UA_StatusCode
+enteringConfirmedStateCallback(UA_Server *server, const UA_NodeId *condition) {
+    /* set ActiveState/Id to false*/
```
basyskom-meerkoetter: see above.

examples/tutorial_server_alarms_conditions.c
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    return retval;
+}
+
+
+
+/** It follows the main server code, making use of the above definitions. */
+
+UA_Boolean running = true;
```
basyskom-meerkoetter: running should be static

src/server/ua_subscription_events.c
jpfr: The tinyCC compiler does not like the array length being left out...
Probably travis will fail.

sqadasi: > 
> 
> The tinyCC compiler does not like the array length being left out...
> Probably travis will fail.

that's right, travis fails when array length is being left empty. I would put the array defintion in ifdef block, so we won't have wrong array length when A&C are disabled.

include/ua_server.h
```diff
@@ -1250,6 +1250,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+/**
+ * set the value of property of condition field.
+ * @param server The server object
+ * @param condition The NodeId of the node representation of the Condition Instance
+ * @param variantValue Variant Value to be written to the Field
+ * @param type Variant Type
+ * @param variableFieldName Name of the Field which has a property
+ * @param variablePropertyName Name of the Field Property in which the value should be written
+ * @return The StatusCode of the UA_Server_triggerEvent method*/
+UA_StatusCode UA_EXPORT
```
jpfr: Can I do all of this "manually" by browsing and writing in the information model?
This is not a critique. It would be rather good if we could implement this function with "public server API".

The comments above for pointers and variants applied as well.

include/ua_server.h
```diff
@@ -1250,6 +1250,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+/**
+ * triggers an event only for an enabled condition. The condition list is updated then with the
+ * last generated EventId.
+ * @param server The server object
+ * @param condition The NodeId of the node representation of the Condition Instance
+ * @param conditionSource The NodeId of the node representation of the Condition Source
+ * @param outEventId last generated EventId
+ * @return The StatusCode of the UA_Server_triggerEvent method*/
+UA_StatusCode UA_EXPORT
+UA_Server_triggerConditionEvent(UA_Server *server, const UA_NodeId condition,
```
jpfr: This checks the state machine internally.
How can we prevent a user from triggering the condition event manually with UA_Server_triggerEvent?
Is this dangerous?

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2383 @@
...
+                                               UA_TwoStateVariableChangeCallback callback,
+                                               UA_TwoStateVariableCallbackType callbackType) {
+    /* Get ConditionSource Entry */
+    UA_ConditionSource_nodeListElement *conditionSourceEntryTmp;
+    LIST_FOREACH(conditionSourceEntryTmp, &server->headConditionSource, listEntry) {
+        if(UA_NodeId_equal(&conditionSourceEntryTmp->conditionSourceId, conditionSource)) {
+            /* Get Condition Entry */
+            UA_Condition_nodeListElement *conditionEntryTmp;
+            LIST_FOREACH(conditionEntryTmp, &conditionSourceEntryTmp->conditionHead, listEntry) {
+                if(UA_NodeId_equal(&conditionEntryTmp->conditionId, condition)) {
```
jpfr: The indent level can be reduced.
By using if(!NodeId_equal) continue;

src/server/ua_subscription.h
```diff
@@ -54,6 +54,43 @@ typedef struct UA_EventNotification {
...
+typedef struct UA_ConditionBranch_nodeListElement {
+    LIST_ENTRY(UA_ConditionBranch_nodeListElement) listEntry;
+    UA_NodeId* conditionBranchId;
+    UA_ByteString lastEventId;
+} UA_ConditionBranch_nodeListElement;
+
+typedef struct UA_Condition_nodeListElement {
+    LIST_ENTRY(UA_Condition_nodeListElement) listEntry;
+    LIST_HEAD(conditionbranchlisthead, UA_ConditionBranch_nodeListElement) conditionBranchHead;
+    UA_NodeId conditionId;
```
jpfr: All of these structures have a nodeid and are represented by a node in the information model.
Can we put a pointer to this structure into the node "context"?
This makes lookup more standard.

And we don't have to loop over long linked lists for every access.

src/server/ua_server_internal.h
```diff
@@ -172,6 +180,24 @@ getTypesHierarchy(UA_Nodestore *ns, const UA_NodeId *leafType, size_t leafTypeSi
...
 
+
+/* Filters an event according to the filter specified by mon and then adds it to
+ * mons notification queue */
+UA_StatusCode UA_EXPORT
+UA_Event_addEventToMonitoredItem(UA_Server *server, const UA_NodeId *event, UA_MonitoredItem *mon);
+
+/* generates a unique event id */
+UA_StatusCode UA_EXPORT
+UA_Event_generateEventId(UA_Server *server, UA_ByteString *generatedId);
```
jpfr: Events automatically have a unique id when they are generated.
They are represented by a node in the information model.
So the id is automatically unique.

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2383 @@
...
+/**
+ * Gets the NodeId of a Field (e.g. Severity)
+ */
+static UA_StatusCode
+getConditionFieldNodeId(UA_Server *server,
+                        const UA_NodeId *conditionNodeId,
+                        const UA_QualifiedName* fieldName,
+                        UA_NodeId *outFieldNodeId) {
+    UA_BrowsePathResult bpr = UA_Server_browseSimplifiedBrowsePath(server, *conditionNodeId, 1, fieldName);
+    if(bpr.statusCode != UA_STATUSCODE_GOOD || bpr.targetsSize < 1) {
```
jpfr: Can this return good if targetsSize == 0?

sqadasi:
@jpfr thank you for your comments!
I am working currently to fix the issues:) I also attached a draft describing the important parts of the code and its logic. Unfortunatly, the thesis is due to 03.2019, therefore a real high level documentation is not yet finished. I will provide you with one as soon as it is done:)
[draft_doku_A_C_PR_v_0_1.pdf](https://github.com/open62541/open62541/files/2693909/draft_doku_A_C_PR_v_0_1.pdf)

include/ua_server.h
```diff
@@ -1250,6 +1250,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+ * set the value of condition field.
+ * @param server The server object
+ * @param condition The NodeId of the node representation of the Condition Instance
+ * @param variantValue Variant Value to be written to the Field
+ * @param type Variant Type.
+ * @param fieldName Name of the Field in which the value should be written
+ * @return The StatusCode of the UA_Server_triggerEvent method*/
+UA_StatusCode UA_EXPORT
+UA_Server_setConditionField(UA_Server *server, const UA_NodeId *condition,
+                            const void* variantValue, UA_UInt16 type,
```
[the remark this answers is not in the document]
sqadasi: My idea was to spare the user some effort, so the user don't have to create and set variants himself each time they want to set value of a field. So there was no technical reason behind that implementation:)
At the moment this function doesn't support variant arrays as you said (arrays were meant to be in a the next version).

include/ua_server.h
sqadasi: yes, it is possible, because those functions (UA_Server_setConditionField() and UA_Server_setConditionVariableFieldProperty()) use only public server API functions (UA_Server_browseSimplifiedBrowsePath(), UA_Variant_setScalar() and UA_Server_writeValue()).


> This is not a critique. It would be rather good if we could implement this function with "public server API".
I am afraid I didn't understand that part of the comment:(

src/server/ua_subscription.h
sqadasi: good idea, I think it would be possible to do that :)

src/server/ua_subscription_alarms_conditions.c
sqadasi: I implemented it just like in
https://github.com/open62541/open62541/blob/d89e9e55bd810a1d1371dbb94dbe968d318bcdbf/src/server/ua_subscription_events.c#L345
without double checking , my bad...

Now that you mentioned it, I think it can't return good if targetsSize==0
https://github.com/open62541/open62541/blob/b13ad2edb2942939bd5a1aebdd6295c35a69d1a7/src/server/ua_services_view.c#L701

src/server/ua_subscription.h
sqadasi: @jpfr, actually it would've been a good Idea, if no other application used the context of those nodes. When users would like to attach application specific data to the node context, then, in my opinion, it would not be correct to use up the context for internal logic.

jpfr: You are right. I agree.

sqadasi: @jpfr, do you think that using an extra internal context, and leaving the normal context as is, would be a good Idea? I don't know how difficult it is to add such feature to the current implementation, but that would be a nice solution for the linked list looping problem in the Alarms and Conditions implementation (and maybe for other problems as well).

https://github.com/open62541/open62541/pull/2083#discussion_r241076137
for example, a quick solution for that problem would look like this:
(as you see, there are a lot of looping through lists, and that's a real pitfall)

1) In A&C Module:
```C
static void
setIsCallerAC(UA_Server *server,
              const UA_NodeId *condition,
              const UA_NodeId *conditionSource,
              UA_Boolean isCallerAC) {
    /* Get conditionSource Entry */
    UA_ConditionSource_nodeListElement *conditionSourceEntryTmp;
    LIST_FOREACH(conditionSourceEntryTmp, &server->headConditionSource, listEntry) {
        if(!UA_NodeId_equal(&conditionSourceEntryTmp->conditionSourceId, conditionSource))
            continue;
        /* Get Condition Entry */
        UA_Condition_nodeListElement *conditionEntryTmp;
        LIST_FOREACH(conditionEntryTmp, &conditionSourceEntryTmp->conditionHead, listEntry) {
            if(UA_NodeId_equal(&conditionEntryTmp->conditionId, condition)) {// found ConditionId -> branch == NULL
                UA_ConditionBranch_nodeListElement *conditionBranchEntryTmp;
                LIST_FOREACH(conditionBranchEntryTmp, &conditionEntryTmp->conditionBranchHead, listEntry) {
                    if(conditionBranchEntryTmp->conditionBranchId == NULL) {
                        conditionBranchEntryTmp->isCallerAC = isCallerAC;
                        return;
                    }
                }
            }
            else {
                UA_ConditionBranch_nodeListElement *conditionBranchEntryTmp;
                LIST_FOREACH(conditionBranchEntryTmp, &conditionEntryTmp->conditionBranchHead, listEntry) {
                    if(UA_NodeId_equal(conditionBranchEntryTmp->conditionBranchId, condition)) {
                        conditionBranchEntryTmp->isCallerAC = isCallerAC;
                        return;
                    }
                }
            }
        }
    }

    UA_LOG_ERROR(server->config.logger, UA_LOGCATEGORY_USERLAND, "Entry not found in list!");
}

UA_Boolean
isConditionOrBranch(UA_Server *server,
                    const UA_NodeId *condition,
                    const UA_NodeId *conditionSource,
                    UA_Boolean *isCallerAC) {
    /* Get conditionSource Entry */
    UA_ConditionSource_nodeListElement *conditionSourceEntryTmp;
    LIST_FOREACH(conditionSourceEntryTmp, &server->headConditionSource, listEntry) {
        if(!UA_NodeId_equal(&conditionSourceEntryTmp->conditionSourceId, conditionSource))
            continue;
        /* Get Condition Entry */
        UA_Condition_nodeListElement *conditionEntryTmp;
        LIST_FOREACH(conditionEntryTmp, &conditionSourceEntryTmp->conditionHead, listEntry) {
            if(UA_NodeId_equal(&conditionEntryTmp->conditionId, condition)) {// found ConditionId -> branch == NULL
                UA_ConditionBranch_nodeListElement *conditionBranchEntryTmp;
                    LIST_FOREACH(conditionBranchEntryTmp, &conditionEntryTmp->conditionBranchHead, listEntry) {
                        if(conditionBranchEntryTmp->conditionBranchId == NULL) {
                          *isCallerAC = conditionBranchEntryTmp->isCallerAC;
                          return true;
                        }
                    }
            }
            else {
                UA_ConditionBranch_nodeListElement *conditionBranchEntryTmp;
                LIST_FOREACH(conditionBranchEntryTmp, &conditionEntryTmp->conditionBranchHead, listEntry) {
                    if(UA_NodeId_equal(conditionBranchEntryTmp->conditionBranchId, condition)) {
                        *isCallerAC = conditionBranchEntryTmp->isCallerAC;
                        return true;
                    }
                }
            }
        }
    }

    return false;
}
```
```C
UA_StatusCode
UA_Server_triggerConditionEvent(UA_Server *server, const UA_NodeId condition,
                                const UA_NodeId conditionSource, UA_ByteString *outEventId){
    UA_StatusCode retval = UA_STATUSCODE_GOOD;
    UA_ByteString eventId = UA_BYTESTRING_NULL;
    UA_QualifiedName enabledStateField = UA_QUALIFIEDNAME(0, CONDITION_FIELD_ENABLEDSTATE);


    /* Check if enabled */
    if(isTwoStateVariableInTrueState(server, &condition, &enabledStateField)) {
        setIsCallerAC(server, &condition, &conditionSource, true);
        /* Trigger the event for Condition*/
        retval = UA_Server_triggerEvent(server, condition, conditionSource,
                                        &eventId, false); //Condition Nodes should not be deleted after triggering the event
        CONDITION_ASSERT_RETURN_RETVAL(retval, "Triggering condition event failed",);

        setIsCallerAC(server, &condition, &conditionSource, false);
[...]
```

2) In Events Module:
```C
UA_StatusCode
UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_NodeId origin,
                       UA_ByteString *outEventId, const UA_Boolean deleteEventNode) {

#ifdef UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS
    UA_Boolean isCallerAC = false;
    if(isConditionOrBranch(server, &eventNodeId, &origin, &isCallerAC)) {
        if(!isCallerAC) {
          UA_LOG_WARNING(server->config.logger, UA_LOGCATEGORY_SERVER,
                                 "Condition Events: Please use A&C API to trigger Condition Events",
                                  UA_STATUSCODE_BADINVALIDARGUMENT);
          return UA_STATUSCODE_BADINVALIDARGUMENT;
        }
    }
#endif /*UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS*/
[...]
```

davesliu:
Have the Alarms & Condition feature been added to master trunk？

CMakeLists.txt
```diff
@@ -134,7 +134,8 @@ option(UA_ENABLE_METHODCALLS "Enable the Method service set" ON)
 option(UA_ENABLE_NODEMANAGEMENT "Enable dynamic addition and removal of nodes at runtime" ON)
 option(UA_ENABLE_SUBSCRIPTIONS "Enable subscriptions support." ON)
 option(UA_ENABLE_SUBSCRIPTIONS_EVENTS "Enable the use of events. (EXPERIMENTAL)" OFF)
-option(UA_ENABLE_DISCOVERY "Enable Discovery Service (LDS)" ON)
+option(UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS "Enable the use of A&C. (EXPERIMENTAL)" OFF)
+option(UA_ENABLE_DISCOVERY "Enable Discovery Service (LDS)" OFF)
```
Pro: This should not be changed. UA_ENABLE_DISCOVERY is by default ON.

src/server/ua_subscription_events.c
```diff
@@ -85,6 +85,26 @@ UA_Server_createEvent(UA_Server *server, const UA_NodeId eventType, UA_NodeId *o
         return UA_STATUSCODE_BADINVALIDARGUMENT;
     }
 
+
+    /***** Hilscher *****/
```
Pro: This should not be in the main repo?!

src/server/ua_subscription_monitoreditem.c
```diff
@@ -121,6 +121,8 @@ UA_Notification_dequeue(UA_Server *server, UA_Notification *n) {
     TAILQ_REMOVE(&mon->queue, n, listEntry);
     --mon->queueSize;
 
+    mon->eventOverflows = 0;// see open issue #2114
```
Pro: Better add a more detailed comment here, so that one does not need to open the issue to understand why this is necessary

sqadasi:
Hello @jpfr and @Pro , 
it would be nice if you have a look into the updates. I worked through the issues that you mentioned previously, hope didn't forget anything:)
Kind regards

davesliu:
@sqadasi, thanks for your implementation of alarm & condition.
When using your updates in below tree and want to build a single open62541.c file  with set UA_ENABLE_AMALGAMATION to ON. It will generate error information for open62541.c to show the static variable     UA_NodeId parentReferences_events[4]  in line 1550 of ua_subscription_alarms_conditions.c will become shadow variable as another global variable has the same name.

I renamed the variables in line 1550 and the build of single open62541 will success.

https://github.com/open62541/open62541/blob/cceb0b14ad76f8fbe1a54c44ac2c5fd6ecc1b163/src/server/ua_subscription_alarms_conditions.c#L1550

davesliu:
@sqadasi  
When I use the UaExpert (an OPC UA client tool) to browse the tutorial_server_alarms_conditions. The content of variables such as "Change Severity Condition 2" can't be displayed, the root cause seems to be that the nodeID is NULL in line 122 of tutorial_server_alarms_conditions.c. When I set the nodeId with a string format nodeID. The content of variable "Change Severity Condition 2" can be displayed correctly.

Similar issue for another two variables in tutorial_server_alarms_conditions.c

https://github.com/open62541/open62541/blob/cceb0b14ad76f8fbe1a54c44ac2c5fd6ecc1b163/examples/tutorial_server_alarms_conditions.c#L122

sqadasi:
@davesliu thank you for the check!
as for https://github.com/open62541/open62541/pull/2083#issuecomment-502129233, you are right. It's a copy-paste mistake:)
regarding https://github.com/open62541/open62541/pull/2083#issuecomment-502133007, UA_NODEID_NULL as a parameter for the requested NodeId will causes the function "UA_Server_addVariableNode" (actually all functions for creating Nodes) to create a Node with a random numeric NodeId. All Nodes in the example have random generated NodIds. I tested it also with UaExpert and it works normally for me. I really can't tell what causes the Problem in your server....

doc/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ generate_rst(${PROJECT_SOURCE_DIR}/examples/tutorial_server_variable.c ${DOC_SRC
 generate_rst(${PROJECT_SOURCE_DIR}/examples/tutorial_server_variabletype.c ${DOC_SRC_DIR}/tutorial_server_variabletype.rst)
 generate_rst(${PROJECT_SOURCE_DIR}/examples/tutorial_server_datasource.c ${DOC_SRC_DIR}/tutorial_server_datasource.rst)
 generate_rst(${PROJECT_SOURCE_DIR}/examples/tutorial_server_events.c ${DOC_SRC_DIR}/tutorial_server_events.rst)
+generate_rst(${PROJECT_SOURCE_DIR}/examples/tutorial_server_events.c ${DOC_SRC_DIR}/tutorial_server_alarms_conditions.rst)
```
jpfr: `tutorial_server_events.c` is probably the wrong source here.

include/open62541/server.h
```diff
@@ -1285,6 +1285,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+ * source and server object. To expose the condition in address space, a hierarchical ReferenceType
+ * should be passed to create the reference to condition source. Otherwise, UA_NODEID_NULL should be
+ * passed to make the condition unexposed.
+ * @param server The server object
+ * @param conditionType The NodeId of the node representation of the ConditionType
+ * @param conditionName The name of the condition to be created
+ * @param conditionSource The NodeId of the Condition Source (Parent of the Condition)
+ * @param hierarchialReferenceType The NodeId of Hierarchical ReferenceType between Condition and its source
+ * @param outNodeId The NodeId of the created Condition
+ * @return The StatusCode of the UA_Server_triggerEvent method */
```
jpfr: Why the result of `UA_Server_triggerEvent`?
What happens if the Condition node cannot be created?

include/open62541/server.h
```diff
@@ -1285,6 +1285,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+/**
+ * triggers an event only for an enabled condition. The condition list is updated then with the
+ * last generated EventId.
+ * @param server The server object
+ * @param condition The NodeId of the node representation of the Condition Instance
+ * @param conditionSource The NodeId of the node representation of the Condition Source
+ * @param outEventId last generated EventId
+ * @return The StatusCode of the UA_Server_triggerEvent method*/
+UA_StatusCode UA_EXPORT
+UA_Server_triggerConditionEvent(UA_Server *server, const UA_NodeId condition,
```
jpfr: Why can't we just use the normal `UA_Server_triggerEvent` with the respective `eventNodeId` and `originId`?

include/open62541/server.h
```diff
@@ -1285,6 +1285,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+ * Function used to set a user specific callback to TwoStateVariable Fields of
+ * a condition. The callbacks will be called before triggering the events when
+ * transition to true State of EnabledState/Id, AckedState/Id, ConfirmedState/Id
+ * and ActiveState/Id occurs.
+ * @param server The server object
+ * @param condition The NodeId of the node representation of the Condition Instance
+ * @param conditionSource The NodeId of the node representation of the Condition Source
+ * @param removeBranch (Not Implemented yet)
+ * @param callback User specific callback function
+ * @param callback Callback function type, indicates where it should be called
```
jpfr: @param callback is used twice

include/open62541/server.h
```diff
@@ -1285,6 +1285,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+ * is not implemented yet)
+ * @param server The server object
+ * @param condition The NodeId of the node representation of the Condition Instance
+ * @param conditionType The NodeId of the node representation of the Condition Type
+ * from which the optional field comes
+ * @param fieldName Name of the optional field
+ * @param outOptionalVariable The NodeId of the created field (Variable Node)
+ * @return The StatusCode of the UA_Server_triggerEvent method*/
+UA_StatusCode UA_EXPORT
+UA_Server_addConditionOptionalField(UA_Server *server, const UA_NodeId condition,
```
jpfr: What is the difference between `UA_Server_writeObjectProperty_scalar` and `UA_Server_setConditionVariableFieldProperty`?

include/open62541/server.h
```diff
@@ -1285,6 +1285,114 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId, const UA_
...
+  UA_ENTERING_ACKEDSTATE,
+  UA_ENTERING_CONFIRMEDSTATE,
+  UA_ENTERING_ACTIVESTATE
+} UA_TwoStateVariableCallbackType;
+
+/**
+ * callback prototype to set user specific callbacks
+ */
+typedef UA_StatusCode
+(*UA_TwoStateVariableChangeCallback)(UA_Server *server, const UA_NodeId *condition);
```
jpfr: Is there an example where the UA_TwoStateVariableChangeCallback is used?
Wouldn't I need the `leaving` callbacktypes in addition to the `entering` ones?

include/open62541/server.h
sqadasi: `UA_Server_setConditionVariableFieldProperty ` is almost the same as `UA_Server_writeObjectProperty_scalar`. The difference is that the first sets a property of a Variable Node (in this case a Condition Field). It takes the Condition NodeId, Field Name and Field Property Name as input parameter. Using `UA_Server_writeObjectProperty_scalar` (for users) will require more code in my opinion, because all Fields are generated automatically when we create a Condition Node und we don't track the NodeIds of the created Fields. So the user should find the correct NodeIds first using their BrowseName then perform `UA_Server_writeObjectProperty_scalar`.

`UA_Server_setConditionField` and
`UA_Server_setConditionVariableFieldProperty ` use only public API functions, so the user can implement their own logic without problems. Inside the A&C module those function could be used staticly. What do you think is best?

include/open62541/server.h
sqadasi: The idea behind that callback is to give the user the possibilty to set additional Condition Fields, which are not set by default when Condition Fields from Type TwoStateVariableType or ConditionVariableType change. For example, when enabling a Condition, UA Spec. doesn't specify what happens to the Retain Field. It could be set to true or stay false. Of course we could set Retain Field to true separatly and then trigger an Event manually. But we will end up with two Events instead of one. The same for AcknowledgeableState, the user my want to change the severity or deactivate the alarm when the AcknowledgeableState changes to true.
So It wasn't meant to change something on entering and also on leaving the callback.

It's just a proposal based on my understanding, so feel free to correct me, if I got it wrong:)

include/open62541/server.h
sqadasi: copy paste error in the description...

include/open62541/server.h
sqadasi: The last generated EventIds should be stored in an intern list so we could manage the Conditions. (EvenIds are needed for most Condition method calls)

include/ua_server.h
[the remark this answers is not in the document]
Pro: I would also prefer to pass a Variant here. Like this we have a more consistent API. @sqadasi can you change that. If one passes a variant array, check this inside the method and show an error. Mark it with `//TODO`

Same for the other methods where you use `const void* variantValue, UA_UInt16 type`

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2575 @@
...
+        /* Get Condition Entry */
+        UA_Condition_nodeListElement *conditionEntryTmp;
+        LIST_FOREACH(conditionEntryTmp, &conditionSourceEntryTmp->conditionHead, listEntry) {
+            if(UA_NodeId_equal(&conditionEntryTmp->conditionId, condition)) {
+                return getonditionTwoStateVariableCallback(server, condition, conditionEntryTmp, removeBranch, callbackType);
+            }
+            else {
+                UA_ConditionBranch_nodeListElement *conditionBranchEntryTmp;
+                LIST_FOREACH(conditionBranchEntryTmp, &conditionEntryTmp->conditionBranchHead, listEntry) {
+                    if(UA_NodeId_equal(conditionBranchEntryTmp->conditionBranchId, condition))
```
andranik-s: I believe that checking for a null pointer is needed here. I'm getting a segfault here since the first branch (trunk) has its id == NULL.
```suggestion
                    if(conditionBranchEntryTmp->conditionBranchId != NULL &&
                       UA_NodeId_equal(conditionBranchEntryTmp->conditionBranchId, condition))
```

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2575 @@
...
+                default:
+                    return UA_STATUSCODE_BADNOTFOUND;
+            }
+        }
+    }
+    return UA_STATUSCODE_BADNOTFOUND;
+}
+
+static UA_StatusCode
+getonditionTwoStateVariableCallback(UA_Server *server,
```
andranik-s: Typo
```suggestion
getConditionTwoStateVariableCallback(UA_Server *server,
```

src/server/ua_subscription_events.c
```diff
@@ -373,25 +391,46 @@ UA_Event_addEventToMonitoredItem(UA_Server *server, const UA_NodeId *event,
...
                        UA_ByteString *outEventId, const UA_Boolean deleteEventNode) {
-    /* Check that the origin node exists */
-    const UA_Node *originNode = UA_Nodestore_getNode(server->nsCtx, &origin);
-    if(!originNode) {
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_USERLAND,
-                     "Origin node for event does not exist.");
-        return UA_STATUSCODE_BADNOTFOUND;
+#ifdef UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS
+    UA_Boolean isCallerAC = false;
+    if(isConditionOrBranch(server, &eventNodeId, &origin, &isCallerAC)) {
```
andranik-s: `isConditionOrBranch` has three nested `LIST_FOREACH`, so if someone enables A&C this will traverse condition lists (that might be quite long with branches) every time they want to trigger an event. And `isCallerAC` field is updated throughout the rest of the code.
As I understand it, all that is needed just to make sure that a user won't trigger a condition event using this function. As far as I am concerned this seems like overkill.

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2575 @@
...
+    CONDITION_ASSERT_RETURN_VOID(retval, "ConditionSource not found", UA_NodeId_deleteMembers(&conditionNode););
+
+    UA_ActiveState lastActiveState = UA_INACTIVE, currentActiveState = UA_INACTIVE;
+    UA_Boolean isLimitalarm = false;
+
+    retval = getConditionActiveState(server, &conditionSource, &conditionNode, &lastActiveState, &currentActiveState, &isLimitalarm);
+    CONDITION_ASSERT_RETURN_VOID(retval, "ActiveState transition check failed", UA_NodeId_deleteMembers(&conditionNode);
+                                 UA_NodeId_deleteMembers(&conditionSource););
+
+    if(isLimitalarm == false) {
```
andranik-s: Could you please explain the idea behind this code?

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2575 @@
...
+    CONDITION_ASSERT_RETURN_VOID(retval, "Set Condition LAstSeverity failed", UA_NodeId_deleteMembers(&condition);
+                                 UA_NodeId_deleteMembers(&conditionSource););
+
+    /* Set SourceTimestamp */
+    UA_Variant_setScalar(&value, &outLastSeverity.sourceTimeStamp, &UA_TYPES[UA_TYPES_DATETIME]);
+    retval = UA_Server_setConditionVariableFieldProperty(server, condition, &value, fieldLastSeverity, fieldSourceTimeStamp);
+    CONDITION_ASSERT_RETURN_VOID(retval, "Set LastSeverity SourceTimestamp failed", UA_NodeId_deleteMembers(&condition);
+                                 UA_NodeId_deleteMembers(&conditionSource););
+
+    /* Update lastSeverity in
```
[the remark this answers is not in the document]
sqadasi: Hello @Nanahara, thank you for checking the code!
Could you please specify where exactly sholuld I put this condition check?

src/server/ua_subscription_alarms_conditions.c
sqadasi: you are right:)
The missing Null pointer check was because the previous implementation of `UA_NodeId_equal` did that check.

src/server/ua_subscription_alarms_conditions.c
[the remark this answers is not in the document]
Nanahara: Oh sorry I think the code line would be underlined by my comment

The condition is in line **1658** in `ua_subscription_alarms_condition.c`

I think you should replace `if(subscription == NULL)` by `if(monitoredItem == NULL)`

src/server/ua_subscription_events.c
sqadasi: @andranik-s thank you for reviewing the code.
I agree with you on this point, that it isn't a nice solution. I discussed that already with @jpfr in:
https://github.com/open62541/open62541/pull/2083#discussion_r248270489

The Problem is when a user triggers a Condition Event using the normal `UA_Server_triggerEvent` function, we will lose track of the last EventId and we won't be able to use the Condition Methods.
another solution would be to create an extra flat list to store all the Condition and their Branches NodeId's, but the pitfall would be that we hav to manage two lists...
Maybe you have a better Idea?:)

src/server/ua_subscription_alarms_conditions.c
sqadasi: The idea was to use it later with LimitAlarmType Conditions. At this point it will always evaluate to false to update the ActiveState.
In the Future implementation, if evaluates to true, we don't need to update the ActiveState, because it should be updated in another callback specific for the LimitAlarms.

Pro:
IMHO this PR is in a good state. I did not review every line, but overall it looks good.

@jpfr Should we squash-merge this to avoid further effort in keeping it up-to-date? Or you you plan to review it once more?

src/server/ua_subscription_events.c
andranik-s: @sqadasi IMHO this check is not nesessary at all. Just warning users in documentation that this function should not be used with condition events seems OK to me.

Alternatively, event type could be used to check for conditions. It will require to create an additional (static) function to trigger condition events inside your code, but I don't think it is a problem. Though, I am not sure it is a very good idea because it will not let users trigger "raw" condition events should they want that for whatever reason.

src/server/ua_subscription_alarms_conditions.c
andranik-s: @sqadasi thank you for the explanation

src/server/ua_subscription_events.c
sqadasi: @andranik-s thank you for answering and sorry for my late respond...
By using the EventType we have to do two steps. First getting the EventType NodeId and then checking the path of it, whether it's a SubType of ConditionType. Here we could get the problem of iterating through the Node path with `getTypesHierarchy`, which will be used for the second time in the Event Triggering function and I think it won't be more effective if we have deeply nested ConditionTypes.

IMHO I think there will be always someone who will use the wrong function by mistake, even with a good documented API, so I am not sure whether it's a good idea to remove that check at all :)

Maybe @jpfr and @Pro could share their opinions with us:)

src/server/ua_subscription_alarms_conditions.c
```diff
@@ -0,0 +1,2576 @@
...
+    UA_Variant_setScalar(&value, &outLastSeverity.lastSeverity, &UA_TYPES[UA_TYPES_UINT16]);
+    retval = UA_Server_setConditionField(server, condition, &value, fieldLastSeverity);
+    CONDITION_ASSERT_RETURN_VOID(retval, "Set Condition LAstSeverity failed", UA_NodeId_deleteMembers(&condition);
+                                 UA_NodeId_deleteMembers(&conditionSource););
+
+    /* Set SourceTimestamp */
+    UA_Variant_setScalar(&value, &outLastSeverity.sourceTimeStamp, &UA_TYPES[UA_TYPES_DATETIME]);
+    retval = UA_Server_setConditionVariableFieldProperty(server, condition, &value, fieldLastSeverity, fieldSourceTimeStamp);
+    CONDITION_ASSERT_RETURN_VOID(retval, "Set LastSeverity SourceTimestamp failed", UA_NodeId_deleteMembers(&condition);
+
```
sqadasi: Not needed anymore, because it is allowed now to trigger Events of abstract EventTypes.

juangburgos:
Hi guys, I made an alternative A&C implementation on Qt/C++ that has branch support if somebody is iterested in testing, or maybe you have some feedback:

<https://github.com/juangburgos/QUaServer#Alarms>

Issue #2082, opened 2018-09-24
sendSymmetricChunk() writes to send buffer without checking its boundaries


jonasgreen88:
Description
===========
We have added an encrypted endpoint to our server. When we connect to the endpoint we almost immediately encounter a fatal crash of our device. We have now tracked the crash to the stack that may write out of the boundaries of the send buffer. In our configuration the message chunks are the minimum size 8196 bytes which probably increases the chance to encounter this issue.

The function sendSymmetricChunk() writes both the padding and the signature to the send buffer without verifying that it doesn't exceeds the send buffer boundaries. In our case when we encountered the device crash the msg size + padding + signature > 8196 bytes.
https://github.com/open62541/open62541/blob/c9c2088124d6b3ff4e3068921812f1abdac40536/src/ua_securechannel.c#L704

We are planing to test a solution that will verify that the buffer boundaries are not exceeded when writing the padding and the signature and return BADENCODINGLIMITSEXCEEDED if there not enough space. Do you think that's the right solution for this issue?

When returning this error code we think the error will be handled at this point, is that right?
https://github.com/open62541/open62541/blob/c9c2088124d6b3ff4e3068921812f1abdac40536/src/ua_types_encoding_binary.c#L1418
If that is the case, the last added member will be removed and the stack will try to send the chunk again. But imagine that the last member is a small one and just removing that member doesn't free up enough space of the chunk for the padding and the signature. Then the sendSymmetricChunk() function will fail again because of the same reason. In that case the message will be treated as "too large". Shouldn't the stack try to remove some more members from the encoded data before giving up sending the message chunk or maybe even reserve space for the padding and signature before encoding the data to send?

We can help out making a fix for this issue but may need some guiding from you who designed the encoding algorithm to get a good solution.

Background Information / Reproduction Steps
===========================================
Run the stack with message chunk size 8196 bytes and connect to a SignAndEncrypt endpoint.

Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): 43246510c0803dfd24fb42bf9a4e0e4a90679a96 (0.3 branch)
 - [x] Other OPC UA SDKs used (client or server): 
 - [x] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [x] Self-contained code example attached
 - [x] Critical issue

jonasgreen88:
[...]

PR #2084, opened 2018-09-25
Reduce the size of the msg chunks to fit header and footer

jonasgreen88:
Before encoding the response into the send buffer space was reserved
for the header, the signature (if applicable) and 2 bytes for the
padding (if applicable). When reserving space for the header, the
end of the buffer was moved out of bounds of the send buffer instead
of reducing the size available for encoded data. 2 bytes padding is
also not enough for worst-case padding scenarios.

With this fix the end buffer position is calculated correctly when
reserving space for the message header and enough space is reserved
for the padding (if the chunk shall be encrypted).

### DIFF
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -607,7 +607,7 @@ setBufPos(UA_MessageContext *mc) {
          * padding (1 <= paddingSize <= encryptionBlockSize).
          */
         size_t maxEncryptDataSize = mc->messageBuffer.length-UA_SECURE_CONVERSATION_MESSAGE_HEADER_LENGTH-UA_SYMMETRIC_ALG_SECURITY_HEADER_LENGTH;
-        mc->buf_end -= (maxEncryptDataSize % encryptionBlockSize) + 1;
+        mc->buf_end -= (maxEncryptDataSize % encryptionBlockSize);
     }
 }
 
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -599,7 +599,7 @@ setBufPos(UA_MessageContext *mc) {
         /* PaddingSize and ExtraPaddingSize fields */
         size_t encryptionBlockSize = securityPolicy->symmetricModule.cryptoModule.encryptionAlgorithm.
             getLocalBlockSize(securityPolicy, channel->channelContext );
-        mc->buf_end -= 1 + (encryptionBlockSize >> 8) ? 1 : 0;
+        mc->buf_end -= 1 + ((encryptionBlockSize >> 8) ? 1 : 0);
 
         /* Reduce the message body size with the remainder of the operation
          * maxEncryptedDataSize modulo EncryptionBlockSize to get a whole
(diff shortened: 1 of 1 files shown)


jonasgreen88:
Solves issue #2082

Infinity95:
Looks good. Though, isn't it a bit aggressive to remove one complete block size from the buffer?
Wouldn't it be better to make the buffer fit exactly? I.e. cut off buffer so that the remaining buffer minus the size of the parts that are not encrypted is equal to the largest size that is 0 modulo the block size, without overflowing the buffer?

jonasgreen88:
@Infinity95 Thanks for the feedback! I think I got the math right now to reserve just enough data to fit the padding in the message chunk.

jpfr:
The SecureChannel has already changed quite a bit from 0.3.
Can we still backport once the fix is solid?

Also a unit test would be nice.
To prevent regressions in the future.

Good job @jonasgreen88 !!

jonasgreen88:
@Infinity95 Unfortunately I was to quick uploading the change that removes the +1. Without that +1 our device crashes due to the boundaries of the send buffer breaks when writing the footer of the message chunk. Should the +1 be there or may it be some other field that is forgotten to reserve space for?

@jpfr I develop this fix on the 0.3 branch as that's branch we are using in our product. The only thing I've changed to make it build on the master is how to access the sendBufferSize in the UA_assert() of the sendSymmetricChunk() function.

Infinity95:
It might be that the padding is calculated incorrectly for the case where the padding should be zero. I currently can't check if thats the case. You could check if that's the case, otherwise i'll have a look at it tomorrow.

jonasgreen88:
So I added the +1 byte to reserve for padding as it doesn't work without it. If maximum message body size only is reduced with the remainder of maxEncryptDataSize mod blockSize, the server is allowed to fill the message body so the sequence header + message body + PaddingSize field + signature together occupies a whole number of encryption blocks.

In this case the calculatePaddingSym() function will return 16 bytes of padding. Together with the PaddingSize field it will be 17 bytes. The result is that the message chunk will be padded with a complete encryption block (which there is no available size for in the message chunk).

If looking into the padding calculation in calculatePaddingSym(), one can see that it can never return 0.
``
UA_UInt16 padding = (UA_UInt16)(encryptionBlockSize - ((bytesToWrite + signatureSize + 1) % encryptionBlockSize));``
As it never returns 0, the PaddingSize field will always contain a value between 1 and encryptionBlockSize. I.e. there will always be at least one pad byte except the PaddingSize field. Therefore one more byte must be reserved initially in the message chunk.

Infinity95:
You're right. Although i find it questionable that the specification says to always have at least one byte of padding in addition to the padding size byte. Seems like a waste to me.

jonasgreen88:
@Infinity95 I agree, the PaddingSize field itself could be treated as padding and be the minimum padding to add.

@jpfr A test would be great. I took a quick look at the tests and I'm not really sure how to verify it properly. Can it be enough to update the current test case of each security policy by reducing the message chunk size to minimum 8192 bytes and add a large variable node to read which will be transfered over two message chunks? If it is ok, I would like this PR to be submitted once you guys have accepted the change so it can be cherry-picked to the 0.3 branch as well. Then I can add a test case in a new PR once we decided how to verify it. Would that be ok?

jpfr:
Looks good. Thanks again @jonasgreen88 .

A simple way to test this would be the following:

- Create a write request with a string value.
- Increase the length of the string value so that the encoded chunk length is close to a multiple of the block size.
- Repeat sending the write request, always increasing the string length by one. Until we roll over into the next block.

The same can be done also at the maxchunksize "rollover".
That way we can be sure to have snuffed out all issues.
At least for our server/client combination.

@Infinity95 You want to take up the tests?

Infinity95:
@jpfr I think i have enough on my hands with the network stuff at the moment

PR #2096, opened 2018-10-05
Fix log level in MonitoredItem_delete

StalderT:


StalderT:
@jpfr can we remove the log like in 0.4 ?

jpfr:
Can we make it symmetric?
If we log the creation, then we should also log the deletion at the same level.

StalderT:
> Can we make it symmetric?
> If we log the creation, then we should also log the deletion at the same level.

Yes. The creation logs nothing. So, I remove the log. I have change the PR title.

jpfr:
Logs are a good thing. What is the problem with this one?
Why not rather add one for creating monitoreditems?

The nodes that are subscribed would be good to know in an audit...

What are the criteria for adding or removing log messages?

StalderT:
> Logs are a good thing. What is the problem with this one?

No problem for me if the log is not a warning. I revert the commit to the original state (Fix log level in MonitoredItem_delete). When I remove 100'000 items, my system slows down because it displays 100'000 warning.

> Why not rather add one for creating monitoreditems?

No problem for me. If someone would make this?

jpfr:
I added two commits to get a "nice" log history for subscriptions and monitoreditems.

StalderT:
@jpfr thanks!

PR #2121, opened 2018-10-18
[WIP] Automatically generate Changelog

Pro:
As discussed in our last TelCo I had a look at the automatic generation of the Changelog.

I found https://github.com/github-changelog-generator/github-changelog-generator which is quite nice.
Based on the labels on each issue and PR there will be a corresponding line in the resulting Changelog.

It is generated by simply installing the ruby gem, and then executing `github_changelog_generator` in the repo. To improve the performance you should add the 

We still need to clean up already closed issues and PRs to give them better titles and maybe change the labels, but all in all this changelog would be quite helpful.

Just have a look at the `CHANGELOG.md` file in this PR.

Please leave any comment here!


Pro:
What we need to do to clean up the changelog, is to start at Issue/PR 1000 (all the issues below that are before 0.3 release) and work up to the current head:
* Rename the PR/Issue to an easy graspable title
* Assign the correct labels: 
   - Ignore: `question`, `duplicate`, `invalid`, `wontfix`, `ignore changelog`
   - `bug`
   - `enhancement`

And then do this for all the new issues and PRs

jpfr:
I find the generated changelog less than ideal.
We need a way to present users with the important changes first.
Such as breaking API when we cannot avoid them.

Otherwise the important stuff will get lost in the noise of bugfix and house-keeping commits.

Maybe we can look at the options during the next conference call.

Issue #2119, opened 2018-10-18
XML Nodeset compiler: generated code emits various errors, array nodes are not created


mtenberge:
**Edit**: the example case in the original issue body is no longer up-to-date. See the newer discussion messages for an updated case.

---

What am I trying to do?:
- I created a minimalistic project in Siemens TIA Portal for an S7-1500 PLC (which includes an OPC-UA server).
- I exported the OPC-UA server configuration as an UANodeSet xml file.
- Goal: I want to run an open62541 OPC-UA server and load this UANodeSet, so it has the same nodes as the Siemens PLC.

**Self-contained example**: [simatic-opcua-demo.zip](https://github.com/open62541/open62541/files/2491840/simatic-opcua-demo.zip)

- Extract the zip, and place the open62541 tree under the folder with that name
- Run `original.sh` to encounter the first problems.
- I made some workarounds for these errors, to try to get a bit further. They are described below. Run `workarounds.sh` to see the situation with those workarounds applied. At this point I'm really stuck...

Problems and workarounds:
=====
1 The generated namespace setup code for my UaNodeSet XML file does not compile.
---
--> Run `original.sh` to see what happens.
(you should see various compilation errors while compiling `ua_namespace_simatic.c`, such as `error: ‘UA_TYPES_SIMATIC_BYTE’ undeclared (first use in this function); did you mean ‘UA_TYPES_SIMATIC’?`)

What happened?
Most UAVariable nodes in the Siemens file make use of Alias datatypes (such as _BYTE_, see line 28 of the XML file). These aliases point to nodes in the XML file, such as `ns=3;i=3001`. These nodes are defined as Subtype of the standard OPC data types, such as Byte (i=3). This is done via a reverse `HasSubtype` reference.
The first problem here is, that array index constants are generated for these types (for example UA_BYTE) by the nodeset compiler, but these types (including their constants or their array positions) are not produced by the type compiler.

Workaround: I modified the Aliases, so they point directly to the standard OPC-UA types, instead of to the Siemens subtypes. See `tools/replace_aliases.sh`.

What more?
The enum values node ns=3;i=6083 is not generated (correctly), so the UAVariable node ns=3;i=6071 encounters an undefined type `UA_TYPES_SIMATIC_SIMATICOPERATINGSTATE`.

Workaround: I replaced this data type by a simple Int. See the last line of  `tools/replace_aliases.sh`.

What more?
[...]

PR #2123, opened 2018-10-19
[Review] Correctly handle datatypes with string NodeIds

Pro:
Related to #2119 where datatypes node ids can also be string values with quotations within the string


jpfr:
Looks great! Merged.

Issue #2140, opened 2018-10-25
Handle certificate chains in the OpenSecureChannel and CreateSession requests


jonasgreen88:
Description
===========
I'm attending the OPC IOP Workshop this week at Siemens in Nürnberg. With one client (using Eclipse Milo SDK) our device returned BadCertificateInvalid. This error code is only returned at one location in the stack: https://github.com/open62541/open62541/blob/e2d4b0bf0355dfe8a064889f69307169959059b9/src/server/ua_services_session.c#L78

We found out that the client included the complete certificate chain (in this case the device certificate + the CA certificate) in the _clientCertificate_ field of the request. But it did only return the device certificate in the CreateSessionRequest. When the stack compares the _remoteCertificate_ of the secure channel and the _clientCertificate_ of the CreateSessionRequest it fails. See attached Wireshark log.
[certificate-chain-opensecurechannel.zip](https://github.com/open62541/open62541/files/2514498/certificate-chain-opensecurechannel.zip)

This is a valid behavior that I think the stack must support. The OPC UA specification Part 6 (1.04), chapter 6.2.3, says:

> However, applications have the option of including a partial or complete chain whenever they pass a Certificate to a peer during the SecureChannel negotiation and during the CreateSession/ActivateSession handshake. All OPC UA applications shall accept partial or complete chains in any field that contains a DER encoded Certificate.
>
> Chains are stored in a ByteString by simply appending the DER encoded form of the Certificates. The first Certificate shall be the end Certificate followed by its issuer. If the root CA is sent as part of the chain it is last Certificate appended to the ByteString.

So, I have done a small investigation what needs to be fixed. Feel free to comment and extend these results.

Secure channel
-----------------
The _SenderCertificate_ of the Asymmetric algorithm security header can contain either just the device certificate of the client, part of the chain or the complete chain CA->device certificate. In part 6 (1.04), chapter 6.7.2.3, table 42, the description of the _SenderCertificate_ says:

> If the Certificate is signed by a CA, the DER encoded CA Certificate may be appended after the Certificate in the byte array. If the CA Certificate is also signed by another CA this process is repeated until the entire Certificate chain is in the buffer or if MaxSenderCertificateSize limit is reached (the process stops after the last whole Certificate that can be added without exceeding the MaxSenderCertificateSize limit).
Receivers can extract the Certificates from the byte array by using the Certificate size contained in DER header (see X.509 v3).
**Receivers that do not handle Certificate chains shall ignore the extra bytes.**

[...]

PR #2141, opened 2018-10-26
Compare end certificates of channel and CreateSessionRequest

jonasgreen88:
A client are allowed to include a partitial or a complete certificate
chain both in the OpenSecureChannelRequest and in the
CreateSessionRequest. When comparing the certificates while
processing the CreateSessionRequest this must be considered.
The comparition has been updated to use the compareCertificate
function of the channelModule of the SecurityPolicy plugin API.

### DIFF
--- a/plugins/ua_securitypolicy_basic128rsa15.c
+++ b/plugins/ua_securitypolicy_basic128rsa15.c
@@ -718,6 +718,7 @@ channelContext_compareCertificate_sp_basic128rsa15(const Basic128Rsa15_ChannelCo
     const UA_SecurityPolicy *securityPolicy = cc->policyContext->securityPolicy;
 
     mbedtls_x509_crt cert;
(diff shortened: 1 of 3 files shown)


jonasgreen88:
This solved the issue described in #2140 (when the client sent us the certificate chain in the OpenSecureChannelRequest and the CreateSessionRequest only contained the end certificate). But I think we need to do additional changes to also handle certificate chains in the CreateSessionRequest. But that could maybe be made in another PR?

jpfr:
> But that could maybe be made in another PR?

:+1:

jonasgreen88:
Okey, I might have found the issue. The certificate parsed by mbedtls in channelContext_compareCertificate_sp_() is never freed. Try to call mbedtls_x509_crt_free() before returning from that function.

jpfr:
The fix is in #2169.

PR #2142, opened 2018-10-26
Fix-gcc-V4.9.x-warning -maybe-uninitialized

mlang-de:
GCC version 4.9.x complains about variable finalLength (compiler setting -Wmaybe-uninitialized)


jpfr:
Thanks. Merged.

PR #2147, opened 2018-10-28
Use defines UA_TRUE/UA_FALSE instead of true/false

mlang-de:


jpfr:
I find true/false more readable.
Is there a technical reason for UA_TRUE and UA_FALSE or is this a matter of different tastes?

jpfr:
You are right. We should aim for consistency. And not break existing code....
Maybe we can add a deprecated flag to the define. And remove it in a future release.

mlang-de:
This sounds good to me! I would replace the use of the defines and try to add an deprecated_attribute on the defines. Should I split These actions in to 2 PullRequests? One the define replaced and the second for the deprecated_attribute? To be honest, I guess the deprecate_attribute is much Compiler specific and I only familiar with gcc.

mlang-de:
Done as discussed by PR  #2173. 
Sorry, I didn't found time this week.

PR #2148, opened 2018-10-28
Fix PC-lint warning: Declaration of symbol 'index' hides symbol 'index(...)'

mlang-de:
PC-lint complains about the naming:
Declaration of symbol 'index' hides symbol 'index(const char *, int)' (file ..\arm-none-eabi-gcc\4.5.2\arm-none-eabi\include\string.h)


Pro:
Thanks! We should also try to add linting tests to the CI again.

PR #2165, opened 2018-10-31
Add wrapper macros for datatype and nodeset generator

Pro:
This PR adds some wrapper macros for the `nodeset_compiler.py` and `generate_datatypes.py` script.

This makes custom nodeset generation easier for users.


jpfr:
👍

PR #2179, opened 2018-11-06
[Review] Config: Add custom logger initialization when creating server config

### LINKED ISSUE #1836: Unable to set logger when creating server
Consider the following code (from the tutorial examples):

```cpp
UA_Boolean running = true;
UA_ServerConfig *config = UA_ServerConfig_new_default();
config->logger = &my_fancy_logging_function;
UA_Server_run(server, &running);
// ...
```

I want to set a custom logging function (`my_fancy_logging_function`).

This is not possible since `UA_ServerConfig_new_default()` already uses `UA_Log_Stdout` to initialize the network layer. So setting `config->logger` after `UA_ServerConfig_new_default()` doesn't have any effect.

Is this a known flaw? Does this exceed the "default" config and so I need generating my own config manually?

References: https://github.com/open62541/open62541/issues/1139, https://github.com/open62541/open62541/pull/1566, https://github.com/open62541/open62541/pull/1641, @janitza-thbe, @jpfr, @Pro

### LINKED ISSUE #2132: Need a logger parameter added to all the UA_ServerConfig create functions
Description
===========
If you use a custom logger function you potentially miss log calls generated inside any of the `UA_ServerConfig` create functions.

Background Information
The problem is that when calling a `UA_ServerConfig` create function e.g. `UA_ServerConfig_new_minimal()`  the `conf->logger` field is set to the default logger function internally which then may be called multiple times before the create function returns.

This field can only be updated **after** the create function returns
`conf->logger=my_logger;`

To fix this a logger parameter must be  added to all the `UA_ServerConfig` create functions.

Checklist
=========

 - [x] open62541 Version: 0.3-rc2
 - [x] Operating system: All
 - [ ] Critical issue

Pro:
Fixes #1836 and fixes #2132

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -329,11 +329,13 @@ addDefaultNetworkLayers(UA_ServerConfig *conf, UA_UInt16 portNumber, UA_UInt32 s
 }
 
 UA_ServerConfig *
-UA_ServerConfig_new_customBuffer(UA_UInt16 portNumber,
+UA_ServerConfig_new_customLogger(UA_UInt16 portNumber,
                                  const UA_ByteString *certificate,
                                  UA_UInt32 sendBufferSize,
-                                 UA_UInt32 recvBufferSize) {
+                                 UA_UInt32 recvBufferSize, UA_Logger logger) {
     UA_ServerConfig *conf = createDefaultConfig();
+    if (logger != NULL)
+        conf->logger = logger;
 
     UA_StatusCode retval = UA_Nodestore_default_new(&conf->nodestore);
     if(retval != UA_STATUSCODE_GOOD) {
--- a/plugins/ua_config_default.h
+++ b/plugins/ua_config_default.h
@@ -25,6 +25,30 @@ extern const UA_EXPORT UA_ConnectionConfig UA_ConnectionConfig_default;
 /* Default Server Config */
 /*************************/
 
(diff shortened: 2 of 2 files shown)


jpfr:
I agree that we need to work in issue #2132.

But it would be unwise to let the config-creation functions go the way of the crappy Win32-API.
With many many arguments that are mostly left to NULL.
Where do we stop adding more arguments for possible configuration-items?

Can we encourage people to make changes to the returned config?
We could have a UA_ServerConfig_setLogger.
But that is overkill. Since everybody can just set the variable.

Pro:
> Can we encourage people to make changes to the returned config?
> We could have a UA_ServerConfig_setLogger.
> But that is overkill. Since everybody can just set the variable.

I agree with you, that users should use the returned config. The problem is, that within that method, the network layers for example already use the logger which is created by the default config. So there is currently no way to change the logger, e.g., to a file logger and log all the messages, including the network layer output.

Otherwise we need to think of a two way configuration step, where we firstly only return the default config, the user can then change the config, and then we initialize the network layers. With this we can also put the buffer size into the config, without passing it as a parameter.

jpfr:
Okay. I agree with you.
Let's talk about a config overhaul when 0.3 is out.

Pro:
Closing in favour of #2263

Pro:
Closing this in favour of #2263 which correctly handles the logger initialization also for the network layer

PR #2180, opened 2018-11-06
Build using clang on Windows

escherstair:
This PR adds a job to AppVeyor so that it builds using clang on Windows


escherstair:
Commit 246cfc9f6a7dfb7848e73ee15d789709a8ce644b fixes the issue but now a different problem comes from clang.
```

[ 13%] Building C object CMakeFiles/open62541-plugins.dir/arch/ua_network_tcp.c.obj
C:\projects\open62541\arch\ua_network_tcp.c:432:37: error: format specifies type 'int' but the argument has type 'SOCKET' (aka 'unsigned long long') [-Werror,-Wformat]
                    (int)newsockfd, layer->serverSockets[i]);
                                    ^~~~~~~~~~~~~~~~~~~~~~~
C:\projects\open62541\arch\ua_network_tcp.c:445:26: error: format specifies type 'int' but the argument has type 'SOCKET' (aka 'unsigned long long') [-Werror,-Wformat]
                         e->connection.sockfd);
                         ^~~~~~~~~~~~~~~~~~~~
C:\projects\open62541\arch\ua_network_tcp.c:458:21: error: format specifies type 'int' but the argument has type 'SOCKET' (aka 'unsigned long long') [-Werror,-Wformat]
                    e->connection.sockfd);
                    ^~~~~~~~~~~~~~~~~~~~
C:\projects\open62541\arch\ua_network_tcp.c:471:25: error: format specifies type 'int' but the argument has type 'SOCKET' (aka 'unsigned long long') [-Werror,-Wformat]
                        e->connection.sockfd);
                        ^~~~~~~~~~~~~~~~~~~~
4 errors generated.
```
It seems that clang warns if variadic parameters of
`UA_LOG_TRACE(UA_Logger logger, UA_LogCategory category, const char *msg, ...) `
are not int.

Does someone has an idea on how to fix this?

I see that for another project, [the fix](https://github.com/apple/turicreate/pull/775/files#diff-cd268334767f54a1764d571e7b07f2c4) was to add
```
COMPILE_FLAGS_EXTRA_CLANG
    -Wno-format
```
in CMakeLists.txt

But I'm not sure about this approach because a wrong handling of data type with format specifiers can produce Undefined Behavior (in some circumstances).

Pro:
Since these are only (trace) outputs to the command line, you can also just cast the sockfd.

This is the current code:

```
        UA_LOG_TRACE(layer->logger, UA_LOGCATEGORY_NETWORK,
                    "Connection %i | New TCP connection on server socket %i",
(int)newsockfd, layer->serverSockets[i]);
```

As you can see `newsockfd` is already casted, so it would make sense to also cast `serverSockets` to int.

escherstair:
Forgot to mention that explicit casting is the first thing that came to my mind.
After I did it, a lot of similar errors came out from other files, related to `UA_LOG_TRACE_CHANNEL `macro.
I'm not sure if this is the right way to proceed, but I can give it a try.

escherstair:
Now the build on AppVeyor works and it adds clang compiler.
I think the build error on travis is one of the random errors discussed in a [separate issue](https://github.com/open62541/open62541/issues/2195).
Could someone retrigger a travis build?

escherstair:
From my side everything is ok.

jpfr:
Great! Thanks for the contribution.
I see this was a lot of work.

PR #2184, opened 2018-11-07
[0.3] Add Boolean to the numeric data types author: @jpfr

StalderT:
#2181


StalderT:
> It includes a public API change and the API of 0.3 was already freezed.

I did not pay attention. With my all apologies. It's fixed.

PR #2187, opened 2018-11-08
[Review] Architecture: Allow overriding memory allocation functions

Pro:
With this change someone can override malloc/free/realloc/calloc with his own implementation by using preprocessor defines.

An example usage is found in: #2188

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -440,6 +440,7 @@ include_directories(${PROJECT_SOURCE_DIR}/include
 
 set(exported_headers ${exported_headers}
                      ${PROJECT_BINARY_DIR}/src_generated/ua_config.h
+                     ${PROJECT_SOURCE_DIR}/arch/ua_architecture_base.h
                      ${ua_architecture_headers_beginning}
                      )
 
--- a/arch/posix/ua_architecture.h
+++ b/arch/posix/ua_architecture.h
@@ -10,6 +10,8 @@
 #ifndef PLUGINS_ARCH_POSIX_UA_ARCHITECTURE_H_
 #define PLUGINS_ARCH_POSIX_UA_ARCHITECTURE_H_
 
+#include "ua_architecture_base.h"
+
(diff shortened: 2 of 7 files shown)


jpfr:
This was possible before as well. And we used it in projects.

Now we have to edit
 arch/ua_architecture_base.h
instead of
arch/vxworks/ua_architecture.h

How is that better?

Pro:
Can you post a quick example how you did it before?
Couldn't think of any good way to do it WITHOUT changing anything in the code, and I would be happy to remove this change.

With this PR you do not need to change any code at all (which is not possible for the fuzzer tests for example).
You can override the malloc by adding `-DUA_malloc=UA_my_override`

See the referenced PR for an example

jpfr:
Hey @Pro ,

by doing `-DUA_malloc=UA_my_override`, you still have to include some header that defines `UA_my_override`. Isn't that a code change?

The commit is fine. I just fear the complexity of /arch in general.
I do not fully understand the architecture mechanism in all detail any more.

Pro:
The nice thing here is that you do not need to make a code change.

https://github.com/open62541/open62541/blob/014f9e9cffeda8ed9919148af75154e74e30b019/arch/ua_architecture_base.h

is included first.

If you use `-DUA_malloc=UA_my_override` this header file will automagically add the definition for that function. 

E.g. have a loom at
```
#ifdef UA_malloc
void* UA_malloc(unsigned long size);
#endif
```

Since UA_malloc is defined, it will be changed by the preprocessor to:

```
#ifdef UA_malloc
void* UA_my_override(unsigned long size);
#endif
```

And within the architectures, the definition of malloc will be skipped since it is already defined.

With this you only need to include your own .c file which implements that function.

If this is OK to merge, than I can extend the arch documentation to clarify that concept. Also please write where you have issues with understanding the concept, because then probably also others have issues there. I will extend the docs accordingly.

jpfr:
Ok to merge.

PR #2188, opened 2018-11-08
[Review] Out of memory simulation for the fuzzer

Pro:
Based on #2187 we define our own memory allocation wrapper and keep track of current amount of allocated memory.

The fuzzer can control the available memory with its byte array: The last four bytes represent a 32-Bit number representing the available memory size.

With this approach the fuzzer should be able to find issues in the paths where memory allocation fails. These paths are currently not checked.

See also: https://github.com/google/oss-fuzz/issues/1675


jpfr:
Nice!

Pro:
To merge this we first need to find a better solition for #2187.

Depends on #2187

Gfast2:
Just drop by to say Hello.
This Repo is actually my Favorites! I've learn tons of stuff from here. I hope some day I can do something on this.

Issue #2193, opened 2018-11-08
Method call within a FunctionalGroup not work


mlang-de:
The DI specification (Device Intergration - Part 100, 1.01) allows to reference a method from a FunctionalGroup.
![image](https://user-images.githubusercontent.com/20433237/48215177-f83f8980-e381-11e8-9f2e-c029cea90ed7.png)
![image](https://user-images.githubusercontent.com/20433237/48215824-5faa0900-e383-11e8-8fac-b5d3ba63523e.png)
The current implementation in function callWithMethodAndObject() verify:
![image](https://user-images.githubusercontent.com/20433237/48215245-2329dd80-e382-11e8-9df4-a66d92093d61.png)

Is it meaningful to implement a logic like:
1. parent object must have a HasTypeDefinition reference to FunctionalGroupType
2. parent object must have a Organize reference to the method

This kind of information modelling will be used at least in some Companion Specification for now.

jpfr:
#2194 was merged.

PR #2194, opened 2018-11-09
[Review] Add capability to call/reference a method from a FunctionalGroup

mlang-de:
See issue #2193 for further details.

Should this part be switchable in cmake?
[Update] In release mode with gcc 4.5.2 the difference is 0,4kB.

### DIFF
--- a/src/server/ua_services_method.c
+++ b/src/server/ua_services_method.c
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
  *    Copyright 2015 (c) Chris Iatrou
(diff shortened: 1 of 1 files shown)


jpfr:
I'm a bit hesitant to put specifics of the DI namespace into the core service definition.
But we should support the case obviously.

Let's put this on the agenda for the conference call.

Pro:
image

From the spec:
`A method shall be a component of an Object`

mlang-de:
DI specify:
image
[...] FunctionalGroupType is used to organize Parameters and Methods [...]

mlang-de:
@jpfr @Pro How could this PR be handled? Some preferred solution like a plugin/callback which could be configured?

mlang-de:
@jpfr Should I do some defines around this parts? So an user can decide via the CMake file to use the DI-method support or not.

jpfr:
Give a clear reference to the DI spec in the comments.
Mention that this is a *workaround* done only for DI.

jpfr:
This is still earmarked for v1.1.
Can be merged when the CI goes green.

jpfr:
Finally merged.
Thanks for being patient. 👍

PR #2200, opened 2018-11-09
[WIP] PubSub information model extension

andreasebner:
This feature update introduces the PubSub configuration over the information model methods. In addition the existing pubsub information model functionality will be extended. The new UA_PUBSUB_INFORMATIONMODEL_METHODS Flags enables the pubsub methods.

Currently the PubSubConnection representation was extended and the corresponding methods are added. Tests for the information model driven PubSub configuration are also part of this PR.


jpfr:
I did a test drive with this. Looks good!

Only one thing. We still have a loop where the writer can be browsed infinitely deep.
The OPC standard forbids directed cycles in the information model.

And the Utthunga tool seems to crash because of this.

andreasebner:
> 
> 
> I did a test drive with this. Looks good!
> 
> Only one thing. We still have a loop where the writer can be browsed infinitely deep.
> The OPC standard forbids directed cycles in the information model.
> 
> And the Utthunga tool seems to crash because of this.

The infinity loop problem is solved and the Utthunga configuration tool is working. This means you can connect to the server with the tool and browse over the open62541-API created PubSub information model nodes. Start e.g. with tutorial_pubsub_publish and connect with the Utthunga tool to the server. The functionality of the configuration tool methods will be increasly added with this PR.

PR #2203, opened 2018-11-10
Prepare python script for usage of define UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT

mlang-de:
Only the inserted dataTypes in list whitelistFuncAttrWarnUnusedResult will be generated with the define UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT in front of the function declaration.
In file ua_types_generated_handling.h, for example:
```C
UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT static UA_INLINE UA_StatusCode
UA_ByteString_copy(const UA_ByteString *src, UA_ByteString *dst) {
    return UA_copy(src, dst, &UA_TYPES[UA_TYPES_BYTESTRING]);
}
```

### DIFF
--- a/tools/generate_datatypes.py
+++ b/tools/generate_datatypes.py
@@ -55,6 +55,8 @@ builtin_overlayable = {"Boolean": "true",
                        "offsetof(UA_Guid, data3) == (sizeof(UA_UInt16) + sizeof(UA_UInt32)) && " + \
                        "offsetof(UA_Guid, data4) == (2*sizeof(UA_UInt32)))"}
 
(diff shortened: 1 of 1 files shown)


jpfr:
Great! Just had to retrigger one of the CI runs.

So what is the plan now? Gradually add types to `whitelistFuncAttrWarnUnusedResult` and fix the warnings?

mlang-de:
Exaclty, I would prefer for each function modification an own PR?

PR #2202, opened 2018-11-09
prevent reentrant renew securechannel

StalderT:
prevent reentrant renew secure channel

fix #2199 #1844

PR #2206, opened 2018-11-11
[master] prevent reentrant renew securechannel

StalderT:
#2202


jpfr:
This has already received a "clean fix" on master.
Chunks are already processed in order.
Also if they renewSecureChannel was called "reentant".

See also the chunks handling in the SecureChannel.

PR #2211, opened 2018-11-13
Fix: Do not ignore the first network adapter in mdns

ccvca:
The first adapter was ignored by mdns.

### DIFF
--- a/src/server/ua_mdns.c
+++ b/src/server/ua_mdns.c
@@ -279,7 +279,7 @@ mdns_is_self_announce(UA_Server *server, struct serverOnNetwork_list_entry *entr
 
 #ifdef _WIN32
         /* Iterate through all of the adapters */
-        IP_ADAPTER_ADDRESSES* adapter = adapter_addresses->Next;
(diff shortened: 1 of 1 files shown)


ccvca:
Ok, looks like now the test run into this problem: #2211 and the server could not be found anymore.

jpfr:
Should we add another filter rule? Why can’t we use the first network layer?

ccvca:
The Problem is, that the current implementation ignores the first avaliable networkinterface, which in my case is the Ethernet-Port of my Laptop. So the mdns-Package does not contain the correct IP-Adress.

adapter_address is a single linked list, which is incremented before the for-loop in https://github.com/open62541/open62541/blob/767093a63d631726cd4ef3b846d324c80d5467e5/src/server/ua_server_discovery_mdns.c#L282-L283 So the first interface is not considered.

Pro:
Honestly, I'm not sure why I skipped the first adapter. I guess that is was not intentional, so this should be fixed.

ccvca:
As this request breaks the CI, I would suggest to merge this after/combined with the fix of #2212.

Pro:
Incuded in #2344

PR #2213, opened 2018-11-14
some server have problems with RequestHeader timeoutHint=0

jrcoding:
if requestHeader timeoutHint is '0' the corresponding server should treat it as no timeout, but some server misbehave and really treat it as zero timeout, which is of course a fault in the server.

To make the client more tolerant a large timeoutHint, i.e. 60000, would help.


jpfr:
Mered. Thanks!

Issue #2216, opened 2018-11-15
Shadowed newEntry in UA_NodeMap_insertNode


cabralfortiss:
Description
===========
The newEntry variable in UA_NodeMap_insertNode is named the same as a function which produces shadow errors in some compilers. 

Background Information / Reproduction Steps
===========================================
A very similar issue is https://github.com/open62541/open62541/issues/1250#issuecomment-336565666 which was fixed with https://github.com/open62541/open62541/commit/c86c8e833f2e0911f64727a059a87a6f3102bb6b

jpfr:
Fixed via #2217

PR #2217, opened 2018-11-15
Rename newEntry that produces shadow errors in some compilers

cabralfortiss:
The newEntry variable in UA_NodeMap_insertNode is named the same as a function which produces shadow errors in some compilers. Now, the name is similar to UA_NodeMap_replaceNode.

This is for issue #2216


jpfr:
Thanks!

Issue #1092, opened 2017-06-29
Cannot connect to server with CTT


Infinity95:
Description
===========
With the current master branch i cannot connect to the server with the Compliance Test Tool anymore. (BadCommunicationError)
Maybe this has something to do with the added ipv6 support? (Connecting with UAExpert works)

I'm using the config from the ctt repo: https://github.com/open62541/open62541-ctt
As far as i checked i don't think the server ever recieves any message.

Background Information / Reproduction Steps
===========================================
build master, run server and try running the ctt.

Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): fadd1dd09d8689e609339467a9c36651e19daa69
 - [x] Other OPC UA SDKs used (client or server):  UA 1.02 Compliance Test Tool
 - [x] Operating system: Windows 10
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jpfr:
I cannot confirm. Works fine on my box with the current config from the repository.
Can you remove your CMake build directory and configure+build from scratch?
Sometimes it gets confused when files are renamed, and so on.

Infinity95:
I will check tomorrow, and see if that does the trick. (I thought i already did a clean build, but i'm not sure)
Although i find it strange, that this would only affect the CTT tool, since UAExpert works just fine.

Infinity95:
I just deleted my build directory and rebuilt from scratch. I then did a fresh clone of the open62541-ctt repo and loaded the config into the CTT. I then started the server and tried connecting with the CTT, and again all i got was a BadCommunicationError. 

Connecting with UAExpert still works. I also found out, that the client example cannot connect to the server as well. This results in the error message:
`[06/30/2017 13:13:42.143] warning/network       Connection to opc.tcp://localhost:16664 failed. Error: 0: Der Vorgang wurde erfolgreich beendet.`

Infinity95:
Ok, i found the reason it wasn't working. This was entirely my fail. I still had another opc tool running that was blocking the 4840 port, and since it was on autostart, it of course also blocked the port after restarting. Now that it's gone, everything works fine.

Infinity95:
[...]

PR #2233, opened 2018-11-20
change port used by Server and Client examples

escherstair:
Readme.md contains **Example Server Implementation** and **Example Client Implementation** that uses a default port.
Port 4840 is the standard one for the OPC UA server discovery.
I think it's better to use 16664 that is the one used in
https://github.com/open62541/open62541/blob/7b3824f8d451f7b1025bee774efbbe81e350b370/examples/server.c#L116

Using port 4840 brings to an issue like the one described in #1092 
`warning/network Error during binding of the server socket`

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ int main(int argc, char** argv)
 {
     signal(SIGINT, signalHandler); /* catch ctrl-c */
 
-    /* Create a server with one network layer listening on port 4840 */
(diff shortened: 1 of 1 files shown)


Pro:
I am against changing it to 16664.

The examples should also use the default port 4840, since that's the one specified.

Also, you want to merge into 0.2 branch which will only changed for really necessary bugfixes.

PR #2237, opened 2018-11-23
[Review] PubSub: Only add timer if interval > 0

Pro:
The callback should only be added if publishingInterval is set to > 0.

@andreasebner please check

Also note, that the call to `UA_WriterGroup_publishCallback(server, writerGroup);` will do nothing, since the writerGroup does not have a writer right after creation and in the tutorial the writer is added afterwards.

Later on we may need a User-API method which allows to manually publish one single data set (without interval).

### DIFF
--- a/src/pubsub/ua_pubsub.c
+++ b/src/pubsub/ua_pubsub.c
@@ -1152,12 +1152,15 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
  */
 UA_StatusCode
 UA_WriterGroup_addPublishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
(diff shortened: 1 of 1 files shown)


jpfr:
This is covered implicitly.

When the interval is negative, then addRepeatedCallback returns BADINTERNALERROR.
In that case, I also would not call UA_WriterGroup_publishCallback before returning the status code.

See alos here: #2239.

Pro:
> When the interval is negative, then addRepeatedCallback returns BADINTERNALERROR.

Hm, yes... I added this because passing 0 as the PublishInterval (i.e. manually publish a data set) would return an error code.

But I just saw, that you then could just skip calling `UA_WriterGroup_addPublishCallback`.

I'll close as wontfix.

PR #2238, opened 2018-11-23
[WIP] - Support of a subset of part 8 "DataAccess"

Nik625:
With the content of this pull requests a subset of the OPC UA Specification Part 8 "DataAccess" is supported. The subset is defined among the following Conformance Units (as defined in the Compliance Test Tool): _AnalogItemType_, _AnalogArrayType_, _DataItemType_, _DiscreteItemType_, _SemanticsChanged_ and _PercentDeadbandFilter_. The "_ValuePrecision_" for variables of type "_DateTime_" is only implemented as a very first draft.

See also #2267 for the feature branch.

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -380,79 +380,52 @@ ReadWithNode(const UA_Node *node, UA_Server *server, UA_Session *session,
     }
 }
 
-static UA_StatusCode
-Operation_Read(UA_Server *server, UA_Session *session, UA_MessageContext *mc,
-               UA_TimestampsToReturn timestampsToReturn, const UA_ReadValueId *id) {
-    UA_DataValue dv;
-    UA_DataValue_init(&dv);
-
+static void
+Operation_Read(UA_Server *server, UA_Session *session, UA_ReadRequest *request,
+               UA_ReadValueId *rvi, UA_DataValue *result) {
     /* Get the node */
-    const UA_Node *node = UA_Nodestore_getNode(server->nsCtx, &id->nodeId);
+    const UA_Node *node = UA_Nodestore_getNode(server->nsCtx, &rvi->nodeId);
 
     /* Perform the read operation */
     if(node) {
-        ReadWithNode(node, server, session, timestampsToReturn, id, &dv);
+        ReadWithNode(node, server, session, request->timestampsToReturn, rvi, result);
+        UA_Nodestore_releaseNode(server->nsCtx, node);
     } else {
-        dv.hasStatus = true;
-        dv.status = UA_STATUSCODE_BADNODEIDUNKNOWN;
+        result->hasStatus = true;
+        result->status = UA_STATUSCODE_BADNODEIDUNKNOWN;
     }
-
-    /* Encode (and send) the results */
-    UA_StatusCode retval = UA_MessageContext_encode(mc, &dv, &UA_TYPES[UA_TYPES_DATAVALUE]);
-
-    /* Free copied data and release the node */
-    UA_Variant_deleteMembers(&dv.value);
-    UA_Nodestore_releaseNode(server->nsCtx, node);
-    return retval;
 }
 
-UA_StatusCode Service_Read(UA_Server *server, UA_Session *session, UA_MessageContext *mc,
-                           const UA_ReadRequest *request, UA_ResponseHeader *responseHeader) {
-    UA_LOG_DEBUG_SESSION(&server->config.logger, session,
-                         "Processing ReadRequest");
+void
+Service_Read(UA_Server *server, UA_Session *session,
+             const UA_ReadRequest *request, UA_ReadResponse *response) {
+    UA_LOG_DEBUG_SESSION(&server->config.logger, session, "Processing ReadRequest");
 
     /* Check if the timestampstoreturn is valid */
-    if(request->timestampsToReturn > UA_TIMESTAMPSTORETURN_NEITHER)
-        responseHeader->serviceResult = UA_STATUSCODE_BADTIMESTAMPSTORETURNINVALID;
-
-    if(request->nodesToReadSize == 0)
-        responseHeader->serviceResult = UA_STATUSCODE_BADNOTHINGTODO;
+    if(request->timestampsToReturn < 0 ||
+       request->timestampsToReturn > UA_TIMESTAMPSTORETURN_NEITHER) {
+        response->responseHeader.serviceResult = UA_STATUSCODE_BADTIMESTAMPSTORETURNINVALID;
+        return;
+    }
 
     /* Check if maxAge is valid */
-    if(request->maxAge < 0)
-        responseHeader->serviceResult = UA_STATUSCODE_BADMAXAGEINVALID;
+    if(request->maxAge < 0) {
+        response->responseHeader.serviceResult = UA_STATUSCODE_BADMAXAGEINVALID;
+        return;
+    }
 
     /* Check if there are too many operations */
     if(server->config.maxNodesPerRead != 0 &&
-       request->nodesToReadSize > server->config.maxNodesPerRead)
-        responseHeader->serviceResult = UA_STATUSCODE_BADTOOMANYOPERATIONS;
-
-    /* Encode the response header */
-    UA_StatusCode retval =
-        UA_MessageContext_encode(mc, responseHeader, &UA_TYPES[UA_TYPES_RESPONSEHEADER]);
-    if(retval != UA_STATUSCODE_GOOD)
-        return retval;
-
-    /* Process nothing if we return an error code for the entire service */
-    UA_Int32 arraySize = (UA_Int32)request->nodesToReadSize;
-    if(responseHeader->serviceResult != UA_STATUSCODE_GOOD)
-        arraySize = 0;
-
-    /* Process all ReadValueIds */
-    retval = UA_MessageContext_encode(mc, &arraySize, &UA_TYPES[UA_TYPES_INT32]);
-    if(retval != UA_STATUSCODE_GOOD)
-        return retval;
-
-    for(UA_Int32 i = 0; i < arraySize; i++) {
-        retval = Operation_Read(server, session, mc, request->timestampsToReturn,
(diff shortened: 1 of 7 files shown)


arch/ua_network_tcp.c
```diff
@@ -525,20 +525,22 @@ UA_ServerNetworkLayer
 UA_ServerNetworkLayerTCP(UA_ConnectionConfig config, UA_UInt16 port, UA_Logger logger) {
     UA_ServerNetworkLayer nl;
     memset(&nl, 0, sizeof(UA_ServerNetworkLayer));
+    nl.deleteMembers = ServerNetworkLayerTCP_deleteMembers;
```
basyskom-meerkoetter: does this "belong" into this series or is this something to break out into a PR of its own?
This is a fix to an overall(?) issues found by oss-fuzz?

examples/tutorial_server_dataaccess.c
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Integrating a DataAccess example VarieableType with Properties
+ * DataAccess defines new Variable Types and Properties for describing real time data.
+ * In this example a VariableType node is added with its respective DataAccess Properties
+*/
+#include "open62541.h"
+#include <signal.h>
+#include "ua_server_dataaccess.h"
+
+UA_Boolean running = true;
```
basyskom-meerkoetter: static UA_Boolean running = true
to limit the scope of the symbol to this translation unit (just like it is already done for the functions)?

examples/tutorial_server_dataaccess.c
```diff
@@ -0,0 +1,91 @@
...
+int
+main(void) {
+    signal(SIGINT,  stopHandler);
+    signal(SIGTERM, stopHandler);
+
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    UA_ServerConfig *config = UA_ServerConfig_new_default();
+    UA_Server *server = UA_Server_new(config);
+
+    retval |= addDataAccessNode(server);
```
basyskom-meerkoetter: does it make sense to enter the server loop if addDataAccessNode() failed?

examples/tutorial_server_dataaccess.c
```diff
@@ -0,0 +1,91 @@
...
+    retval |= UA_Server_run(server, &running);
+
+    UA_Server_delete(server);
+    UA_ServerConfig_delete(config);
+    return (int)retval;
+
```
basyskom-meerkoetter: superfluous empty line (before })

src/server/ua_services_attribute.c
```diff
@@ -1050,6 +1058,270 @@ writeValueAttributeWithRange(UA_VariableNode *node, const UA_DataValue *value,
...
+                const UA_VariableNode* fNode = (const UA_VariableNode*) UA_Nodestore_get(server, &node->nodeId);
+                newNotification->data.value.hasValue = true;
+                newNotification->mon = mon;
+                UA_DataValue_copy(&(fNode->value.data.value), &newNotification->data.value);
+                newNotification->data.value.hasStatus = true;
+                newNotification->data.value.status = UA_STATUSCODE_GOOD;
+                UA_Nodestore_release(server, (const UA_Node*)fNode);
+                /* Enqueue the new notification */
+                UA_Notification_enqueue(server, sub, mon, newNotification);
+			    return UA_STATUSCODE_GOOD;
```
basyskom-meerkoetter: return is indented wronly

arch/ua_network_tcp.c
basyskom-meerkoetter: does this "belong" into this series or is this something to break out into a PR of its own?
This is a fix to an overall(?) issues found by oss-fuzz?

examples/tutorial_server_dataaccess.c
basyskom-meerkoetter: static UA_Boolean running = true
to limit the scope of the symbol to this translation unit (just like it is already done for the functions)?

examples/tutorial_server_dataaccess.c
```diff
@@ -0,0 +1,91 @@
...
+main(void) {
+    signal(SIGINT,  stopHandler);
+    signal(SIGTERM, stopHandler);
+
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    UA_ServerConfig *config = UA_ServerConfig_new_default();
+    UA_Server *server = UA_Server_new(config);
+
+    retval |= addDataAccessNode(server);
+    retval |= UA_Server_run(server, &running);
```
basyskom-meerkoetter: does it make sense to enter the server loop if addDataAccessNode() failed?

examples/tutorial_server_dataaccess.c
basyskom-meerkoetter: superfluous empty line (before })

src/server/ua_services_attribute.c
```diff
@@ -1050,6 +1058,270 @@ writeValueAttributeWithRange(UA_VariableNode *node, const UA_DataValue *value,
...
+    if(UA_STATUSCODE_GOOD == bpr.statusCode || bpr.targetsSize >= 1) {
+        UA_Variant tVar;
+        UA_Variant_init(&tVar);
+        retval = UA_Server_readValue(server, bpr.targets->targetId.nodeId, &tVar);
+        if(retval == UA_STATUSCODE_GOOD) {
+            UA_Range* euRange = (UA_Range*) tVar.data;
+            retval = UA_Variant_checkRange(&(value->value), euRange);
+            if(retval == UA_STATUSCODE_GOOD) {
+                writeValueAttributeWithoutRange(node, &adjustedValue);
+                enqueueSemanticChangedNotification(server, node, session, UA_STATUSCODE_GOOD| 0x4000);
```
basyskom-meerkoetter: what is | 0x4000?

jpfr:
Hello there,

can you indicate when another round of reviews is required?

Pro:
When #2267 is merged, this should be rebased onto master

Nik625:
Unfortunately due to a lot of work, this will be continued approximately in mid of february from my side.

Nik625:
Regarding the fragmentation into several feature PRs i only see the **SemanticsChanged** feature existing alongside the DataAccess extension. Part 8 of the specification  references this as _"Servers that implement Data Access shall set this Bit in notifications if certain Properties defined in this standard change."_ and the changed properties where the bit is set are DataAccess specific properties. Thus, I see this feature integrated in the general DataAccess functionalities.

Nik625:
@jpfr Maybe you can explain again what you meant with splitting up Core DA functionalities and the information model. Many core functionalities rely on the existence of the VariableTypes defined in the informationmodel. Thanks in advance!

jpfr:
`UA_ENABLE_DA` is a very coarse-grained build flag.
It could be that some people want deadband filtering. But **not** the full gamut of DA features.

This could be `UA_ENABLE_SUBSCRIPTIONS_DEADBAND` or similar.

Then we can have a list of individual features that collectively cover the DA specification.
And maybe a meta-build-flag `UA_ENABLE_DA` that enables all of them.

Can you make a list of such feature-flags?
Then we can merge the features individually.

I would like to avoid merging a 5k line monster PR in one go.

Nik625:
image


The build flags are defined as follows:

- **UA_ENABLE_DA** remains as the meta-build flag for the entire DA functionality (this requires _UA_ENABLE_DA_INFORMATIONMODEL_ to be enabled to depict the demo nodes in the address space)
- **UA_ENABLE_DA_INFORMATIONMODEL** enables the xml-descriptions of the types and sets the informationmodel. No other enabled build flags are compulsory for this functionality to work.
- **UA_ENABLE_DA_ANALOGITEMS** enables only the AnalogItemType with its underlying properties but also sets the required informationmodel. The datatypes are referenced by a separate file datatypes_da_analog.txt. This functionality requires the informationmodel to depict the demo nodes in the address space. If no nodes are created in the adress space, this functionality would still be available but would obviously have no use. This is why it should not be activated without enabling - _UA_ENABLE_DA_INFORMATIONMODEL_.

- **UA_ENABLE_DA_SEMANTICCHANGED** enables the enqueuing of new data change notifications with their semanticchanged bit set in the statuscode. This function requires the dataypes to be referenced by the respective file (datatypes_da_analog.txt.). In its core functionality the implementation checks for the existence of relevant underlying properties of the AnalogItemType which are modelled by the informationmodel (such as _Range_). Thus _UA_ENABLE_DA_INFORMATIONMODEL_ needs to be switched on.
- **UA_ENABLE_SUBSCRIPTIONS_DEADBAND** enables the percent deadband data change filter for MonitoredItems in Subscriptions. The percentage is relative to the defined range property and thus requires the range datatype to be enabled through setting the datatype file (e.g _UA_ENABLE_ANALOGITEMS_). The user needs to create this property as a variable node and thus needs the xml definition of the according range datatype (Enable of _UA_ENABLE_DA_INFORMATIONMODEL_ is required in this case aswell).
- The _greyed out build flags_ are shown to fulfill the depiction. Their mere representation is again coppled to the informationmodel, but besides these flags do not define new datatypes that could exist independent of the other parts. Thus, integrating such build flags is not recommended.

The given structure makes it difficult to find appropriate, independent build flags to separate the functionality. Separating the different datatypes would still result in the need of integrating a partial informationmodel in order to be able to actually create these datatypes and use the functionalities behind them.

Ultimately, a separation into independent features could be realized internally, but users would need to enable more than their desired build flag to benefit from the functionalities.

jpfr:
@Nik625 Great to see this pick up again!

The PR is marked as [Review]. There are now a lot of "fix commits".
Can you squash/rebase them out to have a clean history?

The dependency graph you show is really nice. Because it is a DAG (directed acyclic graph).
So we can have features that "build on top of each other".
It should be possible to have one commit per feature in this PR:

- Rebase on the current master
- Squash all commits into a single commit.
- Do a `git reset --soft HEAD^`, then a `git reset` to get all changes as "unadded/unstaged changes"
- Add selectively with `git add -p` and commit

ccvca:
@Nik625 Do you closed this by accident?

PR #2247, opened 2018-11-28
WIP: Separate SecurityPolicies from Endpoints in the server

jonasgreen88:
Currently the SecurityPolicies supported by the server are strongly
tied together with the endpoints. To be able to reuse the
SecurityPolicy instances of the server also in the Access Control
plugin to handle encrypted passwords the SecurityPolicies are moved
to the server config struct.

### DIFF
--- a/include/ua_plugin_securitypolicy.h
+++ b/include/ua_plugin_securitypolicy.h
@@ -111,11 +111,12 @@ typedef struct {
      * key.
      *
      * @param securityPolicy the securityPolicy the function is invoked on.
-     * @param channelContext the channelContext which contains information about
-     *                       the keys needed to decrypt the message.
+     * @param securityPolicyContext the securityPolicyContext which contains
+     *                              information about the keys needed to decrypt
+     *                              the message.
      * @param data the data to decrypt. The decryption is done in place. */
     UA_StatusCode (*decrypt)(const UA_SecurityPolicy *securityPolicy,
-                             void *channelContext,
+                             void *securityPolicyContext,
                              UA_ByteString *data) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
 
     /* Returns the length of the key used locally to encrypt messages in bits
@@ -366,11 +367,6 @@ struct UA_SecurityPolicy {
     void (*deleteMembers)(UA_SecurityPolicy *policy);
 };
 
-typedef struct {
-    UA_SecurityPolicy securityPolicy;
-    UA_EndpointDescription endpointDescription;
-} UA_Endpoint;
-
 /* Gets the number of bytes that are needed by the encryption function in
  * addition to the length of the plaintext message. This is needed, since
  * most RSA encryption methods have their own padding mechanism included.
@@ -386,6 +382,14 @@ UA_SecurityPolicy_getRemoteAsymEncryptionBufferLengthOverhead(const UA_SecurityP
                                                               const void *channelContext,
                                                               size_t maxEncryptionLength);
 
+/* Gets the a pointer to the context of a security policy supported by the
+ * server matched by the security policy uri.
+ *
+ * @param server the server context.
+ * @param securityPolicyUri the security policy to get the context of. */
+UA_SecurityPolicy *
+UA_SecurityPolicy_getSecurityPolicyByUri(const UA_Server *server,
+                                         UA_ByteString *securityPolicyUri);
 
 typedef UA_StatusCode
 (*UA_SecurityPolicy_Func)(UA_SecurityPolicy *policy,
--- a/include/ua_plugin_securitypolicy.h
+++ b/include/ua_plugin_securitypolicy.h
@@ -111,12 +111,11 @@ typedef struct {
      * key.
      *
      * @param securityPolicy the securityPolicy the function is invoked on.
-     * @param securityPolicyContext the securityPolicyContext which contains
-     *                              information about the keys needed to decrypt
-     *                              the message.
+     * @param channelContext the channelContext which contains information about
+     *                       the keys needed to decrypt the message.
      * @param data the data to decrypt. The decryption is done in place. */
     UA_StatusCode (*decrypt)(const UA_SecurityPolicy *securityPolicy,
-                             void *securityPolicyContext,
+                             void *channelContext,
                              UA_ByteString *data) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
 
     /* Returns the length of the key used locally to encrypt messages in bits
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -333,6 +333,8 @@ UA_Server_updateCertificate(UA_Server *server,
             UA_String_deleteMembers(&ed->serverCertificate);
             UA_String_copy(newCertificate, &ed->serverCertificate);
             UA_SecurityPolicy *sp = UA_SecurityPolicy_getSecurityPolicyByUri(server, &server->config.endpoints[i].securityPolicyUri);
+            if(!sp)
+                return UA_STATUSCODE_BADINTERNALERROR;
             sp->updateCertificateAndPrivateKey(sp, *newCertificate, *newPrivateKey);
         }
         i++;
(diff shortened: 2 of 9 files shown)


src/server/ua_server_binary.c
```diff
@@ -647,15 +647,18 @@ createSecureChannel(void *application, UA_Connection *connection,
...
     for(size_t i = 0; i < server->config.endpointsSize; ++i) {
-        UA_Endpoint *endpointCandidate = &server->config.endpoints[i];
+        UA_EndpointDescription *endpointCandidate = &server->config.endpoints[i];
         if(!UA_ByteString_equal(&asymHeader->securityPolicyUri,
-                                &endpointCandidate->securityPolicy.policyUri))
+                                &endpointCandidate->securityPolicyUri))
             continue;
-        retval = endpointCandidate->securityPolicy.asymmetricModule.
-            compareCertificateThumbprint(&endpointCandidate->securityPolicy,
+        securityPolicy = UA_SecurityPolicy_getSecurityPolicyByUri(server,
```
Infinity95: Maybe it would be better to make the certificate comparison a part of the get function?
Because a specific policy instance is uniquely identified by the endpoint and the certificate.

Also, maybe it would be best to make a policyManager plugin(?) so that we can add or remove policys on the fly?

jpfr:
> Also, maybe it would be best to make a policyManager plugin(?) so that we can add or remove policys on the fly?

Why not make the policies a linked list in the configuration?
Then we can add them on the fly in the server/client config.
Without breaking pointers from existing channels.

src/server/ua_server_binary.c
jonasgreen88: > Maybe it would be better to make the certificate comparison a part of the get function?
> Because a specific policy instance is uniquely identified by the endpoint and the certificate.

I tried to find any statement in the specification that confirms this but without success. Do you mean that a server can have several instances of a SecurityPolicy with different certificates? The only thing I can find in the specification is in Part 4, Table 11 - OpenSecureChannel Service Parameters: "The SecurityPolicy is identified by the securityPolicyUri". The idea with this PR is to be able to use the SecurityPolicies also in the access control plugin to decrypt passwords. In the ActivateSessionRequest no ReceiverCertificateThumbprint is included.

> Also, maybe it would be best to make a policyManager plugin(?) so that we can add or remove policys on the fly?

Sure, but what is the use case to add and remove the SecurityPolicies on the fly? Can you share some more details of how this manager should look like?

jonasgreen88:
Ping @Infinity95, @jpfr: How to continue on this PR? Is a plugin manager needed?

src/server/ua_server_binary.c
jpfr: There are cases where we can have several "competing" policies with the same URI.

- If the server is visible in different networks (towards the internet and internally) with different certificates.
- When the GDS updates the server certificate, there could be some time with parallel policy instances.

Some if this stuff is far-fetched. Lets keep this PR simple.
And add this to the lookup when the need arises.

jpfr: Please test if the returned SecurityPolicy is null.
A configuration error should not crash the server.

src/server/ua_server.c
```diff
@@ -321,11 +321,11 @@ UA_Server_updateCertificate(UA_Server *server,
 
     size_t i = 0;
     while (i < server->config.endpointsSize) {
-        UA_EndpointDescription *ed = &server->config.endpoints[i].endpointDescription;
+        UA_EndpointDescription *ed = &server->config.endpoints[i];
         if (UA_ByteString_equal(&ed->serverCertificate, oldCertificate)) {
             UA_String_deleteMembers(&ed->serverCertificate);
             UA_String_copy(newCertificate, &ed->serverCertificate);
-            UA_SecurityPolicy *sp = &server->config.endpoints[i].securityPolicy;
+            UA_SecurityPolicy *sp = UA_SecurityPolicy_getSecurityPolicyByUri(server, &server->config.endpoints[i].securityPolicyUri);
```
jpfr: Please test if the returned SecurityPolicy is null.
A configuration error should not crash the server.

include/ua_plugin_securitypolicy.h
```diff
@@ -386,6 +382,14 @@ UA_SecurityPolicy_getRemoteAsymEncryptionBufferLengthOverhead(const UA_SecurityP
                                                               const void *channelContext,
                                                               size_t maxEncryptionLength);
 
+/* Gets the a pointer to the context of a security policy supported by the
+ * server matched by the security policy uri.
+ *
+ * @param server the server context.
+ * @param securityPolicyUri the security policy to get the context of. */
+UA_SecurityPolicy *
+UA_SecurityPolicy_getSecurityPolicyByUri(const UA_Server *server,
```
jpfr: Can we move this to server_internal.h or similar?
We don't need this in the public API. So lets keep the surface small.

include/ua_plugin_securitypolicy.h
```diff
@@ -111,11 +111,12 @@ typedef struct {
...
      *
      * @param securityPolicy the securityPolicy the function is invoked on.
-     * @param channelContext the channelContext which contains information about
-     *                       the keys needed to decrypt the message.
+     * @param securityPolicyContext the securityPolicyContext which contains
+     *                              information about the keys needed to decrypt
+     *                              the message.
      * @param data the data to decrypt. The decryption is done in place. */
     UA_StatusCode (*decrypt)(const UA_SecurityPolicy *securityPolicy,
-                             void *channelContext,
```
jpfr: This is a (good) simplification and a big bug-magnet (bad).
It is impossible to know for "the new guy" when to use the channelcontext and when to use the policycontext.

Currently, symmetric and asymmetric encryption are both a "cryptoModule" and share this API.
The symmetric encryption needs to channel context in this place.
We should differentiate their APIs to make the difference between channelcontext and policycontext visible.

**Can we move this change to a second PR?**

include/ua_plugin_securitypolicy.h
jonasgreen88: This interface (or something similar) is needed by the accesscontrol plugin when adding support for encrypted passwords. I kept it now, but can remove it if you think some other solution would be better.

include/ua_plugin_securitypolicy.h
jonasgreen88: Sure. I can add an issue for this change.

jonasgreen88:
So, Travis CI passed on the branch of my forked repository. I don't know why it isn't completed on this PR. What's left to do on this PR:

- Discuss if the UA_SecurityPolicy_getSecurityPolicyByUri() function should be removed? Can the SecurityPolicies be provided by the activateSession() callback function instead? Or is it enough that the policies can be reached via the server structure pointer already provided to the activateSession() function? Does some other implementation (e.g. Pub/Sub) need access to the policies as well?
- Merge latest master (can I do it now or should the implementation be approved first?)
- Squash all commits together (or is it done automatically when merged to master?)

@jpfr Do you have something to add?

jpfr:
> Discuss if the UA_SecurityPolicy_getSecurityPolicyByUri() function should be removed? Can the SecurityPolicies be provided by the activateSession() callback function instead? Or is it enough that the policies can be reached via the server structure pointer already provided to the activateSession() function?

getSecurityPolicyByUri will potentially be used in several places. Not only for processing of OPN, but also decryption of authentication UserToken. I would keep it. But as an internal method that is not necessarily API-stable.

> Does some other implementation (e.g. Pub/Sub) need access to the policies as well?

Yes. Decryption of UserToken, PubSub, potentially http transport, etc.
If the policies-array is simply part of the server (client) configuration, then we should cover all potential cases.

> Merge latest master (can I do it now or should the implementation be approved first?)
    Squash all commits together (or is it done automatically when merged to master?)

👍 For merging.
If you don't have any specific commit history you want to keep, then we can squash the PR during merging.

Infinity95:
@jonasgreen88 An application (the server in this case) is identified by an ApplicationInstanceCertificate. The standard does not state, that an application may only have a single one, and it seems plausible to me, that an application may have multiple certificates, issued by different CAs.
A SecurityPolicy is identified by the SecurityPolicyURI, yes. But in our case we have an _Instance_ of a SecurityPolicy, which is identified by the URI _and_ the certificate hash.

Also, i think it might be a good idea to store the Policies in a linked list like @jpfr recommended.
That way we can dynamically add and delete the policies during runtime.
Maybe do this in a separate PR?

jpfr:
Squashed and merged.
Thanks!

PR #2263, opened 2018-12-02
Make logging a proper plugin with state and cleanup method

jpfr:


arch/ua_network_tcp.c
```diff
@@ -613,18 +614,18 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
     connection->sockfd = (UA_Int32) clientsockfd; /* cast for win32 */
 
     if(clientsockfd == UA_INVALID_SOCKET) {
-            UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Could not create client socket: %s", strerror(UA_ERRNO));
-            ClientNetworkLayerTCP_close(connection);
-            return UA_STATUSCODE_BADDISCONNECT;
+        UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
```
Pro: this should use `logger`

arch/ua_network_tcp.c
```diff
@@ -613,18 +614,18 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
...
     }
 
     /* Non blocking connect to be able to timeout */
-    if (UA_socket_set_nonblocking(clientsockfd) != UA_STATUSCODE_GOOD) {
-            UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
-                            "Could not set the client socket to nonblocking");
-            ClientNetworkLayerTCP_close(connection);
-            return UA_STATUSCODE_BADDISCONNECT;
+    if(UA_socket_set_nonblocking(clientsockfd) != UA_STATUSCODE_GOOD) {
+        UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
```
Pro: this should use `logger`

arch/ua_network_tcp.c
```diff
@@ -634,7 +635,7 @@ UA_StatusCode UA_ClientConnectionTCP_poll(UA_Client *client, void *data) {
     if ((error == -1) && (UA_ERRNO != UA_ERR_CONNECTION_PROGRESS)) {
             ClientNetworkLayerTCP_close(connection);
             UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
```
Pro: Same here and in the rest of the file

arch/ua_network_tcp.c
jpfr: Absolutely.
I would go so far not to include  #include "ua_log_stdout.h" here at all.

Infinity95:
Couldn't we make the logger a singleton so that we can access it wherever we want to?
That way we wouldn't have to have all those pesky logger variables in all kinds of places.
IMHO that would clean up the code a bit.

Might make the logging calls easier as well. Example:
```
static UA_INLINE UA_FORMAT(3,4) void
UA_LOG_WARNING(UA_LogCategory category, const char *msg, ...) {
#if UA_LOGLEVEL <= 400
    UA_Logger logger* = UA_GetLogger();
    va_list args; va_start(args, msg);
    logger->log(UA_LOGLEVEL_WARNING, category, msg, args);
    va_end(args);
#endif
}
```
```
UA_LOG_WARNING(category, "my log message");
```

jpfr:
> Couldn't we make the logger a singleton so that we can access it wherever we want to?

Singletons in C are global variables.
http://wiki.c2.com/?GlobalVariablesAreBad

Especially, global variables that are not a constant are bad...
You might have several instances of client / server in one process.
At which point can you shut down the (stateful) logger?

jpfr:
@Pro After trying several times, I am unable to find the source of the problem on OSX.
It is related to the posix architecture definitions.

Do you have a hunch of how to fix this?

Pro:
@jpfr this must be related to the removal of `ua_log_stdout.h`
I reverted that change in one commit (https://github.com/open62541/open62541/pull/2263/commits/ad4f3e6ab1efe33950c6c7561a747bd1ace8955c) and now OS X is green, since the issue is not related to your changes. The macro redefinition is already happening on master, check e.g. https://travis-ci.org/open62541/open62541/jobs/462374570
Looks like it is not failing even though there are warnings and errors in the build.

So let's keep my commit in this branch and I will add a new PR which tries to solve the issue with the queue.h.

PR #2269, opened 2018-12-04
Move plugins to directories for better organisation

Infinity95:


plugins/networking/ua_network_pubsub_ethernet.c
```diff
@@ -5,15 +5,15 @@
  *   Copyright 2018 (c) Kontron Europe GmbH (Author: Rudolf Hoyler)
  */
 
-#include "ua_plugin_network.h"
-#include "ua_log_stdout.h"
-#include "ua_util.h"
+#include "../../include/ua_plugin_network.h"
```
Pro: Relative paths should be avoided for includes

plugins/networking/ua_network_pubsub_ethernet.h
```diff
@@ -7,7 +7,7 @@
 #ifndef UA_NETWORK_PUBSUB_ETHERNET_H_
 #define UA_NETWORK_PUBSUB_ETHERNET_H_
 
-#include "ua_plugin_pubsub.h"
+#include "../../include/ua_plugin_pubsub.h"
```
Pro: Remove relaitve path part

tests/pubsub/check_pubsub_publishspeed.c
```diff
@@ -14,7 +14,7 @@
 #include "ua_types.h"
 #include "ua_pubsub.h"
 #include "ua_config_default.h"
-#include "ua_network_pubsub_udp.h"
+#include "networking/ua_network_pubsub_udp.h"
```
Pro: This will also be problematic for CMake instlall if Amalgamation is disabled

plugins/networking/ua_network_pubsub_ethernet.c
Infinity95: woops. trusted my ide a bit too much doing the refactor. will change.

Pro:
@Infinity95 there are now some conflicts. Can you please rebase?

PR #2271, opened 2018-12-04
WIP: Networking code overhaul

Infinity95:
This pull request aims to overhaul large parts of the networking code to reduce coupling and unify client and server networking.
I will be updating this pull request with a more detailed description as work progresses.

~I made a small python prototype to test the structure which can be found in the concepts directory.~
Deleted, because not needed anymore.

Steps until completion:
- [x] Python structure demo
- [x] Scaffolding in the form of structures and type definitions (probably done)
- [x] NetworkManager implementation
- [x] TCP ListenerSocket implementation
- [x] TCP DataSocket implementation (client, server differs only in the initialization function)
- [x] Swap old networking code for new networking code. _This will probably take the most effort and require some modifications of the api that i devised because of unforeseen circumstances_
- [x] Fix existing unit tests **server side tests that do not involve the client are up and running**
- [ ] Write unit tests for the new networking code
- [x] ~After the server side is up and running again start working on the client ~(separate PR)~ A separate PR will most probably not be possible. There are too many dependencies to make a clean cut imho.~
- [x] Quick and dirty api change on the client side. Needs cleaning up in separate PR


Infinity95:
Continued in #2408

PR #2288, opened 2018-12-11
Fix: mdns record when strlen(hostname) + strlen(servername) is 61 or 62

ccvca:
This fixes the case, that servername + hostname have a length of 61 or 62.

outServiceDomain is a buffer of length 63 + 24
so maxLen is 63
When Servername + hostname in total are 61 or 62 UA_snprintf will cut off the last (two) letter(s).

f.e.
servername:
123456789_123456789_123456789_123456789_123456789_123456789
hostname:
abc
will result in 
123456789_123456789_123456789_123456789_123456789_123456789-a\0
and the next UA_snprintf will insert after the \0


jpfr:
Thx!

PR #2298, opened 2018-12-14
history_server: Add a tutorial for server side history data

peter-rustler-basyskom:


examples/tutorial_server_historicaldata.c
```diff
@@ -0,0 +1,121 @@
...
+
+    /*
+     * We want the values stored in the database, when the nodes value is set.
+     */
+    setting.historizingUpdateStrategy = UA_HISTORIZINGUPDATESTRATEGY_VALUESET;
+    /*
+     * At the end we register the node for gathering data in the database.
+     */
+    retval = gathering.registerNodeId(server, gathering.context, &outNodeId, setting);
+    fprintf(stderr, "registerNodeId %s\n", UA_StatusCode_name(retval));
```
jpfr: Can we use UA_LOG_INFO(UA_Log_Stdout, ...) here as well?

examples/tutorial_server_historicaldata.c
```diff
@@ -0,0 +1,121 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+#include "open62541.h"
```
basyskom-meerkoetter: due to recent changes to how the examples are handled you can't use the combined header here.
needs to become something like this

-#include "open62541.h"
+#include <ua_server.h>
+#include <ua_config_default.h>
+#include <ua_log_stdout.h>
+#include <ua_plugin_historydatabase.h>
+#include <historydata/ua_plugin_history_data_gathering.h>
+#include <historydata/ua_historydatagathering_default.h>
+#include <historydata/ua_historydatabase_default.h>
+#include <historydata/ua_historydatabackend_memory.h>

peter-rustler-basyskom: Done

Pro: Not completely correct ;)

```
#include <historydata/ua_plugin_history_data_gathering.h>
#include <historydata/ua_historydatagathering_default.h>
#include <historydata/ua_historydatabase_default.h>
#include <historydata/ua_historydatabackend_memory.h>
```

Should not use the historydata folder, since it will not work with the installed target. make install just copies all the header files into one folder and then this code will fail.

You need to change it to 

```
#include <ua_plugin_history_data_gathering.h>
#include <ua_historydatagathering_default.h>
#include <ua_historydatabase_default.h>
#include <ua_historydatabackend_memory.h>
```

peter-rustler-basyskom:
This should be ready for integration, now.

examples/tutorial_server_historicaldata.c
peter-rustler-basyskom: Is this really true?

It will not build if I change that.

Maybe, this is true for examples that will be build out of library source, but not if build inside library source.

peter-rustler-basyskom: I tried if the ci is successful building it without folder names in the includes.
It went wrong.
I reverted that commit.
So, path need to be there.

peter-rustler-basyskom:
I tried to build it without folder names in the includes and it does not work.
I have reverted that.

I think this should be ready for integration, when Ci is green.

Pro:
I think the subfolders of the `plugins` folder should also be added as include_directory in CMake, i.e. `plugins/historydata`

Pro:
Otherwise there may be issues after the make install

peter-rustler-basyskom:
Ok, I try that.

peter-rustler-basyskom:
I added it in the Makefile,txt from the examples.
Please review and if all is good and CI is green, merge.
Thanks.

Pro:
Cool!! Thanks, can be merged when green.

PR #2302, opened 2018-12-17
Use separate header files instead of open62541.h amalgamation

Pro:
Based on #2301 

For packaging and installation we are recommending the non-amalgamated version. See also discussion in #2296


Pro:
This PR is now ready (finally).

It took some to to get CI all green.

Now the examples are all using separate headers and `make install` is installing the correct files.

@jpfr I think we already discussed that in the telco, but can you quickly comment, if this is fine, to use the separate headers in the examples?

The .debian package #2397 will then also use the separate headers.
Amalgamation is not the recommended way to install, but still supported.

jpfr:
Yes! Good to merge.
I will comment separately in the debian package PR.

PR #2305, opened 2018-12-18
Fix value rank compatibility check

basyskom-jvoe:
The number of array dimensions must be set to 1 only if the value is not scalar and if the array dimensions size is 0. Omitting the second check breaks the compatibility check for arrays with more than one dimension.


jpfr:
Thanks! Good to merge when green.
Master has this already...

PR #2310, opened 2018-12-19
Attempt to fix crash in read response

lkaino:
In some cases when server was rebooted during a read, the callback would access NULL results.

This fixes the crash, but doesn't call the registered application callback in case results is empty.


ccvca:
If I get this right, threre is a memory leak in that case, because the callback is not removed from the list.
That's also the case that the server respond without a value:
https://github.com/open62541/open62541/blob/b0aafd8590cc9d420eef3193b1a521fab726e160/src/client/ua_client_highlevel.c#L492-L494

Or is it expected that the Server will answer with a correct response after a fault one is received?

lkaino:
Changed now so that the callback is removed in both failure cases.

This is assuming that we are not expecting another response from the server.

jpfr:
See #2321

PR #2326, opened 2018-12-25
Use zip tree data structure for ua_timer

jpfr:
Despite all kinds of tricks, the linked list approach was slowing us down for many repeated callbacks,


StalderT:
@jpfr  Great. Can we backport this change to 0.3? Maybe without changing the API (`UA_Timer_removeRepeatedCallback` and `UA_Timer_addRepeatedCallback` will always return `UA_STATUSCODE_GOOD`).

Or else create a branch 0.3.1 with new features like client encryption? What do you think?

Pro:
I would vote against backporting improvements like this, and only backport really necessary bug fixes. Porting features to multiple branches is quite some effort at current stage since the internal Api changes quite often due to new features. I would say we focus on the current master branch and try to release 0.4 faster.

Issue #2331, opened 2018-12-28
Unit test subscription_events randomly fails


StalderT:
Description
===========

Unit test subscription_events randomly fails

see : 

https://travis-ci.org/open62541/open62541/jobs/473063187#L5215
https://travis-ci.org/open62541/open62541/jobs/473039607#L8243

Checklist
=========
Please provide the following information:

 - [X] open62541 Version (release number or git tag): master

Pro:
@aDogCalledSpot can you have another look?

Pro:
See also:

https://travis-ci.org/open62541/open62541/jobs/474423874#L5222

Issue #2330, opened 2018-12-28
Unit test historical_data test randomly crash (SEGFAULT)


StalderT:
Description
===========

Unit test historical_data test randomly crash (SEGFAULT) see : https://travis-ci.org/open62541/open62541/jobs/473063184#L9687

Checklist
=========
Please provide the following information:

 - [X] open62541 Version (release number or git tag): master

Pro:
See also https://travis-ci.org/open62541/open62541/jobs/474423872#L8280

Pro:
![image](https://user-images.githubusercontent.com/251973/50638819-673ef600-0f5f-11e9-9d84-78f00d19a41b.png)

Found the issue. There's a race condition:
`startPoll` is creating a monitored item, and `stopPoll` is deleting it. Meanwile `serverloop` is executing `UA_Server_run_iterate` which is accessing the deleted timer object for the monitored item.

I'll create a fix for that.

StalderT:
Thanks

PR #2333, opened 2019-01-02
Process opc bit

basyskom-swid:
This change enables the generate_datatypes.py script to process opc:Bit fields in Structures. The fields can then be used as switches for optional Fields (Field attribute SwitchField), which will only be encoded/decoded if the corresponding bit is set to the correct attribute SwitchValue. If no SwitchValue is specified it defaults to 1.
SwitchFields other than opc:Bit flags (e.g. UInt32 SwitchFields) are not yet supported.


basyskom-swid:
I can not reproduce the test errors in the CI. Does anybody have a hint which combination of variables (platform/compiler/cmake config/...) leads to the issue?

StalderT:
> I can not reproduce the test errors in the CI. Does anybody have a hint which combination of variables (platform/compiler/cmake config/...) leads to the issue?

https://github.com/open62541/open62541/issues/2331
https://github.com/open62541/open62541/issues/2330

basyskom-swid:
Thanks.

jpfr:
Hey there,

looks good! Now lets see if we can simplify the structure representations a bit.
If we want to support the 1.04 spec fully, then we need to take care not to accumulate complexity here.

- I don't see why we need to track the position of the switchfield in every member.
We can just count the position in the switchfield when we looping over the members.
- The fact that a switchfield is included in the structure can be covered by adding `padding` to the first member. So we don't have to represent switchfields as members at all.
- This is not mentioned in the standard. But for consistency, arrays could also be optional members.
- For a "union structure" all fields are treated the same.

With the above, we need only one additional bit for every member.
The overall struct length stays the same as bitfields sizes are at least one byte long.

```c
typedef struct {
#ifdef UA_ENABLE_TYPENAMES
    const char *memberName;
#endif
    UA_UInt16 memberTypeIndex;    /* Index of the member in the array of data types */
    UA_Byte   padding;            /* How much padding is there before this
                                     member element? For arrays this is the
                                     padding before the size_t length member.
                                     (No padding between size_t and the
                                     following ptr.) */
    UA_Boolean namespaceZero : 1; /* The type of the member is defined in
                                     namespace zero. In this implementation,
                                     types from custom namespace may contain
                                     members from the same namespace or
                                     namespace zero only.*/
    UA_Boolean isArray       : 1; /* The member is an array */
    UA_Boolean isOptional        : 1; /* The member is a flag and will be encoded in the encoding mask*/
} UA_DataTypeMember;
```

In the future, the `UA_DataType` should be explicit about its nature.
So we can have clear "switch" cases and don't require funny interactions between various flags.

```c
enum UA_DataTypeKind {
    UA_DATATYPETYPE_BUILTIN = 0, /* Including decimal (still unsupported) */
    UA_DATATYPETYPE_ENUMERATION = 1,
    UA_DATATYPETYPE_STRUCTURE = 2,
    UA_DATATYPETYPE_OPTSTRUCTURE = 3,
    UA_DATATYPETYPE_UNION = 4,
    /* The following are bitfields that indicate simplified handling */
    UA_DATATYPETYPE_POINTERFREE = 32, /* The type (and its members) contains no
                                    pointers that need to be freed */
    UA_DATATYPETYPE_OVERLAYABLE = 64 /* The type has the identical memory layout in
                                    memory and on the binary stream */
}
```

This enum can be inserted as one-byte bitfield.
At the right position so we don't provoke unnecessary padding between struct members.

```c
struct UA_DataType {
#ifdef UA_ENABLE_TYPENAMES
    const char *typeName;
#endif
    UA_NodeId  typeId;           /* The nodeid of the type */
    UA_UInt16  memSize;          /* Size of the struct in memory */
    UA_UInt16  typeIndex;        /* Index of the type in the datatypetable */
    UA_DataTypeKind kind : 8;   /* Which kind of datatype is this? One byte enum/bitfield */
    UA_Byte    membersSize;      /* How many members does the type have? */
    UA_UInt16  binaryEncodingId; /* NodeId of datatype when encoded as binary */
    //UA_UInt16  xmlEncodingId;  /* NodeId of datatype when encoded as XML */
    UA_DataTypeMember *members;
};
```

What do you think @basyskom-swid ?
Would this structure contain all the information you need?

basyskom-swid:
@jpfr I based the implementation off of the 1.0.3 spec, which (as I see now) is significantly different from the 1.0.4 spec w.r.t. optional fields, i.e. in 1.0.4 the encoding mask is always exactly 32 bits with a maximum of 32 optional fields whereas in 1.0.3 it was a multiple of 8 bits to accommodate all defined bits.
I will read up on the new spec and your comments and reply then, thanks for the feedback!

basyskom-swid:
@jpfr I have another question about the encoding mask. Part 6 V 1.0.3 reads:

jpfr:
Oh wow.

I was not aware of this incompatibility between 1.03 and 1.04.
This is not a good sign for a standard as important as OPC UA...

I would aim for 1.04. The schema has become much simpler.
And the commercial vendors support 1.04 today...

Sorry if this causes inconvenience for you.

basyskom-swid:
It is suprising, we were under the impression that backward compatibility should never be compromised. But as far as I understand it, there's no way to satisfy both conditions without querying supported versions.
I'd also say we aim for 1.0.4, I'll update the pull request as soon as I have it all in place.

Pro:
One short hint: CI is randomly failing. That's already fixed on master. So rebasing this PR onto master will fix this.

basyskom-swid:
@jpfr
[...]
> If we want to support the 1.04 spec fully, then we need to take care not to accumulate complexity here.

Agreed

>     * I don't see why we need to track the position of the switchfield in every member.
>       We can just count the position in the switchfield when looping over the members.
I can't quite wrap my head around the wording in (1.0.4, Part 5, Annex E 2.6), which I read in a way that permits multiple fields to reference the same SwitchField, which in turn would necessitate tracking flag positions for members and (1.0.4, Part 6, 5.2.7) which says `...Each optional field is assigned exactly one bit...`

>     * The fact that a switchfield is included in the structure can be covered by adding `padding` to the first member. So we don't have to represent switchfields as members at all.

This is a question I had for the already implemented LengthField as well: Since LengthField and SwitchField can reference arbitrary integer fields in the structure, can we really just exclude those from the serialization?

>     * This is not mentioned in the standard. But for consistency, arrays could also be optional members.

>     * For a "union structure" all fields are treated the same.

This I don't understand, could you elaborate?

[...]

Pro:
Note that there was an amendment for the 1.04 spec regarding the bits:
https://opcfoundation-onlineapplications.org/mantis/view.php?id=3426

image


Not sure if that's relevant here.

basyskom-swid:
That is an errata doc for 1.0.3, which has, as far as I can tell, been integrated into 1.0.4 anyway.

jpfr:
>  I think I have now understood how the binary mapping is supposed to work

👍 

> Since LengthField and SwitchField can reference arbitrary integer fields in the structure, can we really just exclude those from the serialization?

They are not variable-sized in 1.04. So a fixed padding should work.

> For a "union structure" all fields are treated the same.
This I don't understand, could you elaborate?

In unions, all fields are automatically optional.
So you wouldn't need the isOptional flag for members.

basyskom-swid:
My current issue:

With the 1.0.4 spec the technical definition of the encoding mask itself has been simplified.
It is a single unsigned 32 bit integer with exactly one bit reserved for every optional field up to a maximum of 32 optional fields.
The encoding mask is added as soon as any member of a structure has a `SwitchField` parameter.

Unfortunately it seems non-trivial to dynamically compute the encoding mask for a given instance of a structure in a general way without packing a lot of information into the member descriptors (offset in the type, bitshift, bitlength for bit fields and switchField + operand to use for the members to switch etc.)

My main problem here are the `opc:Bit` fields, which I tried mapping to packed C bytes, which I am not sure is the correct way to do it. Unfortunately the `opc:Bit` fields are not always single bits, but can have a `Length` parameter to expand to multiple bits and have values greater than `1`.

This fact, in conjunction with the `SwitchValue` and `SwitchOperand` parameters also mentioned in the spec lead me to believe, that it would be easier to actually generate a concrete UA_<type>_encodingMask(<type> *instance) function for structure types with optional fields.

The question now is, how these functions could be passed down into the UA_calcSize/encode/decode.
Or maybe add calcSize/encode/decode variants with encodingMask parameters?

jpfr:
> Unfortunately it seems non-trivial to dynamically compute the encoding mask for a given instance of a structure in a general way without packing a lot of information into the member descriptors (offset in the type, bitshift, bitlength for bit fields and switchField + operand to use for the members to switch etc.)

I don't see much of a problem.
The XML-seems more complicated than the binary encoding.
To me, the binary encoding is the "ground truth".

A single boolean needed for the members: Indicating whether they are optional.
There is no field-length in v1.04. Just the position in the the int32 bitfield.
The position in the bitfield is available by just keeping track during the iterations over the members.

basyskom-swid:
@jpfr: What you say is correct for the encoding mask itself, which is present for decoding, but for encoding and length calculations we need to create a proper encoding mask for an instance of a struct. This decision is sometimes based on `opc:Bit` fields, which can actually comprise of more than one bit (e.g. spec 1.0.4, part 5, annex 4, 2). As far as I understand the `opc:Bit` fields are proper fields in a struct and need to be encoded/decoded as well., otherwise the example above does not make sense.

The input for the decision whether a  member is optional are the SwitchField, SwitchValue and SwitchOperand parameters.

Let's assume this structure:
```
    <opc:StructuredType BaseType="ua:ExtensionObject" Name="ObjectWithBitsOptionalFields">
        <opc:Field TypeName="opc:Bit" Name="OptionOne"/>
        <opc:Field TypeName="opc:Bit" Name="OptionTwo" Length="2"/>
        <opc:Field TypeName="opc:Bit" Name="Reserved1" Length="5"/>
        <opc:Field TypeName="opc:Byte" Name="OptionalByteTwo" SwitchField="OptionOne"/>
        <opc:Field TypeName="opc:Byte" Name="OptionalByteOne" SwitchField="OptionTwo" SwitchValue="2" SwitchOperand="LessThanOrEqual"/>
    </opc:StructuredType>
```

The C-struct might look like this:
```
typedef struct {
    UA_Byte optionOne : 1;
    UA_Byte optionTwo : 2;
    UA_Byte optionalByteOne;
    UA_Byte optionalByteTwo;
} UA_ObjectWithBitsOptionalFields;
```
For the values
```
{
  0, // OptionOne
  2, // OptionTwo
  11, // optionalByteOne
  23 // optionalByteTwo
}
```
this would result in the following encoding
```
0x00 0x00 0x00 0x02 // 4 byte encoding mask, OptionTwo is true
0x04 // packed bitfield containing OptionOne and OptionTwo, with OptionTwo being 2
0x17 // optionalByteTwo, since OptionTwo <= 2
```

The problem I see is generically checking the conditions for the members to create the encoding mask. My current idea would be to generate e.g. a `uint32_t UA_ObjectWithBitsOptionalFields_encodingMask(*src)` into the `*_encoding_binary.h`:
```
uint32_t UA_ObjectWithBitsOptionalFields_encodingMask(*src) {
  uint32_t encodingMask = 0;
  if(src->OptionOne != 0) {
   encodingMask |= 0x01;
  }
  if(src->OptionTwo <= 2) {
    encodingMask |= 0x02;
  }
  return encodingMask;
}
```
Since this would be required for a small subset of types, I would add `UA_calcSizeBinaryMasked(*src, type, mask)` and `UA_encodeBinaryMasked(src*, type, mask, ...)`

I'm not seeing a simple solution here, maybe I'm betriebsblind?

basyskom-swid:
@jpfr: I understand the spec such that the encoding mask is an _additional_ 32 bit at the start of the encoded binary message. It is not directly defined by any bit fields in the defined StructuredType, those bitfields are only used to determine whether a field is enabled or not during encoding.

Bitfields need to be treated like (basically) small integer values with bytewise packing.

The `opc:Bit` fields can explicitly have more than one bit (as mentioned above, see spec 1.0.4, part 5, annex 4, number 2). I think something like this would be a correctly defined struct:

```
    <opc:StructuredType BaseType="ua:ExtensionObject" Name="OptionObject">
        <opc:Field TypeName="opc:Bit" Name="MasterSwitch"/>
        <opc:Field Length="7" TypeName="opc:Bit" Name="Reserved"/>
        <opc:Field TypeName="opc:Byte" Name="O1" SwitchField="MasterSwitch"/>
        <opc:Field TypeName="opc:Byte" Name="O2" SwitchField="MasterSwitch"/>
        <opc:Field TypeName="opc:Byte" Name="O3" SwitchField="MasterSwitch"/>
        <opc:Field TypeName="opc:Byte" Name="O4" SwitchField="MasterSwitch"/>
        <opc:Field TypeName="opc:Byte" Name="O5" SwitchField="MasterSwitch"/>
       ...
        <opc:Field TypeName="opc:Byte" Name="O32" SwitchField="MasterSwitch"/>
    </opc:StructuredType>
```

If `MasterSwitch` is 1, this would encode as follows:
```
0xff 0xff 0xff 0xff // full encoding mask, all 32 optinal fields are active
0x01 // MasterSwitch bitField with the padding
0x<value of O1>
0x<value of O2>
...
0x<value of O32>
```

If `MasterSwitch` is 0:
```
0x00 0x00 0x00 0x00 // Empty encoding mask
0x00 // MasterSwitch bitField with the padding
// Nothing else, everything optional has been disabled
```

Another problem I have thought about is nested structs with optional fields, which would _require_ dynamic computation of the encoding mask in recursive calls of e.g. UA_calcBinarySize.

jpfr:
> I understand the spec such that the encoding mask is an additional 32 bit at the start of the encoded binary message. It is not directly defined by any bit fields in the defined StructuredType, those bitfields are only used to determine whether a field is enabled or not during encoding.

There are no bitfields inside struct.
There is only a bitfield at the beginning of the struct.

> Another problem I have thought about is nested structs with optional fields, which would require dynamic computation of the encoding mask in recursive calls of e.g. UA_calcBinarySize.

UA_calcBinarySize recursively walks over the members. I don't see a problem here.
Of course this will continue to be recursive!

We should make a high-level decision now.
So that you can be as efficient as possible with your work.
I gave two alternatives for the struct with optional fields above.
Can you confirm that you will pursue the first alternative?

```c
typedef struct {
    UA_Int32 hasOptionalByteOne : 1;
    UA_Int32 hasOptionalByteTwo : 1;
    UA_Byte optionalByteOne;
    UA_Byte optionalByteTwo;
} UA_ObjectWithBitsOptionalFields;
```

If not, can you give a counter-example that cannot be represented in the proposed scheme?

basyskom-swid:
> > I understand the spec such that the encoding mask is an additional 32 bit at the start of the encoded binary message. It is not directly defined by any bit fields in the defined StructuredType, those bitfields are only used to determine whether a field is enabled or not during encoding.
> 
> There are no bitfields inside the struct.
> There is only a bitfield at the beginning of the struct.

How would that work with the example in Part 5 Annex E.4 Number 2
```
2. A 16-bit value divided into several fields.
<opc:StructuredType Name="Quality">
<opc:Documentation>An OPC COM-DA quality value.</opc:Documentation>
<opc:Field Name="LimitBits" TypeName="opc:Bit" Length="2" />
<opc:Field Name="QualityBits" TypeName="opc:Bit" Length="6"/>
<opc:Field Name="VendorBits" TypeName="opc:Byte" />
</opc:StructuredType>
```

If bit fields are not proper members of a structure that type does not make sense as the bit fields are never used and carry real information.

[...]

jpfr:
Aaah. Understood.

These are bitfields that are not switchfields.
Based on the original PR description, I had not considered them at all.

Yes, bitfields that are no switchfields should be listed as members.

For these I have a question:
In C, `int a : 1;` is at least 4 byte long and `char b : 1;` is at least one byte long.
Also, only bitfields of the same underlying type are "packed".

How is this in OPC UA? What happens if the field of reserved padding bits is missing?

It would be great if we could just memcpy the underlying type.
So that we can have one "int" member in the array instead of 5 bitfield members.

jpfr:
> I really don't see that they should be treated differently, do you have a pointer to the spec? Might have missed that.

If we have a `UA_DataTypeKind` for structure with optional fields, then we know there is a 32bit field at the beginning. No need to model it as a member field.

> Which in turn only makes sense to me if the information is serialized. Since it says byte boundaries I have opted to use UA_Byte as the underlying type for bitfields.

> I assume that no opc:Bit field has a length > 7, otherwise the user should use a byte or int type

Part 5 contains the following example. Would it also get UA_Byte as the underlying type?
```xml
<opc:OpaqueType Name="Int128" LengthInBits="128" ByteOrderSignificant="true">
    <opc:Documentation>A 128-bit signed integer.</opc:Documentation>
</opc:OpaqueType>
```

> Bitfields can only appear at the beginning of a struct (not sure this assumption is correct)

Not sure either.

> How do we distinguish union fields and optional fields? Is a field a union as soon as it has a SwitchValue?

We see the "kind" of the data type.

-------

We can also merge this piecewise.
If we need more time to figure out edge cases, what would be a nice self-contained feature-set to merge in a first step?

What is the current state of the PR? Ready for review?

jpfr:
Dear Sumedha,

this is an important changes.
The type definitions touches the very core of this project.
So we should invest the time to get it right.

It is now clear that bitfields/switch fields are used in several places:
- Tagged unions
- Structures with optional fields
- Bona-fide bitfields

We should treat these cases separately.
I just pushed PR #2399. It does not add any functionality, but adds a *tag* to every data type.
This "tag" (named datatype-kind) is used for dispatch in the jump tables.

```c
typedef enum {
    UA_DATATYPEKIND_BOOLEAN = 0,
    UA_DATATYPEKIND_SBYTE = 1,
    UA_DATATYPEKIND_BYTE = 2,
    UA_DATATYPEKIND_INT16 = 3,
    UA_DATATYPEKIND_UINT16 = 4,
    UA_DATATYPEKIND_INT32 = 5,
    UA_DATATYPEKIND_UINT32 = 6,
    UA_DATATYPEKIND_INT64 = 7,
    UA_DATATYPEKIND_UINT64 = 8,
    UA_DATATYPEKIND_FLOAT = 9,
    UA_DATATYPEKIND_DOUBLE = 10,
    UA_DATATYPEKIND_STRING = 11,
    UA_DATATYPEKIND_DATETIME = 12,
    UA_DATATYPEKIND_GUID = 13,
    UA_DATATYPEKIND_BYTESTRING = 14,
    UA_DATATYPEKIND_XMLELEMENT = 15,
    UA_DATATYPEKIND_NODEID = 16,
    UA_DATATYPEKIND_EXPANDEDNODEID = 17,
    UA_DATATYPEKIND_STATUSCODE = 18,
    UA_DATATYPEKIND_QUALIFIEDNAME = 19,
    UA_DATATYPEKIND_LOCALIZEDTEXT = 20,
    UA_DATATYPEKIND_EXTENSIONOBJECT = 21,
    UA_DATATYPEKIND_DATAVALUE = 22,
    UA_DATATYPEKIND_VARIANT = 23,
    UA_DATATYPEKIND_DIAGNOSTICINFO = 24,
    UA_DATATYPEKIND_DECIMAL = 25,
    UA_DATATYPEKIND_ENUM = 26,
    UA_DATATYPEKIND_STRUCTURE = 27,
    UA_DATATYPEKIND_OPTSTRUCT = 28, /* struct with optional fields */
    UA_DATATYPEKIND_UNION = 29,
    UA_DATATYPEKIND_BITFIELDCLUSTER = 30 /* bitfields + padding */
} UA_DataTypeKind;
```

So in your case, you would implement the dispatch methods for `UA_DATATYPEKIND_OPTSTRUCT` and `UA_DATATYPEKIND_BITFIELDCLUSTER` separately. Ideally we can even put them into separate smaller PR.

Does the general structure of PR #2399 work for you?
We can also take this offline with a short phone call.

basyskom-swid:
Just a heads up, I was unavailable last week and will have a look at the tags now.

basyskom-swid:
Simply rebasing onto master does not seem like a good way forward. I will close this PR for now and try to split the changes up into multiple, self-contained pull requests.

Issue #1914, opened 2018-07-11
Nodeset Compiler: Generate type defintion for custom datatypes instead of using binary encoding hack


Pro:
This is a replacement for #1609:
Currently the nodeset compiler is using binary-encoded ExtensionObjects to pass custom datatypes from XML to the server initialization. This should be replaced by apropriately generated the custom datatype code and then setting the variable correspondingly. The nodeset compiler should only use public API.

Comment by @jpfr:
The ideal solution would be to not manually encode values in the generated code at all.
We have all the facilities in place to do that now.

In the generated code:
1. Print the type for a new custom structure
2. Print the type description (with macros referring to the custom structure to get the padding right)
3. Add the custom type description to the server
4. Add the variable node with our standard API

crafti5:
is it possible to use nodeset compiler with customs enumeration? I'm stuck here with

`retVal |= UA_encodeBinary(`
 
And my Visual Studio says "error C3861: 'UA_encodeBinary': identifier not found". Can I do or change something here?

Pro:
Fixed in #2337

Issue #1420, opened 2017-11-30
Provide proper information for ExtensionObject decoding


Pro:
For custom datatypes and NS0 generation we are using ExtensionObjects.

UaExpert can currently not display the content of these objects. Other server implementations are able to do that.
see also: #1414 

I'm not sure if UaExpert also uses the `Opc.Ua` node (i=7617) to decode extension objects. It is more or less the content of the `Opc.Ua.DataTypes.bsd` file. Or it could also suffice to add the corresponding data type under `BaseDataTypes`.

This issue is open for grabs.

Pro:
> The UAExpert is a generic client that constructs some nice GUI elements on the fly.
In 90% of the cases, the client contains some hard-coded logic that acts on the received value.
So they have the structure representation compiled in. See the example in our repository for how to do that client-side.

> There is a way to represent the structure layout as a dictionary-object with special child nodes. Probably the UAExpert is using that to generate GUI elements.
See Chapter 5.8 for Part 3 of the OPC UA Spec (can be downloaded from the Foundation's Website with a free registered account). We could generate such a dictionary-object from the UA_DataType type description. But we currently don't.

https://github.com/open62541/open62541/issues/1414#issuecomment-348235627

WoutvdBerg:
Some extra information about the handling of Extensionobjects in UAExpert.

https://forum.unified-automation.com/post6064.html?hilit=datatype%20person#p6064

Pro:
Same as #1414

kjeldflarup:
What is the status, is UAExpert still a no go to display ExtensionObjects?   
I tried to test with examples/custom_datatype/server_types_custom.c but UAExpert does not seem to understand the 3Dpoint. 

Are there alternatives to UAExpert

kjeldflarup:
I tried with FreeOpcUa client. It also complains about something. 
```
    raise UaError("Non array Variant of type {0} cannot have value None".format(self.VariantType))
opcua.ua.uaerrors._base.UaError: Non array Variant of type VariantType.ExtensionObject cannot have value None
uaclient.mainwindow - WARNING - showing error: %s')
```

matkonnerth:
OPC UA 1.04 Part 3 mentions at 5.8, that the Attribute "DataTypeDefinition" is mandatory for structs and enums. This attribute contains a struct of StructureDefinition (for struct datatypes) or EnumDefinition (for enums).

As I understand, this struct contains the information, how the DataType is encoded.

[...]

PR #2337, opened 2019-01-02
[REVIEW] #1914 - remove binary encoding from namespace generation

matkonnerth:
This pr shold remove the necessity of calling ua_encodebinary during namespace generation with the nodeset compiler.

I have added a testnodeset in examples/nodeset. I will move this to tests folder and write some tests.

### DIFF
--- a/examples/nodeset/CMakeLists.txt
+++ b/examples/nodeset/CMakeLists.txt
@@ -25,6 +25,21 @@ if(UA_COMPILE_AS_CXX)
     set_source_files_properties(${PROJECT_BINARY_DIR}/src_generated/ua_namespace_example.c PROPERTIES LANGUAGE CXX)
 endif()
 
+if(UA_NAMESPACE_ZERO STREQUAL "FULL")
+    ua_generate_nodeset_and_datatypes(
+        NAME "testnodeset"
+        FILE_CSV "${PROJECT_SOURCE_DIR}/examples/nodeset/testnodeset.csv"
+        FILE_BSD "${PROJECT_SOURCE_DIR}/examples/nodeset/testtypes.bsd"
+        NAMESPACE_IDX 2
+        FILE_NS "${PROJECT_SOURCE_DIR}/examples/nodeset/testnodeset.xml"
+        INTERNAL
+    )
+    add_example(server_testnodeset server_testnodeset.c 
+        ${PROJECT_BINARY_DIR}/src_generated/ua_namespace_testnodeset.c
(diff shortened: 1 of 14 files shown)


jpfr:
Wow! Looks good. Just two questions.

Can you explain why we should have EnumValueType in the minimal types?

We can now add custom datatypes to the server-config at runtime.
I did not see this feature used. Did I miss something?
Otherwise, if the datatype is not "registered", how is decoding possible, e.g. for a write-request?

matkonnerth:
Hi!
1) EnumValueType: there ist a node called EnumValues:
take a look on ua_namespace0.c, here before my pr:
struct {
  UA_Int64 Value;
  UA_LocalizedText DisplayName;
  UA_LocalizedText Description;
} variablenode_ns_0_i_12169_EnumValueType_0_0_struct;

after pr: I use the generated UA_EnumValueType for setting up the value

2) that's interesting, thanks for the hint. I've to take a look on it.

Pro:
Note that the corresponding necessary steps to remove encodeBinary are described in #1914

Pro:
Hey @matkonnerth!
Thank you for your nice PR! Great work! I really like the simplification this PR brings.

I just had a look at it and the changes look really good! I also added an additional commit which removes some unnecessary code and does some further cleanup. Can you have a look at it?

Just as a summary:
With this PR we do not need the encodeBinary in the public API and the generated code since it is using the generated datatypes for initializing extension objects.

@matkonnerth is this PR ready to be merged? If yes, please squash your commits.
As far as I can see there are still some minor issues with parsing `<array1><Double>1.1</Double></array1>`. I would put that fix in a new PR since it is a new addition/feature and not a change of the current code.

matkonnerth:
Hi @Pro!

I have a few more unit tests, I will add them, I think then you can merge this pr in.

One question:
Should I let the server_testnodeset in the examples or remove it from there?

Pro:
Unit tests are always good!

I'd put the testnodeset.xml into the tools/test/nodeset_compiler folder since it's mainly there for unit tests. Maybe you can add a note in the example/nodeset/server_nodeset.xml and say something like `A more complete example can be found in tools/test/nodeset_compiler`. At a later stage we may extend the nodeset example further.

matkonnerth:
I think if CI is green, you can merge it in.

Pro:
Thank you @matkonnerth!!

Another issue closed :+1: 

We would be really happy if you could also contribute a fix for the `<array1><Double>1.1</Double></array1>` initialization and especially #1420

Issue #2322, opened 2018-12-21
[Discovery] mDNS-Response causes server lags


ccvca:
Description
===========
UA_Server_run_iterate could take up to 9 Seconds when a mdns-record is received, which could not be resolved to a hostname (e.g. Illegal AutoIp-Adress 169.254.213.26). During this time the server execution is blocked.

**This offers DDOS-possibilities by sending malicous mdns-Packages.**

Background Information / Reproduction Steps
===========================================
I've catched the code path:

Here the mdns-Part is executed inside UA_Server_run_iterate :
https://github.com/open62541/open62541/blob/f523384344987a167b07d6d9cfd4b4ae844c0090/src/server/ua_server.c#L424-L437

When a new record is received the callback 'mdns_record_received' calls setSrv which tries to resolve the adress:
https://github.com/open62541/open62541/blob/f523384344987a167b07d6d9cfd4b4ae844c0090/src/server/ua_server_discovery_mdns.c#L482-L485
This command could take several seconds, when there is no hostname availiable.

I've added some logs before and after:
```
UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "Before getnameinfo");
        int res = getnameinfo((struct sockaddr*)&sockaddr,
                              sizeof(struct sockaddr_storage),
                              remote_name, sizeof(remote_name),
                              NULL, 0, NI_NAMEREQD);
        UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "After getnameinfo");
```
Which results in 
```
[2018-12-21 13:56:21.307 (UTC+0100)] debug/server       Before getnameinfo
[2018-12-21 13:56:30.314 (UTC+0100)] debug/server       After getnameinfo
```

Possible Solutions?
===========================================
- Do not resolve the hostname and leave it up to the user to do this
- Use an asynchronous version of getnameinfo which is not portable (maybe getaddrinfo_a (linux), WSAAsyncGetHostByAddr (Windows) )
- Add an option in the config, to disable/enable hostname lookup for those who only need the IP

I would prefer the 1st or 3rd option.

Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): 20eccdcaf15162f48a372186e15f654d81331f59
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system: Windows 10, Mingw-w64
 - [x] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
[...]

Issue #2212, opened 2018-11-13
MDNS: Handle multiple A-records


ccvca:
Description
===========
I discovered some problems on windows with multiple network adapters and mdns-discovery

- Mdns finds multiple IP-Adresses on my computer (one for each networkinterface, **even if the adapter is disabled.**)
- When receiving the record only the first A-record seems to be used,

I would prefer two fixes:
- Do not add IP-adresses from disconnected network adapters, as the 169.254.*.* adresses could not be connected by anyone. (Not even on the same machine)
- Either call the ServerOnNetworkCallback-function multiple times (per IP once), or change the datastructure so multiple URLs could be used

Background Information / Reproduction Steps
===========================================
The captured mdns-package, (the 141.-Address is the correct external address of the network adapter, all other adapers are either disabled or disconnected)

![grafik](https://user-images.githubusercontent.com/3939648/48418741-5d1a2b80-e756-11e8-8863-299a2c2eb0ad.png)

Checklist
=========
Please provide the following information:

 - [x] open62541 Version (release number or git tag): 5a06ca2471fa973cddbac90553cd516db0809f04 with #2211 merged
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Windows 10
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [x] Wireshark network dump attached (availiable on request by mail), if the picture is not enought
 - [ ] Self-contained code example attached
 - [ ] Critical issue

ccvca:
I've just think about another solution:
What about sending an individual mDNS-Message per Interface, which contains only the IP-Address of that interface. As mDNS does only works inside an subnet, other IP-Adresses might not be routable between the different connected Networks, especially when private IP-Adresses are used (10.0.0.0/8 192.168.0.0/16, 172.16.0.0/12)

Pro:
> What about sending an individual mDNS-Message per Interface, which contains only the IP-Address of that interface.

To me that makes perfectly sense, but it may not be so easy to implement since the mDNS library is handling some of the stuff separately. You can try to create a PR and then we can check the solution. That would be really nice :+1:

Pro:
This will be solved with #2344. We should not use the A record at all since the spec defines that the SRV announce should always include the hostname, and not an IP address. The OS will then resolve the hostname to the corresponding IP.

PR #2344, opened 2019-01-04
mDNS: Use hostname instead of resolving IP address

Pro:
The spec says that mDNS should announce the hostname, and this hostname should be used for the discovery URL.

image

An IP address for the SRV.hostname is not allowed.

Up to now we took the announced IP address and tried to resolve it to a hostname, which may lead to various issues (e.g. #2322, #2212). This procedure was done for every received mDNS record.

### DIFF
--- a/src/server/ua_server_discovery_mdns.c
+++ b/src/server/ua_server_discovery_mdns.c
@@ -30,21 +30,6 @@
 # include <netdb.h> // for recvfrom in cygwin
 #endif
 
-#ifndef UA_STRDUP
-# if defined(__MINGW32__)
-static char *ua_strdup(const char *s) {
-    char *p = (char*)UA_malloc(strlen(s) + 1);
-    if(p) { strcpy(p, s); }
-    return p;
-}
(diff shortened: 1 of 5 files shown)


src/server/ua_server_discovery_mdns.c
```diff
@@ -615,10 +495,14 @@ mdns_create_txt(UA_Server *server, const char *fullServiceDomain, const char *pa
         xht_set(h, "path", "/");
     } else {
         /* path does not contain slash, so add it here */
-        if(path[0] == '/')
-            /* todo: malloc in strdup may fail: return a statuscode */
-            allocPath = UA_STRDUP(path);
-        else {
+        if(path[0] == '/') {
+            allocPath = (char*)UA_malloc(sizeof(path));
```
ccvca: Is sizeof intended here? A few lines later strlen(path) [+ 1] is used for this.

Pro: Oooops, no, should be strlen.... Thanks

PR #2349, opened 2019-01-05
Do not dynamically allocate a simple bool

basyskom-meerkoetter:
This PR is also an RFC.  :-)
Is there a good reason to allocate the bool on the heap? I saw it being done like this in a few places. I would tend to convert them all but want to collect feedback first.

### DIFF
--- a/tests/client/check_client.c
+++ b/tests/client/check_client.c
@@ -20,7 +20,7 @@
 
 UA_Server *server;
 UA_ServerConfig *config;
-UA_Boolean *running;
+UA_Boolean running;
 UA_ServerNetworkLayer nl;
 THREAD_HANDLE server_thread;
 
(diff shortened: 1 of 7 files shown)


Pro:
I'd also with the stack version. At least I don't see any reason why it should be on the heap.

basyskom-meerkoetter:
pushed a more complete patch where all instances of this pattern are converted

jpfr:
We originally had an issue where users ran `UA_Server_run(server, &running)` in a background thread.
And wondered why this crashed.
Of course, stack-values cannot be accessed from another thread.
This came up a few times.

👍 For the initiative of @basyskom-meerkoetter ! This makes the code simpler.

In addition we should add a big warning to `UA_Server_run`.

Pro:
Thank you very much!

PR #2354, opened 2019-01-07
Fix: Segfault when path == NULL

ccvca:
Handle path== NULL when parsing a mdns message (which was not sent by open62541).


ccvca:
I've no idea, how to fix this Codacy error and why it has not found it before.

Pro:
Codacy only checks the diff for new flaws, but not the already existing code. strlen was included in your diff, thus codacy thins you added it.

Anyways, I just disabled the Codacy strlen error, since we can not do it in another way here.

Thanks for the fix

ccvca:
strlen has been there before (without an error by codacy).

Pro:
The error was there before, and is also listed in the issue list on codacy, but it was not enforced, since nobody touched the code around it since enabling codacy.

PR #2359, opened 2019-01-09
Provide information for ExtensionObject decoding

matkonnerth:
adds DataTypeDictionary to the nodeset.
Fixes https://github.com/open62541/open62541/issues/1420

### DIFF
--- a/examples/nodeset/testnodeset.csv
+++ b/examples/nodeset/testnodeset.csv
@@ -1,5 +1,7 @@
 Structs,5001,Object
 Point_Encoding_DefaultBinary,5002,Object
+NestedPoint_Encoding_DefaultBinary,5003,Object
+PointWithArray_Encoding_DefaultBinary,5004,Object
 BaseDataTypes,5100,Object
 BaseDataTypes_Double,5101,Variable
 BaseDataTypes_Uint32,5102,Variable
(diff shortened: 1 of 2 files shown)


Pro:
Doesn't this still use the Base64 encoded .bsd DataTypeDictionary here:
https://github.com/open62541/open62541/pull/2359/commits/34afc3afca82cff5220a8d57b8ce285907a99f8e#diff-e1a0a8ac64e8dad0583b0ab1bc4984c1R76

I tried removing that, but then UaExpert does not show the content of the Point anymore.

As you already mentioned here https://github.com/open62541/open62541/issues/1420#issuecomment-452802314, DataTypeDictionary is deprecated.

Were you able to construct a nodeset which uses DataTypeDefinition and UaExpert is able to show the content?

According to https://forum.unified-automation.com/topic2712.html#p7157 UaExpert should be able to use the DataTypeDefinition starting with Version 1.5.0

matkonnerth:
You are right, it is using the base64 encoded bsd DataTypeDictionary.

As I mentioned in https://github.com/open62541/open62541/issues/1420#issuecomment-451426677, we would have to extend the DataTypeNodes and DataTypeAttributes and implement the reading of DataTypeDefinition attribute from DataTypeNode.

Or do I misunderstand something here?

Pro:
As far as I can tell you are right with your statement, that we probably need to extend the DataTypeNodes and DataTypeAttributes.

I will merge this PR for now, but keep #1420 open. If you are interested in providing a PR for this extension, we would be really happy!

Thank you for your support!

PR #2362, opened 2019-01-11
Client history update

peter-rustler-basyskom:
This adds client and server historical update.


jpfr:
Most of the PR is probably outdated now that the server-side was merged.
The client part looks good...

PR #2367, opened 2019-01-14
PROPOSAL: fix debian multiarch and several lintian errors

jrcoding:
**Please take this pull request as a PROPOSAL**  
for a discussion to make the debianization more robust. 

I'm using a cowbuilder/pbuilder environment to create debian packages in a safe minimalistic debian environment and got some errors which I tried to solve. 
With the proposed changes the package builds fine in a cowbuilder/pbuilder setup.

### DIFF
--- a/src/pubsub/ua_pubsub.c
+++ b/src/pubsub/ua_pubsub.c
@@ -1053,9 +1053,11 @@ sendNetworkMessage(UA_PubSubConnection *connection, UA_DataSetMessage *dsm,
  * contained DataSetMessages. */
 void
 UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
+    UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "Publish Callback");
+
     if(!writerGroup) {
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "Publish failed. WriterGroup not found");
+        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                       "Publish failed. WriterGroup not found");
         return;
     }
 
@@ -1065,8 +1067,8 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
 
     /* Binary encoding? */
     if(writerGroup->config.encodingMimeType != UA_PUBSUB_ENCODING_UADP) {
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "Unknown encoding type.");
+        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                       "Publish failed: Unknown encoding type.");
         return;
     }
 
@@ -1074,8 +1076,8 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
     UA_PubSubConnection *connection =
         UA_PubSubConnection_findConnectionbyId(server, writerGroup->linkedConnection);
     if(!connection) {
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "Publish failed. PubSubConnection invalid.");
+        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                       "Publish failed. PubSubConnection invalid.");
         return;
     }
 
@@ -1099,35 +1101,30 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
         UA_PublishedDataSet *pds =
             UA_PublishedDataSet_findPDSbyId(server, dsw->connectedDataSet);
         if(!pds) {
-            UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                         "PubSub Publish: PublishedDataSet not found");
+            UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                           "PubSub Publish: PublishedDataSet not found");
             continue;
         }
 
-        /* If promoted fields are contained in the PublishedDataSet, then this
-         * DSM must encapsulated in one NM. Send it right away. */
-        if(pds->promotedFieldsCount > 0 || maxDSM == 1) {
-            UA_DataSetMessage dsm;
-            UA_StatusCode res = UA_DataSetWriter_generateDataSetMessage(server, &dsm, dsw);
-            if(res != UA_STATUSCODE_GOOD) {
-                UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                             "PubSub Publish: DataSetMessage creation failed");
-                continue;
-            }
-
-            res = sendNetworkMessage(connection, &dsm, &dsw->config.dataSetWriterId, 1);
-            if(res != UA_STATUSCODE_GOOD)
-                UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                             "PubSub Publish: Could not send a NetworkMessage");
-            UA_DataSetMessage_free(&dsm);
-            continue;
-        }
-
-        UA_StatusCode res2 =
+        /* Generate the DSM */
+        UA_StatusCode res =
             UA_DataSetWriter_generateDataSetMessage(server, &dsmStore[dsmCount], dsw);
-        if(res2 != UA_STATUSCODE_GOOD) {
-            UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                         "PubSub Publish: DataSetMessage creation failed");
+        if(res != UA_STATUSCODE_GOOD) {
+            UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                           "PubSub Publish: DataSetMessage creation failed");
+            continue;
+        }
+
+        /* Send right away if there is only this DSM in a NM. If promoted fields
+         * are contained in the PublishedDataSet, then this DSM must go into a
+         * dedicated NM as well. */
+        if(pds->promotedFieldsCount > 0 || maxDSM == 1) {
+            res = sendNetworkMessage(connection, &dsmStore[dsmCount],
+                                     &dsw->config.dataSetWriterId, 1);
+            if(res != UA_STATUSCODE_GOOD)
+                UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                               "PubSub Publish: Could not send a NetworkMessage");
+            UA_DataSetMessage_free(&dsmStore[dsmCount]);
             continue;
         }
 
@@ -1144,8 +1141,8 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
         UA_StatusCode res3 = sendNetworkMessage(connection, &dsmStore[i * maxDSM],
                                                 &dsWriterIds[i * maxDSM], nmDsmCount);
         if(res3 != UA_STATUSCODE_GOOD)
-            UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                         "PubSub Publish: Sending a NetworkMessage failed");
+            UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                           "PubSub Publish: Sending a NetworkMessage failed");
     }
 
     /* Clean up DSM */


jrcoding:
Before merging we should think about the numbering scheme in debian/changelog...

jpfr:
> Before merging we should think about the numbering scheme in debian/changelog...

What do you propose?

Pro:
Before merging, I need to check a bit more, I'm OoO.

Especially: How should our package be named?

libopen62541
open62541

This PR changes this to the lib version and also removes the dbg package which we should keep

jrcoding:
lintian complains if a library package doesn't start with libxxxxx because it's not debian standard
The dbg package will be created automatically by debhelper if compat level is >=9

ralphlange:
@jrcoding was asking me for a comment, so I had a look at the [Debian policy, chapter "Shared Libraries"](https://www.debian.org/doc/debian-policy/ch-sharedlibs.html), which states that:

> Normally, the run-time shared library and its `SONAME` symlink should be placed in a package named libraryname<i>soversion</i>, where _soversion_ is the version number in the `SONAME` of the shared library. Alternatively, if it would be confusing to directly append soversion to libraryname (if, for example, libraryname itself ends in a number), you should use libraryname-soversion instead.

In this case: the installed library is `libopen62541.so.0.4.0`, its `SONAME` is `0`, so the Debian package name should be `libopen62541-0` - which is exactly what Lintian suggests. This is just the base name; the upstream package version (probably `0.4.0~dev` atm) and Debian revision are added.

_Disclaimer:_ Of course, this is just a policy, and only applies strictly to packages in the Debian release. Still, packaging that follows those rules is less likely to create trouble, and changing established things is always more work than starting in the right way.

Pro:
@jrcoding and @ralphlange thanks for your valuable feedback! We are not debian experts, thus we are happy for any input.

I don't mind changing the name to libopen62541 to be more consistant. Also adding the SONAME version to the package seems a good idea.

Then we probably also need to rename this file for the -dev package:
https://github.com/open62541/open62541/pull/2367/files#diff-9003a6a91c6df992c132226a0a3bf357

@jrcoding could you update the PR (if the renaming for -dev is necessary).

Regarding the `compat` version:
If I remember correctly, I think there were problems with a newer version on the launchpad build system. The daily builds did require some older version:
https://launchpad.net/~open62541-team/+archive/ubuntu/daily
I'm not sure though. We can simply test with compat >=9

ralphlange:
The `SONAME` in the library package name is important to allow installing multiple ABI versions of a shared library in parallel without clashing. (So that e.g. `oldapplication` can depend on `libopen62541-0` and `newapplication` can depend on `libopen62541-1` and happily reside next to each other, while both library packages can be updated independently.)

The name of most other related packages (-dev, -doc etc.) is less regulated and could in this case be either `libopen62541` or `open62541`. Personally I would lean towards the latter because it matches the project name, but there are examples for both within the Debian distribution.

debian/open62541-dev.install
```diff
@@ -1,5 +1,2 @@
-usr/include/*
-usr/lib/lib{*.a,*.so,*.so.*}
-usr/lib/cmake/open62541/*
```
Pro: I guess we need this line since we have some custom CMake macros which are added to the dev package.

You can check the content here:

https://launchpad.net/~open62541-team/+archive/ubuntu/daily/+files/open62541-dev_0.4-0~201901132115~ubuntu19.04.1_arm64.deb

jrcoding: hmm, not sure that I've got it...
For the open62541-dev package is a dependency in debian/control to libopen62541 and the dependency ensures that the lib will be installed if installation of open62541-dev is requested. Therefore everything in usr/lib (even usr/lib/cmake/open62541/*) will be installed if one installs open62541-dev.

jrcoding:
o.k., I will try to prepare an updated PR  including the following:
taking into account the remarks from @ralphlange 

sourcepackage name: `open62541`
dev package name: `open62541-dev`
lib package name: `libopen62541-0`

open62541-dev :
changes from `architecture all` to `any` 
includes `unversioned shared library link`
includes `everything in usr/include`
includes `everything in usr/share`

any changes and/or anything else...???

ralphlange:
I would opt for
- proper Multiarch support
- a doc package

Should we work on this together (off-line) and re-ping this PR when it's all shiny?!

jrcoding:
yes, you are welcome, good idea....

jpfr:
Dear all,

thanks for this! It is great if we can support other Debian packages to use open62541 with some a sane default feature-set and build options.

Developers that require all the bells and whistles better use the github source.
A lot of configuration options are set at compile-time and baked into the library.

Is it okay to use **lib** as a postfix?

- open62541-lib (for use by other Debian binary packages)
- open62541-dev (to build those other projects and for developers that are happy with the default compile-time settings)
- libopen62541-doc (pdf + html docs)
- libopen62541-dbg (debug symbols)
- open62541-tools (see the /tools folder: ua2json, nodeset compiler, self-signed certs, and so on.)
   - These are standalone applications that may use open62541-lib internally)

Multiarch is a good idea. But can we do it in a second PR?
Just so that us mere mortals can follow along.

ralphlange:
Except for the library package name that Debian policy mandates as `libopen62541-0` (see comment above) we'll try to stick with consistently using `open62541`. (The debug symbols might have to closely follow the binary package they apply to.) No worries - we'll make this as Debian-normal as possible.

Let's wait how much change is actually needed for Multiarch - shouldn't be a big deal nowadays.

Two real questions for upstream:

1. When we do a separate -tools package, are there necessary development tools that should rather go to -dev instead? Or should -dev "Recommend:" -tools? (That's a hint meaning "better also install _that_ as it is often used together with _this_".)

2. What is the mainstream configuration that Debian users should get when installing the packages, and how do we set this configuration? (Should be driven by strict usability criteria: which is the configuration that will work for most perspective users?)

jrcoding:
And some more questions for upstream:

  1. is it feasible and reasonable to configure open62541 in a way to get both libraries (shared and static) build and installed?
  2. should we include the amalgamated source and maybe header and install it for example in /usr/share (may be helpfull for some users of the lib?)
  3. should multithreaded be configured or is it still experimental

Pro:
> And some more questions for upstream:
> 
>     1. is it feasible and reasonable to configure open62541 in a way to get both libraries (shared and static) build and installed?

Is it standard for other packages to install both versions? I guess (without having a lot of expertise knowledge in this area) that a shared lib is more the standard way?

>     2. should we include the amalgamated source and maybe header and install it for example in /usr/share (may be helpfull for some users of the lib?)

By default we recommend to install the NON-amalgamated version. Also all our examples expect the non-amalgamated headers. See here for further discussion on that:

Also have a look at #2307 and #2296 and #2302

> 
>     3. should multithreaded be configured or is it still experimental


Multithreaded is still very experimental, thus it should be disabled. The recommended CMake is this one:
https://open62541.org/doc/current/installing.html

ralphlange:
Sneak peek at work in progress: https://github.com/ralphlange/open62541/tree/feature/debianization

_Careful: better consider that branch read-only at the moment, as it is highly unstable - I am rewriting commits frequently._

So far: bump to dh 9, Multiarch, proper minimalistic lib and -dev packages, misc small improvements.

I tried a minimal invasive way to add building the static library to CMake.
But I am not a CMake person at all: if you know a better or shorter or otherwise preferred way, please tell me.

Question:
Is that minimal -dev package useful as it is, or are there things from `tools` or `usr/share/open62541/tools` that should be in the -dev package? (Otherwise they will end up in -tools.)

Next:
-tools package.

ralphlange:
Update: we're moving into the right direction.

At this point, the minimal library and -dev packages look good, Lintian is happy, the Travis build succeeds, and @jrcoding is able to compile and run a test application against the installed packages.

(The stuff from `usr/shar/open62541/tools` seems not to be needed for development and will be added to the -tools package.)

Pro:
IMHO we do not need a separate tools package, since all the stuff from the tools dir is basically required for advanced development. E.g. the nodeset compiler and the corresponding CMake macros are required, if you want to add your own nodeset to the server.

ralphlange:
The lib-dev package should be kept minimal, because any application using the lib will have lib-dev as a build dependency, and for a simple and fast build you will need the headers, but not the advanced development tools (and their dependencies like python).

It seems to me there are three phases in application development (that map to use cases for the packaging)
Develop: create application sources and configuration using advanced development tools.
Build: compile application using headers, static library or shared library (through unversioned link).
Run: execute application binary using shared library (through SONAME link).

To stay simple, I would propose to start with the following minimal deb package set to match the use cases:
**open62541**: user facing package for (human) advanced development of applications
contains: all tools
depends on: libopen62541-dev
suggests: open62541-doc
**open62541-doc**: documentation package
contains: all documentation
**libopen62541-dev**: low level library development package
contains: headers, static lib, unversioned shared link, (cmake things if required for building applications)
depends on: libopen62541-0
**libopen62541-0**: shared library package
contains: shared library, SONAME versioned link

The user instructions are clear and easy: on Debian/Ubuntu, `apt install open62541`
A packaged application would add `Depends: libopen62541-0` and `Build-Depends: libopen62541-dev` to only pull the minimal required set of packages for its build.

Should the client and server libraries ever be separated (which - looking at the size of the shared lib - might be happening at some point?), lib and lib-dev would be split into common, server and client parts, and two additional user facing packages (open62541-server and open62541-client) could depend on the right set of tools and low level packages. Migration of packaging would be pretty standard in such a case.

Sorry for my epic comments - but I want upstream to be aware of what is done and why. (And have it documented. Add as a README in the debian folder?)

ralphlange:
Question for upstream:

To verify my packaging attempts, I would like to have a small example application that is CMake based and uses some of the tools.
The code in `open62541/examples` looks like what I need, but doesn't build for me and I didn't find enough documentation to get it building.

Any hints on how to get something to test the development package(s)?

jpfr:
> The code in open62541/examples looks like what I need, but doesn't build for me and I didn't find enough documentation to get it building.

Can you tell us where you got blocked?
The examples should be against the shared-object library with the public headers...

ralphlange:
Here's what I tried (not much experience with CMake, so this might be naïve)...
_Environment: Debian Testing (buster/sid), gcc 8.2.0, cmake 3.11.2_
```
ralph@debian-testing:~/work/open62541$ cd examples
ralph@debian-testing:~/work/open62541/examples$ mkdir build
ralph@debian-testing:~/work/open62541/examples$ cd build
ralph@debian-testing:~/work/open62541/examples/build$ cmake ..
-- The C compiler identification is GNU 8.2.0
-- The CXX compiler identification is GNU 8.2.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:31 (assign_source_group):
  Unknown CMake command "assign_source_group".
  CMakeLists.txt:44 (add_example)


CMake Warning (dev) in CMakeLists.txt:
  No cmake_minimum_required command is present.  A line of code such as

    cmake_minimum_required(VERSION 3.11)

  should be added at the top of the file.  The version specified may be lower
  if you wish to support older CMake versions for this project.  For more
  information run "cmake --help-policy CMP0000".
This warning is for project developers.  Use -Wno-dev to suppress it.

See also "/home/ralph/work/open62541/examples/build/CMakeFiles/CMakeOutput.log".
```

After adding `cmake_minimum_required(VERSION 2.8.8)` (as in the packaging), I still get

```
ralph@debian-testing:~/work/open62541/examples/build$ cmake ..
-- The C compiler identification is GNU 8.2.0
-- The CXX compiler identification is GNU 8.2.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:32 (assign_source_group):
  Unknown CMake command "assign_source_group".
  CMakeLists.txt:45 (add_example)


See also "/home/ralph/work/open62541/examples/build/CMakeFiles/CMakeOutput.log".
```

I can see that `assign_source_group` is defined in the project's cmake macros one level up, but as the example uses it without doing any `include` or `add_subdirectory` I decided that this is out of my cmake-fu and rather ask.

ralphlange:
I am assuming "yes" for the last question.

Here are more things - for the tools package - where you will have to help me:

- What in the mixed bag under `open62541/tools` is actually a development tool that is needed or useful for developing an application, i.e. should be in the `open62541-tools` package? The node compiler probably. Anything else?
- As far as I can see, the node compiler needs the UA nodeset definitions (under `deps/ua-nodeset`) to work. Should these also be packaged? If yes, how?

ralphlange:
Should we start from what CMake is configured to do?
```
set(UA_install_tools_dirs "tools/certs"
    "tools/nodeset_compiler"
    "tools/schema"
    "deps/ua-nodeset")

set(UA_install_tools_files "tools/generate_datatypes.py"
    "tools/generate_nodeid_header.py"
    "tools/generate_statuscode_descriptions.py")
```
But @jpfr also mentioned `ua2json` which currently is not part of the CMake configuration?!

Also: which of the scripts in these tools are called from the command line? (That slightly changes their packaging.)

Pro:
> I can see that `assign_source_group` is defined in the project's cmake macros one level up, but as the example uses it without doing any `include` or `add_subdirectory` I decided that this is outside of my cmake-fu and rather ask.

You can not directly build inside the examples folder. That CMakeLists is included by the main CMakeLists file. Better use the following steps:

```bash
cd open62541
mkdir build && cd build
cmake -DUA_BUILD_EXAMPLES=ON ..
```

That should build the examples with the library.

> For the Travis-CI builds, I need to switch the underlying Ubuntu distribution for the job that builds the Debian packages from trusty to xenial. [...]
> To not affect the setup for the existing builds, I would suggest that I create an additional Travis-CI job

Sounds good to me. Let's see how the build looks inside the PR then we can modify if necessary.

> - What in the mixed bag under `open62541/tools` is actually a development tool that is needed or useful for developing an application, i.e. should be in the `open62541-tools` package? The node compiler probably. Anything else?

Everything which is copied by the `make install` command inside the `tools` dir should be a development dependency. 

See https://github.com/open62541/open62541/blob/master/CMakeLists.txt#L1121

```cmake
set(UA_install_tools_dirs "tools/certs"
    "tools/nodeset_compiler"
    "tools/schema"
    "deps/ua-nodeset")

set(UA_install_tools_files "tools/generate_datatypes.py"
    "tools/generate_nodeid_header.py"
    "tools/generate_statuscode_descriptions.py")
```

I would recommend that you use the output of make install for the packaging. Therefore you should not need to care which stuff is required from the tools directory. If we have this list also somewhere else (e.g. in the debian config files) this will definitely lead to inconsistencies as soon as someone changes the CMakeLists.
The same holds true for all the headers and remaining stuff. I.e. you debian config should not specify its own list of files for any package at all.

> - As far as I can see, the node compiler needs the UA nodeset definitions (under `deps/ua-nodeset`) to work. Should these also be packaged? If yes, how?

I'd put this also in the tools package. Have a look at this comment:
https://github.com/open62541/open62541/pull/2307#issuecomment-448544779
There you'll find a list of files which are installed by default using the make install command. Even better execute `cmake -DCMAKE_INSTALL_PREFIX=$HOME/some_dir .. && make -j install` then you will see what is currently installed.

The resulting debian packages should reproduce the same file structure if possible.

ralphlange:
Thanks for the detailed answer!!
I understand where you're trying to go to, and we will try to make it work that way as far as possible.

Remaining packaging issues:
- _Tools:_ Debian requires all user facing commands (everything the user is expected to run interactively from command line) to be installed in `/usr/bin`, to not have any file name extension and to have a man page. As you want the original names and locations of the tools to be kept, I will add shell script wrappers in `/usr/bin` that call the scripts in their real location. So: which of the scripts are interactive user commands?
- _ua-nodeset:_ What copyright/license applies to these files? (Debian is picky about that.)
- _ua2json:_ Should that be built and installed as a tool? Being an architecture-dependent binary this needs a separate package from the python scripts (that are architecture-independent).

The examples not being able to get built stand-alone makes them unusable for being packaged as part of the documentation and - even worse - unusable for validating the Debian packages.
Without having something that uses the tools and the library from the installed packages, we can't do any QA on them.
Should I create a separate issue and/or PR for making the examples stand-alone?

Last not least:
- Travis: I also cleaned-up the lists of pre-installed packages, matching it to the actual build-dependencies and installing the python related packages in the correct version for the python being used. In principle, these changes are mostly independent from the Debian packaging revamp. Do you want them as a separate PR?

ralphlange:
For the record, the answer to the 2nd bullet (ua-nodeset license): the MIT license applies.
See: https://github.com/OPCFoundation/UA-Nodeset/issues/11

Pro:
>     * _Tools:_ Debian requires all user facing commands [...]. So: which of the scripts are interactive user commands?

IMHO the only script which can be called by the user and which would make sense to provide is the ua2json tool. All the other scripts (nodeset_compiler.py, generate_datatypes.py, ...) are normally called directly from CMake and probably should not be available as an interactive command. We can also change this later if we see the need to provide a script as a command.

>     * _ua-nodeset:_ What copyright/license applies to these files? (Debian is picky about that.)
>

MIT. See https://github.com/OPCFoundation/UA-Nodeset/issues/11

 
>     * _ua2json:_ Should that be built and installed as a tool? Being an architecture-dependent binary this needs a separate package from the python scripts (that are architecture-independent).

For the first PR I would say we leave those special cases out and try to make the PR as slim as possible. We can always add additional commands/tools/files with additional PRs.


> The examples not being able to get built stand-alone makes them unusable for being packaged as part of the documentation and - even worse - unusable for validating the Debian packages.
> Should I create a separate issue and/or PR for making the examples stand-alone?

Yes, you are correct. In general the examples should be buildable standalone. Our current approach was that the user creates his own CMake project and then copies the examples to his local folder structure. It would be nice if someone could just get the examples folder and then build everything inside using the commands as you did above.
For that we need to add a bit more logic to the exampleS/CMakeLists.txt that it detects if it is called inside the main repo or as a standalone. If the latter case aplies, then it should use find_package to find open62541 dev files.
See here https://open62541.org/doc/current/installing.html

It would be nice if you can create a new issue for this, or even better a PR for review :)


>     * Travis: I also cleaned-up the lists of pre-installed packages, matching it to the actual build-dependencies and installing the python related packages in the correct version for the python being used. In principle, these changes are mostly independent from the Debian packaging revamp. Do you want them as a separate PR?

Yes, definitely! We always like cleanup actions! Please create a separate PR for this so that we can merge it in.


Thank you for all your support!!!

PR #2371, opened 2019-01-16
server_historical_update: Add api to alter raw data.

peter-rustler-basyskom:
This is the server side of the history update functionality.
See the https://github.com/open62541/open62541/pull/2362 pull request for tests.

### DIFF
--- a/include/ua_plugin_access_control.h
+++ b/include/ua_plugin_access_control.h
@@ -88,6 +88,21 @@ struct UA_AccessControl {
     UA_Boolean (*allowDeleteReference)(UA_Server *server, UA_AccessControl *ac,
                                        const UA_NodeId *sessionId, void *sessionContext,
                                        const UA_DeleteReferencesItem *item);
+
+    /* Allow insert,replace,update of historical data */
+    UA_Boolean (*allowHistoryUpdateUpdateData)(UA_Server *server, UA_AccessControl *ac,
+                                               const UA_NodeId *sessionId, void *sessionContext,
+                                               const UA_NodeId *nodeId,
+                                               UA_PerformUpdateType performInsertReplace,
+                                               const UA_DataValue *value);
+
+    /* Allow delete of historical data */
+    UA_Boolean (*allowHistoryUpdateDeleteRawModified)(UA_Server *server, UA_AccessControl *ac,
+                                                      const UA_NodeId *sessionId, void *sessionContext,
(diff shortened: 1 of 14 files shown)


include/ua_plugin_access_control.h
```diff
@@ -88,6 +88,31 @@ struct UA_AccessControl {
     UA_Boolean (*allowDeleteReference)(UA_Server *server, UA_AccessControl *ac,
                                        const UA_NodeId *sessionId, void *sessionContext,
                                        const UA_DeleteReferencesItem *item);
+
```
jpfr: This a pretty detailed API. And it will become a lot more verbose when more features are added.
Can we model the API according to the DataType definitions for HistoryUpdate?

Part 11 defines possible content for HistoryUpdateDetails:
- UpdateDataDetails
- UpdateStructureDataDetails
- UpdateEventDetails
- DeleteRawModifiedDetails
- DeleteAtTimeDetails
- DeleteEventDetails

src/server/ua_services_attribute.c
```diff
@@ -1511,6 +1511,54 @@ Service_HistoryRead(UA_Server *server, UA_Session *session,
                                            response, historyData);
     UA_free(historyData);
 }
+
+void
+Service_HistoryUpdate(UA_Server *server, UA_Session *session,
+                    const UA_HistoryUpdateRequest *request,
+                    UA_HistoryUpdateResponse *response) {
+    response->resultsSize = request->historyUpdateDetailsSize;
+    response->results = (UA_HistoryUpdateResult*)UA_Array_new(response->resultsSize, &UA_TYPES[UA_TYPES_HISTORYUPDATERESULT]);
```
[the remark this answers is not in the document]
peter-rustler-basyskom: Done

include/ua_plugin_access_control.h
peter-rustler-basyskom: Ok, I see your problem here.

But the details corner is not the right place to connect the access control.

In UpdateDataDetails you can have more than one update.  For example 3 times a replace and 2 updates and 1 insert. With that you can not say: Insert is OK and replace is not.......

But I found a way which suites better, so we have only one access function for UpdateDataDetails and DeleteRawModifiedDetails.

jpfr:
Perfect PR!
Happy to merge. 👍

PR #2378, opened 2019-01-18
Bugfix: delete subscriptions that are closed by the server

wbitesi:
If the OPC-UA server has closed the session without the client knowing, the
subscription should be deleted.

We observed this behavior with open62541 UA Client against a Beckhoff twincat OPC-UA server when the connection was lost (briefly). The bugfix simply checks if the subscription is already removed on the open62541 client side (and some outstanding publish background checks are coming in) or if the subscription is still active on the client but closed on the server.


jpfr:
Can you trigger the UA_Client_DeleteSubscriptionCallback in this code path as well?

wbitesi:
In UA_Client_Subscription_deleteInternal(UA_Client*, UA_Client_Subscription) (ua_client_subscriptions.c:106) the delete callback is already called.

jpfr:
Right, Thanks for the fix!

Issue #1936, opened 2018-07-21
securityPolicyEndPoint can not disable None/None connection


7486913:
Description
===========
When I use the following function, it works very well but UaExpert still can connect to my OPC Server Successfully without any certificate. 

In UaExpert, set  **Security Policy**  as None/**Message Security Mode** as None, it will pass authority 
and connect to OPC Server.

 UA_ServerConfig* config = UA_ServerConfig_new_basic256sha256(serverPort,&certificate,&privateKey,
                                      trustlist,trustlistSize,
                                      revocationList,revocationListSize);

I note there are two times connection in logs, so  change codes in ua_securechannel.c.
UA_StatusCode
UA_SecureChannel_init(UA_SecureChannel *channel,
                      const UA_SecurityPolicy *securityPolicy,
                      const UA_ByteString *remoteCertificate) {
    if(channel == NULL || securityPolicy == NULL || remoteCertificate == NULL)
        return UA_STATUSCODE_BADINTERNALERROR;

    /* Linked lists are also initialized by zeroing out */
    memset(channel, 0, sizeof(UA_SecureChannel));
    channel->state = UA_SECURECHANNELSTATE_FRESH;
    channel->securityPolicy = securityPolicy;

    UA_StatusCode retval;
    if(channel->securityPolicy->certificateVerification != NULL) {
        retval = channel->securityPolicy->certificateVerification->
            verifyCertificate(channel->securityPolicy->certificateVerification->context, remoteCertificate);

        if(retval != UA_STATUSCODE_GOOD)
            return retval;
    } else {
        UA_LOG_WARNING(channel->securityPolicy->logger, UA_LOGCATEGORY_SECURITYPOLICY, "No PKI plugin set. "
            "Accepting all certificates");
		
    }
	//first connect
	static int time = 0;
	if (time == 0) {
		time = 1;
	}
	else{
        //second connect if remoteCertificate is still null, close connection.
		time = 0;
		if (remoteCertificate->length == 0)
			return UA_STATUSCODE_BADINTERNALERROR;
	}
	

    retval = securityPolicy->channelModule.
        newContext(securityPolicy, remoteCertificate, &channel->channelContext);
    if(retval != UA_STATUSCODE_GOOD)
        return retval;

    retval = UA_ByteString_copy(remoteCertificate, &channel->remoteCertificate);
    if(retval != UA_STATUSCODE_GOOD)
        return retval;

[...]

PR #2379, opened 2019-01-18
[Review] Added possibility to use SecurityPolicy None for discovery only

jonasgreen88:
If no endpoint with SecurityPolicy None is available, it shall be
possible to open a secure channel with SecurityPolicy None to perform
the discovery services GetEndpoints, FindServers and
FindServersOnNetwork. See issue #1936 for details.

### DIFF
--- a/include/open62541/server_config.h
+++ b/include/open62541/server_config.h
@@ -140,6 +140,15 @@ struct UA_ServerConfig {
     size_t endpointsSize;
     UA_EndpointDescription *endpoints;
 
+    /* Only allow the following discovery services to be executed on a
+     * SecureChannel with SecurityPolicyNone: GetEndpointsRequest,
+     * FindServersRequest and FindServersOnNetworkRequest.
+     *
+     * Only enable this option if there is no endpoint with SecurityPolicy#None
+     * in the endpoints list. The SecurityPolicy#None must be present in the
+     * securityPolicies list. */
+    UA_Boolean securityPolicyNoneDiscoveryOnly;
+
     /* Node Lifecycle callbacks */
     UA_GlobalNodeLifecycle nodeLifecycle;
 
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -311,7 +311,7 @@ processHEL(UA_Server *server, UA_SecureChannel *channel, const UA_ByteString *ms
     ackMessage.sendBufferSize = channel->config.sendBufferSize;
     ackMessage.maxMessageSize = channel->config.localMaxMessageSize;
     ackMessage.maxChunkCount = channel->config.localMaxChunkCount;
-    
+
     UA_TcpMessageHeader ackHeader;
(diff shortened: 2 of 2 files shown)


jpfr:
What is the advantage over keeping the Endoint for #None and disallowing the session in the AccessControl plugin?
https://github.com/open62541/open62541/blob/master/include/ua_plugin_access_control.h#L40

jonasgreen88:
> What is the advantage over keeping the Endoint for #None and disallowing the session in the AccessControl plugin?
> [/include/ua_plugin_access_control.h@`master`#L40](https://github.com/open62541/open62541/blob/master/include/ua_plugin_access_control.h#L40)

The main reason for us why we would like to add this hidden DiscoveryOnly endpoint is that the server shouldn't present endpoints that the client will be unable to create sessions on and to only allow a minimum of services without message security (GetEndpoints, FindServers and FindServerOnNetwork).

sqadasi:
It would be nice to have this feature. However, the current implementation forces NonSecure endpoint to be always only for discovery, which could be a problem for different users, who want to keep NonSecure endpoint also for creating sessions.
And as @jpfr said, there is no advantage in keeping/showing the NonSecure endpoint if we are not allowed to create session on it.
So, I would suggest creating the discoveryOnly flag inside the plugin (SecurityPolicy struct), so the user would be able to set it in their application. Inside the function `processMSG` we could access the discoveryOnly flag using the passed SecureChannel parameter to block creating the session, but not discover (just as @jonasgreen88 implemented). In addition, if the discoveryOnly flag is set by user, we should block giving that endpoint on `Service_GetEndpoints` responce, so the UA CLient won't know about it at all.

jonasgreen88:
> It would be nice to have this feature. However, the current implementation forces NonSecure endpoint to be always only for discovery, which could be a problem for different users, who want to keep NonSecure endpoint also for creating sessions.

No it doesn't. If you want an endpoint with SecurityPolicy None you just have to configure the server with such an endpoint. If you only configure your server to support SecurityPolicyNone (but no endpoint with this policy) the clients can only setup a secure channel with SecurityPolicy None for discovery.

> And as @jpfr said, there is no advantage in keeping/showing the NonSecure endpoint if we are not allowed to create session on it.

This PR does not show an endpoint with SecurityPolicy None that does not allow to create sessions on it.

> So, I would suggest creating the discoveryOnly flag inside the plugin (SecurityPolicy struct), so the user would be able to set it in their application. Inside the function `processMSG` we could access the discoveryOnly flag using the passed SecureChannel parameter to block creating the session, but not discover (just as @jonasgreen88 implemented). In addition, if the discoveryOnly flag is set by user, we should block giving that endpoint on `Service_GetEndpoints` responce, so the UA CLient won't know about it at all.

This is exactly now we use it and its the intention it should work this way. If there is an endpoint created with SecurityPolicy None it is possible to create sessions on it. If there is no endpoint with this policy, but the server supports SecurityPolicy None the server only allows a secure channel with SecurityPolicy None to perform discovery.

jpfr:
@jonasgreen88 to rebase and move the boolean flag to the server config.

jpfr:
Merged to the 1.1 branch. Master is next.

Issue #2350, opened 2019-01-07
Make the asymetric decrypt API independent from the channelcontext


jonasgreen88:
The current encryption algorithm API is depends on the channelcontext. The symmetric and the asymmetric algorithms shares the same API. To be able to use this API to also decrypt passwords of an UserIdentityToken this dependency must be removed for the asymmetric algorithm. The activate session context does not have any knowledge about the channelcontext. It only knows the security policy of the token.

https://github.com/open62541/open62541/blob/3d708b820ca8637fd40682405cde9bcd2f4629bf/include/ua_plugin_securitypolicy.h#L117

PR #2390, opened 2019-01-25
Added support for server to descrypt passwords

jonasgreen88:
This PR contains an update of the accesscontrol plugin implementation to handle encrypted passwords. This example access control plugin will create one UserIdentityToken per SecurityPolicy supported by the server.

To be able to decrypt the passwords the issue #2350 needs to be solved. The asymmetric encryption API must be updated to not depend on the channel context.

### DIFF
--- a/examples/access_control/server_access_control.c
+++ b/examples/access_control/server_access_control.c
@@ -59,7 +59,8 @@ int main(void) {
     UA_ServerConfig *config = UA_ServerConfig_new_default();
     // disable anonymous logins, enable two user/password logins
     config->accessControl.deleteMembers(&config->accessControl);
-    if (UA_AccessControl_default(&config->accessControl, false, 2, logins) != UA_STATUSCODE_GOOD) {
+    if (UA_AccessControl_default(&config->accessControl, false, 2, logins,
+                                 config->securityPoliciesSize, config->securityPolicies) != UA_STATUSCODE_GOOD) {
         return EXIT_FAILURE;
     }
 
--- a/include/open62541/plugin/accesscontrol.h
+++ b/include/open62541/plugin/accesscontrol.h
@@ -13,6 +13,11 @@
 
 _UA_BEGIN_DECLS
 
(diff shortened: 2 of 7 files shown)


jpfr:
That is a great feature!

Just one request: Can we decode the UserToken before forwarding to the access control plugin?
The encryption functionality is always the same. So we should not require every possible access control policy to replicate it....

jonasgreen88:
> That is a great feature!
> 
> Just one request: Can we decode the UserToken before forwarding to the access control plugin?
> The encryption functionality is always the same. So we should not require every possible access control policy to replicate it....

Sure, that's a good point. I put it in the access control plugin as no parsing of the UserIdentityToken was done in the stack. Should all UserTokenPolicies be parsed/decrypted in the stack and forwarded as input arguments to the ActivateSession callback or should just the password be decrypted and put back decrypted into the UserIdentityToken so the ActivateSession callback receives the UserIdentityToken as it would have with an unecrypted UserIdentityToken?

jpfr:
I would prefer to put it back in.
So we can have an easy interface of just an ExtensionObject.
Otherwise we have to fan out the interface to all the different options.

jonasgreen88:
@jpfr I have moved the decryption of the passwords inside the stack. It is only the Fuzzer test that fails. Does something have to be updated to make that test pass? All other unit tests seems to pass.

I have not been able to build the stack with mbedTLS in my Windows environment. So this particular implementation has not been tested with encrypted passwords. Does the client of this stack support encrypted passwords? In that case we can maybe build a unittest for encrypted passwords.

jpfr:
Hey @jonasgreen88 can you rebase?
This is looking good, let's bring it in.

jonasgreen88:
@jpfr Great. I squashed the two commits before rebasing to avoid resolving merge conflicts with the first commit that implemented the decryption of passwords in the access control plugin. I hope this works.

jpfr:
Merged via #2812.

PR #2395, opened 2019-01-28
JSON: fix decoding of structured types

lukeMlng:
- For nested structured types the correct _UA_DataType_ is now forwarded in _decodeEntry_.
- The correct size is now set for decoded arrays.


lukeMlng:
> 
> 
> > TODO: is the length ALWAYS one size_t before the array value?
> 
> Yes, it is.
> 
> Ready to merge?

Yes, then it is ready.

PR #2411, opened 2019-02-01
CI: Only use sanitize leak if valgrind disabled

Pro:
This is a follow-up fix for #2400 

It should fix current build errors on master:
https://travis-ci.org/open62541/open62541/jobs/487317314#L7549


Infinity95:
Shouldnt it be enough to only disable asan and leave the other sanitizers enabled?

jpfr:
This is blocking other PR. So I'm merging.

If we want the other sanitizers with valgrind, feel free to add.

Pro:
I tried other combinations by e.g. just removing the leak sanitizer, but still valgrind failed.

Another option would be to use two combinations in the unit tests:
* gcc+valgrind
* clang+sanitizer

Like this we have the full features of both worlds.

PR #2423, opened 2019-02-04
Resolve port conflict

basyskom-meerkoetter:
Registering server complains that the port 4840 is already in use (as the LDS is
running there).

./discovery_server_register
[...]
[2019-02-04 22:45:20.070 (UTC+0100)] info/userland      Node read the.answer
[2019-02-04 22:45:20.070 (UTC+0100)] info/userland      read value 42
[2019-02-04 22:45:20.070 (UTC+0100)] warn/network       Error binding a server socket: Address already in use
[2019-02-04 22:45:20.070 (UTC+0100)] warn/network       Error binding a server socket: Address already in use
[2019-02-04 22:45:20.070 (UTC+0100)] info/network       TCP network layer listening on opc.tcp://prokyon:4840/

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -144,15 +144,11 @@ if(UA_BUILD_SELFSIGNED_CERTIFICATE)
   target_link_libraries(server_certificate open62541 ${open62541_LIBRARIES})
 endif()
 
-if(UA_ENABLE_DISCOVERY)
+if(UA_ENABLE_DISCOVERY_MULTICAST)
     add_example(discovery_server_lds discovery/server_lds.c)
-
     add_example(discovery_server_register discovery/server_register.c)
-
     add_example(discovery_client_find_servers discovery/client_find_servers.c)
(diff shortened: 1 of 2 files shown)


basyskom-meerkoetter:
Is also noticed that the discovery client will not work unless the open62541 build config has UA_ENABLE_DISCOVERY_MULTICAST.

The call to UA_Client_findServersOnNetwork() will fail, the example will terminate with a non-obvious message.

The LDS will print: [2019-02-04 22:37:58.705 (UTC+0100)] info/channel       Connection 6 | SecureChannel 3 | Unknown request with type identifier 12208

Is this to be expected (and should the example be only be compiled if UA_ENABLE_DISCOVERY_MULTICAST is there?

Pro:
Yes, we should only build the example if multicast is on. Can you update this PR to include that fix?

Thank you!!!

basyskom-meerkoetter:
@Pro updated the PR

PR #2428, opened 2019-02-06
Examples: small improvements

basyskom-meerkoetter:
Working on improving the examples a bit.

### DIFF
--- a/examples/access_control/client_access_control.c
+++ b/examples/access_control/client_access_control.c
@@ -28,10 +28,8 @@ int main(void) {
     newVariableAttributes.description = UA_LOCALIZEDTEXT_ALLOC("en-US", "NewVariable desc");
     newVariableAttributes.displayName = UA_LOCALIZEDTEXT_ALLOC("en-US", "NewVariable");
     newVariableAttributes.dataType = UA_TYPES[UA_TYPES_UINT32].typeId;
-    {
-        UA_UInt32 value = 50;
-        UA_Variant_setScalarCopy(&newVariableAttributes.value, &value, &UA_TYPES[UA_TYPES_UINT32]);
-    }
+    UA_UInt32 value = 50;
+    UA_Variant_setScalarCopy(&newVariableAttributes.value, &value, &UA_TYPES[UA_TYPES_UINT32]);
 
     UA_StatusCode retCode;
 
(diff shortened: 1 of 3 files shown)


examples/access_control/client_access_control.c
```diff
@@ -28,10 +28,8 @@ int main(void) {
     newVariableAttributes.description = UA_LOCALIZEDTEXT_ALLOC("en-US", "NewVariable desc");
     newVariableAttributes.displayName = UA_LOCALIZEDTEXT_ALLOC("en-US", "NewVariable");
     newVariableAttributes.dataType = UA_TYPES[UA_TYPES_UINT32].typeId;
-    {
-        UA_UInt32 value = 50;
-        UA_Variant_setScalarCopy(&newVariableAttributes.value, &value, &UA_TYPES[UA_TYPES_UINT32]);
-    }
```
ccvca: The intention of these brackats were that you need the value-Variable only at this place and it will be removed from the stack afterwards. But I could also understand that it might be harder to read.

basyskom-meerkoetter: @ccvc thank you for your feedback :-)

I understand what this scope does. People sometimes use this style/trick to get around name clashes. I don't see the point of scoping a single int-variable. This is unusual. I would argue that an example should be simple and easy to understand.

Nitpick. Its not removed from the stack. It is (at compile time) removed from the scope (so you program can't reference it anymore). It is removed from the stack when the stackframe of the containing function is removed.

ccvca: > Nitpick. Its not removed from the stack. It is (at compile time) removed from the scope (so you program can't reference it anymore). It is removed from the stack when the stackframe of the containing function is removed.

Thanks, I had assumed, that in the following code both "value"-Variables share the same memory.
```
{
  UA_UInt32 value = 50;
  UA_Variant_setScalarCopy(&newVariableAttributes.value, &value, UA_TYPES[UA_TYPES_UINT32]);
}

{
  UA_UInt32 value = 40;
  UA_Variant_setScalarCopy(&newVariableAttributes.value, &value, UA_TYPES[UA_TYPES_UINT32]);
}
```

jpfr:
Merged. Thanks!

PR #2430, opened 2019-02-07
Travis clean up

ralphlange:
A bunch of small changes to the Travis build, trying to make things more obvious without changing the actual builds too much:
- fix some misleading output in the build script
- make the static code analysis build a special build like the others
- group the pre-installs in .travis.yml by the feature they are needed for and remove unnecessary ones
- use Travis variables for setting compilers (reduce number of specific env variables for jobs)
- only set env variables for jobs that are needed (i.e. differ from default)

Things that still could be done (considered too disruptive for this PR):
- rename MINGW to CROSS (as it also activates cross builds for linux and raspi)
- create documentation only in the build(s) that actually use it (reduces pre-installs)

Note: The lower part of the diff (below) is misleading - the indentation of a huge block was removed (the analysis build if clause became short) which shows every line of that block as changed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,6 @@ matrix:
             - xutils-dev
             - zip
       env:
-        - ANALYZE=false
         - CC=gcc-4.8
         - CXX=g++-4.8
         - PYTHON=python2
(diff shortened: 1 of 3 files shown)


Pro:
Very cool! Thank you very much for cleaning up this somewhat mess! The travis build matrix is now much cleaner. The total build time is now even reduced by almost 20 minutes.

ralphlange:
Welcome!

PR #2441, opened 2019-02-11
use GNUInstallDirs to handle multiarch correctly

jrcoding:
Installing libraries in a multiarch environment requires to set the multiarch tuple (i.e. x86_64-linux-gnu) correctly. CMake provides GNUInstallDirs for this reason. 
See also  https://wiki.debian.org/Multiarch/Implementation#CMake


jrcoding:
I close this for the moment, same reason as #2443

PR #2442, opened 2019-02-11
don't install bytecompiled python files

jrcoding:
Lintian throws an error when trying to Installi bytecompiled python files.
If the `.pyc` files really needed one should create them at package installation time in a postins script.


Pro:
Can be merged when green!
Thanks!

PR #2443, opened 2019-02-11
Debian install doc excluding temp build files

jrcoding:
When packaging for Debian this installs the needed stuff for the `HTML` at `/usr/share/doc/open62541/open62541.html` and `PDF` at `/usr/share/doc/open62541` which is the Debian standard location. Installation of the temp build files will be excluded.


jpfr:
Yes. 👍 
We should stick to the conventions.

Ready to merge when the build is green.
The current build break is here:
```
CMake Error at doc/cmake_install.cmake:40 (file):
  file INSTALL cannot find
  "/home/travis/build/open62541/open62541/obj-x86_64-linux-gnu/doc_latex/open62541.pdf".
  cmake_install.cmake:189 (include)
```

jrcoding:
Ooo, yes, two steps to fast, sorry...
This only works if the debian build files will be updated appropriate. This should definitely be part of an updated debianization PR (coming soon). 

I close it and reopen it if it becomes not part of the coming debianization PR for some reason.
Sorry for the noise...

Issue #2309, opened 2018-12-19
Move public include files into include/open62541


Pro:
Separate issue for the comment from @jpfr here https://github.com/open62541/open62541/pull/2307#issuecomment-448536613

------
I propose that we
- move our "public" header files to /include/open62541 in our source tree
- install them to /usr/include/open62541
- drop the ua_ prefix on public header files

So we can use
```c
#include <open62541/server.h>
```
for both the examples compiled against the source tree and against the installed library.

Pro:
> I propose that we
> 
>   *  move our "public" header files to /include/open62541 in our source tree

This is probably not so easy since we are not only installing the content of our `include` dir, but also the "public" files from `deps` and `arch`. This would mean that within deps we also need a separate `open62541` folder, and the same for every architecture.

>   *  install them to /usr/include/open62541
>   *  drop the ua_ prefix on public header files

Dropping the `ua_` prefix would be really nice, let's see if we find a solution for your first point.

Pro:
Additional comment from @jpfr (https://github.com/open62541/open62541/pull/2307#issuecomment-448551900)

----------

Imho the package can be "slimmed".
For discussion of removal:

* Internal headers (that are not stable within a release family)
* Precompiled python bytecode .pyc
* Schemas (internal as well as supplied by the OPC Foundation repository)
* Example server / client

Pro:
This is already impelmented on 1.0 and master

PR #2448, opened 2019-02-12
[REVIEW] Move headers to /usr/include/open62541

jpfr:
This is a big fat breaking change.
In order to facilitate packaging for Linux/UNIXes, we should move our headers to a dedicated directory.
Also the project has grown quite a bit.
So we need to rethink what a sensible organization for the code should look like if we grow even more.

The proposed changes are three-fold:

- Including will be done as `<open62541/server.h>`, i.e. add an organizing folder and drop the prefix for public headers
   The include will look the same for in-repo development and for installed packages.
- Plugin interface definitions are in a subfolder, eg. `<open62541/plugin/log.h>`
- Plugin implementation definitions match the folder structure and prefix
   `<open62541/plugin/log_stdout.h>`
   But plugins are still managed in a separate /plugins folder.
   The merging to /usr/include/open62541 is done only in the package.

For discussion within the core-team.


Pro:
Fixes #2309

I'm also in favour of doing it that way. Like this we are more conformant to the default way how it should be done.

It will lead to breaking changes, but I fear there's not really a way around it.

Pro:
@jrcoding and @ralphlange this change may also affect your current work in cleaning up the debian package. It will make it easier afterwards. Probably it makes sense that we first merge this PR and then your debian fixes.

Pro:
FYI @jpfr I'm currently checking some issues in this PR regarding the nodeset compiler dependencies. I'll add some more commits and try to get it to green. Please don't force push ;)

Pro:
@jpfr looks like this huge PR is finally running though.

Can be merged when green.

PR #2463, opened 2019-02-18
Decrease the size of UA_DataValue

basyskom-meerkoetter:
Shrink the size from 96 bytes to 80 bytes by reordering members to avoid padding
(64bit linux).


jpfr:
Nice improvement!

PR #2473, opened 2019-02-20
Review: Fix invalid cast in case of writing object attribute EventNotifier

### LINKED ISSUE #2471: Writing EventNotifier attribute crashes the server later
## Description

When writing to the EventNotifier attribute, the server overwrites memory locations that should not be touched. This leads to server crash later on.

- Create an object node
- Call UA_Server_writeEventNotifier on that object node
- Trigger an event on some child of that object node
- Server crash

This is due to the following:

`copyAttributeIntoNode` casts the node to `UA_ViewNode`, but the node is an `UA_ObjectNode`. The memory layout of these is not the same, so the `monitoredItemQueue` field of the object node is overwritten with the value written.

## Checklist

 - [x] open62541 Version (release number or git tag): master branch

jojastahl:
Fix #2471


jpfr:
Thanks for the fix!

PR #2451, opened 2019-02-13
[REVIEW] enable standalone examples build

jrcoding:
**FEATURE please review**
This is my first attempt to enable building the examples standalone.

This PR assumes that open62541 lib, examples  and tools are already installed at a standard place.

With the following workflow one can build the examples standalone
```
    # cd <to-my-local-test-dir>
    # cp -a /usr/share/open62541/examples .
    # OR
    # cp -a <path-to-open62541-src-tree/examples .
    # cd examples
    # mkdir build
    # cd build
    # cmake -DUA_EXAMPLES_STANDALONE=ON ..
    # OR
    # cmake -DUA_EXAMPLES_STANDALONE=ON -DUA_NAMESPACE_ZERO=FULL ..
    # make
``` 
Review, hints and comments are welcome....

PR #2480, opened 2019-02-21
[REVIEW] build examples as subproject

jrcoding:
In this PR the examples are defined as subproject to open62541. One can build the examples  standalone or, as before, during the build of the whole open62541 project.

It's assumed that open62541 lib, examples and tools are already installed at a standard place.
The following workflow builds the examples standalone
```
    # cp -a /usr/share/open62541/examples .
    # OR
    # cp -a <path-to-open62541-src-tree>/examples .
    # cd examples
    # mkdir build
    # cd build
    # cmake ..
    # OR
    # cmake -DUA_NAMESPACE_ZERO=FULL  ..
    # make
```
@Pro @ralphlange 
I've decided to make a new PR instead modifying #2451 

Please review and/or comment ...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -933,7 +933,7 @@ else()
     include_directories_private("${PROJECT_BINARY_DIR}"
                                 "${PROJECT_SOURCE_DIR}/deps"
                                 "${PROJECT_BINARY_DIR}/src_generated")
-    if(UA_ENABLE_ENCRYPTION) 
+    if(UA_ENABLE_ENCRYPTION)
         include_directories_private(${MBEDTLS_INCLUDE_DIRS})
     endif()
 
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,8 +1,57 @@
-set(examples_headers
-    ${examples_headers}
-    ${PROJECT_SOURCE_DIR}/examples/common.h
-    )
+cmake_minimum_required(VERSION 3.0)
+project(open62541-examples)
 
+# This examples folder can also be built standalone.
+# First install open62541 using `make install` then
+# copy this folder to any other location and call CMake directly:
+#
+# cp -r open62541/examples $HOME/open62541_examples
+# cd $HOME/open62541_examples
+# mkdir build && cd build
+# cmake -DUA_NAMESPACE_ZERO=FULL ..
+# make -j
+
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    # Examples are built standalone.
+
+    find_package(open62541 REQUIRED)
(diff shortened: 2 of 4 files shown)


Pro:
@jrcoding I updated your PR and made some additional changes and simplifications. Thanks again!

I tested locally, and the examples folder can now be build standalone. This is now also tested for every commit in CI:
https://github.com/open62541/open62541/pull/2480/files#diff-3056cef31c2701d267a0058c8fa125c6R83

This is possible thanks to some recent contributions by especially @traversaro (#2425, #2447)

jrcoding:
@Pro 
during the first forced push ( from 3c4930d to 17d4534 ) one line in the top level `CMakeLists.txt`
was deleted accidentally that installes the `tools/cmake` directory to `/usr/share/open62541/` . 

I think the directory `tools/cmake ` should be installed otherwise the standalone build will not work if we don't have the project directory available any more.

The `tools/cmake` are needed or we will miss some cmake functions (i.e. assign_source_group) used for a standalone build? 

CMakeLists.txt in PR before forced push  from 3c4930d to 17d4534
```
[...snip...]
set(UA_install_tools_dirs "tools/certs"
    "tools/cmake"    
    "tools/nodeset_compiler"
    "tools/schema"
    "deps/ua-nodeset")
[...snip...]

```

CMakeLists.txt in PR after forced push  from 3c4930d to 17d4534
```
[...snip...]
set(UA_install_tools_dirs "tools/certs"
    "tools/nodeset_compiler"
    "tools/schema"
    "deps/ua-nodeset")
[...snip...]

```

Pro:
No, I deleted it intentionally.
The `tools/cmake` directory contains cmake macros which are mainly used for the stack development, but should not be installed at all. These may lead to conflicts or uncertainties for users.

The assign_source_group macro is also not nexessarily needed for standalone example build. Assign_source_group mainly restructures Visual studio projects.

jrcoding:
Hmm, when I try to build the examples without having `tools/cmake ` installed or available I get something like this
```
[...snip...]
-- Found PythonInterp: /usr/bin/python (found version "2.7.15") 
CMake Error at CMakeLists.txt:30 (include):
  include could not find load file:

    AssignSourceGroup

CMake Error at CMakeLists.txt:68 (assign_source_group):
  Unknown CMake command "assign_source_group".
  CMakeLists.txt:81 (add_example)

[...snip...]
```
If `assign_source_group` is not necessary for Linux should we then introduce something like  ` if VisualStudio` or what do you think is the correct way to prevent the mentioned error?

jrcoding:
sorry, I just saw you already defined an **empty function** `assign_source_group` in your changes, that's of course is the answer to my question...

PR #2481, opened 2019-02-21
Fix: UA_client always uses custom datatypes when decoding objects

wbitesi:
The UA_Client did not use custom data types stored in UA_ClientConfig.customDataTypes to decode the stream.


jpfr:
Thanks for the fix!

PR #2488, opened 2019-02-22
Stabilize the order of node generation code produced by the Nodeset Compiler

### LINKED ISSUE #2487: The ua_namespace_xxx.c files are generated differently on every run
## Description

The C file generated to create the nodes of a NodeSet2.xml file is different on every nodeset compiler run.

## Background Information

The order of the genrated `function_xxx_n_begin` and `function_xxx_n_finish` is different on every compiler run. The actual function content is the same. For easy comparision it would be nice if the order could be stabilized.

As far as I can see, the file is generated by `/tools/nodeset_compiler/backend_open62541.py`. The nodes seem to be sorted prior to generation. So I do not understand why the order is different on every run.

jojastahl:
Fix #2487 

When sorting, a set was used which produced a on stable sort. I changed this to use a list.

### DIFF
--- a/tools/nodeset_compiler/backend_open62541.py
+++ b/tools/nodeset_compiler/backend_open62541.py
@@ -72,8 +72,8 @@ def sortNodes(nodeset):
     # used in a reference, it must exist. A Variable node may point to a
     # DataTypeNode in the datatype attribute and not via an explicit reference.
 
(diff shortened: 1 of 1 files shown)


jpfr:
Nice one!

PR #2489, opened 2019-02-22
Make default connection configurable

cabralfortiss:
The default configuration of a connection is hardcoded to 65535. In some devices with low memory this is a problem. 

This PR adds the possibility to set the default values as preprocessor variables, and if not provided, use the old ones.

Also, the client connection uses the default connection variable instead of setting it field by field

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -751,11 +751,7 @@ UA_ClientConfig_setDefault(UA_ClientConfig *config) {
     config->logger.context = NULL;
     config->logger.clear = UA_Log_Stdout_clear;
 
-    config->localConnectionConfig.protocolVersion = 0;
-    config->localConnectionConfig.sendBufferSize = 65535;
-    config->localConnectionConfig.recvBufferSize = 65535;
-    config->localConnectionConfig.maxMessageSize = 0; /* 0 -> unlimited */
-    config->localConnectionConfig.maxChunkCount = 0; /* 0 -> unlimited */
+    config->localConnectionConfig = UA_ConnectionConfig_default;
 
     /* Certificate Verification that accepts every certificate. Can be
      * overwritten when the policy is specialized. */


plugins/ua_config_default.c
```diff
@@ -751,11 +767,7 @@ UA_ClientConfig_setDefault(UA_ClientConfig *config) {
     config->logger.context = NULL;
     config->logger.clear = UA_Log_Stdout_clear;
 
-    config->localConnectionConfig.protocolVersion = 0;
-    config->localConnectionConfig.sendBufferSize = 65535;
-    config->localConnectionConfig.recvBufferSize = 65535;
-    config->localConnectionConfig.maxMessageSize = 0; /* 0 -> unlimited */
-    config->localConnectionConfig.maxChunkCount = 0; /* 0 -> unlimited */
+    config->localConnectionConfig = UA_ConnectionConfig_default;
```
jpfr: This change is good!

plugins/ua_config_default.c
```diff
@@ -20,6 +20,22 @@
...
+# define UA_RECV_BUFFER_SIZE_DEFAULT 65535 /* 64k per chunk */
+#endif
+
+#ifndef UA_MAX_MESSAGE_SIZE_DEFAULT
+# define UA_MAX_MESSAGE_SIZE_DEFAULT 0 /*  0 -> unlimited */
+#endif
+
+#ifndef UA_MAX_CHUNK_COUNT_DEFAULT
+# define UA_MAX_CHUNK_COUNT_DEFAULT 0 /*  0 -> unlimited */
+#endif
```
jpfr: Please do not add these macros. They add needless complexity.
We should keep compile-time flags to a minimum.

If an embedded developer has little ram available, then he can always override the settings once the client has been instantiated (before starting it).
Or by writing a custom function `UA_ClientConfig_setMyEmbeddedDefault`.

cabralfortiss: Ok, I'll do that. Thanks!

jpfr:
Merged. Thanks!

PR #2493, opened 2019-02-22
Remove dependency on python-six

jpfr:
Python3 is now very common. We can expect it for the dev environments
where code is generated. Remove workarounds and the python-six dependency.


jpfr:
Made this a less invasive change.
The dependency on python-six was removed.
Python2 is still okay.

PR #2497, opened 2019-02-24
Add two more tests to the "service - call" suite

basyskom-meerkoetter:


tests/server/check_services_call.c
```diff
@@ -103,6 +103,30 @@ START_TEST(callMethodAndObjectExistsButMethodHasWrongNodeClass) {
...
+START_TEST(callMethodOnUnrelatedObject) {
+    UA_CallMethodRequest callMethodRequest;
+    UA_CallMethodRequest_init(&callMethodRequest);
+    callMethodRequest.methodId = UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_REQUESTSERVERSTATECHANGE);
+    callMethodRequest.objectId = UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER);  // not connected via hasComponent
+
+    UA_CallMethodResult result;
+    UA_CallMethodResult_init(&result);
+    result = UA_Server_call(server, &callMethodRequest);
+    ck_assert_int_eq(result.statusCode, UA_STATUSCODE_BADMETHODINVALID);
```
basyskom-meerkoetter: @jpfr do you have an idea why this assertion fails?

the method is not connected to the object via a hasComponent reference.
Still the code in ua_service_call.c:145-159 validates and doesn't return with BADMETHODINVALID.

jpfr: In that case, can you set a breakpoint on `found = true;` in line 155?
This will give the mis-classified node.

basyskom-meerkoetter: Sure, will try.
But the assumption is (so far) that the unit-test is correct and the code is ua_services_call.c has a bug?

basyskom-meerkoetter: My fault. The test-case was wrong

PR #2504, opened 2019-02-26
DatatypeGenerator: Check for valid node id in csv

Pro:
If there's a type definition in the .bsd file, but there is no corresponding
entry in the .csv, then the old code silently set the type id to NODEID_NULL.

This will then cause the server to fail on a later stage. To avoid this,
the datatype generator should check this immediately and inform the user.


jpfr:
Looks good!
Can be merged from my side.

PR #2506, opened 2019-02-27
Fix #2502

jpfr:
This adds another layer of protection instead of just removing the assert.


MSNexploder:
Just did a test run and was unable to trigger the assert anymore.
@jpfr Thanks for the fast (and more complete) fix. 👍🎉

jpfr:
You're welcome!
Thanks for reporting.

PR #1810, opened 2018-05-28
Add all attributes of VariableType nodes in the nodeset compiler

basyskom-jvoe:
This patch extends the nodeset compiler to support all possible attributes of the VariableType node class.

Using the DataType id from the xml description instead of BaseDataType causes an error when adding two nodes from Opc.Ua.NodeSet2.xml (i=17306 and i=17292 fail with "AddNodes: The value is incompatible with the datatype of the VariableType").
Both VariableType nodes use String as DataType attribute and are a subtype of i=16309. i=16309 has DataType i=862 which is a subtype of Struct.

Could there be something wrong with the type checking in ua_services_nodemanagement.c?
Page 30 of part 3 of the spec states "The VariableType NodeClass also defines a set of Attributes that describe the default or initial value of its instance Variables". Doesn't this imply that the DataType attribute can be changed in a variable instance?

PR #2519, opened 2019-03-01
Update UA-Nodeset submodule to most recent 1.04 commit 

Pro:
Note that this PR can only be merged if https://github.com/OPCFoundation/UA-Nodeset/pull/21 is fixed!


Pro:
After merging this, #1810 can also be merged.

Pro:
@jpfr note that this PR changes deps/ua-nodeset to my fork of the UA-Nodeset. For now we need to use that one, until the issue https://github.com/OPCFoundation/UA-Nodeset/pull/21 is fixed, then we can switch back to the original one.

You probably need to do a `git submodule sync` before `git submodule update`. New clones only need a `git submodule update --init --recursive`

PR #2523, opened 2019-03-02
Use constant time memcmp to verify signatures to mitigate timing attacks

paroga:
I just found this security issue while skimming the code. Maybe there are similar issues in other parts of the code too.


jpfr:
Thanks!

PR #2525, opened 2019-03-02
JSON: cleanup and avoid code duplication in decoding functions

paroga:
* Introduce macros for checking types and null values.
* Return decoding error for Guid and DateTime null values.
* Do not set the destination variable for null values, since the calling
  function is responsible for initalizing it before calling.
* Allways CHECK_TOKEN_BOUNDS before GET_TOKEN.


jpfr:
Nice improvement!

PR #2527, opened 2019-03-03
JSON: raise error when decoding Boolean null values

paroga:
The OPC UA specification 1.04 part 6 does not allow boolean values
encoded as "null". Boolean fields with null values in structs must
be omitted in the resulting JSON object, which is different to an
encoded null value.


jpfr:
Perfect PR.
Thanks!

PR #2528, opened 2019-03-03
JSON: encode UInt64 and Int64 as string

paroga:
According to the OPC UA specification 1.04 part 6 paragraph 5.4.2.3:
"Int64 and UInt64 values shall be formatted as a decimal number encoded
as a JSON string."


jpfr:
👍

PR #2529, opened 2019-03-03
JSON: support nanoseconds when encosing/decoding DateTime

paroga:
The OPC UA specification 1.04 part 6 at paragraph 5.4.2.6 does not
define any restrictions of the fraction part of ISO 8601. To support
the whole precison of DateTime we need to support at least 100ns.


jpfr:
👍

PR #2531, opened 2019-03-03
JSON: do not encode fileds of DataValue with default value

paroga:
According to the OPC UA specification 1.04 part 6 paragraph 5.4.2.18:
"If a field has a null or default value it is omitted."

### DIFF
--- a/src/ua_types_encoding_json.c
+++ b/src/ua_types_encoding_json.c
@@ -1261,51 +1261,57 @@ ENCODE_JSON(Variant) {
 
 /* DataValue */
 ENCODE_JSON(DataValue) {
-    if(!src->hasServerPicoseconds && !src->hasServerTimestamp &&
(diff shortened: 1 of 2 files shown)


jpfr:
Thanks for the contribution.

There is a difference between not having a timestamp and a timestamp at exactly 0.
But both are now encoded the same.
Anyway, the spec leaves no room for interpretation here.

paroga:
> There is a difference between not having a timestamp and a timestamp at exactly 0.

I know what you mean, but it's the same behaviour, which is expected from the binary encoding. According to 5.2.2.17 Table 16 first row: _0x10 False if the Source Picoseconds is 0._ and _0x20 False if the Server Picoseconds is 0._ IMHO the spec leaves there no room for interpretation either. So the current implementation of the binary encoding does not match the requirements of the spec.

PR #2534, opened 2019-03-04
Ignore null fields in DataValue during binary encoding

paroga:
According to the OPC UA specification 1.04 part 6 paragraph 5.2.2.17
fields with all zero content should not be encoded.

### DIFF
--- a/src/ua_types_encoding_binary.c
+++ b/src/ua_types_encoding_binary.c
@@ -1129,14 +1129,21 @@ DECODE_BINARY(Variant) {
 
 /* DataValue */
 ENCODE_BINARY(DataValue) {
+    UA_Boolean hasValue = src->hasValue && src->value.type != NULL;
+    UA_Boolean hasStatus = src->hasStatus && src->status;
+    UA_Boolean hasSourceTimestamp = src->hasSourceTimestamp && src->sourceTimestamp;
+    UA_Boolean hasSourcePicoseconds = src->hasSourcePicoseconds && src->sourcePicoseconds;
+    UA_Boolean hasServerTimestamp = src->hasServerTimestamp && src->serverTimestamp;
+    UA_Boolean hasServerPicoseconds = src->hasServerPicoseconds && src->serverPicoseconds;
+
     /* Set up the encoding mask */
     u8 encodingMask = (u8)
-        (((u8)src->hasValue) |
-         ((u8)src->hasStatus << 1) |
-         ((u8)src->hasSourceTimestamp << 2) |
-         ((u8)src->hasServerTimestamp << 3) |
-         ((u8)src->hasSourcePicoseconds << 4) |
-         ((u8)src->hasServerPicoseconds << 5));
+        (((u8)hasValue) |
+         ((u8)hasStatus << 1) |
+         ((u8)hasSourceTimestamp << 2) |
+         ((u8)hasServerTimestamp << 3) |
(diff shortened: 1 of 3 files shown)


jpfr:
No. Different from the JSON encoding, the binary encoding does not mention that values can be omitted if they are zero.

The standard only refers to the bit in the encoding mask to decide on encoding or not.

paroga:
OPC UA specification 1.04 part 6 paragraph 5.2.2.17:
```
The mask has the following bits:
0x01 False if the Value is Null.
0x02 False if the StatusCode is Good.
0x04 False if the Source Timestamp is DateTime.MinValue.
0x08 False if the Server Timestamp is DateTime.MinValue.
0x10 False if the Source Picoseconds is 0.
0x20 False if the Server Picoseconds is 0.
```

Can you point me to the sentence, which states something different?

If you have a look at the official ANSI-C stack, you will see that it's done in a way like i say:
https://github.com/OPCFoundation/UA-AnsiC-Legacy/blob/f01acfab3da583645221f9a30a9ff88af21ec1aa/Stack/stackcore/opcua_binaryencoder.c#L2133

jpfr:
You are right about the spec.
But I don't want to throw out picoseconds information if zero picoseconds were measured.

In applications that care about picoseconds (e.g. physics), there is a big difference between
    not measuring picoseconds at all
and
    having measured zero picoseconds.
But this difference is lost in the encoding.

Escalated to the OPC Foundation.
https://apps.opcfoundation.org/mantis/view.php?id=4651

paroga:
But you're doing the same "wrong" thing with the 100ns resolution too:
https://github.com/open62541/open62541/blob/0067e1f703cda19ea29edac816735b61214e927b/arch/posix/ua_clock.c#L23 only gets microsecond resolution, but is stored in a DateTime with 100ns resolution. OPC UA has no information about the exact precision of DateTime on the binary transport layer. AFAIK information about the precision must be handled on an other level.

jpfr:
The POSIX clock having only usec resolution does not mean we have to force that on everybody.
Another architecture could support better clock resolution.
Also, you might use an external sensor that returns measurement values with an independent timestamp.

In your case, when a value comes with no date set, can I assume that the measurement was taken on Jan 1st 1601, the epoch date of DateTime?

paroga:
No, you can not assume that: OPC UA defines that DateTime.MinValue and DateTime.MaxValue are special values, which DO NOT represent actual dates (independent of the actual value, which is different between XML/JSON and Binary encoding: `0001-01-01T00:00:00Z` vs. `1601-01-01 12:00AM`).

I know that we could implement a more accurate clock, but that's not my point. I just say, that you have the same problem of "pretending a high resolution" already with the DateTime value, independent of the picosecond part. You must look at the picosecond just as additional digits to the DateTime which are always present, but can be dropped on the wire to save bandwidth.

jpfr:
Part 6, 5.2.2.5:
> A date/time value is encoded as 0 if either
> 1) The value is equal to or earlier than 1601-01-01 12:00AM UTC.
> 2) The value is the earliest date that can be represented with the DevelopmentPlatform’s
encoding.

jpfr:
Also, DateTime is a signed int64. So DateTime.MinValue is not zero but the smallest (negative) int64.

jpfr:
Closing this. The discussion in the official OPCF mantis is clear.

PR #2549, opened 2019-03-06
Async monitored items

wbitesi:
Implemented client async functions for subscriptions/monitored items.
I tried to reduce the number of modifications of the original code by splitting each synchronous function in separate parts.

Longer explanation what and why it is implemented that way:

Split synchronous function (SFN) into the following parts:
a) prepare the call
b) execute the call synchronously
c) handle the call result

The asynchronous function (AFN) implements the same schema as SFN; it replace
the synchronous call by an asynchronous one.  It appends the c) part as
ClientAsyncCallback to the asynchronous call.  It also receives a user-defined
callback, that has to be executed after part c) of the AFN has run.

The difficulty lays in the memory management of parts a) and c).  In the
synchronous application, the function parameters and the local variables are
(except multithreading) safe and cannot be modified or deleted by the outside.

In the asynchronous case, part c) is executed after the AFN has returned.
Therefore, we must preserve the necessary state (parameters, local variables) on
the heap and manage the memory internally.  Since part c) is not user provided,
we have to ensure that everything is cleaned up appropriately.

I extended the CustomCallback to hold client-internal data along with a deleter
for that data.  The custom callback holds the usercall back and the ptr to its
userdata as well (which is called after the client-internal handler has run).

I updated the CustomCallback usage (read node async functions) according to the
extension.

### DIFF
--- a/src/client/ua_client_internal.h
+++ b/src/client/ua_client_internal.h
@@ -112,16 +112,19 @@ void UA_Client_AsyncService_cancel(UA_Client *client, AsyncServiceCall *ac,
 
 void UA_Client_AsyncService_removeAll(UA_Client *client, UA_StatusCode statusCode);
 
+typedef void (*CustomCallbackDataDeleter)(void *);
+
 typedef struct CustomCallback {
     LIST_ENTRY(CustomCallback)
     pointers;
     //to find the correct callback
     UA_UInt32 callbackId;
 
-    UA_ClientAsyncServiceCallback callback;
+    UA_ClientAsyncServiceCallback userCallback;
+    void *userData;
 
-    UA_AttributeId attributeId;
-    const UA_DataType *outDataType;
+    void *clientData;
+    CustomCallbackDataDeleter clientDataDeleter;
 } CustomCallback;
 
 struct UA_Client {
@@ -168,6 +171,34 @@ struct UA_Client {
     UA_DateTime lastConnectivityCheck;
     UA_Boolean pendingConnectivityCheck;
 };
+
+static UA_INLINE CustomCallback *
+CustomCallback_new(void) {
+    CustomCallback *cc = (CustomCallback *)UA_malloc(sizeof(CustomCallback));
+    if (cc)
+        memset(cc, 0, sizeof(CustomCallback));
+    return cc;
+}
+
+// removes the callback from the client and frees it and its clientData (if clientDataDeleter is set)
+static UA_INLINE void
+CustomCallback_remove(CustomCallback *cc, bool removeList) {
+    if(removeList)
+        LIST_REMOVE(cc, pointers);
+    if(cc->clientDataDeleter && cc->clientData)
+        cc->clientDataDeleter(cc->clientData);
+    UA_free(cc);
+}
+
+static UA_INLINE CustomCallback *
+UA_Client_findCustomCallback(UA_Client *client, UA_UInt32 requestId) {
+    CustomCallback *cc;
+    LIST_FOREACH(cc, &client->customCallbacks, pointers) {
+        if(cc->callbackId == requestId)
+            break;
+    }
+    return cc;
+}
 
 void
 setClientState(UA_Client *client, UA_ClientState state);
--- a/src/client/ua_client_internal.h
+++ b/src/client/ua_client_internal.h
@@ -172,26 +172,30 @@ struct UA_Client {
     UA_Boolean pendingConnectivityCheck;
 };
 
-static UA_INLINE CustomCallback *CustomCallback_new(void) {
+static UA_INLINE CustomCallback *
+CustomCallback_new(void) {
     CustomCallback *cc = (CustomCallback *)UA_malloc(sizeof(CustomCallback));
-    if (cc)
+    if(cc)
         memset(cc, 0, sizeof(CustomCallback));
     return cc;
 }
 
-// removes the callback from the client and frees it and its clientData (if clientDataDeleter is set)
-static UA_INLINE void CustomCallback_remove(CustomCallback *cc, bool removeList) {
-    if (removeList)
+// removes the callback from the client and frees it and its clientData (if
+// clientDataDeleter is set)
+static UA_INLINE void
+CustomCallback_remove(CustomCallback *cc, bool removeList) {
+    if(removeList)
         LIST_REMOVE(cc, pointers);
-    if (cc->clientDataDeleter && cc->clientData)
+    if(cc->clientDataDeleter && cc->clientData)
         cc->clientDataDeleter(cc->clientData);
     UA_free(cc);
 }
 
-static UA_INLINE CustomCallback *UA_Client_findCustomCallback(UA_Client *client, UA_UInt32 requestId) {
+static UA_INLINE CustomCallback *
+UA_Client_findCustomCallback(UA_Client *client, UA_UInt32 requestId) {
     CustomCallback *cc;
-    LIST_FOREACH (cc, &client->customCallbacks, pointers) {
-        if (cc->callbackId == requestId)
+    LIST_FOREACH(cc, &client->customCallbacks, pointers) {
+        if(cc->callbackId == requestId)
             break;
     }
     return cc;
--- a/src/client/ua_client_internal.h
+++ b/src/client/ua_client_internal.h
@@ -112,8 +112,6 @@ void UA_Client_AsyncService_cancel(UA_Client *client, AsyncServiceCall *ac,
 
 void UA_Client_AsyncService_removeAll(UA_Client *client, UA_StatusCode statusCode);
 
-typedef void (*CustomCallbackDataDeleter)(void *);
-
 typedef struct CustomCallback {
     LIST_ENTRY(CustomCallback)
     pointers;
@@ -123,8 +121,8 @@ typedef struct CustomCallback {
     UA_ClientAsyncServiceCallback userCallback;
     void *userData;
 
+    bool isAsync;
     void *clientData;
-    CustomCallbackDataDeleter clientDataDeleter;
 } CustomCallback;
 
 struct UA_Client {
@@ -171,17 +169,6 @@ struct UA_Client {
     UA_DateTime lastConnectivityCheck;
     UA_Boolean pendingConnectivityCheck;
 };
-
-// removes the callback from the client and frees it and its clientData (if
-// clientDataDeleter is set)
-static UA_INLINE void
-CustomCallback_remove(CustomCallback *cc, bool removeList) {
-    if(removeList)
-        LIST_REMOVE(cc, pointers);
-    if(cc->clientDataDeleter && cc->clientData)
-        cc->clientDataDeleter(cc->clientData);
-    UA_free(cc);
-}
 
 static UA_INLINE CustomCallback *
 UA_Client_findCustomCallback(UA_Client *client, UA_UInt32 requestId) {
--- a/src/client/ua_client_internal.h
+++ b/src/client/ua_client_internal.h
@@ -172,8 +172,7 @@ struct UA_Client {
     UA_Boolean pendingConnectivityCheck;
 };
 
-static UA_INLINE CustomCallback *
-CustomCallback_new(void) {
+static UA_INLINE CustomCallback *CustomCallback_new(void) {
     CustomCallback *cc = (CustomCallback *)UA_malloc(sizeof(CustomCallback));
     if (cc)
         memset(cc, 0, sizeof(CustomCallback));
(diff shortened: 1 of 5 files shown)


basyskom-meerkoetter:
Please consider adding tests for your new code. See tests/client/ for what is already there.

wbitesi:
I'll take a look at the unit tests and try to add similar async based ones.

I'm not sure about the cppcheck warnings. When i run the cppcheck command (in travis_linux.sh) on my system using gcc-6.3.0 i dont get the warning as in the travis build. I dont understand why it isnt portable to have pointer arithmetic on void**. I thought that the c compiler knows the size of a void pointer.

Infinity95:
Pointer arithmetic for void pointers is not legal in c or c++.
See https://stackoverflow.com/questions/3523145/pointer-arithmetic-for-void-pointer-in-c for an explanation.
Basically some compilers allow it as an extension and assume a size of one byte for void, which is why this is not portable.

wbitesi:
I have added test cases that check if subscriptions and monitored items are added/removed correctly.

The pull request should not fail (everything runs correctly), but the clang-format fails with "clang-format did not modify any files".

Pro:
If you do a new push and/or rebase onto master, the clang-format will be changed to allowed fail. It's currently under testing and should not block new PRs.
I will have a look at the false negative clang-format in the meantime.

wbitesi:
I wanted to ask if there is anything open for me to do with the pull request or if it is ready for merging and will be done if someone has time to do it?

src/client/ua_client_internal.h
```diff
@@ -111,16 +111,19 @@ void UA_Client_AsyncService_cancel(UA_Client *client, AsyncServiceCall *ac,
 
 void UA_Client_AsyncService_removeAll(UA_Client *client, UA_StatusCode statusCode);
 
+typedef void (*CustomCallbackDataDeleter)(void *);
```
jpfr: Why do we need this guy?
If there is a timeout, we can just set a statuscode in the responseheader and call the "normal" callback.

src/client/ua_client_internal.h
```diff
@@ -168,6 +171,31 @@ struct UA_Client {
     UA_Boolean pendingConnectivityCheck;
 };
 
+static UA_INLINE CustomCallback *CustomCallback_new(void) {
```
jpfr: You can use UA_calloc to get zeroed memory.
So we probably don't need this method.

src/client/ua_client_subscriptions.c
```diff
@@ -29,39 +70,62 @@ UA_Client_Subscriptions_create(UA_Client *client,
...
-    /* Allocate the internal representation */
-    UA_Client_Subscription *newSub = (UA_Client_Subscription*)
-        UA_malloc(sizeof(UA_Client_Subscription));
-    if(!newSub) {
+
+    CustomCallback *cc = CustomCallback_new();
+    if (!cc) {
         response.responseHeader.serviceResult = UA_STATUSCODE_BADOUTOFMEMORY;
-        return response;
+        goto cleanup;
```
jpfr: You can immediately return here.
There is no callback to clean up.

src/client/ua_client_subscriptions.c
```diff
@@ -29,39 +70,62 @@ UA_Client_Subscriptions_create(UA_Client *client,
...
-    newSub->statusChangeCallback = statusChangeCallback;
-    newSub->deleteCallback = deleteCallback;
-    newSub->publishingInterval = response.revisedPublishingInterval;
-    newSub->maxKeepAliveCount = response.revisedMaxKeepAliveCount;
-    LIST_INIT(&newSub->monitoredItems);
-    LIST_INSERT_HEAD(&client->subscriptions, newSub, listEntry);
+    __Subscriptions_create_handler(client, cc, 0, &response);
 
+    return response;
+cleanup:
```
jpfr: The goto is then used in only one place.
Move up.

src/client/ua_client_subscriptions.c
```diff
@@ -29,39 +70,62 @@ UA_Client_Subscriptions_create(UA_Client *client,
...
 
+UA_StatusCode UA_EXPORT UA_Client_Subscriptions_create_async(
+    UA_Client *client, const UA_CreateSubscriptionRequest request, void *subscriptionContext,
+    UA_Client_StatusChangeNotificationCallback statusChangeCallback,
+    UA_Client_DeleteSubscriptionCallback deleteCallback, UA_ClientAsyncServiceCallback createCallback, void *userdata,
+    UA_UInt32 *requestId) {
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    CustomCallback *cc = CustomCallback_new();
+    if (!cc) {
+        retval = UA_STATUSCODE_BADOUTOFMEMORY;
```
jpfr: There is no callback to remove.
Directly return.

src/client/ua_client_subscriptions.c
```diff
@@ -29,39 +70,62 @@ UA_Client_Subscriptions_create(UA_Client *client,
...
+    if (retval != UA_STATUSCODE_GOOD) {
+        goto cleanup;
+    }
+
+    /* Send the request as asynchronous service call */
+    return __UA_Client_AsyncService(client, &request, &UA_TYPES[UA_TYPES_CREATESUBSCRIPTIONREQUEST],
+                                    __Subscriptions_create_handler, &UA_TYPES[UA_TYPES_CREATESUBSCRIPTIONRESPONSE], cc,
+                                    requestId);
+
+cleanup:
```
jpfr: The goto is used in only one place.
Move up.

src/client/ua_client_subscriptions.c
```diff
@@ -111,61 +210,134 @@ UA_Client_Subscription_deleteInternal(UA_Client *client, UA_Client_Subscription
...
-    for(size_t i = 0; i < request.subscriptionIdsSize; i++) {
-        if(response.results[i] != UA_STATUSCODE_GOOD &&
-           response.results[i] != UA_STATUSCODE_BADSUBSCRIPTIONIDINVALID) {
+    for (size_t i = 0; i < request->subscriptionIdsSize; i++) {
+        if (response->results[i] != UA_STATUSCODE_GOOD &&
+            response->results[i] != UA_STATUSCODE_BADSUBSCRIPTIONIDINVALID) {
             /* Something was wrong, reinsert the subscription in the list */
             if (subs[i])
                 LIST_INSERT_HEAD(&client->subscriptions, subs[i], listEntry);
+            subs[i] = NULL;
```
jpfr: Why subs[i] = NULL?

src/client/ua_client_subscriptions.c
```diff
@@ -111,61 +210,134 @@ UA_Client_Subscription_deleteInternal(UA_Client *client, UA_Client_Subscription
...
-                        request.subscriptionIds[i]);
+            UA_LOG_INFO(&client->config.logger, UA_LOGCATEGORY_CLIENT, "No internal representation of subscription %u",
+                        delData->request->subscriptionIds[i]);
             continue;
         } else {
             LIST_INSERT_HEAD(&client->subscriptions, subs[i], listEntry);
         }
 
         UA_Client_Subscription_deleteInternal(client, subs[i]);
+        subs[i] = NULL;
```
jpfr: subs[i] = NULL?

src/client/ua_client_subscriptions.c
```diff
@@ -111,61 +210,134 @@ UA_Client_Subscription_deleteInternal(UA_Client *client, UA_Client_Subscription
...
                 LIST_INSERT_HEAD(&client->subscriptions, subs[i], listEntry);
+            subs[i] = NULL;
             continue;
         }
 
         if(!subs[i]) {
-            UA_LOG_INFO(&client->config.logger, UA_LOGCATEGORY_CLIENT,
-                        "No internal representation of subscription %u",
-                        request.subscriptionIds[i]);
+            UA_LOG_INFO(&client->config.logger, UA_LOGCATEGORY_CLIENT, "No internal representation of subscription %u",
```
jpfr: Keep lines short.
Avoid cosmetic changes in a feature-commit.

src/client/ua_client_subscriptions.c
```diff
@@ -111,61 +210,134 @@ UA_Client_Subscription_deleteInternal(UA_Client *client, UA_Client_Subscription
...
         if (subs[i]) {
             LIST_INSERT_HEAD(&client->subscriptions, subs[i], listEntry);
         }
     }
+    if (cc->userCallback)
+        cc->userCallback(client, cc->userData, requestId, response);
+    CustomCallback_remove(cc, false);
+}
+
+UA_StatusCode UA_EXPORT UA_Client_Subscriptions_delete_async(UA_Client *client,
```
jpfr: UA_EXPORT is not needed in .c file.
Move method name to a new line.

src/client/ua_client_subscriptions.c
```diff
@@ -21,6 +21,47 @@
...
+    newSub->publishingInterval = response->revisedPublishingInterval;
+    newSub->maxKeepAliveCount = response->revisedMaxKeepAliveCount;
+
+    LIST_INIT(&newSub->monitoredItems);
+    LIST_INSERT_HEAD(&client->subscriptions, newSub, listEntry);
+
+cleanup:
+    if (cc->userCallback)
+        cc->userCallback(client, cc->userData, requestId, response);
+    CustomCallback_remove(cc, false);
```
jpfr: The handler should not have to remove the callback.
This can be done "outside" in the generic handling of async callbacks.

Then, we can also stack-allocate the CustomCallback when call the handler in the synchronous case.

src/client/ua_client_subscriptions.c
```diff
@@ -283,21 +468,148 @@ __UA_Client_MonitoredItems_create(UA_Client *client,
...
+    void *userdata, UA_UInt32 *requestId) {
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    CustomCallback *cc = CustomCallback_new();
+    if (!cc) {
+        retval = UA_STATUSCODE_BADOUTOFMEMORY;
+        goto cleanup;
+    }
+    cc->userCallback = createCallback;
+    cc->userData = userdata;
+    MonitoredItems_CreateData *data = (MonitoredItems_CreateData *)UA_malloc(sizeof(MonitoredItems_CreateData));
```
jpfr: use UA_calloc.

src/client/ua_client_internal.h
wbitesi: I'll remove this method and replace it with UA_callocs

src/client/ua_client_internal.h
wbitesi: My motivation was to have the delete function in the custom callback to be able to free the memory if the handler isnt called. It looks like that the handler is always called (remove UA_Client_AsyncService_cancel) so i can remove it from the CustomCallback.

src/client/ua_client_subscriptions.c
wbitesi: In the cleanup part, I add all subs[i] that are not null. the cleanup part is always called independent if there was an error or not. This just ensures that previously added subscriptions are not added again.

I could jump to line 330 to only free the memory and skipping the for loop at the beginning? I'm open to suggestions

src/client/ua_client_subscriptions.c
wbitesi: same as above

src/client/ua_client_subscriptions.c
wbitesi: That happened through old clang-format settings. I'm using the new clang-format config and now it doesnt make such changes. Will be fixed later

src/client/ua_client_subscriptions.c
wbitesi: (same thing with clang format)
The existing functions did use UA_EXPORT too, should i remove UA_EXPORT them from my added new functions or from all in ua_client_subscriptions.c?

src/client/ua_client_subscriptions.c
wbitesi: I'm not happy with my approach to delete it directly in the handler either (I previously tried to stack allocate on synchronous functions too). Didnt get it to work properly.

I could add the routine to automatically free a stored customcallback using the requestId in processAsyncMessage (ua_client.c:241). However this would require for every async callback to check (find, delete, remove from list) if there is a customcallback stored, even if there is none saved.

What do you think?

src/client/ua_client_subscriptions.c
wbitesi: done

src/client/ua_client_internal.h
wbitesi: done

src/client/ua_client_internal.h
wbitesi: I have replaced the deleter with a flag to distinguish between "async customcallbacks) on the heap and (sync customcallbacks) on the stack.

src/client/ua_client_subscriptions.c
wbitesi: see first comment (CustomCallbackDataDeleter)

src/client/ua_client_subscriptions.c
jpfr: UA_EXPORT is only for headers. Not for .c files.

wbitesi: I have removed all UA_EXPORT occurrences in the ua_client_subscriptions.c file

src/client/ua_client_subscriptions.c
wbitesi: Fixed

wbitesi:
I have updated the pull request according to the comments. I hope the points are addressed well.

Everything tests/builds fine except the clang-analyzer. I don't know why there should be a potential memory leak (which is not shown in the tests). Maybe someone has an idea what i can do to fix this?

Infinity95:
At a glance it looks like you aren't calling the cleanup part if your function returns normally. Maybe you could check if that's the cause.

wbitesi:
I have found the thing that clang-analyzer pointed to a memory leak and fixed it with an assertion (with #ifdef __clang_analyzer__). Now everything builds fine.

I have not marked the review as resolved so jpfr can take a look if he's satisfied with the changes, although I think they are now fixed.

wbitesi:
I have left the review open for jfp to check if the changes made after his comments are alright.

I have the async functions working in production since April and it works. Is the PR ready for merging or is something still to do?

jpfr:
Nice!
This turned out great.
Thanks for the push @wbitesi!

PR #2558, opened 2019-03-08
WIP: Enable Static Code Analysis

bbara:
Hey,

part of previous PR #1732.
Enables static code analysis during the build process.
Fixes are required that the build runs without warnings/errors with enabled
code analysis and previous clang-format.
Additional CI analysis step looks if committed code is still clang-formatable by
running the build process with reduced code analysis.


Pro:
Thanks for your work! Sorry, I'm a bit confused by your recent PRs...

There's #2560 which runs git-clang-format on the diff.

Here you are adding more tools:
- clang-tidy
- clang-format (again, but as a cmake target)
- cppformat (which is already executed)
- lint

One general thing:
The additional stuff and target definition should be in another CMakeLists.txt, e.g. inside the tools folder. Or you make a separate file in the `cmake\linting.cmake` folder.

Can you clarify the following points:
1. cppcheck is already executed by travis here: https://github.com/open62541/open62541/pull/2558/files#diff-3056cef31c2701d267a0058c8fa125c6R138  Why did you add it again?
2. Can the clang-tidy configuration settings be moved to a config file. Passing them directly in CMake makes the command and the list of settings very crowded. You now have three places where different clang-tidy options are used:
  - https://github.com/open62541/open62541/pull/2558/files#diff-af3b638bc2a3e6c650974192a53c7291R368
  - https://github.com/open62541/open62541/pull/2558/files#diff-af3b638bc2a3e6c650974192a53c7291R377
  - https://github.com/open62541/open62541/pull/2558/files#diff-af3b638bc2a3e6c650974192a53c7291R426
3. IMHO it's better to define two targets, instead of introducing a new CMake variable. E.g. `clang-tidy-reduced` and `clang-tidy-full` (instead of the `lint` target)
4. Why do we need an additional call to clang-format (https://github.com/open62541/open62541/pull/2558/files#diff-3056cef31c2701d267a0058c8fa125c6R134), if we already have it in #2560

bbara:
General Thing:
Yep, can provide that. 
On a previous version of #1732, which is still found in my fork [1], 
I already divided several things into helper cmake files.
But due to some open discussions and the quite long time,
it was quite time-consuming to keep it up to date 
(nearly every change of the CMake resulted in a merge conflict).
Therefore I kept everything in the main CMake but within one section,
so that it can easily be moved to another file if wanted.
So that's the reason why everything is still in the main CMake file.

Now, the other clarifications:
1.)
In fact, it's the same from a CI point of view.
It is just more convenient if the static code analyze process happens during the build process
on a local machine (direct feedback; integrated in IDEs).

2.)
Yep, similar to clang-format, it is possible to store them in a .clang-tidy file.
Can provide that too.

3.) 
You can do that, but the idea of the integration into CMake is to have them running
in parallel to your build. So e.g. if you use clang, which uses the LLVM-backend exactly
like clang-tidy, the conversion must be done just one time -> it is not a separate step anymore.

4.)
After #2560 is merged, this might not be required anymore, as described in [2].
It just guarantees that the whole code base is buildable after the clang-format changes 
(currently around 17k affected LOC).
Even though git clang-format looks at "local" problems, "global" problems (like [3])
will likely not be covered.
I understand the problem with merge compatibility with forks and PRs,
but I am not sure how effective the long-term asymptotic approach is.
Might be an idea for #2560, to display the changed LOCs after a complete clang-format.
These should in fact go asymptotically against zero.

[1] https://github.com/bbara/open62541/tree/1680_2
[2] https://github.com/open62541/open62541/pull/1732#discussion_r263763691
[3] https://github.com/open62541/open62541/pull/1732#discussion_r263759993

Pro:
> Now, the other clarifications:
> 1.)
> In fact, it's the same from a CI point of view.
> It is just more convenient if the static code analyze process happens during the build process
> on a local machine (direct feedback; integrated in IDEs).

Yes, agree! Then let's keep the new cmake target. In travis we should then simply call the cmake target here: https://github.com/open62541/open62541/pull/2558/files#diff-3056cef31c2701d267a0058c8fa125c6R138

```
make cppcheck 2> cppcheck.txt
```

And then also adapt the entry in the cmake that we have the same arguments as there already are in travis/add the same config to your cppcheck-suppressions.txt

> 
> 2.)
> Yep, similar to clang-format, it is possible to store them in a .clang-tidy file.
> Can provide that too.

Yes please. Keeping the options in a separate file makes the CMakeLists.txt cleaner

> 3.)
> You can do that, but the idea of the integration into CMake is to have them running
> in parallel to your build. So e.g. if you use clang, which uses the LLVM-backend exactly
> like clang-tidy, the conversion must be done just one time -> it is not a separate step anymore.

Ok, let's rename the `lint` target to `clang-tidy`. Then it's more obvious.

> 4.)
> After #2560 is merged, this might not be required anymore, as described in [2].
> It just guarantees that the whole code base is buildable after the clang-format changes
> (currently around 17k affected LOC).
> Even though git clang-format looks at "local" problems, "global" problems (like [3])
> will likely not be covered.
> I understand the problem with merge compatibility with forks and PRs,
> but I am not sure how effective the long-term asymptotic approach is.
> Might be an idea for #2560, to display the changed LOCs after a complete clang-format.
> These should in fact go asymptotically against zero.
> 
> [1] [bbara/open62541@`1680_2`](https://github.com/bbara/open62541/tree/1680_2)
> [2] [#1732 (comment)](https://github.com/open62541/open62541/pull/1732#discussion_r263763691)
> [3] [#1732 (comment)](https://github.com/open62541/open62541/pull/1732#discussion_r263759993)

Calling clang-format for the whole code base in every PR is not a viable option. That makes contributions very impracticable. So let's remove this additional call during CI and just use it on the diff, as it is in #2560 


One general thing about this PR:
Try to separate your changes into separate commits, at least for every script and the code changes. That makes it easier to review.

bbara:
Will separate asap.

For the diff-lint idea:
We require some kind of parser for the git diff, like done in [1].
As soon as we have the changed files and respective lines,
we can filter the output of clang-tidy, cppcheck and cpplint for exactly these files+lines.
I think that is the easiest way...

[1] https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/git-clang-format

Pro:
@bbara I did some more cleanup and fixes. Let's first merge #2566 and then continue with this PR.

As a first step I would only add the cpplint and clang-tidy to the CI check.
clang-format should not be checked for now.

Pro:
I will merge this in the current state since this PR only adds some linting targets and does not change any of the code.

In a future PR we can gradually use these targets and linting.

Pro:
I will merge this in the current state since this PR only adds some linting targets and does not change any of the code.

In a future PR we can gradually use these targets and linting.

PR #2559, opened 2019-03-08
WIP: Integrate clang-format-ci

bbara:
Hey,

as discussed in #1732, this is a first try to integrate the clang-format-ci.

Edit: 
clang-format-ci [1] currently requires swift and just works on PRs coming from the same project.

[1] https://github.com/material-foundation/clang-format-ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -309,8 +309,13 @@ script: |
  }
  trap errorTrap 0
 
- if [ ${TRAVIS_OS_NAME} == "linux" ]; then sh ./tools/travis/travis_linux_script.sh; fi
- if [ ${TRAVIS_OS_NAME} == "osx" ]; then sh ./tools/travis/travis_osx_script.sh; fi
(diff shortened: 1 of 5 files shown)


bbara:
While this is a good idea, it is "difficult" to integrate with the current design of GitHub API tokens and TravisCI.
Since writing comments require a token but secure tokens of Travis can not be used on PRs coming
from forks, this is currently not practicable yet.
A workaround that comes to my mind would be some kind of additional server that stores
a token of a "machine user" locally and that provides an interface that can be triggered from Travis.
Maybe there is some easier approach?

Pro:
Hm, yes, you are right. The access token is not available for PRs, so we would only have the suggestions for internal PRs.

The cleanest approach is to create a github app (https://developer.github.com/apps/), which is able to run clang-format and then posts suggestions. Since that's out of scope for our project, we will have to wait until someone else provides it.

Let's close this PR in favour of the current solution in #2560

PR #2560, opened 2019-03-08
Run clang-format check for PR diff

bbara:
Hey,

since #2559 focuses on a more convenient way by adding comment support,
this is a pre-stage and just breaks the CI if the code format does not align with
the specified clang-format style.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -182,6 +182,22 @@ matrix:
         - ANALYZE=true
         - PYTHON=python3
 #
+# clang-format code analysis
+    - os: linux
+      compiler: clang-6.0
+      addons:
+        apt:
+          sources:
+            - llvm-toolchain-trusty-6.0
+            - ubuntu-toolchain-r-test
+          packages:
+            - clang-6.0
+            - clang-tidy-6.0
+            - clang-format-6.0
+      env:
+        - CLANG_FORMAT=true
+        - PYTHON=python3
+#
 # cpplint checking
     #- os: linux
     #  compiler: gcc
--- a/tools/travis/travis_linux_after_success.sh
+++ b/tools/travis/travis_linux_after_success.sh
@@ -2,7 +2,7 @@
 set -ev
 
 # We only want to push the docs once, so we just take the travis run where MINGW=true which is only enabled once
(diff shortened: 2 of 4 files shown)


Pro:
Thanks for the PR!

I modified it to create a new entry in the build matrix, which solely runs clang-format. This makes it easier to spot an issue.

As clarification for others:
This PR adds clang-format run to every pull request and checks that the changed lines conform to the style defined in our main .clang_format file (https://github.com/open62541/open62541/blob/master/.clang-format)

The suggested changes are then written to the build output in travis

Pro:
I just had a look at the output of clang-format, hmmm, that is not really helpful and it makes it really hard to review pull requests.
https://travis-ci.org/open62541/open62541/jobs/503799540#L1900

How do I know from this line that the return has a wrong indentation:
https://github.com/open62541/open62541/pull/2559/commits/aa93e7ff1be1bbdc2c097f87cd79c147cc4d692d#diff-6ad32046c79b2c74ce2b938c870442a9R73

I just added a `echo -en` maybe that gives a multiline output, but still, not really perfect...

Are there other possible outputs which explain the issue better, e.g., "Wrong indentation" or so?

Or should we only use clang-tidy instead of clang-format?

bbara:
clang-tidy would be an option, but we have to bring the git diff output
somehow in a clang-tidy understandable format (JSON string as explained in [1]; -line-filter),
since it usually is also file-based.

[1] https://clang.llvm.org/extra/clang-tidy/

bbara:
Don't know if this works on Travis.
Tried it locally, it's pretty ugly for now and additionally the format violations are not cached 
by clang-tidy in general.
cpplint has no possibility to filter for lines, so we might have to wrap the cpplint output and 
perform some filtering if we want to use it instead (similar to the Proof-of-Concept of clang-tidy).

tools/git-clang-tidy
```diff
@@ -0,0 +1,597 @@
+#!/usr/bin/env python
```
Pro: Do we really need our own copy of this file?
This means additional maintenance effort if there is e.g. a newer version with fixes...

bbara: I actually modified it to run clang-tidy instead of clang-format.
Was just a try if this works better.

Pro:
I updated your PR to only include clang-format and to optimize the output.

It will look like this:
https://travis-ci.org/Pro/open62541/jobs/504594785#L708

Now we will get at least an error in CI if someone does not adhere to the clang format.

@jpfr FYI

Issue #2374, opened 2019-01-17
Redefine the max number of subscriptions/monitored item server config options


jonasgreen88:
## Description
The server configuration does allow the user to configure the maximum number of subscriptions per session and the maximum number of monitored items per subscription to support. I have always been thinking that OPC UA specified that all sessions shall support the same number of subscriptions and all subscriptions shall support the same number of monitored items. But I think I'm wrong after I have read the conformance units of the [StandardDataChangeSubscription2017 facet](http://opcfoundation.org/UA-Profile/Server/StandardDataChangeSubscription2017).

The following conformance unit states otherwise I think:

**Subscription Minimum 02**
Support at least 2 Subscriptions per Session.
This number has to be supported for at least half of the minimum required sessions.

**Monitor Items 100**
Support at least 100 MonitoredItems per Subscription. This number has to be supported for at least half of the required Subscriptions for half of the required Sessions.

The conformance units above states the number of subscriptions per session and the number of monitored items per subscription a server must support to conform to the StandardDataChangeSubscription2017 facet. But they also state that not all of the sessions and the subscriptions must support this amount of resources.

Imaging a device that in total can support maximum 100 monitored items due to memory constraints. With the current configuration options of open62541 it can't support the StandardDataChangeSubscription2017 as it must support 2 subscriptions as well. Open62541 only allows the user to configure 2 subscriptions and 100 monitored items per subscription. That sums up to in total 200 monitored items.

I would like to suggest that we redefine the configurations options `maxSubscriptionsPerSession` and `maxMonitoredItemsPerSubscription` to define the total number of each resource the server can handle. This change would make it possible to only support 100 monitored items in total and allow a single subscription to allocate all of those. Even if there won't be any monitored item resources left for the second subscription it is enough to pass the conformance unit.

What do you think about this?

jonasgreen88:
Merged

PR #2561, opened 2019-03-11
[Review] Making the resource limitations of subscriptions global

jonasgreen88:
By making the resource limitations for subscriptions and monitored
items global in the server, the sessions can share the resources
of the server instead of splitting the resources evenly over the
supported sessions and subscriptions.
This change allow servers to support profile requirements with less
resources. E.g. the Embedded UA server profile sets minimum
requirements on number of subscriptions for a session and number of
monitored items for a subscription. But the profile does not require
all sessions or all subscriptions to meet these requirements
concurrently.

### DIFF
--- a/include/open62541/server_config.h
+++ b/include/open62541/server_config.h
@@ -151,6 +151,7 @@ struct UA_ServerConfig {
 
     /* Limits for Subscriptions */
     UA_UInt32 maxSubscriptions;
+    UA_UInt32 maxSubscriptionsPerSession;
     UA_DurationRange publishingIntervalLimits; /* in ms (must not be less than 5) */
     UA_UInt32Range lifeTimeCountLimits;
     UA_UInt32Range keepAliveCountLimits;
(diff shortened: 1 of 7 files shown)


jonasgreen88:
This PR has target to solve #2374. The resource sharing of subscriptions and monitored items is solved. How to handle publish requests resources has been more difficult to figure out. The Embedded UA profile has a requirement of minimum number of publish requests that violates a requirement of the OPC UA specification part 4. I have posted a question on the OPC forum about this without any response so far: https://opcfoundation.org/forum/opc-ua-standard/minimum-standard-datachange-subscription-resource-config/.

jpfr:
Very nice feature!
Can we keep maxSubscriptionsPerSession?
So that no one session can gobble up all resources for the other sessions.
At least to make it configurable.

jonasgreen88:
> Very nice feature!
> Can we keep maxSubscriptionsPerSession?
> So that no one session can gobble up all resources for the other sessions.
> At least to make it configurable.

Good idea. I also kept the possibility to limit monitoredItems per subscriptions. If a user only wants to use the global limitations it will be possible to set the subscriptions per session and monitoredItems per subscription limits to 0.

jpfr:
Looks good!
Thanks for shepherding.

PR #2569, opened 2019-03-12
[FIX] Check timestamp values in request header

opcua-tsn-team-kalycito:
- The server application should check the timestamp value in the
   header according to CTT test cases in the version 1.03.341.385
 - Timestamp is checked in the request headers of following APIs
   Service_BrowseNext(), Service_TranslateBrowsePathsToNodeIds(),
   Service_RegisterNodes(), Service_UnregisterNodes()
 - This fix will resolve the following test cases
 - ViewTranslateBrowsePath -> Err-024.js
 - ViewRegisterNodes -> Err-009.js, Err-014.js
 - View Minimum Continuation Point 01 -> Err-010.js

Signed-off-by: Jayanth Velusamy <jayanth.v@kalycito.com>

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -12,6 +12,7 @@
  *    Copyright 2016 (c) TorbenD
  *    Copyright 2017 (c) frax2222
  *    Copyright 2017 (c) Mark Giraud, Fraunhofer IOSB
+ *    Copyright 2019 (c) Kalycito Infotech Private Limited
  */
 
 #include "ua_server_internal.h"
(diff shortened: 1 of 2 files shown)


opcua-tsn-team-kalycito:
The updated CTT configuration and results can be found [here. ](https://github.com/open62541/open62541-ctt/pull/1)

jpfr:
Dear @opcua-tsn-team-kalycito team,

instead of adding the test to every single service, can you add it before the dispatch to the individual services? Note that every response structure begins with the ResponseHeader.
So you can cast.

https://github.com/open62541/open62541/blob/269d1ffbdc589416ae83b80bb2b4ceee9f75788d/src/server/ua_server_binary.c#L447

Also, please ensure that the unit tests continue to work.
Are the timestamps added to every request in the client?

opcua-tsn-team-kalycito:
Hi @jpfr 

Thanks for the comments. We have moved the timestamp check before dispatching to the individual services.  

> Also, please ensure that the unit tests continue to work.
> Are the timestamps added to every request in the client?

No, Timestamps are not added to every request in client, so the unit test are failing with BadInvalidTimestamp status code.

jpfr:
Thanks, the change is more elegant than I had hoped for.

> No, Timestamps are not added to every request in client, so the unit test are failing with BadInvalidTimestamp status code.

We cannot merge the PR if there are failing unit tests.

This could be caused by our "deterministic clock" to reproduce timings issues in the unit tests.
Can we simply initialize the unit-test-clock at a non-zero timestamp?

https://github.com/open62541/open62541/blob/0ca3d8a5794314d866091466c54e2473e1ea9025/tests/testing-plugins/testing_clock.c#L8

jpfr:
Great! Thanks for the contribution. 👍

PR #2572, opened 2019-03-14
Add Basic256 SecurityPolicy

jpfr:


plugins/securityPolicies/ua_securitypolicies.h
```diff
@@ -34,6 +35,12 @@ UA_SecurityPolicy_Basic256Sha256(UA_SecurityPolicy *policy,
                                  const UA_ByteString localPrivateKey,
                                  const UA_Logger *logger);
 
+/* Internal definitions for reuse between policies */
+UA_StatusCode
+generateKey_sha1p(mbedtls_md_context_t *sha1MdContext,
```
Infinity95: We don't need to expose the mbedtls specific context here.
It suffices to allocate an md_context inside the generate key function and then deallocate it again.
That way the generateKey function can be generic and agnostic from the library that is used to implement it.

plugins/securityPolicies/ua_securitypolicy_basic256.c
```diff
@@ -0,0 +1,974 @@
...
+    return UA_STATUSCODE_GOOD;
+}
+
+/*******************/
+/* SymmetricModule */
+/*******************/
+
+static void
+md_hmac_Basic256(mbedtls_md_context_t *context, const UA_ByteString *key,
+                 const UA_ByteString *in, unsigned char out[32]) {
```
Infinity95: The out array should be declared as out[20] or rather out[UA_SHA1_LENGTH]

plugins/securityPolicies/ua_securitypolicies.h
jpfr: Moved to an internal "common" header.

plugins/securityPolicies/ua_securitypolicy_basic256.c
jpfr: The same method now works for sha1 and sha256.
This is all captured by the md_context.
Made `out` into a char ptr to be generic.

Issue #2565, opened 2019-03-12
Add possibility to disable the DataChange Subscription retransmission queue


jonasgreen88:
## Description
The DataChange Subscription server requirement "Subscription Publish Min 02" of Part 7 (1.04) has the following requirement: "Support of a NotificationMessage retransmission queue is not required; if not available the Republish Service returns Bad_MessageNotAvailable.".

Currently, there  is no support to disable the retransmission queue. If configuring the queue length to 0 it means unlimited queue size. For resource constrained devices this could be a good option to save some memory and reduce the footprint of the OPC UA stack.

To keep the queue size configuration option consistent with other size options I think it would be good to let 0 be unlimited queue size. Instead I think a boolean config option can be added to server config to let the user configure if the server should support the retransmission queue.

https://github.com/open62541/open62541/blob/7722d7d2345c23b5eecb1d446c4b980617338c72/include/ua_server_config.h#L158

## Background Information / Reproduction Steps
Make it possible for resource constrained devices to support a limited feature set required by profiles.

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): 7722d7d 
 - [x] Other OPC UA SDKs used (client or server): N/A
 - [x] Operating system: N/A
 - [x] Logs (with `UA_LOGLEVEL` set as low as necessary) attached: N/A
 - [x] Wireshark network dump attached: N/A
 - [x] Self-contained code example attached:
 - [ ] Critical issue

jpfr:
The feature was done by @jonasgreen88 and is available on master.

PR #2573, opened 2019-03-14
Added possibility to disable retransmission queue of sessions

jonasgreen88:
The Micro Embedded UA profile allow devices to not implement the
retransmission queue of DataChange Subscriptions. Previously there 
has been no possibility to disable this queue. Disabling the queue
can save some memory on a resource constrained device.

### DIFF
--- a/include/ua_server_config.h
+++ b/include/ua_server_config.h
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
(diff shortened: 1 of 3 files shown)


jonasgreen88:
Solves #2565.

When I disable the retransmission queue I get the following error in the log of UaExpert when subscribing to a parameter: "List of available sequence numbers is empty, at least current sequence number has to be available".

I found out the the KEPServerEX from Kepware has an option to disable the retransmission queue as well and I got the same error in UaExpert. I have posted a question on the OPC UA forum to find out if this error from UaExpert is valid: https://opcfoundation.org/forum/opc-ua-standard/shall-the-availablesequence-numbers-field-of-publish-responses-be-empty-when-retransmission-queue-is-not-implemented/.

The specification (Part 4) doesn't really specify this case explicitly.

jpfr:
Great feature for resource-constrained servers.
Thanks @jonasgreen88 !

PR #2580, opened 2019-03-18
Always include security policies in the amalgamated source file

basyskom-jvoe:
For users of the amalgamated distribution, it would be a nice feature to be able to enable and disable security support at build time.


jpfr:
Thanks!
Support for Basic256 was added in the meantime --> PR does not merge.

basyskom-jvoe:
Thanks for the merge!

PR #2592, opened 2019-03-20
Clean up implementation specific behavior and some other code smells

Infinity95:


jpfr:
Woah. Did you use a linter tool to find all those?

Infinity95:
My IDE found some of those. The rest are from clang-tidy, which is also integrated into the IDE. :+1: 
I can then step through all found issues by pressing f2. Very neat feature.

Pro:
Fyi we now also have custom make targets to run clang-tidy and clang-format:
https://github.com/open62541/open62541/blob/master/tools/cmake/linting_target.cmake#L3

At some point we may be able to just enable this clang-tidy build in CI and check every commit.

Infinity95:
This can be merged now, if you want.

PR #2597, opened 2019-03-22
Unify memory allocation calls

cabralfortiss:
I found some malloc and free functions that should be UA_malloc and UA_free to keep the memory allocation functions unified throught the code

An issue though, is the pressence of malloc in deps/base64.c This I cannot change because, from what I understood, it works kind of as an independent file. I'm not sure the dependencies of other files and projects on this file.

This mismatch in base64.c might cause some problems when freeing the memory allocated there

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -796,13 +796,13 @@ UA_ClientConfig_setDefault(UA_ClientConfig *config) {
         return UA_STATUSCODE_BADINTERNALERROR;
     }
 
-    config->securityPolicies = (UA_SecurityPolicy*)malloc(sizeof(UA_SecurityPolicy));
+    config->securityPolicies = (UA_SecurityPolicy*)UA_malloc(sizeof(UA_SecurityPolicy));
     if(!config->securityPolicies)
         return UA_STATUSCODE_BADOUTOFMEMORY;
     UA_StatusCode retval = UA_SecurityPolicy_None(config->securityPolicies, NULL,
                                                   UA_BYTESTRING_NULL, &config->logger);
     if(retval != UA_STATUSCODE_GOOD) {
-        free(config->securityPolicies);
+        UA_free(config->securityPolicies);
         config->securityPolicies = NULL;
         return retval;
     }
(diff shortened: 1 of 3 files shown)


basyskom-meerkoetter:
If you don't want to patch the 3rd party you could use the preprocessor to do the replacement

cabralfortiss:
Is not that I don't want to change it, but I don't know how exactly to change, since I don't know how the change will affect other parts.

Using preprocessor I see as a hack. For a system wich does not use "normal" malloc, the user will have to add the define again somewhere when compiling.

Pro:
We aim for a unified usage of `UA_malloc` and friends. Therefore every file we include in the stack should use this macro definition.

Can you try to use a preprocessor define for the `deps/base64.c`.

I.e., in the CMake something like:

```
set_source_files_properties(deps/base64.c PROPERTIES COMPILE_FLAGS 
-Dmalloc=UA_malloc
-Dfree=UA_free
-Dcalloc=UA_calloc
-Drealloc=UA_realloc
)
```

But then I guess it will complain that it does not know about `UA_malloc` in that context. Hmm

jpfr:
Initially the `UA_malloc` macros were internal API.
This is why the plugins could not access it.

Now the arch definitions are kinda public.
At least, there is no explicit documentation about ti.

@Pro 
Do we now have three categories of API?
Internal, public and plugin-visible?

Pro:
> @Pro
> Do we now have three categories of API?
> Internal, public and plugin-visible?

No, it's still internal and public. Before we had the arch folder, one could change the config.h file to redefine macros. Now this is switched to the arch folder (which are not plugins, but a basic implementation). The idea here is that one needs to change the `UA_socket` macro for example. And of course here one can also replace the `UA_malloc` with an architecture specific memory allocation. IMHO this is a pre-configuration independent of plugins.

Pro:
I would say if someone needs to use a specific memory allocation, e.g. for freeRTOS, the corresponding architecture file should include additional settings like this one:

```
set_source_files_properties(${PROJECT_SOURCE_DIR}/deps/base64.c PROPERTIES COMPILE_FLAGS "-Dmalloc=malloc_freertos -Dfree=free_freertos -Dcalloc=calloc_freertos -Drealloc=realloc_freertos")
```

I quickly tried to include this in our main CMakeLists using `-Dmalloc=UA_malloc` but then base64.c complains that it does not know UA_malloc.

PR #2614, opened 2019-03-29
Ctt fixes

opcua-tsn-team-kalycito:
Contains fixes for CTT

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -519,8 +519,9 @@ processMSG(UA_Server *server, UA_SecureChannel *channel,
         session = &anonymousSession;
     }
 
-    /* Trying to use a non-activated session? */
-    if(sessionRequired && !session->activated) {
+    /* Trying to use a non-activated session?
+     * Do not allow if request is of type CloseSessionRequest */
+    if(sessionRequired && !session->activated && requestType != &UA_TYPES[UA_TYPES_CLOSESESSIONREQUEST]) {
         UA_LOG_WARNING_SESSION(&server->config.logger, session,
                                "Calling service %i on a non-activated session",
                                requestType->binaryEncodingId);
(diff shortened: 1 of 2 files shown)


src/server/ua_server_binary.c
```diff
@@ -519,8 +519,9 @@ processMSG(UA_Server *server, UA_SecureChannel *channel,
         session = &anonymousSession;
     }
 
-    /* Trying to use a non-activated session? */
-    if(sessionRequired && !session->activated) {
+    /* Trying to use a non-activated session?
+     * Do not allow if request is of type CloseSessionRequest */
+    if(sessionRequired && !session->activated && strcmp(requestType->typeName, "CloseSessionRequest")) {
```
jpfr: I'm not sure this has the intended effect.
This allows access if
- The service requires a session
- The session is not activated
- The service is **not** a CloseSessionRequest
Please close this pontential security hole.

Can you move the test for session->activated here?
https://github.com/open62541/open62541/blob/4209f74e2176f75f67b4f80962909c69cdb54126/src/server/ua_services_session.c#L368

jpfr: The strcmp is very indirect logic. Also, there is a build-flag to disable the typenames.
To save ROM on embedded targets. Another solution is
```c
if(sessionRequired && !session->activated && requestType != &UA_TYPES[UA_TYPES_CLOSESESSIONREQUEST])
```
That's also a bit faster.

src/server/ua_services_session.c
```diff
@@ -213,6 +213,11 @@ checkSignature(const UA_Server *server, const UA_SecureChannel *channel,
        channel->securityMode != UA_MESSAGESECURITYMODE_SIGNANDENCRYPT)
         return UA_STATUSCODE_GOOD;
 
+    /* Check for zero signature length in client signature */
+    if(!(request->clientSignature.signature.length)) {
```
jpfr: Can you make this an explicit test?
```c
if(request->clientSignature.signature.length == 0)
    return UA_STATUSCODE_BADAPPLICATIONSIGNATUREINVALID;
```

opcua-tsn-team-kalycito: Done!

src/server/ua_server_binary.c
opcua-tsn-team-kalycito: Done!

jpfr:
Merged. Thanks!

PR #2616, opened 2019-04-01
[WIP] Enable limited browse of node references in nodestore

jonasgreen88:
From the network's point of view OPC UA offers the possibility for
both the client and the server to limit the maximum number of
references to return for a browse request (maxReferencesPerNode).
Without this limitation it wouldn't be possible for many devices to
have nodes with thousands of references due to memory constraints
when building up a browse response.

In open62541 the limitation is in the nodestore API. This proposed
change allow custom nodestores to implement a browse API instead of
returning all references when the stack requests a node. This browse
API is more transparent in what references that are requested and
allows the nodestore to just return the requested references.

### DIFF
--- a/include/open62541/plugin/nodestore.h
+++ b/include/open62541/plugin/nodestore.h
@@ -73,6 +73,15 @@ typedef struct {
     UA_ExpandedNodeId *targetIds;
 } UA_NodeReferenceKind;
 
+/* Reference type, direction and target */
+typedef struct {
+    UA_NodeId referenceTypeId;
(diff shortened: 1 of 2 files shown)


jonasgreen88:
This change is not complete, but posted here to discuss the solution and decide if this is the right way to solve it.

jpfr:
We now can retrieve a node with just a subset of referencetypes (and directions) from the nodestore.
So part of this PR is already covered.
Please reopen / rebase if more is required on top of that.

PR #2630, opened 2019-04-05
Add support for optional DisplayName in NodeSet2.xml

### LINKED ISSUE #2627: OPC UA NodeSet Update Draft - Please check for impacts on this Stack.
Can you take a look at the NodeSet draft in master branch and see if the changes have broken anything for you?
 
DisplayName now omitted if same as BrowseName;
DataTypeDefintion.Name a String rather than a QualifiedName;

Pro:
Fixes #2627
With this commit we support optional display name in the NodeSet2.xml.
If it is not given, we take the name part of the BrowseName.

NOTE: This PR changes the ua-nodeset submodule from the relase branch 1.04 to the current master.

@jpfr do we want to use the 1.04 branch, or current master branch for the nodeset?


jpfr:
1.04.
So we can get fixes on 1.04 without having to review all other changes.

But good to see that we work on the master node set as well.

jpfr:
Looks good. Some things to look at:

Tabs and spaces after if-conditions..

We remove nodes for historic data and events. Are these features tested in the CI?

Maybe we can have feature-based minimal node sets that we pull in on request.

jpfr:
> Maybe we can have feature-based minimal node sets that we pull in on request.

Scratch that. We already have `Opc.Ua.NodeSet2.PubSubMinimal.xml`.
Once the 1.04 nodeset is current, we have to put some more mandatory nodes to the pubsub nodeset.
They were added in a recent amendment.

Pro:
Closing this in favour of #2639.

PR #2631, opened 2019-04-06
[FEATURE] Subscriber implementation

opcua-tsn-team-kalycito:
- Add ReaderGroup, DataSetReader, TargetVariables functionality
- The data filtered from the DataSetReader gets updated in the target variable available in the information model
- Multiple ReaderGroup and DataSetReader are not verified

### DIFF
--- a/src/pubsub/ua_pubsub.c
+++ b/src/pubsub/ua_pubsub.c
@@ -4,8 +4,10 @@
  *
  * Copyright (c) 2017-2018 Fraunhofer IOSB (Author: Andreas Ebner)
  * Copyright (c) 2019 Fraunhofer IOSB (Author: Julius Pfrommer)
+ * Copyright (c) 2019 Kalycito Infotech Private Limited
  */
 
+#include <open62541/server_config_default.h>
 #include "server/ua_server_internal.h"
 
 #ifdef UA_ENABLE_PUBSUB /* conditional compilation */
@@ -21,8 +23,11 @@
 #endif
 
 #define UA_MAX_STACKBUF 512 /* Max size of network messages on the stack */
+#define UA_MAX_SIZENAME 64  /* Max size of Qualified Name of Subscribed Variable */
 
 /* Forward declaration */
+static size_t
+iterateVariableAttrDimension(UA_VariableAttributes varAttr);
 static void
 UA_WriterGroup_deleteMembers(UA_Server *server, UA_WriterGroup *writerGroup);
 static void
@@ -107,6 +112,12 @@ UA_PubSubConnection_deleteMembers(UA_Server *server, UA_PubSubConnection *connec
     LIST_FOREACH_SAFE(writerGroup, &connection->writerGroups, listEntry, tmpWriterGroup){
         UA_Server_removeWriterGroup(server, writerGroup->identifier);
     }
+    /* remove contained ReaderGroups */
+    UA_ReaderGroup *readerGroups, *tmpReaderGroup;
+    LIST_FOREACH_SAFE(readerGroups, &connection->readerGroups, listEntry, tmpReaderGroup){
+      UA_Server_removeReaderGroup(server, readerGroups->identifier);
+    }
+
     UA_NodeId_deleteMembers(&connection->identifier);
     if(connection->channel){
         connection->channel->close(connection->channel);
@@ -129,7 +140,7 @@ UA_Server_addWriterGroup(UA_Server *server, const UA_NodeId connection,
 
     //allocate memory for new WriterGroup
     UA_WriterGroup *newWriterGroup = (UA_WriterGroup *) UA_calloc(1, sizeof(UA_WriterGroup));
-    if (!newWriterGroup)
+    if(!newWriterGroup)
         return UA_STATUSCODE_BADOUTOFMEMORY;
 
     newWriterGroup->linkedConnection = currentConnectionContext->identifier;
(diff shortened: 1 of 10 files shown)


jpfr:
Thanks for this.

Can you remove all the *purely cosmetic* changes?
These make a review quite difficult.

opcua-tsn-team-kalycito:
@jpfr We have removed the purely cosmetic changes.

tools/schema/datatypes_minimal.txt
```diff
@@ -24,6 +24,7 @@ DataValue
 Variant
 DiagnosticInfo
 NodeClass
+ReferenceNode
```
jpfr: All these DataTypes should not be added to datatypes_minimal.txt but to datatypes_pubsub.txt.
Check individually if they are really needed for PubSub. Otherwise we add dead weight to the binaries.

tests/pubsub/check_pubsub_subscribe.c
```diff
@@ -0,0 +1,451 @@
...
+    tcase_add_test(tc_add_pubsub_readergroup, UpdateDataSetReaderConfigWithInvalidConfig);
+    tcase_add_test(tc_add_pubsub_readergroup, UpdateDataSetReaderConfigWithInvalidId);
+    tcase_add_test(tc_add_pubsub_readergroup, UpdateDataSetReaderConfigWithValidConfiguration);
+    tcase_add_test(tc_add_pubsub_readergroup, GetDataSetReaderConfigWithValidConfiguration);
+    tcase_add_test(tc_add_pubsub_readergroup, GetDataSetReaderConfigWithInvalidConfiguration);
+    tcase_add_test(tc_add_pubsub_readergroup, GetDataSetReaderConfigWithInvalidIdentifier);
+    tcase_add_test(tc_add_pubsub_readergroup, CreateTargetVariableWithInvalidConfiguration);
+    tcase_add_test(tc_add_pubsub_readergroup, AddTargetVariableWithInvalidConfiguration);
+
+    Suite *suite = suite_create("PubSub readerGroups/reader/Fields handling and publishing");
```
jpfr: Nice tests!
It would be perfect if eventually we could have a test also for the receiving.
So publisher and subscriber working together.

src/pubsub/ua_pubsub.c
```diff
@@ -4,8 +4,10 @@
  *
  * Copyright (c) 2017-2018 Fraunhofer IOSB (Author: Andreas Ebner)
  * Copyright (c) 2019 Fraunhofer IOSB (Author: Julius Pfrommer)
+ * Copyright (c) 2019 Kalycito Infotech Private Limited
  */
 
+#include <open62541/plugin/log_stdout.h>
```
jpfr: Don't add arch-specific plugins in the main code.
Use the logger from the server config instead.

src/pubsub/ua_pubsub.c
```diff
@@ -182,6 +185,1098 @@ UA_Server_removeWriterGroup(UA_Server *server, const UA_NodeId writerGroup){
...
+    }
+
+#ifdef UA_ENABLE_PUBSUB_INFORMATIONMODEL
+    //ToDo:RemoveGroupRepresentation(server, &readerGroup->identifier);
+#endif
+
+    //UA_PubSubReaderGroup_delete also removes itself from the list
+    UA_PubSubReaderGroup_delete(server, readerGroup);
+    //Remove readerGroup from Connection
+    LIST_REMOVE (readerGroup, listEntry);
```
jpfr: Avoid spaces before braces to be consistent with the remaining library.
At least for function calls and macros (and also before if/for/...).

src/pubsub/ua_pubsub.c
```diff
@@ -182,6 +185,1098 @@ UA_Server_removeWriterGroup(UA_Server *server, const UA_NodeId writerGroup){
...
+    if (pMsg->payloadHeaderEnabled) {
+        anzDataSets = pMsg->payloadHeader.dataSetPayloadHeader.count;
+    }
+
+    for (UA_Byte iterator = 0; iterator < anzDataSets; iterator++) {
+        UA_PubSubDataSetReader* dataSetReaderErg = NULL;
+        if (publisherIdAvailabe || writerGroupIdAvailabe || dataSetWriterIdAvailabe) {
+            UA_PubSubReaderGroup* readerGroup = NULL;
+            LIST_FOREACH(readerGroup, &pConnection->readerGroups, listEntry) {
+                UA_PubSubDataSetReader *tmpReader;
```
jpfr: Here we have nested code with:

- three for-loops
- four if-statements
- a switch statement

That is way too complex.
I don't see any bugs. But I can also not see the absence of bugs. ;-)

Factor out a dedicated method like "getDataSetReadByAttribute" or whatever the correct name should be.

src/pubsub/ua_pubsub.c
```diff
@@ -182,6 +185,1098 @@ UA_Server_removeWriterGroup(UA_Server *server, const UA_NodeId writerGroup){
...
+                    if (publisherIdAvailabe && (tmpReader->config.publisherId.type != NULL)) {
+                        switch (pMsg->publisherIdType) {
+                            case UA_PUBLISHERDATATYPE_BYTE:
+                                if (tmpReader->config.publisherId.type == &UA_TYPES[UA_TYPES_BYTE]) {
+                                    if (pMsg->publisherId.publisherIdByte == (*(UA_Byte*)tmpReader->config.publisherId.data)) {
+                                        UA_LOG_INFO(&server->config.logger, UA_LOGCATEGORY_SERVER, "DataSetReader found with PublisherId");
+                                        dataSetReaderErg = tmpReader;
+                                    }
+                                }
+                                break;
```
jpfr: If you find a dataSetReaderErg, you should not only break the switch-case, but also the LIST_FOREACH loops.
This is not done imho.

If the search is factored out into a dedicated method, you can just return after the first match.

src/pubsub/ua_pubsub.c
```diff
@@ -182,6 +185,1098 @@ UA_Server_removeWriterGroup(UA_Server *server, const UA_NodeId writerGroup){
...
+        dataSetClassIdAvailable = true;
+    }
+
+    // ToDo: Here some filtering is possible.
+    UA_Byte anzDataSets = 1;
+    if (pMsg->payloadHeaderEnabled) {
+        anzDataSets = pMsg->payloadHeader.dataSetPayloadHeader.count;
+    }
+
+    for (UA_Byte iterator = 0; iterator < anzDataSets; iterator++) {
```
jpfr: The iterator is only used after line 467.
The entire search for the UA_PubSubDataSetReader does not use it.
Can we pull out the search for UA_PubSubDataSetReader before the outermost loop?

src/pubsub/ua_pubsub.c
```diff
@@ -182,6 +185,1098 @@ UA_Server_removeWriterGroup(UA_Server *server, const UA_NodeId writerGroup){
...
+        UA_VariableAttributes vAttr = UA_VariableAttributes_default;
+        vAttr.valueRank = pDataSetReader->config.dataSetMetaData.fields[iteratorField].valueRank;
+        if (pDataSetReader->config.dataSetMetaData.fields[iteratorField].arrayDimensionsSize > 0) {
+            retVal = UA_Array_copy(pDataSetReader->config.dataSetMetaData.fields[iteratorField].arrayDimensions, pDataSetReader->config.dataSetMetaData.fields[iteratorField].arrayDimensionsSize, (void**)&vAttr.arrayDimensions, &UA_TYPES[UA_TYPES_UINT32]);
+            if (retVal == UA_STATUSCODE_GOOD) {
+                vAttr.arrayDimensionsSize = pDataSetReader->config.dataSetMetaData.fields[iteratorField].arrayDimensionsSize;
+            }
+        }
+
+        switch (pDataSetReader->config.dataSetMetaData.fields[iteratorField].builtInType) {
```
jpfr: There is a lot of code duplication going on here.

Don't.

Instead, find the correct dataype in the UA_TYPES array.
And then use the same code path for type.

src/pubsub/ua_pubsub.c
```diff
@@ -182,6 +185,1098 @@ UA_Server_removeWriterGroup(UA_Server *server, const UA_NodeId writerGroup){
...
+ * @param server
+ * @param dataSetReader
+ * @param dataSetMsg
+ */
+void UA_PubSubDataSetReader_process(UA_Server *server, UA_PubSubDataSetReader *dataSetReader, UA_DataSetMessage* dataSetMsg) {
+    if ((dataSetReader == NULL) || (dataSetMsg == NULL) || (server == NULL)) {
+        return;
+    }
+
+    if (dataSetMsg->header.dataSetMessageValid)
```
jpfr: Return early if the message is not valid --> saves one level of indendation of the code.

src/pubsub/ua_pubsub.c
```diff
@@ -336,7 +1431,7 @@ UA_Server_removeDataSetField(UA_Server *server, const UA_NodeId dsf) {
     parentPublishedDataSet->fieldSize--;
     if(currentField->config.field.variable.promotedField)
         parentPublishedDataSet->promotedFieldsCount--;
-
```
jpfr: Cosmetic changes in this file...

opcua-tsn-team-kalycito:
Thanks for your comments.

@jpfr We have addressed the review comments this Pull Request.
Kindly have a look.

jpfr:
:+1:

PR #2661, opened 2019-04-23
[CTT Fix] Remove localhost from hostname in server certificate

opcua-tsn-team-kalycito:
- Currently generated server certificate (server_cert.der) has
   hostname set to localhost and IP Address set to 127.0.0.1
 - In new CTT, using "localhost" as hostname in a certificate or
   EndpointUrl is not allowed
 - So modified Python script and the corresponding configuration
   file (localhost.cnf) to dynamically identify the network
   interfaces and the corresponding IP addresses

Signed-off-by: Asish Ganesh <asish.g@kalycito.com>

### DIFF
--- a/tools/certs/create_self-signed.py
+++ b/tools/certs/create_self-signed.py
@@ -1,8 +1,12 @@
 #!/usr/bin/env python
 # This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this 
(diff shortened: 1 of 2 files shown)


jpfr:
Thanks for this!

opcua-tsn-team-kalycito:
The modified python script requires netifaces package to build. Please add the following command in the build steps section of the documentation:

`pip install netifaces`

Thanks
opcua-tsn-team-kalycito

PR #2674, opened 2019-04-26
[FIX] Add security level value for corresponding message security mode

opcua-tsn-team-kalycito:
- Currently security level value not initialized corresponding to
   the message security mode
 - In CTT, "Security Certificate Validation" test cases gets skipped
   since security level value is zero for all security modes
 - So modified the security level value corresponding to the message
   security mode

Signed-off-by: Asish Ganesh <asish.g@kalycito.com>


jpfr:
Great! Merged.

PR #2679, opened 2019-04-27
Handle NULL in UA_STRING and UA_BYTESTRING

jpfr:


Pro:
@jpfr is there a reason why you removed the connect loop example?

We had some questions in the past how to detect and reconnect broken connections...

jpfr:
@Pro Right. I didn't see that purpose. Restored.

We need a bare-bones "let's see if we can connect" client.
That also supports username/password + selectable securitypolicies that can be driven from the command line.

Issue #2702, opened 2019-05-03
Request with timestamp=0 causes Prosys OPC UA Client to fail


alex-lotz:
## Description

I know this is maybe not entirely related to open62541, but I am using the Prosys OPC UA Client that has worked fine with an older checkout of the open62541 library. After updating my local git clone, the Prosys client fails to connect to any server examples from the open62541 library. While attempting to connect, the Prosys Client displays the error message "Cannot discover applications from ...", which I interpret such that the client fails reading the server space for some reason.

Maybe this is not an issue of the open62541 library by itself, but something must have been changed that prevents the Prosys client to work. Do you have any experience with the Prosys Client, or can you recommend any other generic clients that particularly work on Linux?

## Background Information / Reproduction Steps

For reproducing the error, I cloned the most recent update of the open62541 library and built the examples in a standard way using cmake. I have been using the "server_mainloop" example for testing. After starting the server_mainloop example, I tried to connect the Prosys client with the above mentioned error.

Used CMake options:

```bash
cmake -DUA_BUILD_EXAMPLES=ON ..
```

## Checklist

Please provide the following information:

 - [X] open62541 (most recent update of the master branch, from 3rd may 2019):
 - [X] Other OPC UA SDKs used: [Prosys OPC UA Client](https://www.prosysopc.com/products/opc-ua-client/)
 - [X] Operating system: Ubuntu 18.04 64 Bit

Pro:
Thanks for the report!

Do you see any output on the open62541 Server side?

Can you try to build with `-DUA_LOGLEVEL=100` (=Trace) and post the output of the server here?

alex-lotz:
```
[2019-05-03 12:06:18.456 (UTC+0200)] info/network	TCP network layer listening on opc.tcp://nb-alotz:4840/
[2019-05-03 12:06:25.730 (UTC+0200)] trace/network	Connection 5 | New TCP connection on server socket 3
[2019-05-03 12:06:25.730 (UTC+0200)] info/network	Connection 5 | New connection over TCP from 127.0.0.1
[2019-05-03 12:06:25.780 (UTC+0200)] trace/network	Connection 5 | Activity on the socket
[2019-05-03 12:06:25.780 (UTC+0200)] trace/network	Connection 5 | Received a packet.
[2019-05-03 12:06:25.780 (UTC+0200)] trace/network	Connection 5 | No channel attached to the connection. Process the chunk directly
[...]

PR #2703, opened 2019-05-03
[ON-HOLD] Stack: Add warning if client sends request with 0 timestamp

Pro:
On hold until #2702 is resolved


jpfr:
Fixed via #2705.

PR #2709, opened 2019-05-06
Stack: Correctly pass pointer to encoding for DiagnosticInfo

Pro:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12889

Credit to oss-fuzz


jpfr:
👍

PR #2718, opened 2019-05-08
Fix setup of namespace 1 URI

### LINKED ISSUE #2692: Server configuration and initialisation flaws
## Description

The server configuration possibilities have some flaws in th current master, IMHO.

## The concerns

1. Calling `UA_Server_new` already does some initialisation, but the configuration is not yet setup. This leads to the server namespace index 1 to refer uri `UA_STRING_NULL`. This is since fd08d012d95b5837c90b18ba3fd7230b2c4d8ef6 as far as I can tell.

2. The custom logger cannot be setup before calling any of the `UA_ServerConfig_setXXX` methods, as they cleanup the complete config again. You could setup the logger after this call, but several other things, like `UA_SecurityPolicy` and `UA_ServerNetworkLayerTCP` copy this before you can change it.

3. `UA_ServerConfig_setXXX` do usefull things where many of them you don't want to repeat in your code. But they use a relative fixed default configuration. For example, you cannot setup a (dynamic) custom `config->applicationDescription`. You can change it afterwards, but 

   a. `UA_Server_new` uses the applicationUri before you can ever get access to the config. (See point 1).

   b. `createEndpoint` copies the applicationDescription into the endpoints that are created by UA_ServerConfig_setXXX kind of function before you can change it.

## Proposal

Split the monolithic UA_ServerConfig_setXXX functions so that someone who is interested to do so can setup a config default, and then probably change some aspects, add his network layers and enpoint configuration with some helpers that hide most of the complexity but leave some more customization possibilities.

1. Expose `setDefaultConfig` to the library consumer as `UA_ServerConfig_setDefaults`. This function shouldn't setup anything which copies something of its init values, especially `logger` and `applicationDescription`.

2. The library consumer could modify the values which he doesn't like from the defaults.

3. Expose `addDefaultNetworkLayers` as `UA_ServerConfig_addNetworkLayerTCP`

4. Introduce `UA_ServerConfig_addSecurityPolicy_None`, `UA_ServerConfig_addSecurityPolicy_Basic128Rsa15`, `UA_ServerConfig_addSecurityPolicy_Basic256` and `UA_ServerConfig_addSecurityPolicy_UA_SecurityPolicy_Basic256Sha256`

5. Add `UA_ServerConfig_addSecurityPolicy_all`

6. Introduce `UA_ServerConfig_addEndpoint` which takes the securityPolicyUri and securityMode and uses createEndpoint to setup the endpoint.

7. Add `UA_ServerConfig_addDefaultEndpoints`

8. `UA_Server_new` must not do anything that is based on config. This must be moved to `UA_Server_run_startup`.

Any comments welcome. I would probably prepare a PR if we agree on how / what to do.

jojastahl:
Fix namespace 1 initialisation problem that was introduced in commit fd08d012d95b5837c90b18ba3fd7230b2c4d8ef6.

Fixes #2692

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -34,6 +34,15 @@
 /* Namespace Handling */
 /**********************/
 
+/*
+ * The NS1 Uri can be changed by the user to some custom string.
+ * This method is called to initialize the NS1 Uri if it is not set before to the default Application URI.
+ *
+ * This is done as soon as the Namespace Array is read or written via node value read / write services,
+ * or UA_Server_addNamespace, UA_Server_getNamespaceByName or UA_Server_run_startup is called.
+ *
+ * Therefore one has to set the custom NS1 URI before one of the previously mentioned steps.
+ */
 void setupNs1Uri(UA_Server *server) {
     if (!server->namespaces[1].data) {
         UA_String_copy(&server->config.applicationDescription.applicationUri, &server->namespaces[1]);
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -405,6 +405,9 @@ verifyServerApplicationURI(const UA_Server *server) {
 
 UA_StatusCode
 UA_Server_run_startup(UA_Server *server) {
+    /* ensure that the uri for ns1 is set up from the app description */
+    setupNs1Uri(server);
+    
     if(server->state > UA_SERVERLIFECYCLE_FRESH)
         return UA_STATUSCODE_GOOD;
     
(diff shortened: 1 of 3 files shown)


jojastahl:
>We should document at which point the ns1 URI is finalized in the server.
This is as soon as the Namespace Array is read or written via node value read / write services, or UA_Server_addNamespace, UA_Server_getNamespaceByName or UA_Server_run_startup is called.

>What would be required to allow changes later in the lifecycle?
I do not know if this should be possible. Currently writing to the namespace array via node value write service actively checks that existing namespaces are not changed. So I'm afraid, that changing a namespace URI later on is an allowed operation as this changes the identification of all nodes referring to its index.

Pro:
I added some more comments. This can be merged as soon as CI is green

Issue #2688, opened 2019-04-29
Modifying monitored items with open62541 v0.3.0


basyskom-jvoe:
According to [this comment](https://github.com/open62541/open62541/blob/0.3/include/ua_client_subscriptions.h#L171), UA_Client_MonitoredItems_modify() only operates on the server and does not change the local monitored item representation in the client.

When a monitored item is created using UA_Client_MonitoredItems_createDataChange(), the client handle from the request [is overwritten](https://github.com/open62541/open62541/blob/0.3/src/client/ua_client_subscriptions.c#L248).

The client handle is required for a call to UA_Client_MonitoredItems_modify(). If the client handle does not match the client handle used to create the monitored item on the server, all data changes following the modification of the monitored item are not dispatched to the data change callback because the server uses the client handle from the monitored item modify request.

What is the recommended way to deal with the client handle when creating and modifying a monitored item with the monitored items API in open62541 v0.3?

basyskom-jvoe:
The only option I can think of is attempting to track the value of client->monitoredItemHandles in the application that uses open62541 to create the monitored items. This is very error prone.

For the open62541 side, the least intrusive option would be to add a getter like `UA_UInt32 UA_Client_getLastMonitoredItemHandle(UA_Client *client)` to the library. This could be used to determine the client handle after checking if the monitored item has been successfully created.

Just using the value from the request could break existing applications that don't set the client handle and expect open62541 to handle this.

PR #2746, opened 2019-05-17
Fix modification of monitored items

basyskom-jvoe:
A proposed fix for issue #2688

Setting the client handle from the monitored item id before sending the
modify request to the server solves that problem.

Note: The same issue should also exist on master.

### DIFF
--- a/include/ua_client_subscriptions.h
+++ b/include/ua_client_subscriptions.h
@@ -101,8 +101,8 @@ UA_Client_Subscriptions_setPublishingMode(UA_Client *client,
  * forward Event notifications from that node.
  *
  * During the creation of a MonitoredItem, the server may return changed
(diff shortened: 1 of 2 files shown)


jpfr:
This is a good idea.

Two wishes:

- The remaining "high level api" for subscriptions does not directly expose the requests/responses. It would be consistent to just take the MonitoredItem ID and the MonitoringParameters as arguments.
- We should document that the clientHandle parameter can not be set by the user.

jpfr:
Update:
Oh. We do expose the requests/responses in many cases.
Scrap the first point.

basyskom-jvoe:
I have added some documentation

basyskom-jvoe:
@jpfr Thanks for the merge!

PR #2748, opened 2019-05-21
Nodestore Switch

JGrothoff:
Added a nodestore switch with example as plugin.

Based on the namespace index a nodestore is picked. If no nodestore is linked for a namespace the default nodestore is used. The nodestore switch links namespace indices to nodestores. So that every access to a node is redirected, based on its namespace index.
The linking between nodestores and and namespaces may be altered during runtime (With the use of UA_Server_run_iterate for example). This enables the user to have a persistent and different stores for nodes (for example in a database or file) or to transform objects into OPC UA nodes. Backup scenarios are also possible.
This solution is also used in ACPLT/RTE to transform the internel object model to UA.


jpfr:
Thanks @JGrothoff !

Allowing custom nodestores is a big feature that other SDKs usually don't offer.
Good to see this example happening.

This PR could be done more "minimally invasive".
Especially, the feature can be implemented without touching the core server code at all.
Just in its own plugin!

- Clone `open62541/plugins/ua_nodestore_default.c` to `open62541/plugins/ua_nodestore_custom.c` or similar
- Add a new example (in the CMakeLists.txt) that switches out `ua_nodestore_default.c` in the plugin list that it wants to link against
- Inside `ua_nodestore_custom.c` you have full freedom. You can even define new public API for that plugin in something like `ua_nodestore_custom.h`.

Maybe we can have a self-contained example where the switch is actually used.
For example by defining a long array of integers where each entry is translated to a variable when it is looked up:

    'ns=3;i=<index>' returns a VariableNode with the content from the array at <index>

JGrothoff:
Hi and thanks for your response.

We need to touch the server only to get the nsCtxt, as it is created during server startup and because it is not part of the config anymore.

> Inside ua_nodestore_custom.c you have full freedom. You can even define new public API for that plugin in something like ua_nodestore_custom.h.
We already added the "nodestore_switch.c", so we could use it to integrate a clone of the default nodestore.

> Clone open62541/plugins/ua_nodestore_default.c to open62541/plugins/ua_nodestore_custom.c or similar
We didn't clone the default nodestore, because this would lead to code duplication. So every change in the default nodestore has to be performed manually in the "nodestore_custom" or in the "nodestore_switch" respectively. Maybe there is a better solution to clone the default nodestore to the nodestore_switch, so that the cloning is explicit? (Kind of git submodule?) 

> Maybe we can have a self-contained example where the switch is actually used.
This is possible, i had an example like this two years ago. But an easier example, which i like to provide first is an unlink of the ns1 nodestore, then a restart of the server and ns1 nodestore is linked again. In a future example we could map the ns1 nodestore memory to a file and load it again between the server restart to show how to build a persistent nodestore. Instead of mapping to a file a database could also be used. Maybe i can provide both examples: one to show a custom nodestore and one to show the persistence.

But more important for us is to have a unit-test first.

jpfr:
Sounds very good.

The spec v1.04 defines a binary nodeset format. (Copied verbatim from the unified automation sdk...)

This can lead to a drastic reduction in required Ram and Rom together with a custom nodestore.

I think nobody is working on it yet.

JGrothoff:
@jpfr So do you recomend to duplicate the code from nodestore_default to nodestore_switch instead of reusing it?

jpfr:
> So do you recomend to duplicate the code from nodestore_default to nodestore_switch instead of reusing it?

Yes. The original nodestore is only 400 lines anyway.
It is better to have less indirection in the base library.
And a slightly more verbose example for the special case nodestore.

matkonnerth:
Hi @jpfr, @JGrothoff ,
I've done some prototyping with loading a nodeset provided as xml with the libxml sax parser (https://github.com/matkonnerth/nodesetLoader/blob/master/README.md). The nodes parsed from the nodeset are then to the open62541 default nodestore via the public api (UA_Server_addNode...). Here's the import example with open: https://github.com/matkonnerth/open62541/tree/nodesetImport/examples/import 

It supports loading multiple nodeset.xml. You have to provide it in their dependent order (e.g. DI Model - PLCOpen  model).
Please consider, that's only experimental (loading extensions, values of variables, ... is currently not supported).

A much better way would be adding the nodes directly to a custom nodestore, which will result in speedup of nodeset loading. What do you think of, may I use the nodestore switch for this?

mlang-de:
@jpfr 
> The spec v1.04 defines a binary nodeset format. (Copied verbatim from the unified automation sdk...)

I know this is OT, but where is this specified in 1.04?
br

JGrothoff:
Hi @matkonnerth, thanks for your reply.

> A much better way would be adding the nodes directly to a custom nodestore, which will result in speedup of nodeset loading. What do you think of, may I use the nodestore switch for this?
 
Yes and no: you can use the nodestore switch, with the default nodestore via the nodestore API for that. But i guess the speed up isn’t worse it. The public API takes care of many things, like setting references in Forward and Backward directions, checking types and so on. But if you‘re loading a nodeset, that is consistent, like the mentioned ones you are good to go.
Still i‘m very interested to see, if there would be a speed up and how your code would evolve.
So if you want to try it, i could provide a small example for you and we could add it as an example for open62541 later on.

Some other considerations on your idea -regarding the nodestore:
+ public Server API —> Could be easily changed to a Client API, so that you can load a nodeset to any server in the network
+ Nodestore Switch —> The lifecycle of the nodestore gets decoupled from the server lifecycle: You need to load the nodes for these namespaces only once and could even have multiple Servers refer to the same nodestore, as it is multithreading capable.
+ Custom Nodestore: You could load the xml file to an xml database like base-x or into an optimized xml memory. Then the nodeset loading is decoupled from server startup and is optimized. You can even centralize the nodeset storage and let multiple servers read from this nodestore (via sql, x-queries, ...) But the runtime access to the nodes will be much slower and needs transformation (XML to UA Node). And you also need to change encoding of custom datatypes.

Actually, i‘m quite busy in the moment and will work on this PR in end of July.

Cheers,
Julian

matkonnerth:
Hi @JGrothoff,
as you mentioned, I would do the checks on the model during generating the model upfront loading it. Without the fastNodes pr the stacks takes quite a while to load large nodesets (I'm confronted with nodeset with ~100k nodes).

It would be nice, if you could advise with a small example, then I can go on extending it.

jpfr:
> I know this is OT, but where is this specified in 1.04?

Here is the doc from unified.
https://documentation.unified-automation.com/uasdkhp/1.1.0/html/md_opcua_binary_fileformat.html

You are right. It's not part of 1.04.
Then I must have seen it on some roadmap.

Anyway, some vendors use the binary nodeset format to exchange server configuration.
With respect to PubSub, where the config has a nodeset representation.

JGrothoff:
Hi,
i have added a more verbose example with a simple custom nodestore.
I'm going to add a unit test in the future. For now it would be nice to run the check_nodestore test a second time with UA_ENABLE_CUSTOM_NODESTORE and UA_ENABLE_NODESTORE_SWITCH.
So the nodestore interface compatibility of the switch and hence the consistency of the nodestore interface and its redirection to the default nodestore would be checked.
How can we achieve this?

@jpfr  Can you review the code so far and merge, so i can add the tests in a second PR later?

Maybe we should refactor UA_ENABLE_CUSTOM_NODESTORE to UA_ENABLE_NODESTORE_CUSTOM?

PR #2751, opened 2019-05-22
Fix: Add UA_EXPORT to UA_SECURITY_POLICY_NONE_URI for shared builds

ccvca:
Missing UA_EXPORT that causes problems when building a shared library in windows.


jpfr:
Kudos for finding this. 👍 
Thanks for the PR.

PR #2764, opened 2019-05-27
Remove optional description from Minimal Nodeset and add XML_Schema node

Pro:
The current 1.04 nodeset removed all the node descriptions. This PR aligns the minimal nodeset with the current master.

In addition the XML Schema (i=92) node is added to allow using Reduced nodeset with the model compiler.
Also `HasInterface` and `BaseInterfaceType` nodes are added to reduced nodeset.

(see https://opcua.rocks/custom-information-models/)


jpfr:
Nice!

Pro:
Included in #2743

PR #2767, opened 2019-05-28
CMake: Cleanup of cmake code for doc generation

Pro:
Replaces #2552 

With this code, the copy command is replaced with a custom command which makes sure that the dependencies are correctly set.


jpfr:
👍

PR #2794, opened 2019-06-06
Fixes for CTT

opcua-tsn-team-kalycito:
- Update the ServerArray with same value as NamespaceArray
 - Add support for reading multiple CRL files to server_ctt


jpfr:
:+1:

PR #2809, opened 2019-06-12
Fuzz: Reproduce fuzz results in default environrment

jpfr:


Pro:
@jpfr there are two ways to reproduce test-cases on the current master branch:

Use the oss-fuzz infra:
```
python $HOME/repos/oss-fuzz/infra/helper.py build_fuzzers --sanitizer address open62541 $HOME/repos/open62541 && python $HOME/repos/oss-fuzz/infra/helper.py reproduce open62541 fuzz_binary_message -runs=2  -rss_limit_mb=2048 $HOME/Downloads/clusterfuzz-testcase-fuzz_binary_message-5765057679458304
```

And more specifically, you can just pass a binary/testcase file to the `fuzz_binary_message` executable, which is already created. That's exactly what you are trying to add with this PR, isn't it?

i.e. build with:

```cmake
-DCMAKE_BUILD_TYPE=Debug
-DUA_BUILD_EXAMPLES=ON
-DUA_ENABLE_AMALGAMATION=OFF
-DUA_NAMESPACE_ZERO=FULL
-DUA_BUILD_FUZZING=ON
-DUA_BUILD_UNIT_TESTS=ON
```

and then run:

```bash
./fuzz_binary_message -runs=2  -rss_limit_mb=2048 $HOME/Downloads/clusterfuzz-testcase-fuzz_binary_message-5670554250510336
```

I don't really see how this PR improves that

jpfr:
Not all my machines run Linux.
Stable Clang on Windows does not yet come with libFuzzer.

This is meant to reduce friction when debugging oss-fuzz testcases.

The way the custom malloc is handled is a bit unorthodox.
I'll bang this into review shape and get back to you.
We can still reject if the overhead is not worth it.

Pro:
Ahh, right, didn't think about Windows. For that we need this PR, so all good, thanks for the clarification!

The custom memory manager was more like a hack to get it into fuzzing. Im happy for any refactoring/improvements!

Pro:
@jpfr I just looked through your changes and have a few comments:

First of all you are mixing various changes into one huge PR. I'd prefer that you split this PR into multiple smaller PRs for easier review:

- Bugfixes
- Custom Memory Manager Changes
- Fuzz Changes / Reproduce
- Architecture folder changes

In addition to that, I have the following comments:

- Custom Memory Manager:
You changes will invalidate the currently created corpus, but this should be fine, for the sake of simplifying the memory manager

- Architecture folder changes:
In the current version it's possibly to add a new architecture by just adding a new folder. With this you do not need to change the CMake code of the current CMakeLists. (except of adding ad `add_subdirectory`)
After your changes one needs to update the arch/CMakeLists file. Is this desired?
Also a specific architecture can not remove definitions anymore. I'd prefer that you move these changes into a separate PR and then we include @cabralfortiss in the review, since he did quite some work there.

jpfr:
Hey @Pro ,

welcome back!

> I'd prefer that you split this PR into multiple smaller PRs for easier review.

Jup. This is already under way.
The View-Commits are still listed here, but were actually merged in a separate PR this morning.

> You changes will invalidate the currently created corpus, but this should be fine, for the sake of simplifying the memory manager.

The main change is that the memory manager counts allocated memory without removing freed memory from the count. Because with this, the fuzzer can precisely target each malloc to fail.
Regardless of whether other memory has been freed just before.
Let's evaluate this independent from the other changes.

>  After your changes one needs to update the arch/CMakeLists file. Is this desired?

This was an experiment in the simplification of /arch. Partly successful, partly not.

- The "include hell" of the architectures could be reduced quite a bit. With even more to come with the incoming networking overhaul. Have a look at the /arch/<arch> folders in this branch. They are **so simple** now.
- The original idea of using global properties to relay the architecture settings back into the main context is much better than variables with "parent_scope"
- We can go back to one-CMakelists.txt per arch. But remember that network_tcp.c is used by all of them. If we treat the archs separately, we might include it multiple times into the sources.
In both cases, I prefer to set the global property variables directly instead of hiding this mechanism behind macros.

As you said, let's dissect this into individual PR and discuss.

By the way, even with great care, all arch changes probably bork some stuff that is not under automated testing. We should either have a build env for every arch, at best in a docker container, or dedicated arch-maintainers that can make test runs. Otherwise, less-used architectures will be perpetually left behind.

mlgiraud:
@jpfr Keep in mind that my networking overhaul removes network_tcp.c completely. The relevant code is then a plugin in the plugins directory.

Pro:
@jpfr ping.

Did you already split this PR up, as mentioned above?

PR #2812, opened 2019-06-12
Encoded passwort in the server (independent from the SecureChannel)

jpfr:


jpfr:
@jonasgreen88 

This is a rebased version of your implementation in #2390.
Can you review, try out and greenlight?

Do you have something like a unit test for this as well?

jonasgreen88:
@jpfr 

Sorry for late response. I'm just about to merge the current master to our internal repository. Before that I'm unable to try it out. We have product tests verifying this solution in our product, but no unit test for this repo.

PR #2833, opened 2019-06-19
CMake fixes for branch 0.3

DerDakon:
I tried using the (cross-)compiled 0.3.0 in another CMake project, which failed as the config file was not found automatically. This fixes this and a few minor issues along the way.

PR #2851, opened 2019-06-28
CMake: use GNUInstallDirs to set library destination

Pro:
This will know the right places for many systems without manual intervention.

Ports #2833 to master


jpfr:
Merged via #2158

PR #2853, opened 2019-06-28
For 1 dimension arrays, initializing arrayDimensionsSize may confuse some OPC clients.

aminosu:
PR #1810 caused to loose the fix done in PR #1978...
I've re-introduced that fix and added a unit test to avoid this in the future.
In our case the problem showed in two OPC-UA Client (“Unified Automation UaExpert” and “Integration Objects OPC UA Explorer”):
when trying to do a Method Call, no Input / Output Arguments were shown (UA Explorer told they’re Opc.Ua.Matrix instead of array).


jpfr:
Nice, thanks.

Issue #2774, opened 2019-05-29
LDS-ME does not work on Windows devices


royalheadwear:
## Description

Hello.

I'm currently trying to run the example LDS-Server on multiple device-combinations with different operating systems. I am using the following device-combinations:

| # | first device | second device |
|----------------------------|------------|---------------|
| 1 | Raspberry Pi with raspian stretch | Raspberry Pi with raspian stretch |
| 2 | Raspberry Pi with raspian stretch | Desktop PC with Windows 10 |
| 3 | Desktop PC with Windows 10 | Raspberry Pi with raspian stretch |

All devices are connected via Ethernet.

I'm just using the following three applications from the examples:

 - ``discovery_server_lds``
 - ``discovery_server_multicast``
 - ``discovery_client_find_servers``

I always start an instance of the ``discovery_server_lds`` on each device. Additionally I run an instance of the ``discovery_server_multicast`` on the first device. Then I run ``discovery_client_find_servers`` on the second device to show if I can find the running server instances.

With device-combination 1 everything works fine. But if I replace one of the devices with a Windows 10 machine, the servers on the other device can not be found. As far as I understood the LDS-ME-feature can be used to find servers and endpoints in the local network no matter on which device the server is running - as long as there is a LDS-ME on that device, too.

## Background Information / Reproduction Steps

I used the following CMake options to generate the library files and compile the examples:

``` BUILD_SHARED_LIBS                OFF
 CLANG_FORMAT_EXE                 CLANG_FORMAT_EXE-NOTFOUND
 CMAKE_BUILD_TYPE                 Debug
 CMAKE_INSTALL_PREFIX             /usr/local
 LIB_INSTALL_DIR                  lib
 MDNSD_LOGLEVEL                   300
 UA_ARCHITECTURE                  posix
 UA_BUILD_EXAMPLES                ON
 UA_BUILD_TOOLS                   OFF
 UA_BUILD_UNIT_TESTS              OFF
 UA_ENABLE_AMALGAMATION           OFF
 UA_ENABLE_COVERAGE               OFF
 UA_ENABLE_DISCOVERY              ON
 UA_ENABLE_DISCOVERY_MULTICAST    ON
 UA_ENABLE_ENCRYPTION             OFF
 UA_ENABLE_EXPERIMENTAL_HISTORI   OFF
 UA_ENABLE_HISTORIZING            OFF
 UA_ENABLE_METHODCALLS            ON
[...]

PR #2408, opened 2019-01-31
[REVIEW] Networking code overhaul

mlgiraud:
Continuation of #2271 after rebasing onto master.
I had to solve some massive conflicts and a lot of stuff is now broken again, so i will continue fixing on this branch and leave the old one as is.

PR #2856, opened 2019-06-29
[Review] First attempt of mDNS interface selection (relates to #2774)

Stasik0:
I made a most-easy way of setting the outbound inerface by entering exact IP. In case user wants some more sugar like selecting interface from string, e.g. "eth0", she will need to do it in user-space.

#2774

### DIFF
--- a/examples/discovery/server_multicast.c
+++ b/examples/discovery/server_multicast.c
@@ -251,6 +251,8 @@ int main(int argc, char **argv) {
 
     config->discovery.mdns.mdnsServerName = UA_String_fromChars("Sample Multicast Server");
 
+    config->discovery.mdnsInterface = UA_String_fromChars("docker0");
+
     // See http://www.opcfoundation.org/UA/schemas/1.03/ServerCapabilities.csv
     // For a LDS server, you should only indicate the LDS capability.
     // If this instance is an LDS and at the same time a normal OPC UA server, you also have to indicate
--- a/examples/discovery/server_multicast.c
+++ b/examples/discovery/server_multicast.c
@@ -252,7 +252,7 @@ int main(int argc, char **argv) {
     config->discovery.mdns.mdnsServerName = UA_String_fromChars("Sample Multicast Server");
 
     //setting custom outbound interface
-    config->discovery.mdnsInterfaceIP = UA_String_fromChars("42.42.42.42"); //this line will produce an error, and set the interface to 0.0.0.0
+    config->discovery.mdnsInterfaceIP = UA_String_fromChars("42.42.42.42"); //this line will produce an error and set the interface to 0.0.0.0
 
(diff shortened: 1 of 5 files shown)


ccvca:
I have only skimmed over the previous communication #2774 , wouldn't it make more sense to send on all interfaces where the OPC UA Server is reachable as fallback? Selecting the first one seems random to me.

Pro:
That's a valid point...

Currently we are just adding a tcp socket for every interface we find:
https://github.com/open62541/open62541/blob/master/arch/network_tcp.c#L359
In the future this should be changed to give the user an option to select a specific interface by specifying an IP address where the server should listen (i.e., 0.0.0.0 is the default and means all interfaces). That should then also be part of #2408 here:
https://github.com/open62541/open62541/pull/2408/files#diff-431422d938c278cd653aa380ecdcfd40R292 (ping @mlgiraud)


For mDNS we could use this information and if the tcp listener interface is set to:
- 0.0.0.0:
	- default: check all interfaces which have multicast enabled, and send mDNS on all of them
    - in addition the user may limit the interface selection to a subset of these
- specific ip: only send mDNS packages on the interface which has that ip

Does that make sense?

The next issue is then, what about the case where a server only has UDP interface, or even ETH?
Maybe the most intuitive configuration option is, to configure the mDNS socket independent of the listener sockets.

I.e.:
- Default behaviour: check all interfaces which have multicast enabled, and send mDNS on all of them
- The user can select one or more interfaces where mDNS is sent

Stasik0:
> I have only skimmed over the previous communication #2774 , wouldn't it make more sense to send on all interfaces where the OPC UA Server is reachable as fallback? Selecting the first one seems random to me.

This PR extends the default 0.0.0.0 behavior to be able to specify one non-random interface.

> Default behaviour: check all interfaces which have multicast enabled, and send mDNS on all of them
Here is some example linux-code https://github.com/troglobit/mdnsd/blob/b12ac76c456fb146a697bc3170e67a9ffc7dedbd/src/addr.c
Is there any posix way to do that?

+1 for @Pro's proposal. But it is done in #2408, right?

Do we still want to have this PR merged in? Maybe we should replace the one IP-String by an array of strings for later?

jpfr:
👍 For merging this.
Many people have that problem on boxes with many interfaces (LAN, Wifi, VMWare, etc.)

And we need the hint for #2048.
There, we start with "normal" TCP for the new network manager and then bring over mDNS and other communication gradually over time.

@Stasik0 Welcome back!
Can you squash out your fix-commits in a rebase?

@Pro You're the master of mDNS. Can you do the merge when you are okay?

Pro:
I opened a new issue #2864 to collect the ideas from this PR and squash-merged these changes.
Thanks @Stasik0

PR #2872, opened 2019-07-07
[REVIEW] add support for websocket transport

matkonnerth:
This pr tries to bring up websocket transport. It depends heavily on libwebsockets (https://github.com/warmcat/libwebsockets)

Can be enabled with UA_ENABLE_WEBSOCKET_SERVER

It would be interesting, if anybody knows a opc ua client implementation, that supports binary encoding over websockets.


jpfr:
Oh wow. This is a welcome surprise. 👍

matkonnerth:
Do you know if there are other clients available, which support ua binary encoding over websocket transport? At the moment I'm testing with a generic websocket client, opc ua messages recorded with wireshark. Looks not bad, but is painful ;)

Pro:
Thanks for this nice addition!

I also do not know about any OPC UA Websocket client.

The most straight forward way is probably to collaborate with the node-opcua project on this.
I already created an issue there to ask about their status:
https://github.com/node-opcua/node-opcua/issues/638

matkonnerth:
I think it's time for a first review. Supports websocket transport for server. I've only tested manually with a proprietary generic client.

Things which are currently missing:
- add endpoint for websocket transport in ua_config_default.c
- add encryption

For me these are points for further PRs.

CMakeLists.txt
```diff
@@ -113,7 +113,14 @@ set(ua_architecture_headers ${ua_architecture_headers}
     ${PROJECT_SOURCE_DIR}/include/open62541/architecture_functions.h
 )
 
+if(UA_ENABLE_WEBSOCKET_SERVER)
+    add_definitions(-DUA_ENABLE_WEBSOCKET_SERVER)
```
Pro: This should be put into `include/open62541/config.h.in`, because then it will always be defined automatically and you do not have to add it to every gcc call.

I'll update this PR and then merge it.

Pro:
I just updated this PR.
We can merge it when CI is green to avoid bit rot.

@matkonnerth in future PRs the following points need to be addressed:

* add endpoint for websocket transport in ua_config_default.c
* add encryption
* enable `UA_ENABLE_WEBSOCKET_SERVER` in CI and make sure that option is tested.
* Write unit tests

mlgiraud:
I'd like to remark that encryption should already be supported because the securechannel is created on top of the websockets connection.

Also, since i'm going to have to rewrite this PR for my network overhaul: Which client did you use to test your implementation? I couldn't find any open source clients with websockets support.

matkonnerth:
I've used a proprietary websocket client, unfortunately it's not open source. In the long term I must think of how to test the websocket transport.

Short term solution is that I test your changes.

mlgiraud:
@matkonnerth I just finished a first version of the port. You can try it out here: #2408 
Please let me know if you discover any issues or memory leaks.

mlgiraud:
@matkonnerth Any update on testing?

juangburgos:
Hi, quick question. Only server-side is implemented here for websockets right? Client-side missing? Is not possible to test open62541 client with open62541 server through websockets right now?

PR #2879, opened 2019-07-10
Fix: compatibility with CMake version 3.0.2

jrcoding:
With **cmake 3.0.2** the command **cmake -E copy ...** doesn't copy a list of files but only single files therefore the **add_custom_command** fails on i.e. debian8.
If the **DEPENDS in add_custom_command** is not important here and we want to be compatible with **cmake 3.0.2** the easiest cure would be to go back to the **file(COPY...** command.


Pro:
Depends should be used here to only copy, when the files changed.

Either we stick to the `file(COPY` command, or we use `cmake -E copy` for every single file in docs. 

I have no strong opinion here, as long as it works :)

jrcoding:
if it won't do any harm, then please merge :-)

PR #2889, opened 2019-07-16
Allow abstract events

### LINKED ISSUE #2880: Server cannot create instance of abstract event, e.g., GeneralModelChangeEventType
## Description
My open62541-based server creates and removes nodes in the server AddressSpace at runtime. To notify the client side, I'd like to use the GeneralModelChangeEventType, but the creation of an event of that type fails. I'm using UaExpert V.1.5.1 on the client side.

Trying to create the event using UA_Server_createEvent fails because GeneralModelChangeEventType is abstract. I'd like to point out that the standard says, you are allowed to instantiate abstract EventTypes, as long as they don't appear in the AddressSpace. See OPC UA Part 3, chapter 4.6.2 EventTypes:
> EventTypes defined in this document are specified as abstract and therefore never instantiated in the AddressSpace. Event occurrences of those EventTypes are only exposed via a Subscription. EventTypes exist in the AddressSpace to allow Clients to discover the EventType. This information is used by a client when establishing and working with Event Subscriptions. EventTypes defined by other parts of this series of standards or companion specifications as well as Server specific EventTypes may be defined as not abstract and therefore instances of those EventTypes may be visible in the AddressSpace although Events of those EventTypes are also accessible via the Event Notification mechanisms.

Because UA_Server_createEvent tries to create a node with no parent node, I think this requirement is fulfilled.

Another reference going in that direction can be found in chapter 9.32.7 Guidelines for ModelChangeEvents:
>Two types of ModelChangeEvents are defined: the BaseModelChangeEvent that does not contain any information about the changes and the GeneralModelChangeEvent that identifies the changed Nodes via an array. The precision used depends on both the capability of the OPC UA Server and the nature of the update. An OPC UA Server may use either ModelChangeEvent type depending on circumstances. It may also define subtypes of these EventTypes adding additional information.

In other words: inheriting from GeneralModelChangeEvent is optional, I may also use GeneralModelChangeEvent directly.

I also tried this with the Unified Automation ANSI C Demo Server, which provides methods for dynamic node creation/deletion in its address space at Root\Objects\Demo\0008_DynamicNodes. Those methods instantiate GeneralModelChangeEvents directly.

So IMHO this is a bug, do you agree? Or is there a different way to create instances of those abstract event types?

Used CMake options:

```bash
cmake -DUA_ENABLE_AMALGAMATION=OFF -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_SUBSCRIPTIONS_EVENTS=ON
```
Code snippet:
```
UA_NodeId eventID;

// mServer points to a UA_Server object
UA_StatusCode result = UA_Server_createEvent(mServer, UA_NODEID_NUMERIC(0, UA_NS0ID_GENERALMODELCHANGEEVENTTYPE), &eventID);

// UA_STATUSCODE_BADTYPEDEFINITIONINVALID is returned
```

## Checklist
 - [x] open62541 Version: commit 21e32dd9973096133a2fcf826809518e5fd0bdfc
 - [x] Other OPC UA SDKs used (client or server): UaExpert V1.5.1
 - [x] Operating system: Linux
 - [ ] Critical issue

aDogCalledSpot:
Fixes #2880

### DIFF
--- a/src/client/ua_client_subscriptions.c
+++ b/src/client/ua_client_subscriptions.c
@@ -366,7 +366,14 @@ UA_Client_MonitoredItems_createEvent(UA_Client *client, UA_UInt32 subscriptionId
     UA_CreateMonitoredItemsResponse response = 
        UA_Client_MonitoredItems_createEvents(client, request, &context,
                                              &callback, &deleteCallback);
+    UA_StatusCode retval = response.responseHeader.serviceResult;
     UA_MonitoredItemCreateResult result;
+    UA_MonitoredItemCreateResult_init(&result);
+    if(retval != UA_STATUSCODE_GOOD) {
+        UA_CreateMonitoredItemsResponse_deleteMembers(&response);
+        result.statusCode = retval;
+        return result;
+    }
     UA_MonitoredItemCreateResult_copy(response.results , &result);
     UA_CreateMonitoredItemsResponse_deleteMembers(&response);
     return result;
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -729,13 +729,16 @@ AddNode_addRefs(UA_Server *server, UA_Session *session, const UA_NodeId *nodeId,
                 if(retval != UA_STATUSCODE_GOOD)
                     goto cleanup;
                 /* Object node created of an abstract ObjectType. Only allowed
-                 * if within BaseObjectType folder */
+                 * if within BaseObjectType folder or if it's an event (subType of BaseEventType) */
                 const UA_NodeId objectTypes = UA_NODEID_NUMERIC(0, UA_NS0ID_BASEOBJECTTYPE);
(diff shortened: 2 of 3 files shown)


inspektor-scherlein:
Hi @aDogCalledSpot,
first of all: thanks for taking care of this. With this fix, it's now possible to create Events with an abstract event type. But I'm pretty sure that's only allowed, if the resulting node has no parent (i.e. it is not visible in the server address space). Please see the corresponding sections in the specification I quoted in the [issue](https://github.com/open62541/open62541/issues/2880).

So the condition would probably have to look like:
```
if(!isInBaseObjectType && !(isInBaseEventType && UA_NodeId_isNull(parentNodeId)))
```
Do you agree?

Pro:
@inspektor-scherlein yep, this improves the proposed fix.

@aDogCalledSpot can you please update the PR and add a unit test which also checks:
- if event has parent and is abstract -> fail
- if event has no parent and is abstract -> success
- if event has parent and is not abstract -> success

PR #2908, opened 2019-07-23
Debian: Do not copy lib into open62541-dev to avoid overriding open62541 package

### LINKED ISSUE #2906: Install error with package manager
[error_log.txt](https://github.com/open62541/open62541/files/3421218/error_log.txt)
##  Description
Installing the packages from the daily build fails on (l)Ubuntu

sudo add-apt-repository ppa:open62541-team/daily
sudo apt-get update
sudo apt-get install open62541-dev

## Additional Notes
The package open62541 installed without problems. Based on the error log the fault is probably that the shared libraries are in both packages.

Pro:
Fixes #2906

### DIFF
--- a/debian/open62541-dev.install
+++ b/debian/open62541-dev.install
@@ -1,5 +1,5 @@
-usr/include/*
-usr/lib/*/lib{*.a,*.so,*.so.*}
+usr/include
+usr/lib/*/{*.a,*.so}
 usr/lib/*/cmake/open62541/*
 usr/lib/*/pkgconfig/open62541.pc
 usr/share/open62541/*
--- a/debian/open62541.install
+++ b/debian/open62541.install
@@ -1,1 +1,1 @@
-usr/lib/*/lib{*.a,*.so,*.so.*}
+usr/lib/*/*.so.*


jrcoding:
@Pro 
Hello, for my understanding to the open62541-dev package belongs every thing needed to compile and link and to the library package everything needed at execution time.

This means:

 **open62541-dev.install**
`usr/include`
`usr/lib/*/*.a`
`usr/lib/*/*.so`
`usr/lib/*/cmake/open62541/*`
`usr/lib/*/pkgconfig/open62541.pc`
`usr/share/open62541/*`


 **open62541.install**
`usr/lib/*/*.so.*`

Pro:
But then you can not install open62541 and open62541-dev at the same time, since you will get the same override error.
With the config in this PR, open62541-dev depends on open62541, this means when installing the dev package it will also install the lib package.

Isn't that the correct way?
(I'm not an Ubuntu packaging expert, happy for any feedback!)

jrcoding:
@Pro 
hmmm, please help, can't see any override ...

**open62541-dev** installs only `.a` and `.so` respectively `/usr/lib/x86_64-linux-gnu/libopen62541.a` and `/usr/lib/x86_64-linux-gnu/libopen62541.so`, **open62541** installs the remaining `.so.*` stuff respectively `/usr/lib/x86_64-linux-gnu/libopen62541.so.0` and `libopen62541.so.1.0.0` 

...am I wrong?

If you only install **open62541** you will see in the file system
`/usr/lib/x86_64-linux-gnu/libopen62541.so.0 -> libopen62541.so.1.0.0`
`/usr/lib/x86_64-linux-gnu/libopen62541.so.1.0.0`
and when you install **open62541-dev** you will see
`/usr/lib/x86_64-linux-gnu/libopen62541.a`
`/usr/lib/x86_64-linux-gnu/libopen62541.so -> libopen62541.so.0`
`/usr/lib/x86_64-linux-gnu/libopen62541.so.0 -> libopen62541.so.1.0.0`
`/usr/lib/x86_64-linux-gnu/libopen62541.so.1.0.0`
because it automatically also installs **open62541** because of the dependency on `libopen62541` defined for **open62541-dev** in the `debian/control` file

I use it this way with my local branch of open62541 and it works fine so far on Debian...

BTW: sorry that our proposal for a fine tuned debianization is not available up to now but my colleague, who is much more experienced than I regarding the Debian bells and whistles, is very busy at the moment. 
But our proposal will definitely come but it takes some more time, unfortunately.

Pro:
Oh, sorry, I misunderstood your previous post.

Can you clarify, why the `.so` entry should be in the `-dev` package and not in the main package?

That's just a link to the `.so.0` and should also be created for the lib package?

So I would go for:

 **open62541-dev.install**
`usr/include`
`usr/lib/*/*.a`
`usr/lib/*/cmake/open62541/*`
`usr/lib/*/pkgconfig/open62541.pc`
`usr/share/open62541/*`


 **open62541.install**
`usr/lib/*/lib{*.so,*.so.*}`

Pro:
Ah, I think I found the explanation. `.so` is only used by the linker/compiler. Therefore it is only required in the -dev package.

https://unix.stackexchange.com/a/10338

jrcoding:
> Ah, I think I found the explanation. `.so` is only used by the linker/compiler. Therefore it is only required in the -dev package.
> 
> https://unix.stackexchange.com/a/10338

that's it...

PR #2921, opened 2019-07-24
[CTT] Disable deprecated security policies & Remove not-supported nodes from namespace

opcua-tsn-team-kalycito:


src/server/ua_server_ns0.c
```diff
@@ -649,7 +649,30 @@ UA_Server_initNS0(UA_Server *server) {
 
 #ifdef UA_GENERATED_NAMESPACE_ZERO
     /* Load nodes and references generated from the XML ns0 definition */
-    retVal = namespace0_generated(server);
+    retVal |= namespace0_generated(server);
+
+    /* Remove unsupported nodes */
+    retVal |= UA_Server_deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_SETSUBSCRIPTIONDURABLE), true);
```
Pro: Why do we need to delete the nodes here?
What does `unsupported` mean?

Are these only unsupported by the CTT, but in general still valid nodes?
Or are these deprecated nodes?

Can you clarify that in the comment please.

I think we should ignore the return value for the deleteNode calls if it is `UA_STATUSCODE_BADNODEIDUNKNOWN`.

Especially if the nodes are deprecated, then future nodesets may just omit them and this code will fail.

jpfr:
Some of the nodes for server capabilities in the full ns0 contain a null value.
They are unused atm.
We have to either link them to the actual server config or remove them.
We do the latter as a quick fix for now.

You are right. The return value should be ignored.
Otherwise the minimal ns0 will fail.

Pro:
@jpfr please do not force merge if unit tests fail! Otherwise we may quickly run into bigger issues. Now master is broken...

PR #2945, opened 2019-07-31
Remove garbage value asssignment for target variables in Subscriber

opcua-tsn-team-kalycito:
- Remove the garbage value assignment for variables created using UA_Server_DataSetReader_addTargetVariables | Values will be assigned as NULL by default
 - Unit test cases to verify the published and subscribed data for the following datatypes
    - Int32
    - Int64
    - Boolean


jpfr:
Very nice. 👍 
The code was simplified quite a bit.

PR #2955, opened 2019-08-01
[Review] Modify Send-/Recv-buffer handling + MaxMsgSize

mlang-de:
> UA Spec. Part 6, 1.04, chapter 7.1.3: Establishing a connection describes the negotiation and does not require any error message to be returned by the Server if it can't deal with the Client buffer sizes it responds with what it can and the Client decides.

See https://opcfoundation.org/forum/opc-ua-standard/client-send-maxmessagesize-smaller-then-defined-min-buffer-size-of-8192/.

There is no lower limit for the MaxMessageSize. The current implemented 1460 in this pull request should be discussed. 
OPCF Forum post:

> 
> MaxMessageSize could be smaller than the ReceiveBufferSize/SendBufferSize because it only constrains the unencrypted message body.The signature and MessageChunk headers are not included.
> 
> This also means the 8192 byte limit does not apply to the MaxMessageSize  - the limit only applies to the ReceiveBufferSize/SendBufferSize.

I'm not sure which minimum MaxMessageSize make sense.

This PullRequest finding was reported by a commercial test platform.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -17,8 +17,6 @@
 #define UA_SESSION_LOCALNONCELENGTH 32
 #define MAX_DATA_SIZE 4096
 
-static void closeSession(UA_Client *client);
-static UA_StatusCode createSessionAsync(UA_Client *client);
 
 static UA_SecurityPolicy *
 getSecurityPolicy(UA_Client *client, UA_String policyUri) {
--- a/src/server/.#ua_server_internal.h
+++ b/src/server/.#ua_server_internal.h
@@ -1,0 +1,1 @@
+jpfr@virtualbox.698:1568887719
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -285,6 +285,10 @@ processHEL(UA_Server *server, UA_SecureChannel *channel, const UA_ByteString *ms
     /* Currently not checked */
     UA_String_clear(&helloMessage.endpointUrl);
 
+    /* Parameterize the connection. The TcpHelloMessage casts to a
+     * TcpAcknowledgeMessage. */
+    UA_SecureChannel_processHELACK(channel, (UA_TcpAcknowledgeMessage*)&helloMessage);
+
     /* Get the send buffer from the network layer */
     UA_Connection *connection = channel->connection;
     UA_ByteString ack_msg;
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -285,17 +285,6 @@ processHEL(UA_Server *server, UA_SecureChannel *channel, const UA_ByteString *ms
     /* Currently not checked */
     UA_String_clear(&helloMessage.endpointUrl);
 
-    /* Parameterize the connection. The TcpHelloMessage casts to a
-     * TcpAcknowledgeMessage. */
-    retval = UA_SecureChannel_processHELACK(channel,
-                                            (UA_TcpAcknowledgeMessage*)&helloMessage);
-    if(retval != UA_STATUSCODE_GOOD) {
-        UA_LOG_INFO(&server->config.logger, UA_LOGCATEGORY_NETWORK,
-                    "Connection %i | Error during the HEL/ACK handshake",
-                    (int)(channel->connection->sockfd));
-        return retval;
-    }
-
     /* Get the send buffer from the network layer */
     UA_Connection *connection = channel->connection;
     UA_ByteString ack_msg;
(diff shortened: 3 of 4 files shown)


jpfr:
Hello @mlang-de ,

that's a good improvement. Thanks.

Where does 1460 come from?
OpenSecureChannel with a Certificate?

I would have expected a power of two...

mlang-de:
1460 is the maximum TCP-Payload in one full frame.

I'm not sure how meaningful it is regarding security.
If a certificate has 4096 Byte + "X" Byte for the Header, this should be the minimum MaxMessageSize?
But if I look into Part 6 Figure 11 in combination with 

> MessageSize is the maximum length of an unencrypted Message that can be split into multiple MessageChunks.

it is the "Data To Encrypt" part which defines the minimum MaxMessageSize.

The certificate does not apply to the MaxMessageSize?

jpfr:
> 1460 is the maximum TCP-Payload in one full frame.

Oh cool. That is a very nice constant to use then.

> MessageSize is the maximum length of an unencrypted Message that can be split into multiple MessageChunks.

The OPN messages (to open a SecureChannel) cannot be chunked.
So the certificate needs to fit in one chunk.

Of course, certificates can have basically arbitrary size.
We should give recommendations for developers that fit most cases.

mlang-de:
If I'm correct the code must be extenende with a greater then comparison to protect the server itself:

```c
    if(connection->config.maxMessageSize != 0)
    {
        if(connection->config.maxMessageSize < 1460)
            connection->config.maxMessageSize = 1460;
        else if (connection->config.maxMessageSize > localConfig->maxMessageSize)
            connection->config.maxMessageSize = localConfig->maxMessageSize;
    }

Should the 1460 be moved to a define?

> We should give recommendations for developers that fit most cases.

At the moment the default configurations are "0" which means unlimited. This limit is very use case depending. Without file transfer the maxMessageSize could be much smaller then vice versa.
Which recommendation you think about?

mlgiraud:
IMHO we shouldn't impose a restriction on the `maxMessageSize` because the standard doesn't specify anything regarding the size, except for 0 being unlimited.
We can recommend some values in the documentation, but it should be left to the developer to configure the value to something sensible.

mlang-de:
@mlgiraud I agree with you for maxMessageSize and its upper level. This is not limited in the current PullRequest. What is your opinion in the lower level? Accept all? Even maxMessageSize = 1? This lower limit is not configureable yet.

mlgiraud:
What i meant to say was, that we shouldn't restrict the value at all, since the standard doesn't state anything about that. We could maybe print a warning if the value is less than the one you suggested.

mlang-de:
This is a good idea, but unfortuantly in file ua_connection.c the logger is not available yet.
What is the preferred solution? Extend the function parameter from UA_Connection_processHELACK or handled the warning print out in the server function processHEL?

mlang-de:
For me a send/recv buffer from at least 32kB works for me in relation to CTT.

jpfr:
Very nice!
Squashed and added a new commit since processHELACK can no longer return an error code.

jpfr:
Hum. With this patch, the CTT fail a lof of cases.

The server gives `Processing the message failed with error BadResponseTooLarge`.

mlang-de:
I fixed the maxMessageSize negotiation. The CTT fails was triggered by using maxMessageSize = 0 as connection configuration.
For me the CTT runs goods except 1 script with the (8 days old) master.

mlang-de:
@jpfr I am confused. Where is the server function UA_SecureChannel_processHELACK or its equivalent? In function processHEL() the channel requested client buffer sizes will be assigned without any check? Please advice.

mlgiraud:
I'm not sure why you are confused. Is this not what you are looking for? https://github.com/open62541/open62541/blob/ba5cf14f9c39a48fc2cc0fd9a129399b240a3b26/src/ua_securechannel.c#L136
This function is called on server and client side.

mlang-de:
You are right. It seems that my branch is broken. Thank you.

How shall this work, that client and server uses the same logic?
IMHO the client send the HEL-Msg with desired Buffersizes and MaxMessageSize.
The server retrieves these values and calls UA_Connection_processHELACK. In this function the server checks the values and maybe revise these and send it back to the client. From the server point of view this function never will fail, because the client shall decide if the revised value are ok for the client side. 
Since this is specified it is IMHO not possible to reuse the function for client and server side.

See https://opcfoundation.org/forum/opc-ua-standard/client-send-maxmessagesize-smaller-then-defined-min-buffer-size-of-8192/

`7.1.3 Establishing a connection describes the negotiation and does not require any error message to be returned by the Server if it can't deal with the Client buffer sizes it responds with what it can and the Client decides.`

mlang-de:
I am to silly for git. If we have decide what client and server have to do I will create a new PR from the current master.

mlang-de:
See PR https://github.com/open62541/open62541/pull/3975

PR #2960, opened 2019-08-02
docs: Update contributing guidelines for conventional commits

Pro:
This PR updates the contributing guidelines to clarify the commit message structure.

It recommends to stick with https://www.conventionalcommits.org

This allows us to automatically generate the changelog and in the future use semantic releases more easily.

To check if at least the PR title or any commit inside the PR conforms with the guideline, I just enabled https://github.com/probot/semantic-pull-requests

This will add another CI status. Currently it is not required, but I suggest to set it as required after some time testing it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,97 @@ The following are the minimal requirements that every PR needs to meet.
   link automatically appears on the page of the first PR. In addition, the CLA
   text can be accessed [here](https://cla-assistant.io/open62541/open62541).
 
-### Commit and PR Hygiene
+## Commit and PR Hygiene
+
+We have very precise rules over how our git commit messages can be formatted.  This leads to **more
+readable messages** that are easy to follow when looking through the **project history**.  But also,
+we use the git commit messages to **generate the change log**.
+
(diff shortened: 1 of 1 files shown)


Pro:
@jpfr if you agree using conventional commits format, we can merge this.

One of the main advantages is, that we can later on automatically generate changelogs, and if desired we can also switch to automatic semantic releases.

On the newest PRs the Semantic Pull Request checker is already enabled.

jpfr:
Actually, I think it is not the best to have automatic releases for every (feat) and (BREAKING CHANGE) commit. We need a few weeks of testing and release candidates before a truly **stable** release.

Otherwise we need to be a lot stricter on the master branch.

Also, since the project growse a lot, we also do refactoring and code cleanups.
Those don't fit into the nomenclature...

Pro:
@jpfr agree, that automatic releases do not make much sense at the current stage.

Anyways, should we try to stick witch conventional commit messages? These seem to be well accepted by the OSS community. Then I would merge this PR which extends the docs on the commit message format.

We do not need to force new contributors to use these, but still should provide a corresponding contribution guideline.

PR #2971, opened 2019-08-06
[REVIEW] Provide Datatypedefinition for structures on read request

matkonnerth:
With this PR the read request for Attribute DataTypeDefinition is handled and the StructureDefinition is calculated.


jpfr:
Oh nice!

Small comments:
- Keep the switch-case short. You can break out another function
- I did not understand the behaviour of getDataTypeIdFromIndex for the case of custom types.
With custom types, the index either refers to the ns0 types or to the same "custom array".
See how the type is looked up for members here: https://github.com/open62541/open62541/blob/master/src/ua_types.c#L971
- Use UA_calloc for def.fields. So you don't have to multiply size_t (can overflow in general, so we avoid it everywhere).
- Don't UA_STRING_ALLOC the type name. Just use UA_STRING. So you don't have to clean up the heap at the end. We know the memory access is safe. You can cast via (char*)(uintptr_t) to get the const out.

We can merge once this is done and the history squashed.
Keep up the good work!

matkonnerth:
Thank you for the comments, I appreciate the hint with the custom types.

For a successful build UA_NAMESPACE_ZERO=Minimal or Reduced it would be necessary to add StructureDefinition, SructureField and StructureType (maybe later on the same for enums) to tools/schema/datatypes_minimal.txt. Or should I add a compile switch to the case UA_DatypeDefinition? I would simply add it, but I'm not in a environment with ram/rom pressure put on it ;)

jpfr:
For devices with just a few kb we can remove the type names from the struct.

> UA_ENABLE_TYPENAMES
https://github.com/open62541/open62541/blob/master/CMakeLists.txt#L267

Of course, we need the type (and member) names for this feature to make sense.

We could rename the build flag to UA_ENABLE_TYPEDESCRIPTION and bundle it with this feature.

jpfr:
And have a separate datatypes_typedescription.txt that is added in CMake if the feature is enabled...

matkonnerth:
to be merged when CI is green

PR #2973, opened 2019-08-07
Update pubsub_udp.h

chuongvhn:
Fix compile error due to missing ";"


jpfr:
The line was correct.
Ua_export refers to the line below.

PR #2982, opened 2019-08-08
Fix UA_ClientConnectionTCP_init

melkorb:
memset(tcpClientConnection, 0, sizeof(TCPClientConnection)) added to initialize TCPClientConnection struct members. If TCPClientConnection members not initialized and UA_parseEndpointUrl return error ClientNetworkLayerTCP_free will raise exception.


jpfr:
Thanks for the fix. 👍

PR #2983, opened 2019-08-09
[PubSub] Configuration freeze functionality and PubSub states

andreasebner:
This PR introduces a PubSub configuration freeze concept. This new feature will be used for the ongoing PubSub RT implementation to enable the buffering of pregenerated NetworkMessages and DataSetMessages. Since changes in the PubSub configuration will invalidate the buffered frames, the pubsub configuration can be frozen after the configuration phase. 

**This PR changes also the behavior after adding a new Writer Group. Currently the publish of DataSets is started automatically and the WriterGroup state is set to 'OPERATIONAL'. With the latest changes the default state is 'DISABLED' and the WriterGroup won't publish packages until the new setWriterGroupOperational function is called. These change might cause in problems with existing code since the publishing of packages isn't started automatically.**

The configuration freeze implementation uses the standard defined PubSub state machine (1.0.4 - 6.2.1). This PR contains an first implementation to realize the states and transitions.

API changes (added):
The freeze is always executed on the WriterGroup. This function can be used for lock and unlock the entire WG-linked PubSub configuration (PubSubConnection, PublishedDataSet, DataSetFields, and the DataSetWriters. 
_**UA_StatusCode UA_Server_freezeWriterGroupConfiguration(UA_Server *server, const UA_NodeId writerGroup);**_
_**UA_StatusCode UA_Server_unfreezeWriterGroupConfiguration(UA_Server *server, const UA_NodeId writerGroup);
UA_StatusCode UA_Server_setWriterGroupOperational(UA_Server *server, const UA_NodeId writerGroup);
UA_StatusCode UA_Server_setWriterGroupDisabled(UA_Server *server, const UA_NodeId writerGroup);**_


src/pubsub/ua_pubsub.h
```diff
@@ -86,6 +89,26 @@ typedef struct UA_DataSetWriterSample{
...
 
+typedef enum {
+    UA_PUBSUB_DATASETMESSAGE_TIMESTAMP,
+    UA_PUBSUB_DATASETMESSAGE_WRITERID,
+    UA_PUBSUB_DATASETMESSAGE_PAYLOAD
+    /* TODO add other defined DSM contents */
+} UA_DataSetMessageField;
+
+typedef struct UA_DataSetMessageOffsetDescription{
+    UA_DataSetMessageField *field;
```
jpfr: Why is this a pointer?

src/pubsub/ua_pubsub.h
```diff
@@ -86,6 +89,26 @@ typedef struct UA_DataSetWriterSample{
 } UA_DataSetWriterSample;
 #endif
 
+typedef enum {
+    UA_PUBSUB_DATASETMESSAGE_TIMESTAMP,
+    UA_PUBSUB_DATASETMESSAGE_WRITERID,
```
jpfr: The writer id will not change once the WriterGroup is "frozen".
Why would we ever have to change it?

src/pubsub/ua_pubsub.h
```diff
@@ -86,6 +89,26 @@ typedef struct UA_DataSetWriterSample{
...
+    UA_PUBSUB_DATASETMESSAGE_PAYLOAD
+    /* TODO add other defined DSM contents */
+} UA_DataSetMessageField;
+
+typedef struct UA_DataSetMessageOffsetDescription{
+    UA_DataSetMessageField *field;
+    UA_UInt16 fieldSize;
```
jpfr: This should not be the field `size`, but the `offset` starting at the beginning of the NetworkMessage.
Because there are some parts of the message (e.g. bitfield headers) we won't touch and don't see here.
So we cannot compute the start position of the fields only from their length.

src/pubsub/ua_pubsub.h
andreasebner: Yes, you are right. The code to generate the offset structures are currently not implemented and i tried to use well known header fields for example.

src/pubsub/ua_pubsub.h
andreasebner: This was an mistake. I will fix it.

src/pubsub/ua_pubsub.h
andreasebner: Will be pushed with the next commit.

Issue #2923, opened 2019-07-25
Semantic release numbers


jrcoding:
## Description
At the moment the **version of the open62541 shared libraries** respectively the **SONAME** will be defined in `CMake.txt` by 

`# Generate properly versioned shared library links on Linux
SET_TARGET_PROPERTIES(open62541 PROPERTIES SOVERSION 0 VERSION "${OPEN62541_VER_MAJOR}.${OPEN62541_VER_MINOR}.${OPEN62541_VER_PATCH}")
`

In this case the SONAME of the produced shared library will be generated as `libopen62541.so.0`

AFAIK the SONAME must change if the ABI and/or API changes to be sure that applications must be compiled and linked new because of incompatible changes in the library.

To organize this a solution could be the so called **Semantic Versioning** (amongst others explained @ https://semver.org/)
>
> Given a version number MAJOR.MINOR.PATCH, increment the:
>
>    MAJOR version when you make incompatible API changes,
>    MINOR version when you add functionality in a backwards-compatible manner, and
>    PATCH version when you make backwards-compatible bug fixes.

As far as I see at the moment the shared library API and/or ABI changes already if the \<minor_number\> changes,  _am I right?_

## Question:
Does it make sense and is it reasonable to ensure being compatible with **Semantic Versioning** (beginning with version 1.0.0) ?

## Remark:
This is also important for the scheme how to build  debian packages and how to deal with shared objects on Linux

## Checklist

Please provide the following information:

 - [x] open62541 Version (1.0):
 - [x] Operating system: Linux (all flavours)
 - [x] Critical issue

Pro:
> Does it make sense and is it reasonable to ensure being compatible with **Semantic Versioning** (beginning with version 1.0.0) ?

Short answer: Yes

Long answer: I'm definitely in favour of semantic versioning. We need to find a good way to keep the Changelog updated and document all the major and minor changes.

It would be best if there are some automatic tools/bots which help us with that and check pull requests correspondingly.

I don't mind if we reach version 10.0 in a year, honestly do not care that much about version numbers, as long as they make sense = stick to semver.

@jpfr maybe the 1.0 release is a good starting point to enforce semver?

Pro:
Maybe integrating https://github.com/semantic-release/semantic-release would be a good idea?

jrcoding:
@ralphlange @Pro @jpfr 

[...]

PR #2985, opened 2019-08-09
feat: [Debian] libname fitting debian policy

jrcoding:
This patch sets `SOVERSION` of the library equal to `OPEN62541_VER_MAJOR` and in addition names the library like `libopen62541-<SOVERSION>` i.e. `libopen62541-1` compliant with the Debian policy.

This refers to [the comment from @ralphlange ](https://github.com/open62541/open62541/pull/2367#issuecomment-454344066) and the [Debian policy, chapter "Shared Libraries"](https://www.debian.org/doc/debian-policy/ch-sharedlibs.html).

Of course _"this is not a must"_ but it's more compliant to the Debian policy.


CMakeLists.txt
```diff
@@ -1057,7 +1059,7 @@ if (UA_ENABLE_DISCOVERY_MULTICAST)
     target_compile_definitions(open62541 PRIVATE -DMDNSD_DYNAMIC_LINKING_EXPORT)
 endif()
 # Generate properly versioned shared library links on Linux
-SET_TARGET_PROPERTIES(open62541 PROPERTIES SOVERSION 0 VERSION "${OPEN62541_VER_MAJOR}.${OPEN62541_VER_MINOR}.${OPEN62541_VER_PATCH}")
+SET_TARGET_PROPERTIES(open62541 PROPERTIES SOVERSION "${SONAME}" VERSION "${OPEN62541_VER_MAJOR}.${OPEN62541_VER_MINOR}.${OPEN62541_VER_PATCH}")
```
Pro: Can't we use `${OPEN62541_VER_MAJOR}` directly in here, instead of creating an additional `SONAME` variable?

debian/control
```diff
@@ -20,7 +20,7 @@ Depends: open62541 (= ${binary:Version}), ${misc:Depends}
 Description: Development header files for open62541
  open62541 is an open source C (C99) implementation of the OPC UA standard
 
-Package: open62541
+Package: libopen62541-1
```
Pro: This should be automated. If we change the major version, that part should be updated correspondingly.
We should try to integrate this into the python script from PR #2980
https://github.com/open62541/open62541/pull/2980/files#diff-c83ebe4f93ca1086557dbd0e52e2ae3c

Otherwise creating new releases will be error prone and one might forget to update the version here.

CMakeLists.txt
jrcoding: > Can't we use `${OPEN62541_VER_MAJOR}` directly in here, instead of creating an additional `SONAME` variable?

yes, of course, we can do it...

BTW: 
I think we should agree to change `${OPEN62541_VER_MAJOR}` whenever the ABI and/or API changes to be sure having a valid `SONAME` , it's the #2923 issue

debian/control
jrcoding: > This should be automated. If we change the major version, that part should be updated correspondingly.
> We should try to integrate this into the python script from PR #2980
> https://github.com/open62541/open62541/pull/2980/files#diff-c83ebe4f93ca1086557dbd0e52e2ae3c
> 
> Otherwise creating new releases will be error prone and one might forget to update the version here.
> 
> Maybe we can have a `debian/control.template` which is then read by the python script and updated correspondingly. That script then also renames the file `debian/libopen62541.template` correspondingly.

Yes, I agree, would be nice to have!

AFAIK this unfortunately doesn't work building the package the debian way (i.e. with pbuilder etc.) because  `debian/control` and `debian/changelog` will be referenced before `cmake` will be called!?
But for sure we can use a template for the lib-install-file  `debian/lib62541-<soname>.install` !

Pro: Hmm, yes, right... For the launchpad builds we may need to update the recipe to execute the python script, if that's even possible. Let's merge #2980 first, then we can have a follow-up fix which includes this PR with the comments

jrcoding: > > This should be automated. If we change the major version, that part should be updated correspondingly.
> > We should try to integrate this into the python script from PR #2980
> > https://github.com/open62541/open62541/pull/2980/files#diff-c83ebe4f93ca1086557dbd0e52e2ae3c
> > Otherwise creating new releases will be error prone and one might forget to update the version here.
> > Maybe we can have a `debian/control.template` which is then read by the python script and updated correspondingly. That script then also renames the file `debian/libopen62541.template` correspondingly.
> 
> Yes, I agree, would be nice to have!
> 
> AFAIK this unfortunately doesn't work building the package the debian way (i.e. with pbuilder etc.) because `debian/control` and `debian/changelog` will be referenced before `cmake` will be called!?
> But for sure we can use a template for the lib-install-file `debian/lib62541-<soname>.install` !

BTW: the nice side effect using a template for `debian/lib62541-<soname>.install` is that an error occurs during debian build if the `debian/control` doesn't correspond to the generated `debian/lib62541-<soname>.install`. This solves the problem if one forget  to update `debian/changelog` and `debian/control` in some way when the major version changes. Unfortunately AFAIK one has to correct `debian/changelog` and `debian/control` manually.

Pro:
This is now part of #2987 
Thank you for the proposed fix! Unfortunately I had to rewrite the commits a bit more therefore I couldn't include your commit directly

Issue #1910, opened 2018-07-10
CI for freeRTOS


Pro:
After the Architecture PR #1595 we need Continuous Integration for FreeRTOS+LWIP

cabralfortiss:
Here a comment about this issue. Just to organize

https://github.com/open62541/open62541/pull/2511#issuecomment-500227225

Pro:
@cmbahadir in https://github.com/open62541/open62541/pull/2511#issuecomment-500227225 you note some infos for integrating freeRTOS in Travis CI.

To comment on the necessary changes:

> * Location of the "task.h" should be "freertos/task.h" on open62541.c.

Is it necessary to change this in the file. Can't you just set the additional include directory for the compiler?

> * #define UA_ARCHITECTURE_FREERTOSLWIP has to be defined under open62541.c

This should be done through compiler flags, i.e., `gcc -DUA_ARCHITECTURE_FREERTOSLWIP ...`

> * And also #define UA_IPV6 LWIP_IPV6 should be commented out since IPV6 is not supported.

For this there should be an additional ifdef for the FreeRTOS architecture.
(@cabralfortiss ping)

Based on my comments it shoudl not be necessary to use sed or other hacks to modify the code. Just by parametrizing the compiler correctly, everything should work.

@cmbahadir would you like to propose a PR which adds freeRTOS builds to travis?
We will of course support you!

cmbahadir:
Hey @Pro ,

Ok i will check it, but it may take some time since currently ı am working on something else.

By the way it will be a CI with a xtensa-toolchain which is specific to espressif(ESP32) but it still uses freeRTOS-LWIP so ı guess it is not a major problem to test the integration.

PR #2993, opened 2019-08-13
feat: Lock-based enhancement using mutexes

ckmk14:
This PR implements a first approach to enable the parallel access of the public server api. 
A lock-based enhancement using mutexes is established to avoid race-conditions when accessing data structures in parallel threads.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,12 +192,12 @@ if(UA_ENABLE_DISCOVERY_MULTICAST AND NOT UA_ENABLE_DISCOVERY)
 endif()
 
 # Advanced options
-option(UA_ENABLE_MULTITHREADING "Enable multithreading (EXPERIMENTAL)" OFF)
-mark_as_advanced(UA_ENABLE_MULTITHREADING)
-
 option(UA_ENABLE_IMMUTABLE_NODES "Nodes in the information model are not edited but copied and replaced" OFF)
 mark_as_advanced(UA_ENABLE_IMMUTABLE_NODES)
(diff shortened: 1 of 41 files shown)


jpfr:
Looks really good!
One minor wish: We should document what the levels 0/100/200 mean.
In the .rst docs and very short also in the CMake variable description.

@Pro This build is failing on Azure.
Is this a general problem or only concerning this PR?
Looks like a general problem to me...

Pro:
Yeah, just saw the error today. Will fix it today evening. It's a general caching issue

jpfr:
Very nice! Merged.
We can remove the EXPERIMENTAL flag to MULTITHREADING=100 once the server API is covered.

cmbahadir:
Hello All,

This merge breaks freertos build pipeline of me.

https://travis-ci.org/cmbahadir/opcua-esp32/builds/560100952

Seems like UA_MULTITHREAD is not enabled for other architectures as long as it is set as 100.

[ua_architecture.h](https://github.com/ckmk14/open62541/blob/cada90600c7766657ba351ce11e864f43df608d7/arch/posix/ua_architecture.h#L150)

ckmk14:
Hello @cmbahadir 

you are absolutely right. The macros are not defined in all the different architecture.h files. Freertos is one of them. I will add the missing definitions.

Pro:
@cmbahadir would you be able to provide a PR which adds FreeRTOS to out build pipeline?
Then we can avoid issues like this in the future.
See also #1910

cmbahadir:
Hi @Pro created PR #3023

PR #2994, opened 2019-08-13
[review] Unify license headers

ichrispa:
There are some leftover antiquated licenses floating around in my python scripts. Some still contain revision information from my diplima thesis' svn repo... I think we should get rid of these and clearly mark the files as MPL-2.0 to comply /w the rest of the project. 

Reported by Rahul Mohan to me via mail.


jpfr:
Nice. This is much better!

PR #2999, opened 2019-08-14
ci: Migrate from Appveyor to Azure Pipelines

Pro:
This moves most of the builds from Appveyor to the more efficient
and significantly faster Azure Pipelines system.

**We will get much faster Windows builds**

For now Azure Pipelines by default only supports VS2015 and VS2017,
therefore the builds for VS2013 and VS2008 still remain on Appveyor.

These can be transferred at a later stage.


jpfr:
Yay! Very cool.

PR #3014, opened 2019-08-19
[Review] New information model callbacks

inspektor-scherlein:
Hello everyone,
I would like to propose the introduction of two additional information model callbacks, located in UA_GlobalNodeLifecycle. The motivation for those two are as follows:

## onDefineNodeId
During node creation, a recursive check if all mandatory children exist is performed. If a mandatory child is missing, it is created with a random numeric node id. Currently, the only way to influence those node ids is to manually create the whole subtree of a node and define each node id before UA_Server_addNode_finish is called. Having control over the created node ids might be very helpful, especially when the server uses string-based node ids that follow a certain pattern. So this callback is called just before a random node id would be created. It allows to define a custom node id instead.

## onCreateOptionalChild
This one goes into the same direction. Mandatory children are created recursively during node instantiation. Optional children are ignored. But as there already is a recursive check, I felt that a callback with the semantics "Hey, I've got an optional child here, should I instantiate it?" would come in quite handy. This makes node instantiation a lot easier in dynamic address spaces.

If those callbacks are not used, the behavior remains the same as before. If they are, they make the lives of those easier, who are dealing with dynamic address spaces. Also, having a mechanism to influence node id creation makes it easier to fulfil the requirement, that node ids should not change after a server restart. 

Thanks in advance for reviewing this PR, I hope it will make its way.


include/open62541/server.h
```diff
@@ -670,7 +670,47 @@ typedef struct {
...
+     * @param server The server executing the callback
+     * @param sessionId The identifier of the session
+     * @param sessionContext Additional data attached to the session in the
+     *        access control layer
+     * @param sourceNodeId Source node from the type definition. If the new node
+     *        shall be created, it will be a copy of this node.
+     * @param targetParentNodeId Parent of the potential new child node
+     * @param referenceTypeId Identifies the reference type which that the parent
+     *        node has to the new node. */
+    UA_Boolean (*onCreateOptionalChild)(UA_Server *server,
```
jpfr: Very nice. The on-prefix can be omitted for consistency with the other lifecycle method.
So `createOptionalChild`.

include/open62541/server.h
```diff
@@ -670,7 +670,47 @@ typedef struct {
...
+     *
+     * @param server The server executing the callback
+     * @param sessionId The identifier of the session
+     * @param sessionContext Additional data attached to the session in the
+     *        access control layer
+     * @param sourceNodeId Source node of the copy operation
+     * @param targetParentNodeId Parent node of the new node
+     * @param referenceTypeId Identifies the reference type which that the parent
+     *        node has to the new node. */
+    void (*onDefineNodeId)(UA_Server *server,
```
jpfr: Again, the on-prefix can be omitted. I'd prefer something like `generateChildNodeId`.

jpfr: Should return an UA_StatusCode.
It is possible that malloc fails if we want a string NodeId.

include/open62541/server.h
```diff
@@ -670,7 +670,47 @@ typedef struct {
...
+    UA_Boolean (*onCreateOptionalChild)(UA_Server *server,
+                                        const UA_NodeId *sessionId,
+                                        void *sessionContext,
+                                        const UA_NodeId *sourceNodeId,
+                                        const UA_NodeId *targetParentNodeId,
+                                        const UA_NodeId *referenceTypeId);
+
+    /* Can be NULL. Called when a node is to be copied during recursive
+     * node instantiation. Allows definition of the NodeId for the new node.
+     * If this callback is not used, random numeric NodeIds will be generated.
```
jpfr: We can have the callback **and** allow random NodeId generation.
Rephrase as follows:
If the callback is set to NULL or the resulting NodeId is UA_NODEID_NULL, then a random NodeId will be generated.

src/server/ua_services_nodemanagement.c
```diff
@@ -477,6 +489,18 @@ copyChild(UA_Server *server, UA_Session *session, const UA_NodeId *destinationNo
...
 
+        if (server->config.nodeLifecycle.generateChildNodeId) {
+            retval = server->config.nodeLifecycle.generateChildNodeId(server,
+                                                                      &session->sessionId, session->sessionHandle,
+                                                                      &rd->nodeId.nodeId,
+                                                                      destinationNodeId,
+                                                                      &rd->referenceTypeId,
+                                                                      &node->nodeId);
+            if(retval != UA_STATUSCODE_GOOD)
+                return retval;
```
jpfr: Now you have to delete the node, if you return here.
Otherwise we lose memory!

inspektor-scherlein: Oops. Sorry, I'll have a look.

inspektor-scherlein:
I think it should be OK now. Please let me know if any further changes are required.

jpfr:
The automated builds are still failing...

jpfr:
Perfect now. Merged.

Thanks a lot @inspektor-scherlein !

inspektor-scherlein:
My pleasure.

PR #3019, opened 2019-08-21
CMake: remove needless x${VAR} tricks

DerDakon:


Pro:
We introduced them because of this:
https://github.com/google/nsync/blob/master/CMakeLists.txt#L75

Any reason why we should remove it now?

DerDakon:
That's why I removed the variable expansion on the other side:

```
if ("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
if (CMAKE_C_COMPILER_ID STREQUAL "MSVC")
```

The former could lead to double evaluation, the latter will not. So the "bad" list of CMake versions includes those up to and including 3.0, which is sadly a supported version. Requiring at least 3.1 would solve this, as well as allowing to use CMAKE_C_STANDARD (see #3021).

The relevant thing is [policy CMP0054](https://cmake.org/cmake/help/v3.1/policy/CMP0054.html).

jpfr:
Good initiative.
Ubuntu, Debian and Fedora are all good.
Only the current CentOS is still on CMake 2.8. This will change in the next months.

This is mainly cosmetics. Let's freeze and merge once all major distros have CMake 3.1 in their stable release.

DerDakon:
If CentOS is on 2.8 then you can't build there anyway as the minimum required version already is 3.0.

jpfr:
I think we can increase the requirement to make 3.1 now.

@DerDakon would you rebase your change?

DerDakon:
Done. The Azure errors look unrelated.

PR #2511, opened 2019-02-28
WIP: Cleanup of freeRTOS and LwIP

cabralfortiss:
I want to use this PR to fix all issues related to freeRTOS and LwIP: 
fixes #1859 
fixes #1864 
fixes #2227 
fixes #2460. 

Users working with freeRTOS that confirmed the PR works:

- [X] @embbo (https://github.com/open62541/open62541/pull/2511#issuecomment-475769413)
- [ ] @nidhal-mars (not active in the PR)
- [ ] @FishEatingCat (not active in the PR)
- [X] @VladBraz125 (https://github.com/open62541/open62541/pull/2511#issuecomment-479844345)
- [X] @rudoq32 (https://github.com/open62541/open62541/pull/2511#issuecomment-474811430)
- [ ] @sammer-h (not active in the PR)

@Pro I'll wait to see if people have problems with this patch before thinking in merging

-----------------

I took out all the time related functions that are not part of freeRTOS but from linux. The server is always in 1970 but worked for me. 

I also removed unused includes that were messing compilations. 

Regarding LwIP I took all defines out, but let an explanation of what is needed.

Some info:

How to compile? THere are two ways: 
- Select the cross compiler in CMake, set the flags needed for compilation (can be difficult) and then run make in the folder and the library should be generated. This method can be hard to do because you need to specify the include files and some other configurations. 
 - generate the open6254.h and open6254.c files with the freeRTOSLwIP architecure and then put these files in your project in your IDE that you're using for compiling. This is the best easiest way of doing it. I'll focus on this.

This documentation assumes that you have a basic example using LwIP and freeRTOS that works fine, and you only want to add an opcua task to it.

So basically, pull this branch. In CMake select freertosLwIP as architecture and just select the native compilers. Now try to compile as always. Compilation will fail, but the open62541.h and open62541.c will be generated. 

Import these files in your IDE that you're using. Add the open62541.c file for compilation in the IDE, and add the variable UA_ARCHITECTURE_FREERTOSLWIP to the compilation. 

When compiling LwIP you need a file called lwipopts.h. Here you put all configuration variables for compiling time. You need to make sure that you have the following configurations there:

 * #define LWIP_COMPAT_SOCKETS 0 // Don't do name define-transformation in networking function names.
 * #define LWIP_SOCKET 1 // Enable Socket API (normally already set)
 * #define LWIP_DNS 1 // enable the lwip_getaddrinfo function, struct addrinfo and more.
[...]

Issue #2893, opened 2019-07-18
Compile freeRTOS with open62541 errors


mrburen121:
## Description
After creating the amalgamation header/source files and included them in my project I get some compiling errors. Did I forgot to include some parameters/files?

Language: -std=gnu99 (-std=c99 is also not working)
Compiler flags: -c -fmessage-length=0 -fno-builtin -ffunction-sections -fdata-sections -UA_ARCHITECTURE_FREERTOSLWIP

```
In file included from ../src/open62541.c:28:
../src/open62541.c: In function 'UA_PubSubChannelUDPMC_open':
/home/chris/GitHub/OPCUA-RTOSServer/lwip_freertos_tcpecho/inc/open62541.h:284:27: warning: implicit declaration of function 'ip4addr_aton'; did you mean 'ipaddr_aton'? [-Wimplicit-function-declaration]
      (((af) == AF_INET) ? ip4addr_aton((src),(ip4_addr_t*)(dst)) : 0)
                           ^~~~~~~~~~~~
../src/open62541.c:56076:9: note: in expansion of macro 'UA_inet_pton'
         UA_inet_pton(AF_INET, addressAsChar, &imr_interface);
         ^~~~~~~~~~~~
/home/chris/GitHub/OPCUA-RTOSServer/lwip_freertos_tcpecho/inc/open62541.h:284:47: error: 'ip4_addr_t' undeclared (first use in this function); did you mean 'ip_addr_t'?
      (((af) == AF_INET) ? ip4addr_aton((src),(ip4_addr_t*)(dst)) : 0)
                                               ^~~~~~~~~~
../src/open62541.c:56076:9: note: in expansion of macro 'UA_inet_pton'
         UA_inet_pton(AF_INET, addressAsChar, &imr_interface);
         ^~~~~~~~~~~~
/home/chris/GitHub/OPCUA-RTOSServer/lwip_freertos_tcpecho/inc/open62541.h:284:47: note: each undeclared identifier is reported only once for each function it appears in
      (((af) == AF_INET) ? ip4addr_aton((src),(ip4_addr_t*)(dst)) : 0)
                                               ^~~~~~~~~~
../src/open62541.c:56076:9: note: in expansion of macro 'UA_inet_pton'
         UA_inet_pton(AF_INET, addressAsChar, &imr_interface);
         ^~~~~~~~~~~~
[...]

Issue #2887, opened 2019-07-15
create freertos amalgamation files


mrburen121:
## Description
I get a error when using cmake to create the freertos amalgamation files.

## Background Information / Reproduction Steps
Used CMake options:
`cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUA_ARCHITECTURE=freertosLWIP -DUA_ENABLE_AMALGAMATION=ON -DUA_ENABLE_PUBSUB=ON -DUA_ARCH_FREERTOS_USE_OWN_MEMORY_FUNCTIONS=ON ..`

This is the output:
```
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PythonInterp: /usr/bin/python3.7 (found version "3.7.3") 
-- Found Git: /usr/bin/git (found version "2.20.1") 
-- Architectures included in amalgamation: freertosLWIP
-- The selected architecture is: freertosLWIP
-- Performing Test CC_HAS_-fstack-protector-strong
-- Performing Test CC_HAS_-fstack-protector-strong - Success
-- Performing Test CC_HAS_-fstack-clash-protection
-- Performing Test CC_HAS_-fstack-clash-protection - Success
-- Performing Test CC_HAS_-mcet -fcf-protection
-- Performing Test CC_HAS_-mcet -fcf-protection - Failed
-- Performing Test CC_HAS_-D_FORTIFY_SOURCE=2
-- Performing Test CC_HAS_-D_FORTIFY_SOURCE=2 - Success
-- Could NOT find Sphinx (missing: SPHINX_EXECUTABLE) 
-- Could NOT find LATEX (missing: LATEX_COMPILER) 
-- Configuring done
-- Generating done
```

While running the make command I get this error:
```
Scanning dependencies of target open62541-object
[ 65%] Building C object CMakeFiles/open62541-object.dir/open62541.c.o
[...]

PR #3023, opened 2019-08-21
[Review] Add FREERTOSLWIP build to travis ci. 

cmbahadir:
Created this PR with reference to [#2993 Comment](https://github.com/open62541/open62541/pull/2993#issuecomment-523176846) but with below concerns;

Still need to externally modify the LWIP_IPV6 macro with sed. See #2511 and #1910

tcpip.h compile error is ignored which does not have any effect on freertos build. It is raised by the used xtensa-gcc toolchain.

python version 2.7.5 raises warnings on line;
python -m pip install --user -r $IDF_PATH/requirements.txt

where there is no warning for 2.7.12 in my local here.
https://github.com/cmbahadir/opcua-esp32/blob/open62541_ci/.travis.yml


jpfr:
Very nice!

This depends on your repository https://github.com/cmbahadir/opcua-esp32.git
So the build on our end depends on the new dependency being always up to date.

How much effort would it be to make the FreeRTOS build self-contained in our repo?
In the first step, it could be enough to build the library as shared objects and not make a "flashable" binary.

In a second step, we could use one of the FreeRTOS environments for Linux/POSIX to also run the unit tests for a FreeRTOS-y environment.

https://www.freertos.org/FreeRTOS-simulator-for-Linux.html
https://github.com/nqd/freertos_linux_devl

cmbahadir:
Hi @jpfr yes it is not good to be my repo dependent but was the fastest way. And probably ı wont update it for a while :)

I got to check how much work needed to make it self contained. I will let you know.

cmbahadir:
Hi @jpfr ,

I have checked the freertos posix simulator on linux and seen that it is not only enough to build the Freertos as shared object but also lwip needed and porting of lwip is specific to an architecture.

@cabralfortiss  
Linux with FREERTOS simulator results with  lwip/tcpip.h undefined reference error which also exists whenever the UA_ARCHITECTURE is selected as "FREERTOSLWIP" w/o any simulator.

I couldn't dive in the details but seems to me that the main bottleneck for a CI of FREERTOSLWIP is LWIP porting.

tools/travis/travis_linux_script.sh
```diff
@@ -1,6 +1,21 @@
 #!/bin/bash
 set -e
 
+if [ "$ARCH" = "freertoslwip" ]; then
+    echo -e "\r\n==Compile multithreaded version==" && echo -en 'travis_fold:start:script.build.freertoslwip\\r'
+    export PATH="/home/travis/build/open62541/esp/xtensa-esp32-elf/bin:$PATH"
+    export IDF_PATH=/home/travis/build/open62541/esp-idf
+    cd $IDF_PATH/examples/opcua-esp32
+    cp /home/travis/build/open62541/open62541/build/open62541.c components/open62541lib
+    sed -i '/#define UA_IPV6 LWIP_IPV6/d' /home/travis/build/open62541/open62541/build/open62541.h
```
Pro: Instead of using sed to change the `LWIP_IPV6` define, can you modify your PR to change this line:
https://github.com/open62541/open62541/blob/46270f270d7a231e7be24711a760d271c77c5842/arch/common/ua_lwip.h#L30
into something like this:
```
#ifdef LWIP_IPV6
#define UA_IPV6 LWIP_IPV6
#endif

File: tools/travis/travis_linux_before_install.sh [hunk 2]
```diff
@@ -1,6 +1,35 @@
 #!/bin/bash
 set -e
 
+if [ "$ARCH" = "freertoslwip" ]; then
+    echo -e "\r\n==Compile multithreaded version==" && echo -en 'travis_fold:start:before_install.build.multithread\\r'
+    mkdir -p build && cd build
+    cmake \
```
Pro: cmake and make commands should not be executed in the install script, but in the `travis_linux.sh` script.

Then you should be able to build the whole open62541 stack and not ignore the build error

tools/travis/travis_linux_before_install.sh
```diff
@@ -1,6 +1,35 @@
...
+    -DUA_ARCHITECTURE:STRING=freertosLWIP \
+    -DUA_ENABLE_PUBSUB:BOOL=OFF \
+    -DUA_ENABLE_METHODCALLS:BOOL=ON \
+    -DUA_ENABLE_NODEMANAGEMENT:BOOL=ON \
+    -DUA_ENABLE_AMALGAMATION:STRING=ON \
+    -DUA_ENABLE_PUBSUB:BOOL=ON \
+    -DUA_LOGLEVEL:STRING=600 ..
+	# Compile error ignored and does not cause any problems. Related Issues: #2887 and #2893
+    make -j || true
```
Pro: hmm, this does not look so good. If make fails, then it's bad. Shouldn't be ignored.

First installing esp-idf and then building the stack should solve it?

tools/travis/travis_linux_before_install.sh
```diff
@@ -1,6 +1,35 @@
...
+	git clone --recursive https://github.com/espressif/esp-idf.git esp-idf
+	cd esp-idf && git checkout tags/v3.2
+	git submodule update --init && cd ..
+	export IDF_PATH=/home/travis/build/open62541/esp-idf
+	python -m pip install --user -r $IDF_PATH/requirements.txt
+	mkdir esp && cd esp
+	wget https://dl.espressif.com/dl/xtensa-esp32-elf-linux64-1.22.0-80-g6c4433a-5.2.0.tar.gz
+	tar -xzf xtensa-esp32-elf-linux64-1.22.0-80-g6c4433a-5.2.0.tar.gz
+	cd .. 
+	git clone https://github.com/cmbahadir/opcua-esp32.git opcua-esp32
```
Pro: Why do we need this additional dependency/repo?

tools/travis/travis_linux_before_install.sh
```diff
@@ -11,7 +40,7 @@ if ! [ -z ${CLANG_FORMAT+x} ]; then
     exit 0
 fi
 
-if [ -z ${DOCKER+x} ] && [ -z ${SONAR+x} ]; then
+if [ -z ${DOCKER+x} ] && [ -z ${SONAR+x} ] &&  [ -z ${ARCH} ]; then
```
Pro: This change is not needed because you have an `exit 0` at the end of your block above.

tools/travis/travis_linux_script.sh
```diff
@@ -1,6 +1,21 @@
...
+    cp /home/travis/build/open62541/open62541/build/open62541.h components/open62541lib/include
+    rm -r components/tcpip_adapter components/lwip components/freertos
```
Pro: Can you add some comments why one needs to remove these folders?

tools/travis/travis_linux_script.sh
```diff
@@ -1,6 +1,21 @@
...
+    cp -r $IDF_PATH/components/tcpip_adapter components/ && cp -r $IDF_PATH/components/lwip components/ && cp -r $IDF_PATH/components/freertos components/ 
+    cp components/freertos/include/freertos/* components/freertos/include
```
Pro: This is also a bit strange, to copy include files. If this is necessary, than it means the the include directories are not correctly set somewhere. Either in the code itself (`#include ...`) or in the build setup.

tools/travis/travis_linux_before_install.sh
cmbahadir: > 4. Optional step in a second PR when the previous ones are implemented: use the xtensa toolchain with an example code to build a flashable binary

Actually this repo of me is an example code to build a flashable binary.

cmbahadir:
> Thank you for your Pull Request and your work!
> 
> It would be really nice to include a FreeRTOS build into CI. I added some review comments.
> 
> In general, you are somehow mixing the install script and the main build script.
> 
> Can we do it like this?
> 
> 1. `travis_linux_before_install.sh` downloads and installs the `esp-idf` stuff + xtensa toolchain and prepares all the dependencies
> 2. The environment variables should be set in the build matrix, not in the script itself
> 3. `travis_linux_script.sh` just runs cmake and make to create the shared library
> 4. Optional step in a second PR when the previous ones are implemented: use the xtensa toolchain with an example code to build a flashable binary
> 
> @cmbahadir would you have time to update this PR and do some more tests?



> Thank you for your Pull Request and your work!
> 
> It would be really nice to include a FreeRTOS build into CI. I added some review comments.
> 
> In general, you are somehow mixing the install script and the main build script.
> 
> Can we do it like this?
> 
> 1. `travis_linux_before_install.sh` downloads and installs the `esp-idf` stuff + xtensa toolchain and prepares all the dependencies
> 2. The environment variables should be set in the build matrix, not in the script itself
> 3. `travis_linux_script.sh` just runs cmake and make to create the shared library
> 4. Optional step in a second PR when the previous ones are implemented: use the xtensa toolchain with an example code to build a flashable binary
> 
> @cmbahadir would you have time to update this PR and do some more tests?


Hello @Pro ,

Thanks for the review i will try to fix them but there are a few constraints in my mind;

1. I can fix it.
2. IDF_PATH environment variable can be set after clonning esp-idf. So not sure if i can move it to build matrix. (may not be impossible.)
3. I will move all make and cmake to travis_linux_scripts.
4. Actually this repo of me (https://github.com/cmbahadir/opcua-esp32) is an example code to build a flashable binary but can be removed for now to be considered later.

tools/travis/travis_linux_script.sh
cmbahadir: Macro added with https://github.com/open62541/open62541/pull/3023/commits/5f31676c03d8fe957e07f31f0cd336b460b4cec9

But needs review if its the right place to add.

tools/travis/travis_linux_before_install.sh
cmbahadir: Fixed with https://github.com/open62541/open62541/pull/3023/commits/5f31676c03d8fe957e07f31f0cd336b460b4cec9

tools/travis/travis_linux_before_install.sh
cmbahadir: I couldn't find a way to compile open62541 library without this error and only needed files are amalgamated open62541.c/h files to build a flashable binary with xtensa toolchain, so;

```
/home/travis/build/cmbahadir/open62541/build/open62541.h:239:24: fatal error: lwip/tcpip.h: No such file or directory
 #include <lwip/tcpip.h>
```
error is ignored. 

There are discussions on #2887 and #2893.

tools/travis/travis_linux_before_install.sh
cmbahadir: Its a sample application to build a flashable binary with xtensa-toolchain.

tools/travis/travis_linux_script.sh
cmbahadir: Fixed with https://github.com/open62541/open62541/pull/3023/commits/5f31676c03d8fe957e07f31f0cd336b460b4cec9

tools/travis/travis_linux_before_install.sh
cmbahadir: Fixed with https://github.com/open62541/open62541/pull/3023/commits/5f31676c03d8fe957e07f31f0cd336b460b4cec9

tools/travis/travis_linux_before_install.sh
cabralfortiss: The problem is that he needs only the amalgamated files to then compile with the right tools. 
I see two possible solutions: 
1- The cleanest is to avoid this error by compiling open62541 properly as a library and then use the library in the project compilation (this instead of the using the amalgamated files in the project). This requires that open62541 is compiled with the xtensa tools, giving the right flags. Unfortunately I couldn't understand the "make" for the project later on. 
2- Create a CMake target in open62541 which only generates the amalgamated files and doesn't compile.

I see point 2 very specific. Point 1 is the best option, but also requires a good understanding of the building process of each part of the system.

cmbahadir: I would suggest to go with option 2 since build systems for each target maybe different. So rather than covering every toolchains configurations with different flags, it would be  enough to provide the the get amalgamated .c./.h files. 

And it will make the current "tcpip.h" error invisible to the user., so user will be more confident about what he is doing.

Just an opinion. :)

Pro:
@cmbahadir thank you for all the updates!
I'll take your current state and try to make it a bit more compact.

cmbahadir:
> @cmbahadir thank you for all the updates!
> I'll take your current state and try to make it a bit more compact.

Hi @Pro  yes please i need help. I do not want to occupy the CI more :) and i can fix and squash commits after a review.

Pro:
Replaced by #3205

PR #3051, opened 2019-08-28
PubSub: Fix inconsistent type usage for publishId

dapaulid:
This pull request fixes an issue with inconsistent type usage (UInt32 vs. UInt16) for `publishId` which resulted in mismatching subscriptions on big-endian systems (PPC) so that no subscriptions could be received.

The commit introduces a small change in the API (I had to change the subscriber tutorial and test) by using `UA_TYPES_UINT32` instead of `UA_TYPES_UINT16` for the publisherId type.

If this is not desired, another possibility would be to use an ugly cast instead:
`nm.publisherId.publisherIdUInt16 = (UA_UInt16) connection->config->publisherId.numeric`

But I really think it's better to use UInt32 consistently here.

### DIFF
--- a/examples/pubsub/tutorial_pubsub_subscribe.c
+++ b/examples/pubsub/tutorial_pubsub_subscribe.c
@@ -92,8 +92,8 @@ addDataSetReader(UA_Server *server) {
     /* The following parameters are used to show that the data published by
      * tutorial_pubsub_publish.c is being subscribed and is being updated in
      * the information model */
-    UA_UInt16 publisherIdentifier = 2234;
-    readerConfig.publisherId.type = &UA_TYPES[UA_TYPES_UINT16];
+    UA_UInt32 publisherIdentifier = 2234;
+    readerConfig.publisherId.type = &UA_TYPES[UA_TYPES_UINT32];
     readerConfig.publisherId.data = &publisherIdentifier;
     readerConfig.writerGroupId    = 100;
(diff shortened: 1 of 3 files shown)


jpfr:
The following types of PublisherId are supported in the spec:

> Bit range 0-2: PublisherId Type
000 The PublisherId is of DataType Byte, This is the default value if ExtendedFlags1 is omitted
001 The PublisherId is of DataType UInt16
010 The PublisherId is of DataType UInt32
011 The PublisherId is of DataType UInt64
100 The PublisherId is of DataType String

We should support all of these.

The comparison in https://github.com/open62541/open62541/pull/3051/files#diff-e9f20e061662bf275f05ad0a52d5ca74L1957 is broken.
Valid values are
```c
typedef enum {
    UA_PUBLISHERDATATYPE_BYTE = 0,
    UA_PUBLISHERDATATYPE_UINT16 = 1,
    UA_PUBLISHERDATATYPE_UINT32 = 2,
    UA_PUBLISHERDATATYPE_UINT64 = 3,
    UA_PUBLISHERDATATYPE_STRING = 4
} UA_PublisherIdDatatype;
```

andreasebner:
A overhaul for the publisherID type support is on the way https://github.com/open62541/open62541/pull/3867. Therefore we close this PR.

andreasebner:
A overhaul for the publisherID type support is on the way #3867. Therefore we close this PR.

PR #3060, opened 2019-08-30
Fix #3048 Initialize arrays in addDiscoveryUrl and return error for customHostname

### LINKED ISSUE #3048: Memory leak when UA_gethostname fails in addDiscoveryUrl
In function https://github.com/open62541/open62541/blob/master/plugins/ua_config_default.c#L236
(I copy the code here for easier reading)
```C
static UA_StatusCode
addDiscoveryUrl(UA_ServerConfig *config, UA_UInt16 portNumber) {
    config->applicationDescription.discoveryUrlsSize = 1;
    UA_String *discurl = (UA_String *) UA_Array_new(1, &UA_TYPES[UA_TYPES_STRING]);
    char discoveryUrlBuffer[220];
    if (config->customHostname.length) {
        UA_snprintf(discoveryUrlBuffer, 220, "opc.tcp://%.*s:%d/",
                    (int)config->customHostname.length,
                    config->customHostname.data,
                    portNumber);
    } else {
        char hostnameBuffer[200];
        if(UA_gethostname(hostnameBuffer, 200) == 0) {
            UA_snprintf(discoveryUrlBuffer, 220, "opc.tcp://%s:%d/", hostnameBuffer, portNumber);
        } else {
            UA_LOG_ERROR(&config->logger, UA_LOGCATEGORY_NETWORK, "Could not get the hostname");
        }
    }
    discurl[0] = UA_String_fromChars(discoveryUrlBuffer);
    config->applicationDescription.discoveryUrls = discurl;
    return UA_STATUSCODE_GOOD;
}
```

Since discoveryUrlBuffer is not initialized, if config->customHostname.length is zero and later UA_gethostname fails (case coming from freeRTOS where this function is not implemented),  discurl[0] = UA_String_fromChars(discoveryUrlBuffer); will still be executed which will call strlen on discoveryUrlBuffer.

I see two options:
1. Return an error when UA_gethostname fails
2. Initialize discoveryUrlBuffer first character to '\0' so the discoveryURL becomes "opc.tcp://:4840/" although this doesn't look nice

And in parallalel, in the freeRTOS arch (and other arch that don't have UA_gethostname), instead of returning an error, return a fixed string which can later (or before) be changed anyhow, but won't produce an error or a funny discoveryURL like "opc.tcp://:4840/" in case the user forgets or is too lazy to change the hostname

cabralfortiss:
Fix #3048

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -231,24 +231,26 @@ addDefaultNetworkLayers(UA_ServerConfig *conf, UA_UInt16 portNumber,
 
 static UA_StatusCode
 addDiscoveryUrl(UA_ServerConfig *config, UA_UInt16 portNumber) {
-       config->applicationDescription.discoveryUrlsSize = 1;
-    UA_String *discurl = (UA_String *) UA_Array_new(1, &UA_TYPES[UA_TYPES_STRING]);
-    char discoveryUrlBuffer[220];
+    char discoveryUrlBuffer[220]; //FIXME: Remove magic number
+    memset(discoveryUrlBuffer, 0, 220);
     if (config->customHostname.length) {
         UA_snprintf(discoveryUrlBuffer, 220, "opc.tcp://%.*s:%d/",
                                                  (int)config->customHostname.length,
                                                  config->customHostname.data,
                                                  portNumber);
     } else {
-    char hostnameBuffer[200];
-       if(UA_gethostname(hostnameBuffer, 200) == 0) {
-               UA_snprintf(discoveryUrlBuffer, 220, "opc.tcp://%s:%d/", hostnameBuffer, portNumber);
-       } else {
+        char hostnameBuffer[200]; //FIXME: Remove magic number
+        memset(hostnameBuffer, 0, 200);
+        if(UA_gethostname(hostnameBuffer, 200) == 0) {
+            UA_snprintf(discoveryUrlBuffer, 220, "opc.tcp://%s:%d/", hostnameBuffer, portNumber);
+        } else {
             UA_LOG_ERROR(&config->logger, UA_LOGCATEGORY_NETWORK, "Could not get the hostname");
+            return UA_STATUSCODE_BADRESOURCEUNAVAILABLE;
         }
     }
-    discurl[0] = UA_String_fromChars(discoveryUrlBuffer);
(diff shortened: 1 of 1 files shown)


plugins/ua_config_default.c
```diff
@@ -231,24 +231,26 @@ addDefaultNetworkLayers(UA_ServerConfig *conf, UA_UInt16 portNumber,
 
 static UA_StatusCode
 addDiscoveryUrl(UA_ServerConfig *config, UA_UInt16 portNumber) {
-       config->applicationDescription.discoveryUrlsSize = 1;
-    UA_String *discurl = (UA_String *) UA_Array_new(1, &UA_TYPES[UA_TYPES_STRING]);
-    char discoveryUrlBuffer[220];
+    char discoveryUrlBuffer[220]; //FIXME: Remove magic number
+    memset(discoveryUrlBuffer, 0, 220);
```
Pro:
```suggestion
    discoveryUrlBuffer[0] = 0;
```

plugins/ua_config_default.c
```diff
@@ -231,24 +231,26 @@ addDefaultNetworkLayers(UA_ServerConfig *conf, UA_UInt16 portNumber,
...
                                                  (int)config->customHostname.length,
                                                  config->customHostname.data,
                                                  portNumber);
     } else {
-    char hostnameBuffer[200];
-       if(UA_gethostname(hostnameBuffer, 200) == 0) {
-               UA_snprintf(discoveryUrlBuffer, 220, "opc.tcp://%s:%d/", hostnameBuffer, portNumber);
-       } else {
+        char hostnameBuffer[200]; //FIXME: Remove magic number
+        memset(hostnameBuffer, 0, 200);
```
cabralfortiss: should I do the same here?

cabralfortiss:
> 
> 
> I guess the magic numbers `220` and `200` can be replaced with `255`. Should be more obvious?!

I don't know actually if there's a maximum length and I have no idea why these values where chosen, but I assume if discoveryUrlBuffer is 255, then hostnameBuffer should be sizeof(discoveryUrlBuffer) - sizeof("opc.tcp://:/") - MAX_PORT_LENGTH

Pro:
Can probably be closed because it is superseded by #3062

cabralfortiss:
Yes, thanks!

PR #3062, opened 2019-08-30
fix(server): Correctly initialize application description discovery urls for custom hostname

### LINKED ISSUE #3048: Memory leak when UA_gethostname fails in addDiscoveryUrl

Pro:
fixes #3048

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -367,7 +367,6 @@ ServerNetworkLayerTCP_start(UA_ServerNetworkLayer *nl, const UA_String *customHo
     /* Get the discovery url from the hostname */
     UA_String du = UA_STRING_NULL;
     char discoveryUrlBuffer[256];
-    char hostnameBuffer[256];
     if (customHostname->length) {
         du.length = (size_t)UA_snprintf(discoveryUrlBuffer, 255, "opc.tcp://%.*s:%d/",
                                         (int)customHostname->length,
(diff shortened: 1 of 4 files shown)


cabralfortiss:
You should also update here 

https://github.com/open62541/open62541/blob/1666968d9c6a4bdd7b5b7591dc804604e471d376/arch/network_tcp.c#L367-L385

about the initialization of arrays and return an error if UA_gethostname fails, otherwise the leak will still be there.

And since we're here, I'd recommend to put hostnameBuffer inside the else to reduce its scope

Pro:
Done. The initialization is not needed here, since `discoveryUrlBuffer` will always be set.

cabralfortiss:
> 
> 
> Done. The initialization is not needed here, since `discoveryUrlBuffer` will always be set.

Tru dat! looks fine for me now. I will test it next week how it works with freeRTOS

PR #3069, opened 2019-09-03
fix(debian): remove braces in debhelper config file

jrcoding:
Braces should be removed here because _lintian_ states 
```
 N:    This debhelper config file appears to use shell brace expansion (such as
 N:    {foo,bar}) to specify files. This happens to work due to an accident of
 N:    implementation but is not a supported feature. Only ?, *, and [...] are
 N:    supported.
 N:    
```
As long as this package produces only the shared lib the affected line should be reduced
 to ```usr/lib/*/*.so```

### DIFF
--- a/debian/libopen62541-dev.install-template
+++ b/debian/libopen62541-dev.install-template
@@ -1,5 +1,5 @@
 usr/include
-usr/lib/*/{*.a,*.so}
+usr/lib/*/*.so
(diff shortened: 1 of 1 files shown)


Pro:
@jrcoding thanks for the fix! I just rebased your branch onto 1.0 to include this in the upcoming release this week.

PR #3072, opened 2019-09-03
Fix lgtm issue

escherstair:
This PR fixes some issues reported by LGTM (Recommendations).
This is WIP


escherstair:
My fault.
passing parameters as const reference is not legal in C (C++ only).
The whole PR can be deleted.

PR #3073, opened 2019-09-03
fix(server): Do not allow node having itself as parent. Fixes stack-overflow

Pro:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16141
Credit to oss-fuzz


jpfr:
👍 Nice!

PR #3076, opened 2019-09-03
docs: Add better documentation on third-party licenses

Pro:


jpfr:
Looks good!

The typo `thrid` appears two times.
The jsmn files should have an additonal license header on top of the code files.

Pro:
jsmn had a new release where they switched to header-only library. That file already includes the corresponding license header. I did not want to include that change since it leads to additional compile warnings. That should be in a separate PR. I'll fix the typo

PR #3078, opened 2019-09-04
fix(debian): debian control has obsolete dbg package

jrcoding:
This fixes ...

```
I: open62541 source: debian-control-has-obsolete-dbg-package libopen62541-1-dbg
N: 
N:    The debian/control file includes declaration of -dbg package.
N:    
N:    Debug package are now autogenerated avoiding waste of miror space.
N:    
N:    Please note that -dbg should be dropped from control file and must not
N:    be transitioned to a dummy package depending on -dbgsym.
N:    
N:    Refer to https://wiki.debian.org/AutomaticDebugPackages for details.
N:
```

it bumps compat to 9 and works since Debian 8 (jessie) which is old-stable, should be old enough :-)


jrcoding:
hmmm, think I made a mistake ??? I don't wanted to commit into master but into 1.0, I wil try it once more...

PR #3079, opened 2019-09-04
fix(debian): debian control has obsolete dbg package

jrcoding:
This fixes ...
```
I: open62541 source: debian-control-has-obsolete-dbg-package libopen62541-1-dbg
N: 
N:    The debian/control file includes declaration of -dbg package.
N:    
N:    Debug package are now autogenerated avoiding waste of miror space.
N:    
N:    Please note that -dbg should be dropped from control file and must not
N:    be transitioned to a dummy package depending on -dbgsym.
N:    
N:    Refer to https://wiki.debian.org/AutomaticDebugPackages for details.
N:
```
it bumps compat to 9 and works since Debian 8 (jessie) which is old-stable, should be old enough...

Hopefully this PR is now related to the 1.0 branch :-)


Pro:
Can we keep the compat level at 7? If I remember correctly Launchpad had issues with newer versions

jrcoding:
> Can we keep the compat level at 7? If I remember correctly Launchpad had issues with newer versions

yes, of course, if you are sure that launchpad has problems with `compat > 7`  I will include a lintian-override for this tag to prevent this lintian message.

Feel free to close the PR

Pro:
Let's merge it as it is, if the build this night fails, I will revert that commit.

PR #3080, opened 2019-09-04
Merge 1.0

jpfr:


jpfr:
Hmm. Some of the compiler options changes make problems.
 ```
/usr/bin/ld: /tmp/ccwZWlaI.ltrans0.ltrans.o: relocation R_X86_64_PC32 against symbol `UA_ObjectAttributes_default' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
````
Did you every see that problem before @Pro?

Pro:
I guess this is due to the change in #3077 

@jrcoding do you know what could be the issue here? (see https://travis-ci.org/open62541/open62541/jobs/580802134)

jpfr:
The same problem occurs in a different context as well: #3085

jrcoding:
sorry, I'm some days off, next week I will try the build on my local Debian builder environment and look what happens...

Pro:
I created a revert PR here: #3087

Let's revert that change for now so that we can continue merging PRs. Strange that this error didn't show up while merging the original PR.

I also tested the same commands on my local ubuntu install, and there the error also doesn't show up...

Pro:
Can be merged when green. Includes the revert.

jrcoding:
> I guess this is due to the change in #3077
> 
> @jrcoding do you know what could be the issue here? (see https://travis-ci.org/open62541/open62541/jobs/580802134)

Hardening enables, among other options, the generation of position independent code (PIC) for shared libs and position independent executables (PIE) for programs.

The error here (#3080) and in #3085 is the injection of the PIE flag which should only be used if compiling objects for an executable **but not** if compiling objects for shared libs! 
The magic in debhelper and/or cmake which detects the type of build and sets the flag correctly seems to fail in this case.

Because this package provides shared libs one can safely disable PIE with `hardening=-pie`

PR #3084, opened 2019-09-05
create clientNonce on each SecureChannel renew

HuberHans:
client is disconnected by server on second renew SecureChannel since clientNonce is reused
OPC Unified Architecture, Part 4 (Table 7 defines the parameters for the Service)
clientNonce:
A random number that shall not be used in any other request. A new clientNonce shall be generated for each time a SecureChannel is renewed.


jpfr:
@mlgiraud Can you sign-off on the change?

jpfr:
This needs to go into 1.0 as well.

jpfr:
Wonderful.
Thanks for looking into it @HuberHans !

PR #3092, opened 2019-09-10
fix(debian): hardening all but no PIE flag

jrcoding:
This enables hardening during debian package build and fixes the error due to #3077 described in #3080 and #3085 .
This fix disables the PIE flag explicitly.


jpfr:
Great!

PR #3093, opened 2019-09-10
fix bugs related to "remove multicast record" and "async read value"

morthlee:
[fix] bug in stopMulticastDiscoveryServer(): error caused by removing record when using custom hostname;

[fix] async read value attribute bug: crash caused by error pointer


andreasebner:
The proposed changes are resolved in recent versions.

PR #3098, opened 2019-09-11
build: mark mbedTLS libraries as imported

chhitz:
This prevents absolute paths for these libraries in open62541Targets.cmake


jpfr:
Cool! Merged.

PR #3105, opened 2019-09-15
Fix O(n^n) runtime for adding many references to one node

jpfr:


opcua-tsn-team-kalycito:
@jpfr Thanks for the fix.
This reduced the startup time by 85%.

PR #3110, opened 2019-09-16
Allow customization of the ua-nodeset dependency path

vladimiroltean:
Some build systems simply can't do a git clone with submodules, as
open62541 currently requires. So simply allow builders to specify
UA_NODESET_DIR as the absolute (or relative) path to the ua-nodeset
project.

Signed-off-by: Vladimir Oltean <olteanv@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,11 @@ include(GNUInstallDirs)
 
 # Set when installed via make install
 set(open62541_TOOLS_DIR ${PROJECT_SOURCE_DIR}/tools)
-set(open62541_NODESET_DIR ${PROJECT_SOURCE_DIR}/deps/ua-nodeset)
+if(NOT UA_NODESET_DIR)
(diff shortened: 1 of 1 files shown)


jpfr:
The change looks good to me. We should merge it once green.
Thanks @vladimiroltean.

In addition, @vladimiroltean, did you see the pack branch?
https://github.com/open62541/open62541/tree/pack/master
It comes with the external dependencies "pre-loaded".
That could save some time.

vladimiroltean:
> In addition, @vladimiroltean, did you see the pack branch?
> https://github.com/open62541/open62541/tree/pack/master
> It comes with the external dependencies "pre-loaded".
> That could save some time.

Well, another feature of said build system is that it won't accept building from a branch, just a commit ID or a tag. I guess the reason why is obvious.

GoetzGoerisch:
Relates to #3310

PR #3129, opened 2019-09-24
Create ModelChangeEvent during AddNodes

jpfr:


mlang-de:
@jpfr Will this modification trigger a model change event each time the public functions UA_Server_add* are called?
(Except bootstrap is true and the node is not already in the tree.)

jpfr:
Yes.
This will be added to all services in the NodeManagement service set.

Some clients automatically monitor this event.
For example UAExpert.

mlang-de:
Ok, but what does this mean for the server startup in which the information model will be builded up? If a server/applicatoin add over 2000 Nodes via the Public Server API and so 2000 Events will be generated? Is it possible to suppress this event during startup via the Public API?

jpfr:
@mlang-de Yes we need a suppression flag for startup.
Also, we need to suppress new events to be emitted when we build up an event (which is represented as an object in the information model before triggering).

andreasebner:
The code path targeted by this issue no longer exists in recent versions.

PR #3135, opened 2019-09-26
Fix buffer size negotiation

basyskom-jvoe:
The server modifies the send and recv buffer sizes for the connection but doesn't
include the changed values into the Acknowledge message.

According to OPC UA Part 6, 7.1.2.4, the buffer sizes shall not be larger than what
the client requested in in the HEL message.

Sending the server's default value causes clients with smaller buffer size limits
(for example Siemens S7-1500) to abort the connection attempt.


mlang-de:
Indirect relates to #2955 where the buffer size negotiation was modified.
Both PRs in combination should be fine.

PR #3136, opened 2019-09-26
ci: Add multiarch (big endian) build on azure

Pro:
This will enable qemu builds on azure and via this also enables unit testing of different architectures.


jpfr:
Is a rebase enough or is there a problem with big endian?

Pro:
I just did a rebase, but I guess there were some other issues with specific architectures. Let's see what this build yields.

Pro:
@jpfr looks like there are some segfaults in encryption. Currently I can not investigate that issue in more detail. Maybe next week.

jpfr:
Could these problems indicate a bug in encryption?
It would be good to get this PR in!

andreasebner:
Azure Pipelines no longer supported.

Issue #2587, opened 2019-03-20
Use NodeSet2.xml Definition fields instead of .bsd files for datatype generator


Pro:
Currently we require Types.bsd files for the datatype_generator.py to generate the corresponding custom datatypes.

In newer NodeSets, the DataType Definition is stored inside the NodeSet2.xml. (e.g. see #2582)

We should use this Definition instead of the .bsd files.
This can be achieved by changing the nodeset_compiler in a way, that it also creates the types files, since it already parses all the information. For this the current datatype_generator.py code should be merged into the nodeset compiler.

jpfr:
We could move generate_datatypes.py as a method into the nodeset compiler.
So that it takes the raw string as input and not a file.
Then, generate_datatypes.py can remain as a standalone wrapper for those with existing .bsd files.

Can we keep the .bsd files for the builtin types baked into the library?
I find it quite educational to show the xml.

Pro:
> We could move generate_datatypes.py as a method into the nodeset compiler.
> So that it takes the raw string as input and not a file.

Maybe we can have the following solution:

Provide a command line switch `datatypes-source` which allows to select options how the datatypes should be generated:
- `bsd` Use the .bsd file which is passed as another parameter
- `encoded` Use the base64 encoded string inside the NodeSet2.xml
- `definition` Use the definition tags of each DataType

Like this we support old Nodesets, and also newer ones which do not come with a .bsd file and also without a base64 encoded type definition.
Also we can keep our builtin types definition as .bsd.

Lo-TW:
> you can use UaModeler: Open the NodeSet2.xml, confirm that it should be converted, and then just export it again. With this, UaModeler will also export a .bsd file

Hi @Pro,
I download UaModeler Version 1.6.5 472, and I try this way. 
1. The UaModeler can't open xml file directly.
2. If I open a project, UaModeler can export xml and export NodeIds(csv) only.

How to export the bsd file in the UaModeler？
Thanks！

PR #3138, opened 2019-09-27
Refactor generate_datatypes

aDogCalledSpot:
In order to allow parsing the types directly from the NodesetXML as discussed in https://github.com/open62541/open62541/issues/2587, some refactoring had to be done first in the nodeset compiler.

There is a `TypeParser` class from which a parser for any kind of format can be implemented by deriving from `TypeParser`. Currently, only parsing from CSV and BSD is supported.

The backend can be swapped out as well, similar to the nodeset compiler. Currently, only using C is supported.

The `generate_datatypes.py` remains for handling arguments and choosing which parser/backend should be used.


jpfr:
Really cool.

You can probably reduce the size of the commit by moving the `tools/generate_datatypes.py` file first and then applying the modifications.

Just `git mv tools/generate_datatypes.py <target-file>`.
Also, we then don't lose history for git blame, and so on.

aDogCalledSpot:
Using `git mv` unfortunately does not change much, as one file is being split into three (of which one is the original file itself). Currently, blame works but history on GitHub is lost.

Codacy doesn't like using `xml.etree` as it leaves vulnerabilities and instead suggests using `defusedxml`. I've refrained from changing this for now as this would introduce a new dependency and `xml.etree` was being used before this refactor as well.

PR #3142, opened 2019-10-01
[Review] (Bug) Fix memory leak if enableRetransmissionQueue is disabled

mlang-de:
Free resources in relation to retransmission queue is available or not.

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -504,8 +504,13 @@ UA_Subscription_publish(UA_Server *server, UA_Subscription *sub) {
     sub->currentKeepAliveCount = 0;
 
     /* Free the response */
-    UA_Array_delete(response->results, response->resultsSize, &UA_TYPES[UA_TYPES_UINT32]);
-    UA_free(pre); /* No need for UA_PublishResponse_deleteMembers */
+    if(retransmission)
+        /* NotificationMessage was moved into retransmission queue */
+        UA_NotificationMessage_init(&response->notificationMessage);
+    response->availableSequenceNumbers = NULL;
+    response->availableSequenceNumbersSize = 0;
+    UA_PublishResponse_clear(&pre->response);
+    UA_free(pre);
 
     /* Repeat sending responses if there are more notifications to send */
     if(moreNotifications)


jpfr:
Oh wow.
Could you reproduce the problem or is it theoretical?

I was under the impression that line 473 catches this case.
https://github.com/open62541/open62541/blob/4aa092c8073fa478ba401aced95ff275c3fa2e3d/src/server/ua_subscription.c#L473

mlang-de:
This line only queue the notification message if a retransmission queue is available. See https://github.com/open62541/open62541/blob/4aa092c8073fa478ba401aced95ff275c3fa2e3d/src/server/ua_subscription.c#L428.
If the message is not queued the allocated notification message memory will be lost. This I could reproduce on our system, but should also possible on each other system as long the server is configured with enableRetransmissionQueue = false. The current default is each time enableRetransmissionQueue = true.

mlang-de:
If enableRetransmissionQueue = false is configured then many monitoredItem/subscription CTT scripts fails because the is no queue anymore to support the unacknowledged sequence numbers.

mlang-de:
@jpfr I prefer to merge this PR and fix the wrong behaviour in a futher PR since this PR will only fix the memory leak does not cause the missbehaviour.

jonasgreen88:
> If enableRetransmissionQueue = false is configured then many monitoredItem/subscription CTT scripts fails because the is no queue anymore to support the unacknowledged sequence numbers.

According to the "Embedded DataChange Subscription Server Facet" a retransmission queue is not required: 

> Support at least 2 Publish Service requests per Session.This number has to be supported for all of the minimum required sessions. Support of a NotificationMessage retransmission queue is not required; if not available the Republish Service returns Bad_MessageNotAvailable.

But OPC UA part 4 has designed the DataChange Subscription feature with the assumption that the retransmission queue is mandatory. So there has been some missunderstanding between the specification that specifies the functionality and the profiles that specifies what is mandatory. There is a issue report on the OPC Mantis covering this missmatch: https://apps.opcfoundation.org/mantis/view.php?id=4795.

> OPC UA Part 4 requires the maintenance of a retransmission queue and the sequence number acknowledgement in Publish. Therefore all text assumes that this is available.
But the Embedded DataChange Subscription Server Facet allows to disable the retransmission queue. But the behaviour in this case in not defined.
See also
https://opcfoundation.org/forum/opc-ua-standard/shall-the-availablesequence-numbers-field-of-publish-responses-be-empty-when-retransmission-queue-is-not-implemented/

I think we should await the result of the OPC Mantis issue report before doing changes in the functionality in open62541 because of CTT failures.

mlang-de:
I agree with you, but the problem is in case of a "wrong/misconfiguration" open62541 configuration the current implementation causes memory leaks.
During read the mantis issue as well as your linked OPC Foundation forum post it is enough to handle only the sended/unacknowlegded sequence numbers. I am right? The current open62541 implementation store the whole network message for this. 
IMHO a solution could be:
- if retransmission queue is disabled use the same list to store the sequence numbers without the whole network message (new PR required, this one here should only fix the memory leak)

jonasgreen88:
> I agree with you, but the problem is in case of a "wrong/misconfiguration" open62541 configuration the current implementation causes memory leaks.
> During read the mantis issue as well as your linked OPC Foundation forum post it is enough to handle only the sended/unacknowlegded sequence numbers. I am right? The current open62541 implementation store the whole network message for this.
> IMHO a solution could be:
> 
> * if retransmission queue is disabled use the same list to store the sequence numbers without the whole network message (new PR required, this one here should only fix the memory leak)

I totally agree that we should fix the memory leak you found and fixed in this PR!

But I'm not sure if your solution to only store the sequence number in the retransmission queue is supported by the specification. The idea with the retransmission queue is that the client shall be able to ask for a retransmission of the network message.
 
This could of course be the solution, but I don't think we should precede the outcome of Mantis issue 4795 and do something before the issue has been solved by OPC Foundation.

mlang-de:
Ok, I read the forum post, mantis and today the spec. You are right. Fix the memory leak is ok for now. The rest should wait for the outcome by the mantis issue.
@jpfr  This issue is ready for review (again :) ).

jpfr:
Rebased on 1.0.

PR #3157, opened 2019-10-09
fix(pubsub): Multiple datasetreader with same Publisher ID

opcua-tsn-team-kalycito:
- Get the required Datasetreader even if multiple datasetreader has same Publisher ID
 - Modified the return value from Datasetreader to status values


jpfr:
Hey @opcua-tsn-team-kalycito ,

a small commit was added to the PR. Please have a quick look.
Making the code as simple as possible is a priority to keep it maintainable.

opcua-tsn-team-kalycito:
Hi @jpfr,

Code looks good. Thanks for the enhancement.

PR #3163, opened 2019-10-10
Fix various oss-fuzz issues

Pro:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=17099

Credit to oss-fuzz


jpfr:
Can you make that a PR for 1.0?

PR #3173, opened 2019-10-14
fix compilation warning about unused variables

haphi:
Compiling with `-Wall -Wextra -Werror`, compiler complains about missing parameters in log functions disabled via `UA_LOGLEVEL`.


jpfr:
Hmm. This broke quite a few builds in the CI.
For which compilers are the fixes intended?

PR #3179, opened 2019-10-16
fix(debian): separate tools from dev

jrcoding:
Building pack/1.0 produces the lintian message

```
I: libopen62541-1-dev: arch-dep-package-has-big-usr-share 104164kB 98%
N: 
N:    The package has a significant amount of architecture-independent data
N:    (over 4MB, or over 2MB and more than 50% of the package) in /usr/share
N:    but is an architecture-dependent package. This is wasteful of mirror
N:    space and bandwidth since it means distributing multiple copies of this
N:    data, one for each architecture.
N:
```
which is correct because the `libopen62541-<soname>-dev` package is architecture dependent and the tools are not. 

This PR splits the previous `libopen62541-<soname>-dev` into the architecture dependent `libopen62541-<soname>-dev` package and the architecture independent `libopen62541-<soname>-tools` package. The `libopen62541-<soname>-tools` package gets a `Recommends: libopen62541-<soname>-dev`.

BTW  I've added some minor changes such as this PR bumps the Debian `Standards-Version:` to the current version `4.4.1` and in addition extends the description of the packages a bit to avoid the following lintian complaints 

```
I: open62541 source: duplicate-long-description libopen62541-1-dev libopen62541-1
N: 
N:    The listed binary packages all share the same extended description. Some
N:    additional information in the extended description explaining what is in
N:    each package and how it differs from the other packages is useful,
N:    particularly for users who aren't familiar with Debian's package naming
N:    conventions.
N:    
```
and
```
I: libopen62541-1: extended-description-is-probably-too-short
N: 
N:    The extended description (the lines after the first line of the
N:    "Description:" field) is only one or two lines long. The extended
N:    description should provide a user with enough information to decide
N:    whether they want to install this package, what it contains, and how it
N:    compares to similar packages. One or two lines is normally not enough to
N:    do this.
N:
```

### DIFF
--- a/debian/control-template
+++ b/debian/control-template
@@ -6,7 +6,7 @@ Build-Depends: debhelper (>= 9),
  git,
  graphviz,
  python (>= 2.7),
-Standards-Version: 4.1.2
+Standards-Version: 4.4.1
 Section: libs
 Homepage: https://open62541.org/
 Vcs-Git: https://github.com/open62541/open62541.git
@@ -16,13 +16,42 @@ Package: libopen62541-<soname>-dev
 Section: libdevel
 Architecture: any
 Multi-Arch: same
-Depends: libopen62541-<soname> (= ${binary:Version}), ${misc:Depends}, python
-Description: Development header files for open62541
- open62541 is an open source C (C99) implementation of the OPC UA standard
+Depends: libopen62541-<soname> (= ${binary:Version}), ${misc:Depends}
+Description: Open source implementation of OPC UA - development files
+ open62541 (http://open62541.org) is an open source and free implementation
+ of OPC UA (OPC Unified Architecture) written in the common subset of the
+ C99 and C++98 languages.
+ The library is usable with all major compilers and provides the necessary
+ tools to implement dedicated OPC UA clients and servers, or to integrate
+ OPC UA-based communication into existing applications.
+ .
+ This package provides the open62541 header and development files
+
+Package: libopen62541-<soname>-tools
+Section: libdevel
+Architecture: all
+Depends: ${misc:Depends}, python
+Recommends: libopen62541-<soname>-dev
+Description: Open source implementation of OPC UA - tools
+ open62541 (http://open62541.org) is an open source and free implementation
+ of OPC UA (OPC Unified Architecture) written in the common subset of the
+ C99 and C++98 languages.
+ The library is usable with all major compilers and provides the necessary
+ tools to implement dedicated OPC UA clients and servers, or to integrate
+ OPC UA-based communication into existing applications.
+ .
+ This package provides some open62541 tools, e.g. the nodeset compiler
 
 Package: libopen62541-<soname>
 Architecture: any
 Multi-Arch: same
 Depends: ${shlibs:Depends}, ${misc:Depends}
-Description: Open source implementation of OPC UA (OPC Unified Architecture) aka IEC 62541
- open62541 is an open source C (C99) implementation of the OPC UA standard
+Description: Open source implementation of OPC UA - shared library
+ open62541 (http://open62541.org) is an open source and free implementation
+ of OPC UA (OPC Unified Architecture) written in the common subset of the
+ C99 and C++98 languages.
+ The library is usable with all major compilers and provides the necessary
+ tools to implement dedicated OPC UA clients and servers, or to integrate
+ OPC UA-based communication into existing applications.
+ .
+ This package provides the open62541 shared library
--- a/debian/control-template
+++ b/debian/control-template
@@ -3,7 +3,6 @@ Priority: optional
 Maintainer: open62541 Team <open62541-core@googlegroups.com>
 Build-Depends: debhelper (>= 9),
  cmake (>= 2.8),
- git,
  graphviz,
  python (>= 2.7),
 Standards-Version: 4.4.1
--- a/debian/libopen62541-dev.install-template
+++ b/debian/libopen62541-dev.install-template
@@ -2,4 +2,3 @@ usr/include
 usr/lib/*/*.so
 usr/lib/*/cmake/open62541/*
 usr/lib/*/pkgconfig/open62541.pc
-usr/share/open62541/*
--- a/debian/libopen62541-tools.install-template
+++ b/debian/libopen62541-tools.install-template
@@ -1,1 +1,6 @@
-usr/share/open62541/tools
+usr/share/open62541/tools/*.py
+usr/share/open62541/tools/certs
+usr/share/open62541/tools/nodeset_compiler
+usr/share/open62541/tools/schema
+usr/share/open62541/tools/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml
+usr/share/open62541/tools/ua-nodeset/Schema/Opc.Ua.Types.bsd
--- a/debian/libopen62541-tools.install-template
+++ b/debian/libopen62541-tools.install-template
@@ -4,3 +4,6 @@ usr/share/open62541/tools/nodeset_compiler
 usr/share/open62541/tools/schema
 usr/share/open62541/tools/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml
 usr/share/open62541/tools/ua-nodeset/Schema/Opc.Ua.Types.bsd
+usr/share/open62541/tools/ua-nodeset/Schema/AttributeIds.csv
+usr/share/open62541/tools/ua-nodeset/Schema/NodeIds.csv
+usr/share/open62541/tools/ua-nodeset/Schema/StatusCode.csv
--- a/debian/libopen62541-tools.install-template
+++ b/debian/libopen62541-tools.install-template
@@ -1,0 +1,1 @@
+usr/share/open62541/tools
--- a/tools/prepare_packaging.py
+++ b/tools/prepare_packaging.py
@@ -81,6 +81,9 @@ install_file_template = os.path.join(debian_path, "libopen62541-dev.install-temp
 install_file = os.path.join(debian_path, "libopen62541-{}-dev.install".format(version_major))
 os.rename(install_file_template, install_file)
 
+install_file_template = os.path.join(debian_path, "libopen62541-tools.install-template")
+install_file = os.path.join(debian_path, "libopen62541-{}-tools.install".format(version_major))
+os.rename(install_file_template, install_file)
 
 # Create rule file and replace template variables
 rule_file_template = os.path.join(debian_path, "rules-template")


jpfr:
Good idea to clean this up!
Here are some thoughts.

The build flag UA_BUILD_TOOLS also generates arch-dependent executables.
For example the `ua2json` shell utility.

From my (limited) experience with debian packaging, a usual structure would be as follows:

- libopen62541: Contains the .so shared object that can be used by other packages
- libopen62541-dev: Contains the header files so that the local user can build his software from source against the library
- libopen62541-doc: Contains the PDF/HTML documentation
-libopen62541-tools: Contains additional tools, such as the nodeset compiler and ua2json

The reason why the package is so big is the inclusion of ua-nodeset XML files.
They are actually maintained and hosted by the OPC Foundation.
Putting them into the open62541 distribution package is somewhat misleading and dishonest.
We don't want these kinds of misunderstanding with the OPC Foundation.

It is literally a one-line command argument to download the nodsets.
We can put that in some prominent position.

The only nodeset XML that should be included are the _Minimal.xml ones.
https://github.com/open62541/open62541/tree/master/tools/schema
The _Minimal.xml are used to generate the information model in the binary library.
So they are needed to generate custom information models that can be loaded into the server.

jrcoding:
> Good idea to clean this up!
> Here are some thoughts.
> 
> The build flag UA_BUILD_TOOLS also generates arch-dependent executables.
> For example the `ua2json` shell utility.
> 
O.k., I will take a look on this...

> From my (limited) experience with debian packaging, a usual structure would be as follows:
> 
>     * libopen62541: Contains the .so shared object that can be used by other packages
> 
>     * libopen62541-dev: Contains the header files so that the local user can build his software from source against the library
> 
>     * libopen62541-doc: Contains the PDF/HTML documentation
> 
>     * libopen62541-tools: Contains additional tools, such as the nodeset compiler and ua2json
> 
> 

Yes, I agree with the proposed structure. 
BTW putting the PDF/HTML docs in a separate package is already on the way but my intention was not to change too much in one PR to get easy understandable changes ...

> The reason why the package is so big is the inclusion of ua-nodeset XML files.
> They are actually maintained and hosted by the OPC Foundation.
> https://github.com/OPCFoundation/UA-Nodeset
> Putting them into the open62541 distribution package is somewhat misleading and dishonest.
> We don't want these kinds of misunderstanding with the OPC Foundation.
> 
> It is literally a one-line command argument to download the nodsets.
> We can put that in some prominent position.
> 
> The only nodeset XML that should be included are the _Minimal.xml ones.
> https://github.com/open62541/open62541/tree/master/tools/schema
> The _Minimal.xml are used to generate the information model in the binary library.
> So they are needed to generate custom information models that can be loaded into the server.

So we can shrink tools down and e.g. put the one-line-command into the description of the tools package?

Here I need some advice on what should be included and what shouldn't...

jrcoding:
> > Good idea to clean this up!
> > Here are some thoughts.
> > The build flag UA_BUILD_TOOLS also generates arch-dependent executables.
> > For example the `ua2json` shell utility.
> 
> * O.k., I will take a look on this...
> 

* At the moment ua2json is not included in tools (see tree in following comment) so tools is still architecture independent.
* Is this correct or should ua2json be included?

> > From my (limited) experience with debian packaging, a usual structure would be as follows:
> > ```
> > * libopen62541: Contains the .so shared object that can be used by other packages
> > 
> > * libopen62541-dev: Contains the header files so that the local user can build his software from source against the library
> > 
> > * libopen62541-doc: Contains the PDF/HTML documentation
> > 
> > * libopen62541-tools: Contains additional tools, such as the nodeset compiler and ua2json
> > ```
> 
> Yes, I agree with the proposed structure.
> BTW putting the PDF/HTML docs in a separate package is already on the way but my intention was not to change too much in one PR to get easy understandable changes ...
> 
> > The reason why the package is so big is the inclusion of ua-nodeset XML files.
> > They are actually maintained and hosted by the OPC Foundation.
> > https://github.com/OPCFoundation/UA-Nodeset
> > Putting them into the open62541 distribution package is somewhat misleading and dishonest.
> > We don't want these kinds of misunderstanding with the OPC Foundation.
> > It is literally a one-line command argument to download the nodsets.
> > We can put that in some prominent position.
> > The only nodeset XML that should be included are the _Minimal.xml ones.
> > https://github.com/open62541/open62541/tree/master/tools/schema
> > The _Minimal.xml are used to generate the information model in the binary library.
> > So they are needed to generate custom information models that can be loaded into the server.
> 
> * So we can shrink tools down and e.g. put the one-line-command into the description of the tools package?
> 
> * Here I need some advice on what should be included and what shouldn't...

* Advice needed: What should be included into tools? (I think we should shrink it to the only needed parts, right?)
* At the moment the tree structure of tools look like this (I've removed the file names from the ua-nodeset tree to make the comment not too verbose)
```
.
└── tools
    ├── certs
    │   ├── create_self-signed.py
    │   └── localhost.cnf
    ├── generate_datatypes.py
    ├── generate_nodeid_header.py
    ├── generate_statuscode_descriptions.py
    ├── nodeset_compiler
    │   ├── backend_graphviz.py
    │   ├── backend_open62541_datatypes.py
    │   ├── backend_open62541_nodes.py
    │   ├── backend_open62541.py
    │   ├── datatypes.py
    │   ├── __init__.py
    │   ├── NodeID_NS0_Base.txt
    │   ├── nodeset_compiler.py
    │   ├── nodeset.py
    │   ├── nodeset_testing.py
    │   ├── nodes.py
    │   ├── opaque_type_mapping.py
    │   └── README.md
    ├── schema
    │   ├── Custom.Opc.Ua.Transport.bsd
    │   ├── datatypes_dataaccess.txt
    │   ├── datatypes_discovery.txt
    │   ├── datatypes_events.txt
    │   ├── datatypes_historizing.txt
    │   ├── datatypes_method.txt
    │   ├── datatypes_minimal.txt
    │   ├── datatypes_pubsub.txt
    │   ├── datatypes_query.txt
    │   ├── datatypes_subscriptions.txt
    │   ├── datatypes_transport.txt
    │   ├── NodeIds.csv
    │   ├── Opc.Ua.NodeSet2.EventsMinimal.xml
    │   ├── Opc.Ua.NodeSet2.HistorizingMinimal.xml
    │   ├── Opc.Ua.NodeSet2.Minimal.xml
    │   ├── Opc.Ua.NodeSet2.PubSubMinimal.xml
    │   ├── Opc.Ua.Types.bsd
    │   └── StatusCode.csv
    └── ua-nodeset
        ├── ADI
        │   ├── ***
        ├── AnsiC
        │   ├── ***
        ├── CSPPlusForMachine
        │   ├── ***
        ├── DI
        │   ├── ***
        ├── DotNet
        │   ├── ***
        ├── FDI
        │   ├── ***
        ├── GDS
        │   ├── ***
        ├── ISA-95
        │   ├── ***
        ├── MDIS
        │   ├── ***
        ├── MTConnect
        │   ├── ***
        ├── PLCopen
        │   ├── ***
        ├── POWERLINK
        │   ├── ***
        ├── readme.md
        ├── Schema
        │   ├── ***
        ├── Sercos
        │   ├── ***
        ├── TMC
        │   ├── ***
        ├── v1.03 NodeSet Errata 2016-04-16.txt
        └── v1.04 NodeSet Errata.txt
```

jpfr:
Take out the `ua-nodeset` folder.
The ua2json utility is a binary executable and can be added to /usr/bin or similar.

jrcoding:
> Take out the `ua-nodeset` folder.

o.k. ... should the installation be removed from the main `CMakeLists.txt` or should only the installation by the Debian package be avoided?

> The ua2json utility is a binary executable and can be added to /usr/bin or similar.

ua2json is not included in the Debian package at the moment because **the Debian build is without `UA_BUILD_TOOLS`** (see  branch `pack/1.0` file `debian/rules`), should this really be changed?

jpfr:
> o.k. ... should the installation be removed from the main CMakeLists.txt or should only the installation by the Debian package be avoided?

tbh, I would not have it in `make install` or the Debian package.
We can of course use parts of ua-nodeset to generate our executables.
@Pro do you have an opinion here?

>  Debian build is without UA_BUILD_TOOLS

In that case I would enable it and add ua2json to the tools package.
More binary shell tools will be coming over time.

jrcoding:
> 
> > Debian build is without UA_BUILD_TOOLS
> 
> In that case I would enable it and add ua2json to the tools package.
> More binary shell tools will be coming over time.

hmmm, `UA_BUILD_TOOLS=ON` only has an effect if also ` UA_ENABLE_JSON_ENCODING=ON` which is marked as `EXPERIMENTAL`. 

I'm afraid it's not a good idea to build the Debian package with `EXPERIMENTAL` stuff `ON`, what is your opinion ???

Pro:
The ua-nodeset is included in the dev package because the nodeset compiler needs at least the full NS0 NodeSet2.xml file if one needs to compile his own nodeset based on that. The CMake macros automatically handle the dependencies there.

I see a few possible solutions to that:

1. Just include the whole ua-nodeset in the dev or tools package. ua-nodeset is under MIT license, so at least from that side there aren't any issues.
2. Only include a subset of ua-nodeset, i.e., the ua-nodeset/Schema folder, which is required. The other files are not really required here.
3. Create a `libopen62541-nodeset` package which includes the ua-nodeset stuff and is a dependency of the tools package
4. Adapt the CMake macro so that one needs to download/clone the ua-nodeset manually, and specify the path to the ua-nodeset directory before using the macros. This will maybe lead to version inconsistencies. E.g., the NS0 used for compiling may differ from the pulled NS0 and therefore lead to errors when initializing user-defined nodeset.

jrcoding:
> The ua-nodeset is included in the dev package because the nodeset compiler needs at least the full NS0 NodeSet2.xml file if one needs to compile his own nodeset based on that. The CMake macros automatically handle the dependencies there.
> 
> I see a few possible solutions to that:
> 
>  1. Just include the whole ua-nodeset in the dev or tools package. ua-nodeset is under MIT license, so at least from that side there aren't any issues.
> 
If we agree on option 1 the PR should already work...

jpfr:
> Only include a subset of ua-nodeset, i.e., the ua-nodeset/Schema folder, which is required. The other files are not really required here.

We only have to include these two files:

- https://github.com/OPCFoundation/UA-Nodeset/blob/master/Schema/Opc.Ua.NodeSet2.xml
- https://github.com/OPCFoundation/UA-Nodeset/blob/master/Schema/Opc.Ua.Types.bsd

jrcoding:
> > Only include a subset of ua-nodeset, i.e., the ua-nodeset/Schema folder, which is required. The other files are not really required here.
> 
> We only have to include these two files:
> 
>     * https://github.com/OPCFoundation/UA-Nodeset/blob/master/Schema/Opc.Ua.NodeSet2.xml
> 
>     * https://github.com/OPCFoundation/UA-Nodeset/blob/master/Schema/Opc.Ua.Types.bsd

...done

Pro:
> > > Only include a subset of ua-nodeset, i.e., the ua-nodeset/Schema folder, which is required. The other files are not really required here.
> > 
> > 
> > We only have to include these two files:
> > ```
> > * https://github.com/OPCFoundation/UA-Nodeset/blob/master/Schema/Opc.Ua.NodeSet2.xml
> > 
> > * https://github.com/OPCFoundation/UA-Nodeset/blob/master/Schema/Opc.Ua.Types.bsd
> > ```
> 
> ...done

And the .csv is also required

jrcoding:
which exactly?  or all .csv files?

jpfr:
Looks good! Merged.

There seems to be some interaction between debian-specific cmake flags, etc.
Can you add a short section to the build docs on how to generate debian/ubuntu packages?

Then, users can create custom debian packages with special features enabled, etc.
I know one big user who would host custom builds in their internal deb source server.

jrcoding:
> Looks good! Merged.
> 
> There seems to be some interaction between debian-specific cmake flags, etc.
> Can you add a short section to the build docs on how to generate debian/ubuntu packages?
> 
> Then, users can create custom debian packages with special features enabled, etc.
> I know one big user who would host custom builds in their internal deb source server.

o.k., coming soon...

PR #3184, opened 2019-10-17
return correct error code when information model setup failed

ursfassler:
This fixes a problem I had. The root of the problem is, that in case of an error while initializing the information model the return values are combined with bit-wise OR.
I tried to use an AnalogItemType in a reduced Namespace 0 where this type is not available. The error code BadTooManySubscriptions (0x80770000) was returned, leading investigation into the wrong direction. The correct error code is BadTypeDefinitionInvalid (0x80630000).
This patch aborts initialization after the first error.


Pro:
Thanks for the PR.

This can probably be achieved the same way as we do it already for the whole nodeset:

https://github.com/open62541/open62541/blob/master/tools/nodeset_compiler/backend_open62541.py#L292

```
UA_StatusCode retval = UA_STATUSCODE_GOOD;
// dummy is required
bool dummy = (!(retVal = function_1()) && !(retVal = function_2()) && !(retVal = function_3())); (void)(dummy);
return retval;
```

I guess that's more efficient than adding ifs.

ursfassler:
@NoelGraf I agreed with the CLA but it looks like there is a problem. I am no longer working at the company with this email address and I am also no longer part of that organization on github.

But I think I already signed the CLA back then and also have contributions with this email address (e.g. 24011388eccc2479d7aa11829e3c8183b0f991aa). Hope this helps, otherwise let me know what I can do to bring this patch upstream.

NoelGraf:
@ursfassler 
As long as Github says that the CLA is not signed, we cannot merge the PR. Since you have already agreed to the CLA, I suggest that I create a new PR and merge your changes with your commit.
Is that ok with you?

ursfassler:
@andreasebner this is ok. Happy if it gets in. :+1:

andreasebner:
Thank you for the contribution. Merged in #5197

PR #3188, opened 2019-10-18
[WIP] feat(PubSub):Time triggered multithreaded application with ETF

opcua-tsn-team-kalycito:
- Run time triggered PubSub application using ETF at a publishing interval
   of 200 microseconds with 64 bytes payload
 -TSN_ETF_publisher and TSN_ETF loopback application are the ETF feature application from this      commit 
 - To build this application with ethernet over UADP enable the following in cmake:  UA_BUILD_EXAMPLES, UA_ENABLE_PUBSUB, UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING, UA_ENABLE_PUBSUB_ETH_UADP
 - Integrate time triggered functionality and publish
   OPCUA PubSub packets at calculated transmission time
 - Handle PubSub application for UDP and Ethernet over UADP
 - Implement Publisher and Subscriber as threads and handle
   in same core
 - Integrate target variable implementation in Subscriber thread
 - TODO: Write unit test cases to ensure no memory leaks are possible


jpfr:
In order to ensure that the message is ready in time for the next qbv window (for TSN), the time-offset for the message preparation has to be explicitly defined. That way, slower processors can use a higher offset and still hit the window.

Explicit time-offsets are a **must requirement** for ETF.

The example in https://github.com/open62541/open62541/blob/master/examples/pubsub_realtime/pubsub_interrupt_publish.c uses a `timer_t` for the interrupt. But I'm not sure if this can be directly synced to the clock of the TSN hardware.

vincentxw:
I see that socket priority is being hardcoded for now. I'd like to point out that Linux socket priorities are commonly used to map packets to VLAN priorities and also hardware queues. Making it configurable per-publisher would prove to be useful.

opcua-tsn-team-kalycito:
@vincentxw Thanks for the suggestion. We will take this into consideration in our future design.

PR #3191, opened 2019-10-22
adding missing submodule init command to Dockerfile

LightSoar:
fix: adding missing submodule init command to Dockerfile

Without this line, `docker build` terminates with the following error message:

> CMake Error at CMakeLists.txt:830 (message):
>   File /opt/open62541/deps/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml not found.
>   You probably need to initialize the git submodule for deps/ua-nodeset.


jpfr:
Thanks!

PR #3194, opened 2019-10-23
fix(pubsub): add size checks before UA_calloc calls

### LINKED ISSUE #3145: UA_calloc() - Implementation defined behaviour
Hello!

I noticed that at some code locations the SDK depends on calloc returning a valid pointer if the size parameter is 0.

According to the man pages: https://linux.die.net/man/3/calloc the calloc function may return NULL or a valid pointer if calloc parameters are 0. The behaviour depends on the implementation.

I am using a custom memory managment implementation for a realtime environment. The overwritten calloc function returns a NULL pointer if the size parameter is 0. 

This leads to an invalid out of memory error at some functions within : 
https://github.com/open62541/open62541/blob/master/src/pubsub/ua_pubsub_writer.c

e.g function UA_WriterGroupConfig_copy: 
Here the function returns BADOUTOFMEMORY if src->groupPropertiesSize is 0 and calloc returns a NULL pointer. 

https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L661-L680

At other code locations this case is handled correctly. 
E.g.: UA_PubSubConnectionConfig_copy within the same file.
https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L46-L59

I think I found following lines of code within https://github.com/open62541/open62541/blob/master/src/pubsub/ua_pubsub_writer.c
 which can lead to implementation dependent behaviour. 

https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L512-L513

https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L672-L675

https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L904-L905

https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L1215-L1216

<br>
<br>

These lines handle the undefined behaviour case: 

https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L46-L48

https://github.com/open62541/open62541/blob/073f417d828eb6db44ebcd22aeefdc1bc5bc7174/src/pubsub/ua_pubsub_writer.c#L324-L325

I have only checked https://github.com/open62541/open62541/blob/master/src/pubsub/ua_pubsub_writer.c

<br>

Could you add additional checks for the implementation defined calloc/malloc return value?

Many thanks in advance
Best regards
Martina

MartinaFelberFEM:
Adding checks to avoid calling UA_calloc with param num == 0.
Because the return value of UA_calloc is in this case implementation defined.

Fixes #3145

### DIFF
--- a/src/pubsub/ua_pubsub_writer.c
+++ b/src/pubsub/ua_pubsub_writer.c
@@ -507,12 +507,14 @@ UA_DataSetWriterConfig_copy(const UA_DataSetWriterConfig *src,
     retVal |= UA_String_copy(&src->name, &dst->name);
     retVal |= UA_String_copy(&src->dataSetName, &dst->dataSetName);
     retVal |= UA_ExtensionObject_copy(&src->messageSettings, &dst->messageSettings);
-    dst->dataSetWriterProperties = (UA_KeyValuePair *)
-        UA_calloc(src->dataSetWriterPropertiesSize, sizeof(UA_KeyValuePair));
-    if(!dst->dataSetWriterProperties)
-        return UA_STATUSCODE_BADOUTOFMEMORY;
-    for(size_t i = 0; i < src->dataSetWriterPropertiesSize; i++){
-        retVal |= UA_KeyValuePair_copy(&src->dataSetWriterProperties[i], &dst->dataSetWriterProperties[i]);
+    if (src->dataSetWriterPropertiesSize > 0) {
+        dst->dataSetWriterProperties = (UA_KeyValuePair *)
+            UA_calloc(src->dataSetWriterPropertiesSize, sizeof(UA_KeyValuePair));
+        if(!dst->dataSetWriterProperties)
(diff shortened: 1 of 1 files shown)


jpfr:
Cool, thanks!

PR #3198, opened 2019-10-23
fix(PubSub): Publishing order of multiple DataSetFields

opcua-tsn-team-kalycito:
- The order of DSF should be the same while creating and publishing and
   the existing implementation publishes reversely
 - Used TAILQ structure for DSF instead of LIST structure


jpfr:
Ich you want to insert at the end, use a tailq structure.
It’s already defined in queue.h.
And without the overhead for insertion.

opcua-tsn-team-kalycito:
@jpfr The requested changes are done. Please have a look at the latest version.

jpfr:
Perfect PR.

PR #3209, opened 2019-10-25
feat(debian): add PDF and HTML docs including examples

jrcoding:
This PR adds a Debian doc package and provides the PDF and HTML documentation for open62541, the installation place, as usual at Debian systems, will be `/usr/share/doc/open62541`.
In addition the package provides the open62541 example application sources at `/usr/share/open62541/examples`.
This PR is based on prototype work of @ralphlange (see https://github.com/ralphlange/open62541/tree/feature/debianization/debian)

### DIFF
--- a/debian/control-template
+++ b/debian/control-template
@@ -5,6 +5,13 @@ Build-Depends: debhelper (>= 9),
  cmake (>= 2.8),
  graphviz,
  python (>= 2.7),
+ python-sphinx,
+ python-sphinx-rtd-theme | base-files (<= 7.2ubuntu5.6~),
+ graphviz,
+ texlive-fonts-recommended,
+ texlive-latex-extra,
+ texlive-generic-extra,
+ latexmk
 Standards-Version: 4.4.1
 Section: libs
 Homepage: https://open62541.org/
@@ -54,3 +61,19 @@ Description: Open source implementation of OPC UA - shared library
  OPC UA-based communication into existing applications.
  .
  This package provides the open62541 shared library
+
+Package: open62541-doc
+Section: doc
+Architecture: all
+Multi-Arch: foreign
+Depends: ${misc:Depends}, ${sphinxdoc:Depends}
+Description: Open source implementation of OPC UA - documentation
+ open62541 (http://open62541.org) is an open source and free implementation
+ of OPC UA (OPC Unified Architecture) written in the common subset of the
+ C99 and C++98 languages.
+ The library is usable with all major compilers and provides the necessary
+ tools to implement dedicated OPC UA clients and servers, or to integrate
+ OPC UA-based communication into existing applications.
+ .
+ This package provides the open62541 documentation as PDF and HTML
+ and example applications
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -86,3 +86,32 @@ add_dependencies(doc open62541)
 
 set_target_properties(doc doc_latex doc_pdf PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD TRUE)
 set_target_properties(doc doc_latex doc_pdf PROPERTIES FOLDER "open62541/doc")
+
+if (UA_PACK_DEBIAN)
+    set(open62541_install_debdoc_html_dir share/doc/open62541/open62541.html)
+    set(open62541_install_debdoc_dir share/doc/open62541)
+
+    set(open62541_debdoc_pdf_files "${DOC_LATEX_DIR}/open62541.pdf")
+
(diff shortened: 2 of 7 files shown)


doc/CMakeLists.txt
```diff
@@ -86,3 +86,32 @@ add_dependencies(doc open62541)
...
+    set(open62541_debdoc_pdf_files "${DOC_LATEX_DIR}/open62541.pdf")
+
+    install(DIRECTORY ${DOC_HTML_DIR}/
+        COMPONENT htmldoc 
+        DESTINATION ${open62541_install_debdoc_dir}/open62541.html 
+        USE_SOURCE_PERMISSIONS
+        FILES_MATCHING 
+        PATTERN "*.*"
+        PATTERN "_sources" EXCLUDE
+        PATTERN "_static/ua-wireshark-pubsub.png" EXCLUDE
```
Pro: Why do we need to exclude all these files?
What happens if the doc is updated and more images are added? Do we need to update this list then too?
I fear that this will lead to inconsistencies, since nobody will look into that CMakeLists.txt when changing the docs...

Can we somehow make this more generic?

ralphlange:
Thanks a lot @jrcoding - I am completely aware that I was dropping the ball on this one.
(Division reorganization, additional duties, ... I just didn't see a slot.)

doc/CMakeLists.txt
ralphlange: Maybe using a few specific patterns to include files would make it shorter and more generic?

debian/control-template
```diff
@@ -5,6 +5,13 @@ Build-Depends: debhelper (>= 9),
  cmake (>= 2.8),
  graphviz,
  python (>= 2.7),
+ python-sphinx,
+ python-sphinx-rtd-theme | base-files (<= 7.2ubuntu5.6~),
+ graphviz,
+ texlive-fonts-recommended,
```
jpfr: Are those dependencies meant for the docs package or only for the generation of the docs package?

jrcoding: these dependencies are Build-Dependencies, they are not needed or requested if one installs the docs

doc/CMakeLists.txt
jrcoding: There are a lot of leftovers after `cmake/sphinx` which should be definitely removed before installation of the docs.
It seems some excludes are not needed anymore because `dh_sphinxdoc` already does the job.

So we can reduce the excludes to 
```
        PATTERN "_sources" EXCLUDE
        PATTERN "CmakeFiles" EXCLUDE
```
This directories definitely never needed (as far as I can see) for the installed docs.
 I will update the PR ...

ralphlange:
@jpfr @Pro:
I still have a commit in my original branch that would make the debian build a separate travis job.
Is that interesting?

(Advantage: the "main" full host and cross build that currently also builds the debian packages would get shorter. Disadvantage: one more travis job.)

Pro:
> @jpfr @Pro:
> I still have a commit in my original branch that would make the debian build a separate travis job.
> Is that interesting?
> 
> (Advantage: the "main" full host and cross build that currently also builds the debian packages would get shorter. Disadvantage: one more travis job.)

Having a separate debian build job would definitely be nice.
Doing it on travis, not so nice. I recently migrated some jobs from Travis to Azure Pipelines, which is much faster and easier to handle.

Before you invest too much time in learning the basics of azure, I will just take over the job and split the debian build into a separate azure pipeline.

jpfr:
> Before you invest too much time in learning the basics of azure, I will just take over the job and split the debian build into a separate azure pipeline.

Great! Just merged this so you can add on top.
Let's merge 1.0 to master once we are happy with the packaging situation.

ralphlange:
Cool. Thanks a lot!
I will have a closer look are what you are doing with Azure, as I am about to start moving some Windows jobs from AppVeyor to Azure in my main work project.

(https://epics-controls.org/ which is hosted at https://github.com/epics-base)

jrcoding:
one more PR will follow, add static lib to dev package...

jrcoding:
@Pro 
seems that package `texlive-generic-extra` is not available 
The log is for an `Ubuntu Eoan` build, right?

The problem appears because of the problem mentioned here
https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1704124.html
and here
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=933290

We need to change the Build-Dependency to ` texlive-generic-extra | texlive-plain-generic`
to achieve  a stable build for older and newer Debian/Ubuntu flavors

I prepared PR #3216 , could you please try it on launchpad?

Issue #3215, opened 2019-10-26
launchpad builds failing


Pro:
Currently the launchpad builds are failing:

https://launchpadlibrarian.net/448601588/buildlog.txt.gz

@jrcoding this was intruduced in #3209

Could you check?

Pro:
Fixed in #3216

PR #3216, opened 2019-10-26
fix(debian):use texlive-plain-generic since debian/bullseye or ubuntu/eoan

jrcoding:
This PR addresses the problem reported by @Pro (launchpad builds failing [#3215](https://launchpadlibrarian.net/448601588/buildlog.txt.gz))

Since  Debian/bullseye or Ubuntu/Eoan the generation of the `open62541-doc` package fails because the Debian package `texlive-generic-extra` is not available anymore and should be replaced by `texlive-plain-generic`

@Pro could you please test it on launchpad?

### DIFF
--- a/debian/control-template
+++ b/debian/control-template
@@ -9,7 +9,7 @@ Build-Depends: debhelper (>= 9),
  graphviz,
  texlive-fonts-recommended,
  texlive-latex-extra,
- texlive-generic-extra,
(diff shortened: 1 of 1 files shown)


Pro:
Just testing a specific commit is not so easy. I will just merge this PR and then let's see.
Thanks for the quick fix!!

Issue #3199, opened 2019-10-23
nodeset-compiler should generate function stubs for method calls


cp-aquila:
## Description
I would like to see the nodeset-compiler to properly handle Method nodes.
Currently the following code is generated:
```c
static UA_StatusCode function_tf_2_finish(UA_Server *server, UA_UInt16* ns) {
#ifdef UA_ENABLE_METHODCALLS
return UA_Server_addMethodNode_finish(server, 
UA_NODEID_NUMERIC(ns[1], 7001)
, NULL, 0, NULL, 0, NULL);
#else
return UA_STATUSCODE_GOOD;
#endif /* UA_ENABLE_METHODCALLS */
}
```
This is missing:
* Input arguments + size
* Output arguments + size
* method function-pointer

My suggestion is to parse the argument info from the .xml and use it.
For the method we could add a prototype declaration to the generated .h file. The user then can add the actual implementation to its own code, using that prototype.
The question is how to name that prototype. Perhaps based on the hierachy of the node, so in this example it would be `methodCallback_Gateway_enterBootloader(.....)`.

This is a feature that I need, and I am willing to implement this myself and open a PR for it.
Right now I am looking for feedback:
* Is this feature already there and I just missed it?
* Is this a stupid idea because of $reason that I am not aware of?
* naming of the callback method
* Anything else I forgot?

## Background Information / Reproduction Steps
```./nodeset_compiler.py -e ../schema/Opc.Ua.NodeSet2.Minimal.xml -x test-function.xml foobar```
With the .xml I used: [test-function.xml.txt](https://github.com/open62541/open62541/files/3762518/test-function.xml.txt)

I did not build it myself. Instead the **1.0 release** was used.

## Checklist
 - [x] open62541 Version (release number or git tag):
 - [x] Other OPC UA SDKs used (client or server): UaModeler
 - [x] Operating system: Windows10
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [x] Self-contained code example attached
 - [ ] Critical issue: nope

jpfr:
Good idea!
Marked as a feature request.

PR #3219, opened 2019-10-28
NodesetCompiler: Add function stubs for methodnodes

cp-aquila:
* Function stubs are generated in the header file
* Calling functions does work (tested with 2x Uint32 Input, 1x Uint32 output)
* ~~Something is still wrong when using UAExpert:~~ grafik When using the default nodeset instead of minimal this error goes away.

Related: https://github.com/open62541/open62541/issues/3199


cp-aquila:
In the end I did not use this code but instead dynamically add nodes during runtime with `UA_Server_add?????Node`. 

With the nodeset-compiler I would have to add code to search for nodes to update them. Using a lookup table was much easier for my application.

PR #3228, opened 2019-10-30
[Review] Add readModified, readProcessed, readAtTime to UA_HistoryDatabase

andranik-s:
This PR adds interfaces for readModified, readProcessed, and readAtTime history requests so that users have ways to implement those requests in their custom history databases. 
No default implementations are provided.


jpfr:
Nice. Looks really good.

PR #3229, opened 2019-10-30
Fix discovery url registration

Volksfest:
Earlier the discovery urls are generated by the server configuration
and the network layers. Now only the server configuration is used.
The network layers are only used as a fall back if the server
configuration does not provide discovery urls.

### DIFF
--- a/include/open62541/plugin/network.h
+++ b/include/open62541/plugin/network.h
@@ -152,6 +152,9 @@ struct UA_ServerNetworkLayer {
     /* Points to external memory, i.e. handled by server or client */
     UA_NetworkStatistics *statistics;
 
+    /* Describes the url of the network layer which is used by the remote to connect to this network layer.
+     * The url will be used for the discovery and endpoint services.
+     */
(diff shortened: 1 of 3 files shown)


jpfr:
Nice!
This could surprise some users.
Can you add documentation to the server config header?

Volksfest:
To be honest, I actually have no idea where the documentation exactly should be and what it should contain as this fix only removes some redundant work.

jpfr:
Two places where this should be added as a comment.

The ApplicationDescription in the server config:
https://github.com/open62541/open62541/blob/c5d7624115d35bd053e7571a121497267b184db5/include/open62541/server_config.h#L96

The definition of the DiscoveryUrl in the network layer:
https://github.com/open62541/open62541/blob/c5d7624115d35bd053e7571a121497267b184db5/include/open62541/plugin/network.h#L153

A few sentences at each location should be enough.
With a cross reference between to say when one will override the other...

Thanks!

PR #3231, opened 2019-10-31
feat(doc): example debian packaging inside docker container

jrcoding:
This PR adds documentation regarding Debian packaging and has two parts
- an example how one can locally build open62541 Debian packages inside a Docker container in a convenient way
- a small comment on how to deal with CMake build options in conjunction with Debian packaging

please feel free to slice and dice it - or to dump it :-)

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -111,6 +111,79 @@ The procedure below works on OpenBSD 5.8 with gcc version 4.8.4, cmake version 3
    cmake ..
    make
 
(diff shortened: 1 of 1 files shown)


Pro:
Nice idea with the debian packaging in docker!
Maybe we should use this approach instead of the direct Travis approach here:
https://github.com/open62541/open62541/blob/master/tools/travis/travis_linux_script.sh#L450

PR #3251, opened 2019-11-07
[Review] Historical events

andranik-s:
This PR adds two new methods to UA_HistoryDatabase: setEvent and readEvent.

It also adds readObjectProperty function which is then used in historical events' implementation. (UA_Server_readObjectProperty/readObjectProperty are now analogous to UA_Server_writeObjectProperty/writeObjectProperty).

Some notes:

1. OPC UA Part 11, clause 5.3.2 says: "A HistoricalEventNode that has Event history available will provide the Property [HistoricalEventFilter]". When an event is triggered the server tries to read the HistoricalEventFilter property of an emitting node (not the origin node) and if the property is present it filters the event object and passes results to setEvent (happens for each emitter). On readEvent the client's filter should be mapped to a filter contained in the HistoricalEventFilter to costruct results properly (ideally some utility function should be added for that).

2. It is possible to historize events without setting HistoricalEventFilter because setEvent is always called and it has eventId as a parameter.

3. willEventNodeBeDeleted is added for the following scenario:
users can implement a very simple in-memory event history database by leaving event nodes in the nodestore; on readEvent they can get the relevant nodes and filter them against a client's filter.

4. It may be justified to make UA_Server_filterEvent public to facilitate the use cases 2 and 3.

### DIFF
--- a/include/open62541/plugin/historydatabase.h
+++ b/include/open62541/plugin/historydatabase.h
@@ -39,6 +39,37 @@ struct UA_HistoryDatabase {
                 UA_Boolean historizing,
                 const UA_DataValue *value);
 
+    /* This function will be called when an event is triggered.
+     * Use it to insert data into your event database.
+     * No default implementation is provided by UA_HistoryDatabase_default.
+     *
+     * server is the server this node lives in.
+     * hdbContext is the context of the UA_HistoryDatabase.
+     * sessionId and sessionContext identify the session which set this value.
+     * originId is the node id of the event's origin.
+     * emitterId is the node id of the event emitter.
+     * eventId is the node id of the event that is being emitted.
+     * willEventNodeBeDeleted specifies whether the event node will be deleted after
+     *                        it has been triggered (this might be relevant for
+     *                        in-memory storage).
+     * historicalEventFilter is the value of the HistoricalEventFilter property of
+     *                       the emitter (OPC UA Part 11, 5.3.2), it is NULL if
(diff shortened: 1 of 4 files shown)


andranik-s:
@jpfr ci/dockercloud failed for some reason but the "Details" link leads to 404, so I'm not sure what I'm supposed to do with it. Apart from that the PR is ready for review.

Pro:
> @jpfr ci/dockercloud failed for some reason but the "Details" link leads to 404, so I'm not sure what I'm supposed to do with it. Apart from that the PR is ready for review.

Rebasing onto current master should fix the dockercloud issue

jpfr:
This is still marked as WIP. Can we merge?

andranik-s:
@jpfr The reason it is still marked as WIP is that I was planning to simplify the implementation. I wanted to delete the code that reads the HistoricalEventFilter property of an object from UA_Server_triggerEvent and make UA_Server_filterEvent (or some version of it) public. That would give more control to users regarding how they want to store and use the filter.

However, if you are happy with the current implementation you can merge. I will delete the WIP status.

jpfr:
The new public API looks solid.
So I'm merging to keep individual PR small.

I haven't fully understood making `UA_Server_filterEvent` public.
Do you have a user scenario for that?

We now have new functionality that we cannot cover with unit tests.
Non-testable parts of the SDK will break over time.
Can we have a very basic historical event storage for unit testing?
It can be just an inefficient linked list.

Issue #3232, opened 2019-10-31
Question: 'Medium' nodeset build


cp-aquila:
## Description
I am developing a UA server application to be run on a Atmel SAME55 CPU. It has only limited SRAM (256k), so I have to be considerate about it.
I am using the UAModeler and the nodeset-compiler to add server-nodes.

I would like to have a *Medium* Nodeset that is as minimal as possible but still supporting the features needed for custom object instances.
With the minimal-modeset the instances of the `Gateway` nodes dont show up when running the server. With the normal nodeset that error is fixed, but I don't have enough RAM for that.

* Can I expand the minimal nodeset with the required features for this to work? 
* If yes, which features do need to be included? 
* How would I build/compile such a base nodeset?

## Background Information / Reproduction Steps
Screenshot of UAModeler:
![grafik](https://user-images.githubusercontent.com/20518746/67964243-d9bc4900-fbf7-11e9-8e6e-1d6d4cf1ded4.png)

## Checklist

Please provide the following information:

 - [x] open62541 Version: 1.0.0
 - [x] Other OPC UA SDKs used (client or server):  UAModeler, UAExpert
 - [x] Operating system: freeRTOS
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue: nope

cp-aquila:
I made some progress on that.
The real issue is the wording of the cmake options:
```# Namespace Zero
set(UA_NAMESPACE_ZERO "REDUCED" CACHE STRING "Completeness of the generated namespace zero (minimal/reduced/full)")
SET_PROPERTY(CACHE UA_NAMESPACE_ZERO PROPERTY STRINGS "MINIMAL" "REDUCED" "FULL")
if(UA_NAMESPACE_ZERO STREQUAL "MINIMAL")
    set(UA_GENERATED_NAMESPACE_ZERO OFF)
else()
    set(UA_GENERATED_NAMESPACE_ZERO ON)
endif()
```

When selecting `REDUCED` the file `Opc.Ua.NodeSet2.Minimal.xml` will get compiled and included. That is misleading, as a dumb user I would expect that the option `MINIMAL` builds the file that has minimal in its name.

I suggest to change this is a next major release. The wording could be somthing like this:
* NANO
* MINIMAL (default)
* FULL

Pro:
[...]

PR #3259, opened 2019-11-08
fix(ns): Rename Minimal.xml to Reduced.xml to match CMake options

Pro:
See also #3232

@jpfr please merge if you agree


jpfr:
Looks good to me.

PR #3262, opened 2019-11-11
Fix type of resuse and loopback options to UA_TYPES_BOOLEAN in tutorial_pubsub_connection.c example

tiropp:
The `resuse` and `loopback` options are defined as type UA_TYPES_UINT32 in the example, but in UA_PubSubChannelUDPMC_open [plugins/ua_pubsub_udp.c] are expected to be of type
BOOLEAN. Using UINT32 for either of them will cause the options not to
be evaluated and hence not used.


jpfr:
Thanks!

PR #3266, opened 2019-11-12
Included plugin/log.h cleanup in UA_Client_delete(..).

wbitesi:
The log plugin provides a cleanup routine which is not called in UA_Client_delete(..). Since automatic cleanup can be enabled/disabled by setting UA_Logger.clear function pointer, UA_Client_delete(..) should automatically cleanup the log plugin.


jpfr:
Great! Thanks.

PR #3270, opened 2019-11-12
Add UA_calloc param size check

MartinaFelberFEM:
Add checks to avoid calling UA_calloc with param num == 0.
Because the return value of UA_calloc is in this case implementation defined.


MartinaFelberFEM:
Hi!
I have a question about the integration checks. I am having problems with the pipeline Windows (VS2017). It often fails and I don't know why. And the strange thing is: If I remove my commits, push and add them again, all checks pass.
Does anybody know more about this?

jpfr:
Very nice! Merged.

CI has improved on the current master.
Please report if problems remain.

PR #3277, opened 2019-11-14
Solved issues related to SecurityToken renewal

jonasgreen88:
* previousSecurityToken structure of UA_SecureChannel was not cleared
after generating the new remote symmetric key. This caused a server
to renew the remote key every time a message chunk was received. If
the client requested a renewal of the secure channel and new nonces
were exchanged, but the client continued to use the "old"
SecurityToken for some more message chunks the server generated a
remote key based on the new nonces when it should have used the key
of the current SecurityToken.
* UA_SecureChannelManager_cleanupTimedOut() changed currently used
SecurityToken to the next one if present. According to Part 4,
chapter 5.5.2 OpenSecureChannel the server shall use the existing
SecurityToken until it expires or the server receives a message
secured with the new SecurityToken. Hence, it shouldn't trigg itself
change the SecurityToken.
* Fixed issues in renew channel test cases of the client

### DIFF
--- a/src/server/ua_securechannel_manager.c
+++ b/src/server/ua_securechannel_manager.c
@@ -87,11 +87,6 @@ UA_SecureChannelManager_cleanupTimedOut(UA_SecureChannelManager *cm,
             removeSecureChannel(cm, entry);
             continue;
         }
-
-        /* Revolve the channel tokens */
-        if(entry->channel.nextSecurityToken.tokenId > 0) {
-            UA_SecureChannel_revolveTokens(&entry->channel);
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


jpfr:
Looks good!
One of the builds fails with this message. Could be a false positive though...
```
/home/travis/build/open62541/open62541/src/client/ua_client_subscriptions.c:605:5: warning: Potential memory leak
    return;
    ^~~~~~
1 warning generated.
```

jpfr:
Can you do a simple rebase on master?
The build issue was fixed...

jpfr:
Perfect. This is backported to 1.0 as well.

Issue #2707, opened 2019-05-06
STM FreeRtos LWIP mbedtls


embbo:
## Description

I am trying to compile for stm32 (LWIP ,Freertos) certificate using mbetls libraries using Eclispe. 
-> Its works fine if I Disabaled  //#define UA_ENABLE_ENCRYPTION
-> after giving all the paths of mbedtls library i still have problem ""mbedtls_platform_entropy_poll' undeclared (first use in this function)""
Can anyone help about it .

Used CMake options:

../cmake .. -G "Eclipse CDT4 - Unix Makefiles" -DUA_ARCHITECTURE=freertosLWIP  -DUA_ENABLE_ENCRYPTION=ON -DUA_ENABLE_AMALGAMATION=ON -DMBEDTLS_INCLUDE_DIRS="/../mbedtls/mbedtls-development/build/include" -DMBEDTLS_LIBRARY="/../mbedtls/mbedtls-development/build/library" -DMBEDX509_LIBRARY="/../mbedtls/mbedtls-development/build/library" -DMBEDCRYPTO_LIBRARY="/../mbedtls/mbedtls-development/build/library"

## Checklist

Please provide the following information:

 - [ ] open62541 Version (release number or git tag):
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

embbo:
@Pro

Pro:
@cabralfortiss maybe you can help out with stm32 stuff?

cabralfortiss:
I never worked with encryption but I gave it a quick try. I pulled the https://github.com/ARMmbed/mbedtls repo, and tried to compiled to my platform. The error I receive is:

#error "Platform entropy sources only work on Unix and Windows, see MBEDTLS_NO_PLATFORM_ENTROPY in config.h"

So I think that's the problem. I took the configs/config-no-entropy.h (the name says literally that it has no entropy) paste it in the include/ folder, renamed as config.h and compiled without any problem. (this isn't news, the procedure is described in the repo). 

[...]

PR #3284, opened 2019-11-18
feat(mbedtls): Allow usage of hardware entropy poll function on microcontrollers

Pro:


saperlot:
@Pro i just worked on this issue without knowing that this was tried to fix. I commented this issue #2707. You could also just not add the source since it will be added anyway in the init function  depending on the mbed configuration. It is within `mbedtls_entropy_init`. This fix leads to 2 identical entropy sources in the context.

PR #3286, opened 2019-11-19
feat(core): Allow setting custom ip address list if platform does not support getifaddr

Pro:
This is the case e.g. on an ESP32 chip


jpfr:
Can we put `UA_HAS_GETIFADDR` in the architecture definitions?

I presume this will be "fixed" for every architecture and not something the end user has to configure.

Pro:
Good Idea. PR is updated

jpfr:
Cool!

PR #3287, opened 2019-11-19
Update pubsub_interrupt_publish.c

just520fun:
line 222: Comparing to line 186, the tv_sec should be second. so "%" is typo, should be replaced by "/". FYI.


jpfr:
Good find!
Can you sign the CLA?

PR #3289, opened 2019-11-20
Compability with generated types of UnifiedAutomation SDK

wbitesi:
We observed generated binary encoding IDs that do not fit into an UA_UInt16, e.g. ns=2;i=543211 => Increasing UA_DataType.binaryEncodingId to UInt32.

In addition, added EnumDefinition and EnumField to decode datatype binary information (UA_ATTRIBUTEID_DATATYPEDEFINITION) from OPC UA servers.

### DIFF
--- a/include/open62541/types.h
+++ b/include/open62541/types.h
@@ -869,7 +869,7 @@ struct UA_DataType {
     UA_UInt32 overlayable      : 1;  /* The type has the identical memory layout
                                       * in memory and on the binary stream. */
     UA_UInt32 membersSize      : 8;  /* How many members does the type have? */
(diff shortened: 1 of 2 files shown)


jpfr:
`ns=2;i=543211` is problematic on many levels.
We also don't store explicit namespaces for the binary encoding.

This is good to merge. But probably we have to switch to full NodeIds...

wbitesi:
Not storing the namespace does work in the current implementation, if the namespace index of the binary encoding id and the nodeid of the datatype is the same UA_findDataTypeByBinaryInternal (ua_types_encoding_binary.c).

jpfr:
Very good!

PR #3290, opened 2019-11-20
Do not overwrite the UA_Client logger when using UA_ClientConfig_setD…

wbitesi:
…efault

Default logger is set in UA_Client_new(). If the logger is replaced after UA_Client creation to a custom logger, and UA_ClientConfig_setDefault(Encryption) is called, the custom logger is overwritten.

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -679,9 +679,11 @@ UA_ClientConfig_setDefault(UA_ClientConfig *config) {
     config->timeout = 5000;
     config->secureChannelLifeTime = 10 * 60 * 1000; /* 10 minutes */
 
-    config->logger.log = UA_Log_Stdout_log;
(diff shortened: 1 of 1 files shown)


wbitesi:
Without this patch I always receive the following message in the default Stdout logger instead of the custom logger:
[2019-11-20 11:03:38.920 (UTC+0100)] warn/userland      AcceptAll Certificate Verification. Any remote certificate will be accepted.

jpfr:
I would do this only if `config->logger.log != NULL`.
Otherwise, some users might suddenly end up with no logger at all!

wbitesi:
Updated: default Stdout logger is set if the config->logger.log is NULL when calling UA_ClientConfig_setDefault(..)

jpfr:
Cool! Merged.

PR #3307, opened 2019-12-02
Refactor: Replace deprecated FindPythonInterp with FindPython

mlgiraud:


jpfr:
It seems this is problematic on older systems.

```
 Could not find a package configuration file provided by "Python" with any

  of the following names:

    PythonConfig.cmake

    python-config.cmake
```

jpfr:
Closing, since we seem to not get this working on Travis.
Please reopen if there is a solution after all.

PR #3309, opened 2019-12-03
Fix: (client async (dis)connect)

wbitesi:
* Removed blocking select call in UA_ClientConnectionTCP_poll() which makes UA_Client_connect_async/UA_Client_run_iterate(..) to use the UA_Connection->connectCallback repeatably 
* UA_Client_disconnect_async(..) waits for the sessionClose Response from the server before closing the TCP connection. Requires UA_Client_run_iterate(..) != UA_STATUSCODE_GOOD for finishing the async handling of the client


jpfr:
Nice!
Can you take a look at the Codacy warnings?

jpfr:
The client logic was fully refactored in the 1.1 cycle.
Please rebase and reopen if this has still relevant changes.

PR #3310, opened 2019-12-03
CMake: Allow UA_NODESET_DIR to be custom defined.

besser82:
This makes it easier for developers to keep all ua-nodesets XML definitions in a common location.


jpfr:
You don’t really need the definition of open62541_NODESET_DIR.
Right?

Pro:
That separate definition is required for the public macros:
https://github.com/open62541/open62541/blob/92e0e8ca8ba036058231a1320165b785b4755c39/tools/cmake/macros_public.cmake#L449

jpfr:
The macros can use UA_NODESET_DIR just as well. We define what the macros use.

We should avoid duplicate variables. And we should avoid using different prefixes (UA_ vs open62541_).

Pro:
> The macros can use UA_NODESET_DIR just as well. We define what the macros use.
> 
> We should avoid duplicate variables. And we should avoid using different prefixes (UA_ vs open62541_).

I agree that this is a bit inconsistent. If we change it now, we should make sure that we keep backwards compatibility and don't introduce a breaking change.

jpfr:
> I agree that this is a bit inconsistent. If we change it now, we should make sure that we keep backwards compatibility and don't introduce a breaking change.

Then lets aim for a clean solution over the next 3 releases.
Overlapping definitions in the build scripts should be avoided.

1. Mark the open62541_ variables as deprecated.
Copy open62541_ to UA_. Issue a warning if both are manually set.

2. Issue a warning if open62541_ is used. Don't use it otherwise.

3. Remove the warning.

andreasebner:
Thank you for contributeing! Feature merged in #5138

PR #3321, opened 2019-12-11
feature: use lld if available to speed up linking

mlgiraud:
Using lld for linking saves about 10 seconds on a full build on my machine.
~I could add an option such that a user can disable lld if desired.~
There is now an advanced option to either use gold, or the default linker.
By default lld is used when it is available.


jpfr:
We should both
- have a workflow that is consistent on all target platforms
- provide good defaults.

You are always testing if the selected linker is available.
We can use the same for a "select best" automatic option.

- Automatic (check availability of Gold, LLD, selects the best linker or the system-default)
- System-Default (does nothing)
- Gold (force gold, don't check availability)
- LLD (force lld, don't check availability)

On a non-Linux system, I would always choose the system-default.

We can put the entire sub-script into a separate file in /tools/cmake.
And set in the parent via (set <VARIABLE> <CONTENT> PARENT_SCOPE).
Otherwise the main CMakeLists.txt gets very convoluted.

PR #3327, opened 2019-12-19
Fix: Error message in triggerConditionEvent

ccvca:
As EnableState.Id is checked instead of the Retain when triggering a condition, I changed the error message.
https://github.com/open62541/open62541/blob/a2209072c72ed25918fbd00f19b14e491745a90b/src/server/ua_subscription_alarms_conditions.c#L2549-L2552

https://github.com/open62541/open62541/blob/a2209072c72ed25918fbd00f19b14e491745a90b/src/server/ua_subscription_alarms_conditions.c#L628


jpfr:
Thanks!

Issue #2136, opened 2018-10-23
Add statistics counters to the stack to make troubleshooting easier


jonasgreen88:
Description
===========
Our Ethernet devices often has a set of statistics counters presented on the internal webpage or through some interface of the protocol. When it comes to Ethernet itself there are interface and media counters. When it comes to specific protocols it can be counter about number of clients connected, how many IO connections that are active etc. It can also be counter indicating errors, such as rejected connections.

My question is if you guys have though about adding something similar to the Open62541 stack. Every part of the stack could have its own set of counters (e.g. secure channel layer, session layer, DataChange subscription implementation etc.). It can be both counter indicating good things but also bad things, such as rejected clients.

The counters should be available through some public API so they can be published on e.g. an internal webpage.

If it sounds interesting to add this feature we maybe can discuss how it should be implemented in this issue. We see this as an important feature for troubleshooting and might be implementing it sooner or later. To prepare for that it would be good to agree upon a solution.

I have two initial proposals how to implement this feature:
1. Add statistics counters to each layer/implementation of the stack that have relevant counters to present. For example, counters of the sessions is implemented in the session layer. Each layer/implementation offers a public API to access the counters.
2. Add a statistics implementation of the stack containing all different counters and it offers a public API to access them. The different layers/implementations of the stack have to interface with the statistics implementation to increment/reset the counters.

Please come up with other proposals or ideas about this feature!

Background Information / Reproduction Steps
===========================================
Statistics counters for troubleshooting devices implementing this stack.

Checklist
=========
Please provide the following information:

 - [ ] open62541 Version (release number or git tag):
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jpfr:
Yes. This is an important feature for professional use.

Would the following API be enough?
Then we can keep a statistics structure in the server/client.
And just give a pointer to the structure that is continuously updated in the background.

```c
struct UA_NetworkStatistics {
    size_t currentConnections;
    size_t closedConnections;
};
struct UA_SecureChannelStatistics {};
struct UA_SessionStatistics {};

struct UA_ServerStatistics {
    UA_NetworkStatistics ns;
    UA_SecureChannelStatistics scs;
   UA_SessionStatistics ss;
};

[...]

PR #3341, opened 2020-01-07
[Review] Added statistic counters for client and server

jonasgreen88:
Counters have been added for the network, secure channel and session
layer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -628,6 +628,7 @@ set(exported_headers ${exported_headers}
                      ${PROJECT_BINARY_DIR}/src_generated/open62541/types_generated.h
                      ${PROJECT_BINARY_DIR}/src_generated/open62541/types_generated_handling.h
                      ${PROJECT_SOURCE_DIR}/include/open62541/util.h
+                     ${PROJECT_SOURCE_DIR}/include/open62541/statistics.h
                      ${PROJECT_SOURCE_DIR}/include/open62541/server.h
                      ${PROJECT_SOURCE_DIR}/include/open62541/plugin/log.h
                      ${PROJECT_SOURCE_DIR}/include/open62541/plugin/network.h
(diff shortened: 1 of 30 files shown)


jonasgreen88:
I have defined a proposal of a set of statistic counters for the network, the secure channel and the session layer according to the discussion in issue #2136.

Open for review to get the counter setup agreed before adding unittests.

mlang-de:
Nice to see such PR!
IMHO it is the first step into direction of the specified ServerDiagnostics.
So if it make sense the specified names shall be used where possible?!

image

jpfr:
Very cool feature!

I agree with @mlang-de that we should stay close to the counters defined in the standard.

Also, can the `_getStatistics` methods directly return the struct instead of pointers?
That's one source for bugs less. ;-)

Keep up the good job!

jonasgreen88:
@jpfr @mlang-de Thanks for the feedback and the hint about the ServerDiagnosticsSummaryType. The session counters are now reflecting this type completely. The network and secure channel counters are harmonizing with the session counters as far as possible.

jpfr:
> The session counters are now reflecting this type completely.

Perfect PR. We can merge when CI goes green.

jonasgreen88:
Solved all obvious CI errors. 

Now there is some problem with Sphinx and documentation generation on Travis:
```
/home/travis/build/open62541/open62541/build/doc_src/nodeset_compiler.rst:383: ERROR: Unknown interpreted text role "numref".
/home/travis/build/open62541/open62541/build/doc_src/protocol.rst:92: ERROR: Unknown interpreted text role "numref".
```

Also clang fails on Travis in some file this PR doesn't touch:
```
/home/travis/build/open62541/open62541/src/client/ua_client_subscriptions.c:607:5: warning: Potential memory leak
    return;
```

Azure past before the last commit, now there is some error I can't figure out. It says all unittests passed.

jpfr:
See #3399.

jonasgreen88:
Merged in #3399

PR #3347, opened 2020-01-08
MSYS2/MinGW build fixes

emersonfxbx:
Hi,
I've made some changes in order to build properly the library on MSYS2/MinGW (as static or shared library (DLL))
- I've added a dependency to GCC libssp. I've got some unsatisfied link errors without it when building a DLL (cmake command line: -DBUILD_SHARED_LIBS=ON)
- The RUNTIME install destination (make install) was changed to the "bin" directory. The libopen62541.dll was being copied to the root of install prefix (in the case of MSYS2/MinGW, it was "/mingw32" or "/mingw64")
- A check for NULL parameter was added to src/ua_types.c. The current version of gcc, with CMakeLists.txt configured options, on MSYS2/MinGW treat as an error the use of the second argument of memcmp() call without checking for NULL.


jpfr:
Thanks a lot for the fix.

I'm a bit unsure about the change
```
-        RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}
+        RUNTIME DESTINATION bin
```
Wouldn't it make more sense to set the prefix correctly in the individual cmake configuration?
This could break other architecture that rely on CMAKE_INSTALL_PREFIX.

emersonfxbx:
Hello,

The cmake documentation (https://cmake.org/cmake/help/v3.13/command/install.html#targets) says the runtime destination config is suitable for binaries (in the case of open62541 library, its DLL when BUILD_SHARED_LIBS=ON).
Without this change, the DLL will be copied to the root of the install prefix (/mingw[32|64] for MSYS2, /usr/local or /usr for cygwin).
Inside MSYS2 and cygwin, the bin directory is always in the PATH environment (and is the directory where other DLLs will be found like the mbedtls ones).

Enviado do Email<https://go.microsoft.com/fwlink/?LinkId=550986> para Windows 10

De: Julius Pfrommer<mailto:notifications@github.com>
Enviado:quinta-feira, 9 de janeiro de 2020 13:37
Para: open62541/open62541<mailto:open62541@noreply.github.com>
Cc:Emerson de Freitas Barcelos<mailto:emersonfxbx@msn.com>; Author<mailto:author@noreply.github.com>
Assunto: Re: [open62541/open62541] MSYS2/MinGW build fixes (#3347)


Thanks a lot for the fix.

I'm a bit unsure about the change

-        RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}

+        RUNTIME DESTINATION bin

Wouldn't it make more sense to set the prefix correctly in the individual cmake configuration?
This could break other architecture that rely on CMAKE_INSTALL_PREFIX.

jpfr:
> Inside MSYS2 and cygwin, the bin directory is always in the PATH environment (and is the directory where other DLLs will be found like the mbedtls ones).

Understood. But you make the change for all environments, not only for MSYS2/Cygwin.

What happens on Linux?
It would be bad if binaries intended for /usr/local/bin (via the install prefix) end up in /usr/bin instead.

traversaro:
> 
> 
> > Inside MSYS2 and cygwin, the bin directory is always in the PATH environment (and is the directory where other DLLs will be found like the mbedtls ones).
> 
> Understood. But you make the change for all environments, not only for MSYS2/Cygwin.
> 
> What happens on Linux?
> It would be bad if binaries intended for /usr/local/bin (via the install prefix) end up in /usr/bin instead.

For consistency I suggest to set as `RUNTIME DESTINATION` the `${CMAKE_INSTALL_BINDIR}`, that is defined by [`GNUInstallDirs`](https://cmake.org/cmake/help/v3.16/module/GNUInstallDirs.html) and would be consistent with the `${CMAKE_INSTALL_LIBDIR}` variable already used.

jpfr:
Sounds like a good compromise. Thanks @traversaro .

@emersonfxbx Does `CMAKE_INSTALL_BINDIR` work for you?

emersonfxbx:
> Sounds like a good compromise. Thanks @traversaro .
> @emersonfxbx Does CMAKE_INSTALL_BINDIR work for you?

Yes. I'll push the change to my branch. Is it going to update the pull request ? This is my first contribution on github. I'm still learning how to use the tools...

PR #3349, opened 2020-01-09
[REVIEW] Added possibility to configure max TCP connections limit

jonasgreen88:
In for example LWIP the number of TCP connection resources must be
configured at compile time. On a device with several protocols that
relies on TCP they must share this common pool of TCP resources. To
avoid one protocol allocating all TCP resources they must limit
themselfs in the number of TCP connections they allocate.

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -236,7 +236,6 @@ ServerNetworkLayerTCP_add(UA_ServerNetworkLayer *nl, ServerNetworkLayerTCP *laye
     /* Allocate and initialize the connection */
     ConnectionEntry *e = (ConnectionEntry*)UA_malloc(sizeof(ConnectionEntry));
     if(!e){
-        UA_close(newsockfd);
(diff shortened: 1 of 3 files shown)


jpfr:
Hey @jonasgreen88 ,

very nice!

Thee comments on the PR.

- The interrface to generate the networklayer has changed.
Please add a comment to describe the new API. Users might otherwise not understand that they can put zero to allow any numbe of connections.
- Removing the UA_close(newsockfd); looks suspicious in the PR diff.
Can you check that it really is a duplicate?
- The PR has to be rebased on master. Probably due to the statistics functionality that was just merged.

Very cool to see all the loose ends being gradually tied up.

jpfr:
And one more thing.

If the server gets flooded with SecureChannels, we `prune` the oldest SecureChannel that has no session attached.

Maybe we can do the same for the Connections. When we reach the maximum number of connections, we remove the oldest connection that has no SecureChannel attached...

jpfr:
Very cool! Merged.

PR #3355, opened 2020-01-10
Various fixes for WebSocket layer

andranik-s:
1. Fix memory access after free
2. Clean up message queue on connection free
3. Prepend the incomplete chunk in WebSocket layer

The last fix is necessary because the following code is actually broken (an issue will be created) https://github.com/open62541/open62541/blob/d02fc4f5a4b60bb30a3c9394643abdd93b1fd59c/src/ua_connection.c#L171-L174


jpfr:
Thanks!

PR #3363, opened 2020-01-14
[WIP] XDP integration in OPC UA Pub/Sub

opcua-tsn-team-kalycito:
This pull request integrates XDP with open62541.
XDP is a BPF based high performance data path which bypasses TCP/IP layers and tranfers the frames from hardware/netdev to user application

 

 - Run time triggered PubSub application using ETF at a publishing interval of 100 microseconds with qbv offset of 25us
 - Integrate time triggered functionality and publish OPCUA PubSub packets at calculated transmission time
 - Application design change to run Publisher and Subscriber in same core(core 2)
 - User application runs in core 3
 - Integrate FPM Publisher application
 - Quick process on the received Network Message using a global variable
 - Configure XDP socket to receive frames from a specified queue
 - Follow **README.txt** file to work on this commit
 
 - TODO:
     Implement spin locks for threads
     Handling all datatypes in Subscriber receive
     Processing of networkUrl in user application
     Write unit test cases to ensure no memory leaks are possible
     Multiple PubSub connection handling


vincentxw:
Hello there, what a great start to bringing AF_XDP into open62541!

I took a look and would like some clarifications on the samples provided.

- In TSN_XDP_mps_loopback.c, is the received data being passed from the subscriber to the publisher, or is the data consumed by the subscriber and where the publisher sends something different back? Do correct me, but from what i can understand its the former?

- Any reason why you're using a VLAN interface instead of the ua_server's built-in VLAN tagging?

I also want to raise some design questions (current and future) since this is the first revision and AF_XDP is a complex and very different socket interface.

- Since this is AF_XDP receive only, would the next revision include AF_XDP transmit? Since you've already come this far in implementing a new transport layer, i dont imagine the transmit to be much different to implement .

- libbpf is statically linked now. Would the next version be using a dynamically linked or is there some intentions/restrictions to use it statically. 

- There a requirement for xdpsock_kern.o, which i can only compile using the kernel source and provide through a specific directory. This prompts me 3 questions: 
  - xdpsock_kern is only used when XDP_SHARED_UMEM is used. Should there be an option to select a subscriber which does not use XDP_SHARED_UMEM, hence reducing this requirement?

  - Should users be able to provide the BPF program themselves (e.g. provide a .c sample in tutorials where user can compile on their own) and specify it via a ReaderGroup Config (since only subcribers would use it)?

  - Should open62541 carry this BPF program within it during its compilation instead? Since it must be this specific xdpsock_kern bpf program and most users may not have a different code for it.

- Couple of things are hardcoded for now (which is fine), such as the publisher's SO_PRIORITY and XDP queue. Will there be reader/writer-group config options created for them? I would imagine this configurability  is crucial for a final implementation.

jpfr:
Merged via #3517.

PR #3370, opened 2020-01-15
[REVIEW] Fix(nc): Enable child nodes of method nodes with UA_ENABLE_METHODCALLS

jonasgreen88:
Child nodes of methods nodes are not affected by the
UA_ENABLE_METHODCALLS in nodesets compiled by the
nodeset_compiler.py.
If UA_ENABLE_METHODCALLS is not set, the method nodes won't be
created but child nodes of the method nodes will. These child nodes
will fail during initialization as their parent node doesn't exist.

An example of this in Opc.Ua.NodeSet2.Reduced.xml is the
GetMonitoredItems node that has two child nodes; OutputArguments and
InputArguments.

This change makes child nodes of method nodes to only be initialized
if UA_ENABLE_METHODCALLS is defined.

### DIFF
--- a/tools/nodeset_compiler/backend_open62541.py
+++ b/tools/nodeset_compiler/backend_open62541.py
@@ -238,7 +238,7 @@ _UA_END_DECLS
                     writec("\n".join(code_global))
                     writec("\n")
                 writec("\nstatic UA_StatusCode function_" + outfilebase + "_" + str(functionNumber) + "_begin(UA_Server *server, UA_UInt16* ns) {")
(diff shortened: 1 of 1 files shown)


jonasgreen88:
@jpfr Forgot to mention this issue on last core-dev meeting. But it would be nice to have merged until next release if possible.

jonasgreen88:
Bump. Possible to merge for 1.3?

jpfr:
Cool,  merged to 1.3.

PR #3371, opened 2020-01-15
allow running multithreading related test when UA_MULTITHREADING is greater then 100

damijans:


jpfr:
Nice!
I squashed the commits into one to keep a clean history.

PR #3373, opened 2020-01-15
support multiple conditions

ursfassler:
I was not able to add multiple (unexposed) conditions. One problem was an error in the handling/detection of condition branches.
The more difficult problem occurs when adding unexposed conditions with the same condition source. In this case, we have to make sure that a HASCONDITION reference is only added once from the condition source to the condition type. I did not found a way to check if this reference already exists. Therefore I just ignore this error.


jpfr:
Thanks @ursfassler !

PR #3374, opened 2020-01-15
fix(server): Correctly set discovery url in GetEndpoints AppDescription

Pro:


jpfr:
You should check if UA_String_new() succeeds.
Or you can use UA_Array_copy to allocate the array and copy the one-element content.

PR #3397, opened 2020-01-27
Optimize Condition Refresh Methods

sqadasi:
Refresh start and end events will be created only once when the first condition is created. Refresh events will be reused to save the time used to create those two events each time refresh methods get called.


jpfr:
Thanks.
Is this WIP or ready to merge?

Overall, the functionality should be covered by a unit test.
Otherwise, the next developer in line will break things.

sqadasi:
@jpfr I would say it is ready to merge.
Regarding the unit test I agree with you completely. I will try to prepare it in the next months.

PR #3399, opened 2020-01-27
Server statistics

jpfr:
Rebase of #3341 for discussion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -628,7 +628,7 @@ set(exported_headers ${exported_headers}
                      ${PROJECT_SOURCE_DIR}/include/open62541/architecture_definitions.h
                      ${PROJECT_BINARY_DIR}/src_generated/open62541/statuscodes.h
                      ${PROJECT_BINARY_DIR}/src_generated/open62541/nodeids.h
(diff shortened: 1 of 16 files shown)


jpfr:
A rebase of #3341 had become necessary and I didn't want to write over the original PR.

The kudos still belong to @jonasgreen88.
So the commit author will be reassigned before merging.

This is currently not complete.
- The NetworkLayer statistics are missing. My hope is to find a solution where the statistics does not leak into the API of every Connection. The NetworkLayer mostly knows internally when a close is not "regular".
- The client statistics are missing (the SecureChannel and Session statistics are not very meaningful here).

Up for discussion.

jonasgreen88:
> A rebase of #3341 had become necessary and I didn't want to write over the original PR.
> 
> The kudos still belong to @jonasgreen88.
> So the commit author will be reassigned before merging.
> 
> This is currently not complete.
> 
> * The NetworkLayer statistics are missing. My hope is to find a solution where the statistics does not leak into the API of every Connection. The NetworkLayer mostly knows internally when a close is not "regular".
> * The client statistics are missing (the SecureChannel and Session statistics are not very meaningful here).
> 
> Up for discussion.

Sounds good to me. Do you finish this PR now?

PR #3402, opened 2020-01-28
[REVIEW] Some fixes and improvements of ws layer

andranik-s:
1. Pad send buffer with LWS_PRE on allocation
2. Minor memory issue fixed
3. Fixed high cpu usage which was due to timeout being ignored
4. Added some notes

### DIFF
--- a/arch/network_ws.c
+++ b/arch/network_ws.c
@@ -53,11 +53,19 @@ connection_getsendbuffer(UA_Connection *connection, size_t length, UA_ByteString
     UA_SecureChannel *channel = connection->channel;
     if(channel && channel->config.sendBufferSize < length)
         return UA_STATUSCODE_BADCOMMUNICATIONERROR;
-    return UA_ByteString_allocBuffer(buf, length);
+    UA_StatusCode retval = UA_ByteString_allocBuffer(buf, LWS_PRE + length);
+    if(retval != UA_STATUSCODE_GOOD)
+        return retval;
+    buf->data += LWS_PRE;
+    buf->length -= LWS_PRE;
+
+    return UA_STATUSCODE_GOOD;
 }
 
 static void
 connection_releasesendbuffer(UA_Connection *connection, UA_ByteString *buf) {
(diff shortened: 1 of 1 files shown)


andranik-s:
Well, those fails are quite weird. As far as I understand UA_ENABLE_WEBSOCKET_SERVER is not enabled during tests, so this PR shouldn't have caused CI to fail. Am I wrong?

andranik-s:
Just force pushed the same thing and CI is happy now somehow

arch/network_ws.c
```diff
@@ -242,7 +242,10 @@ ServerNetworkLayerWS_start(UA_ServerNetworkLayer *nl, const UA_String *customHos
                          "Could not get the hostname");
         }
     }
-    UA_String_copy(&du, &nl->discoveryUrl);
+    // we need discoveryUrl.data as a null-terminated string for vhost_name
+    nl->discoveryUrl.data = (UA_Byte *)UA_malloc(du.length+1);
+    strcpy((char *)nl->discoveryUrl.data, discoveryUrlBuffer);
```
jpfr: du.length is known.
We should use strncpy here.

andranik-s: Codacy doesn't like strncpy:

> Easily used incorrectly; doesn't always \0-terminate or check for invalid pointers [MS-banned] (CWE-120).

andranik-s: @jpfr So what should I do? Leave it or revert it?

And what is going on with CI? Again I believe that this PR could not have affected any tests since UA_ENABLE_WEBSOCKET_SERVER is not on when testing.

jpfr:
Rebased and squashed the last fixup commit.

andranik-s:
@jpfr can you at least cherry pick https://github.com/open62541/open62541/pull/3402/commits/11eb9b71355d1a78cb542a31212184dd5ff3cfe8, because without it the ws layer burns my cpu :smile:

jpfr:
Merged. Thanks for sending in the fix!

PR #3405, opened 2020-01-29
Example client_encryption ignored invalid certificates

kjeldflarup:
Example client_encryption: If one of the specified certificate files is in wrong format, the example ignored this and continued with policy #None

Unfortunately this does only write a short error message.  
The real error comes from `UA_CertificateVerification_Trustlist`, but even with full trace, it does not reveal which certifiate fails.  
Neither does it say anything about other parts of `UA_ClientConfig_setDefaultEncryption` failing which may lead to the same situation.

### DIFF
--- a/examples/encryption/client_encryption.c
+++ b/examples/encryption/client_encryption.c
@@ -44,10 +44,16 @@ int main(int argc, char* argv[]) {
     UA_Client *client = UA_Client_new();
     UA_ClientConfig *cc = UA_Client_getConfig(client);
     cc->securityMode = UA_MESSAGESECURITYMODE_SIGNANDENCRYPT;
-    UA_ClientConfig_setDefaultEncryption(cc, certificate, privateKey,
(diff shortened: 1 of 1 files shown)


kjeldflarup:
Unfortunately I cannot see what fails in the dockercloud. I can't rule out that the test ignored some error.

jpfr:
Looks good to me.
We don't need the newline character in the error message.

When you remove that, let's see if docker has healed.
Otherwise I can force-merge. This PR is clean just by looking at the changes.

waltronix:
@kjeldflarup but besides that, the example client worked fine for you?

kjeldflarup:
@waltronix Its been a while, and I know that I only used the example to verify this exaxt behaviour. So I'm not sure that I tested all aspects of the example.

jpfr (approved):

jpfr:
Nice!

Issue #3351, opened 2020-01-10
Memory Leak ? : UA_Server_addNode_begin, UA_Server_addNode_finish


MartinaFelberFEM:
Hello!

I think there are memory leaks at following functions: 
UA_Server_addNode_begin
UA_Server_addNode_finish

I am using your current master branch (Commit SHA: d02fc4f5a4b60bb30a3c9394643abdd93b1fd59c) on Debian Jessie with following build options:

```bash
cmake .. -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_PUBSUB=YES -DUA_ENABLE_PUBSUB_ETH_UADP=YES -DUA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING=YES -DUA_ENABLE_PUBSUB_INFORMATIONMODEL=YES -DUA_ENABLE_PUBSUB_INFORMATIONMODEL_METHODS=YES
```

I wrote a PubSub application which can be configured, started and stopped multiple times at runtime. 
The server is always running. 

I always get 2 memory leaks when I add a node. 
I wasn't sure if it's my programming mistake, so I enabled the build option **UA_ENABLE_PUBSUB_INFORMATIONMODEL** which adds nodes when a new PubSub connection is setup and removes them when the connection gets removed.

It all works well and I can see that the configured PubSub connection nodes are visible under the Server object on load and get removed after an unload. The address space representation looks fine.
But there are leaks which all occur at methods **add**XXX**Representation** which are adding nodes to address space.

E.g.: https://github.com/open62541/open62541/blob/d02fc4f5a4b60bb30a3c9394643abdd93b1fd59c/src/pubsub/ua_pubsub_ns0.c#L227

addPubSubConnectionRepresentation
addPublishedDataItemsRepresentation
addDataSetWriterRepresentation
...

The leaks accumulate every time I load a new configuration.

```
aef59407   /home/admin/open62541_master/src/server/ua_nodes.c:540 - addReferenceKind
aef595c4   /home/admin/open62541_master/src/server/ua_nodes.c:581 - UA_Node_addReference
aeb5ae8a   /home/admin/open62541_master/src/server/ua_services_nodemanagement.c:1692 - addOneWayReference
aeb3adda   /home/admin/open62541_master/src/server/ua_server_utils.c:241 - UA_Server_editNode
aeb5b0e0   /home/admin/open62541_master/src/server/ua_services_nodemanagement.c:1743 - Operation_addReference
[...]

Issue #3387, opened 2020-01-20
Server, memory leak after deleting parent with lots of children


juangburgos:
## Description & Reproduction Steps

I don't know if I am missing something here, maybe I am using the API wrong. Or maybe the behaviour is as intended.

I have simplified the issue to a small example where I create a parent node and then add a lot of children to it. Then delete the parent node.

```c
#include <open62541.h>
#include <signal.h>

static UA_NodeId createParent(UA_Server *server) 
{
    UA_NodeId parentId; 
    UA_ObjectAttributes oAttr = UA_ObjectAttributes_default;
    oAttr.displayName = UA_LOCALIZEDTEXT("en-US", "Parent");
    UA_Server_addObjectNode(server, UA_NODEID_NULL,
                            UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER),
                            UA_NODEID_NUMERIC(0, UA_NS0ID_ORGANIZES),
                            UA_QUALIFIEDNAME(1, "Parent"), UA_NODEID_NUMERIC(0, UA_NS0ID_BASEOBJECTTYPE),
                            oAttr, NULL, &parentId);
    return parentId;
}

static void addManyChildrenToParent(UA_Server* server, const UA_NodeId parentId, const int numChildren)
{
    UA_VariableAttributes attr = UA_VariableAttributes_default;
    attr.displayName = UA_LOCALIZEDTEXT("en-US", "Child");
    UA_Int32 myInteger = 1;
    UA_Variant_setScalar(&attr.value, &myInteger, &UA_TYPES[UA_TYPES_INT32]);
    for (int i = 0; i < numChildren; i++)
	{
		UA_Server_addVariableNode(server, UA_NODEID_NULL, parentId,
			UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT),
			UA_QUALIFIEDNAME(1, "Child"),
			UA_NODEID_NUMERIC(0, UA_NS0ID_BASEDATAVARIABLETYPE), attr, NULL, NULL);

    }
}

static void deleteParent(UA_Server* server, const UA_NodeId parentId)
{
    UA_Server_deleteNode(server, parentId, true);
}

UA_Boolean running = true;
static void stopHandler(int sign) {
    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "received ctrl-c");
    running = false;
}

int main(void) {
    signal(SIGINT, stopHandler);
[...]

PR #3407, opened 2020-01-30
Shrink down memory buffer after removing target reference from UA_Node

damijans:
Two independent issues were recently submitted complaining about memory leak after certain number of ua_nodes were removed from the server.
While analyzing this issue I have found out the problem is in removing target reference from node. Each node keeps list of references in two level list structure, where first list keeps track of different reference kind (for example references of parent node to its child node) while second list keeps track of target references of the same reference kind.
When new target reference is inserted into this structure the second list is always extended (resized) while when existing reference is removed only number if active reference is decremented, while the memory is not shrunk down.
This schema works sufficiently good if references are equally inserted/removed or if the last operation is insertion (since at that time memory is reallocated), but it does not work good if first a lot of references are inserted and afterwards all these references are removed.

In this patch I have introduced memory buffer reallocation also when target reference is removed. 

fix: #3351 and #3387


jpfr:
Very cool! You really got to the bottom of this.

damijans:
Thanks for review & accepting the patch.

juangburgos:
Thank you for the fix! Will this find its way to branch 1.0 or only master?

Issue #3398, opened 2020-01-27
MQTT Publish with Json


waheedejaz:
## Minor Fix in tutorial_pubsub_mqtt_publish.c 
UA PubSub with MQTT example gives the option of JSON encoding but doesn't send JSON encoded. It always send UADP endcoded data.

## Background Information / Reproduction Steps
Build the solution with examples with PUBSUB and MQTT options. 
Run the example: `tutorial_pubsub_mqtt_publish --json`
Open another terminal and run: `mosquitto_sub -t customTopic`
Mosquitto clients needs to installed on the system. `sudo apt-get install mosquitto-clients` for ubuntu.

The mosquitto sub receives data but these are weird shapes, not json data.

`�e�M��Le�M�� �������
�^�M����^�M�� �������
�竵M����竵M�� �������
6;��M���:��M�� �������
6�D�M����D�M�� �������
2dM���	��M�� �������
`

OS: Ubunut 18
open62541 Release: latest on 27-Jan-2019 or the commit: ba060096bcec7335b812820c3e2a31a9033097bf

## Culprit Code

>Starting from line#130 

```
/* decide whether to use JSON or UADP encoding*/
#ifdef UA_ENABLE_JSON_ENCODING
    if(useJson)
        writerGroupConfig.encodingMimeType = UA_PUBSUB_ENCODING_JSON;
    else
#endif
        writerGroupConfig.encodingMimeType = UA_PUBSUB_ENCODING_UADP;
    writerGroupConfig.messageSettings.encoding             = UA_EXTENSIONOBJECT_DECODED;
    writerGroupConfig.messageSettings.content.decoded.type = &UA_TYPES[UA_TYPES_UADPWRITERGROUPMESSAGEDATATYPE];
    /* The configuration flags for the messages are encapsulated inside the
     * message- and transport settings extension objects. These extension
     * objects are defined by the standard. e.g.
     * UadpWriterGroupMessageDataType */
    UA_UadpWriterGroupMessageDataType *writerGroupMessage  = UA_UadpWriterGroupMessageDataType_new();
    /* Change message settings of writerGroup to send PublisherId,
     * WriterGroupId in GroupHeader and DataSetWriterId in PayloadHeader
     * of NetworkMessage */
    writerGroupMessage->networkMessageContentMask =
        (UA_UadpNetworkMessageContentMask)(UA_UADPNETWORKMESSAGECONTENTMASK_PUBLISHERID |
[...]

PR #3410, opened 2020-01-30
[Review] fix(ex): use correct fields for json encoding

waheedejaz:


waheedejaz:
Issue created here:
https://github.com/open62541/open62541/issues/3398

jpfr:
Very nice. Thanks!

PR #3411, opened 2020-01-31
[TASK] Binary Nodestore implementation

opcua-tsn-team-kalycito:
- Encode nodes present in information model
 - Dump the encoded nodes and lookup table into a file
 - Decode nodes using the lookup table after the server is in listening state
 - Added 2 CMAKE flags
      - UA_ENABLE_ENCODE_AND_DUMP - encoding and dumping of nodes
      - UA_ENABLE_USE_ENCODED_NODES - use to load the information model
 - Update encoded content during run-time
 - Added support to update string (value attribute) of nodes present in encoded binary
 - Added unit tests for encode and decode APIs


include/open62541/plugin/nodestore.h
```diff
@@ -20,6 +21,10 @@
 #include <open62541/server.h>
 #include "ziptree.h"
 
+#if defined UA_ENABLE_ENCODE_AND_DUMP || defined UA_ENABLE_USE_ENCODED_NODES
```
jpfr: The nodestore is a plugin. The server only uses the plugin API.
You can implement the binary nodestore in a plugin without changing this API.

Please, not a single line should change in nodestore.h.

Encapsulation/hiding of internals is important to keep the architecture clean.

include/open62541/server_config.h
```diff
@@ -166,6 +167,13 @@ struct UA_ServerConfig {
     /* Nodestore */
     UA_Nodestore nodestore;
 
+#ifdef UA_ENABLE_USE_ENCODED_NODES
```
jpfr: Move this inside the binary nodestore plugin.
Do not make changes in this file.

plugins/include/open62541/plugin/nodestore_default.h
```diff
@@ -2,6 +2,7 @@
  * See http://creativecommons.org/publicdomain/zero/1.0/ for more information.
  *
  *    Copyright 2019 (c) Julius Pfrommer, Fraunhofer IOSB
+ *    Copyright 2020 (c) Kalycito Infotech Pvt Ltd
```
jpfr: Remove, as no real changes were made in this file.

plugins/ua_nodestore_ziptree.c
```diff
@@ -4,9 +4,11 @@
  *    Copyright 2014-2018 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
  *    Copyright 2017 (c) Julian Grothoff
  *    Copyright 2017 (c) Stefan Profanter, fortiss GmbH
+ *    Copyright 2020 (c) Kalycito Infotech Pvt Ltd
  */
 
 #include <open62541/plugin/nodestore_default.h>
+#include <open62541/plugin/log_stdout.h>
```
jpfr: Please make your own plugin instead of piling on top of the ZipTree Nodestore.
You can start by copying ZipTree and then making changes.

I prefer some code duplication instead of an ifdef-forest.

src/server/ua_services_attribute.c
```diff
@@ -118,7 +119,16 @@ readValueAttributeFromNode(UA_Server *server, UA_Session *session,
                                        session->sessionHandle, &vn->nodeId,
                                        vn->context, rangeptr, &vn->value.data.value);
         UA_LOCK(server->serviceMutex);
+#ifdef UA_ENABLE_USE_ENCODED_NODES
+        if(server->isServerListening == UA_TRUE) {
```
jpfr: No.

This is not the place to decide whether a node is in the binary format.
This has to be done **inside the nodestore plugin**. It should be totally opaque to the server.

tools/nodeset_compiler/backend_open62541_typedefinitions.py
```diff
@@ -414,15 +414,19 @@ def print_encoding(self):
  * on host ''' + platform.uname()[1] + ''' by user ''' + getpass.getuser() + ''' at ''' + time.strftime(
             "%Y-%m-%d %I:%M:%S") + ''' */
 
+#ifndef ''' + self.parser.outname.upper() + '''_GENERATED_ENCODING_BINARY_H_
```
jpfr: Put changes to the nodeset compiler into a separate PR.

plugins/include/open62541/plugin/nodestore_default.h
opcua-tsn-team-kalycito: @jpfr Thanks for the comments.
Can we move the lookupTable structure, encode and decode function declarations to this (nodestore_default.h) file?

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+     UA_NS0ID_SERVER_SERVERSTATUS_BUILDINFO_PRODUCTURI, UA_NS0ID_SERVER_SERVERSTATUS_BUILDINFO_MANUFACTURERNAME,
+     UA_NS0ID_SERVER_SERVERSTATUS_BUILDINFO_PRODUCTNAME, UA_NS0ID_SERVER_SERVERSTATUS_BUILDINFO_SOFTWAREVERSION,
+     UA_NS0ID_SERVER_SERVERSTATUS_BUILDINFO_BUILDNUMBER, UA_NS0ID_SERVER_SERVERSTATUS_BUILDINFO_BUILDDATE
+    };
+
+UA_Boolean isMinimalNodesAdded = UA_FALSE;
+#define OFFSET_MEMORYSIZE    2024  // Offset memory size for the edited encode nodes
+#define MAX_ROW_LENGTH         30  // Maximum length of a row in lookup table
+
+lookUpTable *ltRead;
```
jpfr: Global variables should be avoided in the library.
A user might instantiate more than one server in the same process.

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+    retval |= UA_Boolean_decodeBinary(src, offset, &viewNode->containsNoLoops);
+    return retval;
+}
+
+
+
+static void
+readRowLookuptable(char ltRow [], int length, lookUpTable *lt, int row) {
+    /**
+     * There are four contents per row
```
jpfr: There is a standard text format for NodeIds.
Like `i=13`, `ns=10;s=Hello:World`, `g=09087e75-8e5e-499b-954f-f2a9603db28a`, `ns=1;b=M/RbKBsRVkePCePcx24oRA==`, etc.
See Part 6, Section 5.3.1.10. Let's try to write and parse that.

So we are known to be complete.

Base64 encoding/decoding is included in the dependencies.
And already used for the JSON parser.

Here's a snippet to parse guids of the form `00000000-0000-0000-0000-000000000000` from a char-array `arg`:
```c
    UA_Guid *guid = UA_Guid_new();
    guid->data1 = strtoull(arg, NULL, 16);
    guid->data2 = strtoull(&arg[9], NULL, 16);
    guid->data3 = strtoull(&arg[14], NULL, 16);
    UA_Int16 data4_1 = strtoull(&arg[19], NULL, 16);
    guid->data4[0] = data4_1 >> 8;
    guid->data4[1] = data4_1;
    UA_Int64 data4_2 = strtoull(&arg[24], NULL, 16);
    guid->data4[2] = data4_2 >> 40;
    guid->data4[3] = data4_2 >> 32;
    guid->data4[4] = data4_2 >> 24;
    guid->data4[5] = data4_2 >> 16;
    guid->data4[6] = data4_2 >> 8;
    guid->data4[7] = data4_2;
```

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "The opening of file lookupTable.bin failed");
+    }
+    while((ch = fgetc(fpLookuptable)) != EOF) {
+        if(ch == '\n') {
+            nodeCount++;
+        }
+    }
+    fclose(fpLookuptable);
+    *ltSize = nodeCount;
+    lookUpTable *lt = (lookUpTable *)UA_calloc(*ltSize, sizeof(lookUpTable));
```
jpfr: Check for NULL every time memory is allocated.
And handle the error of course. We are in memory constrained environments.

This method might be simpler if you just pass in the pointer to `lookupTable`.

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+    }
+
+    if(retval != UA_STATUSCODE_GOOD) {
+           UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "The decodeNode failed with error : %s",
+                                          UA_StatusCode_name(retval));
+    }
+
+    NodeEntry *entry = container_of(node, NodeEntry, nodeId);
+    ++entry->refCount;
+    entry->deleted = true; // remove the decoded node from the memory when release is called!
```
jpfr: You could just remove the `deleted` member of the entries.
I think you always want to delete when the refocunt is zero.
So you don't need that at all.

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+    UA_LocalizedText displayName;
+    retval |= UA_LocalizedText_decodeBinary(&encodedBin, &offset, &displayName);
+    UA_LocalizedText description;
+    retval |= UA_LocalizedText_decodeBinary(&encodedBin, &offset, &description);
+    UA_UInt32 writeMask;
+    retval |= UA_UInt32_decodeBinary(&encodedBin, &offset, &writeMask);
+    size_t referencesSize;
+    retval |= UA_UInt64_decodeBinary(&encodedBin, &offset, (UA_UInt64 *)&referencesSize);
+
+    UA_Node* node = zipNsNewNode(ctx, nodeClass);
```
jpfr: You need to handle the case where malloc in zipNsNewNode fails.

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+            encodeEditedNode(node, &encodeBin, ltRead, ltSizeRead);
+        }
+    }
+    cleanupEntry(entry);
+}
+
+UA_Node*
+decodeNode(void *ctx, UA_ByteString encodedBin, size_t offset) {
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    UA_NodeId nodeID;
```
jpfr: If you malloc the new node first, then you can directly decode "into" the node.
Instead of allocating on the stack and then doing a memcpy.

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+
+    return node;
+}
+
+static const UA_Node *
+zipNsGetNode(void *nsCtx, const UA_NodeId *nodeId) {
+    ZipContext *ns = (ZipContext*)nsCtx;
+    NodeEntry dummy;
+    if(isMinimalNodesAdded) {
+        if(nodeId->identifierType == UA_NODEIDTYPE_NUMERIC) {
```
jpfr: You can just use `UA_NodeId_order` to test for equality.
Then you don't have to make a difference between string and numeric here.

Also, if you order the Nodes in the lookuptable (with UA_NodeId_order) after parsing from the file, then you can get O(log n) lookup speed by doing binary search.
Instead of the linear lookup speed you have here.

plugins/ua_nodestore_binary_encoded.c
```diff
@@ -0,0 +1,1163 @@
...
+    default:
+        break;
+    }
+
+    return retval;
+}
+#endif
+
+#ifdef UA_ENABLE_USE_ENCODED_NODES
+#define MINIMALNODECOUNT        45 // Total number of nodes in MINIMAL configuration
```
jpfr: I'm not sure why these 45 nodes get special treatment.
Can you add a bit of documentation?

I'd prefer (by far!) if we could handle all nodes the same for the binary nodestore.
Maybe we can just rip out the ns0 initialization in ua_server_ns0 if the binary nodestore is provided.
So we can assume they are "pre-initialized" in the binary nodestore.

jpfr: By the way, this contradicts my previous position that no changes in /src are permitted. 😄 
But changing one line (init_ns0) is less of a problem than tight integration.

plugins/ua_nodestore_binary_encoded.c
jpfr: Here's an extended snippet for NodeId parsing I still had from an older project.
You can quote form that.

```c
/* Parse NodeId from String */
static UA_StatusCode
parseNodeId(UA_String str, UA_NodeId *id) {
                /* zero-terminated string */
    UA_STACKARRAY(char, s, str.length + 1);
                memcpy(s, str.data, str.length);
                s[str.length] = 0;

                UA_NodeId_init(id);

                /* Int identifier */
                if(sscanf(s, "ns=%hu;i=%u", &id->namespaceIndex, &id->identifier.numeric) == 2 ||
       sscanf(s, "ns=%hu,i=%u", &id->namespaceIndex, &id->identifier.numeric) == 2 ||
       sscanf(s, "i=%u", &id->identifier.numeric) == 1) {
                               id->identifierType = UA_NODEIDTYPE_NUMERIC;
                               return UA_STATUSCODE_GOOD;
                }

    /* String identifier */
    UA_STACKARRAY(char, sid, str.length);
                if(sscanf(s, "ns=%hu;s=%[^\n]", &id->namespaceIndex, sid) == 2 ||
       sscanf(s, "ns=%hu,s=%[^\n]", &id->namespaceIndex, sid) == 2 ||
       sscanf(s, "s=%s", sid) == 1) {
                               id->identifierType = UA_NODEIDTYPE_STRING;
                               id->identifier.string = UA_String_fromChars(sid);
                               return UA_STATUSCODE_GOOD;
    }

    /* ByteString idenifier */
                if(sscanf(s, "ns=%hu;b=%[^\n]", &id->namespaceIndex, sid) == 2 ||
       sscanf(s, "ns=%hu,b=%[^\n]", &id->namespaceIndex, sid) == 2 ||
       sscanf(s, "b=%s", sid) == 1) {
                               id->identifierType = UA_NODEIDTYPE_BYTESTRING;
                               id->identifier.string = UA_String_fromChars(sid);
                               return UA_STATUSCODE_GOOD;
    }

    /* Guid idenifier */
                if(sscanf(s, "ns=%hu;g=%s", &id->namespaceIndex, sid) == 2 ||
       sscanf(s, "ns=%hu,g=%s", &id->namespaceIndex, sid) == 2 ||
       sscanf(s, "g=%s", sid) == 1) {
        UA_StatusCode retval = parseGuid(sid, &id->identifier.guid);
            return retval;
                               id->identifierType = UA_NODEIDTYPE_GUID;
                               return UA_STATUSCODE_GOOD;
    }

}
```

Stasik0:
bump

jonasgreen88:
Hi @opcua-tsn-team-kalycito, Looking forward to this PR. What is the schedule for this PR? What is left to complete it?

jpfr:
A few people are interested in this.
But this won’t be on master until ca. autumn 2023.

Issue #3328, opened 2019-12-19
Alarm & Conditions: Messages are changed when EnableState is set


ccvca:
## Description

When changing the EnableState Member (by code), the Message of the Condition is replaced. It was quite unexpected for me that my previous message was not available and I could not find anything about this in the specification.

Looking at the code, this is also done when adding a comment, which looks weird as client will never be able to correlate the original message witht the comment if it has not been saved before (e.g. connected after comment has been added).

## Background Information / Reproduction Steps

https://github.com/open62541/open62541/blob/ff87c2876d8e53b7f09472a90c731daa789e9a28/src/server/ua_subscription_alarms_conditions.c#L77

https://github.com/open62541/open62541/blob/ff87c2876d8e53b7f09472a90c731daa789e9a28/src/server/ua_subscription_alarms_conditions.c#L760-L764

Used CMake options:

BUILD_SHARED_LIBS:BOOL=OFF
CLANG_FORMAT_EXE:FILEPATH=CLANG_FORMAT_EXE-NOTFOUND
CMAKE_BUILD_TYPE:STRING=Debug
CMAKE_GNUtoMS:BOOL=OFF
CMAKE_INSTALL_PREFIX:PATH=D:/Programmieren/Cpp/Open62541_Tests/install-debug
UA_ARCHITECTURE:STRING=win32
UA_BUILD_EXAMPLES:BOOL=OFF
UA_BUILD_TOOLS:BOOL=OFF
UA_BUILD_UNIT_TESTS:BOOL=OFF
UA_ENABLE_AMALGAMATION:BOOL=OFF
UA_ENABLE_DA:BOOL=ON
UA_ENABLE_DISCOVERY:BOOL=ON
UA_ENABLE_DISCOVERY_MULTICAST:BOOL=OFF
UA_ENABLE_ENCRYPTION:BOOL=OFF
UA_ENABLE_HISTORIZING:BOOL=OFF
UA_ENABLE_METHODCALLS:BOOL=ON
UA_ENABLE_MICRO_EMB_DEV_PROFILE:BOOL=OFF
UA_ENABLE_NODEMANAGEMENT:BOOL=ON
UA_ENABLE_SUBSCRIPTIONS:BOOL=ON
UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS:BOOL=ON
UA_ENABLE_SUBSCRIPTIONS_EVENTS:BOOL=ON
[...]

PR #3415, opened 2020-02-02
Clean up alarms/conditons

jpfr:


jpfr:
Hey @sqadasi ,

there are no unit tests for alarms/conditions at the moment.
The changes in these PR are mostly cosmetic cleanups.
Still, can you verify that everything still works as intended?

sqadasi:
Hello @jpfr ,
I tested your changes with the A&C example using UaExpert. 
As expected behaviour, Condition 2 should be always on the Alarms View (of UaExpert) when monitoring the Server Object. When enabling Condition 1, it should also appear on the Alarms View. However, the actual behaviour is not quite right. Although we get the correct amount of events when calling ConditionRefresh, the Alarms View only shows one Condition (which was last triggered). 
I noticed that all Conditions end up having the same ConditionId for some reason. Also the Refresh Events end up having a ConditionId, although they are normal events.
Unfortunately I can't tell where could be the exact problem without debugging the code...

ccvca:
There are two more points in the current implementation, that make it harder for me to use it, which might be addresses here too:

- #3328 The message is changed when interacting with the event (e.g. a comment is added), This makes it impossible to clients that connectes later on to get the original message (which contains the information in case).
 - There last Argument "const UA_Boolean deleteEventNode" is missing in UA_Server_triggerConditionEvent (compared to UA_Server_triggerEvent). So I didn't find a way to delete expired (no longer used) conditions from the memory. (I do not use conditions on specific node, instead for communication error messages, that are active till they are resolved/ gone)

I can also write a pull request for the first one (changed messages) quite quickly, but I'm unsure about the intention of that behaviour, so I started an issue first.

As I do not have an overview about the memory management in the current implementation, the second point would be a quite some work for me to do a pull request.

sqadasi:
@ccvca 
I tried to answer your first point in a comment inside the issue #3328.
Regarding the second point, conditions are not transient like events (see Part 9 5.5.1). They describe a state of a system, so when they are no longer needed for whatever reason (or better to say the system works normaly) , they should go inactive to represent a normal state, set Retain field false to indicade that the events shouldn't be handled as alarms or they could be completely disabled, if the client don't want to get events from the conditions at all.
So I am not sure whether deleting a condition makes a valid use case...

ccvca:
In my use case, alarms are used for the displayed error-messages on a HMI-Control panel. When there messages are gone, there is absolutely no need to keep the alarm in memory, as it will not be needed anymore. Instead a new alarm will be created, when the next error-message appears on the HMI. As the Alarm might have different types and multiple alarms can occure at the same time, from my point of view it's just an overhead to keep a pool of former alarms to potential reuse them in the future.

jpfr:
Hi @sqadasi, hi @ccvca,

thanks to looking into this.

The aim of this PR is to make the code a bit easier to approach for new devs.
There is just one commit marked "fix" in this PR.
All the others are refactorings. They should not change the code logic.

Are any of the points you uncovered new problems compared to master?
If yes, let's fix the regressions before merging this PR.
If no, let's open dedicated issues and PRs for preexisting bugs.

Good to see we have two alarms/conditions experts. 😄 
Let's have *some* unit tests to make working on this code less scary for the non-expert devs.

ccvca:
I've checked this version with my small example Server and everything has the same behaviour as before.
Sorry for starting the discussion here, I was just thinking, that these two points migtht be added easily.

The discussion is extracted into a different issue: #3419

jpfr:
Perfect @ccvca .
Merging then.

sqadasi:
Hello @jpfr ,
I appologise for the late answer. In my first comment https://github.com/open62541/open62541/pull/3415#issuecomment-581410547 I tried to explain the problem which I found. I guess it should've been better to support my explaining with images.
I used the A&C Example in master to reproduce the error. As you see in the image, RefreshStartEvent has a ConditionId field, which is not correct. RefreshEvents should not have ConditionId fields at all. Actually, all triggered events have the same ConditionId value, thats why only one condition is displayed on the Alarm View in UaExpert.

To reproduce the problem, start the A&C example server, enable condition1, monitore server object in Events View, UaExpert will call ConditionRefresh method. In EventView you can check the Events.

1) RefreshEvent
grafik

2) Condition2
grafik

3) Condition1
grafik

PR #3418, opened 2020-02-05
fix(ex): use correct data type for writer group json message settings

waheedejaz:
datatypes_pubsub.txt: add JsonWriterGroupMessageDataType
tutorial_pubsub_mqtt_publish.c: use correct datatype for WriterGroup json message settings
ua_pubsub_writer.c: UA_Server_addWriterGroup(): validate if correct type for json message settings is passed.

https://github.com/open62541/open62541/issues/3398


jpfr:
Nice! Well-received and merged.

PR #3420, opened 2020-02-05
fix(client): Correctly set endpointUrl in HEL Message

Pro:
There is currently an inconsistency. The client->endpointUrl variable is only used for the async connect. This is just a quick-fix.

Fixup for #3390 See:
https://github.com/open62541/open62541/commit/e93f8ebb698a1dd957952e68b15e09f0e4303ebe#diff-096258576a8c3ac7adc8daaf14b817e2L83


jpfr:
Good find!

PR #3423, opened 2020-02-05
Time triggered multithreaded PubSub application with ETF

opcua-tsn-team-kalycito:
Time triggered PubSub application using ETF at a publishing interval of 250 microseconds with qbv offset of 25us
Integrate time triggered functionality and publish OPCUA PubSub packets at calculated transmission time over ethernet
Application design change introduced:
- To run the application in all processors, a macro NANO_SECONDS_SLEEP_PUB has been introduced, which allots 40% of the cycletime for packet preparation in the publisher
- For example, running the application in 250us, 100us time is allocated for packet preparation and additional 25us for qbv offset
- Running in high end processors like i5 and i7, one can decrease the packet preparation time to 20% by changing NANO_SECONDS_SLEEP_PUB to (CYCLE_TIME * 0.8)
- Publisher and Subscriber threads run in same core(core 2)
- User application thread runs in core 3

Follow README_ETF.txt file available in examples/pubsub_realtime to work on this commit


jpfr:
Looks good. But we have to dissect into smaller PR a bit.

First of all, the ua_pubsub_custom_manager.c  should not be below /src.
Can pull the manager out into a separate PR and move it to /plugin or /example?

More comments to follow.

opcua-tsn-team-kalycito:
> Looks good. But we have to dissect into smaller PR a bit.
> First of all, the ua_pubsub_custom_manager.c  should not be below /src.
> Can pull the manager out into a separate PR and move it to /plugin or /example?
> More comments to follow.

@jpfr 
If we enable UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING macro, example binaries are not generated and results in linking error. Inorder to build examples with the UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING macro enabled, we introduced ua_pubsub_custom_manager.c file which does not have any change compared to the previous implementation (ua_pubsub_manager.c). 
The file ua_pubsub_custom_manager.c can be moved to /plugins, but we think it cannot be raised as a separate PR. Does it sound good to you?
Also, can you please provide all the other comments so that we can address them all at once? Thanks.

jpfr:
This is intentional.
The custom manager makes the manager a user-supplied functionality instead of library-supplied.

So all the examples without a user-supplied manager cannot work when we demand one.

Please fix this first.

Why are you changing this? I don’t see another custom manager implementation that is specific to ETF.

jpfr:
Maybe don’t touch the pubsub manager at all.
If you don’t change the logic.

plugins/include/open62541/plugin/pubsub_realtime_etf.h
```diff
@@ -0,0 +1,74 @@
...
+#include "ua_pubsub.h"
+#include "time.h"
+#include <open62541/plugin/log_stdout.h>
+#include <linux/types.h>
+#ifdef UA_ENABLE_PUBSUB_ETH_UADP
+#include <open62541/plugin/pubsub_ethernet.h>
+#endif
+
+#define   CLOCKIDENTITY                                   CLOCK_TAI
+#define   MULTICAST_ADDRESS                               "224.0.0.32"
```
jpfr: The multicast address is part of the example, not the general functionality.
It should notbe in /plugins/include. Move it to /examples/pubsub_realtime.

Do not include <linux/types.h> and <open62541/plugin/log_stdout.h> here.
This should be part of the implementation. But not part of the interface.

jpfr:
Did you see my last comments?
There is still a lot of Linux-specific stuff in `plugins/include/open62541/plugin/pubsub_ethernet.h`.
But the plugin API doesn't use it.
Please move the Linux definitions into the plugin and out of the interface.

Please address my last comment and this one.
Then we are okay to merge.

opcua-tsn-team-kalycito:
@jpfr 
We have removed the linux-specific headers from plugins/include/open62541/plugin/pubsub_realtime_etf.h, but with respect to /plugins/include/open62541/plugin/pubsub_ethernet.h, we require the linux header if_ether.h to define the ETH_ALEN parameter of the 
UA_PubSubChannelDataEthernet structure. This structure is used in ua_pubsub_realtime_etf.c and in ua_pubsub_ethernet.c plugin files, hence we have moved UA_PubSubChannelDataEthernet structure from ua_pubsub_ethernet.c to pubsub_ethernet.h. Is this right or is there
any other way to approach the same?

Also, PUBSUB_IP_ADDRESS macro definition in pubsub_realtime_etf.h is used in ua_pubsub_udp.c for specifying the interface to send UDP packets with ETF. Hence we have not removed PUBSUB_IP_ADDRESS macro definition from pubsub_realtime_etf.h file

plugins/include/open62541/plugin/pubsub_realtime_etf.h
```diff
@@ -0,0 +1,59 @@
...
+
+#ifdef UA_ENABLE_PUBSUB_REALTIME_PUBLISH_ETF
+
+#include "ua_pubsub.h"
+#include <open62541/plugin/pubsub_udp.h>
+#ifdef UA_ENABLE_PUBSUB_ETH_UADP
+#include <open62541/plugin/pubsub_ethernet.h>
+#endif
+
+#define   PUBSUB_IP_ADDRESS                               "192.168.8.105"
```
jpfr: No. You put application specific IP addresses and cycle time in the plugin header.
As said before, this a no-go.

Please make the plugin API in a way where the configuration is provided by the application and not part of header definitions.

plugins/ua_pubsub_realtime_etf.c
```diff
@@ -0,0 +1,259 @@
...
+#define   MICRO_SECONDS                                   1000
+#define   MILLI_SECONDS                                   1000 * 1000
+#define   SECONDS                                         1000 * 1000 * 1000
+#define   CYCLE_TIME                                      250 * 1000
+#define   QBV_OFFSET                                      25
+#define   SECONDS_INCREMENT                               1
+#define   FAILURE_EXIT                                    -1
+#define   SHIFT_32BITS                                    32
+
+struct timespec        nextCycleStartTime;
```
jpfr: Global variables in a plugin are a no-go.
You (Kalycito) have requests coming in for multiple publishers with different cycle times in one application.
This cannot be done with the cycle time in a global variable.

plugins/ua_pubsub_realtime_etf.c
```diff
@@ -0,0 +1,259 @@
...
+#include <linux/if_packet.h>
+#endif
+#include "poll.h"
+#include "linux/errqueue.h"
+#include "time.h"
+
+#define   MICRO_SECONDS                                   1000
+#define   MILLI_SECONDS                                   1000 * 1000
+#define   SECONDS                                         1000 * 1000 * 1000
+#define   CYCLE_TIME                                      250 * 1000
```
jpfr: The cycle time ought to be configurable.

plugins/ua_pubsub_realtime_etf.c
```diff
@@ -0,0 +1,259 @@
...
+        /* Move to next second and remove it from ns field */
+        timeSpecValue->tv_sec  += SECONDS_INCREMENT;
+        timeSpecValue->tv_nsec -= SECONDS;
+    }
+
+}
+
+/* txtime calculation for UDP */
+UA_StatusCode
+txtimecalc_udp(UA_PubSubChannel *channel, UA_ExtensionObject *transportSettings, const UA_ByteString *buf) {
```
jpfr: You have this nice argument with the transportSettings coming for `txtimecalc_udp` and `txtimecalc_ethernet`.
Put the transport settings there. And not into global variables.

PR #3428, opened 2020-02-07
docs: use correct package name in Ubuntu PPAs

mikini:
*Add major version to package name as used in PPAs
*Elaborate a bit on the Ubuntu/Debian/PPA/APT/deb complexities
*Refer to the Debian PPA instructions


mikini:
The wrong package name seems to be fixed in commit 8b058c2435198a6e10a6a77d997227b89d6ea129.

Did you find any problems in the elaboration about Debian/Ubuntu differences since this was not included/discussed?

PR #3445, opened 2020-02-18
Array ranges in the form x:x

giusguerrini:
Hi,
In some of my applications it's helpful to have array ranges "min:max" where min==max. I don't know if this format is legal, but it simplified somewhat my applilcation.


giusguerrini:
Sorry, used the wrong merge base :-(

PR #3446, opened 2020-02-18
Accept array ranges "min:max" where min==max.

giusguerrini:
Hi all,
in one of my client application it is helpful to describe array ranges in the general form "min:max", but the server doesn't accept ranges when min and max are the same. I slightly relaxed the test so I can use the same range format in all cases. I don't know if the standard explicitly forbids the case of "min==max", but it would look weird to me. Anyway, this small change has simplified my particular problem; I hope it helps other people.
Regards


jpfr:
Yes. We should allow this.
Can you add a unit test to avoid future regressions?

giusguerrini:
File **tests/check_types_range.c** looks like a good place. Can I add a new testcase there?

jpfr:
Yes

giusguerrini:
Done (I hope I did it the right way). Oddly, there is already a test case named "parseRangeMinEqualMax", but it doesn't seem to do what it promises. I left it unchanged and added mine.

giusguerrini:
Uhm...  two more tests failed, but I can't figure out why. Tested on my machine (ubuntu 14 x64), at the moment I can't try vs2017.

PR #3449, opened 2020-02-18
Kludge to hook calls to default logger (UA_Logger_Stdout_). Users can…

giusguerrini:
Hi,
it seems that log redirection is still incomplete. E.g. in my environment (linux) I can successfully intercept most server messages, but some are still generated internally by using the default logger directly (or, worst, by calling **UA_Log_Stdout_log** function directly). This is bad for my application, since I need to collect all messages consistently. So I modified the way the library initializes client and server configurations in such a way that I can redirect the logger by assigning address of my custom **UA_Logger** structure to global pointer **UA_Logger_Stdout**. I am aware that this is a temporary solution (e.g. in general it isn't clean for a library to expose global variables, a registration functon would be better), but solved my problem.
I hope it could help.
Regards

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -43,7 +43,7 @@ UA_Server_new() {
     UA_ServerConfig config;
     memset(&config, 0, sizeof(UA_ServerConfig));
     /* Set a default logger and NodeStore for the initialization */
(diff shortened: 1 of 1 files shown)


andreasebner:
Thank you for the contribution. The logging has been rewritten, so I’m closing this PR.

PR #3455, opened 2020-02-21
Extended the access control plugin with check for browsing nodes

jonasgreen88:
The PermissionType (Part 3) defines 17 permissions. Currently the
access control plugin can control all but 2 of them. Those are Browse
and ReceiveEvents. This change adds check for browse permission of
nodes.

### DIFF
--- a/include/open62541/plugin/accesscontrol.h
+++ b/include/open62541/plugin/accesscontrol.h
@@ -88,6 +88,11 @@ struct UA_AccessControl {
     UA_Boolean (*allowDeleteReference)(UA_Server *server, UA_AccessControl *ac,
                                        const UA_NodeId *sessionId, void *sessionContext,
                                        const UA_DeleteReferencesItem *item);
+
+    /* Allow browsing a node */
+    UA_Boolean (*allowBrowseNode)(UA_Server *server, UA_AccessControl *ac,
+                                  const UA_NodeId *sessionId, void *sessionContext,
+                                  const UA_NodeId *nodeId, void *nodeContext);
 #ifdef UA_ENABLE_HISTORIZING
     /* Allow insert,replace,update of historical data */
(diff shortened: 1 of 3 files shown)


jpfr:
Looks very good!

We have segfaults in the CI.
I think you have to add the method pointer also in the default config.

Can be merged when green.

jonasgreen88:
> We have segfaults in the CI.
> I think you have to add the method pointer also in the default config.

Sorry but I can't find any other default config than ua_accesscontrol_default.c. In there the new function pointer is assigned.

mlang-de:
Nice PR!
Could it be meaningful to return an UA_Statuscode instead of a boolean?
So it could be possible to dynamically "hide and unhide" nodes without delete them.
For instance a Port which contains a device only in the address space if the device is connected.
(compare IO-Link Companion spec. figure 6)
Return code shall be BadNodeIdUnknown for IO-Link. The companion spec defines "Device not  instantiated --> Entry/ Delete eventually available device instance"

jpfr:
@mlang-de Good idea! This would apply to all the access control methods.
If we hide for browsing, we should also be able to hide for reading attributes, etc.
There is also some interaction with *Views* once we get around implementing them.

Let me merge this and we can have this discussion separate.

jonasgreen88:
> Nice PR!
> Could it be meaningful to return an UA_Statuscode instead of a boolean?
> So it could be possible to dynamically "hide and unhide" nodes without delete them.
> For instance a Port which contains a device only in the address space if the device is connected.
> (compare IO-Link Companion spec. figure 6)
> Return code shall be BadNodeIdUnknown for IO-Link. The companion spec defines "Device not instantiated --> Entry/ Delete eventually available device instance"

I like the idea. But isn't this already possible to achieve in the nodestore plugin? If the node should be hidden return a null pointer and the stack will return BadNodeIdUnknown.

mlang-de:
@jonasgreen88 You are right. For the moment it fits! As jpfr already mentoined is will be required in future in combination with Views.

PR #3456, opened 2020-02-21
[REVIEW]: add DataType node in example

matkonnerth:
Currently the datatype node for the 3D Type is not added to the address space, therefore a client has no information how to decode the extension object.

This PR adds the datatype node, UAExpert is able to decode the value.

### DIFF
--- a/examples/custom_datatype/server_types_custom.c
+++ b/examples/custom_datatype/server_types_custom.c
@@ -11,14 +11,26 @@
 #include "custom_datatype.h"
 
 UA_Boolean running = true;
+const UA_NodeId variableTypeId = {
+    1, UA_NODEIDTYPE_NUMERIC, {4243}};
 
 static void stopHandler(int sig) {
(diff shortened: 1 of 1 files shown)


jpfr:
> This PR adds the datatype node, UAExpert is able to decode the value.

I have no idea how this can help UAExpert.
Because the DataType node does not contain any structure information.
Are you sure just adding the "empty" node does help?

But yes, it is a good idea to add to the information model.

matkonnerth:
take a look on it with UAExpert with the datatype node and without it. With the datatype node, a client can read the DataTypeDefinition attribute of the DataTypeNode!

matkonnerth:
DataTypeDefinition is provided via https://github.com/open62541/open62541/pull/2971

jpfr:
What about adding all custom data types during startup?
Then we don't have to do it manually for every example!
However, in that case, the custom datatypes need more metadata.

matkonnerth:
This would be a good idea, however the information about the hierachical reference of a custom struct is missing.

Think it's okay to do it manually in an example. If a nodeset is provided, nodes for the datatypes are generated automatically, think that's the right way.

PR #3457, opened 2020-02-21
PubSub: Avoid invalid free of transport layer

MartinaFelberFEM:
If the build option UA_ENABLE_PUBSUB is enabled but PubSub is not used
(e.g. there are no configured transport layers) we have to avoid a
invalid free.


jpfr:
Thanks!

Issue #2509, opened 2019-02-27
Network Layer start-up result ignored (errors get unnoticed...)


pnikiel:
## Description
Result of Network Layer start is ignored (for example because of TCP/IP port being already in use).
This gives false impression that an OPC-UA server started.

## Background Information / Reproduction Steps
Just try with a simplest possible server:
1. run the first instance (should start just fine),
2. in another terminal, run the second instance
you will see that UA_Server_run or UA_Server_run_startup (depending which one you used) absolutely ignored that TCP/IP port is already taken.

I inspected the source code and the problem comes from function:
void addServerSocket(ServerNetworkLayerTCP *layer, struct addrinfo *ai)
in arch/ua_network_tcp.c
which should return integer instead of void.
This function is called by ServerNetworkLayerTCP_start which is already made to return proper error code but it of course ignores the result of invocation to aforementioned addServerSocket.

By looking at the code I expect the problem to be both in current master as well as in 0.3.0; in my development env where I run things I clearly see this in 0.3.0.

 If there's interest from open62541 management, I could improve it.

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): 0.3.0  and current master
 - [x] Other OPC UA SDKs used (client or server):  none
 - [x] Operating system:  Linux
 - [x] Logs (with `UA_LOGLEVEL` set as low as necessary) attached  below
 - [ ] Wireshark network dump attached   Not necessary
 - [x] Self-contained code example attached   Just use the minimal server example.
 - [ ] Critical issue

Logs:
[2019-02-27 22:06:58.723 (UTC+0100)] warn/network	Error binding a server socket: Address already in use
[2019-02-27 22:06:58.723 (UTC+0100)] warn/network	Error binding a server socket: Address already in use
[2019-02-27 22:06:58.723 (UTC+0100)] info/network	TCP network layer listening on opc.tcp://pcatldcsp11.cern.ch:4841/

I.e. the error was noticed, the server shouldn't have started but it did.

Pro:
Thanks for reporting this!

Can you propose a PR with your changes?

pnikiel:
I will. Let's talk soon. Best Regards!

Infinity95:
This should already be handled differently in #2408.

pnikiel:
Hi, 
could either of you advise me if I should wait for #2408 and retest/propose improvement then, or I go ahead and propose a PR now?

Pro:
[...]

PR #3460, opened 2020-02-24
Fix for #2509: Network Layer start-up result ignored (errors get unnoticed...)

pnikiel:
Hi,
I'm proposing a fix for:
https://github.com/open62541/open62541/issues/2509

I tested this fix trying to start two open62541-based OPC-UA servers on the same port and now the error:
[2020-02-24 11:01:04.254 (UTC+0100)] warn/network	Error binding a server socket: Address already in use
correctly propagates into the return code.

(Previously any error like that would be ignored by the stack).

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -277,7 +277,7 @@ ServerNetworkLayerTCP_add(UA_ServerNetworkLayer *nl, ServerNetworkLayerTCP *laye
     return UA_STATUSCODE_GOOD;
 }
 
-static void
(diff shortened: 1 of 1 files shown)


jpfr:
Well received.
Thanks!

anzeskerjanc:
Hi,

this solution has problem on device with SocketCAN. I get error on starting server. Before was just warning "Error opening the server socket".
Can function returns good status in case of invalid socket (for now this is also my quick fix)?

image

PR #3465, opened 2020-02-26
[WIP] Async operations overhaul

andranik-s:
Goals:
1. Generalize anync request handling to support different types of requests (remove hardcoded UA_CallRequest)
2. Move low level handling (UA_Server_getAsyncOperationNonBlocking, UA_Server_setAsyncOperationResult, etc) inside the library and provide UA_Server_processAsync that could be called inside a user created thread
3. Add async read/write requests
4. Add async history requests
5. ? Get rid of async request timeouts. Reasons:
    1. not all operation results have status codes (compare UA_CallResponse/UA_CallMethodResult and UA_ReadResponse/UA_DataValue), so it is not always possible to signal that a timeout event has occurred
    2. clients can better handle timeouts (inside their callbacks) since they have more info about what is going on
    3. current implementation is faulty because the working thread continues executing the current request after it has timed out

### DIFF
--- a/examples/tutorial_server_method_async.c
+++ b/examples/tutorial_server_method_async.c
@@ -178,16 +178,20 @@ THREAD_CALLBACK(ThreadWorker) {
     while(running) {
         UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
                     "Try to dequeue an async operation");
-        const UA_AsyncOperationRequest* request = NULL;
+        const UA_CallMethodRequest *request = NULL;
         void *context = NULL;
-        UA_AsyncOperationType type;
-        if(UA_Server_getAsyncOperationNonBlocking(globalServer, &type, &request, &context, NULL) == true) {
-            UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "AsyncMethod_Testing: Got entry: OKAY");
-            UA_CallMethodResult response = UA_Server_call(globalServer, &request->callMethodRequest);
-            UA_Server_setAsyncOperationResult(globalServer, (UA_AsyncOperationResponse*)&response,
-                                              context);
-            UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "AsyncMethod_Testing: Call done: OKAY");
-            UA_CallMethodResult_clear(&response);
+        const UA_DataType *type;
+        if(UA_Server_getAsyncOperationNonBlocking(globalServer, &type, (const void**)&request, &context, NULL) == true) {
+            if (type == &UA_TYPES[UA_TYPES_CALLMETHODREQUEST]) {
+                UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "AsyncMethod_Testing: Got entry: OKAY");
+                UA_CallMethodResult response = UA_Server_call(globalServer, request);
+                UA_Server_setAsyncOperationResult(globalServer, &response,
+                                                  context);
+                UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "AsyncMethod_Testing: Call done: OKAY");
+                UA_CallMethodResult_clear(&response);
+            } else {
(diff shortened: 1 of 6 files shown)


jpfr:
Thanks for putting in the work!
Most aspects of the pull request are really cool!
Some parts need discussion. Let me go throught the points:

**1. Generalize anync request handling to support different types of requests (remove hardcoded UA_CallRequest)**

Allowing all types of services with a minimalistic API is good goal.
But there are also downsides to it.

**a)** Sometimes important information is missing in the individual "operation".
For example:
```c
typedef struct {
    UA_RequestHeader requestHeader;
    UA_Double maxAge;
    UA_TimestampsToReturn timestampsToReturn;
    size_t nodesToReadSize;
    UA_ReadValueId *nodesToRead;
} UA_ReadRequest;
```
The UA_ReadValueId does not contain `maxAge` or `timestampsToReturn`.
Se we have to create a custom structure for the async read request.

**b)** Not all Responses have the
- ResponseHeader
- ResultsArray

pattern that is assumed in the code. So exceptions need to be built in anyway.

For example
```c
typedef struct {
    UA_ResponseHeader responseHeader;
    UA_ByteString serverNonce;
    size_t resultsSize;
    UA_StatusCode *results;
    size_t diagnosticInfosSize;
    UA_DiagnosticInfo *diagnosticInfos;
} UA_ActivateSessionResponse;
```

**c)** We don't give hints to users what we expect as a result datatype.
This is easy to look up in the service definition. But this takes time and there is more room for error here.

**2. Move low level handling (UA_Server_getAsyncOperationNonBlocking, UA_Server_setAsyncOperationResult, etc) inside the library and provide UA_Server_processAsync that could be called inside a user created thread**

Has this been implemented?
Maybe I missed that in the diff.

**3. Get rid of async request timeouts**

Don't agree.
CallMethodResult has a StatusCode field.
DataValue also has a StatusCode field.

Yes, clients can handle timeouts internally.
But we want timeouts for individual operations. If some have returned results and we are only waiting for one late straggler. Losing all results in that case is bad.

I see a lot of good approaches in the PR.
Maybe we can separate concerns a little bit and make several smaller PR in lieu of one big chunk of changes.

Looking forward to hearing your opinion on things!

andranik-s:
Thanks for the feedback!

> Allowing all types of services with a minimalistic API

This was the initial goal but for a limited number of request types.

> Sometimes important information is missing in the individual "operation"

This whole idea of separating requests into individual operations is causing a lot of problems and is really hard to generelize. There is a simpler approach. For example, if at least one of UA_CallMethodRequest in UA_CallRequest is async then the whole UA_CallRequest is processed asynchronously. That would require more synchronization in clients code but is so much easier to implement without any special cases. Maybe I should do that as a separate feature instead of trying to adapt the current implementation?

> With the change we don't give type hints to users for what we expect as a result datatype

This is not a problem if all that low level handling is hidden from the clients. They should not manully dequeue operations, call appropriate handlers, and set results. The library can do it for the clients. That is the idea behind UA_Server_processAsync.

> Has this been implemented?

No, it's a WIP. Sorry I did not mention it explicitly. I created this PR just to get some feedback.

> If some have returned results and we are only waiting for one late straggler.

The way I see it those stragglers are users' problem. They have spawned them and so they should deal with them. But I do not insist I just wanted to simplify things a little bit.

pnikiel:
Hi,
hope I'm not bothering too much in your discussion...
I wanted to say that it's great to see this work being done. I'm the primary author of the open62541-compat library (https://github.com/quasar-team/open62541-compat) which adapts open62541 to the API of Unified Automation SDK (UA-SDK), as well as quasar framework where both open62541 and UA-SDK are used to create OPC-UA software for LHC Particle Accelerator complex at CERN.

Having said this, we have great use for this feature so once it settles I'd implement it in open62541-compat.

I'm ready, willing and able to test everything that relates to asynchronous server operations in open62541. If apart from testing there are other things I could contribute in this domain let me know. We really need those asynchronous things. So let me know if I can be of use.

andranik-s:
@pnikiel nice to see some interest for this feature. I will close close this PR and create a new one with a new implementation. Any feedback and help is greatly appreciated.

PR #3467, opened 2020-02-27
fix order of files to amalgamate, before amalgamation would conflict …

pnikiel:
…with multithreading now it works

Hi,
I'm proposing this tiny fix.
It solves the problem that when you do (in the base branch):

cmake -DUA_ENABLE_AMALGAMATION=ON  -DUA_MULTITHREADING=101 ../

it fails to build because amalgamation picks wrong order of headers.

With this simple fix both unamalgamated and amalgamated version build fine with multithreading enabled.


jpfr:
Thanks!

PR #3469, opened 2020-02-27
[REVIEW]add security policy Basic256Sha256 support with OpenSSL provider

wenshan1:
1. I introduce two new macros: UA_ENABLE_ENCRYPTION_MBEDTLS UA_ENABLE_ENCRYPTION_OPENSSL.
The encryption code is divided into two parts: the encryption common code and the concrete encryption algorithm implementation code.
The encryption common code includes:
            src/ua_securechannel.c:415:#ifdef UA_ENABLE_ENCRYPTION
            src/client/ua_client_connect.c:304:#ifdef UA_ENABLE_ENCRYPTION
			plugins/include/open62541/plugin/pki_default.h:19:#ifdef UA_ENABLE_ENCRYPTION
			...
The concrete encryption algorithm implementation code includes the source code written by mbedtls encryption provider or the source code written by opessl encryption provider.
The user of open6254 (such as: the examples) will not care who is the encryption provider, just enable the macro UA_ENABLE_ENCRYPTION to get the encryption function. 
For the previous compatibility and do not modify the test code and examples code, UA_ENABLE_ENCRYPTION is enabled then the macro UA_ENABLE_ENCRYPTION_MBEDTLS is enabled by default. 
2. The user can use the OpenSSL as the security provider like this: cmake ..  -DUA_BUILD_EXAMPLES=ON -DUA_ENABLE_ENCRYPTION_OPENSSL=ON
3. The PKI and the other security policy with OpenSSL will be implemented later.
4. Test: 
    a. as the server, the UA encryption client with MbedTLs can open the secure channel and active the session.
   b. as the server, the opcfoundation CTT tool as the client can open the secure channel, active the session and browse the server namespace.


jpfr:
Loks great. Perfect contribution!

There are only minor issues in the Travis CI.
For example this one:

```
====== CPPCHECK Static Analysis Errors ======

plugins/securityPolicies/openssl/securitypolicy_openssl_common.c(547): performance (redundantAssignment): Variable 'ret' is reassigned a value before the old one has been used.

plugins/securityPolicies/openssl/securitypolicy_openssl_common.c(615): performance (redundantAssignment): Variable 'ret' is reassigned a value before the old one has been used.
```

wenshan1:
Thanks for your review, fix it and reopen the pull request.

ccvca:
There is already an findOpenSSL distributed with cmake:
"cmake-3.11\share\cmake-3.11\Modules\FindOpenSSL.cmake"

https://cmake.org/cmake/help/v3.6/module/FindOpenSSL.html

jpfr:
Hey @wenshan1 ,

I just added PR #3475 to complement your changes.
Please rebase on top of that when #3475 has landed.

With #3475, you can return UA_BadNotSupported for the initialization of unsupported SecurityPolicies. This will be logged, but the configuration does not abort.
No endpoints will be created for these SecurityPolicies.

Furthermore, your "PKI" plugin accepts all certificates.
This could give a false sense of security to users who did not read the code in detail.
Please add this functionality also for OpenSSL.

If possible, we can break out "common functionality" for mbedTLS and OpenSSL from the current implementation. And maybe simplify the code a bit...

Sorry for adding the PKI stuff to your scope so late.
But this is actually critical for users who rely on security.

wenshan1:
@jpfr I'll add the PKI support with openssl. I find the function UA_CertificateVerification_CertFolders() is only added for Linux OS. Is it mandatory to add the function UA_CertificateVerification_CertFolders() support in this pull request? 
I hope my pull request can be merged into the latest release.

PR #3470, opened 2020-02-27
[REVIEW] Attach session to channel when activating if no channel exists

jonasgreen88:
While testing access rights we got UaExpert to perform a sequence
that we belive the open62541 server did not handle correctly:
1. OpenSecureChannel
2. CreateSession
3. ActivateSession
4. Some initial browsing/reading that eventually failed due to
   limited access rights
5. CloseSecureCannel (session is now detached from the channel)
6. OpenSecureChannel
7. ActivateSession (same session id as activated in no. 3)

During the second ActivateSession the server failed the request
in UA_Session_generateNonce() as there was no channel attached to the
session. The session was detached from the secure channel when
UaExpert closed the channel in no. 5. This PR makes the session
attached to the new channel before generating the nonce.

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -10,7 +10,7 @@
  *    Copyright 2017 (c) Stefan Profanter, fortiss GmbH
  *    Copyright 2017-2018 (c) Mark Giraud, Fraunhofer IOSB
  *    Copyright 2019 (c) Kalycito Infotech Private Limited
- *    Copyright 2018-2019 (c) HMS Industrial Networks AB (Author: Jonas Green)
+ *    Copyright 2018-2020 (c) HMS Industrial Networks AB (Author: Jonas Green)
  */
 
 #include "ua_services.h"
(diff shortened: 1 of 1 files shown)


jpfr:
Good find.

This problem is important to prevent in the future.
Can you add a small unit test?

Can be merged then.

jonasgreen88:
I was looking into if it was possible to make a unittest using the client as is but found out that the client currently doesn't support to reactivate a session. Maybe we should implement that in a another PR before adding a unittest.

jpfr:
This PR removes code for the following scenario.

The SecureChannel A opens a Session S.
A new SecureChannel B is opened.
B calls ActivateSession on S while B is still open.
In that case, B is to be detached and closed.

Please change the PR to cover your scenario **and** this one.

jpfr:
This needs a unit test and backport to v1.0

jonasgreen88:
> This PR removes code for the following scenario.
> 
> The SecureChannel A opens a Session S.
> A new SecureChannel B is opened.
> B calls ActivateSession on S while B is still open.
> In that case, B is to be detached and closed.
> 
> Please change the PR to cover your scenario **and** this one.

I'm not sure what you mean, might there be an error in the example:
The SecureChannel A opens a Session S.
A new SecureChannel B is opened.
B calls ActivateSession on S while **A** is still open.
In that case, **A** is to be detached and closed.

I think this case is covered as `session->header.channel != channel` will evaluate true and A will be detached while B will be attached to the session. A will eventually timeout and be closed if there is no activity on that channel.

jpfr:
Merged. Thanks!

PR #3471, opened 2020-02-28
test(openbsd): compile options

bluhm:
On OpenBSD there is not subunit package available.  Do not add it
to the library list to successfully build the tests.

The check package is installed in /usr/local/ so the include files
and libraries should be searched there.  Move this directory to the
end of the list so that other software lying there, like an old
version of open62531, is less likely to interfere with testing.


jpfr:
👍

PR #3473, opened 2020-02-28
doc(openbsd): binary package available

bluhm:
An official OpenBSD port of open62541 exists.  The binary package
is available as snapshot now and will be included in the upcoming
OpenBSD 6.7 release.  Put install instructions into the documention.


jpfr:
👍

PR #3478, opened 2020-02-29
add security policy Basic256Sha256 support and PKI with OpenSSL provider 

wenshan1:
1. I introduce two new macros: UA_ENABLE_ENCRYPTION_MBEDTLS UA_ENABLE_ENCRYPTION_OPENSSL.
The encryption code is divided into two parts: the encryption common code and the concrete encryption algorithm implementation code.
The encryption common code includes:
            src/ua_securechannel.c:415:#ifdef UA_ENABLE_ENCRYPTION
            src/client/ua_client_connect.c:304:#ifdef UA_ENABLE_ENCRYPTION
			plugins/include/open62541/plugin/pki_default.h:19:#ifdef UA_ENABLE_ENCRYPTION
			...
The concrete encryption algorithm implementation code includes the source code written by mbedtls encryption provider or the source code written by opessl encryption provider.
The user of open6254 (such as: the examples) will not care who is the encryption provider, just enable the macro UA_ENABLE_ENCRYPTION to get the encryption function. 
For the previous compatibility and do not modify the test code and examples code, UA_ENABLE_ENCRYPTION is enabled then the macro UA_ENABLE_ENCRYPTION_MBEDTLS is enabled by default. 
2. The user can use the OpenSSL as the security provider like this: cmake ..  -DUA_BUILD_EXAMPLES=ON -DUA_ENABLE_ENCRYPTION_OPENSSL=ON
3.  The other security policy with OpenSSL will be implemented later.
4. The test code will be added later.
5. Test:
    OS environment:  Ubuntu18.04 Linux 
    The PKI is enabled.
    a. as the server with OpenSSL, the UA encryption client with MbedTLs can open the secure channel and active the session.
   b. as the server with OpenSSL, the opcfoundation CTT tool v1.03 as the client can open the secure channel, active the session and browse the server namespace.


jpfr:
The PR looks good. Was it closed intentionally?

wenshan1:
@jpfr After the source code is reviewed in my company, I will re-open the pull request.

PR #3483, opened 2020-03-03
Levelled log

kjeldflarup:
Add UA_Log_Stdout_ErrorLevel which only prints Error level and above regardless of UA_LOGLEVEL being set lower.  

The logger can be enabled like this:
```
UA_ServerConfig *open62541_server_config = UA_Server_getConfig(open62541_server);
UA_ServerConfig_setDefault(open62541_server_config);
open62541_server_config->logger = UA_Log_Stdout_ErrorLevel_; 
```

This also allows log level being a command line parameter.
```
UA_Logger mylogger = {UA_Log_Stdout_loglevelcheck, (void*)mylevel, UA_Log_Stdout_clear};
```

The PR also makes it easier to clone UA_Log_Stdout, as some definitions is moved to `open62541/plugin/log_stdout.h`

The PR is based on 1.0.1, but should also be possible to cherry pick to master.

### DIFF
--- a/plugins/include/open62541/plugin/log_stdout.h
+++ b/plugins/include/open62541/plugin/log_stdout.h
@@ -22,6 +22,11 @@ UA_Log_Stdout_log(void *_, UA_LogLevel level, UA_LogCategory category,
 UA_EXPORT void
 UA_Log_Stdout_clear(void *logContext);
 
+/* By default the client and server is configured with UA_Log_Stdout
+   This constructs a logger with a configurable max log level */
+
+UA_EXPORT UA_Logger UA_Log_Stdout_withLevel(UA_LogLevel minlevel);
+
 _UA_END_DECLS
 
 #endif /* UA_LOG_STDOUT_H_ */
--- a/plugins/ua_log_stdout.c
+++ b/plugins/ua_log_stdout.c
@@ -48,8 +48,14 @@ const char *logCategoryNames[7] = {"network", "channel", "session", "server",
 __attribute__((__format__(__printf__, 4 , 0)))
 #endif
 void
-UA_Log_Stdout_log(void *_, UA_LogLevel level, UA_LogCategory category,
+UA_Log_Stdout_log(void *context, UA_LogLevel level, UA_LogCategory category,
                   const char *msg, va_list args) {
+
+    /* Assume that context is casted to UA_LogLevel */
+    /* TODO we may later change this to a struct with bitfields to filter on category */
+    if ( context != NULL && (UA_LogLevel)(uintptr_t)context > level )
+        return;
+
     UA_Int64 tOffset = UA_DateTime_localTimeUtcOffset();
     UA_DateTimeStruct dts = UA_DateTime_toStruct(UA_DateTime_now() + tOffset);
 
@@ -76,3 +82,12 @@ UA_Log_Stdout_clear(void *logContext) {
 
 const UA_Logger UA_Log_Stdout_ = {UA_Log_Stdout_log, NULL, UA_Log_Stdout_clear};
 const UA_Logger *UA_Log_Stdout = &UA_Log_Stdout_;
+
+/* By default the client and server is configured with UA_Log_Stdout
+   This constructs a logger with a configurable max log level */
+
+UA_Logger UA_Log_Stdout_withLevel(UA_LogLevel minlevel)
+{
+    UA_Logger logger = {UA_Log_Stdout_log, (void*)minlevel, UA_Log_Stdout_clear};
+    return logger;
+}
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -157,6 +157,8 @@ add_example(server_repeated_job server_repeated_job.c)
 
 add_example(server_inheritance server_inheritance.c)
 
+add_example(server_loglevel server_loglevel.c)
+
 if(UA_ENABLE_HISTORIZING)
     add_example(tutorial_server_historicaldata tutorial_server_historicaldata.c)
 endif()
--- a/examples/server_loglevel.c
+++ b/examples/server_loglevel.c
@@ -1,0 +1,77 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+#include <open62541/plugin/log_stdout.h>
+#include <open62541/server.h>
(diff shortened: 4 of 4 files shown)


jpfr:
Good idea.
This could be generalized even further.

The logger has a void context pointer.
We could add the actual max log level to the context. That is evaluated only if a context has been set.

The existing logger can be extended that way with an alternative initialization function that takes the actual log level as an additional argument.

kjeldflarup:
@jpfr 
I did not change `UA_Log_Stdout_log`, because I did not want to break backwards compability, I would hate to break others peoples code.

Actually I did use context, in a very primitive way. To avoid messing with allocating variables, I typecasted it into `UA_LogLevel`.  
Now the meaning of context is defined by the `log` and clear `fields` in `UA_Logger`, thus we can continue with this typecast for now, and later change it to an allocated struct.

The idea of an initialization function is clearly better than exporting `UA_Log_Stdout_ErrorLevel`, and would support encapsulating the usage of context

kjeldflarup:
@jpfr
How much does size and speed matter? I assume this is embedded after all.  

I noticed that we also have `UA_LogCategory`, I would change that to a bitmask. This would allow to select categories like this  `UA_LOGCATEGORY_SERVER | UA_LOGCATEGORY_SECURITYPOLICY`

I could do the same with `UA_LogLevel`, but that does not make much sense.

Even a 16 bit void pointer for `context` could hold that
```
struct {
  level: 4;
  categorymask: 12;
};
```

kjeldflarup:
@jpfr I think the PR is ready for review now

plugins/include/open62541/plugin/log_stdout.h
```diff
@@ -8,9 +8,35 @@
...
 #include <open62541/plugin/log.h>
+#include <open62541/server_config.h>
+#include <open62541/client_config.h>
 
 _UA_BEGIN_DECLS
 
+/* ANSI escape sequences for color output taken from here:
+ * https://stackoverflow.com/questions/3219393/stdlib-and-colored-output-in-c*/
+
+#ifdef UA_ENABLE_LOG_COLORS
```
jpfr, line 19: Keep the defines in the .c file.
They are internal and not part of the API.

plugins/include/open62541/plugin/log_stdout.h
```diff
@@ -8,9 +8,35 @@
...
+# define ANSI_COLOR_RED     ""
+# define ANSI_COLOR_GREEN   ""
+# define ANSI_COLOR_YELLOW  ""
+# define ANSI_COLOR_BLUE    ""
+# define ANSI_COLOR_MAGENTA ""
+# define ANSI_COLOR_CYAN    ""
+# define ANSI_COLOR_RESET   ""
+#endif
+
+extern UA_EXPORT const char *logLevelNames[6];
```
jpfr, line 37: Keep these internal as well.

plugins/include/open62541/plugin/log_stdout.h
```diff
@@ -22,6 +48,15 @@ UA_Log_Stdout_log(void *_, UA_LogLevel level, UA_LogCategory category,
 UA_EXPORT void
 UA_Log_Stdout_clear(void *logContext);
 
+/* By default the client and server is configured with UA_Log_Stdout_log
+   These functions sets the context with a max log level */
+UA_EXPORT void
+UA_ClientConfig_setLog_Stdout_withLevel( UA_ClientConfig *config,
```
jpfr, line 54: I'd rather see a "constructor function" that returns the configured UA_Logger structure.
This can be used for the client and server config setup.

```c
UA_Logger UA_Log_Stdout_withLevel(UA_LogLevel minlevel);
```

Remove the _setLog methods.

plugins/ua_log_stdout.c
```diff
@@ -48,8 +27,14 @@ const char *logCategoryNames[7] = {"network", "channel", "session", "server",
 __attribute__((__format__(__printf__, 4 , 0)))
 #endif
 void
-UA_Log_Stdout_log(void *_, UA_LogLevel level, UA_LogCategory category,
+UA_Log_Stdout_log(void *context, UA_LogLevel level, UA_LogCategory category,
                   const char *msg, va_list args) {
+
+    /* Assume that context is casted to UA_LogLevel */
+    /* TODO we may later change this to a struct with bitfields to filter on category */
+    if ( context != NULL && (UA_LogLevel)context > level )
```
jpfr, line 35: You might have to cast to (uintptr_t) first.
Some compilers warn for a direct cast to enums.

plugins/include/open62541/plugin/log_stdout.h
kjeldflarup: IMHO the `_setLog` is more consistent with how other parts of the client/server is configured. 

You want the users to configure it like this then?
```
conf->logger = UA_Log_Stdout_withLevel(UA_LogLevel minlevel);
```

kjeldflarup:
In many plugins I see stuff like this
```UA_LOG_INFO(UA_Log_Stdout
```

Am I right that these plugins will never respect this log level?

plugins/include/open62541/plugin/log_stdout.h
kjeldflarup: I have added the `UA_Log_Stdout_withLevel` constructor.  
But so far I have kept `UA_ClientConfig_setLog_Stdout_withLevel`, and just changed it to use `UA_Log_Stdout_withLevel`

plugins/include/open62541/plugin/log_stdout.h
kjeldflarup: Moved it latest commit

plugins/include/open62541/plugin/log_stdout.h
kjeldflarup: Moved it latest commit

kjeldflarup:
@jpfr I have updated the PR

kjeldflarup:
@jpfr I have added an example

jpfr:
I don’t like the specialization for the client and server config. The config is public API. We can have users directly set the logger and avoid too much shallow wrapper API.

Why extern the logging category names?

Please rebase on 1.1 and see that the CI is green. Can be merged when the above is done.

Good job!

kjeldflarup:
I have rebased and removed some unused changed.  
I have fixed the CI errors which I understood.

But the three ones left I do not understand what the problem is.

jpfr:
Rebased again.

jpfr:

PR #3409, opened 2020-01-30
[WIP] Networking code overhaul

mlgiraud:
Since the master changed quite a bit, i will continue on a new branch, since it is easier adapting to the changes that way.
Old PR: #2408

Current todo:

- [x] Adapt tests
- [ ] Fix pipelines
- [ ] Fix remaining issues

PR #3488, opened 2020-03-04
refactor securechannel: move asym decrypt to channel

mlgiraud:
This moves the decryption of asymmetrically secured messages back into
the securechannel. This way control flow is unified and we dont need to
"reprocess" chunks if the sequence number check fails.
The control flow is now more obvious.
The only "downside" of this approach imho is, that we need to call the
configure callback into the application.
I find the control flow of this easier to follow than some jumping
around depending on failed checks.

This is currently contained in #3409 and i extracted it to reduce the size of the PR.
@jpfr Can you take a look?

### DIFF
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -16,12 +16,14 @@
 #include <open62541/transport_generated_handling.h>
 #include <open62541/types_generated_encoding_binary.h>
 #include <open62541/types_generated_handling.h>
+#include <open62541/plugin/log_stdout.h>
 
 #include "ua_securechannel.h"
 #include "ua_util_internal.h"
 
 #define UA_BITMASK_MESSAGETYPE 0x00ffffffu
 #define UA_BITMASK_CHUNKTYPE 0xff000000u
+#define UA_MESSAGE_HEADER_SIZE 8
 
 const UA_ByteString UA_SECURITY_POLICY_NONE_URI =
     {47, (UA_Byte *)"http://opcfoundation.org/UA/SecurityPolicy#None"};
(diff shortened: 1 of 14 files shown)


jpfr:
Wow.
That was fast. I’m will look at this later today.

src/ua_securechannel.h
```diff
@@ -96,6 +96,9 @@ struct UA_SecureChannel {
     const UA_SecurityPolicy *securityPolicy;
     void *channelContext; /* For interaction with the security policy */
     UA_Connection *connection;
+    UA_CertificateVerification *certificateVerification;
```
jpfr: The asymmetric header is forwarded to the client/server in the new configure callback.
The verification could be done inside the configure callback.
(Simpler structures are better structures...)

src/ua_securechannel.c
```diff
@@ -637,86 +638,148 @@ processSequenceNumberSym(UA_SecureChannel *channel, UA_UInt32 sequenceNumber) {
...
 
-/* The chunk body begins after the SecureConversationMessageHeader */
 static UA_StatusCode
-decryptAddChunk(UA_SecureChannel *channel, UA_MessageType messageType,
-                UA_ChunkType chunkType, UA_ByteString *chunk,
-                UA_Boolean allowPreviousToken) {
-    /* Has the SecureChannel timed out? */
-    if(channel->state == UA_SECURECHANNELSTATE_CLOSED)
-        return UA_STATUSCODE_BADSECURECHANNELCLOSED;
+getSecurityHeaderAlgorithms(void *application,
```
jpfr: This function is too complicated. I wouldn't know how to use it properly.
It looks as if we try to generalize a functionality that doesn't want to be generalized. ;-)

src/ua_securechannel.h
mlgiraud: I'd rather have a pointer to the certVer in the securechannel than duplicate code to verify the certificate in client and server.
The certificate verification step is an essential part of the securechannel imho and should be located there.

Choosing the correct security policy on the other hand is server specific behavior, and located in the configure call.
On the client side we need to do next to nothing.
Currently only the flag that the handshake is complete is set.

src/ua_securechannel.c
mlgiraud: I inlined it. Should be cleaner now.

jpfr:
Merged via  #3623.

PR #3491, opened 2020-03-05
fix(mdns): Replace dots in hostname with minus to ensure correct domain name

Pro:
There was an issue if someone defined his own custom hostname as the IP Adress, which leads to an invalid dns name


src/server/ua_services_discovery_multicast.c
```diff
@@ -310,6 +310,11 @@ createFullServiceDomain(char *outServiceDomain, size_t maxLen,
                     (int) servernameLen, (char *) servername->data,
                     (int) hostnameLen, (char *) hostname->data);
         offset = servernameLen + hostnameLen + 1;
+        //replace all dots with minus. Otherwise mDNS is not valid
```
wenshan1: c++ comments are not recommended by the standard

PR #3497, opened 2020-03-10
Fix crash: Avoid debug trace of nullpointer at Service_ActivateSession

MartinaFelberFEM:
I moved the debug log further down (after the null pointer check). 
And added an additional log, if the session pointer is invalid. 
I am not sure which log messages are needed.
But at least one can see that Service_ActivateSession has been called.

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -495,7 +495,6 @@ void
 Service_ActivateSession(UA_Server *server, UA_SecureChannel *channel,
                         UA_Session *session, const UA_ActivateSessionRequest *request,
                         UA_ActivateSessionResponse *response) {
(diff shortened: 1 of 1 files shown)


MartinaFelberFEM:
CI checks fail because of valgrind download

fatal: repository 'https://sourceware.org/git/valgrind.git/' not found
Error: An exception occurred within a child process:
  DownloadError: Failed to download resource "valgrind"
Failure while executing; `git clone --branch master https://sourceware.org/git/valgrind.git /Users/travis/Library/Caches/Homebrew/valgrind--git` exited with 128. Here's the output:
Cloning into '/Users/travis/Library/Caches/Homebrew/valgrind--git'...
fatal: repository 'https://sourceware.org/git/valgrind.git/' not found

jpfr:
The fix for the MacOS build is attempted here: #3498

PR #3503, opened 2020-03-11
[Review] feat(core) Support for structures with optional fields and unions

andreasebner:
This PR implements the support for structures with optional fields and unions. The custom datatype client and server examples are extended and should be used to getting started with OptStructs and Unions.

As an side effect of this PR, it's now possible to use open62541 together with OptStruct or Union depended companion specifications like  OPC UA for AutoId. In addition, the AutoId information model is now added to the nodeset-compiler tests. The open62541 reduced information model was extended with the Union Datatype to support Unions by the default cake-configuration.

The  internal representation of Unions and OptStructs is following the suggestion for C/C++ of the OPC UA Standard Part 6 5.2.7 (OptStructs) and 5.2.8 (Unions).

**In detail the following changes are introduced:**
→ Extended open62541 datatype handling for the support of optStructs and unions.
→ Support for the data type ‘Bit’ in the context of structures with optional fields. The usage of ‘Bit’ outside an structure with optional fields is still prevented.
→ The nodeset compiler is extended and can now handle structures with optional fields and unions
-> Added Arrays, OptStructs and Unions to the custom datatype example


src/ua_types_encoding_binary.c
```diff
@@ -1718,6 +1914,95 @@ calcSizeBinaryStructure(const void *p, const UA_DataType *type) {
...
+
+        /* Scalar */
+        s += calcSizeBinaryJumpTable[membertype->typeKind]((const void*)ptr, membertype);
+        ptr += membertype->memSize;
+    }
+
+    return s;
+}
+
+static size_t
```
jpfr: This function is defined in multiple places.
It could be made non-static with a single implementation.

src/ua_types.c
```diff
@@ -1058,8 +1077,8 @@ const UA_copySignature copyJumpTable[UA_DATATYPEKINDS] = {
     (UA_copySignature)copyNotImplemented, /* Decimal */
     (UA_copySignature)copy4Byte, /* Enumeration */
     (UA_copySignature)copyStructure,
-    (UA_copySignature)copyNotImplemented, /* Structure with Optional Fields */
-    (UA_copySignature)copyNotImplemented, /* Union */
+    (UA_copySignature)copyStructure, /* Structure with Optional Fields */
+    (UA_copySignature)copyStructure, /* Union */
```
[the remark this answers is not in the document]
GrigoryFridman: copyStructure should work for unions as expected, because the UInt32 switchField at the beginning is listed as a DataTypeMember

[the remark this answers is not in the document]
jpfr: - Do you generate a "real union" or structure + tag?
- It would be good if we didn't need a member in UA_DataType for the switchfield.

src/ua_types.c
```diff
@@ -1010,13 +1011,37 @@ copyStructure(const void *src, void *dst, const UA_DataType *type) {
     for(size_t i = 0; i < type->membersSize; ++i) {
         const UA_DataTypeMember *m= &type->members[i];
         const UA_DataType *mt = &typelists[!m->namespaceZero][m->memberTypeIndex];
-        if(!m->isArray) {
+        if (m->isOptional){
```
jpfr, line 1014: The compiler makes a "fast path" that has minimum jump instructions.
(Exceptions to the rule exist of course.)

A standalone `if` has a jump-instruction.
For an `if-else`, the first if does not have jump instructions.

This is speed-critical code.
Try to make the most common operations go in the first clause of the `if-else`.

If that makes the code ugly, we can add annotation to help the branch prediction.
But we haven't done this in the past.

```c
#define likely(x)       __builtin_expect((x),1)
#define unlikely(x)     __builtin_expect((x),0)
```

src/ua_types.c
```diff
@@ -1010,13 +1011,37 @@ copyStructure(const void *src, void *dst, const UA_DataType *type) {
...
             ptrs += m->padding;
             ptrd += m->padding;
-            retval |= copyJumpTable[mt->typeKind]((const void*)ptrs, (void*)ptrd, mt);
+            if((!m->isArray && (*(void* const*)ptrs != NULL)) ||
```
jpfr, line 1017: We have optional and array and combinations thereof.
So four different cases.

You check for m->isArray in 5 different places.
That makes the code hard to read.
It should be enough to test this just once.

src/ua_types.c
```diff
@@ -1090,20 +1133,57 @@ clearStructure(void *p, const UA_DataType *type) {
     for(size_t i = 0; i < type->membersSize; ++i) {
         const UA_DataTypeMember *m = &type->members[i];
         const UA_DataType *mt = &typelists[!m->namespaceZero][m->memberTypeIndex];
-        if(!m->isArray) {
-            ptr += m->padding;
-            clearJumpTable[mt->typeKind]((void*)ptr, mt);
-            ptr += mt->memSize;
+        if(m->isOptional) {
```
jpfr, line 1136: See the comments for the copy operation

src/ua_types.c
```diff
@@ -1090,20 +1133,57 @@ clearStructure(void *p, const UA_DataType *type) {
...
+                    ptr += sizeof(void*);
+                } else {
+                    ptr += m->padding;
+                    size_t length = *(size_t*)ptr;
+                    ptr += sizeof(size_t);
+                    UA_Array_delete(*(void**)ptr, length, mt);
+                    ptr += sizeof(void*);
+                }
+
+            } else {
```
jpfr, line 1151: This code is so complicated, add some comments to branches which case they handle.

src/ua_types_encoding_binary.c
```diff
@@ -1345,6 +1347,115 @@ encodeBinaryStruct(const void *src, const UA_DataType *type, Ctx *ctx) {
...
+    status ret;
+    UA_UInt32 selection = *(UA_UInt32*) ptr;
+    if(selection == 0){
+        return encodeWithExchangeBuffer((const void*)ptr, &UA_TYPES[UA_TYPES_UINT32], ctx);
+    }
+    const UA_DataType *typelists[2] = { UA_TYPES, &type[-type->typeIndex] };
+    const UA_DataTypeMember *m = &type->members[selection-1];
+    const UA_DataType *mt = &typelists[!m->namespaceZero][m->memberTypeIndex];
+
+    ret = encodeWithExchangeBuffer((const void*)ptr, &UA_TYPES[UA_TYPES_UINT32], ctx);
```
jpfr, line 1447: The code can be simplified so that encodeWithExchangeBuffer for the switchField occurs only once.

src/ua_types_encoding_binary.c
```diff
@@ -1456,6 +1567,98 @@ decodeBinaryStructure(void *dst, const UA_DataType *type, Ctx *ctx) {
...
+    if(ctx->depth > UA_ENCODING_MAX_RECURSION)
+        return UA_STATUSCODE_BADENCODINGERROR;
+    ctx->depth++;
+    uintptr_t ptr = (uintptr_t)dst;
+    status ret;
+    UA_UInt32 selection = *(UA_UInt32*) ctx->pos;
+    if(selection == 0)
+        return decodeBinaryJumpTable[UA_TYPES_UINT32]((void *UA_RESTRICT)ptr, &UA_TYPES[UA_TYPES_UINT32], ctx);
+    u8 membersSize = type->membersSize;
+    if(selection-1 >= membersSize) return UA_STATUSCODE_BADDECODINGERROR;
```
jpfr, line 1643: Line break

jpfr:
Great!

PR #3505, opened 2020-03-12
PubSub UDP plugin: interface selection

MartinaFelberFEM:
Hello!

I have tested several PubSub connection configurations and I think that the interface selection does not work for the Subscriber.

The configuration data type `UA_NetworkAddressUrlDataType` contains the URL and the network interface address. The network interface address is optional.
If it is given, the current UDP plugin implementation configured the socket for sending operations correctly (setsocket - IP_MULTICAST_IF)

But on regist() it joined the multicast group with interface address INADDR_ANY (setsocket - IP_ADD_MEMBERSHIP)
According to the man pages http://man7.org/linux/man-pages/man7/ip.7.html the kernel chooses an appropriate interface if the interface address is equal to INADDR_ANY.

I have a test hardware with two ethernet interfaces - ETH0 and ETH1.
I have configured both interfaces and want to publish and subscribe via interface ETH1.

**Use case 1:** The UA_NetworkAddressUrlDataType only provides a multicast URL but no interface.
    In this case the default ethernet interface ETH0 has been chosen by the kernel and publishing and subscribing works via ETH0.

**Use case 2:** The UA_NetworkAddressUrlDataType provides a multicast URL and additionally an interface - ETH1.
    In this case the publisher sends the packages correctly via ETH1. But the subscriber receives nothing, because the kernel chooses the default ethernet interface ETH0. So the publisher is configured correctly, but the receiving interface for the Subscriber is determined by the kernel and not by the configuration.

That's why I have added the interface configuration (if given) to the setsockopt IP_ADD_MEMBERSHIP call.

Are there any tests with different ethernet configurations? The patch works for me, but I am not sure if it's correct.
I don't want to accidentally break anything.

Many thanks for your help


jpfr:
Nice!

PR #3511, opened 2020-03-14
Simplify chunk assembly

jpfr:


src/ua_securechannel.c
```diff
@@ -690,83 +520,269 @@ decryptAddChunk(UA_SecureChannel *channel, UA_MessageType messageType,
...
+            break;
+        }
+
+        deleteChunks(&doneChunks); 
+        if(res != UA_STATUSCODE_GOOD)
+            break;
+
+        /* Could we process a full message? Done? */
+        UA_Chunk *next = SIMPLEQ_FIRST(&channel->completeChunks);
+        if(chunk == next)
```
mlgiraud: I think i understand what we are doing here, but clang analyzer thinks this condition is always true, which is not the case.
Might indicate a problem that i don't see or too complex code. (And maybe a bug in clang analyzer of course)

src/client/ua_client_connect.c
```diff
@@ -394,27 +394,29 @@ decodeProcessOPNResponseAsync(void *application, UA_SecureChannel *channel,
...
         UA_Client_disconnect(client);
         return;
     }
 
+   /* Decode and verify the sequence header */
+    UA_SequenceHeader sequenceHeader;
+    retval = UA_SequenceHeader_decodeBinary(msg, &offset, &sequenceHeader);
 #ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
-    retval = processSequenceNumberAsym(channel, sequenceNumber);
+    retval |= processSequenceNumberAsym(channel, sequenceHeader.sequenceNumber);
```
mlgiraud: OR'ing the status codes is a bad idea. If two bad status codes are OR'ed it is impossible to reconstruct both error codes in most cases.

src/ua_securechannel.c
```diff
@@ -690,83 +520,269 @@ decryptAddChunk(UA_SecureChannel *channel, UA_MessageType messageType,
...
+                       void *application, UA_ProcessMessageCallback callback) {
+    /* Only MSG and CLO messages at this point */
+    UA_assert(messageType == UA_MESSAGETYPE_MSG ||
+              messageType == UA_MESSAGETYPE_CLO);
+
+    /* Assemble message */
+    UA_Chunk *chunk;
+    UA_ByteString payload;
+    if(chunksSize > 1) {
+        /* Compute the full message length. Omit the MessageHeader and
```
mlgiraud: We can avoid this loop by calculating the size when the chunks are extracted from the packet, or when the chunksSize variable is calculated.

jpfr:
Thanks for taking a look.
Yes, let's bring this up after our next call.

src/ua_securechannel.c
```diff
@@ -690,83 +520,269 @@ decryptAddChunk(UA_SecureChannel *channel, UA_MessageType messageType,
...
+        return UA_STATUSCODE_GOOD;
+
+    /* Compare that the messageType until the final chunk is the same */
+    UA_MessageType messageType = chunk->messageType;
+
+    /* Do we have a complete message? */
+    UA_ChunkType finalType = UA_CHUNKTYPE_INTERMEDIATE;
+    size_t chunksSize = 0;
+    UA_Chunk *finalChunk = NULL;
+    SIMPLEQ_FOREACH(finalChunk, &channel->completeChunks, pointers) {
```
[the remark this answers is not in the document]
jpfr: Integrate the message size count in here.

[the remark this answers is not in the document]
jpfr: But... padding is still integrated here.

PR #3514, opened 2020-03-17
fix(server): Add warning if fuzzing build is enabled

Pro:
If `UA_BUILD_FUZZING` authentication tokens and other variables are overridden internally. This causes subscriptions and other parts of the code to not work as expected.

Since `UA_BUILD_FUZZING` should not be used for debug or release builds in general, show a warning to avoid accidental usage of this build option.


jpfr:
Rebased.

PR #3518, opened 2020-03-19
[WIP] enable build/test for security source code using openssl in windows

wenshan1:


wenshan1:
The job is canceled for the job running on agent Hosted Agent ran longer than the maximum time of 60 minutes.

jpfr:
Maybe we can add OpenSSL-enabled builds as a different "architecture" to Azure.
So that we start separate build jobs for combinations of OS/Compiler/Crypto.

PR #3523, opened 2020-03-26
[WIP] feat(pubsub) Information model RT variables mapping

andreasebner:
This PR extends the PubSub RT functionality. Currently there is no option to use real information model nodes together with the highest RT levels. In fact the source of the published data is not visible in the information model. 

With the introduced changes the user must link the RT DataSetField with an existing node in the RT information model. To use the highest RT level the user has to provide also static memory for the value. This memory must be used by the user and will not be automatically deleted.


ccvca:
Could this also be achived by extending a node?
At the moment it's possible to get a value in two ways:

- UA_Server_writeValue in advance
- onRead-Callback (Which calls UA_Server_writeValue internally)

Wouldn't it be enought, to extended the variable node value union with a third variable?
https://github.com/open62541/open62541/blob/e2cea8bf405a1a63b1d11024f3b95d8d63ecbca9/include/open62541/plugin/nodestore.h#L196-L202
So add another Variable to the union: "UA_DataValue * dataValueInMemory"?

Background: Achieve an easier binding non-OPC UA-Structures, e.g. an Int32 in memory without using onRead for every variable.

Issue #2865, opened 2019-07-02
OPC UA Pub/Sub over MQTT


canadyre:
## Description
Hi all,

I am trying to work with the OPC UA PubSub with MQTT on a couple of Raspberry Pis. I was wondering if there were any examples of this that anyone has come across? The only examples I can find use UDP over UADP. I also read an issue that said that AMQP was not supported even though it mentions in the Documentation so is that the case with MQTT as well? Any help would be greatly appreciated.

Thanks!

## Background Information / Reproduction Steps

Used CMake options:
UA_ENABLE_PUBSUB ON
UA_ENABLE_PUBSUB_ETH_UADP ON
UA_BUILD_EXAMPLES ON
CMAKE_C_COMPILER /usr/bin/arm-linux-gnueabihf-gcc

## Checklist

Please provide the following information:

 - [0.3 ] open62541 Version (release number or git tag):
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ Raspbian] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

theresa95:
As far as I know, the PubSub mechanism is currently only supported by UADP.

sohailmushtaq:
Hi, 
there is a branch which contains OPC UA PubSub MQTT with examples also. Check the branch pubsub_mqtt_clean. My master thesis is about OPC UA PubSub, so I've worked on PubSub with MQTT, in case you have questions just let me know.

Regards

Sohail

canadyre:
Hi,
Thank you for helping me find that @sohailmushtaq . So when running these examples with mqtt, do you still run tutorial_pubsub_publish, tutorial_pubsub_subscribe and then use tutorial_pubsub_mqtt as the broker? 
Thanks!

sohailmushtaq:
Hi,
No, use only tutotial_pubsub_mqtt. First of all install a MQTT broker, such as Mosquitto (it's easy to install and use, I reccomend that). After you the Broker is running, just execute pubsub_mqtt you should see packets sent from Publisher to Broker on Wireshark. If you want to receive as well you have to configure also the Subscriber. Once both configured and connected to Broker, Publisher will send information through Broker to Subscriber.

jpfr:
We have a current PR for pubsub with MQTT in #2796.

rudoq32:
[...]

PR #3537, opened 2020-04-02
[review] feat(pubsub): add support for transportSettings of dataSetReader

waheedejaz:
server_pubsub.h: add extensionObject for transportSettings in DataSetReaderConfig
ua_network_pubsub_mqtt.c: correct datatype checking regist function and QoS mapping
ua_pubsub_reader.c: copy transportSettings when copying dataSetReader_config and some cosmetic  changes

datatypes_pubsub.txt: add BrokerDataSetReaderTransportDataType and JsonDataSetReaderMessageDataType

#2865


src/pubsub/ua_pubsub_reader.c
```diff
@@ -287,7 +285,7 @@ void UA_ReaderGroup_subscribeCallback(UA_Server *server, UA_ReaderGroup *readerG
         UA_LOG_INFO(&server->config.logger, UA_LOGCATEGORY_SERVER, "Message buffer alloc failed!");
         return;
     }
-
+    buffer.length = 0;
```
jpfr, line 288: Not sure about that one.
Isn't the length of the available buffer checked in the `receive` function?

waheedejaz: buffer length is initialized to 0 here.  After the receive() function,  the length of the buffer is checked to determine if a message is received or not. The check can successful due to garbage value even if the message is not received.

opcua-tsn-team-kalycito:
> Looks good to me.
> Please ok this one @andreasebner, @opcua-tsn-team-kalycito .

@jpfr Looks good to us.

src/pubsub/ua_pubsub_reader.c
waheedejaz: Upon closer look, this is not correct. The buffer.length is the total size of the buffer, not the size of the received buffer. It is used in memset(0) in allocate and delete buffer functions. 

But it is also used to check the length of the received message on the next line. In fact, the decision, whether or not a message is received,  is made by checking this length. Total allocated length is 512 bytes, so this check would always be true whether or not the message is received.  Currently, MQTT channel doesn't have receive function, code just explodes at this point. Both UDPMP and ETHERNET receive functions change the buffer length to 0 or to the received message size. This also causes memory leaks. We can maybe change the structure of  UA_ByteString, add one more variable for current size. In any case, the handling of pubsub receive must be change.

For now, I am removing this line.

jpfr: > Both UDPMP and ETHERNET receive functions change the buffer length to 0 or to the received message size. This also causes memory leaks.

That should not happen. The line that sets the length to zero should also free the memory.
Setting to a smaller positive value is okay as `free` does not care in the end.

Can you open separate issues if the above is confirmed?

waheedejaz: It is being freed at the end of subscriberCallBack() in all cases. I was concerned about the memset(0). Didn't realize that type->size and buffer size are different. 
All good!

jpfr:
The PR does two things:

1. Fixes a few issues around data types.
2. Adds the TransportSettings.

The first looks very good.
For the second one, I see the settings copied over. But they are not deleted at the end of the Reader's lifecycle. Further, nobody is using the TransportSettings at the moment. What are they for?

Is it enough to keep the fixes and use the TransportSettings only during the initialization of the Reader?

waheedejaz:
> I see the settings copied over. But they are not deleted at the end of the Reader's lifecycle. 

TransportSettings deletion added. PR updated.

> Further, nobody is using the TransportSettings at the moment. What are they for?

I am working on MQTT_Subscribe example. Didn't want to put up a lot changes for review at once.

jpfr:
Got it. Small PR are good.
Merged.

Issue #3451, opened 2020-02-20
client :  acknowledge or disable an alarm & condition


kikia1b2:
Is there  an example for client to acknowledge or disable an alarm? 
Use function UA_Client_writeEventNotifierAttribute(UA_Client *client, const UA_NodeId nodeId,
                                      const UA_Byte *newEventNotifier)?
 But I have no idea about how to set these paras to acknowledge or disable an alarm?   
Thank you .

kikia1b2:
@ccvca

schroeder-:
Call the "Acknowledge""(NodeId: i=9111) or "Disable"(NodeId: i=9028) method of the Datatype AcknowledgeableConditionType (NodeId: i=2881)  with the EventNotifier of the event.

kikia1b2:
@schroeder- in tutorial_client_event.c,  alarm attributes (like eventId ) can be got in handler_events(), but how to get the alarm's objectId or conditionId which will be used in the following function to acknowledge the alarm??
	UA_StatusCode retval = UA_Client_call(client, objectId, methodId, 2, &input,  &outputSize,  &output);
Thank you so much.

schroeder-:
For the ConditionId add to your filter: 
```
  selectClauses[x].typeDefinitionId = UA_NODEID_NUMERIC(0, UA_NS0ID_BASEEVENTTYPE);
    selectClauses[x].browsePathSize = 1;
    selectClauses[x].browsePath = (UA_QualifiedName*)
        UA_Array_new(selectClauses[x].browsePathSize, &UA_TYPES[UA_TYPES_QUALIFIEDNAME]);
    if(!selectClauses[x].browsePath) {
        UA_SimpleAttributeOperand_delete(selectClauses);
        return NULL;
    }
    selectClauses[x].attributeId = UA_ATTRIBUTEID_VALUE;
    selectClauses[x].browsePath[0] = UA_QUALIFIEDNAME_ALLOC(0, "ConditionId");
```
Now you will get the ConditionId as NodeId in **handler_events**.

kikia1b2:
@schroeder- ConditionId seems not to be a filed atrribute. Nothing is returned in eventFileds[x]. 
Attached is the modified  tutorial_client_event.c as you said.
Thank you so much.
[tutorial_client_events.zip](https://github.com/open62541/open62541/files/4243608/tutorial_client_events.zip)

schroeder-:
Sry my bad, ConditionId is a special case you get it with this filter:

```
[...]

PR #3540, opened 2020-04-03
Fix ConditionId Search (Alarms and Conditions)

sqadasi:
both BranchId and the passed ConditionId should be equal if we want to set the Event Field "ConditionId" with the main ConditionId.
Using "!UA_NodeId_equal" at this place allows the program to go each time inside the if-clause because BranchId is always equal Null NodeId. As a result, each Event ends up having a ConditionId Field within it.
This should fix the problem in https://github.com/open62541/open62541/pull/3415#issuecomment-582816205 and is related to https://github.com/open62541/open62541/issues/3451#issuecomment-591349071

### DIFF
--- a/src/server/ua_subscription.h
+++ b/src/server/ua_subscription.h
@@ -75,15 +75,6 @@ typedef enum {
   UA_ACTIVE_LOWLOW
 } UA_ActiveState;
 
-/* In the first implementation there will be only one entry in this list
- * conditionBranchId is always NULL. */
-typedef struct UA_ConditionBranch {
-    LIST_ENTRY(UA_ConditionBranch) listEntry;
-    UA_NodeId conditionBranchId;
-    UA_ByteString lastEventId;
-    UA_Boolean isCallerAC;
-} UA_ConditionBranch;
-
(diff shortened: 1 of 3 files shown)


jpfr:
Thanks for the fix.

The code makes a lot of implicit assumptions.
Can you add detailed comments to the use of the structure members for UA_Condition and UA_ConditionBranch?

I was certainly not aware that BranchId always equals Null NodeId and when this is the case.

https://github.com/open62541/open62541/blob/632d490d6dfbec8607ab83e8e12b4ffd677d0e35/src/server/ua_subscription.h#L78

jpfr:
Can be merged with the additional comments.
👍

sqadasi:
@jpfr thank you for the review:) I added some comments as requested.

sqadasi:
@jpfr I've found a related error, which affects events when A&C is disabled. Please don't merge yet.

sqadasi:
should work now:)

jpfr:
Rebased.

jpfr:
Merged. Thanks!

PR #3541, opened 2020-04-03
fix(server): misplaced UA_DEPRECATED macro

jjborter:
Fix error C2143: syntax error : missing ';' before 'type' in MSVC
For MSVC, "UA_DEPRECATED" evaluates to "__declspec(deprecated)" which cannot be placed at the end of a deprecated statement, only in the front.


jpfr:
Thanks for the fix!

PR #3543, opened 2020-04-07
[Review] feat(server): exception-based model

max65482:
Hi!

This PR implements the exception-based model according to OPC UA Part 4 [p. 60] 5.12.1.2.
In short: When setting the sampling interval of a monitored item to 0, the server reports data changes proactively. The value will not be sampled periodically.

For now, only Server calls (e.g. UA_Server_writeValue) trigger a notification. Userland can not trigger notifications for a DataSource yet.

Looking forward to your comments!

Regards,
Max

### DIFF
--- a/include/open62541/plugin/nodestore.h
+++ b/include/open62541/plugin/nodestore.h
@@ -18,6 +18,7 @@
  * / OPC UA services to interact with the information model. */
 
 #include <open62541/server.h>
+#include <open62541_queue.h>
 #include "ziptree.h"
 
 _UA_BEGIN_DECLS
@@ -207,6 +208,9 @@ typedef struct {
     UA_Byte accessLevel;
     UA_Double minimumSamplingInterval;
     UA_Boolean historizing;
+#ifdef UA_ENABLE_SUBSCRIPTIONS
+    LIST_HEAD(MonitoredItems, UA_MonitoredItem) monitoredItems;
+#endif
 } UA_VariableNode;
 
 /**
--- a/include/open62541/plugin/nodestore.h
+++ b/include/open62541/plugin/nodestore.h
@@ -94,19 +94,30 @@ typedef struct {
     UA_ReferenceTargetNameTree refTargetsNameTree;
 } UA_NodeReferenceKind;
 
-#define UA_NODE_BASEATTRIBUTES                  \
-    UA_NodeId nodeId;                           \
-    UA_NodeClass nodeClass;                     \
-    UA_QualifiedName browseName;                \
-    UA_LocalizedText displayName;               \
-    UA_LocalizedText description;               \
-    UA_UInt32 writeMask;                        \
-    size_t referencesSize;                      \
-    UA_NodeReferenceKind *references;           \
-                                                \
-    /* Members specific to open62541 */         \
-    void *context;                              \
-    UA_Boolean constructed; /* Constructors were called */
+struct MonitoredItemsList {
+    struct UA_MonitoredItem *slh_first;
+};
+
+#ifdef UA_ENABLE_SUBSCRIPTIONS
+#define UA_MONITORED_ITEMS_EXTENSION struct MonitoredItemsList monitoredItemQueue;
+#else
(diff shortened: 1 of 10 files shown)


jpfr:
That's a really good feature!

I see three immediate topics I'd like to address before merging.

1. The Event-MonitoredItems are already listening on object nodes.
https://github.com/open62541/open62541/blob/master/include/open62541/plugin/nodestore.h#L275
I see the potential to have a single unified way for attaching MonitoredItems to nodes.
We could have a `struct UA_MonitoredItem *monitoredItemQueue` added to all nodes.
Independent of them being an object, a variable, and so on.

2. If you have the new value at hand, try to drop into the existing method `sampleCallbackWithValue` as soon as possible. This will check internally if the value has really changed. And all kinds of filters can be configured.

3. We have the possibility to attach "external" data sources with a callback.
There the value can change without the server noticing.
We need a way to "trigger" a sample from the outside in such cases.

Keep up the good work!

max65482:
Thanks for your reply!

I have implemented your suggestions 1 and 2.

Point 3 needs a bit more consideration. I am thinking about a simple Server method that triggers all monitored items that are attached to a node: `UA_Server_notifyValueChange(UA_NodeId node);`
What do you think?

mlang-de:
Nice PR!
Shall this only be used/activated if the sampling interval = 0?
IMHO this featured could be used for each sampling interval or not?
Even if the monitored item queue size is bigger then 1 the sampling logic use the old value as long as not UA_Server_notifyValueChange() was called.

max65482:
I think we need to look at two cases:

1. When we have no DataSource attached, there is no reason to sample at all, right? Since the value can only be changed through the API, we can notify about changes instantly.
2. With a DataSource attached, we will have to stick with sampling when a non-zero interval is requested. If I got the specification right, we cannot force the user to actively report changes of a DataSource. It might be an external sensor or any passive component that is unaware of changes itself.

jpfr:
> IMHO this featured could be used for each sampling interval or not?
Even if the monitored item queue size is bigger then 1 the sampling logic use the old value as long as not UA_Server_notifyValueChange() was called.

Good idea. There are positive and negative aspects to be considered though.
That means we have to leave a "hasChanged" flag in every MonitoredItem.
That makes frequent writes (more frequent that the MonitoredItems) more expensive.

> When we have no DataSource attached, there is no reason to sample at all, right? Since the value can only be changed through the API, we can notify about changes instantly.

If there is no datasource and there is a sampling interval, instant sampling could increase the overall work. Because we could sample many many times before the interval runs out.
Also, there is the ABA problems.

If the value is sampled at A, then set to B and reset to A before the next sample, no notification is created. Or only if the filter includes the timestamp of the value for comparison.

> With a DataSource attached, we will have to stick with sampling when a non-zero interval is requested. If I got the specification right, we cannot force the user to actively report changes of a DataSource. It might be an external sensor or any passive component that is unaware of changes itself.

Yes, in embedded systems there can be "passive" components that need to be queried externally for values. Or they write values in a static memory location without a triggering our the control flow.

So we need a public "triggerSampling" methods for zero-interval MonitoredItems with external sources.

max65482:
Ah, I see. Didn't think of the case `writeInteral < samplingInterval`.

Do you have any feelings and thoughts about extending the public API by `UA_Server_notifyValueChange(UA_NodeId node);`?
Maybe with an additional out parameter `UA_Boolean *triggeredNotification`

jpfr:
> Do you have any feelings and thoughts about extending the public API by UA_Server_notifyValueChange(UA_NodeId node);?

I like it.

> Maybe with an additional out parameter UA_Boolean *triggeredNotification

The user doesn't actually care about that, does he?

max65482:
I integrated the method `UA_Server_notifyValueChange(UA_Server *server, UA_NodeId node);`

Please note that MonitoredItems that use `indexRange `are not supported at the moment.

max65482:
Now supporting indexRange.

max65482:
Closing as most aspects of this PR have been covered by #4664.

Issue #3530, opened 2020-03-31
PubSub unicast


Fischer-Thomas:
I just tried to run a server with PubSub unicast and got the message that PubSub creation failed (no multicast address). A few days ago it still worked even though there was the same warning.

https://github.com/open62541/open62541/blob/e2cea8bf405a1a63b1d11024f3b95d8d63ecbca9/plugins/ua_pubsub_udp.c#L114-L140

Is there a reason that unicast is not supported anymore or is it just untested?

jpfr:
This came in as part of PR #3505.
Should be easy to fix. We will still issue a warning though.

Fischer-Thomas:
Hello Julius,

thanks for the fast response.
I just pulled the current master and tried it again.
Sadly it doesn't work. It still tries to join a multicast group even though it recognizes it as unicast.

Here is my output:

![image](https://user-images.githubusercontent.com/45487937/78230222-56b65c00-74d1-11ea-90cb-ffce0c206f94.png)

PR #3553, opened 2020-04-14
fix for unicast

Fischer-Thomas:
As mentioned in [#3530](https://github.com/open62541/open62541/issues/3530) pubsub didn't work for unicast because it tries to join a multicast group even when it is a unicast address.


jpfr:
Great! 👍

PR #3557, opened 2020-04-15
fix(log): Use logger clear callback consistenly

bluhm:
The client did check that logger context is not NULL before calling
the clear() callback.  This does not make sense as the default
config uses UA_Log_Stdout_clear() with a NULL context.  It does
make sense to reset the logger callbacks in the client and server
config clean() functions, so that they cannot be called after that.
Do that consistently for both client and server.


jpfr:
Nice. Thanks.

Issue #3544, opened 2020-04-07
Overriding log_stdout.h with a custom configuration file


saukijan:
## Description

Hello everyone, 

I have been working with open62541 SDK as a server and been looking into implementing a custom configuration for our OPC UA server implementation and part of that custom configuration would be a a custom Logging plugin.

I looked at your [Logging Plugin API](https://open62541.org/doc/1.0/plugin_log.html?highlight=logging) and implemented the following methods in a similar fashion as this pseudo code: 

```c
/* ============================ CustomLogger.h ============================ */
#include <open62541/plugin/log.h>

extern const UA_Logger Custom_Logger_;

void Custom_Logger_log(void *_, UA_LogLevel level, UA_LogCategory category,
                      const char *msg, va_list args);
void Custom_Logger_clear(void *logContext);

/* ============================ CustomLogger.c ============================ */
void Custom_Logger_log(void *_, UA_LogLevel level, UA_LogCategory category,
                      const char *msg, va_list args){
// custom logging api, that I can not show
}
void Custom_Logger_clear(void *logContext) {
// nothing to clear
}
// Forward declaration for ease of use 
const UA_Logger Custom_Logger_clear= {Custom_Logger_log, Custom_Logger_clear, NULL};
```

And do a simple start with my 'main.c' 
```c
#include "CustomLogger.h"
#include <open62541/server.h>

#include <signal.h>
#include <stdlib.h>

static volatile UA_Boolean running = true;
static void stopHandler(int sig) {
    UA_LOG_INFO(Custom_Logger_, UA_LOGCATEGORY_USERLAND, "received ctrl-c");
    running = false;
}

int main(void) {
 signal(SIGINT, stopHandler);
 signal(SIGTERM, stopHandler);

 UA_ServerConfig * configuration_ = (UA_ServerConfig *)malloc(sizeof(UA_ServerConfig)); 
 UA_ServerConfig_setDefault(configuration_ );
 configuration_->logger = Custom_Logger_;

 UA_Server* server = UA_Server_newWithConfig(configuration_);
 UA_StatusCode retval = UA_Server_run(server, &running);
 UA_Server_delete(server);

 return retval == UA_STATUSCODE_GOOD ? EXIT_SUCCESS : EXIT_FAILURE;
}
```

## The Problem

[...]

PR #3567, opened 2020-04-24
[Review] feat(plugin): implement configurable logger for pki_default unit

saukijan:
## Description
This PR removes the use of the standard [log_stdout.h](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/log_stdout.h) from [ua_pki_default.c](https://github.com/open62541/open62541/blob/f71912485ec60a42eebbfd96877bef5268c699e6/plugins/ua_pki_default.c) and implements the use of a more generalized [log.h](https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h) in its place. 

This is achieved by using a `static UA_Logger` variable, which is initially set to `NULL` and set during configuration by one of the following functions: 

* `UA_CertificateVerification_AcceptAll()`
* `UA_CertificateVerification_Trustlist()`
* `UA_CertificateVerification_CertFolders()`

A `NULL` logger is acceptable since  [log.h](https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h)  checks against it, and if it is found, no logging is done, which in principal, allows for full deactivation of the logging plugin during the configuration process.

### DIFF
--- a/examples/server_ctt.c
+++ b/examples/server_ctt.c
@@ -9,10 +9,10 @@
 #define _CRT_SECURE_NO_WARNINGS /* disable fopen deprication warning in msvs */
 #endif
 
-#include <open62541/server.h>
 #include <open62541/plugin/log_stdout.h>
-#include <open62541/server_config_default.h>
 #include <open62541/plugin/pki_default.h>
+#include <open62541/server.h>
+#include <open62541/server_config_default.h>
 
 #include <signal.h>
 #include <stdlib.h>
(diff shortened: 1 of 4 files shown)


saukijan:
This PR needs to change both the default and openssl PKI modules to pass CI tests, thus resolving the `plugins/securityPolicies/openssl/ua_pki_openssl.c` PR mentioned in Issue #3544 as well

saukijan:
This PR is ready for review

jpfr:
Please move the logger from a global pointer into the CertContext struct.
We avoid use the use of non-const global variables in the library.

Good to merge otherwise!

saukijan:
Hey @jpfr, thanks for your input.

> Please move the logger from a global pointer into the CertContext struct.

 I did not notice the `ContextCert` before, it seems a much better fit than global ptr!  I used the [openssl module](https://github.com/open62541/open62541/blob/master/plugins/securityPolicies/openssl/ua_pki_openssl.c#L35) definition as an example and created `DefalutContextCert` structure, to differentiate it from the one defined in `ua_pki_oopenssl.c` 

I also pulled in the latest commits from master, but that seems to have broken unit test runner for me. I keep getting errors from Python telling me that Valgrind logfile do not exist.  Here is the error log: 
[ctest.log](https://github.com/open62541/open62541/files/4668368/ctest.log)

Also azure pipeline keeps failing to download some dependencies for windows setup.

andreasebner:
Thank you for sharing back then. A recent check showed that all the proposed changes have been resolved in the latest versions.

PR #3568, opened 2020-04-24
[Review] refactor(arch): change log_stdout.h into log.h

saukijan:
## Description

This PR changes [ua_log_stdout.h](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/log_stdout.h) header use into it's more generic form [log.h](https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h).

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -29,7 +29,8 @@
 /****************************/
 
 static UA_StatusCode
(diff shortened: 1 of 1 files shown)


jpfr:
In general, the Plugin headers have “speaking names” to make clear what they are for.

Log_stdout is better in that sense.

PR #3569, opened 2020-04-24
[Review] refactor(arch): chagne log_stdout.h to log.h for network_ws.c

saukijan:
## Description

This PR changes [ua_log_stdout.h](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/log_stdout.h) header use into it's more generic form [log.h](https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h) for the [network_ws.c](https://github.com/open62541/open62541/blob/af9767240c10f86b29261d8cdfe3abfd26a63345/arch/network_ws.c)

### DIFF
--- a/arch/network_ws.c
+++ b/arch/network_ws.c
@@ -9,10 +9,12 @@
 #define UA_INTERNAL
 
 #include <open62541/network_ws.h>
-#include <open62541/plugin/log_stdout.h>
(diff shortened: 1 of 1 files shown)


saukijan:
hey @jpfr, I guess this PR is the same as PR #3568 ? If so, close it as well

PR #3570, opened 2020-04-24
[WIP] feat(pubsub): change std_log.h into log.h, implement configurable loger for pubsub_ethernet module

saukijan:
## Description
This PR removes the use of the standard [log_stdout.h](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/log_stdout.h) from [ua_pubsub_ethernet.c](https://github.com/open62541/open62541/blob/master/plugins/ua_pubsub_ethernet.c) and implements the use of a more generalized [log.h](https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h) in its place. 

Simmilarly to [PR#3567](https://github.com/open62541/open62541/pull/3567) this is achieved by using a `static UA_Logger` variable, which is initially set to `NULL` and set during configuration when setting the transport layer by `UA_PubSubTransportLayerEthernet(&config->logger)`

A `NULL` logger is acceptable since  [log.h](https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h)  checks against it, and if it is found, no logging is done, which in principal, allows for full deactivation of the logging plugin during the configuration process.

### DIFF
--- a/plugins/include/open62541/plugin/pubsub_ethernet.h
+++ b/plugins/include/open62541/plugin/pubsub_ethernet.h
@@ -7,12 +7,13 @@
 #ifndef UA_NETWORK_PUBSUB_ETHERNET_H_
 #define UA_NETWORK_PUBSUB_ETHERNET_H_
 
+#include <open62541/plugin/log.h>
 #include <open62541/plugin/pubsub.h>
 
 _UA_BEGIN_DECLS
 
 UA_PubSubTransportLayer UA_EXPORT
-UA_PubSubTransportLayerEthernet(void);
+UA_PubSubTransportLayerEthernet(const UA_Logger *logger);
 
 _UA_END_DECLS
 
--- a/plugins/ua_pubsub_ethernet.c
+++ b/plugins/ua_pubsub_ethernet.c
@@ -6,13 +6,12 @@
  *   Copyright 2019 (c) Wind River Systems, Inc.
  */
 
-#include <open62541/plugin/log_stdout.h>
 #include <open62541/plugin/pubsub_ethernet.h>
 #include <open62541/util.h>
 
 #if defined(__vxworks) || defined(__VXWORKS__)
-#include <netpacket/packet.h>
 #include <netinet/if_ether.h>
+#include <netpacket/packet.h>
 #define ETH_ALEN ETHER_ADDR_LEN
 #else
 #include <linux/if_packet.h>
@@ -22,6 +21,8 @@
 #ifndef ETHERTYPE_UADP
 #define ETHERTYPE_UADP 0xb62c
 #endif
+
+static const UA_Logger *pubsub_ethernet_plugin_logger = NULL;
 
 /* Ethernet network layer specific internal data */
 typedef struct {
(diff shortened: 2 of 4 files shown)


jpfr:
Hey,

Thanks for the effort. I have a few requests from a maintainer point.

- don’t rename to log.h This is one of several possible loggers. Log_stdout is not bad for a logger that writes to stdout.
- don’t use static variables. Several servers can run in parallel in the same process.
We don’t use non-constant static variables ever.
- split essential changes and cosmetic changes in separate commits.
This is impossible to review otherwise.

That said, using the configured logging is very welcome.

saukijan:
Hi @jpfr , 

Thanks for your feed back! I address the issues you mentioned:

> * split essential changes and cosmetic changes in separate commits.
This is impossible to review otherwise.

I removed the clang-formater changes and left in only the PR relevant changes themselves. 

> don’t use static variables. Several servers can run in parallel in the same process.
We don’t use non-constant static variables ever. 

I think my initial comment was not really clear. This change uses a `static const UA_Logger *`, which allows for the pointer itself to be changed, but not it's value and the only way to change the pointer is if entire PubSub Transport Layer is rebuilt. 

> * don’t rename to log.h This is one of several possible loggers. Log_stdout is not bad for a logger that writes to stdout.

I am not sure what you mean by this, [log.h](https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h) is the header that contains the definition for `UA_Logger` struct, which declares the callback footprints that all of the loggers will use, where as [log_stdout.h](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/log_stdout.h) implements said callbacks. 

Correct me if I am wrong, but to me this is a typical Interface and the point of an interface is to separate concrete implementation of a functionality from the user of that functionality.

Thus if [log_stdout.h](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/log_stdout.h) header stays, there is no way to implement a configurable logger, since `UA_Log_Stdout` will be used regardless of what is set within `UA_ServerConfig` structure.

Furthermore, this change still allows for the use of `UA_Log_Stdout` as a logger. Instead of using a direct declaration within [ua_pubsub_ethernet.c](https://github.com/open62541/open62541/blob/master/plugins/ua_pubsub_ethernet.c), this uses the logger that was given during PubSub Transport Layer creation (the call of [UA_PubSubTransportLayerEthernet()](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/pubsub_ethernet.h#L15) function) and by default this is set to `UA_Log_Stdout` by functions like [UA_ServerConfig_setDefault(UA_ServerConfig *config)](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/server_config_default.h#L84)

saukijan:
@jpfr, this PR is ready for review

saukijan:
Ill re-add the [log_stdout.h](https://github.com/open62541/open62541/blob/master/plugins/include/open62541/plugin/log_stdout.h) and try to do something similar to PR #3567 and contexts

saukijan:
Hey @jpfr, I moved `const UA_Logger * logger` to [UA_PubSubConnectionConfig](https://github.com/open62541/open62541/blob/master/include/open62541/server_pubsub.h#L137) struct, however I am unsure who sets it and how [TransportLayerEthernet_addChannel(UA_PubSubConnectionConfig *connectionConfig)](https://github.com/open62541/open62541/blob/master/plugins/ua_pubsub_ethernet.c#L437) is called with `UA_PubSubConnectionConfig` parameter.

jpfr:
The core of the change looks good.
There are some artifacts in the PR that are not needed.

- Changes to submodules
- UA_Boolean configurationFrozen;

Can you clean up the PR and rebase? Then we can merge.

saukijan:
sure thing, I'll commit the fixes next week

saukijan:
sorry for the long delay, applied the changes, ready for review

andreasebner:
Thank you for contributing. The PR is now obsolete due to recent changes in the 1.5 codebase and in master.

Issue #3578, opened 2020-04-30
Connection between the example server_encryption and client_encryption is not encrypted


wenshan1:
## Description
Connection between the example server_encryption and client_encryption is not encrypted even if the Sign and Encrypt is set in the client_encryption.
Please see the Wireshark snapshot: 

![non](https://user-images.githubusercontent.com/5170380/80692500-3bf0fa80-8b04-11ea-8deb-4a91b700069c.PNG)

## Background Information / Reproduction Steps
1.  cd $(OPEN62541) && mkdir build
2.  build the OPEN62541 
3.  Use the tools to generate two certifications: 
    $ ../tools/certs/create_self-signed.py    server
    $ ../tools/certs/create_self-signed.py    client
3.  start the example server_encryption:
   $bin/examples/server_encryption  ~/open62541-cert/server_cert.der  ~/open62541-cert/server_key.der  ~/open62541-cert/client_cert.der
4. start the example client_encryption to connect the server and read data:
     $bin/examples/client_encryption  opc.tcp://localhost:4840  ~/open62541-cert/client_cert.der  ~/open62541-cert/client_key.der  ~/open62541-cert/server_cert.der
    Capturing packets with Wireshark. The wireshark shows: 
     1) the client open a secure channel with http://opcfoundation.org/UA/SecurityPolicy#None
     2) the client get the end point information from the server.
     3) the client create a session and assign the secure channel created before. This is not expected. 
        
5. Use the OPC foundation CTT as the client to connect the server.
6.  Capturing packets with Wireshark, the wireshark shows: 
     the client can connect with the server with the signed and encryption information.
    

Used CMake options:
-DUA_BUILD_EXAMPLES=ON 
-DUA_ENABLE_ENCRYPTION_OPENSSL=ON

```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag):  git master version
 - [ ] Other OPC UA SDKs used (client or server):    
 - [x] Operating system:   Ubuntu18.04
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached   
 - [x] Wireshark network dump attached   
 - [ ] Self-contained code example attached    
 - [x] Critical issue

wenshan1:
The Client code maybe needs to be refactored to support encryption feature.

wenshan1:
[...]

PR #3579, opened 2020-04-30
fix(client) connection between the example server_encryption and client_encryption is not encrypted

wenshan1:
This PR is fix the issue https://github.com/open62541/open62541/issues/3578. 
The client want to connect with the server by encryption.
Root cause: 
  1.   The client open a secure channel with http://opcfoundation.org/UA/SecurityPolicy#None
   2.  The client get the end point information from the server and set it to the client.
   3.  But the client don't close the secure channel, the client still use the secure channel to create/active a new session.

Solution:
1.  If the client isn't set the endpoint information:
     Then  the client open a secure channel with http://opcfoundation.org/UA/SecurityPolicy#None.
     The client get the end point information from the server and set it to the client.
2.  If the client want to connect with encryption:
      set the security policy to the client's channel.
3. connect the server.

Test:  
1. start the example server_encryption:
$bin/examples/server_encryption   ~/open62541-cert/server_cert.der  ~/open62541-cert/server_key.der  ~/open62541-cert/client_cert.der
2. start the example client_encryption to connect the server and read data:
 $bin/examples/client_encryption opc.tcp://localhost:4840 ~/open62541-cert/client_cert.der ~/open62541-cert/client_key.der ~/open62541-cert/server_cert.der
3. Capturing packets with Wireshark, the wireshark shows:
the client can connect with the server with the signed and encryption information.

basic128rsa15

Also test the OPEN62541 client can connect/read data from the server without encryption.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -4,6 +4,7 @@
  *
  *    Copyright 2017-2020 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
  *    Copyright 2017-2019 (c) Fraunhofer IOSB (Author: Mark Giraud)
+ *    Copyright 2020 (c) Wind River Systems, Inc.
  */
 
 #include <open62541/transport_generated.h>
@@ -932,9 +933,11 @@ connectIterate(UA_Client *client, UA_UInt32 timeout) {
         if(client->connectStatus != UA_STATUSCODE_GOOD)
             return client->connectStatus;
 
-        client->connectStatus = UA_SecureChannel_generateNewKeys(&client->channel);
-        if(client->connectStatus != UA_STATUSCODE_GOOD)
-            return client->connectStatus;
+        if (client->channel.remoteNonce.length != 0) {
+            client->connectStatus = UA_SecureChannel_generateNewKeys(&client->channel);
+            if(client->connectStatus != UA_STATUSCODE_GOOD)
+                return client->connectStatus;
+        }
     }
 
     /* Open the SecureChannel */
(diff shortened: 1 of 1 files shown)


wenshan1:
This is a security hole for OPEN62541 client.

jpfr:
Hello @wenshan1 ,

the PR #3549 was merged a few days ago.
This development has started quite some time earlier and also touches on this subject.

Did the PR fix the behavior already?
Otherwise, let's merge your fix on top.

The analysis is very welcome!

wenshan1:
The PR #3549 don't fix this subject. 
Based on the latest source code, I have re-fixed this subject.

This fix Solution:
To open an encrypted security channel, we need to obtain the public certificate and the other end point information from the server first.
If the client isn't set the endpoint information:
a) The client open a secure channel with http://opcfoundation.org/UA/SecurityPolicy#None.
b) The client get the end point information from the server and set it to the client.
c)  The client disconnect the secure channel.

The client open an encrypted security channel with the end point information.

wenshan1:
Check failed because the build environment preparation failed.

jpfr:
Can you have a look at #3592?
That should do it for both the synchronous and asynchronous case.

jpfr:
#3592 has landed.

The fix in #3592 uncovered a bug in the OpenSSL SecurityPolicies.
We had an access to uninitialized data. So there were intermittent fails.
https://github.com/open62541/open62541/commit/391d226c570dd29cb487f467f6941320452b1d96

Can you confirm that fix?

jpfr:
Closed and fix verified via #3590.

PR #3588, opened 2020-05-06
fix(discovery): Correctly filter ServerOnNetwork (ignore case, match multiple)

### LINKED ISSUE #3581: LDS ME  -> ServerOnNetworkCallback returns discovery url with hostname
Hello,

This is not issue it's more like discussion if I understand and use the concepts well.

### 1
I'm running two opc ua servers on two different machines in local network (machines obtain IP address by DHCP). Both servers are LDS with other funcionality like DA, so application type is `_config->applicationDescription.applicationType = UA_APPLICATIONTYPE_SERVER;`

The problem is that `ServerOnNetworkCallback()` returns discovery url with hostname -> `opc.tcp://marek-ubuntu-ntbook:4850` and not with ip address `opc.tcp:192.168.0.221:4850` so the 
registering to LDS server fails, because client can't find server by discovery url with hostname.

What I have tried:
 - initialize discovery urls in server config `_config->applicationDescription.discoveryUrls`- this seems to not work.
 - I know about  `UA_ServerConfig_setCustomHostname()`

What is the correct approach to fix this problem?

### 2
Another issue is that the server is announcing itself every 5s even if I already register to him. So the next registration fails.
image
Should I store already registered LDS and filter out those announcements??

#### [UPDATE]
I found out its repeating just few times and then it stops.
image

But after 5 minutes, the server receives next 36 announcements (as the first time).
Is this good behavior?

### 3
If I want register to more than one announced LDS server, can I use the same client instance for periodic registration?
`UA_StatusCode retVal = UA_Server_addPeriodicServerRegisterCallback(GetServer(), registerClient->GetClient(), endpointUrl, 10 * 60 * 1000, 500, NULL);`

### 4
```
UA_Client_findServersOnNetwork(UA_Client *client, const char *serverUrl,
                               UA_UInt32 startingRecordId, UA_UInt32 maxRecordsToReturn,
                               size_t serverCapabilityFilterSize, UA_String *serverCapabilityFilter,
                               size_t *serverOnNetworkSize, UA_ServerOnNetwork **serverOnNetwork) {
```
`UA_Client_findServersOnNetwork` do not find any servers if I use capability filter.
With discovery client from examples I can find both ua servers, with these capabilities.
image

But If I run method `UA_Client_findServersOnNetwork` with parameters as is shown below, it finds 0 servers ;
```
UA_String* serverCapabilityFilter = (UA_String*)UA_Array_new(3, &UA_TYPES[UA_TYPES_STRING]);
        serverCapabilityFilter[0] = UA_String_fromChars("LDS");
        serverCapabilityFilter[1] = UA_String_fromChars("DA");
        serverCapabilityFilter[2] = UA_String_fromChars("storage");

UA_Client_findServersOnNetwork("opc.tcp://localhost:4850", 0, 0, 3, serverCapabilityFilter , serverSize, servers);
```

FROM OPC UA DOCUMENTATION:
image

Is it not working because I use custom capability like *transport* and *storage* ??

`My point is to filter out correct servers in search phase, without need of search in address space after connection.`

## Checklist
 - [ x ] open62541 Version (release number or git tag): v1.0.1
 - [ Ubuntu 18.04.4 LTS ] Operating system:
 - [ ] Critical issue

Pro:
Fixes #3581

### DIFF
--- a/arch/eCos/ua_architecture.h
+++ b/arch/eCos/ua_architecture.h
@@ -73,6 +73,7 @@ int gethostname_ecos(char* name, size_t len);
 #define UA_realloc realloc
 
 #define UA_snprintf snprintf
+#define UA_strncasecmp strncasecmp
 
 #define UA_LOG_SOCKET_ERRNO_WRAP(LOG) { \
     char *errno_str = strerror(errno); \
--- a/arch/posix/ua_architecture.h
+++ b/arch/posix/ua_architecture.h
@@ -137,6 +137,7 @@ extern void * (*UA_globalRealloc)(void *ptr, size_t size);
 
 #include <stdio.h>
 #define UA_snprintf snprintf
+#define UA_strncasecmp strncasecmp
 
 #define UA_LOG_SOCKET_ERRNO_WRAP(LOG) { \
     char *errno_str = strerror(errno); \
--- a/arch/vxworks/ua_architecture.h
+++ b/arch/vxworks/ua_architecture.h
@@ -99,6 +99,7 @@
 
 #include <stdio.h>
 #define UA_snprintf snprintf
+#define UA_strncasecmp strncasecmp
 
 #define UA_LOG_SOCKET_ERRNO_WRAP(LOG) { \
     char *errno_str = strerror(errno); \
--- a/arch/wec7/ua_architecture.h
+++ b/arch/wec7/ua_architecture.h
@@ -124,6 +124,7 @@ char *strerror(int errnum);
 #endif
 
 #define UA_snprintf(source, size, string, ...) _snprintf_s(source, size, _TRUNCATE, string, __VA_ARGS__)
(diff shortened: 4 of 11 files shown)


Pro:
@jpfr yeah right!
I changed the `UA_String_equals_ignorecase` to internal API and added a corresponding comment.

OK like this, or should I completely remove the `UA_String_equals_ignorecase`?

jpfr:
Looks good!

A last build issue:
```
/home/travis/build/open62541/open62541/tests/server/check_discovery.c:56:23: error: request for implicit conversion from ‘void *’ to ‘UA_String *’ {aka ‘struct <anonymous> *’} not permitted in C++ [-Werror=c++-compat]

     UA_String *caps = UA_Array_new(2, &UA_TYPES[UA_TYPES_STRING]);
```

PR #3592, opened 2020-05-07
Reset SecureChannel if the SecurityPolicy in the Endpoint is different

jpfr:


src/client/ua_client_connect.c
```diff
@@ -756,6 +757,12 @@ responseGetEndpoints(UA_Client *client, void *userdata, UA_UInt32 requestId,
                      "No suitable UserTokenPolicy found for the possible endpoints");
         client->connectStatus = UA_STATUSCODE_BADINTERNALERROR;
     }
+
+    /* Close the SecureChannel if a different SecurityPolicy is defined by the Endpoint */
+    if(client->config.endpoint.securityMode != client->channel.securityMode ||
+       !UA_String_equal(&client->config.endpoint.securityPolicyUri,
+                        &client->channel.securityPolicy->policyUri))
+        closeSecureChannel(client);
```
wenshan1, line 765: Is it needed to initialize a connection by calling initConnect() after the security connection is closed? The code is like this:
   closeSecureChannel(client);
   initConnect(client, endpointUrl);

jpfr: You don't need a second initConnect.
The client automatically recovers the connection, but opens a new SecureChannel with the new SecurityPolicy.

PR #3593, opened 2020-05-07
[review]: feat(ex): add subscribe example for mqtt

waheedejaz:
Note: The received message processing is skipped i.e. address space is not updated upon message reception. Instead, a user callback function displays the received message.

ua_pubsub.h : add support for passing transportSetting and callback in UA_PubSubConnection_regist() function.

check_pubsub_subscribe.c, ua_pubsub_writer.c, tutorial_pubsub_subscribe.c and tutorial_pubsub_mqtt_publish.c have corresponding changes for regist() api.

#2865


jpfr:
@andreasebner Can you shepherd this PR until the merge?

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+static UA_NodeId readerGroupIdentifier;
+static UA_NodeId readerIdentifier;
+
+static UA_DataSetReaderConfig readerConfig;
+void subscriber_message_callback(UA_ByteString *encodedBuffer, UA_ByteString *topic);
+
+void subscriber_message_callback(UA_ByteString *encodedBuffer, UA_ByteString *topic) {
+	UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND, "message received: topic=%s | size=%d.",
+	        topic->data, (int)encodedBuffer->length);
+	UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND, "%s\n.", encodedBuffer->data);
```
andreasebner, line 57: Can we extend this example to match the functionality of the non MQTT subscriber examples? It would be great to decode the received values (most common types) and print the 'real' content.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+
+    if (UA_STATUSCODE_GOOD != retval)
+    {
+    	return retval;
+    }else  {
+    	UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "addPubSubConnection successful.");
+    }
+
+    return retval;
+}
```
andreasebner, line 94: Check only for UA_STATUSCODE_GOOD and remove else case. It's not needed because the method ends returns the retval anyway. Adjust if statement to the open62541 coding conventions.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+
+/* Add ReaderGroup to the created connection */
+static UA_StatusCode
+addReaderGroup(UA_Server *server) {
+    if(server == NULL) {
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
```
andreasebner, line 101: Adjust if statement to the open62541 coding conventions.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+    retval |= UA_Server_addReaderGroup(server, connectionIdentifier, &readerGroupConfig,
+                                       &readerGroupIdentifier);
+    return retval;
+}
+
+/* Define MetaData for TargetVariables */
+static void fillTestDataSetMetaData(UA_DataSetMetaDataType *pMetaData) {
+    if(pMetaData == NULL) {
+        return;
+    }
```
andreasebner, line 117: Adjust if statement to the open62541 coding conventions.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+    pMetaData->fields[3].name =  UA_STRING ("BoolToggle");
+    pMetaData->fields[3].valueRank = -1; /* scalar */
+}
+
+/* Add DataSetReader to the ReaderGroup */
+static UA_StatusCode
+addDataSetReader(UA_Server *server, char *topic) {
+    if(server == NULL) {
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
```
andreasebner, line 167: Adjust if statement to the open62541 coding conventions.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+    transportSettings.content.decoded.type = &UA_TYPES[UA_TYPES_BROKERDATASETREADERTRANSPORTDATATYPE];
+    transportSettings.content.decoded.data = &brokerTransportSettings;
+
+    readerConfig.transportSettings = transportSettings;
+
+    retval |= UA_Server_addDataSetReader(server, readerGroupIdentifier, &readerConfig,
+                                         &readerIdentifier);
+
+    /* Subscribe to MQTT and register the callback function */
+    retval |= UA_PubSubConnection_regist(server, &connectionIdentifier, &transportSettings, &subscriber_message_callback);
```
andreasebner, line 213: I'am a bit unhappy about the changed interface of the regist method (see below). After finishing the Subscriber API the user will not process the value from the subscribe directly in a callback. As far as I know he can only link the incoming value to a node in the information model. We should not add/change an API that will only exist for a short time. Maybe we should move the output from the callback to the receive function (with a hint that this output is for testing and a full implementation is still missing).

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+
+    return retval;
+}
+
+/* Set SubscribedDataSet type to TargetVariables data type
+ * Add subscribedvariables to the DataSetReader */
+static UA_StatusCode
+addSubscribedVariables (UA_Server *server, UA_NodeId dataSetReaderId) {
+    if(server == NULL) {
```
andreasebner, line 222: Adjust if statement to the open62541 coding conventions.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+                             UA_NS0ID_BASEOBJECTTYPE), oAttr, NULL, &folderId);
+
+    retval |= UA_Server_DataSetReader_addTargetVariables (server, &folderId,
+                                                          dataSetReaderId,
+                                                          UA_PUBSUB_SDS_TARGET);
+    UA_free(readerConfig.dataSetMetaData.fields);
+    return retval;
+}
+
+
```
andreasebner, line 255: Avoid multiple blank lines in general.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+static int
+run(UA_String *transportProfile, UA_NetworkAddressUrlDataType *networkAddressUrl) {
+    signal(SIGINT, stopHandler);
+    signal(SIGTERM, stopHandler);
+    /* Return value initialized to Status Good */
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    UA_Server *server = UA_Server_new();
+    UA_ServerConfig *config = UA_Server_getConfig(server);
+    UA_ServerConfig_setMinimal(config, 4801, NULL);
```
andreasebner, line 271: Please use the OPC UA default port.

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,329 @@
...
+    config->pubsubTransportLayersSize++;
+
+    /* API calls */
+    /* Add PubSubConnection */
+    retval |= addPubSubConnection(server, networkAddressUrl);
+    if (retval != UA_STATUSCODE_GOOD)
+    {
+    	UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "Error Name = %s", UA_StatusCode_name(retval));
+    	return EXIT_FAILURE;
+    }
```
andreasebner, line 294: Adjust if statement to the open62541 coding conventions.

examples/pubsub/tutorial_pubsub_subscribe.c
```diff
@@ -57,7 +57,7 @@ addPubSubConnection(UA_Server *server, UA_String *transportProfile,
     if (retval != UA_STATUSCODE_GOOD) {
         return retval;
     }
```
andreasebner, line 59: Adjust if statement to the open62541 coding conventions.

plugins/ua_network_pubsub_mqtt.c
```diff
@@ -189,6 +189,22 @@ UA_PubSubChannelMQTT_unregist(UA_PubSubChannel *channel, UA_ExtensionObject *tra
...
+ * receive a message.
+ *
+ * @return UA_STATUSCODE_GOOD if success
+ */
+static UA_StatusCode
+UA_PubSubChannelMQTT_receive(UA_PubSubChannel *channel, UA_ByteString *buffer,
+                             UA_ExtensionObject *transportSettings, UA_UInt32 timeout) {
+
+    /*TODO: add processing of the received messages to address space*/
+    UA_StatusCode ret = UA_STATUSCODE_GOOD;
```
andreasebner, line 202: Return UA_STATUSCODE_BADNOTIMPLEMENTED instead. For whot purpose ist the following statement there 'buffer->length = 0;'? Idea -> Lets make the user-callback outputs here.

src/pubsub/ua_pubsub.h
```diff
@@ -77,7 +77,8 @@ void
 UA_PubSubConnection_clear(UA_Server *server, UA_PubSubConnection *connection);
 /* Register channel for given connectionIdentifier */
 UA_StatusCode
-UA_PubSubConnection_regist(UA_Server *server, UA_NodeId *connectionIdentifier);
+UA_PubSubConnection_regist(UA_Server *server, UA_NodeId *connectionIdentifier, UA_ExtensionObject *transportSettings,
+                           void (*callback)(UA_ByteString *encodedBuffer, UA_ByteString *topic));
```
andreasebner, line 81: The UA_PubSubConnection_regist is generic for all PubSub transport protocols. The callback should not be needed anymore after finishing the subscribe API and the term and parameter topic is very MQTT specific. We should try to avoid these changes.

jpfr:
Hey @waheedejaz ,

thanks for the contribution.
Do the comments of @andreasebner work for you?

examples/pubsub/tutorial_pubsub_subscribe.c
waheedejaz: Not sure what the problem here is?

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
waheedejaz: This was the easiest way to do it as the regist api matches the channel struct regist function signature here:
[https://github.com/open62541/open62541/blob/ba5cf14f9c39a48fc2cc0fd9a129399b240a3b26/include/open62541/plugin/pubsub.h#L60-L61](https://github.com/open62541/open62541/blob/ba5cf14f9c39a48fc2cc0fd9a129399b240a3b26/include/open62541/plugin/pubsub.h#L60-L61)

Handling the callback in the receive function would mess up more things. In my opinion, we should focus on fixing the pubsup API first.

Here is a suggestion:
This is how the receiver side is handled at the moment:
A repeated callback is created when DataReaderGroup is set to Operational. The callback is hardcoded to be repeated after 5 ms. Callback calls the channel receive function and processes the message if received.
[https://github.com/open62541/open62541/blob/ba5cf14f9c39a48fc2cc0fd9a129399b240a3b26/src/pubsub/ua_pubsub_reader.c#L846-L848](https://github.com/open62541/open62541/blob/ba5cf14f9c39a48fc2cc0fd9a129399b240a3b26/src/pubsub/ua_pubsub_reader.c#L846-L848)

It presents some problems. The foremost being that the address space is not updated as soon as the message arrives or messages can be missed between 5ms cycle. 
I would suggest removing the subscriber_callback of the reader group altogether 
[https://github.com/open62541/open62541/blob/a524accc338c9611a5ec60bf595c1a587243b457/src/pubsub/ua_pubsub_reader.c#L773](https://github.com/open62541/open62541/blob/a524accc338c9611a5ec60bf595c1a587243b457/src/pubsub/ua_pubsub_reader.c#L773)
and handle the receiver side using transport channel callback and yield function.  The yield function is called in UA_Server_run_iterate() function. The MQTT client lib lets the user map a callback function which was mapped to a user-defined function using the regist API in this PR. We can map the same callback 
[https://github.com/open62541/open62541/blob/a524accc338c9611a5ec60bf595c1a587243b457/plugins/mqtt/ua_mqtt_adapter.c#L15](https://github.com/open62541/open62541/blob/a524accc338c9611a5ec60bf595c1a587243b457/plugins/mqtt/ua_mqtt_adapter.c#L15) to ReadGroup_Subscriber_callback() function and rename it to something like ReaderGroup_ProcessIncomingMessage(). All incoming traffic on MQTT would be mapped to this function. 
To support multiple readerGroups and topics, the PubSub channel must maintain the context of all its active ReaderGroups. So, PubSub Channel must store ids of the DataSetReaders. A list of Ids can be added here:
[https://github.com/open62541/open62541/blob/a524accc338c9611a5ec60bf595c1a587243b457/include/open62541/plugin/pubsub.h#L45-L49](https://github.com/open62541/open62541/blob/a524accc338c9611a5ec60bf595c1a587243b457/include/open62541/plugin/pubsub.h#L45-L49)

Considering the MQTT client lib, whenever a message arrives, the lib traverse the list of active DataSetReaders or ReaderGroups, compare the topic name to the reader group transport Settings and if matched invoke the ReaderGroup_subscriberCallback() function. The function would need to adapted a little to take a buffer as an argument in addition to the existing arguments. 

As a result, the PubSubChannel->receive() function will not be required. The same approach can be used for UADP transport also which currently relies on PubSubChannel->receive() function.

Let me know what you think. If it seems ok, then I will create a PR for PubSub API change.

waheedejaz:
@andreasebner Please see the following comment:
https://github.com/open62541/open62541/pull/3593#discussion_r436421841

examples/pubsub/tutorial_pubsub_subscribe.c
andreasebner: We normaly use no brackets for single line if statements ;)

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
andreasebner: Thanks’ for your updates and sorry for my late response. I've started a deep dive into the OPC UA standard and our current implementation. I think especially for the subscriber message processing flow is an overhaul needed. However, changes on these code paths require prudence because the must fit together with several PubSub protocols besides MQTT and should not affect our RT-PubSub work.

> The foremost being that the address space is not updated as soon as the message arrives or messages can be missed between 5ms cycle.

I think we will not lose some messages because the incoming messages should be buffered automatically. You're right, the fastest possible update rate for the information model would be 5ms. It would certainly be possible to process the incoming MQTT traffic in the open62541-mainloop, but this would still be a best effort approach because the user can create the server by UA_Server_run_startup and call the UA_Server_run_interat function by itself.

During the review of your changes I noticed firstly the existence of both functions 'UA_PubSubChannel_regist' and 'UA_PubSubConnection_regist'. From my point of view, we should
mark the UA_PubSubConnection_regist as deprecated and remove this function. The regist function is a low-level call which should not be needed (for the user). The necessary information’s (e.g. topic or multicast address) are part of the OPC UA PubSub config and should be extracted automatically. Therefore, the regist call should only be done by the channel interface (internally).


> So, PubSub Channel must store ids of the DataSetReaders.


During the initial PubSub implementation I tried to add clear abstraction layers. Therefor the channel should contain only generic transport variables and functions.
I think the discussion about the PubSub subscribe overhaul is very important. Your solution seems closely related to MQTT and MQTT-C library. I think your proposal with the yield-call within the server mainloop was related to the yield function of the MQTT-C library. 

We should also think about about resource constraint devices where the MQTT-C mainloop is triggered by the user. Therefore, we should hold the main control flow within our open62541 implementation. 
After reading the OPC UA PubSub specification again, I would propose to remove the link between the DataSetReader and the subscriber callback (sb). In my opinion the sb fits better to the PubSubConnection because all incoming messages must be filtered based on DataSetReader settings which are linked to the PubSubConnection (with the ReaderGroup in between).

The callback will trigger (protocol independent) the receive function of the linked channel. Within the CB we can call ‘DataSetReader_processMSG’ if the filters match the configuration of a DSR. 
Like mentioned before, I would also remove the UA_PubSubConnection_regist because the registration is done during the DataSetReader creation under the hood. The subscribe CB cycle should be configurable for the user but still not be part of the server mainloop in my opinion. If the user need's RT behavior we can allow to trigger the sb by external user code (like shown in the rt examples). 

Today I can’t give you a concrete answer how the API will look like finally and which changes we make. I will start a discussion with the other initial developers of the open62541-PubSub code. The base idea of adding a MQTT subscriber example and compete the subscriber is very important and I have now more clarification about the implementation.

basyskom-dege:
Hello! 

I just wanted to ask about the state of this PR. There seems to be some change requests and concerns about the API changes. Is this resolved somehow? Will there be an update in the future? I tested this locally and it worked fine for me and I would appreciate a subscribe example to work with.

andreasebner:
Thank you for the contribution. The PR #5234 landed now in our mainline and introduces MQTT subscribing.

PR #3594, opened 2020-05-08
fix(client) don't generate new key if the client remote nonce is NULL

wenshan1:
If the client don't get the nonce from the remote server, the UA_SecureChannel_generateNewKeys() should not be called to generate the new key.


jpfr:
👍

PR #3601, opened 2020-05-14
fix(client): optional reqId in client read async

bluhm:
In all other async functions the request id is optional, but
__UA_Client_readAttribute_async crashes if it is NULL.  Pass local
storage to __UA_Client_AsyncService() and set the value of reqId
parameter only if the pointer is not NULL.

For that allocate the custem callback before the call to async
service.  This has the additional advantage that in case of memory
allocation error, the read attribute async has not done half the
work.  It would fail early instead.


jpfr:
Thanks!

Issue #3535, opened 2020-04-02
PubSub: Publish/Subscribe callback design (messageReceiveTimeout)


MartinaFelberFEM:
Hi!

I have had a look at the PubSub publish/subscribe callback implementation and was wondering how an application (particularly the Subscriber) can get any information about the status of sent and received network messages.

I have tested the PubSub implementation in a custom realtime environment with overwritten PubSub callback handling (CMAKE flag UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING). My use case is similar to your example: https://github.com/open62541/open62541/blob/master/examples/pubsub_realtime/TSN_ETF_loopback.c

At the subscriber thread routine the subscriber calls the callback to receive network data. 
https://github.com/open62541/open62541/blob/285637fc519e2aa92142cc62f280d0226da412d1/examples/pubsub_realtime/TSN_ETF_loopback.c#L647-L648

The callback definition does not provide a statuscode or any diagnostic parameter:
https://github.com/open62541/open62541/blob/285637fc519e2aa92142cc62f280d0226da412d1/include/open62541/server.h#L97

So, how can the Subscriber (and Publisher) know if anything went wrong?
I know that there's a PubSubDiagnosticsType specified by the Opc Foundation, which would provide some details [https://github.com/open62541/open62541/issues/3354](url), but in this use case 

- I need to immediately (realtime requirements) react to errors

-  and I need to know if subscribed DataSets have been received within configured MessageReceiveTimeout (see PubSub specification 6.2.8.6)
![image](https://user-images.githubusercontent.com/50486106/78216207-98d3a380-74b9-11ea-9f29-1ca5192966a8.png)

The messageReceiveTimeout can be configured for every DataSet. 

I thought about changing the Callback function defintion (adding status code and a diagnostic parameter or something) to enable the application to notice some errors (encryption/decryption, network message could not be parsed etc.) but I have a problem with the messageReceiveTimeout.

Because the subscribe callback handles a ReaderGroup. A ReaderGroup can contain multiple DataSets with different messageReceiveTimeouts. So I am not sure how to handle this.

https://github.com/open62541/open62541/blob/285637fc519e2aa92142cc62f280d0226da412d1/src/pubsub/ua_pubsub_reader.c#L280-L303

Many thanks for your help

opcua-tsn-team-kalycito:
Hello @MartinaFelberFEM ,
[...]

PR #3616, opened 2020-05-20
[REVIEW] PubSub: Subscriber MessageReceiveTimeout

MartinaFelberFEM:
See issue: https://github.com/open62541/open62541/issues/3535


MartinaFelberFEM:
Hi @andreasebner and @opcua-tsn-team-kalycito !

As mentioned in the last conference call I am trying to implement the handling of MessageReceiveTimeout on the subscriber side. And I thought that a status information about the successful publishing/subscribing operation in general would be nice to have for the application. Particularly in a realtime environment when the _UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING_ build option is enabled.

I am struggling a bit and would like to ask for your help/opinions.

I have started to refactor the Publish/Subscribe callback definition and added a StatusCode. 
`typedef UA_StatusCode (*UA_ServerPubSubCallback)(UA_Server *server, void *data);` 

But this breaks the `UA_Timer_addRepeatedCallback` definition. I am not sure if I should refactor the timer implementation. Or if the new callback definition is a bad idea in general, because it's not necessary. The application could check the PubSubDiagnostics (when this feature is available) or the StatusCode of the targetVariables at the Subscriber side. Maybe that's sufficient?


**Regarding the MessageReceiveTimeout:**
At the moment the UA_ReaderGroup_subscribeCallback is called every 5 ms. The receive method of the underlying UDP plugin uses a 1 ms timeout (select timeout).
I've added a counter to the DataSetReader structure to count the ms since the last received message. So I can calculate the MessageReceiveTimeout for every DataSetReader of a ReaderGroup.
It should be sufficient to call the UA_ReaderGroup_subscribeCallback in a loop (with the 1 ms timeout of the select call) until an error occurs or the state of the ReaderGroup changes to paused/disabled. I hope it works with the UA_Timer callback with 1 ms interval as well.


According to the specification the DataSetReader state shall be set to Error if a MessageReceiveTimeout occurs. This is simple: `UA_DataSetReader_setPubSubState(server, UA_PUBSUBSTATE_ERROR, pDataSetReader);`

Then the TargetVariables shall be updated according to the overrideValue configuration:

image

image

I think the override values are not implemented yet... I could just set the StatusCode to BadNoCommunication for all target variables of the DataSetReader.

```
        UA_WriteValue writeVal;
        UA_WriteValue_init(&writeVal);
        writeVal.attributeId = dataSetReader->subscribedDataSetTarget.targetVariables[i].attributeId;
        writeVal.nodeId = dataSetReader->subscribedDataSetTarget.targetVariables[i].targetNodeId;
        writeVal.value.hasStatus = UA_TRUE;
        writeVal.value.status = UA_STATUSCODE_BADNOCOMMUNICATION;
        retVal = UA_Server_write(server, &writeVal);

```

But a public method UA_Server_GetPubSubState or something to check an error in the application would be nice. Then the application would not have to go through all targetVariables and check if there's a bad statuscode to notice that something went wrong.


**Regarding the Publisher:**
I think if an unrecoverable errors occurs (bad memory allocation, etc.) then the WriterGroup state is set to error. 
I am not sure if decoding errors, etc. should set the state to error as well or just generate a bad statuscode within the DataSetmessage.
Depending on the configuration (Substitute/DefaultValue configured) the publisher writes different values/status codes to the DataSetMessage.

image


Can you give me some input if I should change the callback definition or not? Or if the rest of my thoughts do make any sense?

PS: the current PR implementation is not up to date.

andreasebner:
Thank you for the detailed report and background explanations. A lot of these points are important and currently missing. 

The change of the callback signature is a far-reaching change which probably would have a lot of side effects/code breaks. Why is the StatusCode important for you? /Where should the StausCode be handled (the call happens in the main loop automatically)? Inside the callback we have all options to change the PubSub state or set internal flags to represent an error. I would avoid using the PubSub diagnostics (when available) for the timeout information because the diagnostics will be an optional feature and not be available in the default configuration.

Some background about the repeated callback mechanism: The 5ms timing is part of the PubSub code because in earlier open62541 version, this 5ms was the fastest possible setup https://github.com/open62541/open62541/blob/v0.3.1/src/ua_timer.c#L99-L100. As far as I know, this restriction was removed in later versions. But anyway, the repeated callback mechanism is still a 'best effort' service. Especially if the user itself control the server main loop by calling the iterate function. The 1ms select timeout means that the select call is blocking for 1 ms. It's important to have fast code paths especially by often called repeated callbacks like the subscribe callback. Your MessageReceiveTimeout detection approach sounds working but very resource hungry. We should think about a method with less effort. I’m also not sure if this fine timeout resolution is needed/possible. How do you think about a new independent repeated callback which is capable for all the timeouts/heartbeats? This callback could check the PubSub configuration (including the DataSatReaders) if e.g. a MessageTimeOut happened. The interval of this callback can be made configurable by the user or automatically calculated based on the smallest common divisor of all timeouts/heartbeats (with a checked lower bound). 

With TSN in mind, we are generally trying to keep the basic publish and subscribe publish callbacks as simple as possible to allow a fast execution and code optimizations.
"But a public method UA_Server_GetPubSubState or something to check an error in the application would be nice."
-> You're right, this is missing. We discussed about appending the PubSub status to the PubSub config structure for each entity and make it available by the getConfig functions. But I agree with you, an dedicated function UA_Server_GetPubSubState(UA_Server *sever, UA_NodeId node) is more obvious. I've added this to my ToDo list. 

"I think if an unrecoverable errors occurs (bad memory allocation, etc.) then the WriterGroup state is set to error."
-> Yes, this makes sense to me. The PubSub states are monandry and were included during the RT-freeze concept works. For sure, there are a lot of internal error code paths which should affect the PubSub state.

I hope I was able to clarify a few points and my opinion to some implementation details make sense to you. For sure, these points are for discussion 😉

MartinaFelberFEM:
Many thanks for your answer!

Regarding your comments/questions:

> Why is the StatusCode important for you?

   I am using the build option UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING.
   I start realtime tasks for Writer- and ReaderGroups. Within their task loop I thought it would be nice to be able to react to errors immediately.
   E.g. here: https://github.com/open62541/open62541/blob/2087b670cb1dfe437b004c7ff1dc4bf5e6d74091/examples/pubsub_realtime/pubsub_TSN_loopback.c#L693
If something in the subCallback (or publisher callback) went wrong I could immediately react to it and set an error or something in my realtime application. E.g. Stop machine immediately if we haven't received new data after the MessageReceiveTimeout.
   
But a separate call to a function like UA_Server_GetPubSubState(UA_Server *sever, UA_NodeId node) at the end of each loop iteration to check the PubSub state would be ok too.
   
>    The 1 ms select timeout means that the select call is blocking for 1 ms. It's important to have fast code paths especially by often called repeated callbacks like the subscribe callback. Your MessageReceiveTimeout detection approach sounds working but very resource hungry. We should think about a method with less effort. I’m also not sure if this fine timeout resolution is needed/possible. 

You are right. The 1 ms timeout was just the easiest solution for me at the moment.
   

>    How do you think about a new independent repeated callback which is capable for all the timeouts/heartbeats? This callback could check the PubSub configuration (including the DataSatReaders) if e.g. a MessageTimeOut happened. The interval of this callback can be made configurable by the user or automatically calculated based on the smallest common divisor of all timeouts/heartbeats (with a checked lower bound).

   
Sounds very good to me, just not sure how to do it :)
      Do you think about extending this implementation? https://github.com/open62541/open62541/blob/master/src/ua_timer.h 
      Or providing a second ua_timer implementation for this use case?
      
I am not sure if we need a callback on the subscriber side at all. Wouldn't a loop with a select+timeout call be sufficient?

But you are right, the select timeout should be set to the smallest common divisor to avoid the overhead. In my use case with build option UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING I would call the subscriber callback in an endless loop with the select timeout. Shall the general logic of retrieving the configuration and calculating the smallest common divisor be a separate public interface? For those who want to handle the callbacks themselfs?

Thanks for your support!

andreasebner:
“But a separate call to a function like UA_Server_GetPubSubState(UA_Server *sever, UA_NodeId node) at the end of each loop iteration to check the PubSub state would be ok too.”

- I think this should be our suggested solution. Only one thing I like to mention: This function (in my planning’s) will return the PubSub status of the requested PubSub configuration entity e.g. the current ReaderGroup.

“Sounds very good to me, just not sure how to do it :)
Do you think about extending this implementation? https://github.com/open62541/open62541/blob/master/src/ua_timer.h
Or providing a second ua_timer implementation for this use case?”

- I think there is no reason to extend the ua_timer implementation. I would start with an ‘pubSubReflection-callback’ which is implemented in the PubSubManager. All mentioned things like the heartbeat or timeouts are standard defined behaviors. So, it should not be necessary for the user to add his own code. Therefore, we can check all these requirements with internal open62541 defined code. For sure, e.g. for the heartbeat, the transport layer must support the concrete code to send the heartbeat-message. From my point of view, it would be enough if we define the fastest reflection cycle time with 5ms. If the user needs more resolution we can think about the possibility to call this callback from external in the long term. (Same like publish and subscribe callback).

- What cyclic checks have you find out now? -> Heartbeat, Receive-timeout, ..?

- The logic of this function would be like this snipped: UA_StatusCode pubSubReflectionCallback(…) for each writergroup for each datasetreader for each field -> timeout exceeded? -> Set the DataSetReader state … and so on.

“I am not sure if we need a callback on the subscriber side at all. Wouldn't a loop with a select+timeout call be sufficient?”

- Hey @opcua-tsn-team-kalycito, are there any preffered solutions?
“But you are right, the select timeout should be set to the smallest common divisor to avoid the overhead. In my use case with build option UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING I would call the subscriber callback in an endless loop with the select timeout. Shall the general logic of retrieving the configuration and calculating the smallest common divisor be a separate public interface? For those who want to handle the callbacks themselfs?“

- I think there is no rule of thumbs which are the right timings for the subscriber callback. The user has to decide between blocked resources and the need for RT updates. In my last message, I mentioned the smallest divisor with regard to the ‘reflectionCallback’. The call frequency could be static or be based on the check with the highest frequency (e.g. the smallest timeout).

andreasebner:
FYI: There is an open PR regarding the MQTT subscriptions. I've started to read the PubSub standard again and from my point of view we should overhaul the message handling structure in the subscriber. This proposal should not affect your work here but I think you are very familiar with the standard and therefor your feedback is highly welcome ;)
https://github.com/open62541/open62541/pull/3593/files#r440232003

MartinaFelberFEM:
Hi @andreasebner !

Many thanks for your help and thanks for informing me about the other pull request and your suggestions about the subscriber API overhaul. I'll try to give some feedback if something comes up to my mind.

Regarding this PR:

I have added a getState method for Reader- and WriterGroups and DataSetReaders and -Writers.
```
   UA_Server_WriterGroup_getState
   UA_Server_DataSetWriter_getState
   UA_Server_DataSetReader_getState
   UA_Server_ReaderGroup_getState
```

I think the public DataSetWriter functions do not comply with the other public PubSub component functions names.
   e.g. `UA_Server_getDataSetWriterConfig` vs. `UA_Server_DataSetReader_getConfig`
So I used `UA_Server_DataSetWriter_getState` instead of `UA_Server_getDataSetWriterState`

Sidenote: I think there should be a getState method for the Connection as well, but the UA_PubsubConnection structure does not have a state ...
https://github.com/open62541/open62541/blob/211f1be5040ff6330f518a0a05c9424b4b08bcf1/src/pubsub/ua_pubsub.h#L59-L72

image

-----------------------------------------------------------------------------


Then I tried to implement a simple reflectionCallback. To start with I used the suggested constant 5 ms interval, but I will add a function to calculate the interval for a given configuration.

But I would need your help/review before that.

1. 
I wasn't sure if you thought about 1 global reflectionCallback or 1 reflectionCallback for every connection.
I have started with 1 global reflectionCallback.
I am not sure where/when I should add and remove the reflectionCallback. At the moment I am adding the callback at the first call of 
`UA_Server_addPubSubConnection`. And remove it at `UA_Server_removePubSubConnection `if it's the last connection. Do you think that's ok?

2.
But at this stage (UA_Server_addPubSubConnection) there's no configuration for Readers or Writers yet. So I do not have any information about the publishing intervals and messagereceivetimeout etc.
So I don't know when I can calculate a "good" reflectionCallback interval. 
I could use a default value and calculate a new interval every time when UA_Server_setWriterGroupOperational and UA_Server_setReaderGroupOperational gets called using UA_Timer_changeRepeatedCallbackInterval ... I think I'll try that.

I only get a problem when someone calls `UA_Server_setWriterGroupOperational()` before adding the DataSetWriter configuration. 

I think this could be a general PubSub API problem. 
-> If the application calls `UA_Server_setWriterGroupOperational `before adding the DataSetWriter I think the DataSetWriter pubsub state is still DISABLED (initialization value) and I do not have the configuration information of the DataSetWriter.
But in the examples e.g. server_pubsub_publisher_iop.c or tutorial_pubsub_publish.c the WriterGroup is set to operational before adding the DataSetWriter. The same applies to the reader side ...

3.
  

>  If the user needs more resolution we can think about the possibility to call this callback from external in the long term. (Same like publish and subscribe callback). 

   
I think the reflectionCallback needs to be callable from external like publish and subscribe callback to ensure the realtime requirements and to give the user the possibility to set the task priorities etc. 


Many thanks for your help and review!

MartinaFelberFEM:
Hi @andreasebner and @opcua-tsn-team-kalycito !

I have abandoned the reflection callback implementation due to following reasons:

- I think that ms precision for real-time requirments is necessary

- I can't use one global callback to check every DataSetReader timeout, because there can be many different publishers which are all asynchron. So I can't be in sync with all of them and check the receive timeout. Except the global callback interval would be 1 ms, which would provide a lot of overhead.

- Then I thought about using one reflection callback for every ReaderGroup. So that I can be in sync with the specific WriterGroup. But this adds unnecessary complexity as well, because you have to calculate the gcd for all DataSetReaders within the group and you have to calculate if the timeout has occurred within the callback for reader1, reader2, etc.

That's why I would suggest to use a timer (sort of like a watchdog) for every DataSetReader:
On first receivement of a network message -> start a timer for the DataSetReader.
If we receive another message we stop the timer and start it again.
If we do not receive a message within the timeout, the timeout handler get's called and we set the Reader state to error.
If we are receiving network messages again we set the state back to operational.
Every DataSetReader has it's own timer (Watchdog).

It's easier because we do not have to calculate anything.
The SDK provides a simple mechanism based on the UA_TimerCallback.
The PubSub realtime examples can use linux timer_create (custom handling).

What do you think about that?

Things I am not sure about:

1. I had to remove the call to  UA_WriterGroup_publishCallback when the callback gets added. Because then the Subscriber would have received something and the timeout check would have started immediately, although the connection wasn't running.

https://github.com/open62541/open62541/blob/7c251a31550d913958f16d1fad09c5824bf85aae/src/pubsub/ua_pubsub_writer.c#L2110-L2112

I haven't understood why this happens anyways. Why do you execute the publish and subscribe callback immediately at configuration stage? I am using the custom publish/subscribe handling in my application and I do not start the PubSub connection immediately after configuration. 

2. I have changed the DataSetReader statemachine logic:
https://github.com/open62541/open62541/blob/7c251a31550d913958f16d1fad09c5824bf85aae/src/pubsub/ua_pubsub_reader.c#L1065-L1079
I haven't understood why it has been prevented to change from state error back to another state.
Because I think in general the PubSub statemachine would allow this.

image

image

I think the DataSetReader state shall be set back to operational if we are receiving data again.


And I am having problems with the unittest. It works locally on my machine. But it never works in the CI. 
I don't know why. It seems to be a timing problem.

Many thanks for your help

MartinaFelberFEM:
Hi @jpfr or @andreasebner or someone who could help me, please.

I would need some help regarding my unit test.

I tried to test normal PubSub operation first. Setting a node value, calling UA_Server_run_iterate() and checking afterwards if the value was received at the subscriber node.
As it's implemented here: https://github.com/open62541/open62541/blob/ad957ffc4b09b893c509dcf9bde67a4d20562a98/tests/pubsub/check_pubsub_subscribe.c#L612

https://github.com/open62541/open62541/blob/ad957ffc4b09b893c509dcf9bde67a4d20562a98/tests/pubsub/check_pubsub_subscribe.c#L768-L785

It works fine on my local machine using linux. But it does not work on Windows in the CI build and I don't now why. Maybe because there's something wrong with my UA_comboSleep() implementation? It seems like the PubSub callbacks are not called.

I think that the unit test check_pubsub_subscribe from where I have copied the code is not executed in the CI either. The binary gets built and works on linux, but it's not in the CI build ... 

I tried to reproduce the failing unit test on windows, but I can't build on my local machine. Due to the missing check library for the unit test. I have installed it with vcpkg and vcpkg claims that it's installed, but the cmake target can't be found. I don't know why. 

Could you please help me reproduce the error or give me a hint why UA_Server_run_iterate() does not seem to execute the PubSub callbacks?

Many thanks!

andreasebner:
Hello @MartinaFelberFEM , I checked our Azure CI. We are executing our unit tests on the vs2017 environment. The test check_pubsub_subscribe_rt_levels test is also included there (which uses receive).

MartinaFelberFEM:
Hi @andreasebner !

Many thanks for your help!

I was using 2 connections and I had a wrong UDP address configuration for one connection (interesting that it worked with Linux though, don't know why ...)

Publish and Subscribe is now working and I have implemented your refactoring suggestions.

I think the PR is ready for a first review, but I have following questions:

**1. Use case disable/enable writer- and readergroups at runtime**

I tried to test the following:

- writer- and readergroup are running
- disable readergroup
- writer is still running
- disable writergroup
- enable readergroup

   -> The reader receives messages from the past -> although writer is not currently running, I am not sure if we should empty the buffer in this case, before we enable the reader again, otherwise it receives old messages and the timeout handling starts ...

**2. UpdateConfig**

 I have seen that the update pubsub config feature is not yet finished but implemented in some parts.
 I am not sure if during an update the writer- and readergroups will be disabled and afterwards enabled from the user, or if the SDK should do this.
   

Many thanks for your help and best regards

Martina

plugins/ua_log_stdout.c
```diff
@@ -10,7 +10,7 @@
 
 #include <stdio.h>
 
-#if UA_MULTITHREADING >= 200
+#if UA_MULTITHREADING >= 100
```
andreasebner, line 13: Can you explain the background of this change?

src/pubsub/ua_pubsub.h
```diff
@@ -30,6 +30,14 @@ typedef struct UA_WriterGroup UA_WriterGroup;
 struct UA_ReaderGroup;
 typedef struct UA_ReaderGroup UA_ReaderGroup;
 
+typedef enum  {
+    UA_PUBSUB_COMPONENT_CONNECTION,
+    UA_PUBSUB_COMPONENT_WRITERGROUP,
+    UA_PUBSUB_COMPONENT_DATASETWRITER,
+    UA_PUBSUB_COMPONENT_READERGROUP,
+    UA_PUBSUB_COMPONENT_DATASETREADER
+} UA_PubSubComponentEnumType;
```
andreasebner, line 39: Alternatively, we could create a container structure with this enum type and a union with all possible pubsub types (as ptr). This would prevent the change of all the PubSub structures.

tools/azure-devops/win/build.ps1
```diff
@@ -66,6 +66,7 @@ try {
                 -DUA_ENABLE_PUBSUB:BOOL=ON `
                 -DUA_ENABLE_PUBSUB_DELTAFRAMES:BOOL=ON `
                 -DUA_ENABLE_PUBSUB_INFORMATIONMODEL:BOOL=ON `
+                -DUA_LOGLEVEL:INT=0 `
```
andreasebner, line 69: Can you explain the background of this change a little?

tests/pubsub/check_pubsub_subscribe_msgrcvtimeout.c
```diff
@@ -0,0 +1,1511 @@
...
+static UA_NodeId ExpectedCallbackComponentNodeId;
+static UA_UInt32 MsgRcvTimeoutCallbackTriggerCnt = 0;
+
+static UA_UInt32 ExpectedNoOfTimeouts = 0;
+static UA_NodeId *pExpectedTimeoutIds = 0;
+
+
+/***************************************************************************************************/
+/* Note: unit test-plugins can't be used because PubSub does not work with testing_networklayers, 
+that's why we have copied the implementation for (UA_comboSleep) here */
```
andreasebner, line 23: I think this is fixable and we should allow at least the use of e.g. the timer functions here. Maybe you can open a issue for that.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -1994,8 +2024,10 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
     UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "Publish Callback");
 
     if(!writerGroup) {
-        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
                        "Publish failed. WriterGroup not found");
+        /* TODO: we can't set the state of the specific writerGroup to error, because the pointer is invalid ...
+            we would have to set the connection state to error */
```
andreasebner, line 2030: You're right, but this function is internal and the case writerGroup is NULL should never happen.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -2006,25 +2038,32 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
     /* Binary or Json encoding?  */
     if(writerGroup->config.encodingMimeType != UA_PUBSUB_ENCODING_UADP &&
        writerGroup->config.encodingMimeType != UA_PUBSUB_ENCODING_JSON) {
-        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
                        "Publish failed: Unknown encoding type.");
+        UA_WriterGroup_setPubSubState(server, UA_PUBSUBSTATE_ERROR, writerGroup);
```
andreasebner, line 2043: This change will set the state to Error but the WriterGroup will anyway continue because currently the UA_WriterGroup_setPubSubState only changes the state.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1060,9 +1136,39 @@ UA_DataSetReaderConfig_copy(const UA_DataSetReaderConfig *src,
...
+
+    UA_DataSetReader *currentDataSetReader =
+        UA_ReaderGroup_findDSRbyId(server, dataSetReaderIdentifier);
+    if(!currentDataSetReader)
+        return UA_STATUSCODE_BADNOTFOUND;
+
+    *state = currentDataSetReader->state;
+    return UA_STATUSCODE_GOOD;
+}
+
```
andreasebner, line 1167: Some functions can be formatted a little bit more compact.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -786,17 +803,42 @@ UA_DataSetReader *UA_ReaderGroup_findDSRbyId(UA_Server *server, UA_NodeId identi
...
  * contained DataSetMessages. */
 void UA_ReaderGroup_subscribeCallback(UA_Server *server, UA_ReaderGroup *readerGroup) {
+
+    UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "PubSub subscribe callback");
+
+    if(!readerGroup) {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+            "Subscribe failed. ReaderGroup not found");
+        /* TODO: we can't set the state of the specific ReaderGroup to error, because the pointer is invalid ...
+            we would have to set the connection state to error */
```
andreasebner, line 813: I think this is an edge case because there should be absolutely no reason for this null pointer. The user never gets access to the internal ReaderGroup pointer and therefore I think the error log is enough.

src/pubsub/ua_pubsub_writer.c
MartinaFelberFEM: ok, I have removed the comment at this PR: https://github.com/open62541/open62541/pull/3899

src/pubsub/ua_pubsub_writer.c
MartinaFelberFEM: You are right. I thought setting the state to error would be better than nothing.
Shall I remove the callback in case of an error?

src/pubsub/ua_pubsub_reader.c
MartinaFelberFEM: ok, changed it here: https://github.com/open62541/open62541/pull/3899

src/pubsub/ua_pubsub_reader.c
MartinaFelberFEM: ok, I have removed the comment

tools/azure-devops/win/build.ps1
MartinaFelberFEM: This change will be deleted
I wanted to enable more traces, because I think that PubSub is sometimes not working
I can see that the publisher writes the value to the buffer, but it's not received (sometimes)

MartinaFelberFEM:
Hi @andreasebner !

I have removed the ua_timer mutex implementation, so you can check the implementation.
But I think for multithreading tests a mutex within ua_timer is necessary.

I have changed the callback in server.h to a struct.
I have added the struct in server_pubsub.h
I couldn't think of a better name than "UA_PubSubConfiguration", because I don't really know what you have planned with the struct. I can rename and/or relocate the struct anytime. Just let me know.

MartinaFelberFEM:
Hi @opcua-tsn-team-kalycito !

I would have a question regarding the subscriber.

I tested following use case:

- Publisher and Subscriber are running

- disable Subscriber, Publisher is still running

- wait some time

- enable the Subscriber again -> I think the Subscriber immediately receives the all the "old messages"

Is this correct? Or shall we flush the socket before enabling the subscriber?

Thanks for your help

examples/pubsub/tutorial_pubsub_subscribe.c
```diff
@@ -93,11 +93,12 @@ addDataSetReader(UA_Server *server) {
...
-    readerConfig.publisherId.type = &UA_TYPES[UA_TYPES_UINT16];
-    readerConfig.publisherId.data = &publisherIdentifier;
-    readerConfig.writerGroupId    = 100;
-    readerConfig.dataSetWriterId  = 62541;
+    UA_UInt16 publisherIdentifier      = 2234;
+    readerConfig.publisherId.type      = &UA_TYPES[UA_TYPES_UINT16];
+    readerConfig.publisherId.data      = &publisherIdentifier;
+    readerConfig.writerGroupId         = 100;
+    readerConfig.dataSetWriterId       = 62541;
+    readerConfig.messageReceiveTimeout = 200.0; /* Timeout must be greater than publishing interval of corresponding WriterGroup */
```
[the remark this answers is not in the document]
MartinaFelberFEM: Yes, because it's only used for PubSub component timeout handling -> DataSetReader MessageReceiveTimeout
If this feature is enabled the parameter must be set, otherwise the timeout occurs immediately

examples/pubsub/tutorial_pubsub_subscribe.c
```diff
@@ -227,6 +228,29 @@ static void stopHandler(int sign) {
     running = false;
 }
 
+#ifdef UA_ENABLE_PUBSUB_TIMEOUT_HANDLING
+/* Provide a callback to get notifications about specific PubSub state changes or timeouts.
+    Currently only the MessageReceiveTimeout of the subscriber is supported.
+    Stop the tutorial_pubsub_publish example during operation to trigger the MessageReceiveTimeout and 
+    check the callback invocation here */
+static void PubsubStateChangeCallback(UA_NodeId *pubsubComponentId,
+                                      UA_PubSubState state,
```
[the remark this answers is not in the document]
MartinaFelberFEM: note: name of file has already been tutorial_pubsub_subscribe.c 
shall I still rename it?

include/open62541/server_pubsub.h
```diff
@@ -117,6 +117,15 @@ _UA_BEGIN_DECLS
  * **UA_ENABLE_PUBSUB_CUSTOM_PUBLISH_HANDLING**
  *  Enable the OPC UA PubSub support with custom callback implementation for Publisher and Subscriber. The option will provide flexibility to use the user defined callback mechanism for sending
  *  packets in Publisher and receiving packets in the Subscriber. Disabled by default.
+ * **UA_ENABLE_PUBSUB_TIMEOUT_HANDLING**
+ *  Enable the experimental PubSub timeout handling. This feature provides the MessageReceiveTimeout check of a DataSetReader and can be extended to 
+ *  check other PubSub timeouts as well. It uses the internal server callback implementation. An application can check for PubSub state changes and timeouts 
+ *  by providing a pubsubStateChangeCallback() at the server configuration (UA_PubSubConfiguration).
+ *  Disabled by default.
```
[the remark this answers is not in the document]
MartinaFelberFEM: Well no
State changes can be set anytime/anywhere,  without the timeout implementation

examples/pubsub/server_pubsub_subscribe_monitoring_linux.c
```diff
@@ -0,0 +1,539 @@
...
+
+#ifdef UA_ENABLE_PUBSUB_ETH_UADP
+#include <open62541/plugin/pubsub_ethernet.h>
+#endif
+
+#include <stdio.h>
+#include <signal.h>
+#include <stdlib.h>
+
+#include <signal.h>
```
andreasebner, line 40: Double include of signal.

examples/pubsub/server_pubsub_subscribe_monitoring_linux.c
```diff
@@ -0,0 +1,539 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+/**
+ * IMPORTANT ANNOUNCEMENT
+ * The PubSub Subscriber API is currently not finished. This example can be used
+ * to receive and display values that are published by tutorial_pubsub_publish
+ * example in the TargetVariables of Subscriber Information Model.
```
andreasebner, line 8: Which parts of the internal API are used here? I think there is no reason for 'The PubSub Subscriber API is currently not finished.'. This hint is only for the subscribe-standalone example where the complete message flow is skipped by calling the receive function on the user-level.

examples/pubsub/server_pubsub_subscribe_monitoring_linux.c
```diff
@@ -0,0 +1,539 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
```
andreasebner, line 2: My proposal: Make a subfolder in the pubsub example folder called 'monitoring'. Then move this file there and call it ..._custom_monitoring instead of monitoring_linux.

include/open62541/server.h
```diff
@@ -67,6 +67,11 @@ _UA_BEGIN_DECLS
  *
  * The :ref:`tutorials` provide a good starting point for this. */
 
+#ifdef UA_ENABLE_PUBSUB_MONITORING
+struct UA_PubSubConfiguration;
+typedef struct UA_PubSubConfiguration UA_PubSubConfiguration;
+#endif /* UA_ENABLE_PUBSUB_MONITORING */
+
```
andreasebner, line 74: The PubSubConfiguration is in the long term not monitoring dependent. I would prefer to always include this type and make just a conditional include around the contained 'UA_PubSubMonitoringInterface monitoringInterface;'.

include/open62541/server_pubsub.h
```diff
@@ -179,6 +192,42 @@ struct UA_PubSubConnectionConfig {
 #endif
 };
 
+#ifdef UA_ENABLE_PUBSUB_MONITORING
+
+typedef enum {
+    eMessageReceiveTimeout
```
andreasebner, line 198: Please rename the enum entry to a unambiguously name and follow the conventions for enums.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1449,17 +1536,75 @@ UA_Server_DataSetReader_process(UA_Server *server, UA_DataSetReader *dataSetRead
             }
         }
     }
+
+    UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "DataSetReader '%.*s': received a network message", 
+        (int) dataSetReader->config.name.length, dataSetReader->config.name.data);
+#ifdef UA_ENABLE_PUBSUB_MONITORING
+    /* if previous reader state was error (because we haven't received messages and ran into timeout) we should set the state back to operational */
+    if (dataSetReader->state != UA_PUBSUBSTATE_OPERATIONAL) {
+        UA_DataSetReader_setPubSubState(server, UA_PUBSUBSTATE_OPERATIONAL, dataSetReader);
```
andreasebner, line 1545: Why is the DSR set to OPERATIONAL implicitly?

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1449,17 +1536,75 @@ UA_Server_DataSetReader_process(UA_Server *server, UA_DataSetReader *dataSetRead
...
+    /* start message receive timeout timer */
+    if (server->config.pubsubConfiguration->monitoringInterface.startMonitoring(server, dataSetReader->identifier, 
+            UA_PUBSUB_COMPONENT_DATASETREADER, eMessageReceiveTimeout, dataSetReader) == UA_STATUSCODE_GOOD) {
+        UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "Info: DataSetReader '%.*s': start receive timeout timer",
+            (int) dataSetReader->config.name.length, dataSetReader->config.name.data);
+        dataSetReader->msgRcvTimeoutTimerRunning = UA_TRUE;
+    } else {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                        "Starting Message Receive Timeout timer failed.");
+        UA_DataSetReader_setPubSubState(server, UA_PUBSUBSTATE_ERROR, dataSetReader);
```
andreasebner, line 1567: Just for interest: Did you make some profiling? This code path is called very often.

src/server/ua_server_config.c
```diff
@@ -78,6 +78,13 @@ UA_ServerConfig_clean(UA_ServerConfig *config) {
         config->logger.clear(config->logger.context);
     config->logger.log = NULL;
     config->logger.clear = NULL;
+
+#ifdef UA_ENABLE_PUBSUB_MONITORING
+    if (config->pubsubConfiguration != 0) {
```
andreasebner, line 83: 0 vs. NULL

andreasebner (commented):
Thanks for your great work on this feature. I added some comments and requests for minor changes. From my point of view this PR can be merged afterwards. Since there are some minor changes to a couple core files, @jpfr can you also give your approval to 
the merge?

examples/pubsub/server_pubsub_subscribe_monitoring_linux.c
MartinaFelberFEM: thx, removed

examples/pubsub/server_pubsub_subscribe_monitoring_linux.c
MartinaFelberFEM: ok, removed the sentence

include/open62541/server_pubsub.h
MartinaFelberFEM: sorry for the wrong enum convention
called  it 'UA_PUBSUB_MONITORING_MESSAGE_RECEIVE_TIMEOUT'

src/pubsub/ua_pubsub_reader.c
MartinaFelberFEM: Good question ...
The specification only states the following:
"the DataSetReader State shall be changed to Error_3 until the next DataSetMessage is received. "
https://reference.opcfoundation.org/v104/Core/docs/Part14/6.2.8/#6.2.8.6

So I assumed that after receiving a message again the state should be set back to operational. 

According to the state machine following state changes could happen:
https://reference.opcfoundation.org/v104/Core/docs/Part14/6.2.1/

ERROR - OPERATIONAL -> Error resolved (I thought this was the use case)
ERROR - DISABLED -> I thought that this was an action from the application (manual operation)

I think ERROR - PAUSED would not apply, because it's only used if a parent changes to paused or disabled ...


I changed it to this:
```C
    /* if previous reader state was error (because we haven't received messages and ran into timeout) we should set the state back to operational */
    if (dataSetReader->state == UA_PUBSUBSTATE_ERROR) {
        UA_DataSetReader_setPubSubState(server, UA_PUBSUBSTATE_OPERATIONAL, dataSetReader);
        if (server->config.pubsubConfiguration->pubsubStateChangeCallback != 0) {
            server->config.pubsubConfiguration->pubsubStateChangeCallback(dataSetReader->identifier, UA_PUBSUBSTATE_OPERATIONAL, UA_STATUSCODE_GOOD);
        }
    }
```

But maybe I have misunderstood your question.
If you like we could talk about this tomorrow.

src/pubsub/ua_pubsub_manager.c
```diff
@@ -328,4 +333,243 @@ UA_PubSubManager_removeRepeatedPubSubCallback(UA_Server *server, UA_UInt64 callb
...
+    monitoringInterface->startMonitoring = UA_PubSubComponent_startMonitoring;
+    monitoringInterface->stopMonitoring = UA_PubSubComponent_stopMonitoring;
+    monitoringInterface->updateMonitoringInterval = UA_PubSubComponent_updateMonitoringInterval;
+    monitoringInterface->deleteMonitoring = UA_PubSubComponent_deleteMonitoring;
+    return UA_STATUSCODE_GOOD;
+}
+
+UA_StatusCode
+UA_Server_setPubSubMonitoringCallbacks(UA_Server *server, const UA_PubSubMonitoringInterface * const monitoringInterface) {
+    if ((server == 0) || (monitoringInterface == 0)) {
```
andreasebner, line 567: Please use NULL for all pointer comparisons. There are several compares with 0.

include/open62541/server_pubsub.h
```diff
@@ -714,6 +766,11 @@ UA_Server_setReaderGroupOperational(UA_Server *server, const UA_NodeId readerGro
 UA_StatusCode UA_EXPORT
 UA_Server_setReaderGroupDisabled(UA_Server *server, const UA_NodeId readerGroupId);
 
+#ifdef UA_ENABLE_PUBSUB_MONITORING
+UA_StatusCode UA_EXPORT
+UA_Server_setPubSubMonitoringCallbacks(UA_Server *server, const UA_PubSubMonitoringInterface * const monitoringInterface);
+#endif /* UA_ENABLE_PUBSUB_MONITORING */
```
andreasebner, line 772: Not sure if this API function is needed. The user has access to server config and the function only sets the new PTR.

src/pubsub/ua_pubsub_reader.c
andreasebner: Yes, thanks for your references to the standard.

include/open62541/server_pubsub.h
MartinaFelberFEM: yes you are right. I have removed the function

PR #3617, opened 2020-05-20
Add example of how to manipulate timestamp on a write operation

kjeldflarup:

### DIFF
--- a/examples/server_settimestamp.c
+++ b/examples/server_settimestamp.c
@@ -1,0 +1,94 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+/**
+ * Faking source timestamp
+ * ----------------------------
+ *
+ * This example shows how to write data with a modified source timestamp
+ * When showing the data in UAExpert, the source timestamp will always be half an hour behind the server timestamp
+ */
+
+#include <open62541/plugin/log_stdout.h>
+#include <open62541/server.h>
+#include <open62541/server_config_default.h>
+
+#include <signal.h>
+#include <stdlib.h>
+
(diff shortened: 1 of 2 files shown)


examples/server_settimestamp.c
```diff
@@ -0,0 +1,94 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+/**
+ * Faking source timestamp
+ * ----------------------------
+ *
+ * This example shows how to write data with a modified source timestamp
+ * When showing the data in UAExpert, the source timestamp will always be half an hour being the server timestamp
```
ccvca, line 9: Spelling error?
"be half an hour **being** the server timestamp"
being -> behind

ccvca:
What do you think about setting also a custom StatusCode, so the example shows all features of setting a DataValue instead of a Variant?

examples/server_settimestamp.c
kjeldflarup: fixed.

kjeldflarup:
@ccvca then the example should be called server_writedatavalue.c  
Would'nt that however be very alike this example https://github.com/open62541/open62541/blob/791eb95986709e0fdaeff6d64ea5b8f994198b55/examples/tutorial_server_variable.c#L98

ccvca:
@kjeldflarup So it's not required here. (I didn't know tutorial_server_variable in detail before)

kjeldflarup:
@ccvca A complete example is still a good idea. Like a tutorial.  
But I just had a need to make a proff of concept on the timestamp, and thought that I would share it. It is open for anybody to make the example "complete".

jpfr:
Looking good!
Thanks for the contribution.

PR #3622, opened 2020-05-22
fix(debian package) Enable encryption for the official debian packages

andreasebner:
Currently the official debian packages (https://launchpad.net/~open62541-team/+archive/ubuntu/daily) do not include encryption. This PR adjust the dependecies to build these packages with encryption support. The missing encryption leads to errors like this:

`/usr/bin/c++  -O3 -DNDEBUG   CMakeFiles/testOPCUA.dir/simpleClient.C.o  -o testOPCUA -lopen62541 
CMakeFiles/testOPCUA.dir/simpleClient.C.o: In function `setupClient(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
simpleClient.C:(.text+0x4d8): undefined reference to `UA_ClientConfig_setDefaultEncryption'`

### DIFF
--- a/debian/control-template
+++ b/debian/control-template
@@ -10,7 +10,9 @@ Build-Depends: debhelper (>= 9),
  texlive-fonts-recommended,
  texlive-latex-extra,
  texlive-plain-generic | texlive-generic-extra,
- latexmk
+ latexmk,
+ libmbedtls-dev
+
 Standards-Version: 4.4.1
 Section: libs
 Homepage: https://open62541.org/
@@ -50,7 +52,7 @@ Description: Open source implementation of OPC UA - tools
 Package: libopen62541-<soname>
 Architecture: any
 Multi-Arch: same
-Depends: ${shlibs:Depends}, ${misc:Depends}
+Depends: ${shlibs:Depends}, ${misc:Depends}, libmbedtls10 | libmbedtls12
 Description: Open source implementation of OPC UA - shared library
  open62541 (http://open62541.org) is an open source and free implementation
  of OPC UA (OPC Unified Architecture) written in the common subset of the
--- a/debian/rules-template
+++ b/debian/rules-template
@@ -5,7 +5,7 @@ export DH_VERBOSE = 1
 export DEB_BUILD_MAINT_OPTIONS = hardening=+all,-pie
 
 override_dh_auto_configure:
-	dh_auto_configure -- -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_AMALGAMATION=OFF -DUA_PACK_DEBIAN=ON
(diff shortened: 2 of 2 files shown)


tools/travis/travis_linux_script.sh
```diff
@@ -532,15 +532,15 @@ if [ "$CC" != "tcc" ]; then
...
+		if [ "${TRAVIS_PULL_REQUEST}" = "true" ]; then
 			REAL_BRANCH="${TRAVIS_BRANCH}"
        		echo -en "== Checking branch for packing: BRANCH_FOR_TAG='$BRANCH_FOR_TAG' and TRAVIS_TAG='${TRAVIS_TAG}'. \n"
 			if [ "${TRAVIS_TAG}" = "${TRAVIS_BRANCH}" ]; then
 				REAL_BRANCH="$BRANCH_FOR_TAG"
         		echo -en "== Commit is tag build for '${TRAVIS_TAG}'. Detected branch for tag = '$BRANCH_FOR_TAG' \n"
 			fi
 
-			if [ "${REAL_BRANCH}" = "master" ] || [ "${REAL_BRANCH}" = "1.0" ]; then
+			if [ 1 ]; then
```
jpfr, line 543: We want to create a pack_ branch for every feature branch as well?

debian/control-template
```diff
@@ -11,6 +11,8 @@ Build-Depends: debhelper (>= 9),
  texlive-latex-extra,
  texlive-plain-generic | texlive-generic-extra,
  latexmk
+ libmbedtls-dev
```
jpfr, line 14: This is the build-dependency.
Would we need a runtime dependency as well?

andreasebner: I think we don't need to link the mbedtls explicitly. All targets are depending on 'libopen62541-<soname>' and this package in turn depends on '${shlibs:Depends}'. For my understanding, the runtime dependency should therefore be resolved automatically.

jrcoding: FYI: is there a "comma" missing after latexmk ?

andreasebner: You're right, thanks for the hint. ;)

jrcoding: BTW: I agree with @jpfr that the binary section needs a dependency then.

IMHO the control section of the lib should look like
```
...
Package: libopen62541-<soname>
Architecture: any
Multi-Arch: same
Depends: ${shlibs:Depends}, ${misc:Depends}, libmbedtls10 | libmbedtls12
...
``` 
Without the binary dependency installation of the library will fail!

FYI `libmbedtls10 | libmbedtls12` is important to reflect the different names of the dependency package dependent on the Ubuntu/Debian version

andreasebner: I have now update the PR according to your proposal @jrcoding. Thank's for your help :thumbsup:

PR #3623, opened 2020-05-22
[REVIEW] Feature: decrypt in securechannel

mlgiraud:
Continuation of #3488


mlgiraud:
@jpfr Sorry this is taking so long, but i can only work on this max. 5 hours each week.
I updated the control flow to directly decrypt during extraction, so we now only have the chunk queue and no additional state in the channel.
If all tests pass and you agree with the changes, feel free to merge.

jpfr:
Sure. Will have a look.

mlgiraud:
@jpfr As discusssed i implemented the change to have linear runtime. Can you take a look and merge if you are satisfied?

PR #3630, opened 2020-05-25
test: fix signedness compile errors

bluhm:
My clang 8.0.1 on OpenBSD could not compile the test programs due
to implicit conversion changes signedness errors.  Use the correct
unsigned ck_assert_uint_eq() function and change some data types.

### DIFF
--- a/tests/check_chunking.c
+++ b/tests/check_chunking.c
@@ -51,10 +51,10 @@ START_TEST(encodeArrayIntoFiveChunksShallWork) {
                                            &pos, &end, sendChunkMockUp, NULL);
 
     ck_assert_uint_eq(retval,UA_STATUSCODE_GOOD);
-    ck_assert_int_eq(counter,4); //5 chunks allocated - callback called 4 times
+    ck_assert_uint_eq(counter,4); //5 chunks allocated - callback called 4 times
 
     dataCount += (uintptr_t)(pos - buffers[bufIndex].data);
-    ck_assert_int_eq(UA_calcSizeBinary(&v,&UA_TYPES[UA_TYPES_VARIANT]), dataCount);
+    ck_assert_uint_eq(UA_calcSizeBinary(&v,&UA_TYPES[UA_TYPES_VARIANT]), dataCount);
 
     UA_Variant_deleteMembers(&v);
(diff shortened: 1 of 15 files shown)


jpfr:
Looks good!
Can you enable the signedness-test also for gcc and on our linux CI builds?
So we can prevent regressions in the future.

I will nudge the CI from time to time to shake out internal network problems.

bluhm:
> Can you enable the signedness-test also for gcc and on our linux CI builds?
> So we can prevent regressions in the future.

Turns out, that there were already workarounds for the CLang sign-conversion warnings.  But they were explicitly disabled on OpenBSD.  As this commit fixes the fatal warnings, we can remove the workarounds and enable the warnings everywhere.  I have created pull request #3638 for that.

I think it is easier if my test fix is merged before the compile flag change.  Then I can see how the CI deals with the new warnings and the fixed code.  Although I am not sure how to debug CI failures properly.  On my local OpenBSD system I had to dig into CMake quite deeply to see any output.

jpfr:
Rebased on master to help the CI.

jpfr:
Doesn't build on VS2017 with this error:
```
D:\a\1\s\tests\check_types_builtin.c(627): error C2220: warning treated as error - no 'object' file generated [D:\a\1\s\build\tests\check_types_builtin.vcxproj]
D:\a\1\s\tests\check_types_builtin.c(627): warning C4146: unary minus operator applied to unsigned type, result still unsigned [D:\a\1\s\build\tests\check_types_builtin.vcxproj]
D:\a\1\s\tests\check_types_builtin.c(709): warning C4146: unary minus operator applied to unsigned type, result still unsigned [D:\a\1\s\build\tests\check_types_builtin.vcxproj]
```

bluhm:
I have pushed a fixed commit.  Note that in general compiling with -Werror is a bad idea if you want to support multiple compilers.  You can invest a lot of time shuffling the code around to make all compilers happy.  But you gain little code quality.

jpfr:
> Note that in general compiling with -Werror is a bad idea if you want to support multiple compilers.

Yes, this is effort. But not excessively so.
If we let compiler warnings slip, code quality will suffer.
For example, clang-analyzer warnings are also "compiler warnings".

jpfr:
Merged. Thanks!

PR #3632, opened 2020-05-26
Add 'Aes128Sha256RsaOaep' security policy

johndzilva:
- Aes128Sha256RsaOaep security policy tested against UaExpert, OPCF reference client and UA browser.
- Unit test added for the same
- Symmetric encryption algorithm Uri is corrected in 'Basic256Sha256' policy


jpfr:
Looks great!
I triggered the CI. Some tests failed. But that was Azure-internal and from what I see not related to the contribution.

jpfr:
Rebased and history-squashed.

PR #3633, opened 2020-05-26
build(ex): Don't build pubsub subscriber examples with BUILD_SHARED_LIBS

MartinaFelberFEM:
Functions `UA_PubSubConnection_regist` and `UA_ReaderGroup_findDSRbyId` are used at examples/pubsub/server_pubsub_subscriber_rt_level.c but not exported. So compilation with shared library fails.

Shall these function be moved to server_pubsub.h?


jpfr:
Looks good.
Our CI seems to have connectivity problems. Can be merged when that is resolved.

andreasebner:
One of the core design concepts behind the PubSub API was the avoidance of give the user a pointer to the internal PubSub entities. If the user has direct access to the internal structures our concepts like configuration freeze become useless. Therefore, we splited all PubSub into an internal and config part. The user works only with the unique ID of the PubSub entities. All configurations are made with the config structure (which is copied into the internal object). All ‘findById’ methods are made for internal use and should not be exposed to the user. If this is needed for any case we should extend the config structure or discuss why the direct access is needed.

MartinaFelberFEM:
Thanks for your review!

The `UA_PubSubConnection_regist` function is used in the tutorials/examples. 
regist() finishes the network layer configuration (UDP, etc.)
I don't know why the network layer initialization is split up into _open and _regist calls.

Maybe the _regist() call can be moved to UA_Server_addPubSubConnection() ? 
So the application does not have to call 2 methods - addPubSubConnection and regist.

And there's no UA_PubSubConnection_unregist() function anyways.
I don't know why _addPubSubConnection does not handle the regist() call at the moment.


I need the 2nd function `UA_PublishedDataSet_findPDSbyId` in my application, but I'll think about a workaround.

`UA_ReaderGroup_findDSRbyId` is used at examples/pubsub/server_pubsub_subscrier_rt_level.c. There I get a compile error if I build with shared libs. But I have forgotten it in my PullRequest ...

jpfr:
@andreasebner is right and I was blind 👍 
We should not export internal API.

MartinaFelberFEM:
Regarding `UA_PubSubConnection_regist`: 
Shall UA_Server_addPubSubConnection call this method?

Or shall I rename it to UA_Server_registerPubSubConnection() or something? And move the definition to the public header /include/open62541/server_pubsub.h ?

andreasebner:
The register, send, receive... methods are internal and should in the long term never be called by the user. I'am not 100% sure about the subscriber API status. Maybe the call of the regist function is currently needed by the user (as an preview). If this is the case, we should soon start fixing it to allow an complete pubslish and subscribe configuration by the highlevel API. For example the regist function should only be called for an valid subscriber configuration after enabling it.

MartinaFelberFEM:
I have tried to refactor it, it's just a proposal.
But now I get more undefined references warnings in examples.

_/examples/pubsub/pubsub_subscribe_standalone.c:63: undefined reference to `UA_NetworkMessage_decodeBinary'_

_/examples/pubsub/pubsub_subscribe_standalone.c:105: undefined reference to `UA_NetworkMessage_deleteMembers'_

I think I'll abandon these because the API is not finished and I'll do more harm than good.

I think I only need the UA_PubSubConnection_regist() method. Either exported or implicitly called at UA_Server_addPubSubConnection() to enable the shared library build.

andreasebner:
Thank's for your feedback ;) You have used the pubsub_subscriber_standalone example which has the following comment inside: https://github.com/open62541/open62541/blob/50f8074e654f8223334512f5a66e62854191a3eb/examples/pubsub/pubsub_subscribe_standalone.c#L5-L11
This is because there are calls to internal methods which should later not be called by the user. These methods are also not exported and so the linker warning seems logical. But you are right, we should prevent this error on cmake level by removing it if shared libraries are enabled.

MartinaFelberFEM:
Thanks for your feedback!

I will remove my commits and see if I can prevent the PubSub examples build if shared library build is enabled.

jpfr:
Rebased to trigger the CI.

Good to merge @andreasebner ?

PR #3621, opened 2020-05-22
Add username/password support for MQTT

basyskom-jvoe:

PR #3637, opened 2020-05-27
[Review] Add experimental TLS support for PubSub via MQTT

basyskom-jvoe:
The MQTT plugin is extended with the new connection options mqttUseTLS, mqttCaFilePath, mqttCaPath, mqttClientCertPath and mqttClientKeyPath.

There is only an OpenSSL based implementation for now.

The new feature can be enabled by setting the CMake option UA_ENABLE_MQTT_TLS to ON.


basyskom-jvoe:
The first three commits are from PR #3621 and will be removed in a rebase once that PR has been merged.

basyskom-jvoe:
I have extended the MQTT publish example with TLS. Please feel free to review.

jpfr:
Thanks for this. 😄 
Can you set the PR title to [Review] when you are happy and the CI is green?

andreasebner:
Thanks for this PubSub extension PR! Your work is really cool and fits the section 7.3.5.1 of the PubSub standard: 
"Security with MQTT is primary provided by a TLS connection between the Publisher or Subscriber and the MQTT server, however, this requires that the MQTT server be trusted. For that reason, it may be necessary to provide end-to-end security."

Nevertheless, I think the PR title and the used flags are maybe not fitting your work. For my understanding this PR introduces MQTT End-To-End security but not MQTT TLS. If my understanding is right, we should adjust the flags and description.

basyskom-jvoe:
@andreasebner What this PR is doing is setting up a [TLS](https://tools.ietf.org/html/rfc5246) connection via OpenSSL which is then passed into the 3rd party MQTT client. What title would you suggest?

basyskom-jvoe:
The Docker Cloud check doesn't seem to pass. I always get a 404 error page when I'm trying to check what is wrong.

basyskom-meerkoetter:
> The Docker Cloud check doesn't seem to pass. I always get a 404 error page when I'm trying to check what is wrong.

@jpfr @andreasebner any idea about this? We would need some help to progress here.

basyskom-jvoe:
This PR has been updated to the current master

Fischer-Thomas:
Are there any updates on this PR?
I tested this PR on my end and it works pretty fine for me.

Please merge.

jpfr:
@andreasebner Good to go?

basyskom-meerkoetter:
it would be good to get this into 1.3.

jpfr:
Merged. Thanks for this cool feature!
Do you see a possibility to have an automated test in the CI pipeline?

basyskom-jvoe:
> Merged. Thanks for this cool feature!
> Do you see a possibility to have an automated test in the CI pipeline?

Thanks for the merge!
Would it be possible to run mosquitto or something like that in the CI?
There is also a public mosquitto instance with SSL support on the internet (https://test.mosquitto.org/), but I don't really like the idea of adding an external dependency outside of our control.

jackybek:
I tested using OpenSSL 1.1.1K but received an error : ERROR_MqTt_error during runtime.
 I suspect it is a bug in OpenSSL. 
 
I already verified that the certs, ca and keys are working when I test using mosquitto pub and sub on command prompt using port 8883.

PR #3638, opened 2020-05-27
test: build with sign conversion warning

bluhm:
CLang warns about implicit sign conversion by default.  When building
the tests, it may fail with a hard error.  Remove workaround for
Apple and non-working workaround for OpenBSD.  After the conversions
in the test files have been fixed, enable the warning for all
platforms.  Then the CI should catch regressions.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -11,13 +11,6 @@ if(NOT MSVC AND UA_ENABLE_UNIT_TESTS_MEMCHECK)
     find_package(Valgrind REQUIRED)
 endif()
 
-if(APPLE)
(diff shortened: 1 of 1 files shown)


jpfr:
Rebased on master after merging of the sign fixes.

jpfr:
The line `add_definitions(-Wsign-conversion)` fails for the Visual Studio compilers.

Can you use `check_add_cc_flag` instead?
This is a macro defined in the main CMakeLists.txt and should be usable here as well.

jpfr:
Further simplified. -Wsign-conversion is implied in many warning levels.
We don't have to spell that out explicitly.

PR #3656, opened 2020-06-02
check the return value of UA_malloc 

wenshan1:
Check the return value of UA_malloc in the function UA_PubSubChannelEthernet_send.


jpfr:
Good one. Can you make the PR against 1.1?

wenshan1:
OK, no problem.

> Good one. Can you make the PR against 1.1?

OK.

jpfr:
Was merged on 1.1.
We will rebase it on master from there.

Thanks for providing the fix. 👍

PR #3660, opened 2020-06-02
feat(server): set admin session context

bluhm:
In callbacks the userland needs its context.  The node lifecycle
callbacks often use the server's admin session.  Unfortunatey the
API does not provide a possiblility to set the handle of the admin
session, so the context in the lifecycle callbacks is always NULL.
As the admin session is not part of the config it cannot be set
like the callbacks.  Provide an additional server method
UA_Server_setGlobalNodeLifecycle() that sets both, the callbacks
and the context.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -736,7 +736,11 @@ typedef struct {
                                          const UA_NodeId *targetParentNodeId,
                                          const UA_NodeId *referenceTypeId,
                                          UA_NodeId *targetNodeId);
-    } UA_GlobalNodeLifecycle;
+} UA_GlobalNodeLifecycle;
+
+void UA_EXPORT
+UA_Server_setAdminSessionContext(UA_Server *server,
+                                 void *context);
 
 typedef struct {
     /* Can be NULL. May replace the nodeContext */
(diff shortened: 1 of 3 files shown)


jpfr:
Setting the session context for the adminsession is a good idea.
But please don't combine it with the node lifecycle handling.

The node lifecycle handling is set in the server config.
Accessing the entire server config is already public API and we don't need an extra function for it.

The admin session is the only with a GUID identifier as follows:

    server->adminSession.sessionId.identifierType = UA_NODEIDTYPE_GUID;
    server->adminSession.sessionId.identifier.guid.data1 = 1;

So it can be disambiguated in the callbacks already.
This needs better documentation.

bluhm:
> But please don't combine it with the node lifecycle handling.

Done.  Is the function name ok?

> So it can be disambiguated in the callbacks already.

This would only solve half of the problem.  The user would still need a global variable to map the special guid to the needed context.  Providing a void pointer with all necessary user information is easier to deal with.

jpfr:
> Providing a void pointer with all necessary user information is easier to deal with.

👍

PR #3662, opened 2020-06-02
build: compile with debug on OpenBSD using CLang

bluhm:
OpenBSD with CLang 8.0.1 does not support all sanitizer flags.  Make
them optional by using check_add_cc_flag().  It is not necessary
to put them in CXX flags.  CLang with -fsanitize=undefined requires
to link with c++ instead of cc or ld.  Some other error messages
from libcompiler_rt.a prevent linking on OpenBSD, so disable
-fsanitize=undefined on OpenBSD for now.


jpfr:
Would like to merge this.
Can you rebase?

andreasebner:
This PR looks still valid and ready to merge from an functional point of view. If youre intressted in contributing this feature, it would be nice if you solve the conflict and rebase.

jpfr:
@bluhm is this still relevant?

PR #3666, opened 2020-06-04
fix(ci): Fix ppa build error. Regression from #3622

Pro:
There mustn't be a newline in the first block.

Otherwise the PPA build fails with:

```
dpkg-source: error: syntax error in recipe/debian/control at line 21: block lacks the 'Package' field
dpkg-buildpackage: error: dpkg-source -i -I --before-build recipe gave error exit status 25
```

/cc @andreasebner


andreasebner:
Thank you for the quick fix ;)

PR #3671, opened 2020-06-04
Add UA_Server_deleteCondition 

ccvca:
This PR adds UA_Server_deleteCondition to delete a condition with it's internals from the address space.


jpfr:
Thanks. Looks good.

The two nested `foreach` for the condition and conditionSource looks like a very common pattern.
Is it possible to have reusable internal method for the lookup?

But I would only do it if it leads to an actual reduction in lines-of-code.

ccvca:
> But I would only do it if it leads to an actual reduction in lines-of-code.

After a short look at this, I cannot guarantee, that there will be less loops in the end, as these for-loops are a little bit different on several places.
But as I might do addintional Work on this (https://github.com/open62541/open62541/issues/3328#issuecomment-631333190) I'll have a look for such simplifications.

PR #3675, opened 2020-06-05
Replace unnecessary UA_NS0ID_HASORDEREDCOMPONENT with UA_NS0ID_HASCOMPONENT

ccvca:
There is no reason for using (non-implemented) HasOrderedComponent references for these methods.
Looks like a persistent copy-past error :)


jpfr:
Rebased for merging to 1.1.

ccvca:
This will get into master via #3681?

PR #3678, opened 2020-06-05
fix(core): fix QualifiedName encoding at chunk end

genuaboro:
If the UInt16 namespaceIndex of a QualifiedName did not fit into a
chunk, it was omitted in the next chunk because the return code was not
checked and the string / array encoding would overwrite the last known
good position.

### DIFF
--- a/src/ua_types_encoding_binary.c
+++ b/src/ua_types_encoding_binary.c
@@ -41,12 +41,12 @@
 #define UA_ENCODING_MAX_RECURSION 100
 
 typedef struct {
-    /* Pointers to the current position and the last position in the buffer */
+    /* Pointers to the current and last buffer position */
     u8 *pos;
     const u8 *end;
 
-    u8 **oldpos; /* Sentinel for a lower stacktrace exchanging the buffer */
-    u16 depth;   /* How often did we en-/decoding recurse? */
+    /* How often did we en-/decoding recurse? */
+    u16 depth;
 
     const UA_DataTypeArray *customTypes;
     UA_exchangeEncodeBuffer exchangeBufferCallback;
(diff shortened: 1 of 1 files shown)


jpfr:
The return code is **not ignored**.
From my perspective, the return code is or-combined with the next encoding step.
If either of them fails, then the entire QualifiedName is put into the next chunk.

I can be convinced otherwise. By a minimal example that showcases a bug here.
Such an example should be easy to do by using a very small buffer.

genuaboro:
> The return code is **not ignored**.
> From my perspective, the return code is or-combined with the next encoding step.
> If either of them fails, then the entire QualifiedName is put into the next chunk.

Maybe "ignored" was the wrong word. I was referring to the encoding of other types, where we return directly before encoding strings or arrays (e. g. before encoding a NodeId string identifier or the locale string in a LocalizedText). But after looking at the code again I'm not sure anymore why the namespaceIndex will not be written in the next chunk. I think I'm still missing a piece to understand the encoding. Maybe the return is necessary for a special corner case that I am hitting.

> I can be convinced otherwise. By a minimal example that showcases a bug here.
> Such an example should be easy to do by using a very small buffer.

I am running into this issue when the server sends large browse responses that get chunked. Sometimes the client can't decode them and it seems to happen when the QualifiedName is at the end of a chunk and the server does not decode/send the namespaceIndex.

I'll try to make a test case for it. I assumed too much about the underlying problem because the diff fixed my issue and it seems to be consistent with similar code segments.

jpfr:
The secret lies in `encodeWithExchangeBuffer`.
If encoding fails because we reach the end of the buffer, then we reset to the previous buffer position, send out the chunk and retry for the entire datatype.

jpfr:
You were correct. That was a bug.
Once inside a `encodeWithExchangeBuffer` context, we must not exchange the buffer **and** return UA_STATUSCODE_BADENCODINGLIMITSEXCEEDED to `encodeWithExchangeBuffer`.

Went over the encoding with a fine-grained comb to ensure this doesn't happen anywhere else.
I added a couple more assertions to suss out any such problems with oss-fuzz in the future.

Some more drive-by simplifications and beautification on top.
Rebased for the 1.1 branch.

genuaboro:
> You were correct. That was a bug.
> Once inside a `encodeWithExchangeBuffer` context, we must not exchange the buffer **and** return UA_STATUSCODE_BADENCODINGLIMITSEXCEEDED back to `encodeWithExchangeBuffer`.
> 
> Went over the encoding with a fine-grained comb to ensure this doesn't happen anywhere else.
> I added a couple more assertions to suss out such problems with oss-fuzz and clang-analyzer in the future.
> 
> Some more drive-by simplifications and beautification on top.
> Rebased for the 1.1 branch.

Ok. Thanks for looking into it.

PR #3679, opened 2020-06-07
feat(plugin): Add a syslog logger plugin

jpfr:


bluhm:
I do not think that a library should call openlog(3).  It is the program's job to do this once at startup.  There it can select its ident and facility.  Consider two libraries used by one program each calling openlog(3).  That does not make sense.  I would remove the call from UA_Log_Syslog_withLevel() and write in the documentation that openlog(3) must be called before.

jpfr:
Sounds good.

PR #3685, opened 2020-06-09
fix(core): Correct decrypt of security token on big endian

Pro:
Got this issue report from a contact.

```
memcpy(&tokenSecretLength, decryptedTokenSecret.data, sizeof(UA_UInt32));
```
data contains the length in little endian. On big endian systems simple memcpy doesn't work.

`UA_UInt32_decodeBinary` correctly handles endianess depending on the `UA_BINARY_OVERLAYABLE_INTEGER` value.

Will be ported to master in the next 1.1 to master merge


jpfr:
Nice!

PR #3690, opened 2020-06-10
feat(server:) Use config customHostname to bind server address

bluhm:
It should be possible to bind to a specific address instead of
always using a * socket for TCP listen.  Reuse the customHostname
of the server config.  That means that the hostname in the server
URL must be an IP address or resolveable by DNS.  The client also
puts the hostname of the URL into getaddrinfo(3) so this should not
be a problem.  Behavior changes if the customHostname was used to
create a discoveryUrl before.  Then it will be used to additionally
bind the address now.

I need this feature to build a proxy firewall that has several interface addresses. The OPC UA server should only listen on one of these.  Combining the bind address with the FQDN or IP addess in the URL seems reasonable to me.  If you disagree I could also try to extend the config and pass a second hostname down to the gethostbyname(3) in ServerNetworkLayerTCP_start().

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -355,11 +355,18 @@ addServerSocket(ServerNetworkLayerTCP *layer, struct addrinfo *ai) {
 
 static UA_StatusCode
 ServerNetworkLayerTCP_start(UA_ServerNetworkLayer *nl, const UA_String *customHostname) {
-  UA_initialize_architecture_network();
+    UA_initialize_architecture_network();
 
     ServerNetworkLayerTCP *layer = (ServerNetworkLayerTCP *)nl->handle;
 
     /* Get addrinfo of the server and create server sockets */
+    char hostname[512];
+    if(customHostname->length) {
+        if(customHostname->length >= sizeof(hostname))
+            return UA_STATUSCODE_BADOUTOFMEMORY;
+        memcpy(hostname, customHostname->data, customHostname->length);
+        hostname[customHostname->length] = '\0';
+    }
     char portno[6];
     UA_snprintf(portno, 6, "%d", layer->port);
     struct addrinfo hints, *res;
@@ -368,7 +375,8 @@ ServerNetworkLayerTCP_start(UA_ServerNetworkLayer *nl, const UA_String *customHo
     hints.ai_socktype = SOCK_STREAM;
     hints.ai_flags = AI_PASSIVE;
     hints.ai_protocol = IPPROTO_TCP;
-    if(UA_getaddrinfo(NULL, portno, &hints, &res) != 0)
+    if(UA_getaddrinfo(customHostname->length ? hostname : NULL,
+        portno, &hints, &res) != 0)
         return UA_STATUSCODE_BADINTERNALERROR;
 
     /* There might be serveral addrinfos (for different network cards,
--- a/tests/server/check_server_userspace.c
+++ b/tests/server/check_server_userspace.c
@@ -133,7 +133,7 @@ START_TEST(Server_forEachChildNodeCall) {
 
 
 START_TEST(Server_set_customHostname) {
-    UA_String customHost = UA_STRING("fancy-host");
+    UA_String customHost = UA_STRING("localhost");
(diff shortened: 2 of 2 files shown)


jpfr:
Yes, this is an important one!

Travis gives the following warning:
```
/home/travis/build/open62541/open62541/arch/network_tcp.c:365:18: error: request for implicit conversion from ‘void *’ to ‘char *’ not permitted in C++ [-Werror=c++-compat]

         hostname = UA_malloc(customHostname->length + 1);
```

Maybe we don't need a malloc here and can just use a char[512] static array with a length check.

bluhm:
> Travis gives the following warning:
> /home/travis/build/open62541/open62541/arch/network_tcp.c:365:18: error: request for implicit conversion from ‘void *’ to ‘char *’ not permitted in C++ [-Werror=c++-compat]

I disagree with the compiler warning.  This is valid C code.  Reducing casts to a minimum is a good thing, only then the compiler will catch real errors.  C++ is a different language with different casts.  It has other rules, in C++ I would use new to respect the type system.  Why do you want to program in C with C++ style?

Another thing is that the CI runner often fail, usually in some Windows environment I do not understand.  So I ignore them, which is bad.

> Maybe we don't need a malloc here and can just use a char[512] static array with a length check.

Done that.  NI_MAXHOST is 256 on my system, so 512 should be sufficient for all use cases.  Let's see if it fixes the Travis run.

jpfr:
Nice. Merged.

The history why we also compile as C++ in the CI is the problem with MS Visual Studio C up to ca. 2013. Before many C99 feature where not supported in C-mode. By they way in C++-mode. Hence we recommended compiling as C++ on old Visual Studio versions.

At this time, there is nearly no drawback to this approach.
It is a bit inconvenient to cast malloc'ed values. But that's about it.

ccvca:
>Behavior changes if the customHostname was used to
create a discoveryUrl before. Then it will be used to additionally
bind the address now.

Is there a way to handle the case, that discovery hostname is not identical to the interface?
e.g. in my enviroment I execute an OPC UA Server in a Docker enviroment and expose Port 4840. So I set the customHostname to the the public dns address. Inside docker sind public address is not bindable.

PS: Sorry for writing after the PR was merged, but I just stumble accross this yesterday (when configuring security and certificates).

bluhm:
> Is there a way to handle the case, that discovery hostname is not identical to the interface?

Currently not.

> e.g. in my enviroment I execute an OPC UA Server in a Docker enviroment

Is there no DNS within the container or is the IP missing in the container?
Can you fix these problems?

Within the software we could make binding to the address allow to fail.  If bind to address+port fails, we bind again with port only.  But that is more complex and an error in the address configuration is silently ignored.  Usually I prefer to fail hard.

Another option would be to add an option whether the server should bind to the address or not.  But that means more knobs and the software would be harder to configure.  Also this is a library and the using programs would have to expose the option.

I have also avoided to pass two host names, one for discovery and one for binding as they are usually identical unless you have split DNS.  One hostname for both still looks reasonable to me.

I would prefer to fix it in the running environment.  If that is not possible, we need a good idea.

ccvca:
Docker spans a common NAT-Network, just as an example:
Server:
DNS: opcua.example.com -> A-Record: 192.168.128.1 (a "Cloud IP", which will be bind to an Public IP by the cloud provider)


Docker Container:
DNS: <container name, e.g. ocpua_server, no '.' allowed> -> Docker IP 172.16.0.5

Docker has it's own network and the Port 4840 is forwarded to the internal network (acts like a nat with Port forwarding).

I would prefer to be able to define the binding address(es) separately, as this situation might be quite common in NAT-Networks (e.g. expose the OPC UA-Server in an internal machine network via the Machine-NAT-Gateway). This could also provide support to bind to several specific interfaces (but not all available, e.g. IPv4 and IPv6 of the same interface).

bluhm:
> Docker spans a common NAT-Network, just as an example:

I see your problem now.  Perhaps we can bind to a *-socket if the specific address does not exist.  Could you test pull request #3725 ?

PR #3691, opened 2020-06-10
feat(client): Call callback for failed attribute read

bluhm:
Currently if reading an attribute asynchronously from the client
fails, the error is not reported.  Although it is logged, the calling
program is not informed as the callback is not called.  The program
may hang forever as the callbacks are used to make progress.  After
a call to __UA_Client_readAttribute_async() the callback should be
called in all cases.  So use NULL as value pointer in ValueAttributeRead()
to report failure.  That new behavior may break existing applications,
but those programs would not work properly anyway as they wait for
the callback forever.

An open issue with this approach is, that user land can still not see the cause of the failure.  Perhaps it would be better to pass the status code or the whole read response as an additional parameter.  But that would require a much bigger API change.

### DIFF
--- a/src/client/ua_client_highlevel.c
+++ b/src/client/ua_client_highlevel.c
@@ -796,11 +796,13 @@ void ValueAttributeRead(UA_Client *client, void *userdata,
         }
     }
 
-    /* Could not process, delete the callback anyway */
-    if(!done)
+    /* Could not process, run the callback anyway, but without value */
+    if(!done) {
         UA_LOG_INFO(&client->config.logger, UA_LOGCATEGORY_CLIENT,
                     "Cannot process the response to the async read "
                     "request %" PRIu32, requestId);
+        cc->userCallback(client, cc->userData, requestId, NULL);
+    }
 
     UA_free(cc->clientData);
     LIST_REMOVE(cc, pointers);


jpfr:
Hmm. I would prefer if we could forward a StatusCode instead of just NULL.
So the user can disambiguate a server-fault vs. a lost connection.

So `__UA_Client_readAttribute` can return the proper StatusCode.

bluhm:
> Hmm. I would prefer if we could forward a StatusCode instead of just NULL.
> So the user can disambiguate a server-fault vs. a lost connection.

Done.  I have implemented a new type of callback.

To report the cause of the problem, the status code should be part
of the callback.  The generic UA_ClientAsyncServiceCallback callback
has a sufficient API as it reports a result.  There we have the
response header service result with the necessary information.  The
read attribute callbacks report a specfic type and need an additional
status code parameter.  Introduce a new UA_ClientAsyncAttributeCallback
type for these results.  To make the distincion clear, the existing
UA_ClientAsyncServiceCallback type passes an UA_Response instead

bluhm:
Now the build fails on some platforms with -Werror=cast-function-type.  Before my change the cast was in the void pointer of the callback parameter, now it is in the function type.  Although this is legal C, it is not allowed by your language subset.  What should I do?  Refactor the callback interface to allow multiple callback types.  Use a union to trick the compiler not to issue a cast warning.  Drop the idea to report the status code in the callback.  Change the warning level of the compiler.  Some other suggestion?

bluhm:
I went back to my original approach where I just call the callback with value NULL.  If someone needs the error code, he can still use the generic API callback which gives the complete response.  There all information is provided.  Changing the high level callback created more compiler and casting problems that it solved.

andreasebner:
Thank you for the contribution and pointing out this problem! The client is overhauled since this PR and it will be closed therefore.

PR #3705, opened 2020-06-14
fix(client): Check client config is set before connecting

bluhm:
If the program did not set the client default config before calling
UA_Client_connect(), the program crashed due to a initConnectionFunc
NULL pointer.  This is a regression to the 1.0 behavior.  Better
log the problem in initConnect() and return internal error to give
the user a hint what went wrong.


jpfr:
Rebased on 1.1.
This is good to be included in a service-release for 1.1.

Issue #3364, opened 2020-01-14
Use open62541 with Encryption on Windows without MbedTLS on system path


ccvca:
## Description
After 441ac87bfac31c8d10e126d1e373e92b622dbde3 it's quite hard to build with encryption, when mbedtls is not installed on a system path (e.g. on windows), as the targets that are defined by FindMbedTLS are not exported. (mbedtls, mbedx509, mbedcrypto)

https://github.com/open62541/open62541/blob/d4285862e11db10e018c523ba60d79e97e8cd09d/tools/cmake/FindMbedTLS.cmake#L22

Using the open62541 library in an own project causes that cmake tries to link to the libraries by name, as the target is no longer known. This only works, if the location of libmbedtls.a (...) is on a well known library path.

## Background Information / Reproduction Steps
Build + Install open62541 with Encryption by cmake

Use find_package(open62541 REQUIRED) in an own project

## Possible Solutions
1. Export the targets from FindMbedTLS/ revert 441ac87bfac31c8d10e126d1e373e92b622dbde3 
    - Disadvantage: Absolute paths are bad e.g. when building the deb-Package
2. Update documentation that FindMbedTLS must be used additional to find_package(open62541 REQUIRED) when using encryption
    - Disadvantage: Hard to do, as FindMbedTLS  not on the default seach path
3. include FindMbedTLS in [open62541Config.cmake.in](https://github.com/open62541/open62541/blob/d4285862e11db10e018c523ba60d79e97e8cd09d/tools/cmake/open62541Config.cmake.in)
4. Link static and include it into the open62541 library
    - Disadvantage: I don't know it this is expected by mbedTLS, might require building mbedTLS by open62541 from source

My favorite is the 3rd one, but I didn't find any cmake-examples that have done this in the past...

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): d4285862e11db10e018c523ba60d79e97e8cd09d
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Windows 10
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jadhavshital:
You can use this link to build mbedtls.lib (https://tls.mbed.org/kb/compiling-and-building/using-mbedtls-in-microsoft-visual-studio-2015)
And after that set encryption flag on.
Also set following flags -
[...]

PR #3714, opened 2020-06-17
Search for MbedTLS when using CMake module #3364

ccvca:
If build with encrpytion and MbedTLS backend, search for MbedTLS when including the CMake package in a project.

"mbedtls::" forces cmake to look for the target and do not fall back to library. At the moment when building the library "mbedtls" is a target that resolves to the path of library. But when including open62541 in another project, cmake falls back to library seach, as no target "mbedtls" is defined. This causes problems (especially on windows) when mbedtls is not on the well known library path. (See #3364)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1341,6 +1341,12 @@ endif()
 if(UA_ENABLE_ENCRYPTION)
     list(APPEND open62541_enabled_components "Encryption")
 endif()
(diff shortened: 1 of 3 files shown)


jpfr:
Nice! 👍

sheenamelwani52:
Hi, I want to use open62541 with mbedTLS in other project using cmake. Can you share the example of such CMake file. Do I have to link the mbedtls libraries again in this cmake fie?

PR #3717, opened 2020-06-17
fix(nc): correct nested datatype definition parsing

### LINKED ISSUE #3716: nodeset_compiler: Building of __definition__ breaks when enclosing datatype node is encounted before nested datatype node
## Description
The nodeset_compiler does not correctly constructs the `__definition__` Python attribute if the input UANodeSetXML satisfies the following conditions:
1. The nodeset contains a datatype node in which another datatype node is nested.
2. The enclosing datatype node is located before the nested datatype in the UANodeSetXML.

Point 2 has been observed in the output from the UaModeler tool from Unified Automation, where the order of the nodes in the XML seems to vary from one export operation to the other.

## Background Information (very detailed)
The nodeset compiler parses the nodes in the same order they are encountered in the input XML file. The parsed nodes (`Node` objects) are added to a dictionary that uses NodeIds for keys, and the `Node` objects for values: https://github.com/open62541/open62541/blob/cedb4527b4fcf96e8dd6eb45701f5062f303b1e0/tools/nodeset_compiler/nodeset.py#L302

Later, the definition of the nodes is built, by looping over this dictionary and calling the `buildEncoding` method on each node object:
https://github.com/open62541/open62541/blob/cedb4527b4fcf96e8dd6eb45701f5062f303b1e0/tools/nodeset_compiler/nodeset.py#L312
While the order of iteration of Python dictionary object is not guaranteed to be any specific order, in practice, for the CPython implementation, this is insertion order. 

This means that if the enclosing datatype node is encountered before the nested datatype node, buildEncoding is called first on the enclosing datatype node `Node` object.
When the `buildEncoding` call on the enclosing type encounters the nested type, `buildEncoding` is called on the nested type (**note the `indent+1` argument**):
https://github.com/open62541/open62541/blob/cedb4527b4fcf96e8dd6eb45701f5062f303b1e0/tools/nodeset_compiler/nodes.py#L581

**In this call, the definition of the nested type is built, but not added to the nested datatype node, because `indent != 0`** (note line 620):
https://github.com/open62541/open62541/blob/cedb4527b4fcf96e8dd6eb45701f5062f303b1e0/tools/nodeset_compiler/nodes.py#L614-L622

Also note that in this call to `buildEncoding` on the nested datatype node, the `__encodable__` attribute of the nested datatype is set to `True`:
https://github.com/open62541/open62541/blob/cedb4527b4fcf96e8dd6eb45701f5062f303b1e0/tools/nodeset_compiler/nodes.py#L474

When later the nested datatype node is encountered in the loop over the dictionary of nodes, and its `buildEncoding` method is called with `indent=0`, the method exits early, because the `__encodable__` attribute is already set to `True`, and the method returns early:
https://github.com/open62541/open62541/blob/cedb4527b4fcf96e8dd6eb45701f5062f303b1e0/tools/nodeset_compiler/nodes.py#L467

**So the nested datatype node's `__definition__` attribitue is never set**

## Reproduction steps
Using the following Python driving script:

```Python
from nodeset import NodeSet

ns = NodeSet()
with open('deps/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml', 'rb') as base_ns:
    with open('def_order_bug.xml', 'rb') as bugged_ns:
        ns.addNodeSet(base_ns)
        ns.addNodeSet(bugged_ns)
        nestedStructureType = ns.getNodeByIDString("ns=1;i=3002")
        assert len(nestedStructureType.__definition__) > 0
        print(nestedStructureType.__definition__)
      

```

Using the following input UANodeSet XML file: https://pastebin.com/bCrR95LB
It defines a structure datatype `structureType`, which has one member, which is of another structure type `nestedStructureType` (with nodeid  `ns=1;i=3002`, and which has one `double` member field). The XML definition of the enclosing `structureType` comes **before** the nested type in the XML

Run from the open62541 root directory with the above xml file saved in the same directory:
```
PYTHONPATH=tools/nodeset_compiler/ python def_order_bug.py
```

And notice that the assert fails because the `nestedStructureType` did not have its `__definition__` attribute added.

Modifying the input XML such that the `nestedStructureType` datatype node comes before the `structureType` datatype node in the input XML (https://pastebin.com/hk8htxSi), results in correct behaviour. The asserts succeed, and the `__definition__` is printed:

```Python
[['doubleMember', DataTypeNode(ns=0;i=11)]]
```

## Checklist

 - [x] open62541 Version (release number or git tag): cedb4527b4fcf96e8dd6eb45701f5062f303b1e0
 - [x] Operating system: Fedora 31, Python 3.7.6
 - [ ] Critical issue

wnts:
The nodeset_compiler parsing code now correctly constructs the `__definition__`
Python attribute of a `DataTypeNode` if the input UANodeSetXML satisfies the following conditions:
1. The nodeset contains a datatype node in which another datatype node is nested.
2. The enclosing datatype node is located before the nested datatype in the UANodeSetXML.

Fixes #3716

### DIFF
--- a/tools/nodeset_compiler/nodes.py
+++ b/tools/nodeset_compiler/nodes.py
@@ -617,8 +617,8 @@ class DataTypeNode(Node):
             else:
                 logger.debug( "Encodable as: " + str(self.__baseTypeEncoding__))
                 self.__isEnum__ = False
-                self.__definition__ = typeDict
             logger.debug( "")
+        self.__definition__ = typeDict
         return self.__baseTypeEncoding__
 
 class ViewNode(Node):


jpfr:
Thanks!

PR #3725, opened 2020-06-18
feat(server): bind fallback to unspecified address

bluhm:
In a Docker container the local address is usually not the gobal
address in DNS.  So binding to the published address will fail.  As
a fallback bind to any address but with the given port.

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -322,7 +322,34 @@ addServerSocket(ServerNetworkLayerTCP *layer, struct addrinfo *ai) {
     }
 
     /* Bind socket to address */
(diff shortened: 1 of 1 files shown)


ccvca:
I've just check it and this PR fixes my Docker/NAT setup.

jpfr:
Nice!
Can we add some warning-level logging?
The bound address is important from a security perspective.

bluhm:
> Can we add some warning-level logging?
I used an info log, warning is logged if socket cannot be used at all.  Also make sure that either an info or a warning is logged.

jpfr:
Cool!

PR #3728, opened 2020-06-19
fix(server): Add check for EventNotifier attribute value

opcua-tsn-team-kalycito:
- According to 'Table 42 – EventNotifierType Definition' in
   Part 3 Address Space Model specification, the
   'SubscribeToEvents' bit of EventNotifier attribute should be set
   to 1, to subscribe to Events within an object
 - If the condition is true, monitored item can be created, else it
   returns UA_STATUSCODE_BADNOTSUPPORTED status code to the client

Change-Id: Ib6acfed9c1b1c0d84e8061d8f7feb2028d77d7ba

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -13,6 +13,7 @@
  *    Copyright 2017 (c) Henrik Norrman
  *    Copyright 2017-2018 (c) Thomas Stalder, Blue Time Concept SA
  *    Copyright 2018 (c) Fabian Arndt, Root-Core
+ *    Copyright 2020 (c) Kalycito Infotech Private Limited
  */
 
 #include "ua_server_internal.h"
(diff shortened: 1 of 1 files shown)


jpfr:
Looks good. Three small comments. Can be merged when they are resolved.

Use `readWithReadValue` or similar instead of `UA_Server_readEventNotifier`.
You are already in a lock-protected region with multithreading enabled.
`UA_Server_readEventNotifier` would take the lock a second. We deliberately fail that.

The error case warrants an info-log.
Otherwise an unsuspecting user might take a loong time to figure out the problem.

You `return` early in the error case.
The code is simple of you put the error case in an `if`-bracket.
And the continue for the good-case outside of any if/else.

opcua-tsn-team-kalycito:
> Looks good. Three small comments. Can be merged when they are resolved.
> 
> Use `readWithReadValue` or similar instead of `UA_Server_readEventNotifier`.
> You are already in a lock-protected region if thread-safety for multithreading is enabled.
> `UA_Server_readEventNotifier` would take the lock a second. We deliberately fail that.
> 
> The error case warrants an info-log.
> Otherwise an unsuspecting user might take a loong time to figure out the problem.
> 
> You `return` early in the error case.
> The code is simple of you put the error case in an `if`-bracket.
> And the continue for the good-case outside of any if/else.

Thank you for the review comments. We have addressed all the comments and have provided a PR which can be merged into the master.

jpfr:
Nice! Merged.

jpfr:
A question for my understanding.
Does the EventNotifier flag also impact the "blubbling up" of events in the information model?

opcua-tsn-team-kalycito:
> A question for my understanding.
> Does the EventNotifier flag also impact the "blubbling up" of events in the information model?

No. The EventNotifier flag does not impact the creating of events in the information model. It only decides if the generated event in the information model, can be monitored by a client or not.

PR #3739, opened 2020-06-23
feat(core/pubsub) External Data Source and Information model RT variables

andreasebner:
This PR is re-opened based on the closed PR #3584 due to general updates.

In the past month a lot of work was done to enable OPC UA PubSub with real time capabilities. There were added several concepts and RT configuration options to allow short cycle times. For the 'fastest' RT mode, there is currently no option to use an information model variable as publish source. This restriction was necessary because for the sub-milliseconds cycle times the classical read service had to much overhead. Also our JIT compiler integration for RT OPC UA PubSub is based on simple copy instructions during each publish cycle. Currently the user provides an so called 'static value source' where the latest value is copied from. 

For the support of information model variables as RT source we have to think about a new source for variable nodes. An first approach was implemented in #3523. The general idea in this PR was to redirect each read and write service call, if the node is used in an PubSub RT configuration and contains an static value source. This approach worked fine but the configuration and RT code pathes were a little confusing.

This PR introduces an updated approach for the configuration of the source of variable values. In general each variable needs an source for there values (current options are shown there: https://open62541.org/doc/current/tutorial_server_datasource.html?highlight=data%20source) . In a first discussion we decided to use 'ValueBackend' as general term each type of source. The idea is to provide one function UA_Server_setVariableNode_valueBackend witch had the following structure 'UA_ValueBackend' as argument. 

https://github.com/open62541/open62541/blob/a26d5a038a1326c6632115c641262bf9d04719fe/include/open62541/server.h#L938-L955

The user can call this function instead of  UA_Server_setVariableNode_dataSource and   UA_Server_setVariableNode_valueCallback. In addition i added a new ValueBackend option called 'external'. The user can provide an self managed ptr witch is used by the read and write services. This external source is also used together with the PubSub RT mechanism to allow the mentioned information model variables as rt source.


include/open62541/server.h
```diff
@@ -895,11 +895,70 @@ typedef struct {
...
+
 UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Server_setVariableNode_valueCallback(UA_Server *server,
                                         const UA_NodeId nodeId,
                                         const UA_ValueCallback callback);
 
+typedef enum {
+    UA_VALUEBACKENDTYPE_NONE,
+    UA_VALUEBACKENDTYPE_INTERNAL,
+    UA_VALUEBACKENDTYPE_CALLBACK,
```
jpfr, line 941: Use consistent naming.
Either datasource or data-callback.

include/open62541/server.h
```diff
@@ -895,11 +895,70 @@ typedef struct {
...
+    UA_VALUEBACKENDTYPE_EXTERNAL
+} UA_ValueBackendType;
+
+typedef struct {
+    UA_ValueBackendType backendType;
+    union {
+        UA_DataValue internal;
```
jpfr, line 948: on master, internal can also have a value callback (before read, after write).

jpfr:
Nice!
Looking forward to the existing functionality being shifted to the new backend.

PR #3740, opened 2020-06-23
fix: update mdnsd dependency to recent release (minor CMake fixes)

Pro:


Pro:
Can be merged when green

PR #3750, opened 2020-06-28
fix(client) the string is null when UA_LOG_ERROR_CHANNEL is called

wenshan1:
1. How to reproduce
use the open62541 stack to write an encryption client with a pair of public and key certificates. 
use the open62541 stack to write an encryption server with another pair of public and key certificates, and the client public certificate isn't  in the server's trust list.
The client try to connect to the server, the client is crashed.
2. Root cause and Solution:
The error reason (errMessage.reason) is empty but the return value of UA_TcpErrorMessage_decodeBinary is good when the following code is executed:
```
void
processERRResponse(UA_Client *client, const UA_ByteString *chunk) {
  ......
    UA_TcpErrorMessage errMessage;
    UA_StatusCode res = UA_TcpErrorMessage_decodeBinary(chunk, &offset, &errMessage);
......
        UA_LOG_ERROR_CHANNEL(&client->config.logger, &client->channel,
                "Received an ERR response with StatusCode %s and the following reason: %.*s",
                UA_StatusCode_name(errMessage.error), (int)errMessage.reason.length, 
                errMessage.reason.data);
```
UA_LOG_ERROR_CHANNEL try to format a null string.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -240,9 +240,12 @@ processERRResponse(UA_Client *client, const UA_ByteString *chunk) {
         return;
     }
 
-    UA_LOG_ERROR_CHANNEL(&client->config.logger, &client->channel,
-                         "Received an ERR response with StatusCode %s and the following reason: %.*s",
-                         UA_StatusCode_name(errMessage.error), (int)errMessage.reason.length, errMessage.reason.data);
+    if ((errMessage.reason.length > 0) && (errMessage.reason.data != NULL)) {
+        UA_LOG_ERROR_CHANNEL(&client->config.logger, &client->channel,
+                "Received an ERR response with StatusCode %s and the following reason: %.*s",
+                UA_StatusCode_name(errMessage.error), (int)errMessage.reason.length, 
+                errMessage.reason.data);
+    }
     client->connectStatus = errMessage.error;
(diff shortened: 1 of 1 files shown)


jpfr:
Not sure about this.

We definitely want to log the status code.
Also when the reason is empty.

When the reason length is zero, the null ptr of the String data is never accessed.
Will check the definition of the %*.s format String.

wenshan1:
Reopen it.
UA_LOG_ERROR_CHANNEL will eventually call printf in the form of printf("%*.s",0, NULL).
The behavior of passing a NULL pointer to a C library function is undefined although it can work in Linux and win10.

I copy the c11 standard as the following.
7.1.4 Use of library functions
1 Each of the following statements applies unless explicitly stated otherwise in the detailed
descriptions that follow: If an argument to a function has an invalid value (such as a value
outside the domain of the function, or a pointer outside the address space of the program,
or a **null** pointer, or a pointer to non-modifiable storage when the corresponding
parameter is not const-qualified) or a type (after promotion) not expected by a function
with variable number of arguments, the behavior is undefined.

jpfr:
This pattern occurs in many many places.
Doing just a partial fix is no help.

If you can show that a commonly used libc implementation that crashes upon `printf("%*.s",0, NULL)`, then we change it everywhere.
Until then, this is WONTFIX.

PR #3754, opened 2020-07-01
Fix building websocket with amalgamation

anzeskerjanc:
I get "error C2084: function already has a body" for this functions:
connection_getsendbuffer, connection_releasesendbuffer, connection_releaserecvbuffer.

They are defined in files:
open62541\arch\network_tcp.c
open62541\arch\network_ws.c

These functions are static, so the problem occurs only with enabled amalgamation.
I also added libwebsockets include dir in cmake.


jpfr:
Thanks!

PR #3762, opened 2020-07-02
Cleanup refactoring

jpfr:


Pro:
@jpfr this PR introduces some issues...

1. Why is `UA_ServerConfig_setCustomHostname` now deprecated? (https://github.com/open62541/open62541/pull/3762/commits/794d4233c422e5323d8ac2f076cbab76a769676a)
This method is necessary especially on microcontrollers (like FreeRTOS) which do not have hostname resolution.
See e.g. https://github.com/open62541/open62541/issues/3581#issuecomment-624530235 or https://github.com/Pro/open62541-esp32/blob/master/main/main.c#L128
If there is a corresponding replacement, then this should be mentioned in the comments.

2. Cleanup mDNS Server Config  introduces (again) API backwards compatibility breaks (https://github.com/open62541/open62541/pull/3762/commits/8a4f8ebf1b751f0eab09e1290202b0018d6fec24)
I don't really see a big advantage of this commit, but it introduces a huge disadvantage that the backwards compatibility of the API is again broken.
This makes it really hard to support multiple versions (master, 1.1, 1.0) of the stack since almost every version introduces Backwards Breaks which come most of the time due to refactorings, but not really functionality changes. And we should really try to keep such refactorings to a minimum...

gokhalehemanth:
Since UA_ServerConfig_setCustomHostname is deprecated now, which function should i used to set the host name when working on FreeRTOS?

kimim:
According to the test code, you may set host name directly:

```
START_TEST(Server_set_customHostname) {
    UA_String customHost = UA_STRING("localhost");
    UA_UInt16 port = 10042;

    UA_Server *server = UA_Server_new();
    UA_ServerConfig *config = UA_Server_getConfig(server);
    UA_ServerConfig_setMinimal(config, port, NULL);
    UA_String_clear(&config->customHostname);
    UA_String_copy(&customHost, &config->customHostname);
```

gokhalehemanth:
Thank you, that helps. BTW, which test code did you refer/mention?

kimim:
I found from here: open62541/tests/server/check_server_userspace.c line 147

PR #3767, opened 2020-07-06
 fix(server): Fix quadratic runtime when adding many nodes 

jpfr:


ccvca:
Does this also allow to specify an order of the nodes for a browse request (at least form an internal point of view, I do know that the public API is missing):
https://github.com/open62541/open62541/blob/3c45c130fae4865c70a5d05b4199493f158c7e92/include/open62541/plugin/nodestore.h#L270-L276

If I get it right, the current Zip-Tree implementation is an unordered container like a "HashMap", where the input order is not equal to the output order.

Background: The new amendment "Ordered List" is used in some companion specifications (machinery and machine tool) and utilizes HasOrderedComponent-References.

jpfr:
Hey @ccvca ,

the trees are only for fast-lookup for a known NodeId or BrowseName.
The linked-list defines a fixed ordering. The Browse service respects that ordering.

What I'm unsure is how we can add a new reference at a predefined index in the order.
That must be part of the core spec in my opinion.

ccvca:
> What I'm unsure is how we can add a new reference at a predefined index in the order.
> That must be part of the core spec in my opinion.

I've been thinking about inserting them always at the end and provide methods for moving them, e.g.
`UA_Server_moveReferece(*refToMove, *targetReferenceBefore)`
E.g.:
`A, B, C`
`UA_Server_moveReferece(C, A)`
Results in: `A, C, B`

jpfr:
The first part (adding at the end) is already done...

PR #3772, opened 2020-07-08
[REVIEW] feat(pubsub) Extended support in subscriber on external Data Source and Information model RT variables

opcua-tsn-team-kalycito:
This PR adds feature on top of #3739 i.e., it extends the support of external data source variable (RT configuration) in subscriber.

### DIFF
--- a/examples/pubsub/tutorial_pubsub_subscribe.c
+++ b/examples/pubsub/tutorial_pubsub_subscribe.c
@@ -101,6 +101,7 @@ addDataSetReader(UA_Server *server) {
 
     /* Setting up Meta data configuration in DataSetReader */
     fillTestDataSetMetaData(&readerConfig.dataSetMetaData);
+
     retval |= UA_Server_addDataSetReader(server, readerGroupIdentifier, &readerConfig,
                                          &readerIdentifier);
     return retval;
@@ -110,9 +111,8 @@ addDataSetReader(UA_Server *server) {
  * Add subscribedvariables to the DataSetReader */
 static UA_StatusCode
 addSubscribedVariables (UA_Server *server, UA_NodeId dataSetReaderId) {
-    if(server == NULL) {
+    if(server == NULL)
         return UA_STATUSCODE_BADINTERNALERROR;
-    }
 
     UA_StatusCode retval = UA_STATUSCODE_GOOD;
     UA_NodeId folderId;
@@ -136,9 +136,38 @@ addSubscribedVariables (UA_Server *server, UA_NodeId dataSetReaderId) {
                              folderBrowseName, UA_NODEID_NUMERIC (0,
                              UA_NS0ID_BASEOBJECTTYPE), oAttr, NULL, &folderId);
 
-    retval |= UA_Server_DataSetReader_addTargetVariables (server, &folderId,
-                                                          dataSetReaderId,
-                                                          UA_PUBSUB_SDS_TARGET);
+    /* Create the TargetVariables with respect to DataSetMetaData fields */
+    UA_FieldTargetVariable *targetVars = (UA_FieldTargetVariable *)
+            UA_calloc(readerConfig.dataSetMetaData.fieldsSize, sizeof(UA_FieldTargetVariable));
+    for(size_t i = 0; i < readerConfig.dataSetMetaData.fieldsSize; i++) {
+        /* Variable to subscribe data */
+        UA_VariableAttributes vAttr = UA_VariableAttributes_default;
+        UA_LocalizedText_copy(&readerConfig.dataSetMetaData.fields[i].description,
+                              &vAttr.description);
+        vAttr.displayName.locale = UA_STRING("en-US");
+        vAttr.displayName.text = readerConfig.dataSetMetaData.fields[i].name;
+        vAttr.dataType = readerConfig.dataSetMetaData.fields[i].dataType;
+
+        UA_NodeId newNode;
+        retval |= UA_Server_addVariableNode(server, UA_NODEID_NUMERIC(1, (UA_UInt32)i + 50000),
+                                           folderId,
+                                           UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT),
+                                           UA_QUALIFIEDNAME(1, (char *)readerConfig.dataSetMetaData.fields[i].name.data),
+                                           UA_NODEID_NUMERIC(0, UA_NS0ID_BASEDATAVARIABLETYPE),
+                                           vAttr, NULL, &newNode);
+
+        /* For creating Targetvariables */
+        UA_FieldTargetDataType_init(&targetVars[i].targetVariable);
+        targetVars[i].targetVariable.attributeId  = UA_ATTRIBUTEID_VALUE;
+        targetVars[i].targetVariable.targetNodeId = newNode;
+    }
+
+    retval = UA_Server_DataSetReader_createTargetVariables(server, dataSetReaderId,
+                                                           readerConfig.dataSetMetaData.fieldsSize, targetVars);
+    for(size_t i = 0; i < readerConfig.dataSetMetaData.fieldsSize; i++)
+        UA_FieldTargetDataType_clear(&targetVars[i].targetVariable);
+
+    UA_free(targetVars);
     UA_free(readerConfig.dataSetMetaData.fields);
     return retval;
 }
--- a/include/open62541/server_pubsub.h
+++ b/include/open62541/server_pubsub.h
@@ -492,20 +492,34 @@ UA_Server_getDataSetWriterConfig(UA_Server *server, const UA_NodeId dsw,
 UA_StatusCode UA_EXPORT
 UA_Server_removeDataSetWriter(UA_Server *server, const UA_NodeId dsw);
 
-/**********************************************/
-/*               TargetVariables              */
-/**********************************************/
+/**
+ * SubscribedDataSet
+ * -----------------
+ * SubscribedDataSet describes the processing of the received DataSet. SubscribedDataSet defines which field
+ * in the DataSet is mapped to which Variable in the OPC UA Application. SubscribedDataSet has two sub-types
+ * called the TargetVariablesType and SubscribedDataSetMirrorType.
+ * SubscribedDataSetMirrorType is currently not supported. SubscribedDataSet is set to TargetVariablesType
+ * and then the list of target Variables are created in the Subscriber AddressSpace.
+ * TargetVariables are a list of variables that are to be added in the Subscriber AddressSpace.
+ * It defines a list of Variable mappings between received DataSet fields and added Variables
+ * in the Subscriber AddressSpace. */
 
 typedef struct {
-    UA_Guid dataSetFieldId;
-    UA_String receiverIndexRange;
-    UA_NodeId targetNodeId;
-    UA_UInt32 attributeId;
-    UA_String writeIndexRange;
-    UA_OverrideValueHandling overrideValueHandling;
-    UA_Variant overrideValue;
-    // reference to the target node value backend pointer if rt fixed size is enabled
-    UA_DataValue ** externalDataValueSource;
+    /* Standard-defined FieldTargetDataType */
+    UA_FieldTargetDataType targetVariable;
+
+    /* If realtime-handling is required, set this pointer non-NULL and it will be used
+     * to memcpy the value instead of using the Write service.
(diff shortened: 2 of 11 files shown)


src/pubsub/ua_pubsub_reader.c
```diff
@@ -1218,6 +1207,58 @@ UA_Server_DataSetReader_addTargetVariables(UA_Server *server, UA_NodeId *parentN
...
+    if(dataSetMsg->header.dataSetMessageType == UA_DATASETMESSAGE_DATAKEYFRAME) {
+        if(dataSetMsg->header.fieldEncoding != UA_FIELDENCODING_RAWDATA) {
+            size_t anzFields = dataSetMsg->data.keyFrameData.fieldCount;
+            if(dataSetReader->subscribedDataSetTarget.targetVariablesSize < anzFields)
+                anzFields = dataSetReader->subscribedDataSetTarget.targetVariablesSize;
+
+            for(UA_UInt16 i = 0; i < anzFields; i++) {
+                if(dataSetMsg->data.keyFrameData.dataSetFields[i].hasValue) {
+                    const UA_VariableNode *rtNode = (const UA_VariableNode *)
+                                                     UA_NODESTORE_GET(server, &dataSetReader->subscribedDataSetTarget.targetVariables[i].targetNodeId);
```
jpfr, line 1236: Calling `UA_NODESTORE_GET` is where you loose real time guarantees.
Because this requires synchronisation with the main server.

Instead, move the stuff from the valueBackend directly to the PubSub config.
So you don't have to do the lookup in every cycle.

src/pubsub/ua_pubsub.h
```diff
@@ -200,6 +200,9 @@ typedef struct UA_DataSetReader {
 /* Process Network Message using DataSetReader */
 void UA_Server_DataSetReader_process(UA_Server *server, UA_DataSetReader *dataSetReader, UA_DataSetMessage* dataSetMsg);
 
+/* Process Network Message in RT configuration using DataSetReader */
+void UA_Server_DataSetReader_processRTMessage(UA_Server *server, UA_DataSetReader *dataSetReader);
```
jpfr, line 204: Please use `UA_Server_DataSetReader_process` also for the RT path.

You can split this the RT path out **inside of** UA_Server_DataSetReader_process.
But everything should go through the same entry point for processing.

opcua-tsn-team-kalycito: Done

src/pubsub/ua_pubsub_reader.c
opcua-tsn-team-kalycito: Need some clarifications here. 
I can understand that new struct member UA_DataValue should be created in DSR config so we can prevent using UA_NODESTORE_GET in ua_pubsub_reader.c while receiving. But I could not create UA_DataValue for each fields as there is no DataSetField config in subscriber (it uses fieldmetadata in datasetmetadata) like it is there in publisher.

For this, I have mapped the values directly into the KeyValuePair of FieldMetaData and used this in ua_pubsub_reader.c while receiving. Does this hold good?

include/open62541/server_pubsub.h
```diff
@@ -492,6 +492,48 @@ UA_Server_getDataSetWriterConfig(UA_Server *server, const UA_NodeId dsw,
...
+    UA_FieldTargetVariablesSourceEntry *targetVariables; // Can be modifed to use LIST or TAILQ
+} UA_TargetVariablesSourceEntry;
+
+void UA_EXPORT
+UA_TargetVariablesSource_clear(UA_TargetVariablesSourceEntry *subscribedDataSetTarget);
+
+/* Return Status Code after creating TargetVariables in Subscriber AddressSpace */
+UA_StatusCode UA_EXPORT
+UA_Server_DataSetReader_createTargetVariables(UA_Server *server, UA_NodeId dataSetReaderIdentifier,
+                                              UA_TargetVariablesSourceEntry* targetVariables);
```
andreasebner, line 530: Personally, I would prefer to use the UA_TargetVariablesSource * and a size as input parameter in our API. The **'UA_TargetVariablesSourceEntry** is more internally.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1240,22 +1284,37 @@ UA_Server_DataSetReader_process(UA_Server *server, UA_DataSetReader *dataSetRead
...
             }
 
             UA_StatusCode retVal = UA_STATUSCODE_GOOD;
+            if(rg->config.rtLevel == UA_PUBSUB_RT_FIXED_SIZE) {
+                for(UA_UInt16 i = 0; i < anzFields; i++) {
+                    if(dataSetMsg->data.keyFrameData.dataSetFields[i].hasValue) {
+                        if(dataSetReader->subscribedDataSet.subscribedDataSetTarget.targetVariables[i].attributeId == UA_ATTRIBUTEID_VALUE) {
+                            memcpy((**(dataSetReader->subscribedDataSet.subscribedDataSetTarget.targetVariables[i].externalDataValueSource)).value.data,
+                                   dataSetMsg->data.keyFrameData.dataSetFields[i].value.data,
+                                   dataSetMsg->data.keyFrameData.dataSetFields[i].value.type->memSize);
```
andreasebner, line 1306: Here we are accessing the 'user memory'. The external data source PR allows the user to register callbacks to get informed if there is read or write access. Currently your implementation doesn't trigger this CB.

jpfr:
Looking very good.

As usual, I keep a close eye on public interface definitions.
So I have some wishes there.
Use the following for the new definitions in `server_pubsub.h`.

```c
typedef struct {
    /* Standard-defined FieldTargetDataType */
	UA_FieldTargetDataType targetVariable;

    /* If realtime-handling is required, set this pointer non-NULL and it will be used
     * to memcpy the value instead of using the Write service.
     * If the afterWrite method pointer is set, it will be called after a memcpy update
     * to the value. */
    UA_DataValue **externalDataValue;
    void *externalDataValueContext; /* user-defined pointer */
    void (*afterWrite)(UA_Server *server,
                       const UA_NodeId *dataSetReaderIdentifier,
                       const UA_NodeId *readerGroupIdentifier, 
                       void *externalDataValueContext,
                       const UA_DataValue **externalDataValue);
} UA_FieldTargetVariable;

typedef struct {
    size_t targetVariablesSize;
    UA_FieldTargetVariable *targetVariables;
} UA_TargetVariables;
```

- Reuses standard-defined data types
- Data type names short and to the point
- Allow the user to add a callback method if he wishes to be informed about an incoming value change

Can be merged then.

opcua-tsn-team-kalycito:
@jpfr Thanks for restructuring. Completely in-line with the usage of afterWrite callback which is to notify the write to the user in RT path. Can you clarify why do we need an additional pointer "externalDataValueContext" for memcpy as we are already copying the value inside externalDataValue double pointer? Can we use a bool called "rtValueSourceEnabled" for the check instead of a pointer "externalDataValueContext" so that it is understandable to the user?

jpfr:
> Can you clarify why do we need an additional pointer "externalDataValueContext" for memcpy as we are already copying the value inside externalDataValue double pointer?

It could be that the user wants to trigger an third-party API when an update is received.
Having a context pointer is a standard pattern used in many places of the stack.
You could rename it to `targetVariableContext` or similar of that is more clear.

>  Can we use a bool called "rtValueSourceEnabled" for the check instead of a pointer "externalDataValueContext" so that it is understandable to the user?

The context can be NULL. If rt is enabled or not should depend on the DataValue double-pointer to be non-null.

jpfr:
Thanks for the update.

I had a look at `server_pubsub.h`. The datatype `UA_TargetVariables` was purely a wrapper and not very useful.
See the added commit with some simplifications (but not functional changes).

Now to the problematic part:

In the `UA_DataSetReaderConfig`, the FieldTargetVariables are stored as well.

However,
- this targetVariables in the UA_DataSetReaderConfig are currently not set at all!
- the information is duplicated in the `UA_DataSetReader`. Don't duplicate information!
- changes to the targetVariables are not considered in `UA_Server_DataSetReader_updateConfig`

We should avoid redundant API. It should be enough to set the targetvariables in one place.
If we set the target variables it in the config, then we don't need to have a UA_Server_DataSetReader_createTargetVariables at all!

That is preferrable because it keeps the API surface smaller.

opcua-tsn-team-kalycito:
@jpfr Thanks for the comments.

>  the information is duplicated in the UA_DataSetReader. Don't duplicate information!

Yes, you are right, we should avoid redundant information in DataSetReader.

> this targetVariables in the UA_DataSetReaderConfig are currently not set at all!
> changes to the targetVariables are not considered in UA_Server_DataSetReader_updateConfig

From the usage of TargetVariables, I think we don't need definition in DataSetReaderConfig. As TargetVariables are similar to dataSetFields, the current implementation can be modified to use addTargetVariable and removeTargetVariable API in the future to add or remove the variables (similar to DataSetFields) instead of modifying in updateReaderConfig API.

>  If we set the target variables it in the config, then we don't need to have a UA_Server_DataSetReader_createTargetVariables at all!

UA_Server_DataSetReader_createTargetVariables API is required, as in future, we will be using this API to create the user defined Target variable nodes under TargetVariablesType Object (which will be created under DataSetReader Object). 

For now, I have removed the targetVariables from DataSetReaderConfig and modified TSN examples in the new commit. In case of any design change in TargetVariable, can we do that in a separate PR?

examples/pubsub/tutorial_pubsub_subscribe.c
```diff
@@ -136,9 +108,70 @@ addSubscribedVariables (UA_Server *server, UA_NodeId dataSetReaderId) {
...
+    readerConfig.writerGroupId    = 100;
+    readerConfig.dataSetWriterId  = 62541;
+
+    /* Setting up Meta data configuration in DataSetReader */
+    fillTestDataSetMetaData(&readerConfig.dataSetMetaData);
+    retval |= addSubscribedVariables(server);
+
+    retval |= UA_Server_addDataSetReader(server, readerGroupIdentifier, &readerConfig,
+                                         &readerIdentifier);
+
```
andreasebner, line 170: Maybe we can also show the use of our dedicated function 'createTargetVariables()'.

include/open62541/server_pubsub.h
```diff
@@ -492,6 +492,59 @@ UA_Server_getDataSetWriterConfig(UA_Server *server, const UA_NodeId dsw,
...
+    /* Standard-defined FieldTargetDataType */
+    UA_FieldTargetDataType targetVariable;
+
+    /* If realtime-handling is required, set this pointer non-NULL and it will be used
+     * to memcpy the value instead of using the Write service.
+     * If the afterWrite method pointer is set, it will be called after a memcpy update
+     * to the value. */
+    UA_DataValue **externalDataValue;
+    void *targetVariableContext; /* user-defined pointer */
+    void (*afterWrite)(UA_Server *server,
```
andreasebner, line 523: The afterwrite could maybe be part of the external data source itself (more generic).

include/open62541/server_pubsub.h
```diff
@@ -492,6 +492,59 @@ UA_Server_getDataSetWriterConfig(UA_Server *server, const UA_NodeId dsw,
...
+                       const UA_NodeId *readerIdentifier,
+                       const UA_NodeId *readerGroupIdentifier,
+                       const UA_NodeId *targetVariableIdentifier,
+                       void *targetVariableContext,
+                       UA_DataValue **externalDataValue);
+} UA_FieldTargetVariable;
```
andreasebner, line 529: Maybe UA_TargetVariableField would be more clear.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1091,41 +1134,72 @@ UA_DataSetReader_setPubSubState(UA_Server *server, UA_PubSubState state, UA_Data
...
+    tmp.targetVariables = (UA_FieldTargetVariable*)(uintptr_t)targetVariables;
+    return UA_TargetVariables_copy(&tmp, &dataSetReader->config.subscribedDataSet.subscribedDataSetTarget);
 }
 
-/* Adds Subscribed Variables from the DataSetMetaData for the given DataSet into
- * the given parent node and creates the corresponding data in the
- * targetVariables of the DataSetReader */
-UA_StatusCode
+/* Functionality of addTargetVariables API here is the same with respect to createMirrorVariables.
+ * This API will be used in future to create mirror Variables - TODO */
```
andreasebner, line 1200: Why don't we call it internally addMirrorVariable?

PR #3777, opened 2020-07-09
refactor(client): initialize sockfd with invalid socket

bluhm:
In UA_ClientConnectionTCP_init() the memset of the connection
initialized the sockfd field with 0.  As this is a valid file
decriptor, some other bug caused to close stdin.  If sockfd is
initilized with -1, it will fail earlier which helps debugging.

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -827,6 +827,7 @@ UA_ClientConnectionTCP_init(UA_ConnectionConfig config, const UA_String endpoint
     memset(&connection, 0, sizeof(UA_Connection));
 
     connection.state = UA_CONNECTIONSTATE_OPENING;
+    connection.sockfd = UA_INVALID_SOCKET;
     connection.send = connection_write;
     connection.recv = connection_recv;
     connection.close = ClientNetworkLayerTCP_close;


jpfr:
rebased

jpfr:
Something is odd with that change.
I can't get the build to not timeout on Windows.

cochicde:
I'd say that the problem lies here:

https://github.com/open62541/open62541/blob/master/arch/network_tcp.c#L703-L715

In line 703, the check should be as in Line 710: if(clientsockfd == UA_INVALID_SOCKET)

Reason seems to be as stated [here](https://docs.microsoft.com/en-us/windows/win32/winsock/socket-data-type-2)

`Because the SOCKET type is unsigned, compiling existing source code from, for example, a UNIX environment may lead to compiler warnings about signed/unsigned data type mismatches.

This means, for example, that checking for errors when the socket and accept functions return should not be done by comparing the return value with –1, or seeing if the value is negative (both common and legal approaches in UNIX). Instead, an application should use the manifest constant INVALID_SOCKET as defined in the Winsock2.h header file`

jpfr:
Solved via #3851.
Thanks for the groundwork @bluhm !

PR #3779, opened 2020-07-10
[Review] feat: add support for WebSocket Secure (wss)

demike:
- allow passing certificate and privateKey to the websocket network layer (if set to NULL standard ws is used)
- shows the right protocol (ws or wss) in the discovery url
- adds an usage example (examples/encryption/server_encryption.c)
- works with certificates created by: tools/certs/create_self-signed.py

prerequisits for using wss: 
- UA_ENABLE_WEBSOCKET_SERVER has to be ON
- libwebsockets has to be built with SSL enabled


jpfr:
Nice! Looking forward to merging this.
You see the errors in the CI?

Just set the prefix of the PR name to [Review] when we shall have a look.

demike:
@jpfr could you restart the ci checks. The travis task failed because of apt having network issues

jpfr:
Looking good! Merged.

PR #3788, opened 2020-07-17
fix(client): don't check security policy for anonym token type

JulZimmermann:
Some servers send a securityPolicyUri even with tokenType UA_USERTOKENTYPE_ANONYMOUS which doesn't make sense at all.

An open62541 based client should be able the connect to such servers anyway.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -627,7 +627,8 @@ responseGetEndpoints(UA_Client *client, void *userdata, UA_UInt32 requestId,
             const UA_DataType *tokenType = client->config.userIdentityToken.content.decoded.type;
 
             /* Usertokens also have a security policy... */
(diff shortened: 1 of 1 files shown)


jpfr:
Good find!
Can you send a PR against the 1.1 (or even better 1.0) branch?
We can then forward-merge up to master and do minor releases.

JulZimmermann:
Sure. Thanks for accepting this PR!

Should I create a PR for the 1.0 branch or the 1.0 AND the 1.1 then?

jpfr:
1.0 in this case.
Then we cherry pick the commit on the newer branches.

jpfr:
1.0 in this case.
Then we cherry pick the commit on the newer branches.

JulZimmermann:
Created #3790

Issue #3787, opened 2020-07-17
Client to read custom Datatype/ ExtensionObject


skubidu1324:
## Description
I want to develop a client to read a Struct/ExtensionObject from a Siemens PLC. The struct contains 16 UA_Boolean variables and is defined in root/Types/VariableTypes/BaseVariableType/
#1420 ,  #2369 and #3108 are related but havn't given a solution.  

## Background Information / Reproduction Steps
I followed the steps from `master\examples\custom_datatype`. I created the struct, UA_DataTypeMember and UA_DataType for that specific struct but only for client side because the server is given.

```
#define CustomDataStruct_binary_encoding_id        1

typedef struct {
	UA_Boolean Data01;
        ...
        UA_Boolean Data16;
} CustomDataStruct;

static UA_DataTypeMember CustomDataStruct_members[16] = {
	{
		UA_TYPENAME("Data01")		        /* .memberName */
		UA_TYPES_BOOLEAN,                       /* .memberTypeIndex, points into UA_TYPES since namespaceZero is true */
		0,                                     /* .padding */
		true,                                  /* .namespaceZero, see .memberTypeIndex */
		false,                                 /* .isArray */
		false                                  /* .isOptional */
	},
       ...
}

static const UA_DataType CustomDataStructType = {
	UA_TYPENAME("CustomDataStruct")                         
	UA_NODEID_STRING(3, "VT_\"CustomDataStruct\""), /*nodeid from root/Types/VariableTypes/BaseVariableType/*/
	sizeof(CustomDataStruct),                                        
	0, 
	UA_DATATYPEKIND_STRUCTURE, 
	false,
	false, 
	16, 
	CustomDataStruct_binary_encoding_id, 
	CustomDataStruct_members
};
```
Then added this type to the client configuration as the example shows.

```

    UA_DataType types[1];
    types[0] = CustomDataStructType;
    UA_DataTypeArray customDataTypes = {NULL, 1, types};

    UA_Client *client = UA_Client_new();
    UA_ClientConfig *cc = UA_Client_getConfig(client);
    UA_ClientConfig_setDefault(cc);
    cc->customDataTypes = &customDataTypes;
```

 After reading the value I get the ExtensionObject but can't see any relation to my custom datatype. 

```
    retval = UA_Client_readValueAttribute(client, nodeId, &value);
    if(retval == UA_STATUSCODE_GOOD) {
        CustomDataStruct *p = (CustomDataStruct *)value.data;
    }
```

[...]

PR #3795, opened 2020-07-21
[BREAKING] feat(core): Allow all types of NodeId for the binary encoding identifier

jpfr:
This is a breaking change for those using custom datatypes.
The change is necessary for the use case of #3787.


ccvca:
I just wonder if a list of all breaking changes makes sense (e.g. in the wiki)?
So updating to the newest version is easier.
In a short version, a list with the relevant pull request and thein title should help.
In an extended format, a short "migration" example could also be provided.

PR #3798, opened 2020-07-22
feature: prepare open62541 for fuzzing with llvm 10 and above

mlgiraud:


mlgiraud:
@jpfr Could you restart the pipeline? Something seems to have gone wrong with loading the dependencies.

jpfr:
Nice, thanks.

PR #3809, opened 2020-07-28
Fix: BadTypeDefinitionInvalid for methods in instance declarations

JGrothoff:
recursiveCallConstructors calls constructors for all unconstructed **child** nodes of a (parent) node recursively. Hence it determins the type of the child node in [line 1115](https://github.com/open62541/open62541/blob/master/src/server/ua_services_nodemanagement.c#L1115). So only typed children (variable and object node classes) have to be checked.
Before the fix the decission was based on the parent node class, which leads to errors, if a child is a method node for example.

This error occured while trying to load the PackML Companion Specification for the ExecuteState in PackMLMachineStateMachineType and should even be fixed in upcomming v1.1.2 updates.


jpfr:
Nice. Thanks!

PR #3812, opened 2020-07-30
refactor(plugin): log mbedTLS return code

robertns:


robertns:
> Don't you want to log the errBuff string?

My focus was on error from `mbedtls_x509_crl_parse_file` (and `mbedtls_strerror` isn't called there), but I can try with that as well. Thanks for pointing this out.

jpfr:
Nice, thanks!

PR #3813, opened 2020-07-30
feat(server): Add support for HasInterface on Object

Pro:
A HasInterface reference can be on an ObjectType (already supported) and on an Object (not yet supported).

This PR adds support for HasInterface on an Object.

See https://reference.opcfoundation.org/v104/Core/docs/Amendment7/4.9.2/


jpfr:
Cool!
Can we reuse the `addInterfaceChildren` method als for the ObjectTypes?

Pro:
I kept it separate from `getParentTypeAndInterfaceHierarchy` because separating would probably introduce a small performance impact.
But we can also integrate it into the ObjectType if desired

jpfr:
Both work for me.
Can be merged whenever you are happy. :-)

PR #3818, opened 2020-08-03
[REVIEW] fix(pubsub):  Multiple receive message handling for every UA_ReaderGroup_subscribeCallback

opcua-tsn-team-kalycito:
Previously UA_ReaderGroup_subscribeCallback receives and process only one receive message for every cycle even though multiple messages are queued. The queued messages are processed only in the next cycle. Now subscriber functionality is modified to handle multiple received messages.


include/open62541/plugin/pubsub.h
```diff
@@ -65,7 +65,7 @@ struct UA_PubSubChannel {
 
     /* Receive messages. A regist to the message source is needed before. */
     UA_StatusCode (*receive)(UA_PubSubChannel * channel, UA_ByteString *,
-                             UA_ExtensionObject *transportSettings, UA_UInt32 timeout);
+                             UA_ExtensionObject *transportSettings, UA_UInt32 timeout, UA_UInt16 *noOfRecvMessages);
```
jpfr, line 68: I'm not a fan of this change:
There are two options this can be used.

1. Get the number of messages in the ByteString
We know after decoding if we have reached the end of the message bytestring.
If not, continue with the next message. So this is not required.

2. Wait until noOfRecvMessages have been received.
This is an antipattern. Blocking waits should be avoided.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -773,69 +774,78 @@ UA_DataSetReader *UA_ReaderGroup_findDSRbyId(UA_Server *server, UA_NodeId identi
...
-             */
-            UA_DataSetReader *dataSetReader = LIST_FIRST(&readerGroup->readers);
-            /* Decode only the necessary offset and update the networkMessage */
-            if(UA_NetworkMessage_updateBufferedNwMessage(&dataSetReader->bufferedMessage, &buffer) != UA_STATUSCODE_GOOD) {
-                UA_LOG_INFO(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                             "PubSub receive. Unknown field type.");
-                UA_ByteString_deleteMembers(&buffer);
-                return;
-            }
+    UA_UInt16 *messagesReceived = UA_UInt16_new();
```
jpfr, line 786: This one can be stack-allocated.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -773,69 +774,78 @@ UA_DataSetReader *UA_ReaderGroup_findDSRbyId(UA_Server *server, UA_NodeId identi
...
+    *messagesReceived           = UA_MAX_RECEIVE_MESSAGES;
+    connection->channel->receive(connection->channel, buffer, NULL, 1000, messagesReceived);
+    for (UA_UInt16 iterator = 0; iterator < *messagesReceived; iterator++) {
```
jpfr, line 789: For the loop you can simply check if the decode position is at the end of the received buffer.

opcua-tsn-team-kalycito:
@jpfr Modified as per your comments to handle multiple messages in a single receive buffer. Now this PR is ready for the review.

plugins/ua_pubsub_ethernet.c
```diff
@@ -192,7 +196,7 @@ UA_PubSubChannelEthernet_open(const UA_PubSubConnectionConfig *connectionConfig)
     struct sockaddr_ll sll = { 0 };
     sll.sll_family = AF_PACKET;
     sll.sll_ifindex = channelDataEthernet->ifindex;
-    sll.sll_protocol = htons(ETHERTYPE_UADP);
+    sll.sll_protocol = htons(ETH_P_802_2);
```
allannielsen, line 201: Why this change?

As far as I can see in the spec, this was correct before.

Try have a look at https://reference.opcfoundation.org/v104/Core/docs/Part14/7.3.3/

PR #3824, opened 2020-08-04
feat(client): state callback context

bluhm:
Callbacks that provide a context are much more useful for the callee.
Then it can derive its working data without a global variable.  Add
a stateContext to client config and pass it to the stateCallback
as parameter.

### DIFF
--- a/examples/client_async.c
+++ b/examples/client_async.c
@@ -13,7 +13,7 @@
 /* async connection callback, it only gets called after the completion of the whole
  * connection process*/
 static void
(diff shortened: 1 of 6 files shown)


jpfr:
Humm. We have a context void pointer in the client config.
That was meant for these cases.

Would it work for you to use the context pointer from the config?
If not, please let us understand the usage scenario.

bluhm:
> Would it work for you to use the context pointer from the config?

Yes, clientContext is fine for me.

PR #3827, opened 2020-08-05
Pub sub export api ua pub sub connection regist

chuongvhn:
Export API UA_PubSubConnection_regist(). This is necessary for application code to automatically publish node's value once PubSubConnection is added and server is up.


chuongvhn:
wrong PR, I'll delete

PR #3828, opened 2020-08-05
Export UA_PubSubConnection_regist. This is necessary for application …

chuongvhn:
Export API UA_PubSubConnection_regist. This is necessary for application code to automatically publish node's value once PubSubConnection is added and server is up


andreasebner:
The UA_PubSubConnection functions are our internal low-level networking API and should not be exposed to the user. Since the subscriber API isn't finished yet we have added examples which are using this interal API to receive messages.

Can you please explain the case where this function is needed a littele bit more detailed (or add a code snipped)?

andreasebner:
If you have further problems or if there are any reasons to extend our API please create an issue or PR and refer to this one.

PR #3832, opened 2020-08-05
fix(debian): build should not depend on unversioned essential package

jrcoding:
Lintian throws the error 
> E: open62541 source: build-depends-on-essential-package-without-using-version build-depends: base-files

because of the build dependency on the essential package `base-files` without a specific version

According to Debian policy [4.3](http://www.l3jane.net/doc/debian-policy/policy.html/ch-source.html#package-relationships)  there should only be a build dependency on an essential package if a specific version of the essential package is needed.

In this case the dependency on `base-files` should be removed because no specific version is needed for the build and  the package `base-files`  will be installed anyway because `base-files` is an essential package


jrcoding:
@jpfr -  should we also merge it to  1.1 ?

PR #3838, opened 2020-08-11
Add option to include encryption plugins into the amalgamated files

basyskom-jvoe:
Currently, UA_ENABLE_ENCRYPTION_* must be activated to have the encryption plugins included into the amalgamation files.

These two options allow to include the encryption plugins without activating them. This allows the application including the amalgamation files to decide if encryption should be activated or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -817,7 +817,9 @@ if(UA_ENABLE_PARSING)
     list(APPEND lib_sources ${PROJECT_SOURCE_DIR}/src/ua_types_lex.c)
 endif()
 
-if(UA_ENABLE_ENCRYPTION_MBEDTLS)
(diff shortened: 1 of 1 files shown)


jpfr:
Good idea to include more in the amalgamation.

Adding more and more build flags is difficult.
We have already accumulated a lot of complexity there.

Maybe there is an easy way out.
Can you just always include the encryption plugins in the build?
And surroud the code in the .c files with `UA_ENABLE_ENCRYPTION_MBEDTLS` or similar.

basyskom-jvoe:
Good idea. I have changed the condition to UA_ENABLE_AMALGAMATION.
The guards were already present in all the encryption plugin files.

jpfr:
Great 👍

basyskom-jvoe:
Follow-up: #3845

PR #3842, opened 2020-08-13
Fix segfault in ua_pki_openssl.c

basyskom-jvoe:
If the value is not set to NULL, a heap-use-after-free occurs in UA_CertificateVerification_clear().


jpfr:
Good find.

PR #3841, opened 2020-08-13
[WIP] Tool to test all possible failure points that use malloc, calloc and realloc

cochicde:
This is a tool I've been playing around with. It works as follow:

Once the test I created (check_allocation) is compiled, I added a post command in CMake that executes a python script. This python script (asmParser.py) does the following:
 - objdump the check_allocation executable to read the whole assembly code
 - Parse the assembly code, and create a complete tree of functions calls 
 - It looks at malloc, realloc and calloc, and for each, it starts going backwards recursively to see where these functions are being called from. For each possible place in code, the script stores the assemly address and the source code line and writes this information in a file (one address per line)

The test itself (check_allocation) then:
- reads this file, and for each line, it gets the adddess and soruce code line. It sets the address as the next address where malloc, calloc and realloc should fail in a global variable, and run some tests (for now, it's just setting the config default in a server, and deleteing it later). 
- Each of these functions (malloc, realloc and calloc) are implemented with a wrapper arround it which uses the function "backtrace" to check the stack. If it finds in the stack the addrses to break which was recentrly assigned, it will return NULL from there on, otherwise, it returns the usual malloc, realloc or calloc. 

On my fork, the test is failing for some reason I don't know. I assume there is memory leak, but for some reason I don't see it (I think the valgrind output should be outputed to the terminal, but nothing's there). I tested it locally some time ago, and found the valgrind log output and there are some memory leaks indeed. 

I copy the output of that output for you to see it:

==7385== 65 bytes in 1 blocks are definitely lost in loss record 4 of 11
==7385==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==7385==    by 0x402D75: mallocWrapper (check_allocation.c:94)
==7385==    by 0x403BEF: UA_String_fromChars (ua_types.c:108)
==7385==    by 0x4B8D5D: createEndpoint (ua_config_default.c:91)
==7385==    by 0x4B9AF3: UA_ServerConfig_addEndpoint (ua_config_default.c:380)
==7385==    by 0x4BA10A: UA_ServerConfig_setMinimalCustomBuffer (ua_config_default.c:468)
==7385==    by 0x40291C: UA_ServerConfig_setMinimal (server_config_default.h:61)
==7385==    by 0x402968: UA_ServerConfig_setDefault (server_config_default.h:85)
==7385==    by 0x402A2B: functionToTest (check_allocation.c:53)
[...]

PR #3843, opened 2020-08-14
Fix python regex removing all content

cochicde:
Do not use re.DOTALL so only one line will match and the rest of the content is kept

THis is the reason valgrind is not outputing the content when a failure occur. This can be seen in #3841 where the jobs now output the content after the same change was pushed there


jpfr:
Thanks!

PR #3846, opened 2020-08-17
OpenSSL PEM support

basyskom-jvoe:
PEM currently does not work for the local certificate because the UA_ByteString holding the certificate file content is sent directly to the communication partner and part 6 states that a certificate must be DER encoded.

If necessary, this could be made possible by converting a loaded PEM encoded local certificate to DER.

### DIFF
--- a/plugins/securityPolicies/openssl/securitypolicy_openssl_common.c
+++ b/plugins/securityPolicies/openssl/securitypolicy_openssl_common.c
@@ -197,18 +197,17 @@ UA_Openssl_X509_GetCertificateThumbprint (const UA_ByteString * certficate,
 
 static UA_StatusCode
 UA_Openssl_RSA_Private_Decrypt (UA_ByteString *       data, 
-                               const UA_ByteString * privateKey,
+                               EVP_PKEY * privateKey,
                                UA_Int16              padding) {
     if (data == NULL || privateKey == NULL) {
         return UA_STATUSCODE_BADINVALIDARGUMENT;
     }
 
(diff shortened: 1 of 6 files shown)


plugins/ua_config_default.c
```diff
@@ -807,9 +807,11 @@ UA_ClientConfig_setDefaultEncryption(UA_ClientConfig *config,
                        UA_StatusCode_name(retval));
     }
 
-    if(config->securityPoliciesSize == 0) {
+    // Policy "None" is already present before the encryption related security policies are added
```
jpfr, line 810: Failing to load a certificate should not lead to the removal of policy-none.
This is only required as a malloc cleanup when there is no element in the array.

jpfr (commented):
Looks good with just a minor comment.
What I am still missing is documentation.

The user should know that PEM works for OpenSSL but not with mbedtls.
And he should not have to "discover" that fact.

Keep up the good work!

plugins/ua_config_default.c
basyskom-jvoe: As a user of the API, I would not expect a function named "UA_ClientConfig_setDefaultEncryption()" to give me a client config which can only do the "None" security policy.

But as it would be better not to change behavior from one 1.1 release to the next, I'll restore the old code for this part.

basyskom-jvoe:
Actually, this PR makes the OpenSSL and mbedTLS equal because the mbedTLS functions used to load certificates and private keys take both PEM and DER encoded data.

jpfr:
Thanks for the explanation. Merged. 👍

basyskom-jvoe:
Thanks for the merge!

Issue #3849, opened 2020-08-18
OPCUA PubSub over AMQP


waheedejaz:
## Description
Add support for AMQP transport
## Background Information / Reproduction Steps
Using Qpid Proton lib c
https://github.com/apache/qpid-proton/tree/master/c

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

Please provide the following information:

 - [ ] open62541 Version (release number or git tag):
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

waheedejaz:
Please give an early review. The AMQP publish example works with broker example in qpid proton lib proton/c/examples/broker.

The build system is temporarily modified to link the proton lib as it doesn't build successfully as an object library owing to all the warnings enabled in the open62541 build system.

jackybek:
Hi,
I am trying to evaluate .. but when i download the qpid-proton-cpp tar file, there are only 2 files. 
a. libqpid-proton-cpp.so.12.0.0 
b. libqpid-proton-cpp.so.12 

Seemed that libqpid-proton-cpp.a is missing.
I am getting the following link errors:

...
/usr/bin/ld: //usr/local/lib/libqpid-proton-cpp.so: undefined reference to `pn_data_get_binary'
/usr/bin/ld: //usr/local/lib/libqpid-proton-cpp.so: undefined reference to `pni_message_get_extra'
/usr/bin/ld: //usr/local/lib/libqpid-proton-cpp.so: undefined reference to `pn_condition_copy'
/usr/bin/ld: //usr/local/lib/libqpid-proton-cpp.so: undefined reference to `pn_message_set_content_type'
/usr/bin/ld: //usr/local/lib/libqpid-proton-cpp.so: undefined reference to `pn_message_is_durable'
/usr/bin/ld: //usr/local/lib/libqpid-proton-cpp.so: undefined reference to `pn_message_get_correlation_id'
/usr/bin/ld: //usr/local/lib/libqpid-proton-cpp.so: undefined reference to `pn_ssl_domain_free'
...

Seemed that it is looking for proton lib functions..

During the installation process, i already did the following:
1. sudo lfconfig -v
[...]

PR #3850, opened 2020-08-18
WIP: pubsub over amqp

waheedejaz:
Up for the initial review.

Basic PubSub publish over AMQP is added, works with ay AMQP 1.0 broker

NOTE:
The build system is adapted to link the qpid proton lib. At the moment proton c doesn't successfully build using all the warnings enabled in the open62541 build. This will be fixed later.

#3849


ccvca:
Maybe resolve the compile issues by using "libqpid-proton11-dev" and find_package()? Then there is no need to include all the sources in the open62541-project.

jpfr:
Hey there!

Wow, this looks good!
You say that proton does not support all the compiler warnings that we enable by default.
I say that we don't care. We cannot maintain 28,000 lines of code of external dependency that we don't even know.

Most systems can install proton as a dependency.
So let's handle it like we handle mbedtls/openssl.

https://packages.debian.org/stable/libs/libqpid-proton-cpp12

We can have a closer look when the line-count when only the changes to open62541 itself are in the PR.

waheedejaz:
Proton lib is linked using findPackage(). The previous commit is removed. PR rebased

jpfr:
Thanks. We will review when the [WIP] is removed from the title.
Keep up the good work!

plugins/amqp/ua_pubsub_amqp.c
```diff
@@ -0,0 +1,227 @@
...
+     *   Memory Allocation
+     *******************************************/
+    /* Create a new pubsub channel */
+    UA_PubSubChannel *ua_amqpChannel = (UA_PubSubChannel *) UA_calloc(1, sizeof(UA_PubSubChannel));
+    /* Create a new AMQP Proactor database */
+    UA_AmqpContext *amqpCtx = (UA_AmqpContext *) UA_calloc(1, sizeof(UA_AmqpContext));
+
+    if(!ua_amqpChannel || !amqpCtx){
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "PubSub AMQP Connection creation failed. Out of memory.");
+        UA_Destroy_AmqpChannel(ua_amqpChannel, false);
```
basyskom-meerkoetter, line 72: would this leak amqpCtx on the error path?

plugins/amqp/ua_pubsub_amqp.c
```diff
@@ -0,0 +1,227 @@
...
+        return NULL;
+    }
+
+    /* Initialize sub-elements in AMQP Context */
+    amqpCtx->ua_connection = (UA_Connection*)UA_calloc(1, sizeof(UA_Connection));
+    amqpCtx->driver = (pn_connection_driver_t *) UA_calloc(1, sizeof(pn_connection_driver_t));
+
+    if(!amqpCtx->ua_connection || !amqpCtx->driver) {
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "PubSub AMQP: Context creation failed. Out of memory.");
+        UA_Destroy_AmqpChannel(ua_amqpChannel, false);
```
basyskom-meerkoetter, line 82: also here. amqpCtx and either ua_connection or driver would leak?

plugins/amqp/ua_pubsub_amqp.c
```diff
@@ -0,0 +1,227 @@
...
+    /* Get the url form connection config*/
+    UA_NetworkAddressUrlDataType address;
+
+    if(UA_Variant_hasScalarType(&amqpConnectionConfig->address,
+                                &UA_TYPES[UA_TYPES_NETWORKADDRESSURLDATATYPE])){
+        address = *(UA_NetworkAddressUrlDataType *)amqpConnectionConfig->address.data;
+    } else {
+        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
+                     "PubSub AMQP Connection creation failed. Invalid Address.");
+        return NULL;
```
basyskom-meerkoetter, line 102: leaks resources

jackybek:
I tested your code but realised that the prefix opc.amqp://xxx.xxx.xxx.xxx:5672 was not changed to opc.tcp://xxx.xxx.xxx.xxx:5672 during runtime.
As compared to the implementation by kalycito on pubsub mqtt, when I pass in opc.mqtt://xxx.xxx.xxx.xxx, the code changes the prefix to opc.tcp before sending to the mqtt broker.
This caused failure when open62541 tries to extract the host name from the URL.  The host name returned was ‘empty string’.

cloud-rocket:
@waheedejaz - is this PR still alive?

PR #3851, opened 2020-08-19
[Review] fix(arch): Use UA_INVALID_SOCKET where needed

cochicde:
The initial value of a socket should be UA_INVALID_SOCKET and not 0 nor -1. Also, when checking that, the macro is used

I also removed some non-needed local variables.

This was triggered by #3777


cochicde:
travis failed on a unreacheable server. It should go away when retriggered.

There are no more changes I'll make

jpfr:
restarted

jpfr:
Thanks for this!
Squashed and merged.

PR #3852, opened 2020-08-19
[Review] fix(leak): Fix memory leaks from config_default

cochicde:
Fix leaks signalized by #3841


jpfr:
Great. It looks like your tool helps squashing quite a few bugs that were overlooked by clang-analyzer.
Impressive!

PR #3854, opened 2020-08-20
Add support for PEM encoded local certificates

basyskom-jvoe:
The buffer containing the local certificate is directly added to OPC UA messages without further conversion.
According to Part 6 of the specification, the required encoding for certificates is DER, so a conversion is necessary for PEM encoded local certificates.


jpfr:
👍

PR #3855, opened 2020-08-21
Fix leak in UA_Server_newWithConfig

### LINKED ISSUE #3853: Memory not properly handled in UA_Server_newWithConfig
## Description

Memory not properly handled in UA_Server_newWithConfig

While trying to fix issues regarding malloc failures, I found an issue in this function

https://github.com/open62541/open62541/blob/ee275e79310280a71fad3b21f6d430c0c046091b/src/server/ua_server.c#L341-L350

The problem is that the parameter config has some allocated memory in it (in the hashMap). Now, this function can fail in two places:
 - UA_calloc (line 345) wich returns NULL. The allocated memory in the config parameter should be deallocated by the one who called this function
 - UA_Server_init (Line 349): If this fails, the UA_Server_init will take care of deleteing the whole server, including its config variable (and deallocating the pointers there), which was COPIED in line 348. That means that the function will return NULL, but the memory (which came from the config parameter) in this case is already deallocated.

What happens now, is that the function that called UA_Server_newWithConfig can receive a NULL and won't know if the memory was deallocated already or not. 

My first guess was that the caller can check the parameter he send, and see if the memory pointers are NULL. BUT, since the config variable was copied, the copied config variable in the server might have a NULL pointer in it (failure case 2), but the config parameter varaible will never be NULL at this point, since it has never changed (the pointer variable in it is still non-NULL, although the memory where is pointing was already deallocated)

I tried to clean the config parameter variable in UA_Server_newWithConfig but it's const.

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

 - [ ] open62541 Version (release number or git tag):
 - [ ] Operating system:
 - [ ] Critical issue

cochicde:
Fix #3853

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -311,7 +311,7 @@ UA_ServerConfig_setCustomHostname(UA_ServerConfig *config,
 /* Creates a new server. Moves the config into the server with a shallow copy.
  * The config content is cleared together with the server. */
 UA_Server UA_EXPORT *
-UA_Server_newWithConfig(const UA_ServerConfig *config);
+UA_Server_newWithConfig(UA_ServerConfig *config);
 
 void UA_EXPORT UA_Server_delete(UA_Server *server);
 
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -51,12 +51,7 @@ UA_Server_new() {
         return NULL;
     }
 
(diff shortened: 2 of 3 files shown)


jpfr:
Thanks :-)

Issue #3858, opened 2020-08-27
Check for node in ObjectsFolder fails for references derived from HasComponent


TonjaHeinemann:
## Description
When trying to send an event from a node that is referenced with HasOrderedComponent, sending the event fails with the error message "Node for event must be in ObjectsFolder!"

## Background Information / Reproduction Steps
* Create node structure using HasOrderedComponent
* set the event notifier attribute for the node referenced with HasOrderedComponent
* generate an event from the node referenced with HasOrderedComponent

* see /src/server/ua_subscription_events.c, lines 601 to 605:
    /* Make sure the origin is in the ObjectsFolder (TODO: or in the ViewsFolder) */
    /* Only use Organizes and HasComponent to check if we are below the ObjectsFolder */
    UA_ReferenceTypeSet reftypes =
        UA_ReferenceTypeSet_union(UA_REFTYPESET(UA_REFERENCETYPEINDEX_ORGANIZES),
                                  UA_REFTYPESET(UA_REFERENCETYPEINDEX_HASCOMPONENT));

## Checklist

Please provide the following information:

 - [ x ] open62541 Version (release number or git tag): 1.1.2 / ee275e79
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ x ] Operating system: Windows 10
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jpfr:
Fixed on master via #3861.

andreasebner:
CodePath is no longer relevant. The issue is now fixed.

PR #3861, opened 2020-08-27
fix(server): Add HasOrderedComponent to reftypes checked for events f…

jpfr:
…rom objects


ccvca:
In my opinion all sub types of HasComponent and Organized should be allowed there, including potential extensions of companion specifications:
grafik

Maybe the "includeSubtypes=true" code of the browse request can be used for catching all relevant references.

andreasebner:
I compared your work with the latest master. The change seems to be still valid. How should we proceed?

teccheck:
@jpfr there is an issue #3858. How should we proceed?

andreasebner:
The code path targeted by this issue no longer exists in recent versions.

PR #3866, opened 2020-08-31
[Review] Update ua-nodeset to git tag 'UA-1.04.7-2020-07-15'

ccvca:
Update to OPC UA Nodeset version to 1.4.7.
This adds among other things IOrderedObjectType and OrderedListType.

https://github.com/OPCFoundation/UA-Nodeset/tree/UA-1.04.7-2020-07-15


ccvca:
I'll keep trying to fix these CI errors...

ccvca:
Rebased on master, CI fixed.

This is required, as the new 1.4.7 nodeset contains an "Aliases"-Folder, which is only available for FULL namespace 0.

https://github.com/open62541/open62541/blob/d36e6873928852d67e570a9346841ea0ec121761/tests/server/check_server_userspace.c#L126-L131

jpfr:
Very good.

Just one nitpick: I would like to limit the complexity of the configuration options.
This only affects one of the tests.
Can you see another workaround instead of adding UA_GENERATED_NAMESPACE_ZERO_FULL?

ccvca:
> Can you see another workaround instead of adding UA_GENERATED_NAMESPACE_ZERO_FULL?

I do see two possibilities

1. Add the Aliases-Folder Node to the Reduced Namespace 0, so UA_GENERATED_NAMESPACE_ZERO can be used
2. Change the test, so Aliases is either considered always optional, or do not fail, if unknown nodes are found by browsing the Object-Folder

Both workarounds have their downside, but could be implemented quite simple.

ccvca:
@jpfr Do you see any other alternatives? Do you prefer one of the workarounds?

jpfr:
Nice! Merged.

PR #3869, opened 2020-09-02
Fix: Delete new notification

ccvca:
This enable deleting new notifications, that have no values assigned, e.g. 
```cpp
UA_Notification *pNotification = UA_Notification_new();
// Do nothing with the notification
UA_Notification_delte(pServer, pNotification);
```

This was the case here:
https://github.com/open62541/open62541/blob/03a655667e91cd5ccb8221d919fa2ac6d5a00698/src/server/ua_subscription_events.c#L486-L496


jpfr:
Thanks for this.

PR #3874, opened 2020-09-08
 UA_PubSubChannelMQTT_close: remove unused state check, set null after free

wgge:


andreasebner:
The cleanup procedure seems okay for me. Other PubSub networking layers also don't set the internal pointers to NULL during the cleanup. If the pointers are not set to NULL the user can decide if he likes to reuse it. The discussion maybe could be if the close function should really delete the structures or just close the connection and set the connection state to disconnect. But this is a further discussion. The removed checked is currently maybe unused but there are maybe future situations where this state can be used.

PR #3823, opened 2020-08-04
[Reopened without merge conflicts] feat(pubsub) allow subscribing to NetworkMessages with multiple DataSetMessages

jmachauer-codewerk:
Iterating over all DataSetMessages, searching for all DataSetReaders which correspond to each DSM via DataSetWriterId and applying the processing of the DSM to all DSRs.

PR #3880, opened 2020-09-09
feat(pubsub) allow subscribing to NetworkMessages with multiple DataSetMessages

jmachauer-codewerk:
My last PR #3823 had merge conflicts with some recent changes.

Iterating over all DataSetMessages, searching for all DataSetReaders which correspond to each DSM via DataSetWriterId and applying the processing of the DSM to all DSRs. Corresponding DataSetReaders are filter out when Ids mismatch, allowing for the processing of Networkmessages without DataSetWriterId or WriterGroupId. Where every DataSetReader gets a chance at decoding as long as it did not get filtered out before.


jmachauer-codewerk:
reopened #3887

PR #3887, opened 2020-09-14
feat(pubsub) allow subscribing to NetworkMessages with multiple DataSetMessages

jmachauer-codewerk:
getReadersFromIdentifier returns all fitting readers in an array together with it's size, so that for each datasetmessage we can iterate over the fitting readers and process the message for each of them. The filtering is also adjusted to reject when not matching, so that in the case of missing headers the ids that would have been contained in these headers are not relevant for filtering out readers.


andreasebner:
Thank you for your contribution back then. The feature has since been implemented and is now supported in the latest master and in version 1.5.

PR #3892, opened 2020-09-16
Avoid warnings for generated nodesets with large numeric node IDs

basyskom-jvoe:
Large numeric node IDs in nodesets may cause warnings like ```warning: this decimal constant is unsigned only in ISO C90``` when building the generated code.


jpfr:
Thanks!

PR #3898, opened 2020-09-18
[Review]: Extend support on custom pubsub manager callback mechanism

opcua-tsn-team-kalycito:
It allows the user to handle customized callbacks without any modification in the CMake option. This also allows users to configure multiple pubsub manager callbacks (eg: default timer callback and custom callback) in the same application.

Note: The user must take care of the repetitive calls of pubsub callback in the configured interval.


jpfr:
Looks really good.

Last request for the public API.

- Rename UA_PubSub_customManagerCallback to UA_PubSub_CallbackLifecycle. This is then consistent with the node constructor callback API.
- Remove the UA_CustomPublishCallback. It is identical to UA_ServerCallback. Use that instead.

Can be merged then.

opcua-tsn-team-kalycito:
@jpfr Thanks for the comments.

> Remove the UA_CustomPublishCallback. It is identical to UA_ServerCallback. Use that instead.

UA_ServerCallback declared in server.h file requires server_pubsub.h when pubsub is enabled. So accessing UA_ServerCallback from server_pubsub.h leads to a compilation error.

Declaring UA_ServerCallback in util.h resolves this issue. Is this approach fine for you?

jpfr:
The include order was fixed in #3971.
You can rebase on top of that. And reuse all definitions from server.h in server_pubsub.h.

opcua-tsn-team-kalycito:
Thanks, @jpfr.
Rebased and addressed the comments. Also added a test case on the custom callback usage.

jpfr:
Perfect PR.
A nice simplification overall.

PR #3899, opened 2020-09-18
PubSub component getState

MartinaFelberFEM:
Added a getState method for following PubSub components:
ReaderGroup, DataSetReader, WriterGroup, DataSetWriter

Open questions:

- Naming convention of PubSub public functions
e.g.: UA_Server_removeWriterGroup vs. UA_Server_updateWriterGroupConfig
or UA_Server_getDataSetWriterConfig vs. UA_Server_DataSetWriter_getState

- UA_Server_setWriterGroupOperational and UA_Server_setReaderGroupOperational
These functions set the linked DataSetWriters and -Readers state as well
But in most tutorials the DataSetWriters- and Readers have been added after setting the group to operational, so their state is still disabled. Calling setOperational() function again does not change this at the moment, because nothing happens if the state is already operational
Shall we change the setOperational() methods to set the states of all subcomponents again to operational, even if the group is already operational?
I think this could be a use case if someone wants to add a DataSetWriter- or Reader to an existing group at runtime.
Or shall we add a description that setOperational shall only be done after adding all DataSetWriters or -Readers?

- WriterGroup publishCallback error handling
In case of an error shall we stop the callback?

### DIFF
--- a/src/pubsub/ua_pubsub_reader.c
+++ b/src/pubsub/ua_pubsub_reader.c
@@ -469,6 +469,10 @@ UA_ReaderGroup_setPubSubState(UA_Server *server, UA_PubSubState state, UA_Reader
                 case UA_PUBSUBSTATE_PAUSED:
                     break;
                 case UA_PUBSUBSTATE_OPERATIONAL:
+                    UA_PubSubManager_removeRepeatedPubSubCallback(server, readerGroup->subscribeCallbackId);
+                    LIST_FOREACH(dataSetReader, &readerGroup->readers, listEntry){
+                        UA_DataSetReader_setPubSubState(server, UA_PUBSUBSTATE_ERROR, dataSetReader);
+                    }
                     break;
                 case UA_PUBSUBSTATE_ERROR:
                     return UA_STATUSCODE_GOOD;
@@ -476,6 +480,7 @@ UA_ReaderGroup_setPubSubState(UA_Server *server, UA_PubSubState state, UA_Reader
                     UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
                                    "Received unknown PubSub state!");
             }
+            readerGroup->state = UA_PUBSUBSTATE_ERROR;
             break;
         default:
             UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
@@ -788,15 +793,27 @@ UA_DataSetReader *UA_ReaderGroup_findDSRbyId(UA_Server *server, UA_NodeId identi
 void UA_ReaderGroup_subscribeCallback(UA_Server *server, UA_ReaderGroup *readerGroup) {
     UA_PubSubConnection *connection =
         UA_PubSubConnection_findConnectionbyId(server, readerGroup->linkedConnection);
+    if (!connection) {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER, "SubscribeCallback(): "
+            "Find linked connection failed");
+        UA_ReaderGroup_setPubSubState(server, UA_PUBSUBSTATE_ERROR, readerGroup);
+        return;
+    }
     UA_ByteString buffer;
-
     if(UA_ByteString_allocBuffer(&buffer, 4096) != UA_STATUSCODE_GOOD) {
-        UA_LOG_INFO(&server->config.logger, UA_LOGCATEGORY_SERVER, "Message buffer alloc failed!");
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER, "SubscribeCallback(): Message buffer alloc failed!");
+        UA_ReaderGroup_setPubSubState(server, UA_PUBSUBSTATE_ERROR, readerGroup);
         return;
     }
 
+    UA_StatusCode res = connection->channel->receive(connection->channel, &buffer, NULL, 1000);
+    if (UA_StatusCode_isBad(res)) {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER, "SubscribeCallback(): Connection receive failed!");
+        UA_ReaderGroup_setPubSubState(server, UA_PUBSUBSTATE_ERROR, readerGroup);
+        UA_ByteString_deleteMembers(&buffer);
+        return;
+    } 
     size_t currentPosition = 0;
-    connection->channel->receive(connection->channel, &buffer, NULL, 1000);
     if(buffer.length > 0) {
         if (readerGroup->config.rtLevel == UA_PUBSUB_RT_FIXED_SIZE) {
             do {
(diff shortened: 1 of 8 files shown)


MartinaFelberFEM:
Hi @andreasebner !

As discussed today I now 

- remove the publish and/or reader callback on an error

- Set the DataSetWriter or Reader state to operational at addDataSetWriter/Reader function, if the parent group is already operational

- I hope the operational state check in the publish and subscribeCallback is correct

src/pubsub/ua_pubsub_reader.c
```diff
@@ -374,6 +374,19 @@ UA_Server_ReaderGroup_getConfig(UA_Server *server, UA_NodeId readerGroupIdentifi
     return UA_STATUSCODE_GOOD;
 }
 
+UA_StatusCode
+UA_Server_ReaderGroup_getState(UA_Server *server, UA_NodeId readerGroupIdentifier,
+                               UA_PubSubState *state)
+{
+    if((server == 0) || (state == 0))
```
andreasebner, line 381: Please check for NULL all the time.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -788,15 +806,27 @@ UA_DataSetReader *UA_ReaderGroup_findDSRbyId(UA_Server *server, UA_NodeId identi
...
-
     if(UA_ByteString_allocBuffer(&buffer, 4096) != UA_STATUSCODE_GOOD) {
-        UA_LOG_INFO(&server->config.logger, UA_LOGCATEGORY_SERVER, "Message buffer alloc failed!");
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER, "SubscribeCallback(): Message buffer alloc failed!");
+        UA_ReaderGroup_setPubSubState(server, UA_PUBSUBSTATE_ERROR, readerGroup);
         return;
     }
 
+    UA_StatusCode res = connection->channel->receive(connection->channel, &buffer, NULL, 1000);
+    if ((res & 0x80000000) != 0) { // status is bad
```
andreasebner, line 823: It's not good to read. Why not just !UA_STATUSCODE_GOOD?

src/pubsub/ua_pubsub_writer.c
```diff
@@ -841,6 +841,18 @@ UA_Server_getDataSetWriterConfig(UA_Server *server, const UA_NodeId dsw,
     return retVal;
 }
 
+UA_StatusCode
+UA_Server_DataSetWriter_getState(UA_Server *server, UA_NodeId dataSetWriterIdentifier,
+                               UA_PubSubState *state) {
+    if((server == 0) || (state == 0))
```
andreasebner, line 847: Change for NULL. I will not mention this anymore.

tests/pubsub/check_pubsub_get_state.c
```diff
@@ -0,0 +1,437 @@
...
+    ck_assert_int_eq(UA_PUBSUBSTATE_OPERATIONAL, state);
+
+    ck_assert(UA_Server_setReaderGroupOperational(server, RGId_Conn1_RG1) == UA_STATUSCODE_GOOD);
+    ck_assert_int_eq(UA_STATUSCODE_GOOD, UA_Server_ReaderGroup_getState(server, RGId_Conn1_RG1, &state));
+    ck_assert_int_eq(UA_PUBSUBSTATE_OPERATIONAL, state);
+    ck_assert_int_eq(UA_STATUSCODE_GOOD, UA_Server_DataSetReader_getState(server, DSRId_Conn1_RG1_DSR1, &state));
+    ck_assert_int_eq(UA_PUBSUBSTATE_OPERATIONAL, state);
+
+    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND, "set groups disabled");
+    ck_assert(UA_Server_setWriterGroupDisabled(server, WGId_Conn1_WG1) == UA_STATUSCODE_GOOD);
```
andreasebner, line 319: In the long term we should think about changing our API. Now we have a getState, maybe it would be better if we change this function also to setState with the target state as argument (not for this PR).

andreasebner:
@opcua-tsn-team-kalycito  This PR adds some additional error/state checks (e.g. in in high frequency publish and subscribe functions). From my point of view, it looks okay, because additional effort (beneath the if conditions itself) is only added if something went wrong. Can you confirm if the publishCallback and subscribeCallback looks feasible for you?

src/pubsub/ua_pubsub_reader.c
MartinaFelberFEM: I think the receive function of the UDP plugin for instance returns UA_STATUSCODE_GOODNONCRITICALTIMEOUT and sets the message buffer length to 0 if there are no new packages. https://github.com/open62541/open62541/blob/9fab2726e510f9cab2123047a24b6c401a12a2a6/plugins/ua_pubsub_udp.c#L485

So the returned status code is not UA_STATUSCODE_GOOD, but UA_STATUSCODE_GOODNONCRITICALTIMEOUT. In this case I do not want to print a error message or set the state to error. 
Only if a bad statuscode gets returned.
That's why I have masked the status code, because I think that all Bad codes set their highest bit to '1'

opcua-tsn-team-kalycito:
@andreasebner It looks fine for me as well.

src/pubsub/ua_pubsub_reader.c
jpfr: I would avoid the bitfield magic.

Can you do an explicit comparison against UA_STATUSCODE_GOODNONCRITICALTIMEOUT?

Or we can add `UA_StatusCode_isBad` to make it explicit.

MartinaFelberFEM: Sure, I think I will implement the UA_StatusCode_isBad function, if that's ok.
Because the ua_pubsub_ethernet_etf.c receive implementation can return UA_STATUSCODE_GOODNODATA or UA_STATUSCODE_GOOD or UA_STATUSCODE_GOODNONCRITICALTIMEOUT and then the if() would look more complex

MartinaFelberFEM: Where may I implement it?
ua_util.c and .h?

MartinaFelberFEM: I have added it to ua_util

jpfr: You can add it to types.h / ua_types.c.
So a public fundtionality for UA_StatusCode.

MartinaFelberFEM: ok, thanks
moved it to types.c
I wasn't sure where I should add the test. 
I have added it to check_types_builtin.c

tests/pubsub/check_pubsub_get_state.c
MartinaFelberFEM: ok
may I resolve the conversation?

PR #3900, opened 2020-09-18
fix(nodeset_compiler): only set first char to lowercase of struct member

Pro:
E.g., EUInformation has struct members like `namespaceUri` which would become `namespaceuri` and therefore the generated code doesn't compile.


jpfr:
👍

PR #3926, opened 2020-10-02
Create demo-gui.py

tushar-kamra:


jpfr:
Hey there, looking good.
But this is something you should send to the Python-FreeOPCUA folks to include as an example.

PR #3936, opened 2020-10-05
[Review] Add explicit 0 initialization for enumerations 

mlang-de:
Add =0 to some enumeration definitions


mlang-de:
Please rerun CI after memory leak is fixed which was introduced by commit 4fdee9623f65deb5f09ad7e61afebac93b6ddf9e.

basyskom-meerkoetter:
@mlang-de is this for documentation purposes? C as well as C++ make sure that the first enumerator is 0 (if nothing is set by hand)

mlang-de:
You are right, Is only cosmetic. But in the master some enum declarations will be defined with explicit =0 and some not. My intention was to make it each time the same way to be consistent.

PR #3937, opened 2020-10-05
[Review] Remove trailing whitespaces

mlang-de:
Remove trailing whitespaces from all c-source and header files.


mlang-de:
Please rerun CI after memory leak is fixed which was introduced by commit 4fdee9623f65deb5f09ad7e61afebac93b6ddf9e.

jpfr:
Let’s do this in all public header files.
Not a big fan of mass-commits otherwise.

There are ways to enforce good whitespace practices on changes code. For example by a CI Test that enforces it.

Can you make this PR for public headers only?

mlang-de:
"Can you make this PR for public headers only?"  done by commit 472ce23
What is with the example header and/or deps folder headers? These are public headers too?

"There are ways to enforce good whitespace practices on future code changes. For example by a CI Test that enforces it."
I would appreciate it.

mlang-de:
AppVeyor was cancelled, not sure why. Please retrigger.

jpfr:
Resolved conflicts

PR #3939, opened 2020-10-06
[Review] (bug) missing TCP connection size handling could lead to connection error

mlang-de:
If a client does not complaint closes the connection for instance

- Open Socket
- Open SecureChannel + Open Session
- /* Do something irrelevant stuff here */
- Close Socket (no CloseSecureChannel and/or CloseSession)

Due to the not complaint behaviour the servers TCP layer socket error handling decrement the variable layer->connectionsSize. Since the layer->connectionsSize are not used to be incremented somewhere the layer->connectionsSize is 0 and by decrementing the value is 65535. After this situation no more connection establishment is possible due to the if statement in the ServerNetworkLayerTCP_add function because the check layer->connectionsSize >= layer->maxConnections will fail all time.

This error happens with a Ua Client in a Docker container. This container was shutdown, but the Client does not cleanup. So the Docker container only closes the sockets by itself.
For testing I use open62541 Client with some small hacks to close only the socket and not the securechannel or session.


jpfr:
Merged. 👍

PR #3940, opened 2020-10-07
fix(server): Fix starting server on IPv4-only systems

anzeskerjanc:
This pull request is solution for this comment: https://github.com/open62541/open62541/pull/3460#issuecomment-684856177

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -409,6 +409,9 @@ ServerNetworkLayerTCP_start(UA_ServerNetworkLayer *nl, const UA_Logger *logger,
     hints.ai_family = AF_UNSPEC;
     hints.ai_socktype = SOCK_STREAM;
     hints.ai_flags = AI_PASSIVE;
+#ifdef AI_ADDRCONFIG
+    hints.ai_flags |= AI_ADDRCONFIG;
+#endif
     hints.ai_protocol = IPPROTO_TCP;
     int retcode = UA_getaddrinfo(customHostname->length ? hostname : NULL,
                                  portno, &hints, &res);


basyskom-meerkoetter (commented):
can you please give some reasoning for this chance?

anzeskerjanc:
After https://github.com/open62541/open62541/pull/3460 pull request, I have problem running OpcUa server on system with SocketCAN. 

SocketCAN: is widely used in automotive and embedded world. For communication to user space it uses standard Linux socket interface. Here you can read more about this protocol: https://www.kernel.org/doc/Documentation/networking/can.txt

On start, OpcUa server tries to start listening on all network interfaces. But it fails on CAN interface, because it doesn't support this kind of protocol.
Here is my list of socket interfaces:
image

For me is OK to ignore such network interfaces. Basic problem in https://github.com/open62541/open62541/pull/3460 was detecting already used port. With this commit I don't break that functionality.

jpfr:
There should be two parts to the solution.

- Allow the user to explicitly select devices.
- Log the error but continue starting the server when a socket cannot be opened.

If the return value of addServerSocket should be ignored, then the return value should be void and not a StatusCode.

jpfr:
Logging could also be better.
Can we get the device name and ipv4/ipv6 information?

anzeskerjanc:
I figured out that this is not SocketCAN problem. 

My device (AArch64) doesn't have IPv6 stack.
But for some reason glibc::getaddrinfo returns:
-ai_family: 2, ai_socktype: 1, name: 0.0.0.0
-ai_family: 10, ai_socktype: 1, name: ::

The solution for me is to add AI_ADDRCONFIG in ai_flags (like https://github.com/warmcat/libwebsockets/blob/ed1e558eedc79a07837db2f0a41eb94197ad23fa/lib/core-net/adopt.c#L791).

I committed with wrong username. :( 
Later I can create clean pull request, if we figured out this is OK for all.

jpfr:
Hey @anzeskerjanc , you still want to finish this?

anzeskerjanc:
Hi,

off course I want to finish this.

I didn't receive any feedback, if the solution suits everyone.
After this I can clean up history (keep only last commit) and change commit author.

jpfr:
The change is good.
Please rebase.

and leave a short comment on the meaning of the new flag:

https://man7.org/linux/man-pages/man3/getaddrinfo.3.html

anzeskerjanc:
I have rebased branch and changed commit message. 
Is it okay now and what's wrong with CI?

anzeskerjanc:
OK, all checks have passed.

About AI_ADDRCONFIG flag: 
UA_getaddrinfo will return IPv4 or IPv6 address only if system has at least one configured. It's useful on IPv4-only systems, where UA_socket on IPv6 would always fail ("Error opening the server socket").

jpfr:
Thanks for getting to the bottom of this!
Merged.

PR #3943, opened 2020-10-07
CLIENT : Fix log level in initConnect

StalderT:


jpfr:
How is this a fix?
We should inform the user that he should be using the library differently.

StalderT:
In sample we have :

https://github.com/open62541/open62541/blob/master/examples/client_connect_loop.c#L44-L47

If we start client_connect_loop.c with WARNING LOGS enabled we have the log on each cycle. The log is displayed endlessly.

jpfr:
Run _connect only once and _iterate in a loop. This could be recommended in the warning.

StalderT:
If we lose connection in the loop, _connect will automatically attempt to reestablish communication. The client has nothing more to manage. In my opinion this is really a plus that we cannot ignore. IMHO it would be good to fix this before the release of version 1.2.

PR #3948, opened 2020-10-07
fix(debian): a tools script depends on python3 so package must as well

jrcoding:
`tools/certs/create_self-signed.py` depends on `python3` so the package `libopen62541-1-tools` must depend on `python3 `as well


jpfr:
Thanks!

PR #3965, opened 2020-10-20
[Review] Realtime pubsub xml example

opcua-tsn-team-kalycito:
[WIP] On the top of PR #3898 , added pubsub realtime example with XML data model


examples/pubsub_realtime/nodeset/pubsub_nodeset_rt_publisher.c
```diff
@@ -0,0 +1,681 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+#define _GNU_SOURCE
+
+/* For thread operations */
+#include <pthread.h>
```
andreasebner, line 7: I like this extended example. Since the subscriber and publisher tutorials are quite large and complicated, it would be nice to add a comment block regarding the general purpose, functions, parameters and used files of this examples.

PR #3967, opened 2020-10-21
Update ua-nodeset

ccvca:
Update ua-nodeset to the latest specifications.
New CS: 

- Industrial Automation
- Machinery
- Machine Tool


jpfr:
Thanks

PR #3976, opened 2020-10-27
Fix PEM certificate handling for SecurityPolicy None

basyskom-jvoe:
If UA_ENABLE_ENCRYPTION_MBEDTLS is set, SecurityPolicy None will still just copy the certificate data.
If the input is PEM, the trailing null byte is not added and mbedTLS will fail to parse the data. This breaks the application URI check on server start.

In case of UA_ENABLE_ENCRYPTION_OPENSSL, there is a similar problem
if the PEM data is just copied into the local certificate byte string.


jpfr:
Thanks 👍

PR #3981, opened 2020-10-29
[WIP] PubSub: Publishing offset parameter support

opcua-tsn-team-kalycito:
- Support only to the ethernet transport layer (other layer developments are in progress)


include/open62541/server.h
```diff
@@ -353,7 +353,7 @@ UA_Server_run_shutdown(UA_Server *server);
 /**
  * Timed Callbacks
  * --------------- */
-typedef void (*UA_ServerCallback)(UA_Server *server, void *data);
+typedef void (*UA_ServerCallback)(UA_Server *server, UA_DateTime callbackTime, void *data);
```
andreasebner, line 356: Hey @jpfr , during the offset implementation kalycito started to modify the central 'UA_ServerCallback' signature and added a parameter containing the callback execution time. In the latest implementation approach this is needed to schedule the delayed send out of network packages https://github.com/open62541/open62541/blob/ea1622e1b0fbd3efd1f2e95758b399447c5a25ab/src/pubsub/ua_pubsub_writer.c#L2046-L2050 . From my point of view, the modification will be difficult to merge because a lot of exiting code breaks. @opcua-tsn-team-kalycito , why can’t we get the current ‘UA_DateTime_nowMonotonic’ in the first lines of the publishCallack? If you need additional fields, you can also create a container surrounding the current callback input with additional fields. The current signature contains a void pointer and therefore we can pass in everything. @jpfr can you bring in your opinion to this discussion?

opcua-tsn-team-kalycito: Hi @andreasebner 

>why can’t we get the current ‘UA_DateTime_nowMonotonic’ in the first lines of the publishCallack?

We thought about this but while sending the packets at the sub-micro-second level precision at the configured publishingOffset using ETF, it may cause inaccuracy in publish time due to the wakeup latency of the system/threads.

> If you need additional fields, you can also create a container surrounding the current callback input with additional fields. The current signature contains a void pointer and therefore we can pass in everything.

From the Timer_process() API, we require nextTime for the calculation of publishingTime. When using a container structure with DateTime and WriterGroup, we tend to expose this structure in timer.c and pass nextTime value to DateTime. This may cause every void pointer usage to that container structure which also required lots of core modification.

opcua-tsn-team-kalycito:
@jpfr Please have a look at the implementation and let us know your comments.

PR #3986, opened 2020-11-02
fix(pubsub): Segmentation fault while enabling csv logs in realtime TSN samples

opcua-tsn-team-kalycito:
Modified to stop the application when maximum logging reached


jpfr:
Looking good. 👍

PR #4034, opened 2020-11-26
[Review] Shared libs support to subscriber samples

opcua-tsn-team-kalycito:
Support to subscriber samples to build when the shared library is enabled
 - External API usage of regist has been removed
 - Used regist in addReaderGroup API


src/pubsub/ua_pubsub_reader.c
```diff
@@ -265,6 +265,23 @@ UA_DataSetReader_generateNetworkMessage(UA_PubSubConnection *pubSubConnection, U
...
     return UA_STATUSCODE_GOOD;
 }
+
+UA_StatusCode
+UA_PubSubConnection_regist(UA_Server *server, UA_NodeId *connectionIdentifier) {
+    UA_PubSubConnection *connection =
+        UA_PubSubConnection_findConnectionbyId(server, *connectionIdentifier);
+    if(!connection)
+        return UA_STATUSCODE_BADNOTFOUND;
+
```
jpfr, line 275: Add a check here if the connection is already registered.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -291,6 +308,13 @@ UA_Server_addReaderGroup(UA_Server *server, UA_NodeId connectionIdentifier,
         return UA_STATUSCODE_BADCONFIGURATIONERROR;
     }
 
+    /* Regist (bind) the connection channel if it is not already registered */
```
jpfr, line 311: I a missing the same check for the WriterGroup.
Shouldn't that be handled in a symmetric way?

jpfr (commented):
Looks good in general.

Some inline comments.

How do we handle loss-of-connection at the moment?
Ideally this would lead to the connection being "unregistered" so it will be automatically re-registered.

src/pubsub/ua_pubsub_manager.c
```diff
@@ -116,6 +116,27 @@ UA_Server_removePubSubConnection(UA_Server *server, const UA_NodeId connection)
...
+
+    UA_StatusCode retval = connection->channel->regist(connection->channel, NULL, NULL);
+    if(retval != UA_STATUSCODE_GOOD)
+        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                       "register channel failed: 0x%" PRIx32 "!", retval);
+
+    connection->isRegistered = UA_TRUE;
+    return retval;
+}
+
```
andreasebner, line 139: Why is this function needed? Is there a difference of using the regist function within the PubSubChannel (which is contained in the PubSubConnection)?

andreasebner:
I looked over it and it looks good overall. I added one questen regarding the regist function. All in all this PR can be merged.

PR #4045, opened 2020-12-03
Bugfix for gcc-10 and ETF examples

mwalle:
Fix warnings/errors reported by GCC-10. Also fix the RT publisher/subscriber example


jpfr:
Rebased on 1.2

PR #4050, opened 2020-12-04
fix(plugin): use of uninitialized value

MartinaFelberFEM:
UA_PubSubChannelEthernet_receive() returns an uninitialized value
(retval), if  there is more data available than fits into the
message buffer.

Control flow in case of the error:

Receive buffer size is 4096 - see UA_ReaderGroup_subscribeCallback.
timeout > 0
select resultsize > 0
recvmsg datalen > 0
remainingMessageLength >= 1496 -> false -> there is still more data


jpfr:
Thanks!

PR #4056, opened 2020-12-09
Fix: Build shared library with discovery multicast

ccvca:
Fixes building shared library with multicast discovery.

Before this fix, `mdnsd_config.h` looks like this:
```c
#  ifndef MDNSD_EXPORT
#    ifdef libmdnsd_EXPORTS
        /* We are building this library */
#      define MDNSD_EXPORT __declspec(dllexport)
#    else
        /* We are using this library */
#      define MDNSD_EXPORT __declspec(dllimport)
#    endif
#  endif
```

Now it is the same as building mdns:
```c
#  ifndef MDNSD_EXPORT
#    ifdef MDNSD_DYNAMIC_LINKING_EXPORT
        /* We are building this library */
#      define MDNSD_EXPORT __declspec(dllexport)
#    else
        /* We are using this library */
#      define MDNSD_EXPORT __declspec(dllimport)
#    endif
#  endif
```
So these lines have the desired effect:
https://github.com/open62541/open62541/blob/19c006a330b110d9df4e71f680d2ac90eeedaf95/CMakeLists.txt#L1284-L1287

Added/updated copyright for these past commits by me:
40cb2f0e46ccdb04ecfe4fcf5a133257aeed3ab1
0c10b3ea408741abcb339bf53e9975969ae09c01
a524accc338c9611a5ec60bf595c1a587243b457
1e29c3329a0e4ef649b859da31b0562849277c7f
3598c1677e301de0f48c83b3948a1955aa5c8492

PS: Is it intended that Python and CMake files have no license header?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -688,6 +688,7 @@ if(UA_ENABLE_DISCOVERY_MULTICAST)
     # create a "fake" empty library to generate the export header macros
     add_library(libmdnsd ${PROJECT_SOURCE_DIR}/deps/mdnsd/libmdnsd/mdnsd.h)
     set_property(TARGET libmdnsd PROPERTY LINKER_LANGUAGE CXX)
+    set_property(TARGET libmdnsd PROPERTY DEFINE_SYMBOL "MDNSD_DYNAMIC_LINKING_EXPORT")
     configure_file("deps/mdnsd/libmdnsd/mdnsd_config_extra.in"
                    "${PROJECT_BINARY_DIR}/src_generated/mdnsd_config_extra")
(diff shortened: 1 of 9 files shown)


jpfr:
Thanks for this. 👍 
We have a temporary issue in the CI that is already worked on.
We'll merge this when we can.

ccvca:
If a rebase is required, just write a short comment.

ccvca:
Rebased on 1.2 so the  already merged commit (19c006a) is no longer included.

jpfr:
Merged. For the future, please use git commit titles according to the contribution document.
Thanks for the fix!

PR #4064, opened 2020-12-11
doc(pubsub): Add documentation for PubSub examples

opcua-tsn-team-kalycito:
- Tutorial Subscribe example
 - PubSub TSN examples
 - PubSub nodeset realtime examples


opcua-tsn-team-kalycito:
Incorporated it in #3996.

PR #4067, opened 2020-12-11
refactor(ex): remove UA_ENABLE_AMALGAMATION check

hthiery:
Since a check in the toplevel CMakeLists.txt prevents building of the
examples when UA_ENABLE_AMALGAMATION is enabled the checks in the
example/CMakeLists.txt can be removed.

Signed-off-by: Heiko Thiery <heiko.thiery@gmail.com>


jpfr:
We have disabled building examples globally when the amalgamation is enabled.
Because the use the "individual files" API and not "open62541.h".
So this does not apply...

Please reopen if required.

PR #4071, opened 2020-12-15
Add UA_Client_modifyAsyncCallback

ccvca:
This pull request adds a new function `UA_Client_modifyAsyncCallback`, which can modify the callback of an ongoing async request. 
So a user can modify a callback, when the original callback/ userdata is no longer valid.

In my use case I have an instance, which issues a method call on construction. When the instance is destroyed, before the request is completed, the callback/userdata-Information is no longer valid (userdata is the this-pointer of the instance). With this method, I do set the async callback and userdata to NULL in the destructor, so the response is ignored.

### DIFF
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -15,6 +15,7 @@
  *    Copyright 2016 (c) Lykurg
  *    Copyright 2017 (c) Mark Giraud, Fraunhofer IOSB
  *    Copyright 2018 (c) Kalycito Infotech Private Limited
+ *    Copyright 2020 (c) Christian von Arnim, ISW University of Stuttgart
  */
 
 #include <open62541/types_generated_encoding_binary.h>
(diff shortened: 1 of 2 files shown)


src/client/ua_client.c
```diff
@@ -514,7 +529,7 @@ __UA_Client_AsyncServiceEx(UA_Client *client, const void *request,
     if(client->channel.state != UA_SECURECHANNELSTATE_OPEN) {
         UA_LOG_INFO(&client->config.logger, UA_LOGCATEGORY_CLIENT,
                     "SecureChannel must be connected before sending requests");
-		return UA_STATUSCODE_BADSERVERNOTCONNECTED;
+        return UA_STATUSCODE_BADSERVERNOTCONNECTED;
```
ccvca, line 532: This is just a tab to spaces conversion.

jpfr:
Looks good.
Ready to merge once the Ci fix was ported from 1.2 to master.

Issue #4054, opened 2020-12-07
nodeset compiller: NodeSets without data types cause compile error


ccvca:
## Description
The nodeset compiler creates invalid code, when there are no Datatypes defined in a NodeSet. (e.g. Machinery)

The compiler complains about missing UA_TYPES_MACHINERY, which is not generated, as it is empty (Fixed here: bb8ce4be345b23ab84606e91f05be06d446cfb6e):
```
/home/runner/work/Sample-Server/Sample-Server/Sample-Server/src_generated/namespace_machinery_generated.c:1199:5: error: ‘UA_TYPES_MACHINERY’ undeclared here (not in a function); did you mean ‘customUA_TYPES_MACHINERY’?
 1199 |     UA_TYPES_MACHINERY
      |     ^~~~~~~~~~~~~~~~~~
      |     customUA_TYPES_MACHINERY
```

This was introduced here:
https://github.com/open62541/open62541/commit/c706a2d8dc0f0e53109123bafba9634f4398a24d

I do not known how to solve this, as at the correspoinding code part, there is no information available, if there are types or not.

## Background Information / Reproduction Steps

Compile Machinery with an empty bsd:
```xml
<opc:TypeDictionary xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:tns="http://opcfoundation.org/UA/Machinery/" DefaultByteOrder="LittleEndian" xmlns:opc="http://opcfoundation.org/BinarySchema/" xmlns:ua="http://opcfoundation.org/UA/" TargetNamespace="http://opcfoundation.org/UA/Machinery/">
    <opc:Import Namespace="http://opcfoundation.org/UA/"/>
</opc:TypeDictionary>
```

```cmake
ua_generate_nodeset_and_datatypes(
  NAME "machinery"
  FILE_NS "${open62541_NODESET_DIR}/Machinery/Opc.Ua.Machinery.NodeSet2.xml"
  FILE_CSV "${open62541_NODESET_DIR}/Machinery/NodeIds.csv"
  FILE_BSD "${CMAKE_BINARY_DIR}/model/Opc.Ua.Machinery.Types.bsd"
  OUTPUT_DIR "${PROJECT_SOURCE_DIR}/src_generated"
  NAMESPACE_MAP "3:http://opcfoundation.org/UA/Machinery/"
  DEPENDS di
  )
```

Used CMake options:

```bash
[...]

PR #4078, opened 2020-12-21
Handle nodesets without datatypes (#4054)

ccvca:
This handles nodesets without datatypes (e.g. machinery, #4054). So this code does no longer causes errors:
```c
static UA_DataTypeArray customUA_TYPES_MACHINERY = {
    NULL,
    UA_TYPES_MACHINERY_COUNT,
    UA_TYPES_MACHINERY
};
```

I've also noticed (but not modified), that the types are added multiple times, if nodesets depend on others, e.g. (Industrial Automation depends on DI):

- Calling namespace_di_generated add UA_TYPES_DI
- Calling namespace_ia_generated add UA_TYPES_IA and UA_TYPES_DI (again)

Is this intended? As `namespace_di_generated` must be called before `namespace_ia_generated` this seems to be redundant.

The `if(UA_TYPES_MACHINERY_COUNT)` is required, as the corresponding python-Variable is only available in `backend_open62541_typedefinitions.py` and not in `backend_open62541.py`

PS: Is using a template engine like jinja (https://jinja.palletsprojects.com/en/2.11.x/) an option for the nodeset generation? It would require developers to install a library, but the nodeset generation code shoule be easier.

### DIFF
--- a/tools/nodeset_compiler/backend_open62541.py
+++ b/tools/nodeset_compiler/backend_open62541.py
@@ -309,8 +309,10 @@ UA_StatusCode retVal = UA_STATUSCODE_GOOD;""" % (outfilebase))
     for arr in set(typesArray):
         if arr == "UA_TYPES":
             continue
+        writec("if(" + arr + "_COUNT > 0) {")
         writec("custom" + arr + ".next = UA_Server_getConfig(server)->customDataTypes;")
         writec("UA_Server_getConfig(server)->customDataTypes = &custom" + arr + ";\n")
+        writec("}")
 
     if functionNumber > 0:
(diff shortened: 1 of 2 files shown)


jpfr:
Changed the base to 1.2 to include it in the next release.
That seems to make have made things worse...

ccvca:
Shall I do a cherry-pick/ force-push with 1.2 as base?

jpfr:
Yes. Exactly.

ccvca:
Ok, rebase done. CI had some unrealated apt-issues.

jpfr:
Cool

PR #4105, opened 2021-01-05
[Review] Send ethernet packets with ethertype 0xb62c in PubSub

opcua-tsn-team-kalycito:
- Changed ethertype from LLC to ETHERTYPE_UADP(0xb62c) according to PubSub Specification
 - Handled the padding bytes in the src file

### DIFF
--- a/plugins/ua_pubsub_ethernet.c
+++ b/plugins/ua_pubsub_ethernet.c
@@ -25,6 +25,9 @@
 #include <linux/version.h>
 #endif
 #include "time.h"
+#define ETHERTYPE_UADP                       0xb62c
+#define MIN_ETHERNET_PACKET_SIZE_WITHOUT_FCS 60
+#define VLAN_HEADER_SIZE                     4
 
 /* Ethernet network layer specific internal data */
 typedef struct {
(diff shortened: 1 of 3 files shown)


jpfr:
There are two comments /* TODO: Need to handle padding bytes for UDP */
Please clarify the status.
Does the current code break for udp?

opcua-tsn-team-kalycito:
No, it will not break the current UDP code. For now, UDP packets are never padded by the hardware as the minimum packet size exceeds 60 bytes. 

OPC UA UDP packet size: 
Default minimum header size: MAC header(14 bytes) + IP header(20 bytes) + UDP header(8 bytes)
Network message header size: 31 bytes - Current implementation with all pubId, DSWId and WGId enabled 
Total: 73 bytes (padding will not happen if total data is above 60 bytes)

In the future, OPC UA NetworkMessageHeader size can have minimum size (for eg: keepAlive message) less than 18 bytes which introduces padding while transmission. When this happens, multiple receiving of the messages feature is not handled.

rexsum:
This patch only works when specifying a VLAN otherwise the header still contains a protocol type of LLC. It should not be necessary to specify a VLAN to get the proper Ethernet header as required by Part 14 section 7.3.3 of the OPC UA specification.

opcua-tsn-team-kalycito:
@rexsum Thanks for pointing out the issue. It is resolved now.

PR #4111, opened 2021-01-07
Fix(Server): Check EncryptionAlgorithm is empty

rdyhms:
According to OPCUA spec 1.04 Part 4 page 162 the
EncryptionAlgorithm field in the UserIdentityToken shall be
"No encryption" if the SecurityPolicy is "None".
Currently the stack does not care about this field if the
SecurityPolicy is "None".

This triggers an error in OPC UA CTT Version 1.4.9.396
Security - Security User Name Password - Test Case 010.
In this test the EncryptionAlgorithm in the
UserIdentityToken is set to "abc123" and the test expects
the login attempt to fail.


jpfr:
Looks good!
@rdyhms I rebased on top of 1.2 to include the fix in the next stable release already.

PR #4114, opened 2021-01-08
CMakeLists.txt: fix build without C++

ffontaine:
This will fix the following build failure:

```
CMake Error at CMakeLists.txt:2 (project):
  The CMAKE_CXX_COMPILER:

    /srv/storage/autobuild/run/instance-3/output-1/host/bin/arm-linux-g++

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

Fixes:
 - http://autobuild.buildroot.org/results/86ca6a5a01ecfc7030c6be0da81924436b41d057

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.0...3.12)
-project(open62541)
+project(open62541 C)
 # set(CMAKE_VERBOSE_MAKEFILE ON)
 if(${CMAKE_VERSION} VERSION_LESS 3.12)
     cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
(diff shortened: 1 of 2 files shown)


jpfr:
That's a good improvement.

I see an error on Github Actions (based on a recentish Ubuntu)
> CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_CXX_CREATE_STATIC_LIBRARY

Once the CI is green we can merge.

ffontaine:
The build failure was raised because I didn't tested with `UA_ENABLE_DISCOVERY_MULTICAST`, indeed `libmdnsd` was unconditionally linked as a `CXX` library. With the new PR, the language will change depending on `UA_FORCE_CPP` however as it seems this is a "fake" library perhaps it would be better to always link it as a `C` library?

ffontaine:
Do you have additional comments on this PR?

jpfr:

PR #4120, opened 2021-01-13
Add ReaderGroup KeyValuePair feature

rexsum:
Provide a new feature that provides KeyValuePair properties for a ReaderGroup that mirrors the way the WriteGroup KeyValuePair properties function.

### DIFF
--- a/include/open62541/server_pubsub.h
+++ b/include/open62541/server_pubsub.h
@@ -702,6 +702,8 @@ typedef struct {
     UA_Boolean enableBlockingSocket; // To enable or disable blocking socket option
     UA_UInt32 timeout; // Timeout for receive to wait for the packets
     UA_PubSubRTLevel rtLevel;
+    size_t groupPropertiesSize;
+    UA_KeyValuePair *groupProperties;
 } UA_ReaderGroupConfig;
 
 void UA_EXPORT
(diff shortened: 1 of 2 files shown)


jpfr:
Looks good.
This only adds the configuration option.
I am missing a new feature actually using that new config option.

rexsum:
Perhaps I was mistaken in calling this a "new feature", it's simply a new configuration option that's need by my application. I've been patching the library for several months now and finally decided that I would try submitting my patch to the community. To clarify, are you asking me to provide a test or sample code that used this new configuration option, or do you simply want me to change the title? If you need some sample code that uses this config option, I can work on producing that.

jpfr:
At this time, this is a "dead" configuration option.
That begs the question why we need the added complexity.
I would only merge if the config option is actually useful to users in its current form.

I see two possible ways:
- Add a small feature that uses the config option and document it.
- If the config option is only forwarded to user-defined plugins, document the intended us in the public header.

rexsum:
I have been using this option for months in my own application. It's important for my app to be able to pass parameters into a ReaderGroup so that those values are available to the Subscriber callback function. It functions exactly like the WriteGroup KeyValuePair (from which I basically copied the code). My app is fully multi-threaded and all functions must be fully reentrant so global variables are a no-go. Anyway, I thought that i would share my work and make it available to all since it seemed to me to be a useful missing/overlooked capability. I didn't realize that the bar was so high for submitting a simple code enhancement, I will withdraw this pull request since this is turning into something bigger than I care to deal with.

PR #4130, opened 2021-01-18
Introduction of RAW encoding for PubSub

andreasebner:
The most efficient encoding format for the PubSub based communication is called "RAW Encoding". The values are transmittet without additional meta informations like the type, status or timestamps. The disadvantage is the need for the FieldMetaData to decode the incoming messages. Therefore this PR addes an additional behavoir which allows the subscriber to collect the FieldMetaData from a given OPC UA Server using Client/Server communication. 

The RAW Encoding has no dedicated chapter in Part 14 and is partly introduced in 6.2.3.2. Especialy for time critical applications the RAW encoded formats the RAW encoding allows smaler frames and more throughput. 

This PR is currently WIP and some features are not finished e.g. the use of RAW encoding together with our open62541-RT-fastpath concept.

The PubSub Raw encding was implemented by Fraunhofer IOSB  
on behalf of the Siemens AG.


mlang-de:
Will this feature not be part of v1.2? May at least as experimental state?

jpfr:
1.2 is after the feature-freeze, so we won't have it officially in 1.2.

mlang-de:
Thank you for clarification!

PR #4132, opened 2021-01-20
Add support for method calls with deferred completion.

dapaulid:
This change introduces the 'deferred' bit for method nodes. 

'deferred' method nodes are similar to 'async' methods, with the following difference:
Here, the method is called immediately (by the server thread instead of a worker), but its result is not (yet) sent to the client. Instead, the method can access its pending operation using UA_Server_getLastAsyncResponse and pass ("defer") it to an arbitrary thread or event handler. Eventually, this thread or event handler can call UA_Server_sendAsyncResponse to send the final result to the client. 

The advantage over the 'async' method call is that client code does not need to block/wait in their UA_MethodCallback implementation until the result is ready, while still maintaining only a constant number of worker threads.

Adds the following new API functions:
- UA_Server_setMethodNodeDeferred
- UA_Server_getLastAsyncResponse
- UA_Server_sendAsyncResponse

### DIFF
--- a/include/open62541/plugin/nodestore.h
+++ b/include/open62541/plugin/nodestore.h
@@ -612,6 +612,8 @@ typedef struct {
     UA_MethodCallback method;
 #if UA_MULTITHREADING >= 100
     UA_Boolean async; /* Indicates an async method call */
+    UA_Boolean deferred; /* Indicates a method call with deferred completion, i.e.
+                            started by server thread but completed by some other thread */
 #endif
 } UA_MethodNode;
 
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -1586,6 +1586,16 @@ UA_Server_getNamespaceByName(UA_Server *server, const UA_String namespaceUri,
 * ready. See the examples in ``/examples/tutorial_server_method_async.c`` for
 * the usage.
 *
+* Marking a method as deferred has a similar but slightly different effect:
+* Here, the method is called immediately, but its result is not (yet) sent to the client.
+* Instead, the method can access its pending operation using
+* UA_Server_getLastAsyncResponse and pass ("defer") it to an arbitrary thread
+* or event handler. Eventually, this thread or event handler can call
+* UA_Server_sendAsyncResponse to send the final result to the client.
+* The advantage over the 'async' method call is that client code does not need
+* to block/wait in their UA_MethodCallback implementation until the result is
+* ready, while still maintaining only a constant number of worker threads.
+*
 * Note that the operation can time out (see the asyncOperationTimeout setting in
 * the server config) also when it has been retrieved by the worker. */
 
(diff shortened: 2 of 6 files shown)


jpfr:
This is a very good initative. 👍 
The code looks high-quality as well.

Now a few wishes from my side:

I would prefer if the existing infrastructure is extended instead of duplicated.
You can just extend `UA_AsyncOperationType`, `UA_AsyncOperationRequest` and `UA_AsyncOperationResponse`.
Then you don't have to introduce a duplicate queue+lock+get/set functions.

The term *deferred* is great. But let's use both *async* and *deferred*.
I would prefer *async* so we stay consistent and don't break the existing API.

jpfr:
Humm. On a second reading of the code I have a few more questions.

It looks as if the deferred behaviour could be replicated within the existing *async* API.
Because the "worker" thread in the async API can also create further threads.
Both are a way to forward execution of a method to a separate thread.

Can you elaborate on a usage scenario that is not possible with the existing API....

dapaulid:
Thanks for your feedback!

> Can you elaborate on a usage scenario that is not possible with the existing API....

In my `UA_MethodCallback` function I want to start some long-lasting process (e.g. moving an actuator to a given end position). After the process completed, I want to inform the client about the result (e.g. OK or fail). Therefore, I need to access the input parameters to start the process, but I'm not able to set the output parameters when returning from my callback unless I wait for the process handled by some system thread to complete (which I want to avoid, see further below).

Now with the existing API it is possible to send the response at some later time using `UA_Server_setAsyncOperationResult`, but I need to pass it the `UA_CallMethodResult`. The problem is that this object is only returned by `UA_Server_call`, meaning that it is not yet accessible in my `UA_MethodCallback`. Like this, I have no way to pass the object along to the system thread when starting the process. Passing the object somehow along after the `UA_Server_call` returns would be tricky and prone to race conditions.

To solve this problem, my change introduces the `UA_Server_getLastAsyncResponse` so that the `UA_MethodCallback` has the chance to access and pass along its `UA_AsyncResponse` (that contains the `UA_CallMethodResult`)

> Because the "worker" thread in the async API can also create further threads.

This would mean that for each concurrent async method call, a separate thread is required that needs to wait in the `UA_MethodCallback` until the result is available. In our case, this is bad, as we have potentially many concurrent long-lasting processes (many actuators in the system), but we want to use only a constant number of threads, as they're a valuable resource in our context (custom real-time operating system).

> I would prefer if the existing infrastructure is extended instead of duplicated.

Actually, this was exactly my intention 🙂
What do you consider duplication? From my point of view, the only thing I copied is for the `UA_Server_setMethodNodeDeferred` setter, to introduce the 'deferred' bit.

>You can just extend UA_AsyncOperationType, UA_AsyncOperationRequest and UA_AsyncOperationResponse.
Then you don't have to introduce a duplicate queue+lock+get/set functions.

Not sure if I understand correctly. I didn't introduce a new queue, just refactored the existing `asyncResponses` queue to have an own type, so that I can use the `TAILQ_LAST` macro for it. The lock was necessary because the `asyncResponses` queue can now be accessed by different threads (server thread that creates the `UA_AsyncResponse` and the thread that eventually calls `UA_AsyncManager_sendAsyncResponse`)

> The term deferred is great. But let's not use both async and deferred.
> I would prefer async so we stay consistent and don't break the existing API.

You mean there shouldn't be a 'deferred' bit at all, i.e. merge the two functionalities into one?

As I see it, having a dedicated 'deferred' bit spares programmers the necessity to implement a dedicated worker thread doing the `UA_Server_getAsyncOperationNonBlocking` / `UA_Server_call` loop, as the `UA_MethodCallback` is already called by the server thread, just as in the synchronous case. Additionally, having a separate bit maintains compatibility with the existing async API, as for clients that already use a dedicated worker thread.

dapaulid:
Sorry if my last post was confusing, it turned out longer than I expected 🙂

To clarify, could you please point out which changes (API wise etc.) you want me to rework?

jpfr:
This is handled by the factored asynchronous api in v1.5.
This was a good inspiration.

PR #4136, opened 2021-01-21
[REVIEW] fix(server): check decoded extension object's type for compatibility

robertns:
Check if decoded extension object's type is compatible with the target type. This solves problem with variable write when using Extension object, as well as for method arguments which are structures.


jpfr:
It is good to be able to load structure-datatypes from the generated information model.

But this is mostly a workaround to some quirks of the nodeset compiler.

Can you add a TODO note that we should clean up the nodeset compiler to direcly include structures into the value-variant.
And afterwards clean up this workaround as well.

robertns:
@jpfr Thanks for the response, I pushed the changes (added TODO note). However, continuous-integration/travis-ci check doesn't seem to be working, it is "Expected" each time. Can you please check if there is some general problem (I've seen it on other pull requests as well), or should it be triggered in some way? Thanks.

jpfr:
The cleanup to the nodeset compiler is imminent.
Then we can go to the final version instead of a temporary workaround.

jpfr:
No longer needed after the update to the nodeset compiler.
Still, thanks for the hard work @robertns !

PR #4138, opened 2021-01-23
[WIP] Provide callback for AccessLevel attribute

matkonnerth:
In some situations the write access to a variable node is not possible, e.g. the backend is not connected or device is in a state where change of a certain variable is not possible. For such use cases a callback could be provided, which calculates the accessLevel.

Todos:
add tests
same callback should be possible for the executable flag for methods
Move accessLevel and Callback in an union type

### DIFF
--- a/examples/tutorial_server_variable.c
+++ b/examples/tutorial_server_variable.c
@@ -42,6 +42,40 @@ addVariable(UA_Server *server) {
                               UA_NODEID_NUMERIC(0, UA_NS0ID_BASEDATAVARIABLETYPE), attr, NULL, NULL);
 }
 
+static UA_StatusCode myAccessLevelCallback(UA_Server *server,
+                                                const UA_NodeId *sessionId,
+                                                void *sessionContext,
+                                                const UA_NodeId *nodeId,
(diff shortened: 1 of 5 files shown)


jpfr:
This is a good feature.
Please indicate the relation of this feature with `getUserAccessLevel` in the AccessControl plugin.
Furthermore, there are probably many other cases where we want to override the nodestore.
I would prefer a solution that works for all attributes and not only for AccessLevel.
For example with a bitmask that indicates which attribute is "overridden".

https://github.com/open62541/open62541/blob/8f4f99d3f795738cfb2bd363702021f4503bceee/include/open62541/plugin/accesscontrol.h#L60

matkonnerth:
Makes sense for me, I've changed the callback for reading a generic attribute value. I haven't considered writing an attribute up to now. If this is really necessary, the callback for reading/writing attributes should look like the DataSource callback.

matkonnerth:
have to think of the writeMask and using DataSource struct for attribute read/write

matkonnerth:
switch back to WIP, this is only experimental

Issue #4118, opened 2021-01-13
SimpleOverflowEventType is not necessary


ccvca:
## Description
As https://apps.opcfoundation.org/mantis/view.php?id=4206 points out, it's totally valid to send events notifications of abstract event types.
https://github.com/open62541/open62541/blob/e5eba7bd591642281717be9edbcd0b6d2ba0fe0b/src/server/ua_server_ns0.c#L1104-L1117

From: https://reference.opcfoundation.org/v104/Core/docs/Part3/4.6.2/
> EventTypes defined in this document are specified as abstract and therefore never instantiated in the AddressSpace. Event occurrences of those EventTypes are only exposed via a Subscription.

So all occurences of SimpleOverflowEventType can be replaced by EventQueueOverflowEventType.

## Used CMake options:

```bash
cmake -DUA_ENABLE_SUBSCRIPTIONS_EVENTS :BOOL="True"
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): e5eba7bd591642281717be9edbcd0b6d2ba0fe0b 
 - [x] Other OPC UA SDKs used (client or server): CTT
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

PR #4143, opened 2021-01-27
fix(server): Replace SimpleOverflowEventType by EventQueueOverflowEventType #4118

ccvca:
Fix as discussed here: https://apps.opcfoundation.org/mantis/view.php?id=4206 (#4118)

Shall this be merged to master or 1.2? If the 1.2 should be used for the CTT-Tests of events it might be usefull there, as the issue (#4118)  was initially discovered in conjunction with CTT tests ("unknown event").

PS: I hope that I got the commit message format right this time.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -11,7 +11,7 @@
  *    Copyright 2017 (c) Henrik Norrman
  *    Copyright 2018 (c) Fabian Arndt, Root-Core
  *    Copyright 2017-2020 (c) HMS Industrial Networks AB (Author: Jonas Green)
- *    Copyright 2020 (c) Christian von Arnim, ISW University of Stuttgart  (for VDW and umati)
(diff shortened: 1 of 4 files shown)


jpfr:
Cool! Simplifications are always good. 👍

jpfr:
I am hesitant in bringing changes to 1.2 that might break some applications.
If this is relevant for the CTT, then yes, pleasee backport to 1.2 as well.

jpfr:
>  If this is relevant for the CTT, then yes, pleasee backport to 1.2 as well.

Even better, use `git cherry-pick` so we have no trouble when merging the combined 1.2 fixes back onto master.

ccvca:
> I am hesitant in bringing changes to 1.2 that might break some applications.
> If this is relevant for the CTT, then yes, pleasee backport to 1.2 as well.

We where just informed, that while CTT-developement our server sends "unknown" events of this type. As this might be also a bug in the CTT (subtypes should always be valid), let's keep this in mind. If there are any issues while CTT certification, a quick cherry-pick is always possible.

PR #4144, opened 2021-01-27
[Review] build(deps): update UA-NodeSet

GoetzGoerisch:
Update to the latest [OPCFoundation/UA-Nodeset](https://github.com/OPCFoundation/UA-Nodeset)

New:
Includes .xsd and .bsd files for companion specifications
- PlasticsRubber
- Robotics
- MachineTool

Update:
- Base NodeSet Version 1.04.8 (PublicationDate="2020-11-20")
- PackML Version 1.01 (PublicationDate="2020-10-08")


jpfr:
It's good to keep this up-to-date.

But.. the build breaks when the official nodeset is used.
It seems the problem is with `UA_STATUSCODE_GOOD` now being defined via the csv file as well.
```
Good,0x00000000,"The operation succeeded."
```
We already have that defined in `open62541/types.h`. So the typedef collides.

We can either 
- filter out "good" in the script that generates code from the csv
- remove the definition from types.h

GoetzGoerisch:
Thanks for the suggestion.

Used:
> remove the definition from `types.h`

Changes:
- Removed the manual `UA_STATUSCODE_GOOD` generation
- Deleted outdates statuscode and NodeIds
- All builds requires checked out submodul `deps/ua-nodeset`
- Fixed references of the renamed PLCopen NodeSet

jpfr:
> All builds requires checked out submodul deps/ua-nodeset

Uuh. That's not good.
We should not include the full upstream ua-nodeset repository in our default downloads due to licensing questions.
Still, it would be good if open62541 would continue to build out-of-the-box with just the default download/checkout.
We have gotten bug reports in the past where people downloaded the code as .zip from Github.
And the submodule functionality was therefore lost to them.

I see that you removed three files in /tools/schema
Please just update these three files to the latest upstream version.
Out-of-the-box builds (for the reduced-size information model) should then continue to work.

Otherwise nice and ready to merge!

GoetzGoerisch:
Rebased, looking forward to your feedback @jpfr.

jpfr:
Perfect! Merged.

PR #4157, opened 2021-01-29
refactor(pubsub): Support to DataSetWriterId value in Information model

opcua-tsn-team-kalycito:


andreasebner:
Your current work only set's the fields during the initial creation. You also have to extend the static void
onWrite and onRead methon within ua_pubsub_ns0.c

opcua-tsn-team-kalycito:
Hi @andreasebner,  
We have updated the static void onRead method for DataSetWriterId but we have not done for OnWrite method because DataSetWriterId doesn't have write access.

PR #4158, opened 2021-01-30
fix(arch): Fix architecture definitions for Embarcadero C++Builder clang/win32

ogsadmin:
Embarcadero C++Builder 10.4 (clang 5.0 / win32) has a few incompatibilities in its standard library, which prevents building Open62541 on this compiler.

I would like to suggest two small additions to the architecture definition files to fix build on this compiler. The changes do not break existing behavior and are guarded by `#ifdef __CODEGEARC__` to only become active for this compiler.

I would also like to add a fix for a bug in the gmtime_s() call in the arch/win32/ua_clock.c file - the call uses the wrong parameter order.

### DIFF
--- a/arch/win32/ua_architecture.h
+++ b/arch/win32/ua_architecture.h
@@ -124,7 +124,7 @@ void UA_sleep_ms(unsigned long ms);
 # define UA_realloc realloc
 #endif
 
-#ifdef(__CODEGEARC__)
+#ifdef __CODEGEARC__
 #define _snprintf_s(a,b,c,...) snprintf(a,b,__VA_ARGS__)
 #endif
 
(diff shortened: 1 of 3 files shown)


jpfr:
Looks good! Can be merged like this.
Please click the CLA (in the CI list) so that we can merge.

ogsadmin:
Unfortunately there is nothing to click in the CI list (or I can't find it, sorry). I've signed the CLA when I opened the first PR, but did not get a chance to do it again for this PR...

andreasebner:
Add a "empty commit" which can be removed by an rebase later to restart the CI.

jpfr:
Great. Rebased and now the CLA went green.
👍 For the contribution!

PR #4159, opened 2021-02-01
fix(server): Add decrement for monitoredItemSize when deleting monitoredItem

sqadasi:
having a limited monitoredItemSize leads to a problem when removing monitoredItems without deleting the subscribtion. No further monitoredItems can be added because the server doesn't decrement the stored monitoredItemsSize on monitoredItemDelete. However, deleting the subscribtion sets monitoredItemsSize to 0 again.
This bug is found by running the OPC UA CTT on a server with limited maxMonitoredItemsPerSubscription.

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -649,6 +649,9 @@ Operation_DeleteMonitoredItem(UA_Server *server, UA_Session *session, UA_Subscri
         *result = UA_STATUSCODE_BADMONITOREDITEMIDINVALID;
         return;
     }
+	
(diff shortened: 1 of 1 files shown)


jpfr:
Please rebase for 1.2

sqadasi:
Looks like I messed things up a bit. Would it be easier if we delete this pull request and then I can create a new one for 1.2?

sqadasi:
Looks like this bug was already fixed in https://github.com/open62541/open62541/commit/be3754a8706d9a615fb1cbef735d01fa93e73a6b

PR #4162, opened 2021-02-02
Fix(ci) for ubuntu20.04

ccvca:
Update to ubuntu 20.04 for github actions (latest)
As python2-sphinx and clang++-10 is no longer available, I used python3 and clang-11

The exact version for ubuntu is set, as latest is sometimes ubuntu18.04 and sometimes ubuntu20.04 depending on the repository.

PS: Should the 1.2 branch stay on ubuntu18.04? If not, I can move this to 1.2.


jpfr:
Nice 👍

PR #4153, opened 2021-01-28
Add client side support for localeIds in ActivateSession

basyskom-jvoe:

PR #4165, opened 2021-02-03
[REVIEW] Locale ids support in the server

basyskom-jvoe:
This is a first start at supporting locale IDs
- The locale IDs from ActivateSession are stored in the UA_Session
- The server is extended with a getter for the locale IDs of a session

The client-side support has been cherry-picked from #4153.

This should be sufficient for the Value attribute of custom Variable nodes.
The BrowseName and Description attributes still need a different solution. I see the following options:
- The user can pass an array or map for both attributes instead of the scalar UA_LocalizedText and the server does the selection according to Part 4, 5.6.3.2
- Additional, optional callbacks for reading BrowseName and Description

I would prefer the second solution because this can be realized with less memory footprint (no need to duplicate values for nodes with the same display name or description).

### DIFF
--- a/include/open62541/plugin/nodestore.h
+++ b/include/open62541/plugin/nodestore.h
@@ -402,8 +402,17 @@ struct UA_NodeHead {
     UA_NodeId nodeId;
     UA_NodeClass nodeClass;
     UA_QualifiedName browseName;
+
+    /* A node can have different localizations for displayName and description.
+     * The server selects a suitable localization depending on the locale ids
+     * that are set for the current session.
+     *
+     * Locales are added simply by writing a LocalizedText value with a new
+     * locale. A locale can be removed by writing a LocalizedText value of the
+     * corresponding locale with an empty text field. */
     UA_LocalizedTextListEntry *displayName;
     UA_LocalizedTextListEntry *description;
+
     UA_UInt32 writeMask;
     size_t referencesSize;
     UA_NodeReferenceKind *references;
--- a/include/open62541/plugin/nodestore.h
+++ b/include/open62541/plugin/nodestore.h
@@ -391,13 +391,19 @@ UA_NodeReferenceKind_findTarget(const UA_NodeReferenceKind *rk,
 UA_EXPORT UA_StatusCode
 UA_NodeReferenceKind_switch(UA_NodeReferenceKind *rk);
 
+/* Singly-linked LocalizedText list */
+typedef struct UA_LocalizedTextListEntry {
+    struct UA_LocalizedTextListEntry *next;
+    UA_LocalizedText localizedText;
+} UA_LocalizedTextListEntry;
+
 /* Every Node starts with these attributes */
 struct UA_NodeHead {
     UA_NodeId nodeId;
     UA_NodeClass nodeClass;
     UA_QualifiedName browseName;
-    UA_LocalizedText displayName;
-    UA_LocalizedText description;
+    UA_LocalizedTextListEntry *displayName;
+    UA_LocalizedTextListEntry *description;
     UA_UInt32 writeMask;
     size_t referencesSize;
     UA_NodeReferenceKind *references;
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -488,36 +488,52 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId);
 UA_EXPORT UA_StatusCode UA_THREADSAFE
 UA_Server_closeSession(UA_Server *server, const UA_NodeId *sessionId);
 
-/* Session Parameters: Besides the user-definable session context pointer,
- * so-called session parameters are a way to attach key-value parameters to a
- * session. This enables "plugins" to attach data to a session without impacting
- * the user-definedable session context pointer. */
+/**
+ * Session attributes: Besides the user-definable session context pointer (set
+ * by the AccessControl plugin when the Session is created), a session carries
+ * attributes in a key-value list. Some attributes are present in every session
+ * and shown in the list below. Additional attributes can be manually set as
+ * meta-data.
+ *
+ * Always present as session attributes are:
+ *
+ * - 0:localeIds [UA_String]: List of preferred languages (read-only)
+ * - 0:clientDescription [UA_ApplicationDescription]: Client description (read-only)
+ * - 0:sessionName [String] Client-defined name of the session (read-only)
+ */
 
+/* Returns a shallow copy of the attribute. Don't _clear or _delete the value
+ * variant. Don't use the value once the Session could be already closed in the
+ * background or the attribute of the session replaced. Hence don't use this in a
+ * multi-threaded application. */
+UA_EXPORT UA_StatusCode
+UA_Server_getSessionAttribute(UA_Server *server, const UA_NodeId *sessionId,
+                              const UA_QualifiedName key, UA_Variant *outValue);
+
+/* Return a deep copy of the attribute */
 UA_EXPORT UA_StatusCode UA_THREADSAFE
-UA_Server_setSessionParameter(UA_Server *server, const UA_NodeId *sessionId,
-                              const UA_QualifiedName key,
-                              const UA_Variant *value);
-
-UA_EXPORT void UA_THREADSAFE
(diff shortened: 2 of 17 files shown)


matkonnerth:
this is maybe a related PR from myside: https://github.com/open62541/open62541/pull/4138

basyskom-meerkoetter:
@jpfr any feedback on this PR? I know this it takes some rebasing, but is the general direction sound?

basyskom-meerkoetter:
The feature can be tested with the UaExpert.
Settings -> Configure UaExpert
Set General.LocaleId accordingly

jpfr:
The feature is good. And the code quality very sound as usual.

However I would prefer to keep this out of the API surface of the Nodestore.

How about we put the BrowseName and Description into a singly-linked lists into UA_Node?
Then the nodestore can forward the node over the standard API and the server can pick out the language he prefers.

jpfr:
> How about we put the BrowseName and Description into a singly-linked lists into UA_Node?

What are your thoughts on that @basyskom-meerkoetter ?

basyskom-jvoe:
> > How about we put the BrowseName and Description into a singly-linked lists into UA_Node?
> 
> What are your thoughts on that @basyskom-meerkoetter ?

Using a linked list in UA_Node would increase the memory footprint if there are lots of nodes sharing the same localized text. An advantage of this solution would be that the client doesn't have to do the bookkeeping.

basyskom-jvoe:
@jpfr Do you have an API proposal for your approach (setting the values, getting the value with the best matching locale)?

Stasik0:
@basyskom-jvoe @jpfr to com back to this topic a bit: what would speak against 
- passing the locale from the session to the nodestore
- letting nodestore pick the "right" display names when node is required from the nodestore e.g. from a read service based on the session data
then the nodestore api is almost unchanged (except the one session passing) between the server and the nodestore

for putting mutliple values into a nodestore we could add some nodestore-implementation-specific api which is not part of the API between server and nostresore, but can be triggered from user-application during bootstrapping

basyskom-jvoe:
@jpfr Would it be possible to extract the first part (saving the locale ids in the session and adding API to get the locale ids for a session) into a separate PR and merge it to master or are there any concerns?

basyskom-jvoe:
UA_Server_getSessionById() seems to require an active lock on server->serviceMutex. This could become a problem if the session localeIds getter is also used inside another internal function.

basyskom-jvoe:
The main problem for DisplayName and Description support is that there are dozens of places where other internal functions read the LocalizedText members of the node head. Changing the members to a SLIST_HEAD would require extensive changes.

There are also places where no session is involved. For these, we could always use the first position in the list.
For writing the attributes, we must use the locale to determine if we update an existing entry or add a new one.
The same behavior could be available as public API. As an alternative, we could also ask the user to write the attributes using the write service.

basyskom-jvoe:
@Stasik0 The problem with the approach you suggested is that there is no API in the nodestore to get an attribute. Unfortunately, all functions access the members of the UA_Node directly. Adding such an API would require the same effort as changing the displayName and description members to a SLIST or TAILQ.

jpfr:
Proposal:

1. We store the session locale in the session parameters under the key “locale”.
2. We forward the session id to the nodestore.
3. The nodestore can use the public API UA_Server_getSessionParameter to retrieve the locale.

basyskom-jvoe:
Switching to SLIST has removed the false positive. Inserting at the end of an existing list has become a bit more complicated but I guess this won't be a problem because most access operations will be reads.

basyskom-jvoe:
@jpfr We would really like to bring this feature upstream.

Is the current API concept sound or should we change something?
If the API for DisplayName and Description is still controversial, I'd suggest we move the function to get the locale IDs for a session to a separate PR. This would already enable localizing the value attribute using data source callbacks.

basyskom-jvoe:
@jpfr The PR has been rebased and is ready for review.

src/nodestore_attribute_helper.c
```diff
@@ -0,0 +1,209 @@
...
+UA_StatusCode UA_NodeStore_insertOrUpdateDescriptionForHead(const UA_LocalizedText *value,
+                                                            UA_NodeHead *head)
+{
+    LocalizedTextEntry *lt = NULL;
+    UA_Boolean found = false;
+    SLIST_FOREACH(lt, &head->description, pointers) {
+        if (UA_String_equal(&lt->localizedText.locale, &value->locale)) {
+            UA_LocalizedText_clear(&lt->localizedText);
+
+            if (value->text.length == 0) {
```
jpfr, line 178: I don't understand this.
If `value->text.length == 0` the lookup will never match.
Except for the odd case where an empty description was already stored.

include/open62541/plugin/nodestore.h
```diff
@@ -411,6 +421,16 @@ struct UA_NodeHead {
 #endif
 };
 
+UA_LocalizedText UA_NodeStore_getDisplayNameForHead(const UA_NodeHead *head, UA_Server *server,
```
jpfr, line 424: This should be an internal API to reduce the API surface.
The user typically does not have access to any session ids.

In our method signature we typically go from "big context" to local context.
In that sense, use the server for the first argument, then the node-head, then the value.

jpfr (commented):
The overall approach with the singly-linked list is good.
A few comments inline.

include/open62541/plugin/nodestore.h
basyskom-jvoe: The user always gets the sessionId in callbacks (method, data source, etc...) so I thought it could be handy to have access to this

include/open62541/server.h
```diff
@@ -1795,6 +1795,17 @@ typedef struct {
...
+/**
+* Localization
+* ------------
+*
+* A client may pass an array of locale IDs in the ActivateSession request (Part 4, 5.6.3.2).
+* This information can be used to return localized text to the client depending on its preferred
+* language(s).
+*/
+UA_EXPORT UA_THREADSAFE const UA_LocaleId *
+UA_Server_getLocaleIdsForSession(UA_Server *server, const UA_NodeId *sessionId, size_t *localeIdsSize);
```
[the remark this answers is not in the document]
basyskom-jvoe: @jpfr The user needs access to this information to localize values in data source callbacks, so it can't be kept internal.
Our main use case for this PR is localizing messages and other properties in events.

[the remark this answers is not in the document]
jpfr: Ok, got it.

The method `UA_Server_getLocaleIdsForSession` is marked as thread-safe.
But it isn't.

Please make two versions of this. `UA_Server_getLocaleIdsForSession` which assumes that the session is stable as long as the data is used (and not marked as thread-safe).
And `UA_Server_getLocaleIdsCopyForSession` which generates a copy of the data and which is thread-safe.

src/nodestore_attribute_helper.c
```diff
@@ -0,0 +1,209 @@
...
+UA_LocalizedText UA_NodeStore_getDisplayNameForHead(const UA_NodeHead *head, UA_Server *server,
+                                                    const UA_NodeId *sessionId)
+{
+    UA_LocalizedText result;
+    UA_LocalizedText_init(&result);
+    LocalizedTextEntry *lt = NULL;
+
+    if (server && sessionId) {
+        size_t localeIdsSize = 0;
+        const UA_LocaleId * locales = UA_Server_getLocaleIdsForSession(server, sessionId,
```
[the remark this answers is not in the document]
basyskom-jvoe: @jpfr Do you have a suggestion how to handle this?

src/nodestore_attribute_helper.c
basyskom-jvoe: This is meant to allow removing a translation from the list by passing an empty text with the locale to remove.

jpfr:
Reworked this a little bit and fixed the locking workarounds.

There is now a general way to get session attributes from the userland.
`UA_Server_getSessionAttribute` that is also used to get the locales.

Please validate that this works for you.
Then we can clean the commit history and merge.

basyskom-jvoe:
@jpfr Many thanks for your reviews and modifications - your generic getter approach is much more future proof.

PR #4172, opened 2021-02-08
Update tutorial_pubsub_mqtt.c

jackybek:
based on the condition checked in open62541.h line 89311 function name: UA_PubSubChannelMQTT_regist(), it is comparing transportSettings->content.decoded.type->typeIndex == UA_TYPES_BROKERDATASETREADERTRANSPORTDATATYPE); however the sample code initialise decoded.type wrongly.

After making the above change, i can now subscribe to the json message on the MQTT broker successfully.


jpfr:
In that case, you also need to change the definition of the variable that is referenced.
The datatype reference needs to match the underlying pointer.

c```
UA_BrokerWriterGroupTransportDataType brokerTransportSettings;
```

jpfr:
Are you still on this @jackybek ?

jackybek:
Yes .. based on the above proposed changes I tested using my local copy and it worked.

PR #4178, opened 2021-02-11
[Review] Setup CodeScanning with CodeQL

GoetzGoerisch:
As per discussion with @andreasebner:

Initial setup of codeScanning with CodeQL


jpfr:
Very good addition!
Merged.

PR #4184, opened 2021-02-15
Fix references and change monitoreditems default queue size

jpfr:


jpfr:
@juangburgos, right. Fixed the commit string.

juangburgos:
Thank you!

PR #4187, opened 2021-02-15
fix(sec): fix OpenSSL V1.1 API incompatibility for Windows 32 build 

ogsadmin:
While trying to build Open62541 on Windows 32-bit (using Embarcadero C++Builder 10.4) with a prebuild OpenSSL V1.1.1j (from [https://kb.firedaemon.com/support/solutions/articles/4000121705](https://kb.firedaemon.com/support/solutions/articles/4000121705)), I encountered an issue related to OpenSSL APIs not available in the V1.1 series anymore.

I would like to suggest a small change, which uses the OPENSSL_API_COMPAT to wrap the API version dependent code.

Thanks!


jpfr:
Looking good. Merged.

PR #4189, opened 2021-02-17
Merge v1.2 to master

jpfr:


ccvca:
"fix(ci): set ubuntu to 18.04" may clash with: https://github.com/open62541/open62541/pull/4162

I think 1.2 should build on 18.04 with old clang. But master should be tested with 20.04

jpfr:
@ccvca That was already addressed in the merge commit.
Our process is to merge all commits from 1.2 on master.
And address changes we don‘t want to have on master in a later commit.
That way we can ensure that no patch is lost „in translation“.

PR #4191, opened 2021-02-19
[Review] feat(pubsub): Support baseTime to timer callbacks

opcua-tsn-team-kalycito:
- baseTime and handleCycleMissWithBaseTime variables are added in timer process

### DIFF
--- a/src/ua_timer.c
+++ b/src/ua_timer.c
@@ -38,6 +38,18 @@ cmpId(const UA_UInt64 *a, const UA_UInt64 *b) {
     return ZIP_CMP_MORE;
 }
 
+static UA_DateTime
+calculateNextTime_withPastBaseTime(UA_DateTime currentTime, UA_DateTime baseTime,
+                                   UA_DateTime interval) {
+    UA_DateTime timeToAdd;
+    /* Base time is always lesser than the current time. Take the difference between current and base time */
+    UA_DateTime diffCurrentTimeBaseTime = currentTime - baseTime;
+    /* Take modulo of the diff time with the interval and subtract it with the interval to get the time to be added with the current time.
+     * This added time will match the current time to base time cycle with respect to the interval */
+    timeToAdd = interval - (diffCurrentTimeBaseTime % interval);
+    return (currentTime + timeToAdd);
+}
+
 ZIP_PROTOTYPE(UA_TimerIdZip, UA_TimerEntry, UA_UInt64)
 ZIP_IMPL(UA_TimerIdZip, UA_TimerEntry, idZipfields, UA_UInt64, id, cmpId)
 
@@ -60,7 +72,7 @@ UA_Timer_addTimerEntry(UA_Timer *t, UA_TimerEntry *te, UA_UInt64 *callbackId) {
 
 static UA_StatusCode
 addCallback(UA_Timer *t, UA_ApplicationCallback callback, void *application, void *data,
-            UA_DateTime nextTime, UA_UInt64 interval, UA_UInt64 *callbackId) {
+            UA_DateTime nextTime, UA_UInt64 interval, UA_TimerPolicy timerPolicy, UA_UInt64 *callbackId) {
     /* A callback method needs to be present */
     if(!callback)
         return UA_STATUSCODE_BADINTERNALERROR;
@@ -77,6 +89,7 @@ addCallback(UA_Timer *t, UA_ApplicationCallback callback, void *application, voi
     te->application = application;
     te->data = data;
     te->nextTime = nextTime;
+    te->timerPolicy = timerPolicy;
 
     /* Set the output identifier */
     if(callbackId)
@@ -92,7 +105,7 @@ UA_Timer_addTimedCallback(UA_Timer *t, UA_ApplicationCallback callback,
                           void *application, void *data, UA_DateTime date,
                           UA_UInt64 *callbackId) {
     UA_LOCK(&t->timerMutex);
-    UA_StatusCode res = addCallback(t, callback, application, data, date, 0, callbackId);
+    UA_StatusCode res = addCallback(t, callback, application, data, date, 0, UA_TIMER_HANDLE_CYCLEMISS_WITH_CURRENTTIME, callbackId);
     UA_UNLOCK(&t->timerMutex);
     return res;
 }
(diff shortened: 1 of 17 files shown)


src/ua_timer.c
```diff
@@ -112,10 +114,30 @@ UA_Timer_addRepeatedCallback(UA_Timer *t, UA_ApplicationCallback callback,
     if(interval == 0)
         return UA_STATUSCODE_BADINTERNALERROR;
 
-    UA_DateTime nextTime = UA_DateTime_nowMonotonic() + (UA_DateTime)interval;
+    UA_DateTime currentTime = UA_DateTime_nowMonotonic();
+    UA_DateTime nextTime;
+    if(baseTime <= 0.0) {
```
jpfr, line 119: UA_DateTime is a signed integer.
Please don't rely on 0 as a special date value.

src/ua_timer.c
```diff
@@ -112,10 +114,30 @@ UA_Timer_addRepeatedCallback(UA_Timer *t, UA_ApplicationCallback callback,
...
+        nextTime = currentTime + (UA_DateTime)interval;
+    }
+    else {
+        if(baseTime > (currentTime + UA_DATETIME_SEC)) {
```
jpfr, line 123: The timer uses UA_DateTime_nowMonototic.
So it should be impossible travel back behind the baseTime.

src/ua_timer.c
```diff
@@ -189,9 +225,23 @@ UA_Timer_process(UA_Timer *t, UA_DateTime nowMonotonic,
 
         /* Set the time for the next execution. Prevent an infinite loop by
          * forcing the next processing into the next iteration. */
-        first->nextTime += (UA_Int64)first->interval;
-        if(first->nextTime < nowMonotonic)
-            first->nextTime = nowMonotonic + 1;
+        first->nextTime += (UA_DateTime)first->interval;
+        if(first->nextTime < nowMonotonic) {
+            if(first->handleCycleMissWithBaseTime) {
+                UA_DateTime timeToAdd;
```
jpfr, line 231: I see this logic repeated in several places.
Pull it out into a static functions and reuse that.

src/ua_timer.h
```diff
@@ -33,6 +33,8 @@ typedef struct UA_TimerEntry {
                                                 the interval is zero, the
                                                 callback is not repeated and
                                                 removed after execution. */
+    UA_Boolean handleCycleMissWithBaseTime;  /* Handle the cycle misses with the next
```
jpfr, line 36: The `pahole` tool gives the current alignment of members as this:
```c
struct UA_TimerEntry {
        struct {
                struct UA_TimerEntry * zip_left;         /*     0     8 */
                struct UA_TimerEntry * zip_right;        /*     8     8 */
                unsigned char      rank;                 /*    16     1 */
        } zipfields;                                     /*     0    24 */

        /* XXX last struct has 7 bytes of padding */

        UA_DateTime                nextTime;             /*    24     8 */
        UA_UInt64                  interval;             /*    32     8 */
        UA_ApplicationCallback     callback;             /*    40     8 */
        void *                     application;          /*    48     8 */
        void *                     data;                 /*    56     8 */
        /* --- cacheline 1 boundary (64 bytes) --- */
        struct {
                struct UA_TimerEntry * zip_left;         /*    64     8 */
                struct UA_TimerEntry * zip_right;        /*    72     8 */
                unsigned char      rank;                 /*    80     1 */
        } idZipfields;                                   /*    64    24 */

        /* XXX last struct has 7 bytes of padding */

        UA_UInt64                  id;                   /*    88     8 */

        /* size: 96, cachelines: 2, members: 8 */
        /* paddings: 2, sum paddings: 14 */
        /* last cacheline: 32 bytes */
};
```
If we put `handleCycleMissWithBaseTime` in the current location, we would move regularly used members behind the cacheline. This comes with a significant performance penalty in a "hot path" of the stack.
Please move `handleCycleMissWithBaseTime` immediately behind the `zipfields` members to use up the padding space instead.

The wording `handleCycleMissWithBaseTime` does not say *how* the missed cycles are handled.
Since we have 7 bytes of padding available, we can make the whole thing more explicit.
Define an enum `UA_TimerPolicy` instead of using a boolean.
And call the variables `timerPolicy` in the function arguments. Its more explicit and shorter.

jpfr (commented):
Looking good. See the inline comments.

jpfr (commented):
Hey there,

the code is high quality. However, I have two requests.

1. Split into two commits. One that only changes ua_timer.h/.c. The second commit to apply the changes to PubSub.
2. The split of baseTime / callbackTime is needlessly complicated. In my understanding, when the baseTime is known the callbackTime can be computed.

opcua-tsn-team-kalycito:
Hi @jpfr, 

Thanks for the comments.

Modifications regarding PubSub have been removed. We have added callbackTime variable as this will be useful when the user has not specified the base time. Now, we have found a way to handle this callbackTime even when base time is not specified, which will be seen in the next PR.

src/ua_timer.c
```diff
@@ -139,7 +165,12 @@ UA_Timer_changeRepeatedCallbackInterval(UA_Timer *t, UA_UInt64 callbackId,
     /* Set the repeated callback */
     ZIP_REMOVE(UA_TimerZip, &t->root, te);
     te->interval = (UA_UInt64)(interval_ms * UA_DATETIME_MSEC); /* in 100ns resolution */
-    te->nextTime = UA_DateTime_nowMonotonic() + (UA_DateTime)te->interval;
+    UA_DateTime currentTime = UA_DateTime_nowMonotonic();
+    if(te->nextTime > (currentTime + UA_DATETIME_MSEC))
+        te->nextTime = te->nextTime + (UA_DateTime)te->interval;
+    else
+        te->nextTime = calculateNextTime_withPastBaseTime((currentTime + UA_DATETIME_MSEC), te->nextTime, (UA_DateTime)te->interval);
```
jpfr, line 172: I don't get the logic of this if/else.
You should use the timerPolicy to select how the nextTime is computed.

Put the timerPolicy and the basetime-pointer also in the API of UA_Timer_changeRepeatedCallbackInterval.
When changing the interval, we might change the basetime as well.

jpfr (commented):
This is now looking really good.
One last request. Then we can merge.

jpfr:
Nice. Merged.

PR #4203, opened 2021-02-25
explicit cast to avoid compiler warning

waltronix:
I had the following compiler warning when cross compiling from ubuntu to armv7l with clang
`implicit conversion from 'long long' to 'long double'`
The warning is fixed by explicitly casting the divisor.


jpfr:
Thanks for the contribution 👍

PR #4212, opened 2021-03-01
[REVIEW] Handle errors on client session commands

cochicde:
I saw some functions whose return values are not handled and added some temporary solution. I'm not sure if this the right way of doing it. 

I see the relationship between the session and connection state of the client is not always consistent.

Some of the functions involved in the code change the state of the connection inside of it and some don't.


jpfr:
The intention is good.
Let's get the tests green first.

cochicde:
All green (is travis still working?) I unified the way the retVal and client->connecState are handled when (some) async requests are sent.  

receiveResponseAsync was the problem on my first commit, since it was changing back client->connecState (this is usually changed in the callback function). 

The only discrepancy I see now is at https://github.com/open62541/open62541/blob/master/src/client/ua_client_connect.c#L924 but I didn't dare to change that

jpfr:
Very nice. Merged.
I like the fact that the connectStatus now is mostly handled in just one location (inside of connectIterate).

PR #4213, opened 2021-03-01
[Review] build(deps): update UA-NodeSet

GoetzGoerisch:
Update to the latest UA version 1.04.9 ([Release 21-01-2021](9ba5c136eee04e962444f9bd751c21c6454cbae8))
Includes Updated Machinery 1.01.0 ([Release 25-02-2021](3fff03815de8694cbbadfcabf7e32fdb8d8a3f3c))

PR #4217, opened 2021-03-02
[REVIEW] feat(ci): automatic gitsubmodule updates

GoetzGoerisch:
Following my and other recent updates to the OPC UA NodeSets. #4213 #4144 #3967 

I propose an semi-automatic update of the corresponding gitsubmodules.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gitsubmodule" # See documentation for possible values
+    directory: "/" # Location of package manifests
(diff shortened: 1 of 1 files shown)


jpfr:
Some of these recent updates were problematic.
We had to adjust some of our generation scripts.

More such problematic changes have to anticipated. For example, the .bsd format for datatypes is now deprecated.

Let's not automatic this. The effort to do a PR is small.
And we see the build result before pressing the merge button.

GoetzGoerisch:
@jpfr I understand this explanation.

The PR is generated automatically, but not merged.

Therefore, one sees the build results and if a problematic change happens it results in a failing PR run. 😄

jpfr:
> The PR is generated automatically, but not merged.

In that case, yes please!

GoetzGoerisch:
I assume, the docker cloud failing build is unrelated to this change, but I have no access to asses this, as the "Details" gives me a 404.

GoetzGoerisch:
For our project we are working with this method.
See https://github.com/umati/Sample-Server/pull/108 as an example.

jpfr:
Nice 👍 
Merged.

PR #4225, opened 2021-03-05
[WIP] feat(eventfilter): Extended EventFilter

UranikB:
Concept:

The EventFilter defined in Part 4 (7.17.3 EventFilter) is set during the Event Subscription and enables the client to filter EventNotifications. 
The whereClause parameter of the EventFilter specifies the events.
The selectClause parameter selects the event fields of the events specified by the whereClause. 
EventNotifications containing the selected fields are sent to the client if it passes the whereClause.

It is planned, that this pull request extends the current implementation of the EventFilter.
Currently, only limited filtering is possible. It is possible to select specific event fields and filter according to the type of event. 
However, there is no initial validation as described in the specification, the selection is only possible if the event is of the type BaseEventType, only the OfType operator is implemented and filters with more than one operator are not possible. Furthermore, no EventFilterResults are generated.

This pull request includes the implementation of most of the missing features, but also includes a tutorial for using it as well as tests for the added functionality.

Still missing parts of the specifications are:
The operators „cast“ and „like“
Non-numeric operators for the relational operators and consequently InList and Between
Dynamic validation of WhereClause
Relational operators are only implemented for numeric operands:
GreaterThan(OrEqual)
LessThan(OrEqual)
Equals

The following paragraphs explain where the according features are implemented.

Initial validation:

To verify the validity of a new or updated filter, as the specification demands, an initial validation is performed in the following added functions:
UA_initialWhereClauseValidation
UA_initialSelectClauseValidation

This static validation includes checks for the result codes described in Table 144 of Part 4 and checks on the number of operands. 

The aforementioned functions are implemented and tested, however not included in the subscription process yet. To do so the functions have to be called somewhere in the subscription process (We do not know where exactly yet).

Evaluation validation:

The whereClause of filters containing more than one operator is set as an array of ContentFilterElements which contain the operator and the operands which can be one of three types: ElementOperands, LiteralOperands and SimpleAttributeOperands. 
However, the implicit structure of the whereClause is a tree structure (See Annex B of Part 4). Therefore, the evaluation is performed as a recursive descent.
To implement this, the original function UA_Server_evaluateWhereClauseContentFilter in ua_subscription_events had to be exchanged, because it would not allow for the recursive call as intended. Thereby, the evaluation of the whereClause is now performed in:
UA_Server_evaluateWhereClause

Here the operator is selected by a switch and implemented according to the Table 119 in Part 4. In order to perform the operation, the operands have to be resolved according to their type. This is performed by another added function resolveOperand. Furthermore, two other helper functions are added to perform the comparison for the relational operands and to resolveBooleans.

For the selectClauses an array of SimpleAttributeOperands is used to specify the fields of the events, that passed through the whereClause. These are evaluated after the filtration according to the whereClause in UA_Server_filterEvent.

The whereClause evaluation is tested by a test which constructs a complex whereClause where all operators are included.

Result Generation:

The result of the evaluation is returned in the form of status codes. The status code corresponds to an error and carries information about the error, that happened during the evaluation.
The results are generated during the evaluation. Thus the whereClauseResults are set in the according switch case in UA_Server_evaluateWhereClause and the selectClauseResults are set in UA_Server_filterEvent.


examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+/**
+ * In this C file we will create a ClientFilter example.
+ * Client configured on the server
```
andreasebner, line 5: Please add a newline between licence and description. The description can be more precise: In this tutorial a client setup an event filter on the server .. You should also mention the server event filter example here.

examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
```
andreasebner, line 1: Tutorial are the general examples. Since the event filter are more advances, please remove the "tutorial" and add it as normal client example.

examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
...
+#ifdef UA_ENABLE_SUBSCRIPTIONS
+
+static void
+handler_events(UA_Client *client, UA_UInt32 subId, void *subContext,
+               UA_UInt32 monId, void *monContext,
+               size_t nEventFields, UA_Variant *eventFields) {
+    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND, "Notification");
+
+    /* The context should point to the monId on the stack */
+    UA_assert(*(UA_UInt32*)monContext == monId);
```
andreasebner, line 44: From my point of view, asserts should not be used in the examples.

examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+}
+
+/**
+ * Setting up WhereClauses
+ * ^^^^^^^^^^^^^^^^^^^
+ * The WhereClause defines the filtering criteria of an EventFilter.
+ * It is represented in a ContentFilter Structure.
+ * The ContentFilter structure defines a collection of elements,
```
andreasebner, line 73: please define "elements" here more prcise. Events? Fields?..

examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
...
+    secondNodeId->namespaceIndex = 0;
+    secondNodeId->identifierType = UA_NODEIDTYPE_NUMERIC;
+    secondNodeId->identifier.numeric = UA_NS0ID_EVENTQUEUEOVERFLOWEVENTTYPE;
+    UA_Variant_setScalar(&secondLiteralOperand->value, secondNodeId, &UA_TYPES[UA_TYPES_NODEID]);
+    contentFilter[0].elements[2].filterOperands[0].content.decoded.data = secondLiteralOperand;
+    return contentFilter;
+}
+
+static UA_SimpleAttributeOperand *
+setupSelectClauses(void) {
```
andreasebner, line 139: The "setupWhereClauses" has a description which is missing here.

examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
...
+
+    /* opc.tcp://uademo.prosysopc.com:53530/OPCUA/SimulationServer */
+    /* opc.tcp://opcua.demo-this.com:51210/UA/SampleServer */
+    UA_StatusCode retval = UA_Client_connect(client, "opc.tcp://ammar:4840/");
+    if(retval != UA_STATUSCODE_GOOD) {
+        UA_Client_delete(client);
+        return EXIT_FAILURE;
+    }
+
+#ifdef UA_ENABLE_SUBSCRIPTIONS
```
andreasebner, line 192: I think the overall example makes only sense with enabled subscriptions. Therefore just built it if subscriptions are enabled and remove this check.

examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
...
+        printf("Usage: tutorial_client_events <opc.tcp://server-url>\n");
+        return EXIT_FAILURE;
+    }
+
+    UA_Client *client = UA_Client_new();
+    UA_ClientConfig_setDefault(UA_Client_getConfig(client));
+
+    /* opc.tcp://uademo.prosysopc.com:53530/OPCUA/SimulationServer */
+    /* opc.tcp://opcua.demo-this.com:51210/UA/SampleServer */
+    UA_StatusCode retval = UA_Client_connect(client, "opc.tcp://ammar:4840/");
```
andreasebner, line 186: Please remove "ammar" here and use e.g. localhost as the default. Remove the example server since this example should be used together with the event filter server example.

examples/tutorial_server_eventfilter.c
```diff
@@ -0,0 +1,282 @@
...
+ * ^^^^^^^^^^^^^^^^^^^
+ * In order to set up the event, we can first use ``UA_Server_createEvent`` to give us node representations of the events.
+ * All we need for this is our `EventTypes`. Once we have our event nodes, which are saved internally as `ObjectNodes`,
+ * we can define the attributes the different events have the same way we would define the attributes of an object node. It is not
+ * necessary to define the attributes `EventId`, `ReceiveTime`, `SourceNode` or `EventType` since these are set
+ * automatically by the server. In this example, we will be setting the fields 'Message' and 'Severity' in addition
+ * to `Time` which is needed to make the example UaExpert compliant.
+ */
+static UA_StatusCode
+setUpEvent(UA_Server *server, UA_NodeId *outId, UA_NodeId eventType, UA_Boolean random,size_t index) {
```
andreasebner, line 82: Blank missing.

examples/tutorial_server_eventfilter.c
```diff
@@ -0,0 +1,282 @@
...
+    UA_NodeId* eventNodeId;
+    eventNodeId = (UA_NodeId*)
+        UA_Array_new(eventTypesCount, &UA_TYPES[UA_TYPES_NODEID]);
+    /* set up event */
+    for(size_t i = 0 ; i < eventTypesCount; i++) {
+        retval = setUpEvent(server, &eventNodeId[i], eventTypes[i], false,i);
+        if(retval != UA_STATUSCODE_GOOD) {
+            UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND,
+                           "Creating event failed. StatusCode %s", UA_StatusCode_name(retval));
+            return retval;
```
andreasebner, line 183: If you return here i think we will lose the allocated memory for "eventNodeId". This is a general hint: Check on each error return if there is some cleanup needed.

examples/tutorial_server_eventfilter.c
```diff
@@ -0,0 +1,282 @@
...
+                       "Triggering event failed. StatusCode %s", UA_StatusCode_name(retval));
+    return retval;
+}
+
+/**
+ * Now, all that is left to do is to create a method node which uses our callback. We do not
+ * require any input and as output we will be using the `EventId` we receive from ``triggerEvent``. The `EventId` is
+ * generated by the server internally and is a random unique ID which identifies that specific event.
+ *
+ * This method node will be added to a basic server setup.
```
andreasebner, line 228: I don't get this description "This method node will be added to a basic server setup."

examples/tutorial_server_eventfilter.c
```diff
@@ -0,0 +1,282 @@
...
+        return retval;
+    }
+    retval = UA_Server_triggerEvent(server, eventNodeId,
+                                    UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER),
+                                    NULL, UA_TRUE);
+    if(retval != UA_STATUSCODE_GOOD)
+        UA_LOG_WARNING(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND,
+                       "Triggering event failed. StatusCode %s", UA_StatusCode_name(retval));
+    return retval;
+}
```
andreasebner, line 221: You're defining a lot of complex events but you don't write the properties here. Is there any reason? Because we could youse more interesting where clauses if we have properties with different values.

examples/tutorial_server_eventfilter.c
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * Emitting events by calling methods
+ * ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+ * The following example is an extension of the server event tutorial. We will be
+ * creating two method nodes. One which generates multiple event from the server node and
+ * one which generates a random event
+ *
+ * The event we want to generate should not be simple, in order to allow the client to set
+ * a meaningful filter. `EventTypes` are saved internally as `ObjectTypes`,
+ * so add the a few types as you would new `ObjectTypes`. */
```
andreasebner, line 27: This example can be used together with the event filter example. But the example itself contains no event filter specific definitions. Therefore please don't call it tuturial, mention the client example here and rename it e.g. server_random_events.c

src/server/ua_services_monitoreditem.c
```diff
@@ -402,6 +402,47 @@ Operation_CreateMonitoredItem(UA_Server *server, UA_Session *session,
...
+     * the server can access the EventFilter which is contained in the MonitoredItem.
+     * According to 7.17.3 in Part 4 : The Server shall validate the whereClause and the selectClauses when a Client creates or updates the EventFilter.
+     * The initial validation of the WhereClause and SelectClause are done here.
+     * EventFilterResult isn't currently being used to send Information to the Client.
+     * If the Validation is successful the Server prints a Notification and continues running,
+     * otherwise the connection with the Client will be terminated */
+#ifdef UA_ENABLE_SUBSCRIPTIONS_EVENTS
+    if(newMon->parameters.filter.content.decoded.type == &UA_TYPES[UA_TYPES_EVENTFILTER]) {
+        UA_EventFilter *eventFilter =
+            (UA_EventFilter *)newMon->parameters.filter.content.decoded.data;
```
andreasebner, line 415: Do we need a Null check here?

src/server/ua_services_monitoreditem.c
```diff
@@ -402,6 +402,47 @@ Operation_CreateMonitoredItem(UA_Server *server, UA_Session *session,
...
+        if(eventFilter->whereClause.elementsSize != 0) {
+            UA_ContentFilterResult *contentFilterResult =
+                UA_Server_initialWhereClauseValidation(server, &eventFilter->whereClause);
+            if(contentFilterResult->elementResults[0].statusCode == UA_STATUSCODE_GOOD) {
+                UA_LOG_INFO_SUBSCRIPTION(&server->config.logger, cmc->sub,
+                                         "initialWhereClauseValidation succeeded");
+            }else{
+                result->statusCode = contentFilterResult->elementResults[0].statusCode;
+                UA_ContentFilterResult_delete(contentFilterResult);
+                return;
```
andreasebner, line 425: We can lose memory here: "newMon". Please check every return if there is some cleanup needed.

andreasebner (commented):
Thank's for you contribution! The current review is not complete. Your test case and the canges in ua_subscription_events.c are currently not reviewed. You will receive the review of the second batch soon.

src/server/ua_subscription_events.c
```diff
@@ -293,16 +692,445 @@ UA_Server_evaluateWhereClauseContentFilter(UA_Server *server,
...
+UA_Server_evaluateWhereClause(UA_Server *server,
+                              UA_Session *session,
+                              const UA_NodeId *eventNode,
+                              const UA_ContentFilter *contentFilter,
+                              UA_ContentFilterResult *contentFilterResult,
+                              UA_Variant* valueResult, UA_UInt16 index) {
+    valueResult[index].storageType = UA_VARIANT_DATA;
+    valueResult[index].arrayDimensions = NULL;
+    valueResult[index].arrayDimensionsSize = 0;
+    valueResult[index].arrayLength = 0;
```
andreasebner, line 720: Since you init all values with 0, a memset call would be slimmer.

src/server/ua_subscription_events.c
```diff
@@ -293,16 +692,445 @@ UA_Server_evaluateWhereClauseContentFilter(UA_Server *server,
...
+                contentFilterResult->elementResults[index].operandStatusCodes[1] =
+                    UA_STATUSCODE_BADFILTEROPERANDINVALID;
+                break;
+            }
+
+            /* Check if the operands conform to the operator*/
+            if(!UA_DataType_isNumeric(firstOperand.type) ||
+               !UA_Variant_isScalar(&firstOperand) ||
+               (firstOperand.type == &UA_TYPES[UA_TYPES_DOUBLE]) ||
+               (firstOperand.type == &UA_TYPES[UA_TYPES_FLOAT])) {
```
andreasebner, line 869: Are custom types allowed and handled?

examples/tutorial_client_eventfilter.c
```diff
@@ -0,0 +1,267 @@
...
+#pragma warning(disable:4996) // warning C4996: 'UA_Client_Subscriptions_addMonitoredEvent': was declared deprecated
+#endif
+
+#ifdef __clang__
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
+#ifdef __GNUC__
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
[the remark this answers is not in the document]
UranikB: No they are not needed anymore, they will be deleted in the next commit.

examples/tutorial_client_eventfilter.c
UranikB: We have updated the description.

examples/tutorial_client_eventfilter.c
UranikB: We have changed the names of both tutorials as you wished.

examples/tutorial_client_eventfilter.c
UranikB: "Element" refers to the ContentFilterElement and is described more precisely in the next sentence. Nevertheless, we changed "elements" to "ContentFilterElements" now to avoid any confusion.

examples/tutorial_client_eventfilter.c
UranikB: We added the description in the next commit.

examples/tutorial_client_eventfilter.c
UranikB: We overlooked "Ammar" when we went through the code. The host is now passed as an argument similar to the other tutorials.

examples/tutorial_client_eventfilter.c
UranikB: You are right. The examples are now only built when subscriptions are enabled.

examples/tutorial_server_eventfilter.c
UranikB: We changed the description to "This method node will be added to the server."
Is that any clearer?

examples/tutorial_server_eventfilter.c
UranikB: The properties are simply set in another method (see setUpEvent) for the sake of clarity. 
There is an option to randomize the values of the properties by passing true to the setUpEvent method in the boolean argument "random".

src/server/ua_services_monitoreditem.c
UranikB: No, a null check is not needed as far as we can tell. Why do you believe one might be needed?

src/server/ua_subscription_events.c
```diff
@@ -213,6 +213,405 @@ resolveSimpleAttributeOperand(UA_Server *server, UA_Session *session, const UA_N
...
+    if(type == &UA_TYPES[UA_TYPES_UINT64]){
+        if(variant->type == &UA_TYPES[UA_TYPES_UINT64]){
+            *(UA_UInt64*) data = *(UA_UInt64*) variant->data;
+        }else if(variant->type == &UA_TYPES[UA_TYPES_UINT32]){
+            *(UA_UInt64*) data = *(UA_UInt32*) variant->data;
+        }else if(variant->type == &UA_TYPES[UA_TYPES_UINT16]){
+            *(UA_UInt64*) data = *(UA_UInt16*) variant->data;
+        }else if(variant->type == &UA_TYPES[UA_TYPES_BYTE]){
+            *(UA_UInt64*) data = *(UA_Byte*) variant->data;
+        }else{
```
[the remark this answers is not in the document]
UranikB: The specification defines the conversion rules needed for the implicit conversion in https://reference.opcfoundation.org/v104/Core/docs/Part4/7.4.3/#Table123. 
There only built-in types are specified.

src/server/ua_subscription_events.c
```diff
@@ -213,6 +213,405 @@ resolveSimpleAttributeOperand(UA_Server *server, UA_Session *session, const UA_N
...
+                                       const UA_ContentFilter *contentFilter) {
+    UA_ContentFilterResult *contentFilterResult = UA_ContentFilterResult_new();
+    UA_ContentFilterResult_init(contentFilterResult);
+    contentFilterResult->elementResultsSize = contentFilter->elementsSize;
+    if (contentFilterResult->elementResultsSize !=0){
+        contentFilterResult->elementResults = (UA_ContentFilterElementResult*)
+            UA_Array_new(contentFilterResult->elementResultsSize,&UA_TYPES[UA_TYPES_CONTENTFILTERELEMENTRESULT]);
+        for(size_t i =0; i<contentFilterResult->elementResultsSize; ++i) {
+            UA_ContentFilterElementResult_init(&contentFilterResult->elementResults[i]);
+            contentFilterResult[0].elementResults[i].operandStatusCodes = (UA_StatusCode *)
```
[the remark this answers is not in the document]
UranikB: Is now fixed,

src/server/ua_subscription_events.c
```diff
@@ -213,6 +213,405 @@ resolveSimpleAttributeOperand(UA_Server *server, UA_Session *session, const UA_N
...
+                UA_Array_new(contentFilter->elements->filterOperandsSize,&UA_TYPES[UA_TYPES_STATUSCODE]);
+        }
+    }
+    for(size_t i =0; i<contentFilterResult->elementResultsSize; ++i) {
```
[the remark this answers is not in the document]
UranikB: We checked the code again for format issues, hopefully all are gone now.

src/server/ua_subscription_events.c
```diff
@@ -213,6 +213,405 @@ resolveSimpleAttributeOperand(UA_Server *server, UA_Session *session, const UA_N
...
+            default:
+                contentFilterResult->elementResults[i].statusCode = UA_STATUSCODE_BADFILTEROPERATORUNSUPPORTED;
+                break;
+        }
+    }
+    return contentFilterResult;
+}
+
+/* Performes an initial validation of the selectClause if it was updated or newly created */
+UA_StatusCode*
```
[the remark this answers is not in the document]
UranikB: The function returns the selectClauseResults which consist of an array of UA_StatusCodes.

andreasebner:
The feature is now part of the open62541 repository and was merged bach-wise over time. Thank you very much for your contribution. Since the integrated version was a complete rework, I did not reuse your commits and copy code out of this PR. Your names should be mentioned in the headers. If someone was forgotten, give me a short hint.

andreasebner:
The feature is now part of the open62541 repository and was merged bach-wise over time. Thank you very much for your contribution. Since the integrated version was a complete rework, I did not reuse your commits and copy code out of this PR. Your names should be mentioned in the headers. If someone was forgotten, give me a short hint.

PR #4230, opened 2021-03-07
feat(server) Use RefTree structure during node delete

andreasebner:
This PR extends the current implementation of the DeleteNode operation. Deleting nodes is done in two phases:
1. Calling the Destructors.
2. Delete the child nodes and the node itself.
Both steps are currently using their own loop and browse services calls to walk recursively over the information model. This PR introduces a new step which prepares the RefTree (containing all the child’s of the node) structure with an iterative walk over the information model. This RefTree is used in now used in both phases.

In addition, this PR solves a more substantial problem: If we delete a node, than all childs with only on incoming hierarchical reference can be deleted to. Let's assume the following information model:
A
-B
--C
-C
If we delete node "A" than node "C" can be also deleted because all parents of incoming references are childs of "A". Currently the node "A" and "B" are deleted but "C" will still exist, since there are more than one incoming references. This PR uses the RefTree to check if the parent from incoming references are "nodes to delete". This enables us to decide if node "C" can also be deleted.

### DIFF
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -14,6 +14,7 @@
  *    Copyright 2017-2018 (c) Stefan Profanter, fortiss GmbH
  *    Copyright 2017 (c) Christian von Arnim
  *    Copyright 2017 (c) Henrik Norrman
+ *    Copyright 2021 (c) Fraunhofer IOSB (Author: Andreas Ebner)
  */
 
 #include "ua_server_internal.h"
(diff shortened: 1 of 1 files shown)


src/server/ua_services_nodemanagement.c
```diff
@@ -1833,11 +1841,19 @@ deleteNodeOperation(UA_Server *server, UA_Session *session, void *context,
     UA_NodeId hr = UA_NODEID_NUMERIC(0, UA_NS0ID_HIERARCHICALREFERENCES);
     referenceTypeIndices(server, &hr, &hierarchRefsSet, true);
 
-    recursiveDeconstructNode(server, session, &hierarchRefsSet, &node->head);
-    recursiveDeleteNode(server, session, &hierarchRefsSet, &node->head,
-                        item->deleteTargetReferences);
+    /* The list of childs is needed for the deconstructing and deleting phase.
+     * Within the processNodeLayer we generate a RefTree based set of childs
+     * which can be deleted beside the parent node */
+    RefTree refTree;
```
jpfr, line 1847: You might want to initialize the RefTree...

src/server/ua_services_nodemanagement.c
```diff
@@ -1777,17 +1740,62 @@ recursiveDeleteNode(UA_Server *server, UA_Session *session,
...
     UA_BrowseResult_clear(&br);
+}
 
-    if(removeTargetRefs)
-        removeIncomingReferences(server, session, head);
+static void
+findSetOfSingleRefChilds(UA_Server *server, UA_Session *session,
+                    const UA_ReferenceTypeSet *hierarchRefsSet,
+                    const UA_NodeHead *head, UA_Boolean removeTargetRefs, RefTree *refTree) {
+    UA_StatusCode retval = RefTree_init(refTree);
```
andreasebner, line 1761: @jpfr The initialisiation is done here. But i can move it to the RefTree creation place to make it more visible.

jpfr:
Very nice!

PR #4240, opened 2021-03-12
feat(server): Add modellingRulesOnInstances option, to omit HasModellingRule references in created instances

ccvca:
As the current version of the CCT does not like it, if there are HasModellingRule references on instances (it thinks the node is part of an instance declaration in a type definition), I added an option to suppress these references.

Only Nodes, which are not under the Types folder can omit this reference. In the default configuraiton, HasModellingRule are created in instances as before.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -150,6 +150,16 @@ struct UA_ServerConfig {
     /* Node Lifecycle callbacks */
     UA_GlobalNodeLifecycle nodeLifecycle;
 
+    /** Copy the HasModellingRule reference in instances from the type
+     * definition in UA_Server_addObjectNode and UA_Server_addVariableNode.
+     * Part 3 - 6.4.4
+     * https://reference.opcfoundation.org/v104/Core/docs/Part3/6.4.4/#6.4.4.4
+     *   [...] it is not required that newly created or referenced instances
+     *   based on InstanceDeclarations have a ModellingRule, however, it is
+     *   allowed that they have any ModellingRule independent of the
+     *   ModellingRule of their InstanceDeclaration */
+    UA_Boolean modellingRulesOnInstances;
+
     /**
      * .. note:: See the section for :ref:`node lifecycle
      *    handling<node-lifecycle>`. */
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -191,6 +191,7 @@ setDefaultConfig(UA_ServerConfig *conf) {
     /* conf->nodeLifecycle.destructor = NULL; */
     /* conf->nodeLifecycle.createOptionalChild = NULL; */
     /* conf->nodeLifecycle.generateChildNodeId = NULL; */
+    conf->modellingRulesOnInstances = UA_TRUE;
 
     /* Limits for SecureChannels */
     conf->maxSecureChannels = 40;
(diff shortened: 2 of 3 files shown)


jpfr:
Thanks for this.
Just to be very explicit about it.

When an ObjectType is instantiated under the "Objects" folder, then it should not have ModellingRules.
If the same ObjectType is instantiated within an ObjectType, then the ModellingRules have to be copied over.

Correct?

If the standard is clear about it, then I would go so far as to remove the config option.
That just leads to confusion. Anyhow, it would be good to have the behaviour documented with a reference to the spec section.
And not just in a comment...

Keep up the good work!

ccvca:
I tried to get the parts from the specification:
For ObjectTypes
Part 3 - 5.5.2 (https://reference.opcfoundation.org/v104/Core/docs/Part3/5.5.2/)
>HasSubtype References are used to subtype ObjectTypes. ObjectType subtypes inherit the general semantics from the parent type. The general rules for subtyping apply as defined in Clause 6. It is not required to provide the HasSubtype Reference for the supertype, but it is required that the subtype provides the inverse Reference to its supertype.

So the inverse Reference must be available.

Part 5 - 6.2 (https://reference.opcfoundation.org/v104/Core/docs/Part5/6.2/)
>This ObjectType is the base ObjectType and all other ObjectTypes shall either directly or indirectly inherit from it. However, it might not be possible for Servers to provide all HasSubtype References from this ObjectType to its subtypes, and therefore it is not required to provide this information.

All ObjectTypes are Subtyped from BaseObjectType. The second sentance is a contraction to "shall" in the first sentance, so I assume that this reference is always available.

For VariableTypes
Part 3 - 5.6.5 (https://reference.opcfoundation.org/v104/Core/docs/Part3/5.6.5/)
> HasSubtype References are used to subtype VariableTypes. VariableType subtypes inherit the general semantics from the parent type. The general rules for subtyping are defined in Clause 6. It is not required to provide the HasSubtype Reference for the supertype, but it is required that the subtype provides the inverse Reference to its supertype.

Inverse reference always there.

Part 5 - 7.2 (https://reference.opcfoundation.org/v104/Core/docs/Part5/7.2/)
>The BaseVariableType is the abstract base type for all other VariableTypes. However, only the PropertyType and the BaseDataVariableType directly inherit from this type.
There are no References, except for HasSubtype References, specified for this VariableType.

All VariableTypes  must inherit from the BaseVariableType.

Part 3 - 6.2.1 (https://reference.opcfoundation.org/v104/Core/docs/Part3/6.2.1/)
>An InstanceDeclaration is an Object, Variable or Method that references a ModellingRule with a HasModellingRule Reference and is the TargetNode of a hierarchical Reference from a TypeDefinitionNode or another InstanceDeclaration. The type of an InstanceDeclaration may be abstract, however the instance must be of a concrete type.

Instance declarations do always refer to a type, which is either connected to a BaseObjectType or a BaseVariableType which are both references by the TypesFolder. So a type definition can't be (only) referenced in the objects folder.


Part 3 - 6.4.4 (https://reference.opcfoundation.org/v104/Core/docs/Part3/6.4.4/#6.4.4.4)
>[...] it is not required that newly created or referenced instances based on InstanceDeclarations have a ModellingRule, however, it is allowed that they have any ModellingRule independent of the ModellingRule of their InstanceDeclaration.

So in instances the modelling rules are not required.


As a result, this should be fine, as long as the addess space has references between all object types. Any other opinions on this?

I introduced the new option, to keep the old (correct) behaviour as a default. If it's fine to change it, I can remove it.


PS: @jpfr If you are intrested, I can send you a linkt to the recorded OPC Foundation meeting where this has been discussed.

ccvca:
@jpfr I've updated the comment.

jpfr:
Perfect.

PR #4246, opened 2021-03-16
refactor(pubsub): Cleanup on the pubsub TSN applications

opcua-tsn-team-kalycito:
- Comments added
 - A single base time computation
 - Remove complete blocking on subscriber thread when blocking socket enabled


jpfr:
Looks good.

PR #4251, opened 2021-03-17
[WIP] feat(PubSub): Picoseconds resolution in DataSetMessageHeader 

leonkareem21:
@andreasebner 
We have noticed that the datatype of the field "picoSeconds" (specified in table 81 of chapter 7.2.2.3.4) is not big enough to have the picoseconds saved in. The specified UInt16 only supports numbers up to 65535, however in order to save tens of picosecond we would need a datatype that can support numbers up to 1e11. Should the field not be of type UInt64?


jpfr:
You cannot use system headers like `<error.h>` within the core library.
`UA_DateTime` gives time in 100ns resolution.
The picoSeconds field is intended to be used "on top" of that.

leonkareem21:
> You cannot use system headers like `<error.h>` within the core library.
> `UA_DateTime` gives time in 100ns resolution.
> The picoSeconds field is intended to be used "on top" of that.

Thank you for your comment, @jpfr ,
I have modified the code to not use errno anymore.
The picoSeconds field now correctly displays the amount of 10s of picoseconds passed in the current second.

PR #4252, opened 2021-03-17
[Review] feat(PubSub): FieldMetaData maxStringLength 

leonkareem21:


andreasebner:
Thank you for mentioning the missing maxStringLength attribute. We start working on this feature again, but in addition to fill the field we also need changes in our networking layer.

Issue #4258, opened 2021-03-18
Unable to create monitored item when value status code is 'bad'


GrantEdwards:
## Description

The open62541 server will not create a monitored item if the item's value's status code is bad at the time of the request.

## Background Information / Reproduction Steps

A number of my process-data values have status codes that can be temporarily 'bad' for various reasons (sensor disconnected, value out of range, value marked invalid by user, etc). When values are read in these states, my application code sets the value's status code to various BAD values depending on the situation.

Once a monitored item is created, everything works fine: the client is notified when the value or status changes.

But if the client tries to create a monitored item at a point in time when the value has a BAD status code, the open62541 will refuse to create the monitored item and returns the value's BAD status code.

I don't understand this. Why does the value's status code have any bearing on whether you can monitor it or not?

Are BAD status codes assumed to be permanent and never change to GOOD?

Even if the status remains BAD, the value can still change and can still be useful information.

Used CMake options:

~~~
-- The C compiler identification is GNU 4.6.3
-- The CXX compiler identification is GNU 4.6.3
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /home/nextgen/toolchain/bin/arm-linux-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/nextgen/toolchain/bin/arm-linux-g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PythonInterp: /usr/bin/python (found version "3.8.8") 
-- Found Git: /usr/bin/git (found version "2.26.2") 
-- open62541 Version: 0.0.01.0.1
-- CMAKE_BUILD_TYPE not given; setting to 'Debug'
-- The selected architecture is: posix
-- Found MBEDTLS: /home/nextgen/opcua/mbedtls-2.16.2/include  
-- Performing Test CC_HAS_-fstack-protector-strong
-- Performing Test CC_HAS_-fstack-protector-strong - Failed
-- Performing Test CC_HAS_-fstack-clash-protection
-- Performing Test CC_HAS_-fstack-clash-protection - Failed
-- Performing Test CC_HAS_-mcet -fcf-protection
-- Performing Test CC_HAS_-mcet -fcf-protection - Failed
-- Could NOT find Sphinx (missing: SPHINX_EXECUTABLE) 
[...]

PR #4259, opened 2021-03-18
fix(server): Allow MonitoredItems with an initial StatusCode != Good

jpfr:
Some external sources can report a StatusCode. Allow the generation of
MonitoredItems for most StatusCodes != Good.

The issue was reported by Grant Edwards in #4258.

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -289,18 +289,26 @@ Operation_CreateMonitoredItem(UA_Server *server, UA_Session *session,
         return;
     }
 
-    /* Make an example read to get errors in the itemToMonitor. Allow return
-     * codes "good" and "uncertain", as well as a list of statuscodes that might
-     * be repaired inside the data source. */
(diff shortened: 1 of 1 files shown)


GrantEdwards:
Thanks!

That was fast. I was about to start working on a patch, but it
probably would have taken me half a day to figure out where/what to
change.

I haven't tested the change, but it certainly looks like it will work
for me.

However, it looks like Bad_UserAccessDenied will still cause an abort,
and that's one of the two status codes that the spec specifically says
should not cause a failure:

From part 04 of the spec

> 5.12.2 CreateMonitoredItems

> 5.12.2.1 Description

> [...]

> When a user adds a monitored item that the user is denied read
> access to, the add operation for the item shall succeed and the bad
> status Bad_NotReadable or Bad_UserAccessDenied shall be returned in
> the Publish response.

Do you need to swap UA_STATUSCODE_BADINDEXRANGENODATA and
UA_STATUSCODE_BADUSERACCESSDENIED in your change?

PR #4268, opened 2021-03-22
Support all publisher Id types at connection configuration

MartinaFelberFEM:
According to OpcUa part 14 specification the PublisherId can be of data type unsigned integer or string.

In the connection configuration data type the publisher Id is defined like this:
https://github.com/open62541/open62541/blob/0daf3b6fb1345feb1207332a1b781e6a95e0e96d/include/open62541/server_pubsub.h#L155-L167

UINT32 data type is used to represent publisherId.

In the network message struct the PublisherId is defined like this:
https://github.com/open62541/open62541/blob/0daf3b6fb1345feb1207332a1b781e6a95e0e96d/src/pubsub/ua_pubsub_networkmessage.h#L92-L98

https://github.com/open62541/open62541/blob/0daf3b6fb1345feb1207332a1b781e6a95e0e96d/src/pubsub/ua_pubsub_networkmessage.h#L137-L159

The union contains a GUID as well, that's not defined in the enum and it seems that it's not used anywhere.

The Publisher always sets the Publisher type to UINT16 at generateNetworkMessage function.
https://github.com/open62541/open62541/blob/0daf3b6fb1345feb1207332a1b781e6a95e0e96d/src/pubsub/ua_pubsub_writer.c#L1973-L1975

The reader on the other hand checks for all types at getReaderFromIdentifier function when it processes the received network messages.
https://github.com/open62541/open62541/blob/0daf3b6fb1345feb1207332a1b781e6a95e0e96d/src/pubsub/ua_pubsub_reader.c#L764-L816

The DataSetReaderConfiguration struct uses a Variant to represent the Publisher Id

https://github.com/open62541/open62541/blob/0daf3b6fb1345feb1207332a1b781e6a95e0e96d/include/open62541/server_pubsub.h#L647-L662

If we set the PublisherId type at DataSetReader configuration to anything else then UINT16 (for numeric publisher IDs) then the reader won't identify the received network message correctly, because the network message publisher Id type is always UINT16, and it's not obvious that that's the case.

I am not sure if it's ok to use a variant at UA_PubSubConnectionConfig? like at DataSetReader config ...

Not finished with tests and refactoring ...

### DIFF
--- a/src/pubsub/ua_pubsub_config.c
+++ b/src/pubsub/ua_pubsub_config.c
@@ -153,18 +153,20 @@ static UA_StatusCode
 setConnectionPublisherId(const UA_PubSubConnectionDataType *src,
                          UA_PubSubConnectionConfig *dst) {
     if(src->publisherId.type == &UA_TYPES[UA_TYPES_STRING]) {
-        dst->publisherIdType = UA_PUBSUB_PUBLISHERID_STRING;
+        dst->publisherIdType = UA_PUBLISHERIDTYPE_STRING;
         dst->publisherId.string = *(UA_String*)src->publisherId.data;
-    } else if(src->publisherId.type == &UA_TYPES[UA_TYPES_BYTE] ||
-              src->publisherId.type == &UA_TYPES[UA_TYPES_UINT16] ||
-              src->publisherId.type == &UA_TYPES[UA_TYPES_UINT32]) {
-        dst->publisherIdType = UA_PUBSUB_PUBLISHERID_NUMERIC;
-        dst->publisherId.numeric =  *(UA_UInt32*)src->publisherId.data;
+    } else if(src->publisherId.type == &UA_TYPES[UA_TYPES_BYTE]) {
+        dst->publisherIdType = UA_PUBLISHERIDTYPE_BYTE;
+        dst->publisherId.byte = *((UA_Byte*)src->publisherId.data);
+    } else if(src->publisherId.type == &UA_TYPES[UA_TYPES_UINT16]) {
+        dst->publisherIdType = UA_PUBLISHERIDTYPE_UINT16;
+        dst->publisherId.uint16 = *((UA_UInt16*)src->publisherId.data);
+    } else if(src->publisherId.type == &UA_TYPES[UA_TYPES_UINT32]) {
+        dst->publisherIdType = UA_PUBLISHERIDTYPE_UINT32;
+        dst->publisherId.uint32 = *(UA_UInt32*)src->publisherId.data;
     } else if(src->publisherId.type == &UA_TYPES[UA_TYPES_UINT64]) {
-        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
-                     "[UA_PubSubManager_setConnectionPublisherId] PublisherId is UInt64 "
-                     "(not implemented); Recommended dataType for PublisherId: UInt32");
-        return UA_STATUSCODE_BADNOTIMPLEMENTED;
+        dst->publisherIdType = UA_PUBLISHERIDTYPE_UINT64;
+        dst->publisherId.uint64 = *(UA_UInt64*)src->publisherId.data;
     } else {
         UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
                      "[UA_PubSubManager_setConnectionPublisherId] PublisherId is not valid.");
@@ -1057,14 +1059,36 @@ generatePubSubConnectionDataType(const UA_PubSubConnection *src,
                              &dst->connectionProperties[i]);
     }
 
-    if(src->config->publisherIdType == UA_PUBSUB_PUBLISHERID_NUMERIC) {
-        UA_Variant_setScalarCopy(&dst->publisherId,
-                                 &src->config->publisherId.numeric,
-                                 &UA_TYPES[UA_TYPES_UINT32]);
-    } else if(src->config->publisherIdType == UA_PUBSUB_PUBLISHERID_STRING) {
-        UA_Variant_setScalarCopy(&dst->publisherId,
-                                 &src->config->publisherId.string,
-                                 &UA_TYPES[UA_TYPES_STRING]);
+    switch (src->config->publisherIdType) {
+        case UA_PUBLISHERIDTYPE_BYTE:
+            UA_Variant_setScalarCopy(&dst->publisherId,
+                                     &src->config->publisherId.byte,
+                                     &UA_TYPES[UA_TYPES_BYTE]);
+            break;
+        case UA_PUBLISHERIDTYPE_UINT16:
+            UA_Variant_setScalarCopy(&dst->publisherId,
+                                     &src->config->publisherId.uint16,
+                                     &UA_TYPES[UA_TYPES_UINT16]);
+            break;
+        case UA_PUBLISHERIDTYPE_UINT32:
+            UA_Variant_setScalarCopy(&dst->publisherId,
+                                     &src->config->publisherId.uint32,
+                                     &UA_TYPES[UA_TYPES_UINT32]);
+            break;
+        case UA_PUBLISHERIDTYPE_UINT64:
+            UA_Variant_setScalarCopy(&dst->publisherId,
+                                     &src->config->publisherId.uint64,
+                                     &UA_TYPES[UA_TYPES_UINT64]);
+            break;
+        case UA_PUBLISHERIDTYPE_STRING:
+            UA_Variant_setScalarCopy(&dst->publisherId,
+                                     &src->config->publisherId.string,
+                                     &UA_TYPES[UA_TYPES_STRING]);
+            break;
+        default:
+            UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "generatePubSubConnectionDataType(): publisher Id type is not supported");
+            return UA_STATUSCODE_BADINTERNALERROR;
+            break;
     }
 
     /* Possibly, array size and dimensions of src->config->address and
--- a/src/pubsub/ua_pubsub_config.c
+++ b/src/pubsub/ua_pubsub_config.c
@@ -547,7 +547,7 @@ createReaderGroup(UA_Server *server,
     memset(&config, 0, sizeof(UA_ReaderGroupConfig));
 
     config.name = readerGroupParameters->name;
-    config.securityParameters.securityMode = readerGroupParameters->securityMode;
+    config.securityMode = readerGroupParameters->securityMode;
 
     UA_NodeId readerGroupIdent;
     UA_StatusCode res =
--- a/src/pubsub/ua_pubsub_ns0.c
+++ b/src/pubsub/ua_pubsub_ns0.c
@@ -75,13 +75,31 @@ onRead(UA_Server *server, const UA_NodeId *sessionId, void *sessionContext,
             UA_PubSubConnection_findConnectionbyId(server, *myNodeId);
         switch(nodeContext->elementClassiefier) {
         case UA_NS0ID_PUBSUBCONNECTIONTYPE_PUBLISHERID:
-            if(pubSubConnection->config->publisherIdType == UA_PUBSUB_PUBLISHERID_STRING) {
-                UA_Variant_setScalar(&value, &pubSubConnection->config->publisherId.numeric,
-                                     &UA_TYPES[UA_TYPES_STRING]);
-            } else {
-                UA_Variant_setScalar(&value, &pubSubConnection->config->publisherId.numeric,
-                                     &UA_TYPES[UA_TYPES_UINT32]);
-            }
+            switch (pubSubConnection->config->publisherIdType) {
+                case UA_PUBLISHERIDTYPE_BYTE:
+                    UA_Variant_setScalar(&value, &pubSubConnection->config->publisherId.byte,
+                                         &UA_TYPES[UA_TYPES_BYTE]);
+                    break;
+                case UA_PUBLISHERIDTYPE_UINT16:
+                    UA_Variant_setScalar(&value, &pubSubConnection->config->publisherId.uint16,
+                                         &UA_TYPES[UA_TYPES_UINT16]);
+                    break;
+                case UA_PUBLISHERIDTYPE_UINT32:
+                    UA_Variant_setScalar(&value, &pubSubConnection->config->publisherId.uint32,
+                                         &UA_TYPES[UA_TYPES_UINT32]);
+                    break;
+                case UA_PUBLISHERIDTYPE_UINT64:
+                    UA_Variant_setScalar(&value, &pubSubConnection->config->publisherId.uint64,
+                                         &UA_TYPES[UA_TYPES_UINT64]);
+                    break;
+                case UA_PUBLISHERIDTYPE_STRING:
+                    UA_Variant_setScalar(&value, &pubSubConnection->config->publisherId.string,
+                                         &UA_TYPES[UA_TYPES_STRING]);
+                    break;
+                default:
+                    UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                           "Read error! Unknown PublisherId type.");
+                }
             break;
         default:
             UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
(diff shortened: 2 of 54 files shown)


matkonnerth:
looks good for me, also that integer with 8bytes are supported. in some euromap groups and also in the FLC working group there is the discussion to use a MAC address for unique identification of a publisher, therefore we will need the 8 Byte integer

MartinaFelberFEM:
Thanks for your comment @matkonnerth ! then I will continue with this PR.

jpfr:
The change looks good!
Two requests before we merge.

Since the value can only be a scalar, use an ExtensionObject to store: UA_ExtensionObject_setValue
That has less moving parts like arrays and ArrayDimensions.

Document the valid publisher types in the header for the structure.
So that it is visible in the documentation and not only via error messages.

MartinaFelberFEM:
TODO: test all pubsub examples manually if they still work

MartinaFelberFEM:
Question:
UA_Server_freezeWriterGroupConfiguration() accepts publisherId of type STRING, but           UA_Server_freezeReaderGroupConfiguration() returns an error (BadNotSupported and then BadInternalError) -> what is correct?

As the publisherId is not a DataSetField and the value can't change at runtime I would assume that STRING is ok

MartinaFelberFEM:
Manually tested examples: <br>
   pubsub_publish_encrypted
   pubsub_subscribe_encrypted
   
   tutorial_pubsub_publish_raw
   tutorial_pubsub_subscribe_raw<br>
   
   tutorial_pubsub_publish
   tutorial_pubsub_subscribe<br>
   
   tutorial_pubsub_publish
   server_pubsub_subscribe_custom_monitoring<br>
   
   server_pubsub_publish_rt_level
   server_pubsub_subscribe_rt_level<br>
   
   server_pubsub_rt_field_information_model
   server_pubsub_subscribe_rt_level<br>
   
   tutorial_pubsub_connection<br>
   
   server_pubsub_publisher_iop<br>

MartinaFelberFEM:
Manually tried PubSub file configuration example

```./server_pubsub_file_configuration 4840 server_pubsub_file_configuration```

but decode .bin file failed ...

```
[2021-10-05 11:11:45.779 (UTC+0200)] info/userland      loadPubSubFromFile()
[2021-10-05 11:11:45.781 (UTC+0200)] error/server       [UA_decodeBinFile] decoding UA_Binary failed
[2021-10-05 11:11:45.781 (UTC+0200)] info/userland      loadPubSubFromFile(): failed
```

File: src/pubsub/ua_pubsub_config.c [hunk 1]
```diff
@@ -154,17 +154,19 @@ setConnectionPublisherId(const UA_PubSubConnectionDataType *src,
                          UA_PubSubConnectionConfig *dst) {
     if(src->publisherId.type == &UA_TYPES[UA_TYPES_STRING]) {
         dst->publisherIdType = UA_PUBSUB_PUBLISHERID_STRING;
-        dst->publisherId.string = *(UA_String*)src->publisherId.data;
-    } else if(src->publisherId.type == &UA_TYPES[UA_TYPES_BYTE] ||
-              src->publisherId.type == &UA_TYPES[UA_TYPES_UINT16] ||
-              src->publisherId.type == &UA_TYPES[UA_TYPES_UINT32]) {
-        dst->publisherIdType = UA_PUBSUB_PUBLISHERID_NUMERIC;
-        dst->publisherId.numeric =  *(UA_UInt32*)src->publisherId.data;
+        dst->publisherId.publisherIdString = *(UA_String*)src->publisherId.data;
```
jpfr, line 157: Won't this require a deep copy of the string?

jpfr (commented):
Looks good! A few last comments to consider.
In a few places, I was unsure about shallow copies of strings.
A few unit tests with string ids would help to figure out problems in the valgrind runs...

Keep up the good work!

plugins/ua_pubsub_ethernet.c
```diff
@@ -1247,7 +1247,7 @@ UA_PubSubChannelEthernet_receive(UA_PubSubChannel *channel,
         buffer.length = messageLength;
 
         retval = receiveCallback(channel, receiveCallbackContext, &buffer);
-        if (retval != UA_STATUSCODE_GOOD) {
+        if (UA_StatusCode_isBad(retval)) {
```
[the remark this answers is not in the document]
MartinaFelberFEM: Sorry my mistake
This change has nothing to do with the publisherId implementation.
It should have been a separate commit.

If we have multiple connections within the same multicast group (as unit-test `Test_multiple_connections` checks) then every connection calls receive() and gets the messages from all publishers.
So 1 subscriber receives for example 6 messages, but only 1 of them is the interesting one and all the others are for another Subscriber.

If the message is not for the Subscriber then we get following messages if implementation is still retval != UA_STATUSCODE_GOOD

```

[32minfo/server[0m	PubSub receive. Unknown message received. Will not be processed.
[33mwarn/network[0m	PubSub Connection decode and process failed.
```

And somehow I didn't like the warning: PubSub Connection decode and process failed.
Because it looked like something is wrong but in reality the received message was just for another subscriber.
I wanted to only print the message if there was a "real" error, e.g. decoding failed or any other internal programming error.

If I change the implementation to (UA_StatusCode_isBad(retval))
then I only get 

`[32minfo/server[0m	PubSub receive. Unknown message received. Will not be processed.`

```

receiveCallback return values can be
   GOOD, BAD, UNCERTAIN
   
   decodeAndProcessFunRT
      decodeAndProcessNetworkMessageRT
         UA_NetworkMessage_decodeHeaders
         verifyAndDecryptNetworkMessage

         UA_NetworkMessage_updateBufferedNwMessage
            UA_STATUSCODE_UNCERTAIN
            
         checkReaderIdentifier() - would return UA_STATUSCODE_BADNOTFOUND
               but return value is set to uncertain
               UA_STATUSCODE_UNCERTAIN
   
   decodeAndProcessFun
      decodeAndProcessNetworkMessage
         decodeNetworkMessage
         UA_Server_processNetworkMessage -> returns always GOOD
            checkReaderIdentifier

```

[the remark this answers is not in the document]
MartinaFelberFEM: But maybe warning is not that bad, because the subscriber actually couldn't process message, and it's not an error, so maybe I should not have changed it

I have removed it for now

src/pubsub/ua_pubsub_ns0.c
```diff
@@ -248,15 +266,26 @@ addPubSubConnectionConfig(UA_Server *server, UA_PubSubConnectionDataType *pubsub
...
+    } else if (pubsubConnectionDataType->publisherId.type == &UA_TYPES[UA_TYPES_UINT32]) {
+        connectionConfig.publisherIdType = UA_PUBSUB_PUBLISHERID_UINT32;
+        connectionConfig.publisherId.publisherIdUInt32 = *((UA_UInt32 *) pubsubConnectionDataType->publisherId.data);
+    } else if (pubsubConnectionDataType->publisherId.type == &UA_TYPES[UA_TYPES_UINT64]) {
+        connectionConfig.publisherIdType = UA_PUBSUB_PUBLISHERID_UINT64;
+        connectionConfig.publisherId.publisherIdUInt64 = *((UA_UInt64 *) pubsubConnectionDataType->publisherId.data);
+    } else if (pubsubConnectionDataType->publisherId.type == &UA_TYPES[UA_TYPES_STRING]) {
         connectionConfig.publisherIdType = UA_PUBSUB_PUBLISHERID_STRING;
-        UA_String_copy((UA_String *) pubsubConnectionDataType->publisherId.data, &connectionConfig.publisherId.string);
+        connectionConfig.publisherId.publisherIdString = *((UA_String *) pubsubConnectionDataType->publisherId.data);
```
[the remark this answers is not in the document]
MartinaFelberFEM: I don't understand how I can memcpy the entire union without if/else

Because input UA_PubSubConnectionDataType has publisherId of type Variant
how can I memcpy it and set publisherIdType and publisherId.byte etc. correctly?

or do you mean I should just set the enum type

and then memcpy the pubsubConnectionDataType->publisherId.data to connectionConfig.publisherId?

src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -903,15 +903,8 @@ generateNetworkMessage(UA_PubSubConnection *connection, UA_WriterGroup *wg,
...
-        networkMessage->publisherIdType = UA_PUBLISHERDATATYPE_UINT16;
-        networkMessage->publisherId.publisherIdUInt32 =
-            connection->config->publisherId.numeric;
-    } else if(connection->config->publisherIdType == UA_PUBSUB_PUBLISHERID_STRING) {
-        networkMessage->publisherIdType = UA_PUBLISHERDATATYPE_STRING;
-        networkMessage->publisherId.publisherIdString =
-            connection->config->publisherId.string;
-    }
+    networkMessage->publisherIdType = connection->config->publisherIdType;
+    networkMessage->publisherId = connection->config->publisherId;
```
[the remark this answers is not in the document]
MartinaFelberFEM: At connection creation (UA_Server_addPubSubConnection()) a deep copy of the publisherId has been performed.
Then the Id should be available until connection is destroyed.
So it should be ok to reuse it here.

I have actually had some issues with the string PublisherIds -> but after deep copy at UA_Server_addPubSubConnection() everything worked, and I also checked pointers, and allocated configuration on stack which went out of scope and PubSub operation still worked, because Ids are copied once and then are available until teardown

src/pubsub/ua_pubsub_config.c
MartinaFelberFEM: It shouldn't because `UA_Server_addPubSubConnection() -> UA_PubSubConnectionConfig_copy()` creates the deep copy

I have tried to write a unit-test (see last commit), but I couldn't get it working
I don't have a PubSub binary config file, so I thought I could create UA_PubSubConfigurationDataType structure with unittest, then encode it to bytestring and then use it to simulate PubSub file - already loaded bytestring

But `UA_PubSubManager_loadPubSubConfigFromByteString` fails at decoding and I don't know why

When I try to use `encodePubSubConfiguration` https://github.com/open62541/open62541/blob/c6e4e028a740b50c49a6dd5987ee8e61a52953c9/src/pubsub/ua_pubsub_config.c#L834 to create bytestring of PubSubConfigurationDataType it does not compile because functions `UA_ExtensionObject_calcSizeBinary` and `UA_ExtensionObject_encodeBinary` are used, which are not available anymore I think

So I'm confused why it compiles in the pipeline and how to use it correctly.

MartinaFelberFEM: It looks like `UA_ENABLE_PUBSUB_FILE_CONFIG` is not enabled at any test within https://github.com/open62541/open62541/blob/master/tools/ci.sh

which I think configures the tests right? because the previous travis scripts are obsolete?

src/pubsub/ua_pubsub_ns0.c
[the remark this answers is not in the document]
MartinaFelberFEM: thanks for the suggestion
I set the type within if/else and memcopied PublisherId union after that
Hope that is what you had in mind

src/pubsub/ua_pubsub_config.c
MartinaFelberFEM: Sorry, I was totally confused

I wanted to test PubSub configuration with a file, but I don't have a file, so I created PubSubConfigurationDataType structure in unit-test and tried to encode it to ByteString.
I wanted to use UA_encodeBinary but it didn't work and then I confused a lot of things

But I think it should work now and I have added the test for PubSub file configuration with PublisherId strings

src/pubsub/ua_pubsub_ns0.c
```diff
@@ -248,16 +266,22 @@ addPubSubConnectionConfig(UA_Server *server, UA_PubSubConnectionDataType *pubsub
...
+    } else if (pubsubConnectionDataType->publisherId.type == &UA_TYPES[UA_TYPES_STRING]) {
+        connectionConfig.publisherIdType = UA_PUBLISHERIDTYPE_STRING;
     } else {
-        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER, "Unsupported PublisherId Type used.");
-        //TODO what's the best default behaviour here?
-        connectionConfig.publisherId.numeric = 0;
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER, "Unsupported PublisherId Type used.");
+        return UA_STATUSCODE_BADCONFIGURATIONERROR;
     }
+    memcpy(&connectionConfig.publisherId, pubsubConnectionDataType->publisherId.data, sizeof(connectionConfig.publisherId));
```
jpfr, line 284: This implies a shallow copy of the string id.
That is intended?

src/pubsub/ua_pubsub_reader.c
```diff
@@ -277,20 +277,20 @@ UA_DataSetReader_generateNetworkMessage(UA_PubSubConnection *pubSubConnection, U
...
+        nm->publisherIdType = UA_PUBLISHERIDTYPE_UINT32;
+        nm->publisherId.uint32 = *(UA_UInt32 *) dataSetReader->config.publisherId.data;
         break;
     case UA_DATATYPEKIND_UINT64:
-        nm->publisherIdType = UA_PUBLISHERDATATYPE_UINT64;
-        nm->publisherId.publisherIdUInt64 = *(UA_UInt64 *) dataSetReader->config.publisherId.data;
+        nm->publisherIdType = UA_PUBLISHERIDTYPE_UINT64;
+        nm->publisherId.uint64 = *(UA_UInt64 *) dataSetReader->config.publisherId.data;
         break;
     default:
```
jpfr, line 295: String is missing.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -39,6 +39,9 @@ UA_PubSubConnectionConfig_copy(const UA_PubSubConnectionConfig *src,
                                UA_PubSubConnectionConfig *dst) {
     UA_StatusCode res = UA_STATUSCODE_GOOD;
     memcpy(dst, src, sizeof(UA_PubSubConnectionConfig));
+    if (src->publisherIdType == UA_PUBLISHERIDTYPE_STRING) {
+        UA_String_copy(&src->publisherId.string, &dst->publisherId.string);
```
jpfr, line 43: add `res |=`.

src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -903,15 +903,8 @@ generateNetworkMessage(UA_PubSubConnection *connection, UA_WriterGroup *wg,
...
-    if(connection->config->publisherIdType == UA_PUBSUB_PUBLISHERID_NUMERIC) {
-        networkMessage->publisherIdType = UA_PUBLISHERDATATYPE_UINT16;
-        networkMessage->publisherId.publisherIdUInt32 =
-            connection->config->publisherId.numeric;
-    } else if(connection->config->publisherIdType == UA_PUBSUB_PUBLISHERID_STRING) {
-        networkMessage->publisherIdType = UA_PUBLISHERDATATYPE_STRING;
-        networkMessage->publisherId.publisherIdString =
-            connection->config->publisherId.string;
-    }
+    networkMessage->publisherIdType = connection->config->publisherIdType;
```
jpfr, line 906: Please add a comment that this is a shallow copy of the string.
And that the config needs to be stable until the message is sent.
And that we must not "clean up" the PublisherId in network message after sending.

Also add such a comment to the header file where the networkmessage is defined.

jpfr (commented):
Looks good.
A few quick comments remain.

src/pubsub/ua_pubsub_ns0.c
MartinaFelberFEM: yes shallow copy is intended, because deep copy happens at next line within `UA_Server_addPubSubConnection`

`removeConnectionAction` calls `UA_Server_removePubSubConnection()` calls `UA_PubSubConnection_clear()` calls `UA_PubSubConnectionConfig_clear()` which frees string

src/pubsub/ua_pubsub_reader.c
MartinaFelberFEM: Thanks for review

With fast-path configuration string PublisherId was not supported for the Reader before, that's why I haven't changed it.
But it was supported for the Publisher ... and I haven't understood the reason, because even if string is dynamic the PublisherId is not a dynamic field which changes at runtime ...

UA_Server_freezeReaderGroupConfiguration() -> UA_DataSetReader_generateNetworkMessage()
I have removed the check
```
    if(!UA_DataType_isNumeric(dataSetReader->config.publisherId.type))
        return UA_STATUSCODE_BADNOTSUPPORTED;
```

checkReaderIdentifier() checks string PublisherIds

But subscribing fails with
`PubSub receive. Unknown field type.`

MartinaFelberFEM: I have a problem with fast-path Subscriber string PublisherId implementation:

- I couldn't remove the PublisherId String clear within `UA_NetworkMessage_clear()`, because the `UA_NetworkMessageHeader_decodeBinary` - `UA_String_decodeBinary` allocates memory at every receive() cycle

- therefore I also had to alloc memory (deep copy) at `UA_Server_freezeReaderGroupConfiguration()` - `UA_DataSetReader_generateNetworkMessage` because it uses `UA_NetworkMessage_clear()` at `unfreezeReaderGroupConfiguration`

- But `UA_NetworkMessage_updateBufferedNwMessage` allocates memory with `UA_String_decodeBinary()` into the same prepared (generated at freeze) publisherId string variable ... and in the worst case it uses buffer malloc and at teardown we have to call clear() because otherwise we would not clear the last received string

MartinaFelberFEM: If I would immediately clear the PublisherId String after UA_DataSetReader_generateNetworkMessage() after the size has been calculated then I would not have a problem during subscribe cycle, but at unfreezeReaderGroup I may not free anything
but there the same UA_NetworkMessage_clear() function is used as for subscribe Cycle ...
and then we would free invalid memory

keba-estr:
FYI: I am trying to resume the work of this pull request by creating a pull request for the minor issues resolved by these commits.

keba-estr:
Work is continued in #5223

keba-estr:
@jpfr: Please close this pull request, it has been resolved with the other pull requsts.

andreasebner:
Work is continued in #5223 and #5220

PR #4274, opened 2021-03-23
fix(client) Async connect on windows on small run_iterate timeout

wbitesi:
UA_Client_connectAsync and UA_Client_run_iterate fail on windows when timeout is too small to achieve the TCP handshake. If UA_Client_run_iterate is called twice while the TCP connection opens, return codes for connect differ from POSIX connect.


jpfr:
Thanks. Will merge this 1.1 -> 1.2 -> master.

PR #4275, opened 2021-03-23
fix(client) Async connect on windows on small run_iterate timeout

wbitesi:
UA_Client_connectAsync and UA_Client_run_iterate fail on windows when timeout is too small to achieve the TCP handshake. If UA_Client_run_iterate is called twice while the TCP connection opens, return codes for connect differ from POSIX connect.

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -734,6 +734,17 @@ UA_ClientConnectionTCP_poll(UA_Connection *connection, UA_UInt32 timeout,
     int error = UA_connect(connection->sockfd, tcpConnection->server->ai_addr,
                            tcpConnection->server->ai_addrlen);
 
+#ifdef _WIN32
+    /* https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-connect
+     * - Successfull async connect will fail but set WSALastError() to WSAEISCONN
+     * - Checking on WSAEWOULDBLOCK (UA_ERR_CONNECTION_PROGRESS) is not enough, if async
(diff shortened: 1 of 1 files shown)


jpfr:
Closing. Will be merged from the parallel PR 1.1 -> 1.2 -> master.

jpfr:
The PR #4277 is not running. The change in this PR hangs the windows unit tests.
Please have a look at #4277.

wbitesi:
I have updated the implementation using getsockopt on windows too. Now all tests are running except check_server_discovery. I've never worked with discovery/register so I'm not too aware what is going on in the test cases (especially with the setup/teardown and the UA_faketime specialities in the unit tests).

In one case, the client does not connect because the socket fails with:

2021-03-29T12:15:09.4995031Z   29: [1601-01-01 00:02:43.290 (UTC+0000)] warn/network	Connection to opc.tcp://localhost:4840 failed with error: No connection could be made because the target machine actively refused it.
2021-03-29T12:15:09.4996113Z.

I don't know why the connection reject comes from the server or if this behavior is wanted in the test case. I cannot reproduce the failure on my local system; the failure just does not occur after multiple runs. @jpfr, can you take a look at the test failure and give me a hint what may be the problem with the test case?

jpfr:
The CI looks good now. Great!

Can you explain why it is necessary to not early-abort if create_socket is false?
That should be put in a comment for future developers.

jpfr:
Nice one!
This kind of hardening is very valuable and requires deep domain knowledge.
Good to have it in.

PR #4193, opened 2021-02-20
[WIP] feat(core): Add EventLoop structure

jpfr:

PR #4276, opened 2021-03-23
feat(client) Expose client socket for use with external eventloops

wbitesi:
Using an external eventloop (e.g. libevent) it would be useful to be able to monitor if server responses arrive. Then UA_Client_run_iterate(client, 0) can be used which is especially useful if an application uses multiple UA_Clients. I did not expose UA_Connection because there is internal state that should not be visible to the user of the open62541 library.

I would love to see this exposed in version 1.1 too, because i currently use https://conan.io/center/open62541 which currently supports only 1.0.X and 1.1.X packages.

### DIFF
--- a/include/open62541/client.h
+++ b/include/open62541/client.h
@@ -200,6 +200,19 @@ UA_Client_getContext(UA_Client *client) {
     return UA_Client_getConfig(client)->clientContext; /* Cannot fail */
 }
 
+/* Returns the current socket of the connection. Returns UA_SOCKET_INVALID, if
+ * connection is not opened. The user should monitor the state of the connection
+ * ``UA_Client_getState`` or ``UA_ClientConfig.stateCallback`` to identify when
+ * a socket becomes available or invalid.
+ *
+ * Useful for calling ``UA_Client_run_iterate`` with timeout 0, e.g. when having
+ * multiple UA_Client handles in an application. If the socket activity is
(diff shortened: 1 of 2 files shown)


jpfr:
Sockets are POSIX specific. There are embedded platforms that have TCP, but not via sockets.
It would be better to have this functionality as part of the architecture-specific API.

Note that there is an upcoming PR #4193 that introduces an actual event loop concept to open62541.
With the ability to integrate into your custom event loop.

Until then, can you workaround by accessing the private API to get the socket?

wbitesi:
I know that this PR is not really ideal. I sadly cannot access the private API since I use just the library (public API) and do not modify the open62541 code myself (just consuming the conan package). I use libevent to drive my own eventloop and work around the interface of UA_Client_run_iterate which causes me to frequently call the function without knowing if I have to or not.

Using the public API, i've tried to get the socket using UA_ClientConfig.initConnectionFunc and UA_ClientConfig.pollConnectionFunc. This was not possible because too few types are available through the public API to implement those.

I've seen your PR and also the PR https://github.com/open62541/open62541/pull/3409.  Both PRs look promising with regard to having multiple UA Clients in one thread/eventloop.  I know that this PR exposing the UA_SOCKET is far away from being a perfect solution and is probably obsolete when an eventloop PR finally lands.

jpfr:
The new EventLoop approach on master allows this and much more.

Issue #4223, opened 2021-03-04
AddPublishedItems method call throws segmentation fault


opcua-tsn-team-kalycito:
Hi,

While we are calling the method AddPublishedItems(NodeId:UA_NS0ID_DATASETFOLDERTYPE_ADDPUBLISHEDDATAITEMS) from OPC UA client we are receiving the segmentation fault.

The segmentation fault occurs while clearing the UA_PublishedVariableDataType ([UA_PublishedVariableDataType_clear(variablesToAddField)](https://github.com/open62541/open62541/blob/master/src/pubsub/ua_pubsub_ns0.c#L712)).

AddPublishedItems method call contains four input arguments out of which three input arguments provided from the client reaches the server successfully
but only the fourth input argument structure datatype UA_PublishedVariableDataType shows the garbage value. 

We suspect that the encoding and decoding of the structure UA_PublishedVariableDataType might be the problem. The problem only occurs if we call the AddPublishedItems method from the client but the issue does not occur if we call the method from server.

## Background Information / Reproduction Steps

Used CMake options:

```bash
cmake -DUA_BUILD_EXAMPLES=ON -DUA_ENABLE_PUBSUB=ON -DUA_ENABLE_PUBSUB_ETH_UADP=ON -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_PUBSUB_INFORMATIONMODEL=ON -DUA_ENABLE_PUBSUB_INFORMATIONMODEL_METHODS=ON ..
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): master
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Linux
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

opcua-tsn-team-kalycito:
We spent sometime in understanding the root cause for the issue and found that the issue occurs in decodeBinaryStructure function.
 
So in our case we have to decode the UA_CallRequest structure,
typedef struct {
    UA_RequestHeader requestHeader;
    size_t methodsToCallSize;
    UA_CallMethodRequest *methodsToCall;
} UA_CallRequest;

which further has many structures within it which gets decoded separately.
 
[...]

PR #4278, opened 2021-03-23
[REVIEW] fix(pubsub): AddPublishedItems method call throws segmentation fault #4223

opcua-tsn-team-kalycito:
- While calling the addPublishedItems method from the client the last argument (UA_PublishedVariableDataType) gets corrupted
- Fixed by changing the last argument (UA_PublishedVariableDataType) from variant type to extension object


jpfr:
@andreasebner Ready to merge? The docker build problem can be ignored...

PR #4282, opened 2021-03-25
Added the ability to create types with the same name.

Johnsik:
Added the ability to create types with the same name, but in different namespaces.


jpfr:
This seems to break some unit tests.
Can you check?

PR #4283, opened 2021-03-26
fix(pubsub) interface selection for multicast

Fischer-Thomas:
- use the given interface from the application when subscribing to a multicast addr

### DIFF
--- a/plugins/ua_pubsub_udp.c
+++ b/plugins/ua_pubsub_udp.c
@@ -18,6 +18,7 @@
 typedef struct {
     int ai_family;                    /* Protocol family for socket. IPv4/IPv6 */
     struct sockaddr_storage *ai_addr; /* https://msdn.microsoft.com/de-de/library/windows/desktop/ms740496(v=vs.85).aspx */
+    struct sockaddr_storage *intf_addr;
     UA_UInt32 messageTTL;
     UA_Boolean enableLoopback;
     UA_Boolean enableReuse;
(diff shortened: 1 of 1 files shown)


plugins/ua_pubsub_udp.c
```diff
@@ -177,6 +178,20 @@ UA_PubSubChannelUDPMC_open(const UA_PubSubConnectionConfig *connectionConfig) {
     memcpy(channelDataUDPMC->ai_addr, rp->ai_addr, sizeof(*rp->ai_addr));
     newChannel->handle = channelDataUDPMC; /* Link channel and internal channel data */
 
+    if(address.networkInterface.length > 0) {
+        channelDataUDPMC->intf_addr = (struct sockaddr_storage *)
+            UA_calloc(1, sizeof(struct sockaddr_storage));
+        if(!channelDataUDPMC->intf_addr) {
+            UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
+                        "PubSub Connection creation failed. Out of memory.");
+            UA_close(newChannel->sockfd);
```
jpfr, line 187: Clean up `channelDataUDPMC->ai_addr`  in this error case.
In the error case below clean up both `channelDataUDPMC->ai_addr` and  `channelDataUDPMC->intf_addr`.

plugins/ua_pubsub_udp.c
```diff
@@ -18,6 +18,7 @@
 typedef struct {
     int ai_family;                    /* Protocol family for socket. IPv4/IPv6 */
     struct sockaddr_storage *ai_addr; /* https://msdn.microsoft.com/de-de/library/windows/desktop/ms740496(v=vs.85).aspx */
```
jpfr, line 20: I would prefer if both `sockaddr_storage` were not pointers but direct members.
That removes a lot of the allocation and cleanup logic.

Fischer-Thomas: I agree. I will update the PR and remove the pointers from this struct.

jpfr:
Great contribution!

PR #4284, opened 2021-03-26
PubSub: provide optional usage of memory pool for time critical subscriber parts

MartinaFelberFEM:
Hi!

I have had a look at PubSub Writer and Reader performance on my test hardware.
I am using a custom PubSub  application callback like in your realtime examples with fast-path (ExternalValueCallback) and try to publish/subscribe 256 variables within 1 ms.

I saw that the Publisher needs about 30-40 us and the Reader about 150-160 us to process 256 INT32 variables.
I was wondering why the Reader needed so much more time compared to the Writer.
(My backend variable write / read implementation takes an equal amount of time)

I measured the subscribeCallback steps:

- receive() network message  -> 6 us
- updated buffered network message (decode fields) - 60 us
- write data (memcpy the data to fast-path subscriber variables - ExternalValueCallback pointer) - 8 us
- cleanup (free dataset fields) - 60 us

Decoding the fields and cleanup takes the most time (60 + 60 us)
I thought with fast-path configuration the Reader would directly decode the data to the pointer given by the ExternalValue callback or StaticValueSource.

But this is not the case, because UA_decodeBinary always allocates memory and that's why the cleanup is also necessary.
So in one cycle I have 256 mallocs and frees.

- decodeBinary allocates mem
- then memory is copied to subscriber var
- then the memory is freed

This would not be necessary if UA_decodeBinary would only allocate memory if needed.
Then we could directly decode the data to the given StaticValueSource pointer and would not need to allocate the memory, copy the data and free the memory.

I have implemented a dirty workaround by checking in the UA_decodeBinary function if the destination is a Variant and if the UA_VARIANT_DATA_NODELETE storageType is set. In this case I do not allocate memory and do not free it and the Reader is now almost as fast as the Writer with about 40 us for one subscribe cycle.

I know that my implementation is not usable, it was just a performance test for me.

But now I don't know how to change the UA_decodeBinary function to only allocate memory if necessary.
Or if I should add a second UA_decodeBinary_noAlloc function or something like this ...

Can you help me with this? Can you give me any suggestions?

Many thanks and best regards


MartinaFelberFEM:
I could write a UA_Variant_decodeBinary_noAlloc() function, which does not allocate memory and is only called at fast-path subscriber UA_NetworkMessage_updateBufferedNwMessage().

And delete the Variant_clear, DataValue_clear calls at subscribeCallback() then I wouldn't have to modify the UA_clear() ...

jpfr:
Hey Martina,

this is a good push. However, it involves adding hacks to the decoder.
I think this can be done nearly hack-free!

The option UA_ENABLE_MALLOC_SINGLETON switches all uses of malloc/free et al. to global method pointers.
If you are single-threaded, then these can be set at runtime.

There is a simple arena-based malloc here: https://github.com/open62541/open62541/blob/0db99235b08359a3bb5844cfc2d02d36eadd70c0/examples/pubsub_realtime/bufmalloc.c

In the realtime case, you could switch to the arena-based malloc and restore the normal malloc afterwards.
Just make sure not to free memory that now points to the arena.

MartinaFelberFEM:
Hi!
Many thanks for your response!
I hope I have understood your suggestions.

I would need a multithreading safe implementation.

I have had a look at the bufmalloc.c implementation and I think it can't be used in a multithreading environment with multiple Publishers and Subscribers out of the box. Because there's no locking. And in the examples it's also only used with publisher only (pubsub_interrupt_publish_tsn.c and pubsub_interrupt_publish.c). 

Except I would use the implementation and add a locking mechanism. And it's interesting how long the implementation with the publisher works, because resetMembuf() is never called and there's no free (I think), so sometime the publisher should be out of memory. But you probably could call resetMembuf() and then it would be fine and it's just an example.

But still I would like to avoid the memory allocation (regardless if it's on heap or stack), the copying and the freeing.
Because it's unnecessary. The concept behind ExternalValue should be that the pointer is provided by the application.

And everything is already beautifully prepared (buffer, callbackpointer etc.). It's just not used to it's full potential like in the Writer ...

jpfr:
Hum. Okay. Multi-threaded.

You could use the singleton malloc pointer as a thread-local variable.
So every thread can switch its malloc individually.

Its a bit more work but looks doable.

1. Add a UA_THREAD_LOCAL define that can be adjusted to architectures and compilers.
https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html
2. Use UA_THREAD_LOCAL for the singleton malloc pointers.
3. Extend or fork the membuf malloc to avoid races between threads.
4. Switch in and out of custom malloc in the PubSub realtime path.

MartinaFelberFEM:
Thanks for your help!
I will try that and make time measurements.

But I have changed my hack implementation and added a function UA_Variant_decodeBinary_noAlloc.
And I am working on UA_DataValue_decodeBinary_noAlloc and try to support arrays and structures with no allocation decoding.
Because I would still prefer to not handle any memory operations in the realtime task at all if it can be avoided.

Would these noAlloc functions be ok? Or do you have any other suggestions?
Have there ever been any thoughts about Variant_decoding with preallocated memory?

MartinaFelberFEM:
I try to test it with the examples server_pubsub_publisher_rt_level and server_pubsub_subscriber_rt_level but the subscriber does not successfully receive messages. -> PubSub receive. Unknown field type.

Setting the dataSetFieldContentMask in subscriber to RAWDATA as is done in the Publisher did not help, only removing the RAWDATA option.

MartinaFelberFEM:
@jpfr FYI:
Your original idea of switching the thread-local memory like this:

useMemBuf()
updateBufferedNwMessage
useNormalAlloc()
DataSetReaderProcess
-> only clear if build option BUFALLOC is not defined

did not quite work. I need the clear() methods, because they memset the Variant data pointer. If the Variant is not cleared, the UA_Server_unfreezeReaderGroupConfiguration function tries to free it and this causes a segfault.

I am not sure but I think we can't avoid the clear() calls for every field :(
I have tried to adapt the UA_DataSetMessage_clear() function but this does not work.

The memory is allocated at subscribeCallback() and the pointer has to set back to 0 ...

MartinaFelberFEM:
Regarding the receiveMsgBuffer -> I know you showed me the UA_Byte static thread-local array, but if I use a byte array I would have to change the receive() interface, because it expects a ByteString ...

jpfr:
You can use _init instead of _clear to reset.

MartinaFelberFEM:
Thanks, readability is a bit worse with Variant_init() but I hope it's ok

If you are interested: I measured the reader performance
Before optimization: 160 us
EncodeBinary hack: 40 us
MemPool implementation: 60 us

I am very happy about the solution now, many thanks for your help!

src/pubsub/ua_pubsub_reader.c
```diff
@@ -25,9 +25,15 @@
...
 #endif
 
+#ifdef UA_ENABLE_PUBSUB_BUFMALLOC
+#include "ua_pubsub_bufmalloc.h"
+#endif
+
 #define UA_MAX_SIZENAME           64  /* Max size of Qualified Name of Subscribed Variable */
 #define MIN_PAYLOAD_SIZE_ETHERNET 46
 
+#define READERGROUP_BUFFER_SIZE   4096
```
jpfr, line 35: If the buffer has a fixed size anyway, then I prefer a thread-local static buffer instead of the heap.
Less memory overhead per connection.
And less pressure on the cpu cache.

jpfr (changes requested):
Looks good! Some last comments remain.

jpfr (approved):

jpfr:
Great one!

PR #4289, opened 2021-03-30
Thread local singleton malloc

jpfr:


MartinaFelberFEM:
Thanks for the PR!

Does the UA_THREAD_LOCAL define have to be coupled to the UA_MULTITHREADING define?

I think the malloc singleton has only been used for PubSub until now with bufMalloc.h .c implementation and previously UA_MULTITHREADING > 0 was not necessary for PubSub.
I am not sure what UA_MULTITHREADING changes within the server, or if it's necessary.

And if the bufMalloc implementation would use thread-local memory for it's threads then maybe a locking would not be necessary

jpfr:
> Does the UA_THREAD_LOCAL define have to be coupled to the UA_MULTITHREADING define?

Yes. Without multithreading support, it doesn't make much sense to have thread local variables.
The UA_THREAD_LOCAL define will be simply empty in that case.

MartinaFelberFEM:
Thanks for your answers!

I have tried to use a thread_local memory for every thread. But when I use the __thread specifier I get a segfault.

I have simplified it and tested it without thread local memory -> So just your declaration of UA_THREAD_LOCAL and no changes to bufmalloc.c and still get a segfault.
I have tested it here: https://github.com/MartinaFelberFEM/open62541/commit/642d7d4690ea1f34e213003124073c2d4f807856

It seems that there's a bug in GLIBC (at least I found some issues regarding segfaults and usage of __thread)

I am using gcc 6.3.0 in a container. I'll try another one.

Just for your information. Not sure if this happens at your test platform as well. Or if the pubsub realtime examples are integrated in your test environment.

jpfr:
You should extend the bufmalloc implementation internally with thread_local definitions as well.
Otherwise you get a race condition internally.

Did you check whether the initial values of the Malloy pointers are set correctly in all threads?

MartinaFelberFEM:
I forgot to add the UA_THREAD_LOCAL definition after I have simplified the example.
But it still crashed with a segfault.
The UA_mallocSingleton pointer had the value <__GI___libc_malloc>, so I thought it would be correct.

But: a new gcc version solved the problem. I am now using 8.3.0 and it works :)

Many thanks for your help! :)

PR #4292, opened 2021-03-31
Update ua_services_discovery.c

jackybek:
At line 624 change the condition from : if (client->connection.state != UA_CONNECTIONSTATE_CLOSED) to
if (client->connection.state != UA_CONNECTIONSTATE_ESTABLISHED)


jpfr:
Rebase on master...

jpfr:
The CI builds break.
There seems to be more going on.

Possibly we have to  make discovery fully asynchronous …

PR #4294, opened 2021-03-31
fix Fix memory leaks detected by AddressSanitizer.

nilsRoettger:
AddressSanitizer detected two memory leaks when using open62451. 

- The first one occurs when string IDs are used when creating child nodes. If this is the case, the NodeId must be cleaned up. 
- The second case occurs when using Encryption, when calling UA_AccessControl_default, new memory is allocated, but before that it is not checked if memory has already been allocated elsewhere for the UA_AccessControl struct. Therefore, it must be ensured that the UA_AccessControl is clean at the beginning.

### DIFF
--- a/plugins/ua_accesscontrol_default.c
+++ b/plugins/ua_accesscontrol_default.c
@@ -253,6 +253,7 @@ UA_AccessControl_default(UA_ServerConfig *config, UA_Boolean allowAnonymous,
     UA_LOG_WARNING(&config->logger, UA_LOGCATEGORY_SERVER,
                    "AccessControl: Unconfigured AccessControl. Users have all permissions.");
     UA_AccessControl *ac = &config->accessControl;
+    clear_default(ac);
     ac->clear = clear_default;
     ac->activateSession = activateSession_default;
     ac->closeSession = closeSession_default;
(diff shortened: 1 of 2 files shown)


plugins/ua_accesscontrol_default.c
```diff
@@ -253,6 +253,7 @@ UA_AccessControl_default(UA_ServerConfig *config, UA_Boolean allowAnonymous,
     UA_LOG_WARNING(&config->logger, UA_LOGCATEGORY_SERVER,
                    "AccessControl: Unconfigured AccessControl. Users have all permissions.");
     UA_AccessControl *ac = &config->accessControl;
+    clear_default(ac);
```
jpfr, line 256: We don't know if the previous AccessControl config is from the same plugin.

Here you should check if `ac->clear` is already set. If yes, use that to clean up.
Otherwise, there is no need to clear.

jpfr (changes requested):
Looks good!
See the comment for AccessControl.

Please split into two commits, as separate subsystems are concerned.

jpfr:
Ah, and you have to click the CLA agreement in the list of CI tests.

nilsRoettger:
I have added you comments and created a new Pull Request: https://github.com/open62541/open62541/pull/4308

PR #4308, opened 2021-04-07
Fix: Fix memory leaks detected by AddressSanitizer.

nilsRoettger:
AddressSanitizer detected two memory leaks when using open62451.

- The first one occurs when string IDs are used when creating child nodes. If this is the case, the NodeId must be cleaned up.
- The second case occurs when using Encryption, when calling UA_AccessControl_default, new memory is allocated, but before that it is not checked if memory has already been allocated elsewhere for the UA_AccessControl struct. Therefore, it must be ensured that the UA_AccessControl is clean at the beginning.


jpfr:
Thanks for this!

PR #4315, opened 2021-04-08
Export malloc singleton function pointers

MartinaFelberFEM:
Export malloc singleton function pointers, otherwise there are not
visible if we use shared object library build.


jpfr:
👍

PR #4316, opened 2021-04-08
Update the localhost.cnf

opcua-tsn-team-kalycito:
As per discussion in the OpenSSL issues [forum](https://github.com/openssl/openssl/issues/7754), it says that OpenSSL version 1.1.1 does not require RANDFILE configurations anymore.

 - RANDFILE configuration causes error: 'Can't load /root/.rnd into RNG' (as shown below)
image

 - After the fix the error get resolved (as shown below)
image


jpfr:
Great, thanks.

PR #4322, opened 2021-04-09
fix(pubsub): remove platform dependencies

### LINKED ISSUE #4324: PubSub Ethernet would not compile on specific platforms
## Description
PubSub Ethernet plugin seems to be designed and tested only on Linux and VxWorks. It's not consistent in using architecture wrappers vs. direct system calls. As a result it doesn't compile on our embedded platform.

Used CMake options:
UA_ENABLE_PUBSUB
UA_ENABLE_PUBSUB_ETH_UADP

## Checklist

 - [ ] open62541 Version (release number or git tag): afb74a26f98ef63b17175085fda79763db939c0e
 - [ ] Operating system: custom platform

MurzynJ:
Two new wrappers UA_recvmsg and UA_ioctl are added and reflected to architecture specific files.

In pubsub_ethernet.c replace the direct system calls with wrappers in all places which caused troubles on our platform. Some more refactoring may be needed to support any other platforms in future.

By the way, in ua_pubsub_reader.c replace the platform dependent alloca with an already existing wrapper UA_STACKARRAY.

Closes #4324

### DIFF
--- a/plugins/ua_pubsub_ethernet.c
+++ b/plugins/ua_pubsub_ethernet.c
@@ -14,7 +14,11 @@
 #include <open62541/plugin/log_stdout.h>
 #include <open62541/plugin/pubsub_ethernet.h>
 
-#if defined(__vxworks) || defined(__VXWORKS__)
+#if !defined(UA_ARCHITECTURE_POSIX) && !defined(UA_ARCHITECTURE_VXWORKS)
+/* For anything else than Linux or VxWorks which are specifically handled below,
+ * depend only on headers included by the architecture layer.
+ */
+#elif defined(__vxworks) || defined(__VXWORKS__)
 #include <netpacket/packet.h>
 #include <netinet/if_ether.h>
 #include <ipcom_sock.h>
@@ -762,7 +766,7 @@ UA_PubSubChannelEthernet_open(const UA_PubSubConnectionConfig *connectionConfig)
     }
 #endif
 
-    if(ioctl(sockFd, SIOCGIFINDEX, &ifreq) < 0) {
+    if(UA_ioctl(sockFd, SIOCGIFINDEX, &ifreq) < 0) {
         UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
            "PubSub connection creation failed. Cannot get interface index.");
         UA_close(sockFd);
@@ -773,7 +777,7 @@ UA_PubSubChannelEthernet_open(const UA_PubSubConnectionConfig *connectionConfig)
     channelDataEthernet->ifindex = ifreq.ifr_ifindex;
 
     /* determine own MAC address (source address for send) */
-    if(ioctl(sockFd, SIOCGIFHWADDR, &ifreq) < 0) {
+    if(UA_ioctl(sockFd, SIOCGIFHWADDR, &ifreq) < 0) {
         UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER,
             "PubSub connection creation failed. Cannot determine own MAC address.");
         UA_close(sockFd);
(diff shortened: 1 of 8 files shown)


jpfr:
Looks very good! Just a last build issue Windows:
```
D:\a\1\s\build\open62541.h:30938:39: error: 'struct msghdr' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
30938 | ssize_t UA_recvmsg(int sockfd, struct msghdr *msg, int flags);//equivalent to posix recvmsg implementation
```

MurzynJ:
Ah, right. I forgot about Windows.
But here I'm confused. Winsock doesn't seem to have recvmsg. So, the Ethernet plugin was not able to compile on Windows anyway in the past, is it?
There is an equivalent function in Winsock, called WSARecvMsg, but it's not straightforward to use it. Its function pointer has to be obtained in runtime, and also the parameters are not directly compatible with posix version. I would prefer not to port it myself as I'm not a Windows expert.
Shall we find someone with Windows/Winsock background who can have a look at it, or can we just say that Ethernet plugin doesn't compile on Windows (as it seemed to be the case so far)?

jpfr:
No need to make the Ethernet windows-compatible if that was not the case prior.
But this introduced a build error even with Ethernet PubSub disabled.
Have a look at the CI results. Should be straightforward to fix.

plugins/ua_pubsub_ethernet.c
```diff
@@ -814,7 +818,8 @@ UA_PubSubChannelEthernet_open(const UA_PubSubConnectionConfig *connectionConfig)
         }
     }
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,19,0)
+#if defined(KERNEL_VERSION)
```
jpfr, line 821: Please combine the two #if into one.
Just concatenation the expressions with &&.

plugins/ua_pubsub_ethernet.c
```diff
@@ -926,7 +932,8 @@ UA_PubSubChannelEthernet_unregist(UA_PubSubChannel *channel,
     return UA_STATUSCODE_GOOD;
 }
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,19,0)
+#if defined(KERNEL_VERSION)
```
jpfr, line 935: Merge the two #if.

plugins/ua_pubsub_ethernet.c
```diff
@@ -1049,7 +1057,8 @@ UA_PubSubChannelEthernet_send(UA_PubSubChannel *channel,
     /* copy payload of ethernet message */
     memcpy(ptrCur, buf->data, buf->length);
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,19,0)
+#if defined(KERNEL_VERSION)
```
jpfr, line 1060: Merge the #if.

jpfr (commented):
Looks good now.
Just a last cosmetic improvement in the comments...

plugins/ua_pubsub_ethernet.c
MurzynJ: I've done it on purpose in separate #ifs. At least my compiler (armclang 6.6) complains when I concatenate it. See

ua_pubsub_ethernet.c:830:55: error: function-like macro 'KERNEL_VERSION' is not defined
#if defined(KERNEL_VERSION) && (LINUX_VERSION_CODE >= KERNEL_VERSION(4,19,0))

MurzynJ: Alternatively we can define a wrapper macro that would do both checks in 1 line:

#if defined(KERNEL_VERSION)
#define LINUX_AND_KERNEL_VERSION_HIGHER_OR_EQUAL_4_19_0 (LINUX_VERSION_CODE >= KERNEL_VERSION(4,19,0))
#else
#define LINUX_AND_KERNEL_VERSION_HIGHER_OR_EQUAL_4_19_0 0
#endif

jpfr: Ok. Then lets keep as-is.

jpfr:
Nice 👍 
Merged.

PR #4323, opened 2021-04-09
fix(pubsub): source vs destination address filtering on reception

### LINKED ISSUE #4325: In PubSub Ethernet, unicast frames would not be received due to bad filtering.
## Description
In PubSub Ethernet plugin preliminary filtering of incoming frames is done using the address of PubSubConnection.
Currently the unicast frames are dropped because the destination address doesn't match the configured PubSubConnection's address.

Logic has to be different in case of multicast or unicast PubSubConnections. In case of multicast, the connection's address has to match the frame's destination. In case of unicast, the connection's address has to match the frame's source. Behavior in case of multicast vs. unicast is not clearly specified in the current version of Part14, but that's the only logical way to interpret the address field of PubSubConnection. This point is known to specification authors and it will be clarified in the future releases of the spec.

Used CMake options:
UA_ENABLE_PUBSUB
UA_ENABLE_PUBSUB_ETH_UADP

## Checklist

 - [ ] open62541 Version (release number or git tag): afb74a26f98ef63b17175085fda79763db939c0e
 - [ ] Operating system: any

MurzynJ:
On frame reception, check if current PubSubConnection is for multicast or unicast and based on that apply correct filter condition.

Closes #4325

### DIFF
--- a/plugins/ua_pubsub_ethernet.c
+++ b/plugins/ua_pubsub_ethernet.c
@@ -1188,7 +1188,9 @@ UA_PubSubChannelEthernet_receive(UA_PubSubChannel *channel, UA_ByteString *messa
             break;
         }
         /* Make sure we match our target */
-        if(memcmp(eth_hdr.ether_dhost, channelDataEthernet->targetAddress, ETH_ALEN) != 0) {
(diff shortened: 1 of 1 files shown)


jpfr:
Good find!

PR #4330, opened 2021-04-13
[Review] refactor(pubsub): PubSub method call

opcua-tsn-team-kalycito:
Implementation of ns0 method calls for Publisher and Subscriber

-  Add/Remove PubSub nodes at runtime


jpfr:
Thanks.
You automatically activate new writer groups via
UA_Server_setWriterGroupOperational.

That behavior of automatically activating should be consistent.
Independently of whether it gets added via a method call or internally.

jpfr:
So far the idea was to *not* auto activate.
So the user has the chance to fully configure before the first message is sent out.

PR #4352, opened 2021-04-20
[REVIEW] fix(*): set node's backendType in setValueCallback

robertns:
Set node's backendType field to `UA_VALUEBACKENDTYPE_DATA_SOURCE_CALLBACK` (similarly as in`setExternalValueSource()`). Currently this information is not stored.


jpfr:
@andreasebner ok?

PR #4362, opened 2021-04-21
Added the ability to create types with the same name, but in differen…

Johnsik:
Added the ability to create types with the same name, but in different datasets.


jpfr:
Thanks for this.
The CI fails with `\examples\nodeset\server_testnodeset.c(47): error C2065: 'UA_Point': undeclared identifier`.
Can you fix that and add some documentation on the new behavior?

Johnsik:
I'm build on Windows 10 and MSVC 2019.
My cmake string is `cmake .. -DUA_BUILD_EXAMPLES=ON -DCMAKE_TOOLCHAIN_FILE=C:\Utils\vcpkg\scripts\buildsystems\vcpkg.cmake`
I foresaw that there should be errors like this, but I don't see this example (with server_testnodeset.c) in the build directory.
What I do wrong with cmake?

PR #4367, opened 2021-04-23
feat(server) Made it possible to control if base nodes is initialized.

ED19F:
Added flags to be able to controll if base nodes is initialized. This 
feature makes it easier to work with dynamic information models, 
as you do not always want to use a predefined node store.

### DIFF
--- a/examples/server_ctt.c
+++ b/examples/server_ctt.c
@@ -1238,6 +1238,9 @@ int main(int argc, char **argv) {
         UA_String_fromChars("urn:open62541.server.application");
 
     config.shutdownDelay = 5000.0; /* 5s */
+    config.createNs0BaseNodes = true;
+    config.createNs0Nodes = true;
+    config.initNs0Nodes = true;
 
     server = UA_Server_newWithConfig(&config);
     if(!server) {
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -184,6 +184,12 @@ struct UA_ServerConfig {
 
     /* Nodestore */
     UA_Nodestore nodestore;
+    UA_Boolean   initBaseNodes;
+
+    /* Ns0 */
+    UA_Boolean createNs0BaseNodes;
+    UA_Boolean createNs0Nodes;
+    UA_Boolean initNs0Nodes;
 
     /* Certificate Verification */
     UA_CertificateVerification certificateVerification;
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -134,6 +134,9 @@ setDefaultConfig(UA_ServerConfig *conf) {
         conf->logger = UA_Log_Stdout_;
 
     conf->shutdownDelay = 0.0;
+    conf->createNs0BaseNodes = true;
+    conf->createNs0Nodes = true;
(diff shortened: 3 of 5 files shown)


jonasgreen88:
@jpfr This is a first proposal to discuss what flags we need to control the initialization. When we have decide the scope we can finalize with complete unittests etc.

The use cases we see are:
- The nodestore is pre-initialized and only initialization of NS0 nodes are needed (`createNs0BaseNodes=false`, `createNs0Nodes=false`, `initNs0Nodes=true`)
- The bare minimum of nodes are initialized to e.g. allow to load a NS0 file of the user's choice in runtime before starting the server (for example using the NodesetLoader project). (`createNs0BaseNodes=true`, `createNs0Nodes=false`, `initNs0Nodes=false`)
  - Do you think we can separate the NS 0 initialization (set callbacks, values etc.) to allow it to be executed after the information model has been loaded in runtime by the user (Between the calls to the `UA_Server_new()` and the `UA_Server_Run_Startup()`)

jpfr:
That's a good feature.
But it needs a clearly documented rationale.

The naming is a bit ambiguous.
We use the term "bootstraping" in the code. This could also be a good name for configuration of the initialisation of the "innermost" nodes.

> Currently a NS0 nodeset initialized in runtime after server initialization won't be initialized with callbacks etc.

We could do this initialization in server_run_startup.
So additional work on the sever can be perfomed between creating the server and starting it up.

jonasgreen88:
Thanks for a positive response! We got some more questions and thoughts to discuss before continue the work.

> That's a good feature.
> But it needs a clearly documented rationale.
> 
> The naming is a bit ambiguous.
> We use the term "bootstraping" in the code. This could also be a good name for configuration of the initialisation of the "innermost" nodes.

With _"innermost" nodes_ you mean `createNs0BaseNodes` or also `createNs0Nodes`? Is `bootstrapNs0BaseNodes` what you suggest?

Another thought I got when thinking about this solution for a while is that three different flags gives a lot of combinations, most of them are not wanted and some are even not possible and will result in error (as mentiond in previous post only two combinations are useful in our case). I got another idea that I belive is more user friendly, but before re-designing this PR I would like to discuss it. The idea is to replace these flags with an enumeration. The enumeration should have a set of choices depending on use case. The stack will then do the initialization of nodes based on configured use case. Example of enum elements could be:

- Bootstrap_NS0 // As it works today.
- Pre_loaded_NS0 // nodestore is preinitialized and only initialization of nodes are performed.
- Runtime_load_NS0 // Loads bare minimum (base address space nodes etc.) and do initialization of nodes in UA_Server_Run_Startup(). Allows runtime initialization of nodes between calls to UA_Server_New() and UA_Server_Run_Startup().

The solution won't be as flexible but less error prone I think.

> We could do this initialization in server_run_startup.
> So additional work on the sever can be perfomed between creating the server and starting it up.

I agree. Can we in this PR also break out the initilaization of nodes (dataSourceCallbacks, setting values, deleting nodes not supported by the stack and/or chosen profile etc.) from UA_Server_initNS0() to its own function? The function will either be called from UA_Server_initNS0() or from UA_Server_Run_Startup() depending on configuration?

PR #4371, opened 2021-04-27
Add nodeset prefix to nodeset types

Johnsik:
If I try to use nodeset for example TESTNODESET1 with type Point and the other nodeset TESTNODESET2 with different type Point I've get UA_Point in first case and also UA_Point and the second. So it will be two declaration of two different types with the same name.
I think thats wrong...
I'm suggest add nodeset prefix to all nodet types (except ns0).
In that case I will get in C code two declaration of two different types UA_TESTNODESET1_Point and UA_TESTNODESET2_Point.


GoetzGoerisch:
@Johnsik why do you need this?

For normative _NodeSets_ this behavior would be harmful?

Johnsik:
> @Johnsik why do you need this?
> 
> For normative _NodeSets_ this behavior would be harmful?

No. It will affected only to C names.

NoelGraf:
@jpfr
Is there something that prevents us from having the same type in two nodesets? If not, we should include this change. What do you think?

Johnsik:
> Is there something that prevents us from having the same type in two nodesets? If not, we should include this change. What do you think?

Sorry, my I correct you. Not the same type in two node sets, but the two different types with same non-qualified name in two nodesets.

PR #4376, opened 2021-04-29
fix. #include TYPE_ARRAYs from --import to types_{nodeset}_generated.h

Johnsik:
**types_{nodeset1}_generated.h** header can refer to types from **nodeset2** so it would be reasonably to add `#include "types_{nodeset2}_generated.h"` like it already done in **namespace_{nodeset1}_generated.h** header.
 
 Information of dependences we can get from the first part of --import argument (IMPORT_BSD).


jpfr:
Looks good!
But you should track the additional includes with a set instead of a list.
To prevent duplicates.

Johnsik:
> But you should track the additional includes with a set instead of a list.

Done!

jpfr:
Nice, thanks for this!

PR #4379, opened 2021-04-30
Chunk Padding

jpfr:


src/ua_securechannel_crypto.c
```diff
@@ -225,25 +225,28 @@ hideBytesAsym(const UA_SecureChannel *channel, UA_Byte **buf_start,
...
+     * before the signature. From that compute the maximum usable plaintext
+     * size. */
+    size_t maxEncrypted = (size_t)(*buf_end - *buf_start) +
         UA_SECURECHANNEL_SEQUENCEHEADER_LENGTH;
-    size_t max_blocks = plainTextMaxLength / plainTextBlockSize;
-    size_t max_overhead = max_blocks * (encryptedBlockSize - plainTextBlockSize);
+    size_t max_blocks = maxEncrypted / encryptedBlockSize;
     size_t paddingBytes = (UA_LIKELY(!extraPadding)) ? 1u : 2u;
-    *buf_end -= max_overhead + paddingBytes + signatureSize;
+    *buf_end = *buf_start + (max_blocks * plainTextBlockSize) -
```
mlgiraud, line 248: Are you sure this is correct?
I don't think we need to subtract the sequence header length here, since it is contained in the max_blocks. Might be wrong though.

jpfr: The Sequence Header is encrypted as well.
Let's say we use AES256 with a 32byte block size.
The payload including padding and the signature need to be 32byte-aligned.
So the sequence header has to stay here.

mlgiraud: Yeah but isnt the header already included here since it is used in the max_blocks calculation? It looks to me like it is included twice.

PR #4384, opened 2021-04-30
[WIP] Circular buffer implementation for 24*7 live plot

opcua-tsn-team-kalycito:
- Implement circular buffer for computing round-trip latency of pubsub_TSN applications 
- Two threads for latency csv computation and csv filewrite
- CSV with epoch time and latency in microseconds and nanoseconds
- To initiate long run, follow readme_circularBuffer.txt


jpfr:
Let’s keep this as a PR.
- Not under CI coverage 
- Nobody from the core team is qualified to maintain the munin scripts.
- The changes to the core library are very specific instrumentation and not platform independent.

Very good to have the Feature.
But unmaintainable. So keep in the PR and update when the timings are recomputed.

jpfr:
This is very outdated now with the recent refactoring of PubSub.

PR #4397, opened 2021-05-06
[REVIEW] Adding macros for checking error codes

JanSurft:
Adding the possibility to use specific macros to check for error codes.

#### Check without logging

The following macros do error-code or error-state checks without logging anything and just evaluating an additional parameter.
- `UA_CHECK`
- `UA_CHECK_MEM`

`UA_CHECK_MEM` can be used in cases where an `malloc/calloc/realloc` call return-value is checked.

Here the return code is checked and returns the error code itself if the check encounters an error code.
```c
static UA_StatusCode
foo() {
    UA_StatusCode rv = some_func();
    UA_CHECK(rv, return rv);

    return UA_STATUSCODE_GOOD;
}
```

If there is the need for a cleanup routine one can simply jump to an according label like in this example:

```c
static UA_StatusCode
foo() {
    void *data;

    UA_StatusCode rv = some_func_which_allocates_memory(data);
    // check jumps into cleanup routine
    UA_CHECK(rv, goto error);

    return UA_STATUSCODE_GOOD;
error:
    if (data) {
        UA_free(data);
    }
    return rv;
```

#### Check with logging

A common case is to log a warning, info or error when the error-code implies a faulty state.
There are three macros for this:
- `UA_CHECK_INFO`
- `UA_CHECK_WARN`
- `UA_CHECK_ERROR`
Those macros can be called with a logger, a logging category, a message and variadic arguments.

In this example the return code is checked, while generating a log message with the `WARNING` logger level
and variadic arguments attached.

```c
static
UA_StatusCode
someFunc(UA_Server *server) {

    UA_StatusCode rv = some_func();
    UA_CHECK_WARN(rv, goto error, &server->config.logger, UA_LOGCATEGORY_SERVER, "failed: %s!", "my arg");
    return UA_STATUSCODE_GOOD;
error:
    // do cleanup
    return rv;
}
```

### Output of logging messages

The message of a check-call is generated with additional information and can look like this:

`[1601-01-01 00:02:35.290 (UTC+0000)] warn/server	My log message (/path/to/open62541/src/pubsub/ua_pubsub_reader.c:1207: status-code: GoodNonCriticalTimeout)`

In this case a macro-call such as `UA_CHECK_WARN(..., "My log message")` would suffice. The information of filename, linenumber and status-code string is added by the macro.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -15,6 +15,7 @@
  *    Copyright 2017 (c) Mark Giraud, Fraunhofer IOSB
  *    Copyright 2018 (c) Hilscher Gesellschaft für Systemautomation mbH (Author: Martin Lang)
  *    Copyright 2019 (c) Kalycito Infotech Private Limited
+ *    Copyright 2021 (c) Fraunhofer IOSB (Author: Jan Hermes)
  */
 
 #include "ua_server_internal.h"
@@ -71,14 +72,13 @@ UA_UInt16 addNamespace(UA_Server *server, const UA_String name) {
     /* Make the array bigger */
     UA_String *newNS = (UA_String*)UA_realloc(server->namespaces,
                                               sizeof(UA_String) * (server->namespacesSize + 1));
-    if(!newNS)
-        return 0;
+    UA_CHECK_MEM(newNS, return 0);
+
     server->namespaces = newNS;
 
     /* Copy the namespace string */
     UA_StatusCode retval = UA_String_copy(&name, &server->namespaces[server->namespacesSize]);
-    if(retval != UA_STATUSCODE_GOOD)
-        return 0;
+    UA_CHECK_STATUS(retval, return 0);
 
     /* Announce the change (otherwise, the array appears unchanged) */
     ++server->namespacesSize;
@@ -98,9 +98,8 @@ UA_UInt16 UA_Server_addNamespace(UA_Server *server, const char* name) {
 
 UA_ServerConfig*
 UA_Server_getConfig(UA_Server *server) {
-  if(!server)
-      return NULL;
-  return &server->config;
+    UA_CHECK_MEM(server, return NULL);
+    return &server->config;
 }
 
 UA_StatusCode
@@ -139,20 +138,16 @@ UA_Server_forEachChildNodeCall(UA_Server *server, UA_NodeId parentNodeId,
 
     UA_BrowseResult br = UA_Server_browse(server, 0, &bd);
     UA_StatusCode res = br.statusCode;
-    if(res != UA_STATUSCODE_GOOD) {
-        UA_BrowseResult_clear(&br);
-        return res;
-    }
+    UA_CHECK_STATUS(res, goto cleanup);
 
     for(size_t i = 0; i < br.referencesSize; i++) {
         if(!UA_ExpandedNodeId_isLocal(&br.references[i].nodeId))
             continue;
         res = callback(br.references[i].nodeId.nodeId, !br.references[i].isForward,
                        br.references[i].referenceTypeId, handle);
-        if(res != UA_STATUSCODE_GOOD)
-            break;
+        UA_CHECK_STATUS(res, goto cleanup);
     }
-
+cleanup:
     UA_BrowseResult_clear(&br);
     return res;
 }
@@ -248,15 +243,20 @@ UA_Server_cleanup(UA_Server *server, void *_) {
 /* Server Lifecycle */
 /********************/
 
+static
+UA_INLINE
+UA_Boolean UA_Server_NodestoreIsConfigured(UA_Server *server) {
+    return server->config.nodestore.getNode != NULL;
+}
+
 static UA_Server *
 UA_Server_init(UA_Server *server) {
+
     UA_StatusCode res = UA_STATUSCODE_GOOD;
-    
-    if(!server->config.nodestore.getNode) {
-        UA_LOG_FATAL(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "No Nodestore configured in the server");
-        goto cleanup;
-    }
+    UA_CHECK_FATAL(UA_Server_NodestoreIsConfigured(server), goto cleanup,
+                    &server->config.logger, UA_LOGCATEGORY_SERVER,
+                    "No Nodestore configured in the server"
+                   );
 
     /* Init start time to zero, the actual start time will be sampled in
      * UA_Server_run_startup() */
@@ -284,10 +284,8 @@ UA_Server_init(UA_Server *server) {
     /* Create Namespaces 0 and 1
      * Ns1 will be filled later with the uri from the app description */
     server->namespaces = (UA_String *)UA_Array_new(2, &UA_TYPES[UA_TYPES_STRING]);
-    if(!server->namespaces) {
-        UA_Server_delete(server);
-        return NULL;
-    }
+    UA_CHECK_MEM(server->namespaces, goto cleanup);
+
     server->namespaces[0] = UA_STRING_ALLOC("http://opcfoundation.org/UA/");
     server->namespaces[1] = UA_STRING_NULL;
     server->namespacesSize = 2;
@@ -312,8 +310,7 @@ UA_Server_init(UA_Server *server) {
 
     /* Initialize namespace 0*/
     res = UA_Server_initNS0(server);
-    if(res != UA_STATUSCODE_GOOD)
-        goto cleanup;
+    UA_CHECK_STATUS(res, goto cleanup);
 
 #ifdef UA_ENABLE_PUBSUB
     /* Build PubSub information model */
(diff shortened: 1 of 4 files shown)


jpfr:
Looks good!
So this PR will definitely pass, as the new macros are never used.

Please add a commit that uses the new check macros in a few places.
There are many instances of the check+log combo in the server code.

That way we can suss out potential problems with the different compilers.

src/server/ua_server_async.c
```diff
@@ -204,14 +205,10 @@ UA_AsyncManager_createAsyncResponse(UA_AsyncManager *am, UA_Server *server,
...
-    if(!newentry)
-        return UA_STATUSCODE_BADOUTOFMEMORY;
+    UA_CHECK_MEM(newentry, return UA_STATUSCODE_BADOUTOFMEMORY);
 
     UA_StatusCode res = UA_NodeId_copy(sessionId, &newentry->sessionId);
-    if(res != UA_STATUSCODE_GOOD) {
-        UA_free(newentry);
-        return res;
-    }
+    UA_CHECK(res, goto error);
```
jpfr, line 211: goto is not evil.
But it is still harder to read.
Please put the free in the UA_Check macro as well.

src/server/ua_server_async.c
```diff
@@ -250,19 +251,11 @@ UA_AsyncManager_createAsyncOp(UA_AsyncManager *am, UA_Server *server,
...
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "UA_Server_SetNextAsyncMethod: Mem alloc failed.");
-        return UA_STATUSCODE_BADOUTOFMEMORY;
-    }
+    UA_CHECK_MEM(ao, goto mem_error);
 
     UA_StatusCode result = UA_CallMethodRequest_copy(opRequest, &ao->request);
-    if(result != UA_STATUSCODE_GOOD) {
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+    UA_CHECK_ERROR(result, goto error, &server->config.logger, UA_LOGCATEGORY_SERVER,
```
jpfr, line 257: dito, use free+return in the check macro instead of an additional goto.

src/server/ua_server_async.c
```diff
@@ -250,19 +251,11 @@ UA_AsyncManager_createAsyncOp(UA_AsyncManager *am, UA_Server *server,
     }
 
     UA_AsyncOperation *ao = (UA_AsyncOperation*)UA_calloc(1, sizeof(UA_AsyncOperation));
-    if(!ao) {
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "UA_Server_SetNextAsyncMethod: Mem alloc failed.");
-        return UA_STATUSCODE_BADOUTOFMEMORY;
-    }
+    UA_CHECK_MEM(ao, goto mem_error);
```
jpfr, line 254: Now it gets interesting.
Here we could have to specialize UA_CHECK_MEM_ERROR to remove the goto.
Or we de-specialize UA_CHECK_ERROR to include a logical statement instead of just the variable name.

jpfr (commented):
A few comments.
The file ua_server_async is only compiled when multithreading is enabled.
So proably a few compilers still haven't used the macros.

src/server/ua_server_async.c
JanSurft: You can also review the file `ua_server.c`, it is also refactored with the use of check macros.

Okay it will be changed. Although there are some places in code where cleanup is done in a cleanup routine at the bottom of a function. Because I prever it that way I also modified the instance in this case. Maybe we find a rule of thumb when its is considered good to jump into a cleanup routine and when its not. (e.g. when there are at least two places in the same function that use the same cleanup routine).

You could review such an example in `ua_server.c` on lines 132 - 152.

src/server/ua_server_async.c
JanSurft: Yes, I also thought about that, there are some ideas:

1. Have a specific base macro that just asserts that the given boolean expression is not true.
```c
#define UA_CHECK(A, EVAL) if(!(A)) { EVAL; }

// usage:
void *d = malloc(...);
UA_CHECK(d, return error);

UA_StatusCode rv = do_some();
UA_CHECK(rv == UA_STATUSCODE_GOOD, return error);
```
This would be good for general use of the check macro in places where the status code checking is not that extensively used.

2. Have a specific macro to check the status (like you suggested earlier), referencing the base asserter.

```c
#define UA_CHECK_STATUS(STATUS, EVAL) UA_CHECK((STATUS) == UA_STATUSCODE_GOOD, EVAL)
```

3. Have specific macros for exactly those two cases that include logging for ERROR, INFO, WARNING. this would mean 6 macros for the bare assertion and the one including the error code check. Downside: Macronames kind of explode.

```c
#define UA_CHECK_STATUS_ERROR(...)
#define UA_CHECK_ERROR(...)

#define UA_CHECK_STATUS_INFO(...)
#define UA_CHECK_INFO(...)

#define UA_CHECK_STATUS_WARN(...)
#define UA_CHECK_WARN(...)
```

Or we leave out all the specifications for statuscode and just use the explicit versions. The downside is, that there are about a 1000 uses of `if (... != UA_STATUSCODE_GOOD)` in the code until now so it could reduce at least some of that verbosity.

JanSurft: Although it might be no such big change in verbosity, but it would at least encapslate the specific error checking. E.g. if indeed one might want to check the statuscode in an augmented way than "== UA_STATUSCODE_GOOD".

```c
UA_CHECK_ERROR(rv == UA_STATUSCODE_GOOD, return rv, logger, CATEGORY, "message");

// versus

UA_CHECK_STATUS_ERROR(rv, return rv, logger, CATEGORY, "message");
```

src/server/ua_server.c
```diff
@@ -312,8 +308,7 @@ UA_Server_init(UA_Server *server) {
 
     /* Initialize namespace 0*/
     res = UA_Server_initNS0(server);
-    if(res != UA_STATUSCODE_GOOD)
-        goto cleanup;
```
JanSurft, line 316: Should we also get rid of the jump to the cleanup routine here?

src/server/ua_server_async.c
JanSurft: A huge pro for specific STATUS check macros:
  `UA_StatusCode_name(...)` can be added to the log information without any further effort.

src/ua_util_internal.h
```diff
@@ -86,6 +87,48 @@ typedef UA_UInt64 u64;
...
+
+#define UA_CHECK(STATUSCODE, EVAL_ON_ERROR)                                              \
+    if((STATUSCODE) != UA_STATUSCODE_GOOD) {                                             \
+        EVAL_ON_ERROR;                                                                   \
+    }
+
+#define UA_CHECK_LOG_INTERNAL(STATUSCODE, EVAL, LOG, LOGGER, CAT, MSG, ...)              \
+    UA_MACRO_EXPAND(UA_CHECK(                                                            \
+        STATUSCODE, LOG(LOGGER, CAT, "" MSG "%s (%s:%d: statuscode: %s)", __VA_ARGS__,   \
+                        __FILE__, __LINE__, UA_StatusCode_name(STATUSCODE));             \
```
JanSurft, line 102: adding `UA_StatusCode_name(STATUSCODE)` is a huge benefit for error logging. If there is no specific macro for statuscode checking then this would not be possible.

src/server/ua_server_async.c
jpfr: > Maybe we find a rule of thumb when its is considered good to jump into a cleanup routine and when its not. (e.g. when there are at least two places in the same function that use the same cleanup routine).

I like this rule!
Lets collect best practices in the CONTRIBUTING.md file.

src/server/ua_server_async.c
jpfr: > UA_CHECK_STATUS_ERROR(rv, return rv, logger, CATEGORY, "message");

Right, UA_CHECK_STATUS_ERROR is a good macro to have for the StatusCode name.

It seems that each macro only takes line or so to define.
With the macro nesting you already have in place.

So I would say let's add **all of the above**. ;-)

src/ua_util_internal.h
jpfr: 👍

jpfr:
Hey Jan,

just two minor comments;

1. Type-Checking
The function
```c
static UA_INLINE UA_Boolean isGood(UA_StatusCode code);
```
is exactly right. But it doesn't solve the `UA_CHECK(retval, goto error)` case.
Please add the same static checking for the other input types as well.
```c
static UA_INLINE UA_Boolean isTrue(UA_Boolean expression) { return expression; } // for UA_CHECK
static UA_INLINE UA_Boolean isNonNull(const void *ptr) { return (ptr != NULL); } // for UA_CHECK_MEM
```

2. Clean up history and use the git commit message format from CONTRIBUTING.md
The .gitignore change should stay separate. And maybe some other stuff as well.

Looking forward to bringing this in!

PR #4408, opened 2021-05-13
Ability for types from custom namespace to contain members from any namespace

Johnsik:
Added ability for types from custom namespace to contain members from any namespace (instead of choice the same namespace or namespace zero).


jpfr:
That is a good simplification.

Four wishes:
1. Make it pass the CI on all architectures.
2.  Don't remove the flag `check_add_cc_flag("-Wc++-compat")`.
2. Use a tool like `pahole` to check the new size of `UA_DataType` and paste it here. We should not have it leak into the next cacheline if possible.
3. Run the unit tests before/after the change and paste the timing differences

Johnsik:
> That is a good simplification.
But I'm afraid it doesn't work :)

> 1. Make it pass the CI on all architectures.
It work for me with VS2019 (Windows) and GCC 8.3.0 (Debian) but the same code don't compile in CI.
According ISO/IEC 9899 6.6:
```
9 An address constant is a null pointer, a pointer to an lvalue designating an object of static
storage duration, or a pointer to a function designator; it shall be created explicitly using
the unary & operator or an integer constant cast to pointer type, or implicitly by the use of
an expression of array or function type. The array-subscript [] and member-access .
and -> operators, the address & and indirection * unary operators, and pointer casts may
be used in the creation of an address constant, but the value of an object shall not be
accessed by use of these operators.
```
But UA_TYPES_XX isn't static. So what it work for me but it UB...
May be you have idea what can I do?

> 2. Don't remove the flag `check_add_cc_flag("-Wc++-compat")`.
With this flag don't work tentative definition of type `selfContainingUnionType` in `check_types_custom.c` because it unlegal in C++.

Johnsik:
> 1. Make it pass the CI on all architectures.

It don't work for BUILD_SHARED_LIBS=ON on Windows :(
According ISO/IEC 9899 6.6:
```
9 An address constant is a null pointer, a pointer to an lvalue designating an object of static
storage duration, or a pointer to a function designator; it shall be created explicitly using
the unary & operator or an integer constant cast to pointer type, or implicitly by the use of
an expression of array or function type. The array-subscript [] and member-access .
and -> operators, the address & and indirection * unary operators, and pointer casts may
be used in the creation of an address constant, but the value of an object shall not be
accessed by use of these operators.
```
But UA_TYPES_XX in case of import DLL has attribute __declspec(dllimport) so it isn't constant. 
May be you have idea what can I do?

> 2. Don't remove the flag `check_add_cc_flag("-Wc++-compat")`.

With this flag don't work tentative definition of type `selfContainingUnionType` in `check_types_custom.c` because it illegal in C++. But we don't going to compile it with C++ right?

Johnsik:
> 1. Make it pass the CI on all architectures.

It don't work for BUILD_SHARED_LIBS=ON on Windows :(
According ISO/IEC 9899 6.6:
```
9 An address constant is a null pointer, a pointer to an lvalue designating an object of static
storage duration, or a pointer to a function designator; it shall be created explicitly using
the unary & operator or an integer constant cast to pointer type, or implicitly by the use of
an expression of array or function type. The array-subscript [] and member-access .
and -> operators, the address & and indirection * unary operators, and pointer casts may
be used in the creation of an address constant, but the value of an object shall not be
accessed by use of these operators.
```
But UA_TYPES_XX in case of import DLL has attribute __declspec(dllimport) so it isn't constant. 
May be you have idea what can I do?

> 2. Don't remove the flag `check_add_cc_flag("-Wc++-compat")`.

With this flag don't work tentative definition of type `selfContainingUnionType` in `check_types_custom.c` because it illegal in C++. But we don't going to compile it with C++ right?

Johnsik:
> 1. Make it pass the CI on all architectures.
Done

> 2. Don't remove the flag `check_add_cc_flag("-Wc++-compat")`.
Done

> 4. Run the unit tests before/after the change and paste the timing differences
How I can do it?

Johnsik:
> 3\. Use a tool like `pahole` to check the new size of `UA_DataType` and paste it here. We should not have it leak into the next cacheline if possible.

jpfr:
Its good that you are so persistent on the subject.

I just had a look at the PR.
You added a customer method callback to every datatype member so it can be retrieved from the correct "type array".
That works. But the indirection is non-trival and also quite a bit slower. Especially, we might want to generate datatypes at runtime. But we would need a JIT-Compiler to generate the matching callback method at runtime.

If you definitely need to reference datatypes from another "type array", I think it makes more sense to just copy these datatypes over into the current "type array". The copy needs a bit more memory. But overall not as much as adding a method pointer to every member.

I hope you understand that we need to be very careful about the complexity that we add to the "innermost layer" of the entire SDK.

Johnsik:
Yes I completely agree with you, and how can you see my first version (before last commits) use pointer to datatype but not callback+index.
It was work excellent but not in case of Windows DLL :(
In result compilation of custom_type_example are fails with BUILD_SHARED_LIBS=ON on Windows.

Pointer into DLL data is not constant so it couldn't be used to in struct initializers outside main(). But callback is constant!
In case of static linking libs or in case of Linux shared libs data pointer into lib data also constant but not in case of Windows shared libs.

jpfr:
> Pointer into DLL data is not constant so it couldn't be used to in struct initializers outside main(). But callback is constant!

Yes. That was a nice hack! 👍 
We still can't do it for the above reasons.

Johnsik:
I see the following paths:

1. don't compile DLL and don't use DLL :)
2. initialize datatypes not at compile time, but from main ()
3. initialize datatypes during compilation, but make a reference to another datatype NONE and then fill it in main ()
4. do the same, but with conditional compilation only in case of DLL
5. check when using a reference to datatype == NONE and in case of NONE do datatype = (*nodesetArray) [index]

Obviously, in cases 3-5 instead of datatype* it is necessary to store struct {datatype*, nodesetArray*, index} for all cases, or only for DLL in case of conditional compilation.
nodesetArray* can be formed on the fly as I am doing now using callback() or filled when loading nodeset from main()

Johnsik:
> 1. Make it pass the CI on all architectures.

Done

> 2. Don't remove the flag `check_add_cc_flag("-Wc++-compat")`.

This flag breaks C tentative definition of type selfContainingUnionType in test check_types_custom.c because tentative definition legal in C but illegal in C++. We don't going to compile it with C++ right?

> 3. Use a tool like `pahole` to check the new size of `UA_DataType` and paste it here. We should not have it leak into the next cacheline if possible.

Do you have an example how use it?

> 4. Run the unit tests before/after the change and paste the timing differences

How I can do it?

PR #4411, opened 2021-05-14
[REVIEW] refactor(pubsub): separate decoding and encoding

JanSurft:
Decoding and encoding is separated into different functions
concerened about header, payload, footer decoding/encoding


jpfr:
👍

PR #4415, opened 2021-05-18
refactor(pubsub): encryption decoupled from encode

JanSurft:
The position and portion to be encrypted is now handled
by the caller. Encoding methods decoupled from encryption info.


jpfr:
Github action builds are failing atm with a network issue.
Will restart at a later time.

PR #4418, opened 2021-05-18
[REVIEW] feat(pubsub): decryption

JanSurft:
Depends on #4428 

Adding the possibility to verify and decrypt pubsub networkmessages.


src/pubsub/ua_pubsub_security.h
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Copyright (c) 2017-2018 Fraunhofer IOSB (Author: Jan Hermes)
```
jpfr, line 5: 2017-2018?

src/pubsub/ua_pubsub_security.h
```diff
@@ -0,0 +1,28 @@
...
+#define UA_SERVER_PUBSUB_SECURITY_H
+
+#include <open62541/plugin/log.h>
+#include <open62541/types.h>
+#include <open62541/types_generated.h>
+
+#include "ua_pubsub_networkmessage.h"
+
+UA_StatusCode
+verifyAndDecrypt(const UA_Logger *logger,
```
jpfr, line 18: A header file for a single function, that's a bit too fine-grained.
Merge this into the general pubsub.h header.

src/pubsub/ua_pubsub_security.c
```diff
@@ -0,0 +1,167 @@
...
+    return retval;
+}
+
+UA_StatusCode
+verifyAndDecrypt(const UA_Logger *logger,
+                     UA_MessageSecurityMode securityMode,
+                 UA_ByteString *buffer, const size_t *currentPosition,
+                 const UA_NetworkMessage *currentNetworkMessage,
+                     void *channelContext,
+                 UA_StatusCode (*setNonce)(void *,const UA_ByteString *),
```
jpfr, line 78: Its easier to forward the securitypolicy than 5 separate function pointers that all come from the same securitypolicy.

include/open62541/plugin/log.h
```diff
@@ -44,7 +45,9 @@ typedef enum {
     UA_LOGCATEGORY_SERVER,
     UA_LOGCATEGORY_CLIENT,
     UA_LOGCATEGORY_USERLAND,
-    UA_LOGCATEGORY_SECURITYPOLICY
+    UA_LOGCATEGORY_SECURITYPOLICY,
```
jpfr, line 48: If you adjust these, you also have to adjust the two log-plugins in /plugins.
They probably would get an out-of-array access for some cases right now.
Because the enum is resolved to a string from an array.

To overcome this, we could have a #define with the number of categories.
If that used for `char * arrayNames[UA_LOGCATEGORIES_COUNT] = {...};` or similar, it breaks non-adjusted code in the plugins if we add new categories.

Such things are helpful for groups that maintain their own plugins.

include/open62541/server_pubsub.h
```diff
@@ -4,6 +4,7 @@
  *
  * Copyright (c) 2017-2018 Fraunhofer IOSB (Author: Andreas Ebner)
  * Copyright (c) 2019 Kalycito Infotech Private Limited
+ * Copyright (c) 2017-2018 Fraunhofer IOSB (Author: Jan Hermes)
```
jpfr, line 7: 2021

src/pubsub/ua_pubsub.h
```diff
@@ -293,6 +304,28 @@ UA_ReaderGroup_addSubscribeCallback(UA_Server *server, UA_ReaderGroup *readerGro
...
+                               UA_ReaderGroup *readerGroup);
+#endif
+
+UA_StatusCode
+readNetworkMessage(const UA_Logger *logger,
+                   UA_ByteString *buffer, size_t *currentPosition,
+                   UA_NetworkMessage *currentNetworkMessage,
+                   UA_ReaderGroup *readerGroup);
+
+UA_StatusCode
```
jpfr, line 325: Does this only receive the message or also process it?
Add a short comment to help getting a quick overview on the code.

src/pubsub/ua_pubsub_receiver.c
```diff
@@ -0,0 +1,207 @@
...
+receiveNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
+                      UA_PubSubConnection *connection, size_t previousPosition,
+                      UA_ByteString *buffer, size_t *currentPosition) {
+    size_t paddingBytes = 0;
+    UA_NetworkMessage currentNetworkMessage;
+    memset(&currentNetworkMessage, 0, sizeof(UA_NetworkMessage));
+
+    UA_StatusCode rv = UA_STATUSCODE_GOOD;
+    rv = readNetworkMessage(&server->config.logger,
+                                        buffer,
```
jpfr, line 70: You can save some vertical space here.
Vim's `J` key is you friend. :-)

src/pubsub/ua_pubsub_receiver.c
```diff
@@ -0,0 +1,207 @@
...
+    UA_CHECK_STATUS_WARN(rv, goto error, &server->config.logger, UA_LOGCATEGORY_SERVER,
+                         "Subscribe failed. verify and decrypt network message failed.");
+
+    /* TODO: We already know the ReaderGroup at this point. Now we loose that
+    * information. There is only one place where
+    * UA_PubSubConnection_processNetworkMessage is used. */
+    rv = UA_PubSubConnection_processNetworkMessage(server, connection, &currentNetworkMessage);
+    UA_CHECK_STATUS_WARN(rv, goto error, &server->config.logger, UA_LOGCATEGORY_SERVER,
+        "Subscribe failed. process network message failed.");
+    UA_NetworkMessage_clear(&currentNetworkMessage);
```
jpfr, line 83: Just "fall through" to the cleanup case below (instead of return good).
And the _clear happens automatically.

src/pubsub/ua_pubsub_receiver.c
```diff
@@ -0,0 +1,207 @@
...
+
+    /* Minimum ethernet packet size is 64 bytes where the header size is 14
+     * bytes and FCS size is 4 bytes so remaining minimum payload size of
+     * ethernet packet is 46 bytes */
+    /* TODO: Need to handle padding bytes for UDP */
+    if ((((*currentPosition) - previousPosition) < MIN_PAYLOAD_SIZE_ETHERNET) &&
+        (strncmp((const char *)connection->config->transportProfileUri.data, "http://opcfoundation.org/UA-Profile/Transport/pubsub-eth-uadp", (size_t)(connection->config->transportProfileUri.length)) == 0)) {
```
jpfr, line 90: line length

src/pubsub/ua_pubsub_receiver.c
```diff
@@ -0,0 +1,207 @@
...
+    rv = UA_NetworkMessage_decodeBinary(buffer, currentPosition, currentNetworkMessage);
+    UA_CHECK_STATUS_WARN(rv, return rv, logger, UA_LOGCATEGORY_SERVER,
+                         "Subscribe failed. receive network message failed.");
+
+    return UA_STATUSCODE_GOOD;
+}
+
+static
+UA_StatusCode
+receiveNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
```
jpfr, line 61: This function does not receive anything.
It processes an already received buffer.

Should be called processSomethingFooBar.

src/pubsub/ua_pubsub_receiver.c
```diff
@@ -0,0 +1,207 @@
...
+            UA_DataValue_init(&dsm->data.keyFrameData.dataSetFields[i]);
+#else
+            UA_DataValue_clear(&dsm->data.keyFrameData.dataSetFields[i]);
+#endif
+        }
+    }
+}
+
+UA_StatusCode
+readNetworkMessage(const UA_Logger *logger,
```
jpfr, line 37: read is misleading.
Why not decryptDecodeNetworkMessage?

CMakeLists.txt
```diff
@@ -810,6 +810,7 @@ set(lib_sources ${PROJECT_SOURCE_DIR}/src/ua_types.c
                 ${PROJECT_SOURCE_DIR}/src/pubsub/ua_pubsub_networkmessage.c
                 ${PROJECT_SOURCE_DIR}/src/pubsub/ua_pubsub_writer.c
                 ${PROJECT_SOURCE_DIR}/src/pubsub/ua_pubsub_reader.c
+                ${PROJECT_SOURCE_DIR}/src/pubsub/ua_pubsub_receiver.c
```
jpfr, line 813: Writer and Reader are standard-defined concepts.
The code in _receiver clearly belongs to the reader. I would keep it in _reader.c to avoid confusion.
If there is indeed shared functionality with the writer we can have a file for common code and call it such as well.

jpfr (commented):
Looks good!
I know one should not comment WIP code.
But I couldn't resist and left a few pointers.

👍

CMakeLists.txt
JanSurft: It was my attempt to start a separation of concerns. In the future it might possibly be good to have publisher/subscribe functionality outside of server-coupled code. As the standard describes that server/client and publisher/subscriber are decoupled concepts.

src/pubsub/ua_pubsub_security.c
JanSurft: I attempted to create an easily to unit-test function that knows as little as possible from the higher level concepts (especially the configuration structures).

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1839,4 +1722,270 @@ UA_PubSubConnection_processNetworkMessage(UA_Server *server, UA_PubSubConnection
...
+    buffer.data = ReceiveMsgBuffer;
+
+    UA_StatusCode rv = connection->channel->receive(
+        connection->channel, &buffer, NULL, readerGroup->config.timeout);
+
+    // TODO attention: here rv is ok if UA_STATUSCODE_GOOD != rv
+    UA_CHECK_WARN(!UA_StatusCode_isBad(rv), return rv,
+                  &server->config.logger, UA_LOGCATEGORY_SERVER,
+                  "SubscribeCallback(): Connection receive failed!");
+    if(buffer.length > 0) {
```
jpfr, line 1955: You can move the if(rt) inside the while-loop and reuse the same while-loop for both cases.
That also lets you remove the outer if. Just do a while() instead of do {} while.

src/pubsub/ua_pubsub_security.c
```diff
@@ -0,0 +1,168 @@
...
+                         "PubSub receive. "
+                         "Message is not signed but ReaderGroup requires signature");
+            retval = UA_STATUSCODE_BADSECURITYMODEREJECTED;
+        }
+    }
+    return retval;
+}
+
+UA_StatusCode
+verifyAndDecrypt(const UA_Logger *logger,
```
jpfr, line 78: Vertical space can be reduced.

src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -851,7 +851,23 @@ static UA_StatusCode
 UA_NetworkMessage_decodeBinaryInternal(const UA_ByteString *src, size_t *offset,
                                            UA_NetworkMessage* dst) {
 
-    UA_StatusCode rv = UA_NetworkMessage_decodeHeaders(src, offset, dst);
+    UA_StatusCode rv = UA_STATUSCODE_GOOD;
+
```
jpfr, line 855: This is now used exclusively in UA_NetworkMessage_decodeBinary.
The functionality can be folded in there.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1839,4 +1722,270 @@ UA_PubSubConnection_processNetworkMessage(UA_Server *server, UA_PubSubConnection
...
+            } while((buffer.length) > currentPosition);
+        } else {
+            UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_USERLAND,
+                         "Message received:");
+            do {
+                rv = decodeAndProcessNetworkMessage(server, readerGroup, connection,
+                                                    previousPosition, &buffer,
+                                                    &currentPosition);
+
+                // TODO: check what actions are to be done on error.. nothing?
```
jpfr, line 1978: Yes. Break the loop on error and return rv.

jpfr (commented):
Really good!
It even simplifies the control flow of the reader.
Some last comments. Can be merged once these are addressed.

src/pubsub/ua_pubsub_reader.c
JanSurft: That also was my initial thought, but it broke some tests.

src/pubsub/ua_pubsub_reader.c
JanSurft: This would mean an additional "if" statement inside a while loop.
I wanted to avoid that, although surely it might not be of any relevance.

otherwise I had a solution once which included a function pointer, do that instead?:

```c
    UA_StatusCode (*decodeAndProcessNetworkMessageFun)(UA_Server *server, UA_ReaderGroup *readerGroup,
                                     UA_PubSubConnection *connection, size_t previousPosition,
                                     UA_ByteString *buffer, size_t *currentPosition) = NULL;

    if (readerGroup->config.rtLevel == UA_PUBSUB_RT_FIXED_SIZE) {
        decodeAndProcessNetworkMessageFun = decodeAndProcessNetworkMessageRT;
    } else {
        decodeAndProcessNetworkMessageFun = decodeAndProcessNetworkMessage;
    }

    if(buffer.length > 0) {
        size_t currentPosition = 0;
        size_t previousPosition = 0;
        do {
            rv = decodeAndProcessNetworkMessageFun(server, readerGroup, connection,
                                                      previousPosition, &buffer,
                                                      &currentPosition);
            UA_CHECK_STATUS_WARN(
                rv, (void) 0, &server->config.logger, UA_LOGCATEGORY_SERVER,
                "SubscribeCallback(): receive message failed");

                previousPosition = currentPosition;

        } while((buffer.length) > currentPosition);
    }
```

PR #4422, opened 2021-05-20
[WIP] test(pubsub): fix ignored pubsub subscribe test

JanSurft:
There is a test suite for pubsub subscribe in the source tree that never was
enabled.
Additionally, the tests it ran introduced a segfault which is fixed now.

TODO:
Check if the testing results and iterations of the server are deterministic and not depend on timing.

possible fix: pair call of **SetOperation for reader and writergroup. Because each will start exactly one publish/subscribe callback


andreasebner:
Tank you for mention this issue. It's addressed now in #5083 and the test is fixed.

andreasebner:
Tank you for mention this issue. It's addressed now in #5083 and the test is fixed.

PR #4423, opened 2021-05-20
feature(plugin): aes256 encryption

JanSurft:
Adds a plugin and test for aes256 pubsub encryption.


jpfr:
The commit is called initial support.
Is there a part still missing?

jpfr:
Ready to merge otherwise.

JanSurft:
> The commit is called initial support.
> Is there a part still missing?

Nothing missing, I just missed renaming it after squashing.

PR #4424, opened 2021-05-20
ci: dependant issues (allows dependent PRs)

JanSurft:
Workflow that allows for adding dependencies into a pull request.

An example pullrequest message might look like this:

```
Depends on #4420 

This pull request depends on integrating the pullrequest #4420.
```

The dependent-issues-bot will add a "dependent" label to this pullrequest an block until the dependency is integrated.
This way developers can create multiple separate pullrequests that depend on one another.

An example image of such a pullrequest with the according bot-action:

image

### DIFF
--- a/.github/workflows/dependent-issues.yml
+++ b/.github/workflows/dependent-issues.yml
@@ -1,0 +1,42 @@
+name: Dependent Issues
+
+on:
+  issues:
+    types:
+      - opened
+      - edited
+      - reopened
+  pull_request_target:
+    types:
+      - opened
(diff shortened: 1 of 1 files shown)


jpfr:
Looks good.
But the PR should only contain the dependant issues feature.
Right now it is 500 lines big.

JanSurft:
> Looks good.
> But the PR should only contain the dependant issues feature.
> Right now it is 500 lines big.

Apparently it was based on the aes256 encryption branch, that's fixed.

PR #4433, opened 2021-05-26
fix(ex): fix the usage info for example: tutorial_pubsub_mqtt_publish

winndows:
Fix the usage info for example: tutorial_pubsub_mqtt_publish

Signed-off-by: Pingfang Liao <winndows@163.com>


winndows:
$ git describe --tags --dirty --match v*
v.1.2.2-212-g08b70da4

The above tag is not as expected as format "# v1.2.3-5-g4538abcd" in "tools/prepare_packaging.py", an extra dot is there after "v". 

Does it mean that we need change code for "tools/prepare_packaging.py"?

Thanks!

PR #4435, opened 2021-05-26
[REVIEW] PubSub hardware encryption and decryption using pkcs11

opcua-tsn-team-kalycito:


src/pubsub/ua_pubsub_writer.c
```diff
@@ -4,7 +4,7 @@
  *
  * Copyright (c) 2017-2019 Fraunhofer IOSB (Author: Andreas Ebner)
  * Copyright (c) 2019 Fraunhofer IOSB (Author: Julius Pfrommer)
- * Copyright (c) 2019 Kalycito Infotech Private Limited
+ * Copyright (c) 2021 Kalycito Infotech Private Limited
```
jpfr, line 7: 2019-2021

CMakeLists.txt
```diff
@@ -274,6 +281,12 @@ mark_as_advanced(UA_ENABLE_PUBSUB_BUFMALLOC)
 #RT and Transport PubSub settings
 option(UA_ENABLE_PUBSUB_ETH_UADP "Enable publish/subscribe UADP over Ethernet" OFF)
 
+if(UA_ENABLE_PUBSUB_ENCRYPTION)
```
jpfr, line 284: The TPM implementation should be a SecurityPolicy plugin that can coexist with software encryption.
**Do not make them incompatible**.

include/open62541/plugin/securitypolicy.h
```diff
@@ -377,6 +377,27 @@ struct UA_PubSubSecurityPolicy {
 
 #endif
 
+#ifdef UA_ENABLE_PUBSUB_ENCRYPTION_TPM
```
jpfr, line 380: Use the existing `UA_PubSubSecurityPolicy`.
If required, make a proposal for forwarding a "context" pointer into a new writerGroupContext.

jpfr: Let me stress this, we will not duplicate interfaces for the TPM.
Reuse the `UA_PubSubSecurityPolicy`.

plugins/crypto/pkcs11/pkcs11.h
```diff
@@ -0,0 +1,1765 @@
+/* pkcs11.h
```
jpfr, line 1: I don't like the "code dump".
The file is from here: https://github.com/OpenSC/libp11/blob/master/src/pkcs11.h
It would be much better if that library was installed system-wide.
So we don't have to maintain this going forward.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -2149,6 +2174,12 @@ writeNetworkMessage(UA_WriterGroup *wg, size_t msgSize,
     rv = encryptAndSign(wg, nm, networkMessageStart, payloadStart, footerEnd);
     UA_CHECK_STATUS(rv, return rv);
 
+#endif
+#ifdef UA_ENABLE_PUBSUB_ENCRYPTION_TPM
+
+    rv = encryptTPM(wg, nm, payloadStart, footerEnd);
```
jpfr, line 2180: This is only encryption from what I can read.
What about signing?

jpfr (changes requested):
See my comments below.
In addition, add a test that tests hardware encryption with software decryption.

plugins/include/open62541/plugin/securitypolicy_default.h
```diff
@@ -57,6 +57,17 @@ UA_PubSubSecurityPolicy_Aes256Ctr(UA_PubSubSecurityPolicy *policy,
 
 #endif
 
+#ifdef UA_ENABLE_TPM2_PKCS11
+
+UA_EXPORT UA_StatusCode
+UA_PubSubSecurityPolicy_Aes128CtrTPM(UA_PubSubSecurityPolicy *policy, char *userpin, unsigned long slotId,
+                                     char *encryptionKeyLabel, char *signingKeyLabel, const UA_Logger *logger);
+                                     UA_EXPORT UA_StatusCode
```
jpfr, line 65: Indent

jpfr (commented):
Looks good.
Please also test hardware encryption against the software decryption and vice versa.
We can discuss this out of band.

PR #4436, opened 2021-05-26
[Draft] FileType implementation

opcua-tsn-team-kalycito:
This pull request implements file implementation.
- Methods implemented:
    - Open, Close, Read, Write
    - TODO: SetPosition, GetPosition
- Properties supported
    - Size, OpenCount


include/open62541/server.h
```diff
@@ -1553,6 +1553,71 @@ UA_Server_updateCertificate(UA_Server *server,
                             UA_Boolean closeSessions,
                             UA_Boolean closeSecureChannels);
 
+#ifdef UA_GENERATED_NAMESPACE_ZERO_FULL
```
jpfr, line 1556: Don't add this per default to namespace full.
It should be a separate configuration option.

include/open62541/server.h
```diff
@@ -1553,6 +1553,71 @@ UA_Server_updateCertificate(UA_Server *server,
...
+
+/* Set information on FileType Object instance
+ * 
+ * @param server The server object
+ * @param fileTypeObjectNodeId The NodeId of the fileType Object instance 
+ * @param fileName Name of the file created
+ * @param filePath Path to the file (TODO)
+ * @return UA_STATUSCODE_GOOD on success
+ */
+UA_StatusCode UA_EXPORT
```
jpfr, line 1566: The callbacks you expose in the public API are much too detailed.
For the user it should be enough to have
```c
UA_Server_addFile(UA_Server *server, const UA_NodeId requestedNewNodeId,
                        const UA_NodeId parentNodeId,
                        const UA_NodeId referenceTypeId,
                        const UA_QualifiedName browseName,
                        const UA_ObjectAttributes attr,
                        const UA_String filePath, /* Load the file from here */
                        void *nodeContext, UA_NodeId *outNewNodeId);
```
All the methods and callbacks can be created and set internally from there and don't have to be exposed.

jpfr (changes requested):
Good feature!
See the inline comments.

include/open62541/server.h
opcua-tsn-team-kalycito: Fixed it. Also added one more Public API `UA_Server_registerFileNode` to register the FileType object instance that added through information modeling.

include/open62541/server.h
opcua-tsn-team-kalycito: Fixed it.

include/open62541/server.h
```diff
@@ -1553,6 +1554,38 @@ UA_Server_updateCertificate(UA_Server *server,
...
+/* Add the FileType Object instance to the server information model
+ * and register method callbacks
+ *
+ * @param server The server object
+ * @param requestedNewNodeId The requested NodeId of the fileType Object instance
+ * @param filePath Path to the file
+ * @return UA_STATUSCODE_GOOD on success
+ */
+UA_StatusCode UA_EXPORT
+UA_Server_addFile(UA_Server *server, const UA_NodeId requestedNewNodeId,
```
jpfr, line 1568: For consistency with `registerFileNode`, rename this to `UA_Server_addFileNode`.

src/server/ua_file.c
```diff
@@ -0,0 +1,419 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jpfr, line 1: Rename to `ua_file_nodes.c`.

src/server/ua_file.c
```diff
@@ -0,0 +1,419 @@
...
+        return bpr.statusCode;
+    }
+    UA_StatusCode retval = UA_NodeId_copy(&bpr.targets[0].targetId.nodeId, outFieldNodeId);
+    UA_BrowsePathResult_clear(&bpr);
+    return retval;
+}
+
+static UA_StatusCode
+updateOpenCountProperty(UA_Server *server, UA_FileType* fileObject)
+{
```
jpfr, line 56: Use the usual conventions of the entire library with curly brackets on the same line.

src/server/ua_file.c
```diff
@@ -0,0 +1,419 @@
...
+                  const UA_ObjectAttributes attr,
+                  const UA_String filePath,
+                  void *nodeContext, UA_NodeId *outNewNodeId)
+{
+    UA_StatusCode result = UA_Server_addObjectNode(server, requestedNewNodeId,
+                            parentNodeId, referenceTypeId, browseName,
+                            UA_NODEID_NUMERIC(0, UA_NS0ID_FILETYPE),
+                            attr, nodeContext, outNewNodeId);
+    
+    UA_file *sampleFile = UA_file_open((char*)filePath.data, "w");
```
jpfr, line 384: Error-check everywhere.
This should fail when the file does not exist.
Use proper logging,

src/server/ua_file.c
```diff
@@ -0,0 +1,419 @@
...
+
+    //If NodeId is assigned by server
+    if(outNewNodeId)
+    {
+        result |= setFileTypeInfo(server, *outNewNodeId, filePath);
+        result |= setFileMethodCallbacks(server, *outNewNodeId);
+    }
+
+    //else use the  NodeId requested by user
+    else
```
jpfr, line 395: This will fail if the user does not provider outNewNodeId and if the NodeId has `identifier.numeric==0`.
That is the convention to ask the nodestore to generate a new numeric nodeid.
Use a temporary newNodeId variable.

src/server/ua_file.c
```diff
@@ -0,0 +1,419 @@
...
+UA_Server_readFileCallback(UA_Server *server,
+                            const UA_NodeId *sessionId, void *sessionHandle,
+                            const UA_NodeId *methodId, void *methodContext,
+                            const UA_NodeId *objectId, void *objectContext,
+                            size_t inputSize, const UA_Variant *input,
+                            size_t outputSize, UA_Variant *output)
+{
+    UA_UInt32 fileHandle = *(UA_UInt32 *)input[0].data;
+    UA_Int32 value = *(UA_Int32 *)input[1].data;
+    UA_UInt64 requestedLengthToRead = (UA_UInt64)value;
```
jpfr, line 231: Use size_t for lengthToRead etc.

src/server/ua_file.c
```diff
@@ -0,0 +1,419 @@
...
+
+static UA_StatusCode
+UA_Server_setPositionFileCallback(UA_Server *server,
+                                    const UA_NodeId *sessionId, void *sessionHandle,
+                                    const UA_NodeId *methodId, void *methodContext,
+                                    const UA_NodeId *objectId, void *objectContext,
+                                    size_t inputSize, const UA_Variant *input,
+                                    size_t outputSize, UA_Variant *output)
+{
+    return UA_STATUSCODE_BADNOTIMPLEMENTED;
```
jpfr, line 311: The idea of a file is to transport content that is larger than what fits inside a single message.
You can only do that if you preserve the position in the file across different calls to `read`.

So you need some persistency of that information stored in the user session.
There is a way to get there in a clean fashion. The new PR #4463 adds a key-value configuration system.
Add key-value parameters to the session struct.
And select a nice key-name under which the position for each particular file is stored.

src/server/ua_server.c
```diff
@@ -171,6 +171,24 @@ void UA_Server_delete(UA_Server *server) {
     }
     UA_Array_delete(server->namespaces, server->namespacesSize, &UA_TYPES[UA_TYPES_STRING]);
 
+#ifdef UA_ENABLE_FILETYPE_OBJECT_SUPPORT
+    UA_FileType *fileObject, *fileObject_tmp;
```
jpfr, line 175: Please keep this as much out of ua_server.c as possible.
Set up destructor set for the FileType object type.
So the cleanup happens automatically when the node is deleted.

src/server/ua_server_internal.h
```diff
@@ -65,6 +65,26 @@ typedef enum {
...
+typedef struct UA_FileInfo {
+    LIST_ENTRY(UA_FileInfo) listEntry;
+    UA_file *file;
+    UA_UInt32 filehandle;
+    UA_Byte openFileMode;
+}UA_FileInfo;
+
+typedef struct UA_FileType {
+    LIST_ENTRY(UA_FileType) listEntry;
+    UA_NodeId fileTypeObjectNodeId;
```
jpfr, line 78: fileTypeObjectNodeId or fileNodeId?

src/server/ua_file.c
jpfr: Also limit the number of open files per session.
Or this can be a way to consume arbitrary amounts of memory.

jpfr (changes requested):
This is on a good way.
See above for a new batch of comments.

jpfr:
You can now use this facility to attach data to session:
https://github.com/open62541/open62541/blob/master/include/open62541/server.h#L434

src/server/ua_server_internal.h
ZidCode: Why a list of filetyeobjects in the ua_server_internal.h is necessary?

src/server/ua_file.c
```diff
@@ -0,0 +1,419 @@
...
+    UA_NodeId getPositionNodeId;
+    UA_NodeId writeNodeId;
+    UA_NodeId closeNodeId;
+
+    result |= getFieldNodeId(server, &fileObjectNodeId, &fieldOpenQN, &openNodeId);
+    result |= getFieldNodeId(server, &fileObjectNodeId, &fieldReadQN, &readNodeId);
+    result |= getFieldNodeId(server, &fileObjectNodeId, &fieldSetPositionQN, &setPositionNodeId);
+    result |= getFieldNodeId(server, &fileObjectNodeId, &fieldGetPositionQN, &getPositionNodeId);
+    result |= getFieldNodeId(server, &fileObjectNodeId, &fieldWriteQN, &writeNodeId);
+    result |= getFieldNodeId(server, &fileObjectNodeId, &fieldCloseQN, &closeNodeId);
```
ZidCode, line 341: This always returns the NodeIds of the Type Object. Shouldn't it be the NodeID of the Object itself? Is it necessary to add explicitly the methods to the Object?

src/server/ua_file.c
daniel-petrovic: what should be the (good) default value of max open files per session ?

daniel-petrovic: > The idea of a file is to transport content that is larger than what fits inside a single message. You can only do that if you preserve the position in the file across different calls to `read`.
> 
> So you need some persistency of that information stored in the user session. There is a way to get there in a clean fashion. The new PR #4463 adds a key-value configuration system. Add key-value parameters to the session struct. And select a nice key-name under which the position for each particular file is stored.

Is the persistency of the current position of file descriptor in the session is really needed here ? Because this seems to be implied by the current implementation anyway. As long the Close is not called, each subsequent call to Read will read from the underlying FIL E* stream starting from it's current stream position.

daniel-petrovic:
> Hi,
> 
> just wanted to ask what is the current status of this ticket?
> 
> Cheers

Would also like to know that ... have addressed some of the open issues in [this PR](https://github.com/opcua-tsn-team-kalycito/open62541/pull/22) but I'm not sure if it's still maintained

src/server/ua_file.c
suy: This can only log an error and return the invalid `UA_file*` to the caller.

pberginkonsult:
> > Hi,
> > just wanted to ask what is the current status of this ticket?
> > Cheers
> 
> Would also like to know that ... have addressed some of the open issues in [this PR](https://github.com/opcua-tsn-team-kalycito/open62541/pull/22) but I'm not sure if it's still maintained

And another one that wants to know the status of this PR. Is it something that can be accepted or is it rejected for inclusion in the project?

src/server/ua_server.c
```diff
@@ -167,6 +167,14 @@ void UA_Server_delete(UA_Server *server) {
...
+    LIST_FOREACH_SAFE(fileObject, &server->fileObjects, listEntry, fileObject_tmp)
+    {
+        UA_Server_deleteNode(server, fileObject->fileNodeId, true);
```
daniel-petrovic, line 174: This will likely cause the   UA_assert(++(lock->mutexCounter) == 1); in UA_LOCK which is called in UA_Server_deleteNode to fail.

giokara-oqton:
One more interested party here. Kudos to @daniel-petrovic for making the PR. Would really appreciate seeing this merged.

Darcraytore1:
I am also interested and I'm willing to contribute but I would love to have a new review to know what is lacking to make that PR be merged ? @jpfr @daniel-petrovic

paperino1983:
> @Darcraytore1 @giokara-work I'm currently not working on this. a new review would be needed

Hi @daniel-petrovic, to better understand...
do you mean that you're waiting for a new review by someone or shall it be done by you ?

daniel-petrovic:
> > @Darcraytore1 @giokara-work I'm currently not working on this. a new review would be needed
> 
> Hi @daniel-petrovic, to better understand... do you mean that you're waiting for a new review by someone or shall it be done by you ?

@pberginkonsult shuld be re-reviewed by the project maintainer. I'm not the project maintainer

dlp:
Hello @jpfr, is there any update on whether this will be addressed?
Without this feature, the SDK is increasingly at a competitive disadvantage as time goes on.

xR3b0rn:
I'm currently playing around with file-read and file-write implementations in open62541. From what I've learned, I think this PR, in its current state, is not mergeable. Honestly, I believe a generic synchronous implementation like this shouldn't be part of the open62541 core at all.

My biggest argument against it is that file I/O must not be done synchronously in this context. It blocks the underlying event loop for the duration of the read/write operation. For users dealing with larger files or slower storage, this will inevitably lead to timeouts and dropped connections.

Every solution that comes to mind to solve this asynchronously within the library either contradicts open62541's fundamental architectural design decisions (e.g., forcing thread pools or OS-specific async APIs) or feels like a hack.

Besides that, setting up the few nodes and callbacks is not the hard part. The real challenge is getting the actual reading and writing done properly and asynchronously. That implementation is highly application- and platform-specific, and providing a synchronous "one-size-fits-all" solution here sets a dangerous trap for users.

xR3b0rn:
Just one more critical consideration regarding why this PR is unmergeable: It completely breaks access control.

Imagine a file being opened by User A (who has Admin privileges), and the server returns file handle 5. User B (with only Operator privileges) could simply guess or brute-force this integer and pass 5 to the Read or Write method. Since the handles are not strictly tied to the session that created them, User B can now read or manipulate files they do not have permission for. 

This is a classic IDOR vulnerability.

PR #4449, opened 2021-06-02
fix(pubsub): process all network messages

MartinaFelberFEM:
Hi all!

I wanted to fix the failing PubSub message receive timeout tests, which have been disabled by this commit.
b0fbe615433ce9aa9eb946284900c39fa4b75344

A message receive timeout happened, which was not expected and in this case an assert raised an abort.
A message receive timeout can only happen if something has been received at least once.
Therefore I thought that the PubSub configuration was correct.
In the failing test I setup 3 connections which are all part of the same multicast group.

```
    /*  Writers            : Interval  -> Readers            : Timeout
        ----------------------------------------------------------------------------------
        Conn 1 WG 1 - DSW1 : 30        -> Conn 2 RG 1 - DSR1 : 40
        Conn 1 WG 1 - DSW2 : 30        -> Conn 2 RG 2 - DSR2 : 45
        Conn 2 WG 1 - DSW1 : 20        -> Conn 1 RG 1 - DSR1 : 25
        Conn 2 WG 2 - DSW1 : 10        -> Conn 3 RG 1 - DSR1 : 25
    */
```

I have added some traces and saw that sometimes not all received network messages are processed and therefore the timeout occurred, although the message has been received in time.

I think the problem is the UA_CHECK_STATUS_WARN macro in the receiveBufferedNetworkMessage() function.

If we have multiple connections which are all part of the same multicast group then every connection receives all sent messages, even if they are not meant for them.
In the current implementation the processing of the received buffer is aborted if a message is encountered that's not meant for the currently processed connection.
If this message is always at the beginning of the received frame, then all the other messages which would be interesting for the connection are not processed.

I am not sure if my multicast example is a good real life example or would not be valid.
But if we would have 2 connections (device A and device B) and both are in the same multicast group and both have multiple DataSetReaders and the receiveBufferedNetworkMessage() always checks the wrong DataSetWriterId first, before the expected second one, then I think that the messages would not be processed ....

Not sure about that yet ...

### DIFF
--- a/src/pubsub/ua_pubsub_reader.c
+++ b/src/pubsub/ua_pubsub_reader.c
@@ -1526,7 +1526,9 @@ UA_DataSetReader_process(UA_Server *server,
                         }
                     }
                 }
-
+#ifdef UA_ENABLE_PUBSUB_MONITORING
+                UA_DataSetReader_checkMessageReceiveTimeout(server, dataSetReader);
+#endif /* UA_ENABLE_PUBSUB_MONITORING */
                 return;
             }
 
@@ -1559,6 +1561,17 @@ UA_DataSetReader_process(UA_Server *server,
     UA_LOG_DEBUG(&server->config.logger, UA_LOGCATEGORY_SERVER, "DataSetReader '%.*s': received a network message",
         (int) dataSetReader->config.name.length, dataSetReader->config.name.data);
 #ifdef UA_ENABLE_PUBSUB_MONITORING
+    UA_DataSetReader_checkMessageReceiveTimeout(server, dataSetReader);
+#endif /* UA_ENABLE_PUBSUB_MONITORING */
+}
+
+#ifdef UA_ENABLE_PUBSUB_MONITORING
+
+void
+UA_DataSetReader_checkMessageReceiveTimeout(UA_Server *server, UA_DataSetReader *dataSetReader) {
+    assert(server != 0);
+    assert(dataSetReader != 0);
+
     /* if previous reader state was error (because we haven't received messages and ran into timeout) we should set the state back to operational */
     if(dataSetReader->state == UA_PUBSUBSTATE_ERROR) {
         UA_DataSetReader_setPubSubState(server, UA_PUBSUBSTATE_OPERATIONAL, dataSetReader);
@@ -1590,10 +1603,8 @@ UA_DataSetReader_process(UA_Server *server,
                         "Starting Message Receive Timeout timer failed.");
         UA_DataSetReader_setPubSubState(server, UA_PUBSUBSTATE_ERROR, dataSetReader);
     }
-#endif /* UA_ENABLE_PUBSUB_MONITORING */
 }
 
-#ifdef UA_ENABLE_PUBSUB_MONITORING
 /* Timeout callback for DataSetReader MessageReceiveTimeout handling */
 void
 UA_DataSetReader_handleMessageReceiveTimeout(UA_Server *server, void *dataSetReader) {
--- a/src/pubsub/ua_pubsub_reader.c
+++ b/src/pubsub/ua_pubsub_reader.c
@@ -793,6 +793,9 @@ checkReaderIdentifier(UA_Server *server, UA_NetworkMessage *pMsg, UA_DataSetRead
             break;
         return UA_STATUSCODE_BADNOTFOUND;
     case UA_PUBLISHERDATATYPE_UINT16:
+                UA_LOG_INFO(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                    "FEM: PublisherId = %u vs %u", (UA_UInt32) pMsg->publisherId.publisherIdUInt16, (UA_UInt32) *(UA_UInt16*)reader->config.publisherId.data);
+
         if(reader->config.publisherId.type == &UA_TYPES[UA_TYPES_UINT16] &&
            pMsg->publisherIdType == UA_PUBLISHERDATATYPE_UINT16 &&
            pMsg->publisherId.publisherIdUInt16 == *(UA_UInt16*)reader->config.publisherId.data)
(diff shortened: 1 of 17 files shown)


JanSurft:
Hi Martina,
thanks for correcting the issues with the message recv timeout test! I will add some review to the code for suggested changes.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1975,8 +1975,11 @@ receiveBufferedNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
     while(buffer.length > currentPosition) {
         rv = decodeAndProcessNetworkMessageFun(
             server, readerGroup, connection, previousPosition, &buffer, &currentPosition);
-        UA_CHECK_STATUS_WARN(rv, return rv, &server->config.logger, UA_LOGCATEGORY_SERVER,
-                             "SubscribeCallback(): receive message failed");
+        if ((rv != UA_STATUSCODE_GOOD) && (rv != UA_STATUSCODE_BADNOTFOUND)) {
```
JanSurft, line 1978: 1. We need to intercept the case when `decodeAndProcessNetworkMessageFun` did not progress in the buffer and no error was thrown, otherwise we might end up in an infinite loop (just add a first interception of the buffer progress with according logging and breaking the loop)
2. Would be great to create a separate if-clause for handling the `UA_STATUSCODE_BADNOTFOUND` specifically (including a specified logging message)
3. You can use UA_CHECK_STATUS_WARN for any other bad value that was encountered after the `BAD_NOT_FOUND` case

JanSurft (commented):
The multicast is test is a great special case that we still should test and support, thanks for the fix. I added just some little things that might improve readability and error-proneness.

src/pubsub/ua_pubsub_reader.c
MartinaFelberFEM: Many thanks for your review!

I am not sure if I have understood all of your points:

1. I would have expected that the `decodeAndProcessNetworkMessageFun` would return an error if currentPosition is still 0. 
Shall I add an additional check, after checking the statusCode for BADNOTFOUND and other errors?
`if (previousPosition == currentPosition)`
But this hasn't been checked before I think -> but as you suggest!

2. I have added the if-clause but I print the log at `decodeAndProcessNetworkMessage` as `decodeAndProcessNetworkMessageRT` does. Is this ok?

Log looks like this:
```
[1601-01-01 00:02:48.545 (UTC+0000)] debug/server       PubSub subscribe callback
[1601-01-01 00:02:48.545 (UTC+0000)] info/server        Dataset reader not found. Check PublisherID, WriterGroupID and DatasetWriterID
[1601-01-01 00:02:48.545 (UTC+0000)] info/server        PubSub receive. Unknown message received. Will not be processed.

```

And I haven't used log category warning but info instead, because in the multicast scenario I think it's ok to receive these messages.

If you like and you think it's easier we could also talk about it

MartinaFelberFEM: I have rebased on current master and now it doesn't work anymore.

I think it has something to do with this commit: https://github.com/open62541/open62541/commit/b862f28e4bac4403c39c9cb5d9b436643cde3183

MartinaFelberFEM: because in the past the subscribeCallback() called UA_Server_processNetworkMessage -> which checked getReaderFromIdentifier -> which checked **all DataSetReaders of the ReaderGroup**

But now: decodeAndProcessNetworkMessage directly calls UA_ReaderGroup_processNetworkMessage, and this only checks the current readergroup
and then the full network message has been processed and the next readergroup of the same connection does not receive() anything anymore ...
maybe we just have to add 
    `LIST_FOREACH(readerGroup, &pConnection->readerGroups, listEntry) {`
    
    
    
In general I am not sure if it's a good idea to provide a subscribe callback for every ReaderGroup.
Because if there are multiple readergroups then (at least in the past) the first ReaderGroup received() everything and forwarded the network messages to the right readergroups and dataset readers within it's connection
the other ReaderGroups which were also scheduled had nothing to do in this case -> which causes unnecessary task switches in a realtime environment

I thought it would be simpler if there was one global reader task, which receives all messages for the connection and then forwards it to the correct readersgroups/datasetreaders

@JanSurft What do you think ?

MartinaFelberFEM: I don't understand, somehow it looks like Julius commit would have already solved all these issues

https://github.com/open62541/open62541/commit/b862f28e4bac4403c39c9cb5d9b436643cde3183

but it doesn't look like the current master branch

MartinaFelberFEM: Hmm, the commit removed UA_PubSubConnection_processNetworkMessage  and changed it to UA_ReaderGroup_processNetworkMessage, but I think the connection function is needed, even though we already know the readerGroup

MartinaFelberFEM: Just my thoughts (not sure if I have understood the current implementation correctly)

There's one socket for every connection -> that's ok

BUT: if there are multiple ReaderGroups, then every ReaderGroup has a SubscribeCallback
The ReaderGroup which is scheduled first calls connection->channel->receive() and reads the received data from the socket. -> With it's configured timeout. 

If all ReaderGroups have the same subscribe/corresponding publishing interval then the first 
ReaderGroup receives all network messages, even though they are meant for other ReaderGroups as well.
And then it does not forward the messages to the other ReaderGroups anymore.

And even if we say that it's unlikely that all ReaderGroups of a connection have the same subscribing interval, because it does not make a lot of sense to configure multiple WriterGroups with the same publishing interval, it could still be a problem if the interval difference is small enough and the network messages of other readergroups have already been received ...

MartinaFelberFEM: UA_ReaderGroup_subscribeCallback
receiveBufferedNetworkMessage
calls receive() -> reads everything that's available at the moment on the socket -> maybe also data for other ReaderGroups
decodeAndProcessNetworkMessage

- thinks that it only has network messages for this ReaderGroup
- param connection would be given to check all ReaderGroups, but there are problems with decodeNetworkMessage because verifyAndDecryptNetworkMessage also uses a specific ReaderGroup, UA_NetworkMessage_decodeHeaders would probably work if we call it for multiple ReaderGroups ...

JanSurft:
You can try to rebase your changes on #4480 , the tests should run successfully again.

MartinaFelberFEM:
@JanSurft Many thanks for your fix and sorry that I haven't finished my PR sooner.
I close this pull request because I have so many merge conflicts.
new PR: https://github.com/open62541/open62541/pull/4520

PR #4458, opened 2021-06-07
Subscriber message receive timeout implementation for fast-path

MartinaFelberFEM:


JanSurft:
Hi, if you want to you can use this line as the first phrase in your PR header: 

"Depends on: #4449"

That way, the dependent pull request blocks as long as it is not integrated

Like e.g. in this closed PR: #4427

MartinaFelberFEM:
Thanks for the hint @JanSurft !

Do I have to trigger the bot?

JanSurft:
> Thanks for the hint @JanSurft !
> 
> Do I have to trigger the bot?
I thought it might immediately be triggered after changing the header.
Now I think (hope) the bot will be triggered as soon as you add some new changes to your PR. :)

jpfr:
Nice!

PR #4460, opened 2021-06-08
fix(pubsub): checkReaderIdentifier msg content check

MartinaFelberFEM:
The function checkReaderIdentifier needs the
GroupHeader, WriterGroupId and DataSetWriterId in the received message
to find the correct reader.

The check at the beginning shall verify that all of these are available.
But the check was wrong.
There a segfault can happen if for example the payload header is
disabled.

Now the check should be correct and return BADNOTIMPLEMENTED
if one of the mandatory parameters is not available in the received
message.


MartinaFelberFEM:
build fails at sphinx installation

jpfr:
👍

PR #4464, opened 2021-06-10
fix(pubsub): Fix conversion error

shifty91:
Hi,

the PubSub code with XDP enabled, does not compile. Tested with gcc 10 and libbpf-0.3. Fix it.

Thanks, Kurt


jpfr:
Thanks Kurt 👍

PR #4473, opened 2021-06-15
fix(pubsub): Fix IPv6 Multicast Support

shifty91:
Hi,

currently PubSub over IPv6 Multicast on Linux doesn't work. Enable it.

Tested on Linux v5.10 with the PubSub examples.

Thanks, Kurt

### DIFF
--- a/plugins/ua_pubsub_udp.c
+++ b/plugins/ua_pubsub_udp.c
@@ -18,6 +18,7 @@
 typedef struct {
     int ai_family;                    /* Protocol family for socket. IPv4/IPv6 */
     struct sockaddr_storage ai_addr; /* https://msdn.microsoft.com/de-de/library/windows/desktop/ms740496(v=vs.85).aspx */
+    socklen_t ai_addrlen;            /* Address length */
     struct sockaddr_storage intf_addr;
     UA_UInt32 messageTTL;
     UA_Boolean enableLoopback;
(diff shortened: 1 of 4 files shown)


plugins/ua_pubsub_udp.c
```diff
@@ -530,19 +529,13 @@ UA_PubSubChannelUDPMC_receive(UA_PubSubChannel *channel, UA_ByteString *message,
...
-            }
-#if UA_IPV6
+        ssize_t messageLength = UA_recvfrom(channel->sockfd, (message->data + dataLength), remainingMessageLength, 0, NULL, NULL);
+        if(messageLength > 0){
+            dataLength += (size_t)messageLength;
+            remainingMessageLength -= (size_t)dataLength;
         } else {
-        //TODO implement recieve for IPv6
-#endif
+            retval = UA_STATUSCODE_BADINTERNALERROR;
```
jpfr, line 537: Add a UA_LOG_WARNING in case the receive fails.
See above for a copy/paste example.

The same for the select-call above.

jpfr (commented):
Very nice unborking of the IPV6 code path.
Two minor requests since you are already deep into this subsystem.

1. Add more logging in case of failures. In network_tcp.c we have examples where the errno is translated to readable strings for logging.
2. Extend the PubSub unit tests with an IPv6 example. So we can avoid future regressions for the fixes you just made.

plugins/ua_pubsub_udp.c
shifty91: Make sense. I'll add more logging and a test case. Thanks!

jpfr:
A perfect PR.
👍 @shifty91 !

PR #4477, opened 2021-06-17
[Review] Use Pointer within the DataTypeMember structure

andreasebner:
This PR is based on #4408 and the major work is done by @Johnsik. This PR is intended to fix the mentioned problems in the original PR and cleans up the history.

Currently DataTypes con only contain members of a type, which is declared in ns0 or the same ns like the DataType. This restriction is based on the use of an index which refers to the right type entry in the types array. This PR replaces this index and use a pointer which can now also point to other type arrays.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -172,8 +172,13 @@ if(UA_ENABLE_ENCRYPTION)
     target_include_directories(client_encryption PRIVATE "${PROJECT_SOURCE_DIR}/examples")
 endif()
 
-add_example(custom_datatype_client custom_datatype/client_types_custom.c)
-add_example(custom_datatype_server custom_datatype/server_types_custom.c)
+if (NOT (BUILD_SHARED_LIBS AND WIN32))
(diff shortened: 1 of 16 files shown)


andreasebner:
pa_hole_datatype

@jpfr  This is the pahole output for the UA_DataType structure. The UA_DataTypeMember strucuture was not contained in the pahole output- maybe you can give me a hint how to get it. I was able to replace the pointer with the structure with the "-p" option. But the result did not contain any memory information.

pa_hole_datatype_p_option

jpfr:
UA_DataType:
- the typeindex member can be removed.
- The bitfield can be reduced to a 16bit base type

UA_DataTypeMember:
- Without pahole the padding can be checked with sizeof and offsetof in the gdb debugger. Often padding around 64bit  pointers is the biggest. Check if it helps to bunch the pointers together.

Make sure that jonsik appears author or co-author in the commits.

otherwise very nice!!

jpfr:
Made some small improvements, rebased and merged.
Good job @Johnsik @andreasebner !

PR #4480, opened 2021-06-21
[REVIEW] fix(pubsub): mutiple readers same network message

JanSurft:
This is a (maybe temporary) minimally invasive fix for multiple network messages from different writergroups that arrive at the same time. For this to work the fix enables iteration over all configured readergroups and their dataset readers for decoding (verify and decrypt) and processing each network message in a current buffer.

### DIFF
--- a/src/pubsub/ua_pubsub_reader.c
+++ b/src/pubsub/ua_pubsub_reader.c
@@ -1806,10 +1806,12 @@ decodeNetworkMessage(UA_Server *server,
                 #ifdef UA_DEBUG_DUMP_PKGS
                 UA_dump_hex_pkg(buffer->data, buffer->length);
                 #endif
-                break;
+                /* break out of all loops when first verify & decrypt was successful */
+                goto loops_exit;
             }
         }
     }
+loops_exit:
 
     // TODO check if warning is correct here and error code carries correct info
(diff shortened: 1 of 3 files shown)


src/pubsub/ua_pubsub_reader.c
```diff
@@ -1777,20 +1782,40 @@ decodeNetworkMessage(const UA_Logger *logger,
...
+                                                    buffer,
+                                                    currentPosition,
+                                                    currentNetworkMessage,
+                                                    readerGroup);
+                UA_CHECK_STATUS_WARN(rv, return rv, logger, UA_LOGCATEGORY_SERVER,
+                                     "Subscribe failed. verify and decrypt network message failed.");
+
+                #ifdef UA_DEBUG_DUMP_PKGS
+                UA_dump_hex_pkg(buffer->data, buffer->length);
+                #endif
```
jpfr, line 1808: Break out of both for-loops.
Goto?

jpfr:
Just a minor comment.
Looks good otherwise!

jpfr:
Nice! Merged.
Please take another look at the tests that were changed here.
We don't like commented-out code. If a test is no longer needed, remove it.

PR #4481, opened 2021-06-21
[REVIEW] Pubsub encryption for FPM

opcua-tsn-team-kalycito:


src/pubsub/ua_pubsub_reader.c
```diff
@@ -218,9 +218,10 @@ UA_DataSetReader_generateDataSetMessage(UA_Server *server, UA_DataSetMessage *da
 }
 
 static UA_StatusCode
-UA_DataSetReader_generateNetworkMessage(UA_PubSubConnection *pubSubConnection, UA_DataSetReader *dataSetReader,
+UA_DataSetReader_generateNetworkMessage(UA_PubSubConnection *pubSubConnection, UA_ReaderGroup *readerGroup,
                                         UA_DataSetMessage *dsm, UA_UInt16 *writerId, UA_Byte dsmCount,
                                         UA_NetworkMessage *nm) {
+    UA_DataSetReader *dataSetReader = LIST_FIRST(&readerGroup->readers);
```
jpfr, line 224: Always taking the first reader prevents us generating a network message for the other readers when there is more than one.

Better forward the reader **and** the readergroup.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -2143,12 +2182,32 @@ generateNetworkMessage(UA_PubSubConnection *connection, UA_WriterGroup *wg,
 }
 
 static UA_StatusCode
-sendBufferedNetworkMessage(UA_Server *server, UA_PubSubConnection *connection,
+sendBufferedNetworkMessage(UA_Server *server, UA_WriterGroup *writerGroup,
```
jpfr, line 2185: The function is called `sendBufferedNetworkMessage`. It should do exactly that and not more.

Please pull the encryption out of here and do it in the caller.
Likewise pull `UA_NetworkMessage_updateBufferedMessage` out of here.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -248,6 +249,29 @@ UA_DataSetReader_generateNetworkMessage(UA_PubSubConnection *pubSubConnection, U
...
+#ifdef UA_ENABLE_PUBSUB_ENCRYPTION
+    if(readerGroup->config.securityMode > UA_MESSAGESECURITYMODE_NONE) {
+        nm->securityEnabled = true;
+        nm->securityHeader.networkMessageSigned = true;
+        if(readerGroup->config.securityMode >= UA_MESSAGESECURITYMODE_SIGNANDENCRYPT)
+            nm->securityHeader.networkMessageEncrypted = true;
+        nm->securityHeader.securityTokenId = readerGroup->securityTokenId;
+
+        /* Generate the MessageNonce */
+        UA_ByteString_allocBuffer(&nm->securityHeader.messageNonce, 8);
```
jpfr, line 262: An additional malloc for every generated network message looks like a lot of overhead.
Avoid malloc in the RT path!

You know this is 8byte long. So could be statically allocated and/or reused between network messages.

jpfr (commented):
See the inline comments.

In general, there is now a lot of duplicate functionality between the normal and the RT path.

Encryption/decryption can be performed "in-situ" on the cleartext buffer.
Pull that functionality into separate functions that are called both inside and outside the RT path. Reduce duplicate functionality as much as possible.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1953,14 +1979,84 @@ decodeAndProcessNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
...
+                                                &payLoadPosition,
+                                                &currentNetworkMessage,
+                                                readerGroup);
+            UA_CHECK_STATUS_WARN(rv, return rv, &server->config.logger, UA_LOGCATEGORY_SERVER,
+                                 "Subscribe failed. verify and decrypt network message failed.");
+            UA_NetworkMessage_clear(&currentNetworkMessage);
+            return rv;
+        }
+        else{
+            return UA_STATUSCODE_BADREQUESTTYPEINVALID;
```
jpfr, line 2015: Exit early with this error code instead of putting the else at the end.
That saves one level of indendation for the happy-path.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1953,14 +1979,84 @@ decodeAndProcessNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
...
+        }
```
jpfr, line 2016: End the if-case with return UA_STATUSCODE_GOOD.
That saves one level of indendation instead of having an else clause.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1865,10 +1892,9 @@ decodeNetworkMessage(UA_Server *server,
     UA_dump_hex_pkg(buffer->data, buffer->length);
 #endif
 
-    UA_StatusCode rv = UA_NetworkMessage_decodeHeaders(
-        buffer, currentPosition, currentNetworkMessage);
-    UA_CHECK_STATUS_ERROR(rv, return rv, logger, UA_LOGCATEGORY_SERVER,
-                          "PubSub receive. decoding headers failed");
+    /* Decode the headers and decrypt the NM */
+    UA_StatusCode rv = decodeHeadersAndDecryptNM(server, connection->readerGroups.lh_first, buffer);
```
jpfr, line 1896: This change is wrong.

- The same code-path calls `verifyAndDecryptNetworkMessage` below. So the decrypt routine is called twice on the same buffer.
- I think we need to have currentNetworkMessage updated to contain the decrypted headers. This is not done with the new code.

Just keep the old code here.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -2131,23 +2171,26 @@ generateNetworkMessage(UA_PubSubConnection *connection, UA_WriterGroup *wg,
...
-                 UA_Byte *msgEnd) {
+               UA_Byte *signStart, UA_Byte *encryptStart,
+               UA_Byte *msgEnd) {
     UA_StatusCode rv;
     void *channelContext = wg->securityPolicyContext;
 
+    size_t sigSize = wg->config.securityPolicy->symmetricModule.cryptoModule.
+                     signatureAlgorithm.getLocalSignatureSize(channelContext);
+
+    if(wg->config.rtLevel == UA_PUBSUB_RT_FIXED_SIZE)
```
jpfr, line 2191: I don't understand why this is required.
This should be documented in a good comment.
I prefer to have uniform code paths for rt/non-rt as much as possible.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -2297,20 +2337,41 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
...
+                                              writerGroup->bufferedMessage.encryptBuffer.data + writerGroup->bufferedMessage.encryptBuffer.length);
+
+            if(rv != UA_STATUSCODE_GOOD)
+                UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER, "PubSub Encryption failed");
+        }
+
+        UA_StatusCode res = sendBufferedNetworkMessage(server, connection, &writerGroup->bufferedMessage.encryptBuffer,
+                                                       &writerGroup->config.transportSettings);
+#else
+        UA_StatusCode res = sendBufferedNetworkMessage(server, connection, &writerGroup->bufferedMessage.buffer,
```
jpfr, line 2366: The send part is identical in the else cluase. Pull it out of the ifdef.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -2297,20 +2337,41 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
...
+                                                       &writerGroup->config.transportSettings);
+#endif
+
         if(res == UA_STATUSCODE_GOOD) {
             writerGroup->sequenceNumber++;
         } else {
             UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                       "Publish failed. RT fixed size. sendBufferedNetworkMessage failed");
+                         "Publish failed. sendBufferedNetworkMessage failed");
```
jpfr, line 2374: log the statuscode name as well.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1953,14 +1979,84 @@ decodeAndProcessNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
...
+                                               handle padding bytes */
     }
 
 cleanup:
     UA_NetworkMessage_clear(&currentNetworkMessage);
     return rv;
 }
 
+static UA_StatusCode
+decodeHeadersAndDecryptNM(UA_Server *server, UA_ReaderGroup *readerGroup, UA_ByteString *buffer) {
```
jpfr, line 1991: Do not put UA_NetworkMessage_decodeHeaders and verifyAndDecryptNetworkMessage into a shared function.
This forces users to decode the headers twice. Because this function does not return them.

Just call UA_NetworkMessage_decodeHeaders and verifyAndDecryptNetworkMessage separately in the parent context.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1953,14 +1979,84 @@ decodeAndProcessNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
...
+    else {
+        UA_Boolean processed = false;
+        UA_DataSetReader *reader;
+        /* Search the connection to which the given readergroup is connected to */
+        UA_PubSubConnection *connection =
+            UA_PubSubConnection_findConnectionbyId(server, readerGroup->linkedConnection);
+
+        /* choose a correct readergroup for decrypt/verify this message
+        * (there could be multiple) */
+        LIST_FOREACH(readerGroup, &connection->readerGroups, listEntry) {
```
jpfr, line 2027: The readergroup is already known when the function is called.
At least in the RT case above, the calling readergroup is used.

Either the readergroup is known or not. But the handling should be similar for rt/non-rt.

jpfr (commented):
Looks much improved.
Here's a bunch more comments.

jpfr:
Looking good! Merged.

PR #4483, opened 2021-06-22
fix(nodeset_compiler): proper resolution of alias structures

NoelGraf:
The nodeset compiler will be reworked, and some errors in parsing the XML schema will be fixed. Furthermore, the structure of the nodeset compiler is adapted so far that all necessary data for the resolution of the aliases are present. For this, the BSD is extracted from the XML and prepared. In this way, missing information can be added.

Supporting Features / Changes
* The parsing is based on the XML and not on the generated encoding rule.
* Optional fields are processed correctly
* Unions are processed correctly
* The buildEncoding function is no longer needed.


tools/nodeset_compiler/datatypes.py
```diff
@@ -17,12 +17,14 @@
 import xml.dom.minidom as dom
 from base64 import *
 
-__all__ = ['valueIsInternalType', 'Value', 'Boolean', 'Number', 'Integer',
+from type_parser import *
+
+""" __all__ = ['valueIsInternalType', 'Value', 'Boolean', 'Number', 'Integer',
```
andreasebner, line 22: If it is not needed anymore, than we can remove it.

tools/nodeset_compiler/datatypes.py
```diff
@@ -301,14 +235,115 @@ def __parseXMLSingleValue(self, xmlvalue, parentDataTypeNode, parent, alias=None
...
+
+            childValue = ebodypart.firstChild
+            if not childValue == ebodypart.ELEMENT_NODE:
+                childValue = getNextElementNode(childValue)
+            for e in members:
+                    if isinstance(e, StructMember):
+                        if isinstance(e.member_type, BuiltinType):
+                            if e.is_array:
+                                values = []
+                                # [WIP] The Values still have to be run through!
```
andreasebner, line 320: Still WIP or solved?

tools/schema/Opc.Ua.NodeSet2.Reduced.xml
```diff
@@ -1716,7 +1716,3116 @@
       <Reference ReferenceType="HasTypeDefinition">i=72</Reference>
     </References>
     <Value>
-    <ByteString xmlns="http://opcfoundation.org/UA/2008/02/Types.xsd"></ByteString>
+    <ByteString xmlns="http://opcfoundation.org/UA/2008/02/Types.xsd">PG9wYzpUeXBlRGljdGlvbmFyeQ0KICB4bWxuczpvcGM9Imh0dHA6Ly9vcGNmb3VuZGF0aW9uLm9y
+Zy9CaW5hcnlTY2hlbWEvIg0KICB4bWxuczp4c2k9Imh0dHA6Ly93d3cudzMub3JnLzIwMDEvWE1M
```
andreasebner, line 1720: Would it be also an option that we don't change the reduced nodeset and include insted the ns0-bsd where needed?

andreasebner (commented):
The PR looks quite good. Please fix all the lgtm warnings and look into my review comments. Looking forward to merge it soon.

PR #4495, opened 2021-06-29
[REVIEW] Fix receiving multiple messages in one pubsub callback

JanSurft:
Depends on #4565 
Depends on #4579 

### Before
Receiving multiple messages in one pubsub callback was previously done like so:

1. create a buffer
2. call a recv wrapper on that buffer:
  - fill the buffer with as much raw messages (opcua NetworkMessages) as it can carry and as are messages on the socket
3. process the buffer (decode and process all NetworkMessages)

This behavior led to the situation that the length of single messages in the buffer is unknown upon decoding. Decryption though needs this information for determining the position of the encrypted payload and signature.

### After
This pull request implements the following behavior:
  
  - call a recv wrapper on a functional closure
  - receive each message separately and decode and process with the given closure

### Other

- make `check_pubsub_publish_uadp` independent of timing issues. (use fake timer?)
  - resolved via fake timer

### DIFF
--- a/examples/pubsub/pubsub_subscribe_standalone.c
+++ b/examples/pubsub/pubsub_subscribe_standalone.c
@@ -27,7 +27,7 @@
 
 UA_Boolean running = true;
 static UA_StatusCode
-customDecodeAndProcessCallback(UA_DecodeAndProcessClosure *closure, UA_ByteString *buffer);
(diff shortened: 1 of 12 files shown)


jpfr:
Cool!

PR #4496, opened 2021-06-29
Fix and refactor

mlgiraud:
This PR fixes a memory leak and refactors the code to comply with static analysis tools.


jpfr:
Nice! 👍

PR #4511, opened 2021-07-04
FLC data types model design file

smolinid:
Initial model design file to generate some of the FLC (FX) data types.


andreasebner:
Hi @smolinid, can you tell us what this PR is for? I think all FX related work should currently not be public available since it is still WIP.

smolinid:
> Hi @smolinid, can you tell us what this PR is for? I think all FX related work should currently not be public available since it is still WIP.

Hello! Got it! I will close it then.

PR #4518, opened 2021-07-08
[Review] Fix(plugin): Openssl versions less than 1.1

ccvca:
This cleans up the version specific openssl-Code to a separate file and fixes compatabilit with openssl 1.0 (tested on Ubuntu 16.04 Docker)

I've removes this version distinguish, as the first one works just fine with the new API:
https://github.com/open62541/open62541/blob/b82ad4df1b751e9a301102b32fea518394d00c19/plugins/crypto/openssl/ua_pki_openssl.c#L179-L185

There were some 1.0 bugs, which I tried to fix, but as I do not have CTT available, I was not able to test this.

This code does not work with 1.0, as `X509_get_key_usage` has been introduced in 1.1.
https://github.com/open62541/open62541/blob/b82ad4df1b751e9a301102b32fea518394d00c19/plugins/crypto/openssl/ua_pki_openssl.c#L509-L513

This is my replacement, which should do the same:
https://github.com/open62541/open62541/blob/4cf032d43b3413a82b9db47cd4afffcdfb982c52/plugins/crypto/openssl/ua_pki_openssl.c#L498-L500

This does also not work with 1.0:
https://github.com/open62541/open62541/blob/4cf032d43b3413a82b9db47cd4afffcdfb982c52/plugins/crypto/openssl/ua_pki_openssl.c#L555

This is the fix:
https://github.com/open62541/open62541/blob/4cf032d43b3413a82b9db47cd4afffcdfb982c52/plugins/crypto/openssl/ua_openssl_version_abstraction.h#L27-L29

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -947,6 +947,7 @@ endif()
 if(UA_ENABLE_ENCRYPTION_OPENSSL OR UA_ENABLE_AMALGAMATION)
 list(INSERT default_plugin_sources 0
      ${PROJECT_SOURCE_DIR}/plugins/crypto/openssl/securitypolicy_openssl_common.h
+     ${PROJECT_SOURCE_DIR}/plugins/crypto/openssl/ua_openssl_version_abstraction.h
      ${PROJECT_SOURCE_DIR}/plugins/crypto/openssl/securitypolicy_openssl_common.c
      ${PROJECT_SOURCE_DIR}/plugins/crypto/openssl/ua_openssl_basic128rsa15.c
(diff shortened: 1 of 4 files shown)


jpfr:
Looks great!
Please add the new file to the Cmakelists.txt.
So we can continue to build single file releases.

can be merged then.

ccvca:
https://github.com/open62541/open62541/blob/662e57e9bf354dfb12eb5ef384674ce663c71e8a/CMakeLists.txt#L950
 has been added.

jpfr:
Cool!

PR #4523, opened 2021-07-12
[REVIEW] feat(server) Made it possible to control if base nodes is initialized.

ED19F:
Added flags to be able to controll if base nodes is initialized. This feature makes it easier to work with dynamic information models, as you do not always want to use a predefined node store.


jpfr:
Hey,

I understand the intention and like the change.
But it requires better documentation.

The docs are generated from the header file. 
The meaning of these parameters should be documented there absolutely unambiguous.
It should be possible to understand them without looking at the code.

Keep up the good work!

ED19F:
Hey @jpfr

Got the information that a release will be made soon and we would have liked to include this commit in that release.
Do you have the opportunity to review this change and let me know if you have any comments?

I'm troubleshooting why some test fails, do you see what might cause be the problem?

PR #4532, opened 2021-07-15
(fix) fixing pubsub dataset representation in information model

StenGruener:
A minor fix here: published datasets should be connected via hasComponent reference

image


jpfr:
👍

Issue #2816, opened 2019-06-14
How to create a certification and send this to server


AndyarcherReset:
Hi,
Now I want to create a UA client and then connect to a third part UA server with the Basic128Rsa15 and Basic256.
But seems that do not support this, right?

SebsSa:
Right now, I'm working the other way around.
I want to generate a server and run it with Basic128Rsa15 and only allow a specific third party to use it. (UAExpert in my case)

That's why I'm not an expert, but maybe I can help you anyway. 


Read this through for yourself: 
https://github.com/open62541/open62541/issues/2637

and maybe use :
https://github.com/open62541/open62541/blob/master/tools/certs/create_self-signed.py  and
https://github.com/open62541/open62541/blob/master/examples/encryption/client_encryption.c

jpfr:
This is definitely supported.

See here the example:
https://github.com/open62541/open62541/blob/master/examples/encryption/client_encryption.c

You can define which SecurityPolicy to use in the client config:
https://open62541.org/doc/current/client.html#client-configuration

AndyarcherReset:
> This is definitely supported.
> 
> See here the example:
> https://github.com/open62541/open62541/blob/master/examples/encryption/client_encryption.c
> 
> You can define which SecurityPolicy to use in the client config:
> https://open62541.org/doc/current/client.html#client-configuration


Hi,
From the example client_encryption.c, I just see that can load the certificate.
But how can I create a certificate within the code? Or should I use another way to do this by myself?

jpfr:
https://github.com/open62541/open62541/blob/master/tools/certs/create_self-signed.py

Issue #4372, opened 2021-04-27
Upcoming mbedtls V3 contains breaking changes to API 


GoetzGoerisch:
## Description

With mbedtls moving to V3 they are unifying the internal headers into library/* which breaks current implementations.
see https://github.com/ARMmbed/mbedtls/pull/4164

## Background Information / Reproduction Steps
The [sample server](https://github.com/umati/Sample-Server) uses mbedtls with automated [dependency updates](https://github.com/umati/Sample-Server/pull/222).
The latest changes from ARMmbed/mbedtls@b5939e8 to ARMmbed/mbedtls@12f93f4. breaks the CI, as it upgrades from V2 to V3. 
In detail it is this commit ARMmbed/mbedtls@ea0a865

```
 Move entropy_poll.h to library

`entropy_poll.h` is not supposed to be used by application code and
is therefore being made internal.

```

Used CMake options:

```bash
cmake -DUA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS:BOOL=ON -DUA_ENABLE_SUBSCRIPTIONS_EVENTS:BOOL=ON -DUA_NAMESPACE_ZERO:STRING=FULL -DUA_ENABLE_ENCRYPTION:BOOL=1 -DUA_ENABLE_ENCRYPTION_MBEDTLS:BOOL=1 

```
https://github.com/umati/Sample-Server/blob/2d282e64dcccaf0b2591b6df46436e3c5785a37a/.github/CMakeLists.txt#L28

## Checklist

Please provide the following information:

 - [X] open62541 Version (release number or git tag): 0c2cb1c
 - [ ] Other OPC UA SDKs used (client or server): 
 - [X] Operating system: Linux
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jpfr:
Two options. 

1. We provide code that works for v2 and v3.
Potentially with a compiler macro that comes from a standard header.

2. Fix mbedtls to a long-term support branch.
That should be available as a lot of software has that same problem.

GoetzGoerisch:
Thanks,

would prefer solution 1. 
I'm going to look into it.

GoetzGoerisch:
[...]

PR #4535, opened 2021-07-16
[Review] feat(sec): Create certificate in C using OpenSSL/LibreSSL

ccvca:
Create a self-signed encryption certificate in C (only implemented with OpenSSL/LibreSSL).

For demonstration purpose, `UA_CreateCertificate` is implemented as a fallback in `server_encryption.c`, if no files are passed using the command line.

Resolves: #2816

### DIFF
--- a/plugins/crypto/openssl/ua_openssl_create_certificate.c
+++ b/plugins/crypto/openssl/ua_openssl_create_certificate.c
@@ -1,0 +1,363 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ *    Copyright 2021 (c) Christian von Arnim, ISW University of Stuttgart (for VDW and umati)
+ *
+ */
+
+#include <open62541/plugin/create_certificate.h>
+
+#include "securitypolicy_openssl_common.h"
+
+#if defined(UA_ENABLE_ENCRYPTION_OPENSSL) || defined(UA_ENABLE_ENCRYPTION_LIBRESSL)
+
+
+#include <openssl/pem.h>
+#include <openssl/x509v3.h>
+
+/**
+ * Join an array of UA_String to a single NULL-Terminated UA_String
+ * separated by character sep
+ */
+static UA_StatusCode
+join_string_with_sep(const UA_String *strings, size_t stringsSize,
+                     char sep, UA_String *out) {
+    if(!out)
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+
+    UA_String_clear(out);
+    size_t totalSize = stringsSize;
+    for(size_t iStr = 0; iStr < stringsSize; ++iStr) {
+        totalSize += strings[iStr].length;
+    }
+
+    UA_ByteString_allocBuffer(out, totalSize);
+    if(!out->data) {
+        return UA_STATUSCODE_BADOUTOFMEMORY;
+    }
+
+    size_t pos = 0;
+    for(size_t iStr = 0; iStr < stringsSize; ++iStr) {
+        memcpy(&out->data[pos], strings[iStr].data, strings[iStr].length);
+        pos += strings[iStr].length;
+        out->data[pos] = (UA_Byte) sep;
+        ++pos;
+    }
+    out->data[out->length-1] = 0;
+
+    return UA_STATUSCODE_GOOD;
+}
+
+/**
+ * Search for a character in a string (like strchr).
+ * \todo Handle UTF-8
+ *
+ * \return index of the character or -1 on case of an error.
+ */
+
+static UA_Int32
+UA_String_chr(const UA_String *pUaStr, char needl) {
+    UA_Byte byteNeedl = (UA_Byte)needl;
+    for(size_t i = 0; (size_t)i < pUaStr->length; ++i) {
+        if(pUaStr->data[i] == byteNeedl) {
+            return (UA_Int32) i;
+        }
+    }
+    return -1;
+}
+
+static UA_StatusCode
+add_x509V3ext(X509 *x509, int nid, const char *value) {
+    X509_EXTENSION *ex;
+    X509V3_CTX ctx;
+    X509V3_set_ctx_nodb(&ctx);
+    X509V3_set_ctx(&ctx, x509, x509, NULL, NULL, 0);
+    ex = X509V3_EXT_conf_nid(NULL, &ctx, nid, value);
+    if(!ex)
+        return UA_STATUSCODE_BADINTERNALERROR;
+    X509_add_ext(x509, ex, -1);
+    X509_EXTENSION_free(ex);
+    return UA_STATUSCODE_GOOD;
+}
+
+UA_StatusCode
+UA_CreateCertificate(const UA_Logger *logger,
+                     const UA_String *subject, size_t subjectSize,
+                     const UA_String *subjectAltName, size_t subjectAltNameSize,
+                     size_t keySizeBits, UA_CertificateFormat certFormat,
+                     UA_ByteString *outPrivateKey, UA_ByteString *outCertificate) {
+    if(!outPrivateKey || !outCertificate || !logger || !subjectAltName ||
+       !subject || subjectAltNameSize == 0 || subjectSize == 0 ||
+       (certFormat != UA_CERTIFICATEFORMAT_DER && certFormat != UA_CERTIFICATEFORMAT_PEM ))
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+
+    /* Use the maximum size */
+    if(keySizeBits == 0)
+        keySizeBits = 4096;
+
+    UA_ByteString_init(outPrivateKey);
+    UA_ByteString_init(outCertificate);
+
+    UA_String fullAltSubj = UA_STRING_NULL;
+    UA_Int32 serial = 1;
+
+    /** \TODO: Seed Random generator
+    * See: (https://www.openssl.org/docs/man1.1.0/man3/RAND_add.html) */
+    BIO *memCert = NULL;
+    BIO *memPKey = NULL;
+
+    UA_StatusCode errRet = UA_STATUSCODE_GOOD;
+
+    BIGNUM *exponent = BN_new();
+    EVP_PKEY *pkey = EVP_PKEY_new();
+    X509 *x509 = X509_new();
+    RSA *rsa = RSA_new();
+
+    if(!pkey || !x509 || !exponent || !rsa) {
+        errRet = UA_STATUSCODE_BADOUTOFMEMORY;
+        goto cleanup;
+    }
+
+    UA_LOG_INFO(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                "Create Certificate: Generating RSA key. This may take a while.");
+
+    if(BN_set_word(exponent, RSA_F4) != 1) {
+        UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                     "Create Certificate: Setting RSA exponent failed.");
+        errRet = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    if(RSA_generate_key_ex(rsa, (int) keySizeBits, exponent, NULL) != 1) {
+        UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                     "Create Certificate: Generating RSA key failed.");
+        errRet = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    if(EVP_PKEY_assign_RSA(pkey, rsa) != 1) {
+        UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                     "Create Certificate: Assign RSA key failed.");
+        errRet = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+    /* rsa will be freed by pkey */
+    rsa = NULL;
+
+    /* x509v3 has version 2
+     * (https://www.openssl.org/docs/man1.1.0/man3/X509_set_version.html) */
+    if(X509_set_version(x509, 2) != 1) {
+        UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                     "Create Certificate: Setting version failed.");
+        errRet = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    if(ASN1_INTEGER_set(X509_get_serialNumber(x509), serial) != 1) {
+        UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                     "Create Certificate: Setting serial number failed.");
+        /* Only memory errors are possible */
+        errRet = UA_STATUSCODE_BADOUTOFMEMORY;
+        goto cleanup;
+    }
+
+    if(X509_gmtime_adj(X509_get_notBefore(x509), 0) == NULL) {
+        UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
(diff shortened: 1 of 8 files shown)


GoetzGoerisch:
Missing Mbed TLS support could be implemented after #4372 is solved.

ccvca:
The  "AppVeyor"-Status looks goot, but Github does not show the correct result.

include/open62541/plugin/securitypolicy.h
```diff
@@ -320,6 +321,31 @@ struct UA_SecurityPolicy {
     void (*clear)(UA_SecurityPolicy *policy);
 };
 
+#ifdef UA_ENABLE_ENCRYPTION
```
jpfr, line 324: Move this definition to the plugins/include folder.

include/open62541/plugin/securitypolicy.h
```diff
@@ -320,6 +321,31 @@ struct UA_SecurityPolicy {
...
+enum UA_CertificateFormat {
+    UA_CERTIFICATE_FORMAT_DER,
+    UA_CERTIFICATE_FORMAT_PEM
+};
+
+/**
+ * (Experimental) Create a certificate
```
jpfr, line 331: Mention that this is a self-signed certificate.

include/open62541/plugin/securitypolicy.h
```diff
@@ -320,6 +321,31 @@ struct UA_SecurityPolicy {
...
+ * It is recommended to store the generated certificate in a non volatile memory and
+ * reuse it, so the application can be recognized across several executions.
+ *
+ * \param subject Elements for the subject,
+ *                  e.g. ["C=DE", "O=SampleOrganization", "CN=Open62541Server@localhost"]
+ * \param subjectAltName Elements for SubjectAltName,
+ *                  e.g. ["DNS:localhost", "URI:urn:open62541.server.application"]
+ */
+UA_StatusCode UA_EXPORT
+UA_CreateCertificate(const UA_Logger *logger,
```
jpfr, line 342: Keep consistent with the style of the overall SDK.
Array length are size_t.
The name of the array length is "arrayNameSize".
With a `Size` postfix.

plugins/crypto/mbedtls/ua_mbedtls_create_certificate.c
```diff
@@ -0,0 +1,30 @@
...
+
+UA_StatusCode
+UA_CreateCertificate(const UA_Logger *logger,
+                    UA_String subject[], UA_UInt32 lenSubject,
+                    UA_String subjectAltName[], UA_UInt32 lenSubjectAltName,
+                    UA_ByteString *outPKey, UA_ByteString *outCert,
+                    enum UA_CertificateFormat certFormat) {
+    if(!outPKey || !outCert || !logger || !subjectAltName
+        || !subject || lenSubjectAltName == 0 || lenSubject == 0) {
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
```
jpfr, line 21: If it builds, then it should work.
I prefer to have this function undefined for mbedtls (linker error for a missing function) than to let it fail only at runtime.
That reduces the potential surprise at runtime.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,366 @@
...
+
+#define RSA_KEY_SIZE 4096
+#include <openssl/pem.h>
+#include <openssl/x509v3.h>
+
+/**
+ * Join an array of UA_String to a single NULL-Terminated UA_String
+ * separated by character sep
+ */
+static UA_StatusCode UA_String_join_nullterm(
```
jpfr, line 22: "nullterm" is specialized nomenclature not verybody knows.
Call this something easy to understand. Like join_string_with_sep or similar.

Format this definition according to CONTRIBUTING.md.
See the comments on the public header for the representation of arrays.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,366 @@
...
+    const UA_String strings[], size_t lenStrings,
+    char sep,
+    UA_String *out) {
+    if(!out)
+    {
```
jpfr, line 27: We don't need all these braces/newlines here.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,366 @@
...
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+    }
+    UA_String_clear(out);
+    size_t totalSize = lenStrings;
+    size_t pos = 0;
```
jpfr, line 32: Move the definition of pos below.
Where it is first used.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,366 @@
...
+    for(size_t iStr = 0; iStr < lenStrings; ++iStr) {
+        totalSize += strings[iStr].length;
+    }
+
+    out->length = totalSize;
```
jpfr, line 37: UA_ByteString_allocBuffer is shorter / cleaner.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,366 @@
...
+ *
+ * \return index of the character or -1 on case of an error.
+ */
+
+static UA_Int32
+UA_String_chr(const UA_String *pUaStr, char needl) {
+
+    UA_Byte byteNeedl = (UA_Byte)needl;
+
+    for(UA_Int32 i = 0; (size_t)i < pUaStr->length; ++i) {
```
jpfr, line 66: indices are size_t.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,366 @@
...
+        char field[16];
+        UA_Int32 sep = UA_String_chr(&subject[iSubject], '=');
+        if(sep == -1 || sep == 0 || ((size_t) sep == (subject[iSubject].length - 1)) || sep >= 15) {
+            UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                           "Create Certificate: Subject must contain one '=' with "
+                           "content before and after.");
+            errRet = UA_STATUSCODE_BADINTERNALERROR;
+            goto cleanup;
+        }
+        memcpy(field, subject[iSubject].data, (size_t) sep);
```
jpfr, line 208: You can just replace the '=' with a '\0' in the original string.
Than you don't have to copy to a temporary variable and can just forward the original string with the name.
The replacement can be rolled back later. So that the original string is left identical when returning.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,366 @@
...
+            errRet = UA_ByteString_copy(&tmpPem, outCert);
+            if(errRet != UA_STATUSCODE_GOOD) {
+                UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                            "Create Certificate: Copy PEM Certificate failed.");
+                goto cleanup;
+            }
+            break;
+        }
+        default: {
+            UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
```
jpfr, line 349: set the return code to something bad.
this could be checked early on in the function.

jpfr (commented):
Looks great!
A couple of comments in the technical details.
Can be merged then.

plugins/crypto/openssl/ua_openssl_create_certificate.c
ccvca: This causes a SegFault, when using `UA_STRING_STATIC`.

ccvca:
I've resolved all comments except one, which causes a SegFault.

jpfr:
So why don’t the unit tests fail?

try to pinpoint the problem with Valgrind.
Often times it becomes quite obvious when doing that.

ccvca:
I've investigate the problem a little bit further using this short snippet:
```c
#include <open62541/types.h>

void func(UA_String testStr) {
    UA_Byte orig = testStr.data[1];
    testStr.data[1] = 'D';
    testStr.data[1] = orig;
}

int main(int argc, char* argv[]) {
    UA_String testStr = UA_STRING_STATIC("ABC");
    func(testStr);
    return 0;
}
```

Running this application without an debugger is fine (at least no crash). But gdb doesn't like it:

```txt
GNU gdb (GDB) 8.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "i686-w64-mingw32".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from D:\Programmieren\Cpp\Open62541_Tests\Test_StaticSEGFAULT\build\Test_StaticMem_SegFault.exe...done.
(gdb) r
Starting program: D:\Programmieren\Cpp\Open62541_Tests\Test_StaticSEGFAULT\build\Test_StaticMem_SegFault.exe 
[New Thread 26552.0x461c]
[New Thread 26552.0x5ae4]
[New Thread 26552.0x1fc]

Thread 1 received signal SIGSEGV, Segmentation fault.
func (testStr=...) at D:\Programmieren\Cpp\Open62541_Tests\Test_StaticSEGFAULT\Test_StaticMem_SegFault.cpp:6
6           testStr.data[1] = 'D';
(gdb) c
Continuing.

Thread 1 received signal SIGSEGV, Segmentation fault.
func (testStr=...) at D:\Programmieren\Cpp\Open62541_Tests\Test_StaticSEGFAULT\Test_StaticMem_SegFault.cpp:6
6           testStr.data[1] = 'D';
(gdb) c
Continuing.
[Thread 26552.0x1fc exited with code 3221225477]
[Thread 26552.0x5ae4 exited with code 3221225477]
[Inferior 1 (process 26552) exited with code 030000000005]
(gdb) q
```

DrMemory also does not totally agree on this:
```txt
~~Dr.M~~ Dr. Memory version 2.3.18722
~~Dr.M~~ Running "Test_StaticMem_SegFault.exe"
~~Dr.M~~ Using system call file c:\Portable\DrMemory-Windows-2.3.18722\drmemory\logs\symcache\syscalls_wow64.txt
~~Dr.M~~ 
~~Dr.M~~ Error #1: WARNING: writing to readonly memory 0x00404045-0x00404046
~~Dr.M~~ # 0 func               [D:/Programmieren/Cpp/Open62541_Tests/Test_StaticSEGFAULT/Test_StaticMem_SegFault.cpp:6]
~~Dr.M~~ # 1 main               [D:/Programmieren/Cpp/Open62541_Tests/Test_StaticSEGFAULT/Test_StaticMem_SegFault.cpp:12]
~~Dr.M~~ Note: @0:00:00.365 in thread 12276
~~Dr.M~~ Note: instruction: mov    $0x44 -> (%eax)
```

So I don't think the proposed optimisation is a good idea.

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,365 @@
...
+
+#define RSA_KEY_SIZE 4096
+#include <openssl/pem.h>
+#include <openssl/x509v3.h>
+
+/**
+ * Join an array of UA_String to a single NULL-Terminated UA_String
+ * separated by character sep
+ */
+static UA_StatusCode join_string_with_sep(
```
jpfr, line 24: Fix the alignment.
Use UA_String *strings instead of UA_String strings[].
Do this consistent everywhere.

plugins/include/open62541/plugin/create_certificate.h
```diff
@@ -0,0 +1,39 @@
...
+#ifdef UA_ENABLE_ENCRYPTION
+enum UA_CertificateFormat {
+    UA_CERTIFICATE_FORMAT_DER,
+    UA_CERTIFICATE_FORMAT_PEM
+};
+
+/**
+ * (Experimental) Create a self-signed certificate
+ *
+ * It is recommended to store the generated certificate in a non volatile memory and
```
jpfr, line 23: volatile has a specific meaning in C.
If you mean "store on disk for reuse", just say so.

plugins/include/open62541/plugin/create_certificate.h
```diff
@@ -0,0 +1,39 @@
...
+ * reuse it, so the application can be recognized across several executions.
+ *
+ * \param subject Elements for the subject,
+ *                  e.g. ["C=DE", "O=SampleOrganization", "CN=Open62541Server@localhost"]
+ * \param subjectAltName Elements for SubjectAltName,
+ *                  e.g. ["DNS:localhost", "URI:urn:open62541.server.application"]
+ */
+UA_StatusCode UA_EXPORT
+UA_CreateCertificate(const UA_Logger *logger,
```
jpfr, line 32: Please use pointers instead of array syntax to be consistent with the rest of the SDK.

Is it possible to generalize this function for different key lengths?
That would be great so we don't have to change the API later when a different key length is requested.

jpfr (commented):
Ok. Then leave out the UA_STRING_STATIC part.
This is probably because the string memory is marked as const.
So the compiler puts it into a location that is write-protected.

The const-ness of the string however is not visible in the code part where UA_STRING_STATIC could have been used. By the way, UA_STRING is better than UA_STRING_STATIC. Because UA_STRING uses strlen to find the \0 for the length. 
UA_STRING_STATIC uses sizeof() for the length and only works when you put in a string literal.

Otherwise, I have a few minor comments and one request to generalize the key lenght in the API. Let's bring this in soon!

plugins/include/open62541/plugin/create_certificate.h
ccvca: Parameter is added.
In case a `0` is provided, the largest possible key is generated (4096Bits in the current OPC UA Specification).

ccvca:
I do not see the docker build, but it also failes on the current master.
PS: ghcr.io (github container registry) may be used to provide docker images with github actions instead of an external ci.

jpfr:
Hey @ccvca ,

I just added a commit on top of your branch (I can do that for PR to this repo).
A few cosmetic cleanups and a few resolved bugs that came up from valgrind.

Also, valgrind-testing in the CI is now also enabled for OpenSSL.

Please have a look at the cleanup-commit and fold the changes into your history.

Can be merged when you are done.

This is really cool!

plugins/crypto/openssl/ua_openssl_create_certificate.c
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    if(X509_gmtime_adj(X509_get_notBefore(x509), 0) == NULL) {
+        UA_LOG_ERROR(logger, UA_LOGCATEGORY_SECURECHANNEL,
+                       "Create Certificate: Setting 'not before' failed.");
+        errRet = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    if(X509_gmtime_adj(X509_get_notAfter(x509), (UA_Int64) 60 * 60 * 24 * 356) == NULL) {
```
nokem61, line 172: 356??

ccvca: Changed to 365 :)

ccvca:
> 
> 
> Hey @ccvca ,
> 
> I just added a commit on top of your branch (I can do that for PR to this repo).
> A few cosmetic cleanups and a few resolved bugs that came up from valgrind.
> 
> Also, valgrind-testing in the CI is now also enabled for OpenSSL.
> 
> Please have a look at the cleanup-commit and fold the changes into your history.
> 
> Can be merged when you are done.
> 
> This is really cool!

Thank you.

I've cleaned up the commits.

jpfr:
Nice. Merged.

PR #4538, opened 2021-07-17
fix(nodeset_compiler): Adding the various EventNotifier attributes in the code generation

NoelGraf:
Adding the various EventNotifier attributes in the code generation to indicate if the Node can be used to subscribe to Events or read/write historic Events.


jpfr:
Nice!

PR #4550, opened 2021-07-21
fix(server): revert removal of Dictionaries object

iosb-ina-mr:
The I4AAS companion specification depends on the Dictionaries object of the OPC UA Nodeset2 specification. Commit [bdb233e](https://github.com/open62541/open62541/commit/bdb233e810a7c9f5187c2e53f021c71abdefba80#diff-78df7143ea0c91caf7c6e0969cd003ad1265cbdb359a2fa529228c5263734dfe) removed this node and therefore breaks compatibility to the I4AAS specification.


mlang-de:
How does the I4AAS companion specification depends on the Base Nodeset2 dictionary?
Maybe it is only for me, but I'm interested in understanding the dependency.
TIA.

iosb-ina-mr:
The [I4AAS Specification](https://github.com/OPCFoundation/UA-Nodeset/tree/v1.04/I4AAS) instantiates several objects within the Dictionaries folder as you can see in attached figure. But I can't tell you about the reason why they are there.
Dictionaries

jpfr:
👍

PR #4551, opened 2021-07-22
Add UA_Server_getMethodNodeCallback

jpfr:


include/open62541/server.h
```diff
@@ -1077,9 +1077,21 @@ UA_Server_deleteMonitoredItem(UA_Server *server, UA_UInt32 monitoredItemId);
...
-UA_Server_setMethodNode_callback(UA_Server *server,
-                                 const UA_NodeId methodNodeId,
-                                 UA_MethodCallback methodCallback);
+UA_Server_setMethodNodeCallback(UA_Server *server,
+                                const UA_NodeId methodNodeId,
+                                UA_MethodCallback methodCallback);
+
+/* Backwards compatibility definition */
+#define UA_Server_setMethodNode_callback(server, methodNodeId, methodCallback) \
+    UA_Server_setMethodNodeCallback(server, methodCallback, methodCallback)
```
mlgiraud, line 1086: I think this might be a typing error. The second argument should be the node id, not the callback.

jpfr: whoups. thanks for pointing this out.

PR #4553, opened 2021-07-23
[WIP] Feature/eventloop updated structure

JanSurft:


arch/eventloop_posix_tcp.c
```diff
@@ -167,15 +167,17 @@ TCP_listenSocketCallback(UA_ConnectionManager *cm, UA_FD fd,
 
     /* The socket has opened. Signal it to the application. The callback can
      * switch out the context. So put it into a temp variable.  */
-    void *ctx = cm->initialConnectionContext;
-    cm->connectionCallback(cm, (uintptr_t)fd, &ctx, UA_STATUSCODE_GOOD,
+    void *fdctx_address = NULL;
+    UA_EventLoop_getFDContext(cm->eventSource.eventLoop, fd, &fdctx_address);
+    cm->connectionCallback(cm, (uintptr_t)fd, (void **) fdctx_address, UA_STATUSCODE_GOOD,
```
jpfr, line 172: I don't get the change.
This is about initializing the context of the new connection.
Why do you want to forward the context of the parent fd?
The pointer to the cm and the initialContext should be enough.

At any rate, you want to forward a "true" double-pointer.
So that `connectionCallback` can switch out the pointer to something else if it wants to.

jpfr (commented):
Cool! Just a single comment.
Once we are happy I'd fold some of the commits into a clean history and add you as a co-author.

arch/eventloop_posix_tcp.c
JanSurft: From my point of view, within the TCP_listenSocketCallback, the listensocket will be called with its registered context which can be changed. But this extra work with **getFDContext** done here is unnecessary because the correct address is already passed via _fdcontext_ to this same function.

I modified the code so the existing address is used instead.

JanSurft: Of course its a different situation when we never want to actually call the `cm->connectionCallback` on the listenSocket with its context. 

Then we maybe should change `cm->connectionCallback` to use `newsockfd` and not `fd` shouldn't we?

JanSurft: So there are actually two versions that work and one version that doesn't with the current implementation of the server integration:
Within `TCP_listenSocketCallback:

1. Mixing (Doesn't work): calling `cm->connectionCallback` with the parent `fd` but the soon to be child `ctx` of `newsockfd`
2. Child only (Does work): calling `cm->connectionCallback` with the `newsockfd` and its soon to be child context `ctx`
3. Parent only (Does work): calling `cm->connectionCallback` with the parent `fd` and its already registered `fdcontext` 

Why "mixing" does not work and the others do:

- For the server integration I assume each connection id (e.g. `fd` or `newsockfd` in `TCP_listenSocketCallback`) will always be called with its registered `fdcontext`.
- So if the passed context is the initial one, I will store the connection id within the context so that it can be used during calls of the connection manager.
- If I primarily call the initial context with another connectionId and subsequently call it with a new id it will fail, because the wrong connectionId is registered in the connection context. (i.e. before my change `TCP_listenSocketCallback` would call `cm->connectionCallback` with {`parent-fd`, `child-ctx`} which stores `parent-fd` for any calls to the connection manager and subsequently `TCP_connectionSocketCallback` would call `cm->connectionCallback` with {`child-fd`, `child-ctx`})

JanSurft: If for any reason, we need those calls that use the parent `fd` with the child `ctx` (**Mixing**) the server integration callback can be changed, so that it modifies the `connectionId` of the passed context on every call.

jpfr: >  I assume each connection id (e.g. fd or newsockfd in TCP_listenSocketCallback) will always be called with its registered fdcontext

You are right. It should be that way.
I prefer option 2 (child only).
If required, the parent fd and context can always be passed on via the initialContext.
And that can be solved in the individual implementation "outside" the main eventloop code.

What do you think?

JanSurft: Using the connected socket (the "child" of the listen-socket) and its child-context sounds good to me. After all the callback is named `connectionCallback` which might indicate it will not be called for the initial listen-socket but for any socket that is connected.

Do you think there could be a situation where we want the `cm->connectionCallback` to get the listen-socket and its context for any reason? Because like this the registered context of the listenSocket will never be used anyhwere, is that correct?

Also, what do you mean by 
> If required, the parent fd and context can always be passed on via the initialContext.

?

arch/eventloop_posix_tcp.c
```diff
@@ -266,7 +268,7 @@ TCP_registerListenSocket(UA_ConnectionManager *cm, struct addrinfo *ai) {
         UA_EventLoop_registerFD(cm->eventSource.eventLoop, listenSocket,
                                 UA_POSIX_EVENT_READ,
                                 (void (*)(struct UA_EventSource *, int, void *, short))
-                                TCP_listenSocketCallback, &cm->eventSource, NULL);
+                                TCP_listenSocketCallback, &cm->eventSource, cm->initialConnectionContext);
```
JanSurft, line 269: So revering to my earlier comment about the listen-socket never being called with its context: Is that why here we actually register `NULL` to be the context of the listen-socket, because it is never supposed to be called with any context.

So I can change it to `NULL` again as you had it before?

arch/eventloop_posix_tcp.c
jpfr: > Because like this the registered context of the listenSocket will never be used anyhwere, is that correct? 

Yes. I think that is the case.
At any rate, we forward the connectionManager pointer. That ought to be enough.
And if not, the initialContext can be set up to forward additional information when a new socket goes into the userland callback for the first time.

That also answers the question below.

arch/eventloop_posix_tcp.c
jpfr: Yeah. I think we don't need the initialContext here.
Keeping the pointer to cm/eventSource is enough.

Can be reverted to NULL.

jpfr:
The CI is failing for the multi-threaded build.

Otherwise this looks good now.

I will make a 1.3 branch for the next stable release.
Then we can apply this to master.

arch/eventloop_posix.c
```diff
@@ -0,0 +1,693 @@
...
+        UA_UNLOCK(&el->elMutex);
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
+
+    UA_LOG_INFO(el->logger, UA_LOGCATEGORY_EVENTLOOP, "Starting the EventLoop");
+
+    UA_EventSource *es = el->eventSources;
+    UA_StatusCode res = UA_STATUSCODE_GOOD;
+    while(es) {
+        res |= es->start(es);
```
JanSurft, line 378: > The CI is failing for the multi-threaded build.
> 
> Otherwise this looks good now.
> 
> I will make a 1.3 branch for the next stable release.
> Then we can apply this to master.

Great, regrading multithreading: 

- One multithreading test could be fixed by just adding missing `UA_UNLOCK` calls.
- Another test's (`check_eventloop_tcp`) issue is about the lock being a non-reentrantlock.
  - In the snipped above we call `es->start(es);` within `UA_EventLoop_start` with the lock `elMutex` being in a locked state.
  - `es->start` in case of tcp calls `TCP_eventSourceStart`
  - thereby registering the listen socket via `TCP_registerListenSocketDomainName` -> `TCP_registerListenSocket` -> `UA_EventLoop_registerFD`
  - `UA_EventLoop_registerFD` in turn does a call to `UA_LOCK` which is illegal because it is non-reentrant and allready locked by `UA_EventLoop_start`.
  
**Possible Solutions**

1. make the lock reentrant.
2. call `UA_UNLOCK` before `es->start` and `UA_LOCK` afterwards 
```c
UA_UNLOCK(&el->elMutex);
res |= es->start(es);
UA_LOCK(&el->elMutex);
```

arch/eventloop_posix.c
```diff
@@ -0,0 +1,693 @@
...
+        UA_LOG_WARNING(el->logger, UA_LOGCATEGORY_EVENTLOOP,
+                       "Cannot delete a running EventLoop");
+        UA_UNLOCK(&el->elMutex);
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
+
+    /* Deregister and delete all the EventSources */
+    while(el->eventSources) {
+        UA_EventSource *es = el->eventSources;
+        UA_EventLoop_deregisterEventSource(el, es);
```
JanSurft, line 342: Same as in https://github.com/open62541/open62541/pull/4553/files#r678903699, we do not have a reentrant lock, so `UA_EventLoop_deregisterEventSource` which itself calls LOCK/UNLOCK fails for the already locked mutex of `UA_EventLoop_delete`.

PR #4565, opened 2021-07-28
[REVIEW] Link tests to open62541-testplugins

JanSurft:
There are several tests that currently use the `open62541-plugins` instead of the declared `open62541-testplugins`. This specifically makes a difference in tests where timing relevant data is tested.

For example in `check_pubsub_publish_uadp:326` the server node datetime is tested against a call to `DateTime_now` to be lower than a specific threshold. So it is actually relevant how fast data is decoded and processed on different platforms for this test to run successfully (processing needs to be done in less than `1000000` (magic number?) time-units)

### Update

Some tests needed to be fixed causing a segfault. This is because the DataSetWriter was added before the DataSetFields which causes uninitialized data that is accessed when DeltaFrames is enabled.


jpfr:
Nice!
Just remove the WIP prefix when you are happy.

tests/pubsub/check_pubsub_subscribe_rt_levels.c
```diff
@@ -345,114 +343,122 @@ START_TEST(SetupInvalidPubSubConfig) {
...
-
-    ck_assert(UA_Server_freezeReaderGroupConfiguration(server, readerGroupIdentifier) == UA_STATUSCODE_BADNOTSUPPORTED); // DateTime not supported
-    ck_assert(UA_Server_freezeWriterGroupConfiguration(server, writerGroupIdent) == UA_STATUSCODE_BADNOTSUPPORTED); // Static value source only supported
-    ck_assert(UA_Server_unfreezeReaderGroupConfiguration(server, readerGroupIdentifier) == UA_STATUSCODE_GOOD);
-    ck_assert(UA_Server_unfreezeWriterGroupConfiguration(server, writerGroupIdent) == UA_STATUSCODE_GOOD);
+    UA_DataSetWriterConfig dataSetWriterConfig;
+    memset(&dataSetWriterConfig, 0, sizeof(UA_DataSetWriterConfig));
+    dataSetWriterConfig.name = UA_STRING("Test DataSetWriter");
+    dataSetWriterConfig.dataSetWriterId = 62541;
+    ck_assert(UA_Server_addDataSetWriter(server, writerGroupIdent, publishedDataSetIdent, &dataSetWriterConfig, &dataSetWriterIdent) == UA_STATUSCODE_BADCONFIGURATIONERROR);
```
JanSurft, line 351: If the correct order of calls is used (`addDataSetWriter` after `addDataSetField`), the invalid pubsub configuration is already intercepted at this point, which makes the rest of the nest unnecessary.

JanSurft: Added another test where the commented out codepaths for the reader are tested again.

jpfr:
👍

Issue #4359, opened 2021-04-21
Make encoding APIs public (UA binary and JSON)


jonasgreen88:
## Description

I would like to suggest to expose the APIs of the encoding implementations (currently UA binary and JSON) of the library. These encoding and decoding functions can be used by other purposes, for instance to use the encoding to read/write data of a file or exchange data (transport != OPC UA, e.g. MQTT) using a standardized encoding.

Does anyone see a problem in exposing these APIs? If not, I can make a PR.

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): master
 - [x] Other OPC UA SDKs used (client or server): N/A
 - [x] Operating system: N/A
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jpfr:
Sounds good. But the public API needs to be cleaned up compared to the internal one.

Can you make a proposal?

jpfr:
Then we could also convert the shell tool

/tools/ua2json

to use only the public api.

jonasgreen88:
Sure, we can probably do a try sometime soon. Can we start with the Binary API and take care of the JSON API in a second PR?
When looking in the JSON API I can see that file can probably be separated into a private and a public part and it needs documentation. Can you give more specific examples what you think need to be updated in the Binary API? I guess also some of the generated file need to be mapped as public headers.

jpfr:
The binary encoding is now public API: #4597
JSON comes via the PR of @matkonnerth #4570

PR #4570, opened 2021-07-31
[Review] move the json encoding functions to public includes

matkonnerth:
It would be interesting to use the json decoding/encoding for example to decode a Variant from json description. Related to https://github.com/open62541/open62541/issues/4359

### DIFF
--- a/include/open62541/types.h
+++ b/include/open62541/types.h
@@ -1134,6 +1134,52 @@ UA_decodeBinary(const UA_ByteString *inBuf,
                 void *p, const UA_DataType *type,
                 const UA_DecodeBinaryOptions *options);
 
+typedef struct {
+    const UA_String *namespaces;
+    size_t namespacesSize;
+    const UA_String *serverUris;
+    size_t serverUrisSize;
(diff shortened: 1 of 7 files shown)


jpfr:
Looks good!
I had some comments wrt the API.
Instead of many iterations here, I just added for the binary encoding what I believe the API should be.

The binary encoding was added to the public API via #4597.

Please duplicate/translate the API added to open62541/types.h for the JSON encoding.
That is, translate the API methods and convetions for JSON.
The API signature should remain identical for JSON.
With possibly more members added to the "options-struct" for the JSON encoding.

jpfr:
Nice. This is nearly done now.

A few last requests.

Check that you follow the code style guidelines in CONTRIBUTING.MD.
Especially breaking of overlong lines in the public header (so we generate nice PDF docs) and braces around if/else.

Clean up the commit history. And remove the change to the external dependencies.

Remove the C99 type variable initialization in ua2json.
I’m pretty sure that would break on MSVC if that was activated in the builds.

then we can merge :-)

jpfr:
Nice!

PR #4584, opened 2021-08-07
Add untrustedCertificateRejectionCallback

nokem61:
Simple callback in the moment when we reject an untrusted certificate.

### DIFF
--- a/include/open62541/plugin/pki.h
+++ b/include/open62541/plugin/pki.h
@@ -44,6 +44,10 @@ struct UA_CertificateVerification {
                                           const UA_ByteString *certificate,
                                           const UA_String *applicationURI);
 
+    /* Callback for rejecting an untrusted certificate*/
+    void (*untrustedCertificateRejectionCallback)(
+        const UA_ByteString *rejectedCertificate);
+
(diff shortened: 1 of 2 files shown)


jpfr:
The cert verification plugin does the rejection. Why indicate the result back to the same plugin?

note that you can fork and modify the plugin itself to suit your requirements.

nokem61:
This solution is more universal. It works independently of the plugin. And information about an untrusted certificate is very helpful. You can easily save the certificate to a file, etc.

This solution is very simple and requires little change. I can do it in a plugin.

fmeerkoetter:
as this is something we have implemented over and over again - this is useful API functionality!

nokem61:
> as this is something we have implemented over and over again - this is useful API functionality!

I have the same opinion on this matter.  This is very useful API functionality. Currently, when we change something in the plug-in codes, we must always do it when we want to upload a new version of open62541. And that's how it would be in the configuration codes and you wouldn't have to remember about it anymore.

nokem61:
@jpfr Is the idea definitely rejected?

PR #4587, opened 2021-08-09
Update ua_services_discovery.c

jackybek:
In this function UA_Server_addPeriodicServerRegisterCallback(), line 624 checks is incorrect:

if (client->connection.state != UA_CONNECTIONSTATE_CLOSED)

It should be : if (client->connection.state != UA_CONNECTIONSTATE_ESTABLISHED)

### DIFF
--- a/src/server/ua_services_discovery.c
+++ b/src/server/ua_services_discovery.c
@@ -621,7 +621,7 @@ UA_Server_addPeriodicServerRegisterCallback(UA_Server *server,
     }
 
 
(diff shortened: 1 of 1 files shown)


jackybek:
Hi,
I tested using Matrikon OPCUA explorer and confirmed that the proposed changes are required.

jackybek (commented):
I tested using Matrikon OPCUA Explorer and confirmed that the required changes are needed.  Otherwise Matrikon cannot discover the OPC LDS instance.

PR #4592, opened 2021-08-10
Fix (core): Memory leak during decoding data using string-based NodeIds

wbitesi:


jpfr:
This probably assumed That the NodeId, in it was found among the builtin types, is numeric.

yes, this assumption is not always true for custom types...

wbitesi:
Yes, I found that bug using custom types from a beckhoff OPC UA server. It uses string-based node ids for its custom data types.

jpfr:
All green. Merged.

PR #4593, opened 2021-08-11
fix(nc): Sort references before adding to list of nodes

peschuster:
`Node.references` is of type `set()` which is an unordered collection.
This results in a different order of elements on every run of nodeset compiler although the input did not change.

This change fixes this behaviour and produces identical (i.e. deterministic) output from the same input.


jpfr:
Nice!

PR #4603, opened 2021-08-19
[Review] Permit CreateMonitoredItem service for hidden nodes/statuscode BadNodeIdUnknown

mlang-de:
Part 4, 5.12.2 CreateMonitoredItems: Monitored Nodes can be removed from the AddressSpace after the creation of a MonitoredItem. This does not affect the validity of the MonitoredItem but a Bad_NodeIdUnknown shall be returned in the Publish response. It is possible that the MonitoredItem becomes valid again if the Node is added again to the AddressSpace and the MonitoredItem still exists. 

OPCF Forum:
https://opcfoundation.org/forum/opc-ua-standard/create-monitoreditem-with-node-statuscode-badnodeidunknown/

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
  *    Copyright 2014-2018 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
  *    Copyright 2016-2017 (c) Florian Palm
@@ -95,7 +95,7 @@ Service_SetTriggering(UA_Server *server, UA_Session *session,
         response->responseHeader.serviceResult = UA_STATUSCODE_BADSUBSCRIPTIONIDINVALID;
         return;
     }
-    
+
     /* Get the MonitoredItem */
     UA_MonitoredItem *mon = UA_Subscription_getMonitoredItem(sub, request->triggeringItemId);
     if(!mon) {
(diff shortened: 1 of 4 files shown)


jpfr:
Nice!

The spec sounds as if the node has to be present when the MonitoredItem is created. But could be removed later.
Right?

mlang-de:
Yes I agree with you. But to be sure I asked in the OPCF forum.

> If the CreateMonitoredItem status is bad it means the item was not created.
You can return Good in the CreateMonitoredItem status but Bad_NodeIdUnknown in the first reported change.
This tells the client that the not does not exist now but might exist in the future.

So it is also possible if the Client never see this node before. But is there a need for a check on Server application level?

jpfr:
It seems there are good arguments for both sides.
Let’s make this configurable separately for the initial creation of the MonitoredItem and later at runtime.

Could you add a UA_RuleHandling option for both to the server config?

mlang-de:
I think we need this UA_RuleHandling and additional something like a user callback in which the application can decide whether this is a NodeId which really may exists in future. With the current implementation the CTT fails since for instance some configured invalid also will return service result code == Good if the UA_RuleHandling == ACCEPT all time.

mlang-de:
I'm not really happy with this solution. But may be a additional function could be nicer.

jpfr:
- Switched the base to target the 1.3 branch.
- Added to the 1.3 milestone list

The callback looks a lot like "access control".
A callback where we ask in the userland before creating a MonitoredItem.

I would actually suggest that we move the callback to the AccessControl plugin.
I can think of cases where monitoring is not allowed everywhere.

Also, please add the AttributeId to the argument list for the callback.

mlang-de:
I update this PR until end of next week.

src/server/ua_services_monitoreditem.c
```diff
@@ -215,8 +215,11 @@ checkAdjustMonitoredItemParams(UA_Server *server, UA_Session *session,
                 params->samplingInterval = vn->minimumSamplingInterval;
             UA_NODESTORE_RELEASE(server, node);
         }
+        else{
+          params->samplingInterval = server->config.samplingIntervalLimits.min;
```
mlang-de, line 219: @jpfr This is maybe not correct! Which plugin should be requested in this case? NODESTORE_GET_EX?

PR #4607, opened 2021-08-21
[REVIEW] Hofix: when BPF defined error for undefined variable

JanSurft:
### BUG-Fix for:
    
Open62541 doesn’t compile when BPF headers are available on the system.

The root cause is this “message” variable which no longer exists:

https://github.com/open62541/open62541/blob/master/plugins/ua_pubsub_ethernet.c#L1155

The commit which broke it:

https://github.com/open62541/open62541/commit/5d7b509d26939065b6e8768c647a53366904b5f9#diff-b3a77898cbe9ab570871aaa4393ddeeb80891c56035e2b8c0a45d6f5dfed136aL1115 

### TODO

This fix needs to be tested and compiled against a system with BPF headers available.

### DIFF
--- a/plugins/ua_pubsub_ethernet.c
+++ b/plugins/ua_pubsub_ethernet.c
@@ -1150,13 +1150,6 @@ UA_PubSubChannelEthernet_receive(UA_PubSubChannel *channel,
         }
     }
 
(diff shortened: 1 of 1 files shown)


jpfr:
Cool

PR #4608, opened 2021-08-23
feat(server) Event-Filter Batch 1 (Examples, Static-Filter-Validation)

andreasebner:
This PR is the first batch of the event filter server logic. 

This Batch currently contains:

- New pair of examples for the event-filter mechanism
- Static select clause  validation

The PR is based on the work in #4225.

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -14,6 +14,9 @@
  *    Copyright 2017-2018 (c) Thomas Stalder, Blue Time Concept SA
  *    Copyright 2018 (c) Fabian Arndt, Root-Core
  *    Copyright 2020 (c) Kalycito Infotech Private Limited
+ *    Copyright 2021 (c) Uranik, Berisha
+ *    Copyright 2021 (c) Ammar, Morshed
+ *    Copyright 2021 (c) Fraunhofer IOSB (Author: Andreas Ebner)
  */
 
 #include "ua_server_internal.h"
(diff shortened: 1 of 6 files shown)


src/server/ua_services_monitoreditem.c
```diff
@@ -240,6 +243,45 @@ checkAdjustMonitoredItemParams(UA_Server *server, UA_Session *session,
...
+checkEventFilterParam(UA_Server *server, UA_Session *session,
+                      const UA_MonitoredItem *mon,
+                      UA_MonitoringParameters *params){
+    if(mon->itemToMonitor.attributeId == UA_ATTRIBUTEID_EVENTNOTIFIER) {
+        UA_EventFilter *eventFilter =
+            (UA_EventFilter *) params->filter.content.decoded.data;
+        //check the where clause for logical consistency
+        if(eventFilter->whereClause.elementsSize != 0) {
+            UA_ContentFilterResult contentFilterResult;
+                UA_Event_staticWhereClauseValidation(server, &eventFilter->whereClause,
```
jpfr, line 256: Dedent

src/server/ua_services_monitoreditem.c
```diff
@@ -240,6 +243,45 @@ checkAdjustMonitoredItemParams(UA_Server *server, UA_Session *session,
...
+                    //we can use the detailed contentFilterResult later
+                    UA_StatusCode whereResult =
+                        contentFilterResult.elementResults[i].statusCode;
+                    UA_ContentFilterResult_clear(&contentFilterResult);
+                    return whereResult;
+                }
+            }
+        }
+        //check the select clause for logical consistency
+        UA_STACKARRAY(UA_StatusCode, selectClauseValidationResult,
```
jpfr, line 270: This VLA can be abused to overload the stack.
Think a message with a million filter elements. That’s only a few MByte.
Either
- add a length limit (and use a fixed size temp array which can be more optimized by the compiler)
- Run the checks one-by-one and break upon the first problem. So no array at all.

src/server/ua_subscription_events.c
```diff
@@ -678,4 +678,216 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId,
     return retval;
 }
 
+void
+UA_Event_staticSelectClauseValidation(UA_Server *server,
+                                      const UA_EventFilter *eventFilter,
+                                      UA_StatusCode *result) {
+    /* The selectClause only has to be checked, if the size is not zero */
+    if(eventFilter->selectClausesSize != 0) {
```
jpfr, line 686: Return early and dedent.
Deep nesting is an anti-pattern.

src/server/ua_subscription_events.c
```diff
@@ -678,4 +678,216 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId,
...
+                    continue;
+                }
+            }
+        }
+    }
+}
+
+UA_StatusCode
+UA_Event_staticWhereClauseValidation(UA_Server *server,
+                                     const UA_ContentFilter *contentFilter,
```
jpfr, line 746: Rename the arguments simply “filter” and “result”.
That makes all lines shorter and nicer to read.

src/server/ua_subscription_events.c
```diff
@@ -678,4 +678,216 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId,
...
+    if(contentFilterResult->elementResultsSize == 0)
+        return UA_STATUSCODE_GOOD;
+    contentFilterResult->elementResults =
+        (UA_ContentFilterElementResult *)UA_Array_new(
+            contentFilterResult->elementResultsSize,
+            &UA_TYPES[UA_TYPES_CONTENTFILTERELEMENTRESULT]);
+    if(!contentFilterResult->elementResults)
+        return UA_STATUSCODE_BADOUTOFMEMORY;
+    for(size_t i = 0; i < contentFilterResult->elementResultsSize; ++i) {
+        UA_ContentFilterElementResult_init(&contentFilterResult->elementResults[i]);
```
jpfr, line 759: You have to write “ contentFilterResult->elementResults[i]” a million times.
Just create a pointer to it at the beginning of the loop and call it something short.
For example just “er” for element result.

src/server/ua_subscription_events.c
```diff
@@ -678,4 +678,216 @@ UA_Server_triggerEvent(UA_Server *server, const UA_NodeId eventNodeId,
     return retval;
 }
 
+void
```
jpfr, line 681: Please add a comment that lists the validation steps already implemented.
So that humans can read it. ;-)
That makes future work for “completing” the validation much easier.
Right now It would take a lot of effort to figure out which validation steps might be interesting but still missing.

jpfr (commented):
Nice! Looks good.
Some inline comments below.

PR #4610, opened 2021-08-23
[Review] MonitoredItem sampling callback handle Bad_StatusCodes improper

mlang-de:
If the Node are become hidden during runtime the sampling callback did not report Bad_NodeIdKnown. Instead the callback evaluate a data change continuously.

I'm not sure what the line 90 really checks, but I must extend it.
The UA_ByteString_equal signale a datachange if mon->lastSampledValue is a NULL-ByteString all time.
```c
if(!value->value.type) {
    *changed = UA_ByteString_equal(encoding, &mon->lastSampledValue);
    return UA_STATUSCODE_GOOD;
}
```

### DIFF
--- a/src/server/ua_subscription_datachange.c
+++ b/src/server/ua_subscription_datachange.c
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
  *    Copyright 2017-2020 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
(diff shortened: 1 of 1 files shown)


jpfr:
Good find.

I think the if-clause you edited can be entirely removed.
Changes to the statuscode are picked up later when the DataValue is encoded and the encodings are compared.
So that should lead to the same change-detection for the statuscode without special-casing it.

jpfr:
Looks good. Just two last things for completeness.
- Check for `mon->lastValue.hasStatus` as well
- Check for `if(value->value.type != NULL && UA_DataType_isNumeric(value->value.type)` in the if-clause below.

mlang-de:
@jpfr Should be the part the if clasue "mon->parameters.filter.content.decoded.type == &UA_TYPES[UA_TYPES_DATACHANGEFILTER])" also deleted? Imho not.

jpfr:
Nice!

PR #4620, opened 2021-08-31
feat(core): add attach condition behavior function

sqadasi:
In case of custom nodes, which are entirely managed by the application, the function "UA_Server_attachConditionBehavior" can be used to have the alarms and conditions behavior attached to a normal object node of ConditionType.
An example for a usecase could be wanting to create all children nodes manually to manage the memory and node ids and still being able to have alarms and conditions functionality.


jpfr:
> In case of custom nodes, which are entirely managed by the application,

What is meant by that? All nodes are managed by a Nodestore.
Do you mean a custom Nodestore that cannot persist the event listening information?

PR #4627, opened 2021-09-06
refactor(pubsub): improve error messages

jackybek:
change the error messages to reflect the correct meaning.


jpfr:
Merged. Thanks.

Pleas use the correct format for commit titles in the future.

PR #4628, opened 2021-09-06
refactor(pubsub): handle specific case: SSL_ERROR_SSL

jackybek:


andreasebner:
Hi jackybek,

your change looks nice and can be merged. Since you're quite active, please use our conventional commit scheme by our own. Details are there: "open62541/CONTRIBUTING.md". Please update your PR and i look forward to merge it.

jpfr:
Hmm.
The added case previously fell into the default case.
Where the exact same status is returned.
So the change actually changes nothing.

Okay to close @jackybek ?

jackybek:
Ok.  Please close￼

PR #4479, opened 2021-06-18
Look for HasInterface references in the type hierarchy

basyskom-jvoe:
This commit adds support for interfaces that are not directly
attached to the object type being instantiated.

fix(core): HasInterface reference on base type (#4363)

PR #4632, opened 2021-09-08
Fix HasInterface handling

basyskom-jvoe:
Sub nodes of ObjectType nodes and instances were missing the HasInterface reference.

With this patch, all necessary information should be there.
For verification purposes, I have used the XML file from https://github.com/open62541/open62541/pull/4479

What I would expect:
- TopLevelObjectType should have a child of type MyTestObjectType named ObjectWithInterfaces
- ObjectWithInterfaces in the ObjectType node should have two HasInterface references
- ObjectWithInterfaces in the ObjectType node should have three properties (one own property, one from each of the two interfaces. I assume this is correct according to https://reference.opcfoundation.org/v104/Core/docs/Amendment7/4.9.2/ Figure 7A)
- An instance of TopLevelObjectType should look the same

### DIFF
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -548,6 +548,61 @@ Operation_addReference(UA_Server *server, UA_Session *session, void *context,
                        const UA_AddReferencesItem *item, UA_StatusCode *retval);
 
 static UA_StatusCode
+addRef(UA_Server *server, UA_Session *session, const UA_NodeId *nodeId,
+       const UA_NodeId *referenceTypeId, const UA_NodeId *parentNodeId,
+       UA_Boolean forward) {
+    UA_AddReferencesItem ref_item;
+    UA_AddReferencesItem_init(&ref_item);
+    ref_item.sourceNodeId = *nodeId;
+    ref_item.referenceTypeId = *referenceTypeId;
+    ref_item.isForward = forward;
+    ref_item.targetNodeId.nodeId = *parentNodeId;
+
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    Operation_addReference(server, session, NULL, &ref_item, &retval);
+    return retval;
+}
+
+static UA_StatusCode
+addInterfaceChildren(UA_Server *server, UA_Session *session,
+                const UA_NodeHead *head, const UA_Node *type) {
+    /* Get the hierarchy of the type and all its supertypes */
+    UA_NodeId *hierarchy = NULL;
+    size_t hierarchySize = 0;
+    UA_StatusCode retval = getAllInterfaceChildNodeIds(server, &head->nodeId, &type->head.nodeId,
+                                                              &hierarchy, &hierarchySize);
+    if(retval != UA_STATUSCODE_GOOD)
+        return retval;
+    UA_assert(hierarchySize < 1000);
+
+    /* Copy members of the type and supertypes (and instantiate them) */
+    for(size_t i = 0; i < hierarchySize; ++i) {
+        retval = copyAllChildren(server, session, &hierarchy[i], &head->nodeId);
+        if(retval != UA_STATUSCODE_GOOD) {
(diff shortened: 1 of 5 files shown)


src/server/ua_services_nodemanagement.c
```diff
@@ -26,6 +26,10 @@ setMethodNode_callback(UA_Server *server,
                        const UA_NodeId methodNodeId,
                        UA_MethodCallback methodCallback);
 
+static UA_StatusCode
```
jpfr, line 29: Move the function here instead of adding a forward declaration.

src/server/ua_services_nodemanagement.c
```diff
@@ -665,6 +669,26 @@ copyChild(UA_Server *server, UA_Session *session,
             return retval;
         }
 
+        /* Add HasInterface references to the child */
+        if (rd->nodeClass == UA_NODECLASS_OBJECT && !UA_NodeId_isNull(&rd->typeDefinition.nodeId)) {
+            UA_Node *typeNode = NULL;
+            retval = UA_NODESTORE_GETCOPY(server, &rd->typeDefinition.nodeId, &typeNode);
```
jpfr, line 675: It would be better to not get a copy, but a const reference to the original node.

src/server/ua_services_nodemanagement.c
```diff
@@ -754,6 +778,33 @@ addInterfaceChildren(UA_Server *server, UA_Session *session,
             break;
     }
 
+    if (retval == UA_STATUSCODE_GOOD) {
```
jpfr, line 781: Remove the space after if and for for consistency.

src/server/ua_services_nodemanagement.c
```diff
@@ -754,6 +778,33 @@ addInterfaceChildren(UA_Server *server, UA_Session *session,
...
+        for (size_t i = 0; i < hierarchySize; ++i) {
```
jpfr, line 782: Early quit instead of nesting blocks.

src/server/ua_services_nodemanagement.c
```diff
@@ -754,6 +778,33 @@ addInterfaceChildren(UA_Server *server, UA_Session *session,
...
+            UA_ExpandedNodeId targetId;
+            UA_ExpandedNodeId_init(&targetId);
+            targetId.nodeId = hierarchy[i];
+
+            /*
+             * This is bad, some other thread could sneak in here.
+             * Unfortunately, there is no other choice due to the lock assert.
+             */
+            UA_UNLOCK(&server->serviceMutex);
+            retval = UA_Server_addReference(server, head->nodeId,
```
jpfr, line 792: There is an internal version that does not take locks.
You might have to expose it in ua_server_internal.h.

jpfr (commented):
Looks good !
A few comments…

basyskom-jvoe:
@jpfr Many thanks for the review.
I can't seem to find the internal version of addReference, any hints in which file I can find it?

jpfr:
> The dangerous unlock-lock-cycle is done two layers deeper in Operation_addReference().

At this time, the "big server lock" must not be taken twice.
Therefore we give up the lock temporarily when a callback to user-defined code is called.
That is only dangerous in case that callback does something dumb.
Like deleting the server from which the callback came...

Is this ready for another review?

It would be great to have it covered with a unit test as well.
That prevents future breakage.

basyskom-jvoe:
From how I understand the problem, it's not just callbacks but any open62541 API function which is called from user code and could acquires the lock and change some internal state before the intended internal function can get the lock. This would be fatal in case such a thing happens while some function is iterating over a list and the intruder changes the list.

The code is ready for a second round of review.
I will be on vacation for the next few days. After I am back, I could look into adding unit tests.

basyskom-jvoe:
@jpfr I have extended the PR with a test.

jpfr:
Nice! Merged.

basyskom-jvoe:
@jpfr Thanks for the merge!
Would you see this as a bugfix which should also go to 1.2?

PR #4643, opened 2021-09-13
Fixed wrong calculation of UA_DateTime_localTimeUtcOffset when UA_ARC…

skuep2:
Local time offset in function ``UA_DateTime_localTimeUtcOffset`` is returned as 0, on a FreeRTOS+LWIP architecture, when enabling POSIX compatible time functions (``UA_ARCHITECTURE_FREERTOSLWIP_POSIX_CLOCK``).

I copied the ``UA_DateTime_localTimeUtcOffset`` function from ``arch/posix/ua_clock.c`` to retrieve the actual timezone/UTC offset information.

### DIFF
--- a/arch/freertosLWIP/ua_clock.c
+++ b/arch/freertosLWIP/ua_clock.c
@@ -4,6 +4,7 @@
  *    Copyright 2016-2017 (c) Julius Pfrommer, Fraunhofer IOSB
  *    Copyright 2017 (c) Stefan Profanter, fortiss GmbH
  *    Copyright 2017 (c) Thomas Stalder
+ *    Copyright 2021 (c) Simon Kueppers, 2pi-Labs GmbH
  */
 
 #ifdef UA_ARCHITECTURE_FREERTOSLWIP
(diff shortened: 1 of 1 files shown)


jpfr:
Nice!

goetzj:
With this commit, I am not able to build the library because python is missing. Maybe its not because of this commit. With the last commit (b862f28) of version 1.2 I succeed. Ubuntu V18.04 amd64.
cmake_python3_err

skuep2:
I have the same problem, but it is not related to my commit but rather commit 4d4e206ef58875aedbc39af4274344b849b5a48b.
The problem is that cmake needs to be at least version 3.12 according to https://cmake.org/cmake/help/latest/module/FindPython3.html. However I am running A Ubuntu 18.04 based OS, which ships cmake 3.10.2.
So far the only solutions I can come up with are: 
- keep using the 1.2 branch or 
- use a PPA with a more recent cmake version (e.g. https://apt.kitware.com/)

PR #4647, opened 2021-09-15
Nodeset resolver Tool (for generating custom tailored namespace-zero nodesets)

skuep2:
We are using the open62541 stack in an embedded environment, where RAM is scarce. When implementing a PA-DIM nodeset, I noticed that there are a lot of dependencies, that can thus far only be satisfied by including the full namespace-zero. 
However, because this is a heavy burden on the on-chip memory I was looking for a way to produce a custom tailored namespace-zero that included only the hard dependencies.

I could not find a way to programatically produce such a custom XML file, which is why I created the ``nodeset_resolver`` tool. It heavily reuses the nodeset libraries used by the exquisite ``nodeset_compiler``, which is why I put it into the same folder. Although this is a bit messy, this may be something a folder rename might fix in the future.

It works by loading the given XML files into a nodeset representation and then searches the tree for missing nodeIds. Then, the tool can look up those nodeIds from a reference XML, that is loaded into a separate nodeset representation. All (child-) dependencies are gathered and output. Lastly, it can automagically create an XML nodeset file for the required nodes from the reference XML or even splice it into the existing XML.

What do you think of this idea? It was a bit fiddly to get it working to really get all dependencies from the supplied nodeset files (especially if you don't know a lot about OPC UA yet) but the examples I have written into the README file work very well for our embedded setup. Comments are welcome!

EDIT: Should I include an affiliation in the README (similar to the one for the ``nodeset_compiler``) in case you are interested in this PR?
EDIT2: Also a license should be included in the Python script, right?

### DIFF
--- a/tools/nodeset_compiler/README.md
+++ b/tools/nodeset_compiler/README.md
@@ -43,10 +43,10 @@ Due to memory constraints, you are using the default reduced OPC UA Namespace-Ze
 ````
 nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
   -x .../deps/ua-nodeset/PADIM/Opc.Ua.IRDI.NodeSet2.xml \
-  -x .../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
+  -x .../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml -u
 ````
 
-The output shows a list of missing NodeIds, that can not be resolved from the supplied XML files. Ignoring ``ns=0``, the other missing nodeIds can be supplied by including the OPC UA DI model.
+The output shows a list of missing NodeIds, that can not be resolved from the supplied XML files. Ignoring ``nsu=http://opcfoundation.org/UA/``, the other missing nodeIds can be supplied by including the OPC UA DI model (``nsu=http://opcfoundation.org/UA/DI/``).
 ````
 nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
   -x .../deps/ua-nodeset/DI/Opc.Ua.Di.NodeSet2.xml \
@@ -54,7 +54,7 @@ nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
   -x .../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
 ````
 
-This time, only NodeIds with ``ns=0`` are listed as unresolvable, because they are missing from the default reduced namespace-zero nodeset.
+This time, only NodeIds of namespace-zero are listed as unresolvable, because they are missing from the default reduced namespace-zero nodeset.
 We can now include the full OPC UA namespace-zero nodeset as a reference file to resolve all missing dependencies.
 
 ````
--- a/tools/nodeset_compiler/README.md
+++ b/tools/nodeset_compiler/README.md
@@ -35,7 +35,7 @@ optional arguments:
 
 ## Showing missing dependencies
 Suppose you want to implement a OPC UA PA-DIM nodeset on an embedded platform.
-Due to memory constraints, you are using the default reduced OPC UA Namespace-Zero nodeset.
+Due to memory constraints, you are using the default reduced OPC UA Namespace-Zero nodeset. However you quickly find yourself in a position, where a lot of additional dependencies need to be satisfied.
 
 ````
 nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
@@ -43,7 +43,7 @@ nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
   -x .../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
 ````
 
-The output shows a list of missing NodeIds, that can not be resolved from the supplied XML files. Ignoring ``ns=0``, the missing NodeIds can be solved by including the OPC UA DI model.
+The output shows a list of missing NodeIds, that can not be resolved from the supplied XML files. Ignoring ``ns=0``, the other missing nodeIds can be supplied by including the OPC UA DI model.
 ````
 nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
   -x .../deps/ua-nodeset/DI/Opc.Ua.Di.NodeSet2.xml \
@@ -51,7 +51,7 @@ nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
   -x .../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
 ````
 
-This time, only NodeIds with ``ns=0`` are list as unresolvable, because they are missing from the default reduced namespace-zero nodeset.
+This time, only NodeIds with ``ns=0`` are listed as unresolvable, because they are missing from the default reduced namespace-zero nodeset.
 We can now include the full OPC UA namespace-zero nodeset as a reference file to resolve all missing dependencies.
 
 ````
@@ -63,10 +63,11 @@ nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
 ````
 
 The output now shows all resolved NodeIds, that are required for namespace zero.
+Now the next step includes generating a new namespace-zero that includes the missing nodes.
 
-## Generating Custom NodeSets
-You can now automatically generate an XML file with these resolved NodeIds by pulling in their definition from the reference XML file.
-Additionally, you can automatically merge the resolved node definitions with the existing XML nodeset given with the ``-e`` parameter.
+## Generating a Custom nodeset
+You can now automatically generate an XML file with the resolved nodes by specifying that their definition should be pulled-in from the reference XML file by specifying ``-p``.
+Additionally, you can automatically merge the resolved node definitions with the existing XML nodeset (given with the ``-e`` parameter) by specifying ``-m``.
 
 Using this feature, you can create a custom tailored namespace-zero XML file using the default reduced nodeset as a basis.
 
@@ -79,7 +80,17 @@ nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
   -p -m > Opc.Ua.NodeSet2.Custom.xml
 ````
 
-_Note_: In this scenario, the nodesets given with the ``-x`` parameter will later be used with the ``nodeset_compiler`` to produce the compiled code for implementing the OPC UA PA-DIM nodeset.
+## Using the custom nodeset
+The custom generatead ``Opc.Ua.NodeSet2.Custom.xml`` can now be included as the namespace-zero file during compilation. For example:
+
+````
+cmake -DUA_ARCHITECTURE=freertosLWIP \
+      -DUA_ENABLE_AMALGAMATION=ON \
+      -DUA_NAMESPACE_ZERO=FULL \
+      -DUA_FILE_NS0='path/to/generated/Opc.Ua.NodeSet2.Custom.xml' ../
+````
+
+_Note_: In this scenario, the nodesets given with the ``-x`` parameter will later be used with the ``nodeset_compiler`` to produce the compiled code for implementing the OPC UA PA-DIM nodeset. For example:
 
 ````
 nodeset_compiler.py -e path/to/generated/Opc.Ua.NodeSet2.Custom.xml \
--- a/tools/nodeset_compiler/README.md
+++ b/tools/nodeset_compiler/README.md
@@ -4,9 +4,9 @@ The nodeset compiler is a collection of Python scripts that can parse OPC UA XML
 
 The initial implementation has been contributed by a research project of the chair for Process Control Systems Engineering of the TU Dresden. It was not strictly speaking created as a C generator, but could be easily modified to fulfill this role for open62541. Later on it was extended and improved by the core developers of open62541.
 
-# open62541 nodeset resolver
+# open62541 nodeset Resolver
 
-The nodeset resolver can be used to identify dependencies between nodesets. The main purpose of this tool is to generate custom stripped-down namespace-zero nodesets.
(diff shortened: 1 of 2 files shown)


GoetzGoerisch:
Nice functionality!

But to my knowledge a compliant UA 1.04 server must provide all nodes defined by Part 3 -> NS0

[Reference](https://reference.opcfoundation.org/Core/docs/Part7/5.4/#Table14)  _ConformanceUnit_ Address Space Base

This requirement is going away with UA 1.05. (See https://youtu.be/P-FQOhvBOBU?t=688)

skuep2:
Thanks for your reply! 
I am in no way an expert on OPC UA, however if I understand correctly, the default reduced namespace that comes with open62541 passes the CTT (I assume that is the conformance test?). 
See https://open62541.org/doc/1.2/building.html#build-options under **UA_NAMESPACE_ZERO**

The ``nodeset_resolver``, when used with ``--merge`` just expands the reduced (or any) nodeset with the missing dependencies. I.e. the default reduced nodeset will always be a subset of the final (custom tailored) nodeset.
I guess this was not entirely clear in my first post.

jpfr:
Wow. This looks amazing!

- The examples uses only one `-r` reference file. Can several reference files be used as well?
- From a developer perspective it might be easier to have a whitelist of NodeIds (e.g. a file with one NodeId per line) instead of a partial/invalid nodeset xml to be filled up. That should be at least on option. Note that we might have to use `ExpandedNodeId` because the namespace identifier is rewritten inside the nodeset compiler when several companion specs are loaded.
- Yes, the license should be the MPv2 similar to the rest of the NodeSet compiler.

Looking forward to this!

skuep2:
I am glad you find this tool useful! Thanks for your input.

>     * The examples uses only one `-r` reference file. Can several reference files be used as well?
In the beginning I had support for multiple reference files (to be loaded into the ``referenceNodeSet``). However I ran into issues with the ``--pull`` functionality, where the missing nodes are pulled from the reference XML. This functionality would then be required to work with multiple reference files, which is more than we actually need for our purposes.

>     * From a developer perspective it might be easier to have a whitelist of NodeIds (e.g. a file with one NodeId per line) instead of a partial/invalid nodeset xml to be filled up. That should be at least on option.
I changed the output format of node ids (when not using ``--pull``) to be one node id per line. I think that is what you mean with (generating?) a whitelist of NodeIds. (Or did you refer to _reading_ NodeIds from a whitelist file?) E.g.:
````
./nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
    -x ../../deps/ua-nodeset/DI/Opc.Ua.Di.NodeSet2.xml \
    -x ../../deps/ua-nodeset/PADIM/Opc.Ua.IRDI.NodeSet2.xml \
    -x ../../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml \
    -r ../../deps/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml
INFO:__main__:Preprocessing (existing) ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml
INFO:__main__:Preprocessing ../../deps/ua-nodeset/DI/Opc.Ua.Di.NodeSet2.xml
INFO:__main__:Preprocessing ../../deps/ua-nodeset/PADIM/Opc.Ua.IRDI.NodeSet2.xml
INFO:__main__:Preprocessing ../../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
INFO:__main__:Collecting missing nodes...
INFO:__main__:Collected 28 missing nodes out of 619 used nodes
INFO:__main__:Preprocessing (reference) ../../deps/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml
INFO:__main__:Resolving all dependencies from ../../deps/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml...
INFO:__main__:Resolved 465 required nodes out of 1060 dependent nodes (0 unresolved)
ns=0;i=11616
ns=0;i=11715
ns=0;i=96
ns=0;i=95
ns=0;i=291
ns=0;i=256
ns=0;i=11508
ns=0;i=11510
ns=0;i=17603
ns=0;i=18347
ns=0;i=2771
ns=0;i=2307
ns=0;i=2311
ns=0;i=2310
ns=0;i=2309
[...]
````
> Note that we might have to use `ExpandedNodeId` because the namespace identifier is rewritten inside the nodeset compiler when several companion specs are loaded.

I needed to think about that for a while.. You are totally right. So the reason the nodeset_resolver tool current works _at all_ is, because I am only using nodeset-zero right? Or in other words, the output of the following command reports missing namespace Ids that do not correspond to actual namespace Ids in the OPC UA Specification (due to the rewriting you mention).
````
./nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
    -x ../../deps/ua-nodeset/PADIM/Opc.Ua.IRDI.NodeSet2.xml \
    -x ../../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
INFO:__main__:Preprocessing (existing) ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml
INFO:__main__:Preprocessing ../../deps/ua-nodeset/PADIM/Opc.Ua.IRDI.NodeSet2.xml
INFO:__main__:Preprocessing ../../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
INFO:__main__:Collecting missing nodes...
INFO:__main__:Collected 20 missing nodes out of 251 used nodes
ns=0;i=11715
ns=0;i=11616
ns=0;i=256
ns=0;i=291
ns=0;i=17598
ns=0;i=17594
ns=0;i=17603
ns=2;i=15051
ns=0;i=17590
ns=0;i=17597
ns=2;i=15063
ns=2;i=6244
ns=2;i=1004
ns=0;i=17570
ns=0;i=19084
ns=0;i=15318
ns=0;i=2373
ns=0;i=19077
ns=0;i=2372
ns=0;i=11508
````
NodeIds from ``ns=2`` are reported that actually belong to Opc.Ua.Di.NodeSet2.xml that has ``ns=1`` in the specification.
Do you have an example how ``ExpandedNodeId`` can be used to solve this problem?

**Comment by skuep2** [comment]
````
# removed -x Opc.Ua.Di.NodeSet2.xml to show expanded output
./nodeset_resolver.py -e ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml \
    -x ../../deps/ua-nodeset/PADIM/Opc.Ua.IRDI.NodeSet2.xml \
    -x ../../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml --expanded
INFO:__main__:Preprocessing (existing) ../../tools/schema/Opc.Ua.NodeSet2.Reduced.xml
INFO:__main__:Preprocessing ../../deps/ua-nodeset/PADIM/Opc.Ua.IRDI.NodeSet2.xml
INFO:__main__:Preprocessing ../../deps/ua-nodeset/PADIM/Opc.Ua.PADIM.NodeSet2.xml
INFO:__main__:Collecting missing nodes...
INFO:__main__:Collected 20 missing nodes out of 251 used nodes
nsu=http://opcfoundation.org/UA/;i=11715
nsu=http://opcfoundation.org/UA/;i=11616
nsu=http://opcfoundation.org/UA/;i=256
nsu=http://opcfoundation.org/UA/;i=291
nsu=http://opcfoundation.org/UA/;i=17594
nsu=http://opcfoundation.org/UA/;i=17598
nsu=http://opcfoundation.org/UA/;i=17597
nsu=http://opcfoundation.org/UA/DI/;i=15063
nsu=http://opcfoundation.org/UA/;i=17590
nsu=http://opcfoundation.org/UA/;i=17603
nsu=http://opcfoundation.org/UA/DI/;i=15051
nsu=http://opcfoundation.org/UA/DI/;i=6244
nsu=http://opcfoundation.org/UA/DI/;i=1004
nsu=http://opcfoundation.org/UA/;i=17570
nsu=http://opcfoundation.org/UA/;i=19084
nsu=http://opcfoundation.org/UA/;i=15318
nsu=http://opcfoundation.org/UA/;i=2373
nsu=http://opcfoundation.org/UA/;i=19077
nsu=http://opcfoundation.org/UA/;i=2372
nsu=http://opcfoundation.org/UA/;i=11508
````

**Comment by GoetzGoerisch** [comment]
> I am in no way an expert on OPC UA, however if I understand correctly, the default reduced namespace that comes with open62541 passes the CTT (I assume that is the conformance test?).
> See https://open62541.org/doc/1.2/building.html#build-options under **UA_NAMESPACE_ZERO**

From the certified server: https://github.com/open62541/open62541/blob/1135945fe8eb8e4e9df7694e5d564443a6d4df80/CMakeLists.txt#L152-L154

Therefore for UA 1.04 compliance a server product need to have a full NS0 for ["Micro Embedded Server Profile"](http://opcfoundation.org/UA-Profile/Server/MicroEmbeddedDevice2017) and above. 

As mentioned earlier, this is a great feature for the upcoming UA 1.05 release, which is currently in review.

As Julius said, looking forward to this feature.

**Comment by skuep2** [comment]
Thanks for the infos! I guess you are right, although the "problem" with non-conformity is not necessarily introduced by the ``nodeset_resolver``, but by using the reduced namespace-zero in the first place (which is the default of open62541, if i remember correctly). The ``nodeset_resolver`` just extends the given namespace-zero by the dependencies.

PS: For someone with a heavy embedded background, it is kind of... interesting, that for something called a "micro embedded device profile", requires a full namespace-zero, comprising several megabytes of RAM. This is more RAM, than almost all non-application microcontrollers actually have on-chip.. by a multiple. :-)

**Comment by nogryy** [comment]
Hello,
very good feature! 
With this it was for example possible to integrate the FileType support easily into the reduced namespace ns0. 
Do you have any idea when this feature will be available in a release?
Greetings

**Comment by jpfr** [comment]
Hey there,

we had another look at this. This is really cool.

If we add a "whitelist" argument to the nodeset compiler, then we could use your `walkNodes` as a filter step to only select the nodes that are required for the whitelist.

If we move that internally to the nodeset compiler, that would forgeo the XML import-export intermediate step.
And we could load the full ns0 by default and remove everything that is not needed.

**Comment by skuep2** [comment]
Hey Julius (former Fraunhofer Gesellschaft Colleague ;-)), 
Thanks for your continued interest.
The tool was designed (as a separate tool) with the idea in mind, that I, as a developer _using_ the open62541 library can call the ``nodeset_compiler.py`` _on my own_, to generate my own .c/.h nodeset files, that I will then integrate into my own application.

However, your (and other peoples) comments give me the impression that this is not the way open62541 was designed to operate. It looks like you are using the ``nodeset_compiler.py`` exclusively within the CMake environment, during the namespace-zero generation. And if you want to add other device models, they will be added to this namespace-zero generation (which technically, it is not a namespace-zero anymore in this case). For better understanding, is that what your comment is implying? 
Anyway, I think the idea with a "whitelist" is good, is the term  "whitelist" appropriate for what is happening behind the scenes? 

Our use-case was, that we wanted to use the PADIM device information model, how would I find out, what nodes to whitelist to include _anything_ to do with this particular information model? I think it would require a way of extracting all (?) nodes from the corresponding ``Opc.Ua.PADIM.NodeSet2.xml`` (and ``Opc.Ua.IRDI.NodeSet2.xml``) to give to the compiler. Or would it be enough to just add the PADIM Object node id to the whitelist? For lack of deep insight into all the OPC UA stuff, I sincerely do not know if that would be enough. Or is there any obvious way I have missed?

Also another question I have is, do you suggest to implement ``walkNodes`` within ``nodeset_compiler.py`` to control the namespace-zero generation only or to generate a full nodeset, including PADIM+dependencies+stripped-ns0 as a single monolith integrated into the open26541 library/amalgamate?

I have to admit, I am not that deeply invested into the OPC UA/open62541 development and I am not familiar with certain ways to do or ways, that things were designed to be done (Which is why I never thought about integrating this into the nodeset_compiler). Please excuse my ignorance in some of the points above.

Feel free to use the code! If you need me to do something (sourcecode or organizational stuff) let me know.

**Comment by jpfr** [comment]
Hi @skuep2 ,

the node set compiler can also be used standalone, not only within cmake.
The question is whether your tool is to be used outside of the open62541 ecosystem as well.
Then it makes sense to have XML as an exchange format.

If you just write out the xml to read it back into the node set compiler, then we can reduce the complexity and omit the intermediate xml representation.

The whitelisiting could be done for a list of nodes. Or, since every companion spec has its own namespace (uri), on the level of namespaces.

thanks for giving the go-ahead for working with your code. I think this will become a new feature rather soon.

**Comment by skuep2** [comment]
I think it would be perfectly reasonable to bind the tool to the open62541 ecosystem. The reason I chose XML as the exchange format was, that there was/is no other way to use a customized nodeset-zero than giving ``my-custom-ns0.xml`` to cmake (besides from choosing one of the pre-made ones).

But yeah, during cmake/build this XML will be read again by the nodeset compiler, you are totally right. 
I guess you would want to add some compile options to cmake so you can a) enable tailored-nodeset-zero generation and b) control what nodes are allowed, right?
I like the idea with the namespace (uri)! I did not think of it that way.

Best Regards

**Comment by Schievel1** [comment]
This probably needs a rebase against the newest main. I updated my open62541 submodule and get this: 
```
[  1%] Generating /home/pascal/builds/t3_Y2r2S3/0/jaeger/opcua-mtb4-server/src/generated/Opc.Ua.NodeSet2.CustomNS0.xml
INFO:__main__:Preprocessing (existing) /home/pascal/builds/t3_Y2r2S3/0/jaeger/opcua-mtb4-server/src/ns0reduced/Opc.Ua.NodeSet2.Reduced.xml
INFO:__main__:Preprocessing (existing) /home/pascal/builds/t3_Y2r2S3/0/jaeger/opcua-mtb4-server/open62541/tools/schema/Opc.Ua.NodeSet2.Part8_Subset.xml
INFO:__main__:Preprocessing /home/pascal/builds/t3_Y2r2S3/0/jaeger/opcua-mtb4-server/open62541/deps/ua-nodeset/DI/Opc.Ua.Di.NodeSet2.xml
INFO:__main__:Preprocessing /home/pascal/builds/t3_Y2r2S3/0/jaeger/opcua-mtb4-server/open62541/deps/ua-nodeset/Machinery/Opc.Ua.Machinery.NodeSet2.xml
INFO:__main__:Preprocessing /home/pascal/builds/t3_Y2r2S3/0/jaeger/opcua-mtb4-server/src/generated/mtb4.NodeSet2.xml
  File "/home/pascal/builds/t3_Y2r2S3/0/jaeger/opcua-mtb4-server/tools/nodeset_resolver.py", line 115, in <module>
    if n.sanitize() == False:
       ^^^^^^^^^^
AttributeError: 'ObjectNode' object has no attribute 'sanitize'
```

PR #4648, opened 2021-09-16
feat(PubSub): missing useTLS parameter

jackybek:


jpfr:
Looks good.
Can you confirm that `connect_flags |= MQTT_CONNECT_USETLS;` shall be set for both with and without tls?

jackybek:
For non TLS connection MQTT_CONNECT_USETLS will be false, for TLS
connection MQTT_CONNECT_USETLS will be true

jpfr:
Greta. Merged.

PR #4656, opened 2021-09-19
(feat) Secure MQTT Client connection

jackybek:
line 481 to 487 - suspected cause of error during runtime :  SSL_ERROR_SSL
actual error is triggered by yieldMqtt(), where the mqtt_sync function return 0x80000011.

if i remove these 7 lines, this error disappears.

### DIFF
--- a/plugins/mqtt/ua_mqtt_adapter.c
+++ b/plugins/mqtt/ua_mqtt_adapter.c
@@ -167,7 +167,7 @@ connectMqtt(UA_PubSubChannelDataMQTT* channelData){
 
 #ifdef UA_ENABLE_MQTT_TLS_OPENSSL
         SSL_library_init();
-
+        OpenSSL_add_all_algorithms();
         SSL_load_error_strings();
         ERR_load_crypto_strings();
 
@@ -289,7 +289,8 @@ connectMqtt(UA_PubSubChannelDataMQTT* channelData){
         UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "PubSub MQTT: Connection creation failed. Out of memory (mqtt_client).");
         return UA_STATUSCODE_BADOUTOFMEMORY;
     }
-
+    client->keep_alive = 400;
+    
     /* save reference */
     channelData->mqttClient = client;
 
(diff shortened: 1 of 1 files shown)


jackybek:
What I discover was that the yieldMqtt() function was already called in TransportLyaterMqtt_addChannel().
So if it is called again in connectMqtt() for the mqtt client object, it will trigger errors.

plugins/mqtt/ua_mqtt_adapter.c
```diff
@@ -387,13 +478,13 @@ connectMqtt(UA_PubSubChannelDataMQTT* channelData){
...
     /* sync the first mqtt packets in the buffer to send connection request.
        After that yield must be called frequently to exchange mqtt messages. */
-    UA_StatusCode ret = yieldMqtt(channelData, 100);
-    if(ret != UA_STATUSCODE_GOOD){
-        freeTLS(channelData);
-        UA_free(channelData->connection);
-        UA_free(client);
-        return ret;
-    }
+    //UA_StatusCode ret = yieldMqtt(channelData, 100);
```
jpfr, line 481: Why is this commented out?
If not needed, remove entirely.

plugins/mqtt/ua_mqtt_adapter.c
```diff
@@ -370,11 +371,101 @@ connectMqtt(UA_PubSubChannelDataMQTT* channelData){
...
+
+    UA_Boolean useTLS;
+    useTLS = channelData->mqttUseTLS;
+#endif
+
     /* Connect mqtt with socket fd of networktcp  */
-    mqttErr = mqtt_connect(client, clientId, NULL, NULL, 0, username, password, 0, 400);
+    //mqttErr = mqtt_connect(client, clientId, NULL, NULL, 0, username, password, 0, 400);
+    mqttErr = mqtt_connect(client, clientId, NULL, NULL, 0, username, password,
+                            caFilePath, caPath, clientCertPath, clientKeyPath, useTLS, 0, client->keep_alive);
```
jpfr, line 456: This will fail to build with encryption disabled.
Because you define the variable useTLS behind an ifdef.

jpfr (commented):
Looks good.
Two comments to address.

plugins/mqtt/ua_mqtt_adapter.c
jackybek: removed

plugins/mqtt/ua_mqtt_adapter.c
jackybek: re-organised the code

PR #4661, opened 2021-09-21
LWIP full gethostname implementation

skuep2:
Changed ``gethostname_lwip`` to actually retrieve hostname from LWIP stack, when ``LWIP_NETIF_HOSTNAME`` is defined, for FreeRTOS-LWIP Architecture. 

With this PR, it is no more required to set the server hostname prior to starting the server as described in the docs (at least when LWIP is configured to support hostnames).
Using the corresponding open62541 ``UA_ServerConfig_set_customHostname`` function is deprecated anyway, so this is a much better solution, future wise.


jpfr:
Cool, thanks!

PR #4662, opened 2021-09-22
Fixed compiler warnings (-Wformat) related to printf format specifiers

skuep2:
Wrong format specifiers cause a heap of compiler warnings with -Wformat enabled. I used the correct definitions from ``inttypes.h`` or cast to the correct type (to follow the scheme within the source code file).


jpfr:
Thanks! Adjusted the commit title...

Issue #4497, opened 2021-06-30
Error when building package that uses open62541


joe28965:
## Description
I have written my OPC UA client (opcua_bridge) using open62541. However, when I try and build it (using colcon build) I get the following error:

```
/usr/bin/ld: cannot find -lmbedtls
/usr/bin/ld: cannot find -lmbedx509
/usr/bin/ld: cannot find -lmbedcrypto
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/opcua_bridge.dir/build.make:152: opcua_bridge] Error 1
make[1]: *** [CMakeFiles/Makefile2:78: CMakeFiles/opcua_bridge.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

## Background Information / Reproduction Steps
Operating system: Ubuntu 20.04
Installed using the [debian](https://open62541.org/doc/current/installing.html#debian)

## Fix

I already found how to fix it, I need to run the following:

`sudo apt install libmbedtls-dev`

However, it's not mentioned on the installation page. I also feel like it should be something that's installed automatically when you install open62541.

If you have any other questions, please let me know. I'll be happy to provide more information

jpfr:
Which version of the library did you install?
Where did you download it?

joe28965:
I installed it using the instructions for the debian installation on the [website](https://open62541.org/doc/current/installing.html#debian):

```
sudo add-apt-repository ppa:open62541-team/ppa
sudo apt-get update
sudo apt-get install libopen62541-1-dev
```

Which, if I'm not mistaken, installed version 1.1.5. (at least, when I run `sudo apt-get install libopen62541-1-dev` again, I get `libopen62541-1-dev is already the newest version (20210128T143534~1.1.5~1~ge1650d8a~ubuntu20.04.1).`

jackybek:
Julius - actually the command ‘sudo apt-get install … limbedtls-dev ‘  … can be found in www.kalycito.com/running-opcua-on-raspberry-pi/ as part of the installation steps.  You may want to update the online documents in open62541.org

PR #4669, opened 2021-09-29
fix/debian): add libmbedtls-dev dependency to open62541-dev

jrcoding:
With branch 1.1 there is a dependency on libmbedtls introduced for open62541-dev, but this dependency is missing in the corresponding control file. As result the libmbedtls-dev package won't be automatically installed when installing open62541-dev, as pointed out in #4497. 
This patch adds the related dependency on libmbedtls-dev for open62541-dev and should solve #4497

This patch should be applied to branches 1.1 and master as well.

### DIFF
--- a/debian/control-template
+++ b/debian/control-template
@@ -22,7 +22,7 @@ Package: libopen62541-<soname>-dev
 Section: libdevel
 Architecture: any
 Multi-Arch: same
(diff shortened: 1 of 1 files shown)


jpfr:
Thanks!

PR #4670, opened 2021-09-30
fix/debian): add libmbedtls-dev dependency to open62541-dev

jrcoding:
same as #4669 but for master branch


jpfr:
We will do a merge from 1.2 to master.
This will be included there.

PR #4672, opened 2021-09-30
[REVIEW] fix(server): Allow SecurityPolicyNone without an application certificate

jonasgreen88:
Currently the UA_Server_run_startup() function fails with the status
code UA_STATUSCODE_BADSECURITYCHECKSFAILED if the stack is built with
UA_ENABLE_ENCRYPTION=ON and the server config contains the
SecurityPolicyNone without an application certificate.

This is a problem if you build a product with encryption support but
supports to it run without any secured SecurityPolicies enabled and
no certificates installed.

This fix makes verifyServerApplicationURI() to not verify the
applicationURI of SecurityPolicyNone instances without an application
certificate.


jpfr:
Nice!

PR #4673, opened 2021-10-01
[REVIEW] fix(server/client): Updated log message format strings to solve type missmatch build warnings

jonasgreen88:
The binaryEncodingId variable of the UA_DataType structure has
over time changed from an 16 bit integer to an UA_NodeId type. This
has caused type missmatches between the format string and the
binaryEncodingId variable when it is used in in log messages.


jpfr:
Thanks!

PR #4674, opened 2021-10-01
fix(amalgamation): change include order to fix UA_INLINE missing

kimim:
when amalgamation, the order of include headers causes UA_INLINE missing.
headers include order changed in CMakeLists.txt to fix this issue.


jpfr:
Thanks for looking into this.

We can merge when the build errors in the CI are fixed.

kimim:
this PR fix the problem reported in maillist: https://groups.google.com/g/open62541/c/QyKkkJHd9uo

PR #4677, opened 2021-10-04
Event-Filter Batch 3 - Extended Operators

andreasebner:
This PR is the second batch of the event filter server logic. The changes are based on batch 1 #4608 and #4626 

This Batch currently contains:
- Extended Operand handling during where evaluation
- Extended Operator List (and, or)

The PR is based on the work in #4225


src/server/ua_subscription_events.c
```diff
@@ -234,7 +319,22 @@ UA_Server_evaluateWhereClauseContentFilter(UA_Server *server,
             return UA_STATUSCODE_BADEVENTFILTERINVALID;
         }
         case UA_FILTEROPERATOR_EQUALS:
-        case UA_FILTEROPERATOR_ISNULL:
+        case UA_FILTEROPERATOR_ISNULL: {
```
jpfr, line 322: Looks good!
But please break out the cases into dedicated functions.
Those are much easier to reason about as the number of input arguments is small.
If that function returns a statuscode, we can have only one line per case.

PR #4702, opened 2021-10-15
build(pkgconfig): remove the 'v' prefix from the version string

EmmanuelP:
This prefix does not play nicely with the tools using pkg-config (like meson
for instance).


jpfr:
Ok, cool.
Switched to target to 1.3.

PR #4704, opened 2021-10-15
fix(core): check if output connection is ready before write

EmmanuelP:
Hi,

If an output connection is not ready (buffer filled), it will return immediately
with EAGAIN, and connection_write retries to send data without checking if the
condition has changed. It results of an unwanted high CPU use. Polling with
POLLOUT event fixes this issue.

Only a posix implementation for now.

Please let me know if this work is worthwhile for open62541.

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -61,6 +61,10 @@ connection_write(UA_Connection *connection, UA_ByteString *buf) {
     int flags = 0;
     flags |= MSG_NOSIGNAL;
 
+    struct pollfd poll_fd[1];
+    poll_fd[0].fd = connection->sockfd;
+    poll_fd[0].events = UA_POLLOUT;
+
     /* Send the full buffer. This may require several calls to send */
     size_t nWritten = 0;
     do {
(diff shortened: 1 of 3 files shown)


EmmanuelP:
Hi,

> Most windows builds are failing … Can you look into that?

I'm not sure I'll be able to test or even implement a solution on Windows. Polling on this platform is tricky.

Is a solution where polling is only implemented for the posix platforms acceptable ?

jpfr:
Having the same logic on windows seems doable.
At least the function prototype is the same:
https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsapoll

Can you elaborate on the problems you expect?
I'd prefer to keep the windows port as similar as possible to the regular POSIX port.

EmmanuelP:
> Can you elaborate on the problems you expect?

I'm maintaining a multiplatform vision communication library, and the native Windows port done by a contributor seemed tricky to do correctly. But for some reasons he did not use WSAPoll.

> I'd prefer to keep the windows port as similar as possible to the regular POSIX port.

I have declared the necessary API for Windows. As far as I have understood WSAPoll, and the already defined API wrappers (UA_ERRNO, UA_send...), I think it should work, but I did not perform tests on a Windows platform.

jpfr:
Looks good. This is ready to merge @EmmanuelP ?

jpfr:
@EmmanuelP another thing came up:
We should first try to send. And only when that fails with UA_INTERRUPTED add a short `poll` before a retry.
Otherwise we double the number of roundtrips to the kernel for every send....

EmmanuelP:
> @EmmanuelP another thing came up: We should first try to send. And only when that fails with UA_INTERRUPTED add a short `poll` before a retry. Otherwise we double the number of roundtrips to the kernel for every send....

Yes, right. I'll fix it.

EmmanuelP:
> Yes, right. I'll fix it.

Fixed in the last pull request version. I still have to test it using different machines for the producer and the consumers. I'll keep you informed.

EmmanuelP:
> I still have to test it using different machines for the producer and the consumers. I'll keep you informed.

Test done. I think the patch is ready now.

jpfr:
Cool, rebased for 1.3.

PR #4712, opened 2021-10-20
fix(pubsub): UA_Server_addDataSetReader() init steps and cleanup

MartinaFelberFEM:
The createMonitoring callback needs a valid DataSetReader Id,
but the Id was previously assigned after invocation of the
createMonitoring callback.

Therefore we have changed the init order and also added some
memory cleanups.


jpfr:
Cool!
Please rebase for the 1.3 branch.

PR #4717, opened 2021-10-22
feat(server) automatic copy applicationdescription to the endpoints

andreasebner:
The endpoints contain an application description. If the user changes the server application description, the endpoints are currently not updated. This PR changes the default behavior and copy the latest application description to all the endpoints during the server start.


jpfr:
Let’s do this before 1.3 drops.

andreasebner:
Replaced by #4958

Issue #4651, opened 2021-09-16
CTT issue: View Minimum Continuation Point01 script 012.js fails


mlang-de:
The current master and V1.2.2 fails the test script.
V1.2.1 the test script passed successfully.

## Checklist

Please provide the following information:

 - [ ] open62541 Version (release number or git tag):
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

mlang-de:
Fixed by linked Pull Request.

PR #4724, opened 2021-10-28
fix(core): browseReferences continuation point usage check

sqadasi:
the loop should break when it finds the correct referenceType of the continuation point next target, and not after the first iteration. If no reference found after finishing all iterations, then it is ok to return an error.
This should fix the issue https://github.com/open62541/open62541/issues/4651


mlang-de:
> Thanks! Reformulated for clarity. Not sure this fixes #4651. That could only happen if the nodestore has multiple entries for the same ReferenceType.

If I understand it correct then this happens quite often.
For instance below the "Objects" (ns=0;i=85) base node with the "Server object" and e.g. "Device Set" from DI.
Or below the ParameterSet node from DI where many Variables with HasComponent referenced.

Issue #3082, opened 2019-09-05
Example 'server_nodeset_powerlink' is not working


ckmk14:
## Description
Using the master branch the example 'server_nodeset_powerlink' is not working:

```
info/session	Connection 0 | SecureChannel 0 | Session g=00000001-0000-0000-0000-000000000000 | AddNodes: The value rank of ns=3;i=682 is incomatible with the value rank of the VariableType
info/session	Connection 0 | SecureChannel 0 | Session g=00000001-0000-0000-0000-000000000000 | AddNodes: Type-checking the variable node ns=3;i=682 failed with error code BadTypeMismatch
error/server	Adding the Powerlink namespace failed. Please check previous error output.
```

## Background Information / Reproduction Steps

Used CMake options:

```bash
UA_ARCHITECTURE:STRING=posix
UA_BUILD_EXAMPLES:BOOL=ON
UA_BUILD_TOOLS:BOOL=OFF
UA_BUILD_UNIT_TESTS:BOOL=OFF
UA_ENABLE_AMALGAMATION:BOOL=OFF
UA_ENABLE_COVERAGE:BOOL=OFF
UA_ENABLE_DISCOVERY:BOOL=ON
UA_ENABLE_DISCOVERY_MULTICAST:BOOL=OFF
UA_ENABLE_ENCRYPTION:BOOL=OFF
UA_ENABLE_EXPERIMENTAL_HISTORIZING:BOOL=OFF
UA_ENABLE_HISTORIZING:BOOL=OFF
UA_ENABLE_METHODCALLS:BOOL=ON
UA_ENABLE_MICRO_EMB_DEV_PROFILE:BOOL=OFF
UA_ENABLE_NODEMANAGEMENT:BOOL=ON
UA_ENABLE_QUERY:BOOL=OFF
UA_ENABLE_STATIC_ANALYZER:STRING=OFF
UA_ENABLE_SUBSCRIPTIONS:BOOL=ON
UA_ENABLE_SUBSCRIPTIONS_EVENTS:BOOL=OFF
UA_ENABLE_WEBSOCKET_SERVER:BOOL=OFF
UA_LOGLEVEL:STRING=300
UA_NAMESPACE_ZERO:STRING=FULL

```

NoelGraf:
PR #4737

peschuster:
I think this is actually a bug in the Opc.Ua.POWERLINK.NodeSet2.xml and not in open62541.

[...]

PR #4737, opened 2021-11-01
fix(nc): Inherit the values rank even if the parents have < -1

NoelGraf:
Currently, the Powerlink nodeset example does not run. The problem is that the Value Rank does not match the ValueRank of the VariableType #3082. 

The PowerlinkVariableType is assigned a ValueRank of -1 (default). However, in the XML the node has a subtype BaseDataVariableType (ns=0;i=63), which has a ValueRank of -2.


jpfr:
Looks good!
Switched to target the 1.3 branch.
Can you rebase and only include the relevant commits?

NoelGraf:
Yes, I can do that. Currently, a test in Alarms&Conditions still fails. Here a value rank of -1 is expected, but a -2 is set. Must look at which value is correct.

jpfr:
Nice!

PR #4738, opened 2021-11-01
[REVIEW] fix(server): Delete node with non-existing parent

robertns:
`UA_Server_addVariableNode()` has potential memory leak if iParentId node is non-existent. In that case, `AddNode_addRefs()` will return UA_STATUSCODE_BADPARENTNODEIDINVALID, but the node will be released, not removed.


jpfr:
Looks good!
Switched to target 1.3. But that added more commits from master.
Can you clean up the commit history for the 1.3 target branch?

jpfr:
Updated for 1.3

robertns:
Thanks! I cleaned up commit history, only my commit is now included.

PR #4739, opened 2021-11-01
[REVIEW] Fix #4633 Linux compilation error when enable UA_ENABLE_AMALGAMATION and UA_MULTITHREADING

### LINKED ISSUE #4633: Linux comple open62541.h/.c error with UA_MULTITHREADING  in masterBranch
git log : commit f7e679a330fdb0b66b068692c0f5d8a9695a675f
Used CMake options:
addtional cmake option from default:
**UA_ENABLE_AMALGAMATION 
UA_MULTITHREADING 100**
UA_ENABLE_SUBSCRIPTIONS_EVENTS.

I get Error such as:
_Scanning dependencies of target open62541-object
[ 65%] Building C object CMakeFiles/open62541-object.dir/open62541.c.o
In file included from /tmp/wangyichen/open62541/build/open62541.c:28:
/tmp/wangyichen/open62541/build/open62541.h:512:17: error: expected ‘;’ before ‘void’
  512 | static UA_INLINE void
      |                 ^~~~~
      |                 ;
/tmp/wangyichen/open62541/build/open62541.h:513:1: error: no previous prototype for ‘UA_LOCK_INIT’ [-Werror=missing-prototypes]
  513 | UA_LOCK_INIT(UA_Lock *lock) {
      | ^~~~~~~~~~~~
/tmp/wangyichen/open62541/build/open62541.h:520:17: error: expected ‘;’ before ‘void’
  520 | static UA_INLINE void
      |                 ^~~~~
      |                 ;
/tmp/wangyichen/open62541/build/open62541.h:521:1: error: no previous prototype for ‘UA_LOCK_DESTROY’ [-Werror=missing-prototypes]
  521 | UA_LOCK_DESTROY(UA_Lock *lock) {
      | ^~~~~~~~~~~~~~~
/tmp/wangyichen/open62541/build/open62541.h:526:17: error: expected ‘;’ before ‘void’
  526 | static UA_INLINE void
      |                 ^~~~~
      |                 ;
/tmp/wangyichen/open62541/build/open62541.h:527:1: error: no previous prototype for ‘UA_LOCK’ [-Werror=missing-prototypes]
  527 | UA_LOCK(UA_Lock *lock) {_

**But I complie  UA_MULTITHREADING OK whitout UA_ENABLE_AMALGAMATION. And I try to complie on Realsea 1.2.2 and it works well. Will this bug be fixed soon?**

kimim:
Fix issue #4633 by change UA_INLINE and UA_assert locations.


jpfr:
This duplicates a lot of identical definitions.
I'd prefer to have these in a central location.
That could be include/open62541/architecture_definitions.h or even the config.h

kimim:
has some problems that if #include <assert.h> earlier than <arch>/ua_architecture.h, will cause compilation error.

jpfr:
open62541/architecture_definitions.h and open62541/architecture_functions.h are included at the top of the <arch>/ua_architecture.h files.

There should be a way of solving this without code duplication.

My guess is that the order of the include files in CMakeLists.txt has to be adjusted.
Note that the order of the files there results in the order used inside the open62541.c/.h.

kimim:
Yes. in normal compile,

arch/ua_architecture.h will include 
`open62541/architecture_definitions.h` in around line 38 for windows, 
line 195 for posix.

But when AMALGAMATION happens, open62541/architecture_definitions.h will 
be removed and added a bit later. This is because in amalgamation.py, it 
will filter some include starting with `open62541/`


One possible solution is to split arch/ua_architecture.h into two parts:

- arch/ua_architecture_pre.h

- arch/ua_architecture_post.h

And then include them in config.h:


```

#include "${UA_ARCHITECTURE}/ua_architecture_pre.h"

#include "pen62541/architecture_definitions.h"

#include "${UA_ARCHITECTURE}/ua_architecture_post.h"

```

But this will cause problems in other code which include 
`${UA_ARCHITECTURE}/ua_architecture.h`. Well, luckily, no other code is 
include this.

jpfr:
I found a less invasive solution: #4747

Thanks for bringing this forward.
You are mentioned as the Co-Author for the commits in #4747.

kimim:
great. Glad that this issue is resolved.

Issue #2835, opened 2019-06-20
Historize data in a circular buffer


luibass92:
Goodmorning,

I am a "Historical Data" plugin user and I see that there is no limit on the size of the buffer which historizes data.
I can set an initial size but, each time it is reached, a UA_realloc is done and the buffer is re-allocated with its size doubled.

I would like to configure the historical data in order to have something similar to a circular buffer beacause I want to be sure to not have undesired memory growths.

I managed how to do it with the Gathering, now I can register only a maximum number of nodes and if I overcome that limit an UA_STATUS_BADOUTOFMEMORY is returned.
Now I am trying to do the same with the Backend part but it seems way more complicated.
In this case I do not want to simply return a bad status code, but I want to historize data in a circular buffer which does not double its size but overwrite the oldes data as a circlar buffer would do.

Someone could please help me or give me some tips?

Thanks,
Luigi

jpfr:
@peter-rustler-basyskom 
@basyskom-meerkoetter 

Can you have a look at this?

gcn1996:
I have the same question,it will take too much memory and break the program

luibass92:
Hi everyone,
after several months I have implemented by myself a solution to have a "circular history".

The idea behind it is to have a fixed number of nodes to historize, that means it is mandatory to choose a maximum number of nodes at compile time (and that number will never be exceeded) and for each of that nodes a circular array of values will be saved and, each time a client performs a UA_HistoryReadRequest on a certain node, the current circular buffer (full of values) will be sent by the server.

To do so I started from the original history functions and I did some modifications to obtain the desired result. This was the faster solution for me but it means that most of the code it is very similar to the original implementation. Sorry for that but I did not found a more clever and fast way to implement the algorithms.

The following images describe the differences between the original history and the "circular" one.

Original implementation of OPC UA Historical Data Access: nodes and
corresponding values can grow indefinitely:

![orig_hist](https://user-images.githubusercontent.com/40570739/83265179-6bf1f300-a1c1-11ea-9afc-e38cdbf5702a.PNG)

Modified implementation of OPC UA Historical Data Access: if max
number of nodes is reached (in the image is 3), registration of other nodes is not allowed. If max
number of values is reached (in the image is 4), subsequent values are saved overwriting oldest ones
following the rules of circular buffers

![circ_hist](https://user-images.githubusercontent.com/40570739/83265330-a65b9000-a1c1-11ea-90bc-c609b49cd69d.PNG)

[...]

PR #4740, opened 2021-11-01
[REVIEW] Circular buffer implementation for historydata plugin

luibass92:
The idea behind the circular buffer historydata implementation is to have a fixed number of nodes to historize, that means it is mandatory to choose a maximum number of nodes (and that number will never be exceeded) and for each of that nodes a circular array of values will be saved and, each time a client performs a UA_HistoryReadRequest on a certain node, the current circular buffer (full of values) will be sent by the server.

To do so I started from the original history functions and I did some modifications to obtain the desired result. This was the faster solution for me but it means that most of the code it is very similar to the original implementation. Sorry for that but I did not find a faster way to implement the algorithms.

The following images describe the differences between the original history and the "circular" one.

Original implementation of OPC UA Historical Data Access: nodes and
corresponding values can grow indefinitely:

83265179-6bf1f300-a1c1-11ea-9afc-e38cdbf5702a

Modified implementation of OPC UA Historical Data Access: if max
number of nodes is reached (in the image is 3), registration of other nodes is not allowed. If max
number of values is reached (in the image is 4), subsequent values are saved overwriting oldest ones
following the rules of circular buffers

83265330-a65b9000-a1c1-11ea-90bc-c609b49cd69d

Moreover it is useful to know that I have added a field to the UA_NodeIdStoreContextItem_backend_memory struct in order to make things easier to implement.

To try this implementation all you need to do is to use the consturctors:

`UA_HistoryDataBackend_Memory_Circular`
and
`UA_HistoryDataGathering_Circular`

While the History Database part (interface) will remain untouched.

Original issue: https://github.com/open62541/open62541/issues/2835

### DIFF
--- a/plugins/historydata/ua_history_data_backend_memory.c
+++ b/plugins/historydata/ua_history_data_backend_memory.c
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
  *    Copyright 2018 (c) basysKom GmbH <opensource@basyskom.com> (Author: Peter Rustler)
+ *    Copyright 2021 (c) luibass92 <luibass92@live.it> (Author: Luigi Bassetta)
  */
 
 #include <open62541/plugin/historydata/history_data_backend_memory.h>
@@ -26,6 +27,8 @@ typedef struct {
     UA_DataValueMemoryStoreItem **dataStore;
     size_t storeEnd;
     size_t storeSize;
+    /* New field useful for circular buffer management */
+    size_t lastInserted;
 } UA_NodeIdStoreContextItem_backend_memory;
 
 static void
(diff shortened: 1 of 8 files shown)


jpfr:
Great!
I added the [WIP] prefix to the PR.
You might want to have a look at the builds on different architecture.

Just set the prefix to [Review] when you are happy.
Or trigger us when we should at a particular build in details.

Looking forward to this!

jpfr:
Uhm. It appears the Appveyor build failed due to some infrastructure problems. Re-started that.

luibass92:
> Uhm. It appears the Appveyor build failed due to some infrastructure problems. Re-started that.

Hi @jpfr ,
I see that Clang Static Analyzer fails cause of 2 warnings related to:

```
/home/runner/work/open62541/open62541/plugins/historydata/ua_history_data_gathering_default.c:248:37: warning: Dereference of null pointer [core.NullDereference]
    ctx->dataStore[current].setting = setting;
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
```

This is something that I have implemented, in particular the entire method looks like this:

```
static UA_StatusCode
registerNodeId_gathering_circular(UA_Server *server, void *context,
                                  const UA_NodeId *nodeId,
                                  const UA_HistorizingNodeIdSettings setting) {
    UA_NodeIdStoreContext *ctx = (UA_NodeIdStoreContext *)context;
    if(getNodeIdStoreContextItem_gathering_default(ctx, nodeId)) {
        return UA_STATUSCODE_BADNODEIDEXISTS;
    }
    if(ctx->storeEnd >= ctx->storeSize) {
    }
    UA_NodeId_copy(nodeId, &ctx->dataStore[ctx->storeEnd].nodeId);
    size_t current = ctx->storeEnd;
    ctx->dataStore[current].setting = setting;     /* <------- The warning is generated here */
    ++ctx->storeEnd;
    return UA_STATUSCODE_GOOD;
}
```

But it looks strange because ther is another very similar method that not generates the same warning:

```
static UA_StatusCode
registerNodeId_gathering_default(UA_Server *server,
                                 void *context,
                                 const UA_NodeId *nodeId,
                                 const UA_HistorizingNodeIdSettings setting)
{
    UA_NodeIdStoreContext *ctx = (UA_NodeIdStoreContext*)context;
    if (getNodeIdStoreContextItem_gathering_default(ctx, nodeId)) {
        return UA_STATUSCODE_BADNODEIDEXISTS;
    }
    if (ctx->storeEnd >= ctx->storeSize) {
        size_t newStoreSize = ctx->storeSize * 2;
        ctx->dataStore = (UA_NodeIdStoreContextItem_gathering_default*)UA_realloc(ctx->dataStore,  (newStoreSize * sizeof(UA_NodeIdStoreContextItem_gathering_default)));
        if (!ctx->dataStore) {
            ctx->storeSize = 0;
        }
        memset(&ctx->dataStore[ctx->storeSize], 0, (newStoreSize - ctx->storeSize) * sizeof(UA_NodeIdStoreContextItem_gathering_default));
        ctx->storeSize = newStoreSize;
    }
    UA_NodeId_copy(nodeId, &ctx->dataStore[ctx->storeEnd].nodeId);
    size_t current = ctx->storeEnd;
    ctx->dataStore[current].setting = setting;     /* <------ NO warnings here */
    ++ctx->storeEnd;
    return UA_STATUSCODE_GOOD;
}
```

Am I missing something?
Btw I see that the Clang Static Analyzer is not a "required" step. What I am supposed to do at this point?

Thanks!

luibass92:
@jpfr I fixed the warning, needed another approval

plugins/historydata/ua_history_data_backend_memory.c
```diff
@@ -603,3 +606,268 @@ UA_HistoryDataBackend_Memory_clear(UA_HistoryDataBackend *backend)
...
+    if(continuationPoint->length > 0) {
+        if(continuationPoint->length >= sizeof(size_t)) {
+            skip = *((size_t *)(continuationPoint->data));
+            if(continuationPoint->length > 0) {
+                backendContinuationPoint.length =
+                    continuationPoint->length - sizeof(size_t);
+                backendContinuationPoint.data = continuationPoint->data + sizeof(size_t);
+            }
+        } else {
+            return UA_STATUSCODE_BADCONTINUATIONPOINTINVALID;
```
jpfr, line 746: Treat the error case first.
Then you don't have to indent into an else condition.
Flat code is good.

plugins/historydata/ua_history_data_backend_memory.c
```diff
@@ -603,3 +606,268 @@ UA_HistoryDataBackend_Memory_clear(UA_HistoryDataBackend *backend)
...
+                outResult[counter].sourceTimestamp = start;
+            }
+            ++counter;
+        }
+    }
+    UA_ByteString backendOutContinuationPoint;
+    UA_ByteString_init(&backendOutContinuationPoint);
+    if(endIndex != storeEnd && startIndex != storeEnd) {
+        size_t retval = 0;
+
```
jpfr, line 791: remove the blank line.

plugins/historydata/ua_history_data_backend_memory.c
```diff
@@ -603,3 +606,268 @@ UA_HistoryDataBackend_Memory_clear(UA_HistoryDataBackend *backend)
...
+            if(skip == 0) {
+                valueSize = _resultSize - addFirst - addLast;
+            } else {
+                valueSize = _resultSize - skip - addLast;
+            }
+        }
+
+        UA_StatusCode ret = UA_STATUSCODE_GOOD;
+        if(valueSize > 0)
+
```
jpfr, line 803: remove the blank line.

plugins/historydata/ua_history_data_backend_memory.c
```diff
@@ -603,3 +606,268 @@ UA_HistoryDataBackend_Memory_clear(UA_HistoryDataBackend *backend)
...
+            ret = backend->copyDataValues(
+                server, backend->context, sessionId, sessionContext, nodeId, startIndex,
+                endIndex, reverse, valueSize, range, releaseContinuationPoints,
+                &backendContinuationPoint, &backendOutContinuationPoint, &retval,
+                &outResult[counter]);
+        if(ret != UA_STATUSCODE_GOOD) {
+
```
jpfr, line 810: remove the blank line.

plugins/historydata/ua_history_data_backend_memory.c
```diff
@@ -603,3 +606,268 @@ UA_HistoryDataBackend_Memory_clear(UA_HistoryDataBackend *backend)
...
+            UA_Array_delete(outResult, *resultSize, &UA_TYPES[UA_TYPES_DATAVALUE]);
+            *result = NULL;
+            *resultSize = 0;
+            return ret;
+        }
+        counter += retval;
+    }
+    if(addLast && counter < *resultSize) {
+
```
jpfr, line 819: remove the blank line.

plugins/historydata/ua_history_data_backend_memory.c
```diff
@@ -603,3 +606,268 @@ UA_HistoryDataBackend_Memory_clear(UA_HistoryDataBackend *backend)
...
+        outResult[counter].hasStatus = true;
+        outResult[counter].status = UA_STATUSCODE_BADBOUNDNOTFOUND;
+        outResult[counter].hasSourceTimestamp = true;
+        if(start == LLONG_MIN &&
```
jpfr, line 823: Clean the funny newlines.

plugins/historydata/ua_history_data_backend_memory.c
```diff
@@ -603,3 +606,268 @@ UA_HistoryDataBackend_Memory_clear(UA_HistoryDataBackend *backend)
...
+           storeEnd != backend->firstIndex(server, backend->context, sessionId,
+                                           sessionContext, nodeId)) {
+            outResult[counter].sourceTimestamp =
```
jpfr, line 826: Please document that code.
Why do you subtract one second at the end?

plugins/include/open62541/plugin/historydata/history_data_backend_memory.h
```diff
@@ -17,6 +18,9 @@ _UA_BEGIN_DECLS
 UA_HistoryDataBackend UA_EXPORT
 UA_HistoryDataBackend_Memory(size_t initialNodeIdStoreSize, size_t initialDataStoreSize);
 
+UA_HistoryDataBackend UA_EXPORT
```
jpfr, line 21: Add a short documentation to the public API.
What is the meaning of the arguments?

jpfr (commented):
Looking good!

Please address the (mostly cosmetic) inline comments.
Also add a minimal example to the unit tests as well.

That way we can ensure the code won't break in the future.

plugins/historydata/ua_history_data_backend_memory.c
luibass92: @jpfr  This part of the code is identical to the "normal" implementation. I did not change it because is not the core part of the circular buffer historizing algorithm.
Maybe @peter-rustler-basyskom know tha answer.

Meanwhile I implemented an example and a minimal unit test for the circular buffer, as well as following your suggestions on cosmetics and code documentation.

luibass92:
@jpfr I implemented an example and a minimal unit test for the circular buffer, as well as following your suggestions on cosmetics and code documentation 👍

jpfr:
Looking good!
Thanks for this contribution.

luibass92:
Thanks! I will close the related issue.

rjapeer:
@luibass92 @jpfr 
It seems that the changes were reverted in the 1.3 branch?
My questions:
* Was this done on purpose? 
* Which branch should we use to have the circular buffer available again?

luibass92:
> 

Hi @rjapeer ,
the circular buffer is currently present on the master branch.
I don't know why on 1.3 it has not been included.

PR #4748, opened 2021-11-04
fix(nc): fix ordering of types parsed from xml

olivluca:
Running the generator under windows and linux produced a different types_generated.h (with a different ordering of the types).
Using and OrderedDict for snippets preserves the ordering and gives a more consistent result.


jpfr:
Rebased for the upcoming release branch 1.3 in #4753.
The commit is still yours.

olivluca:
No 1.2.x?

Issue #4156, opened 2021-01-29
Build fails when MSYS generator option is enabled 


opcua-tsn-team-kalycito:
## Description
The objective here is to build the shared libraries of the open62541 stack with MinGW in windows environment. After generating the CMake files, make command fails with the following output.

 
```

Scanning dependencies of target open62541-generator-transport

make[2]: Leaving directory '/c/Users/TESTPC Developer/Documents/test/open62541_latest/open62541/build'

make[2]: Entering directory '/c/Users/TESTPC Developer/Documents/test/open62541_latest/open62541/build'

[ 30%] Generating src_generated/open62541/transport_generated.c, src_generated/open62541/transport_generated.h, src_generated/open62541/transport_generated_handling.h, src_generated/open62541/transport_generated_encoding_binary.h

Traceback (most recent call last):

 File "C:/Users/TESTPC Developer/Documents/test/open62541_latest/open62541/tools/generate_datatypes.py", line 95, in <module>

 parser.create_types()

 File "C:\Users\TESTPC Developer\Documents\test\open62541_latest\open62541\tools/nodeset_compiler/type_parser.py", line 366, in create_types

 self.parse_types()

 File "C:\Users\TESTPC Developer\Documents\test\open62541_latest\open62541\tools/nodeset_compiler/type_parser.py", line 391, in parse_types

 self.parseTypeDefinitions(outname_import, file_import)

 File "C:\Users\TESTPC Developer\Documents\test\open62541_latest\open62541\tools/nodeset_compiler/type_parser.py", line 286, in parseTypeDefinitions

 xmlDoc = etree.iterparse(

 File "C:/dev/msys64/mingw64/lib/python3.8/xml/etree/ElementTree.py", line 1248, in iterparse

 source = open(source, "rb")

OSError: [Errno 22] Invalid argument: 'C;C:\\dev\\msys64\\Users\\TESTPC Developer\\Documents\\test\\open62541_latest\\open62541\\deps\\ua-nodeset\\Schema\\Opc.Ua.Types.bsd'

make[2]: *** [CMakeFiles/open62541-generator-transport.dir/build.make:67: src_generated/open62541/transport_generated.c] Error 1

make[2]: Leaving directory '/c/Users/TESTPC Developer/Documents/test/open62541_latest/open62541/build'

make[1]: *** [CMakeFiles/Makefile2:441: CMakeFiles/open62541-generator-transport.dir/all] Error 2

[...]

PR #4752, opened 2021-11-05
[REVIEW] fix issue #4156, there should be no prefix of a path for python

kimim:
when python fetch the command line parameters, it will take care of `os.pathsep`. 
If there is some characters before `C:/XXX`, such as `--format=TYPES#C:/XXX`, it will convert this to:
`--format#C;C:/Users/Home/XXX`
That means python will consider `C:/XXX` as two paths. This cause the error as #4156 log:

```
OSError: [Errno 22] Invalid argument: 'C;C:\\dev\\msys64\\Users\\TESTPC Developer\\Documents\\test\\open62541_latest\\open62541\\deps\\ua-nodeset\\Schema\\Opc.Ua.Types.bsd'
```

The solution is moving `TYPES#` as postfix:

`--format=C:/XXX#TYPES`


kimim:
too long to wait for the CI results.. 
I need to go sleep now..

andreasebner:
@NoelGraf Could you please review the changes?

jpfr:
I'm hesitant to accept this.
Because it will break generator scripts people have in their own build pipeline.

It appears to me we can fix this by working around the call to `os.pathsep`.
Can't you just wrap the input argument string in a quotes `"` to get around this?

jpfr:
@kimim Could you have a look if there is backwards-compatible workaround?

kimim:
I’ve tried quoting the parameter, but cannot avoid the issue. So I have to move the prefix to postfix. 
It seems the path conversion is happened before python get it.
I can have a try once more.

kimim:
@jpfr I use a simpler solution to solve issue #4156 
Could you help review? Thanks.

kimim:
I have a problem that the CI script is running in powershell, which cannot accept an environment variable in posix way..

jpfr:
I think you need to put `"MSYS2_ARG_CONV_EXCL=--import"` into quotes...

kimim:
Try again. On my local bash mintty terminal, I have many options:
- export MSYS2_ARG_CONV_EXCL=--import && python ...
- env MSYS2_ARG_CONV_EXCL=--import   python ...
- MSYS2_ARG_CONV_EXCL=--import   python ...

But all failed in CI's powershell script

jpfr:
Found a way to set the environment that appears to work.
Can you verify in your environment?

kimim:
this is not working inside mingw64 bash environment.
I've tried this way in CMakeList.txt.
cmake manual says ENV only work for the function level.
Here python is invoked inside `add_custom_command` function.

There is a paradox: 
- we do not need to set `MSYS2_ARG_CONV_EXCL` for powershell, because it will not convert path automatically.
- we need `MSYS2_ARG_CONV_EXCL` for mingw64 shell, because it convert path automatically.

How about add more judgement: if MINGW and MINGW64_Shell.
Then the code will not affect powershell script.

Or another solution, we just write it down in MSYS2 manual:
- please `export MSYS2_ARG_CONV_EXCL=--import` before compilation.

jpfr:
> How about add more judgement: if MINGW and MINGW64_Shell.
Then the code will not affect powershell script.

Sounds good!

kimim:
This commit should work for both powershell + mingw and bash + mingw.

kimim:
@opcua-tsn-team-kalycito could you have a test with this PR?
It works on my machine's MingW64 bash shell.

jpfr:
Seems to work.
If there are issues, let's update separately.
Merging!

PR #4758, opened 2021-11-08
[Review] fix(plugin): Fix validation of ref type index range

cldhms:


jpfr:
Nice!
Switched to target the 1.3 branch.
Can you remove the commits form `master` that are still included?

cldhms:
I closed this one by mistake, see #4760 instead...

PR #4759, opened 2021-11-08
[Review] fix(arch): UA_fd_isset() evaluates to non-zero value

rdyhms:
fix(arch): UA_fd_isset() evaluates to non-zero value

UA_fd_isset() evaluates to a non-zero value if the descriptor is part of the set.

See https://pubs.opengroup.org/onlinepubs/9699919799.2016edition/functions/select.html


jpfr:
Very good.
Switched the target to the 1.3 branch.
Can you remove the commits from `master` that were introduced?

jpfr:
Nice, thanks!

PR #4780, opened 2021-11-17
server(deps) update and fix ua-nodeset (5fdc7d4 to f71b3f4)

andreasebner:
Move ua-nodeset to f71b3f4 (Tag Glass=1.0.0-2022-01-01)
Fix nodeset compiler to handle new empty enumartion, contained in types-bsd.


andreasebner:
@GoetzGoerisch This PR replaces the autogenerated #4718. The dependabot-PR failed because the ua-nodeset update contan an empty enumeration which caused an issue in our nodeset compiler.

jpfr:
Nice!

GoetzGoerisch:
Thanks  @andreasebner!

PR #4802, opened 2021-11-26
[Review] core(deps): Bumps LibreSSL to 3.4.2

GoetzGoerisch:
* fixes [CB-K21-1240](https://www.cert-bund.de/advisoryshort/CB-K21-1240)


jpfr:
👍

PR #4816, opened 2021-12-03
[WIP] feat(server): Forward UA_Arguments of OutputArguments Node in method callback

jonasgreen88:
When using a generic method callback for all methods in a nodestore
it is useful to have access to the UA_Arguments of the
OutputArguments node in the callback when writing the output
variants. In our case we do a translation of method calls to a
subsystem and need to know the OPC UA types when translating the
subsystem output arguments to OPC UA output arguments.

The callback function can find the OutputArguments node itself by
browsing the method NodeId but as the stack has already done
this work just before calling the method callback we see it as a
possibility to forward this information as an argument.

### DIFF
--- a/examples/events/server_random_events.c
+++ b/examples/events/server_random_events.c
@@ -162,7 +162,8 @@ addGenerateSampleEventsMethodCallback(UA_Server *server,
                              const UA_NodeId *methodId, void *methodContext,
                              const UA_NodeId *objectId, void *objectContext,
                              size_t inputSize, const UA_Variant *input,
-                             size_t outputSize, UA_Variant *output) {
+                             size_t outputSize, UA_Variant *output,
+                             UA_Argument *outputArguments) {
     UA_StatusCode retval;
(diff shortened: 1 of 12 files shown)


jpfr:
Hey there,

We really shouldn't break existing code. And there is a lot of code that uses methods.

I much prefer that we provide a dedicated function that looks up the arguments based on the nodeid.
We could even do some sort of "caching" where accessing the arguments node is faster if we call it from inside a method callback.

jonasgreen88:
> Hey there,
> 
> We really shouldn't break existing code. And there is a lot of code that uses methods.
> 
> I much prefer that we provide a dedicated function that looks up the arguments based on the nodeid. We could even do some sort of "caching" where accessing the arguments node is faster if we call it from inside a method callback.

Thanks for the feedback. We do agree and intend to come back with a new implementation in the future.

andreasebner:
Thanks for the concept. If the alternative implementation is ready please open a new PR.

PR #4826, opened 2021-12-08
fix: Fuzzing didn't compile because of missing sources.

mlgiraud:
Also reduce log spam and format code in eventloop.


jpfr:
In general good.
But please keep the white space changes in separate commits.
Also don’t let your editor make auto-changes to whitespaces..
It’s better to be consistent and not mix styles...

mlgiraud:
Done

PR #4829, opened 2021-12-08
nodeset_compiler: Fix parent node reference resolution.

gehwolf:
Currently the parent node and the necessary reference id for the
relation is not generated correctly.
This fixes the issue for the parent reference types 'Organizes',
'HasProperty', 'HasComponent' and 'HasHistoricalConfiguration'.
However this is only a hot fix and the code needs more testing and
refactoring.


jpfr:
There are a lot of tests failing with this.
Can you unpack what you think the problem is?

Note that there can be more than one (hierarchical) reference pointing upwards.

gehwolf:
sorry for my late response, I think I have a fix or more a workaround. I will  try to add a test example in tests/nodeset_compiler/ which should point out my problem.

andreasebner:
@NoelGraf Could you please review this changes?

tools/nodeset_compiler/nodes.py
```diff
@@ -123,6 +125,17 @@ def parseXML(self, xmlelement):
...
                     self.parseXMLReferences(x)
 
+        # determine parent reference type
+        PARENT_REFERENCE_NODES = [
+                NodeId("ns=0;i=35"), NodeId("ns=0;i=45"),
+                NodeId("ns=0;i=46"), NodeId("ns=0;i=47"),
+                NodeId("ns=0;i=56"),
+        ]
+        for ref in self.references:
+            if ref.referenceType in PARENT_REFERENCE_NODES and not ref.isForward:
```
NoelGraf, line 135: The referenceType is of type str. Therefore the check will never be true. And what happens if more references match the condition?

NoelGraf:
@gehwolf 
We want to merge your changes. However, I cannot fully understand your changes. Therefore, you need to explain your changes. Maybe give me a short example that shows the problem and your solution, or is there an open issue that shows the problem?

andreasebner:
Hi, thanks for your contribution. We tried to follow your changes but without further comment we are not abel to merge it. I'am not sure if these changes are correct since the specification says: "The NodeId of the Node that is the parent of the Node within the information model. This field is used to indicate that a tight coupling exists between the Node and its parent (e.g. when the parent is deleted the child is deleted as well). This information does not appear in the AddressSpace and is intended for use by design tools."  

Feel free to reopen the PR with further comments ;)

PR #4831, opened 2021-12-09
refactor: use consistent CMakeLists.txt style.

mlgiraud:
The formatting style in the cmake files was inconsitent. This PR uses the prevalent style. The formatting style will also be added into the .idea code formatting file in a different PR.


jpfr:
Not a fan of monster white space cleanup commits.
The problem is that merging fixes from 1.3 to master becomes much harder.

I much prefer to have automated checks for this.
But only for changed lines of code.

PR #4834, opened 2021-12-09
[REVIEW] fix(server): fix no match for TranslateBrowsePathsToNodeIds

cldhms:
If no matches are found a size of 0 is passed to UA_realloc(). The
behavior of realloc if the size is 0 is implementation defined and will
either return NULL or a valid pointer. Fix so bad out of memory is not
returned in case a NULL pointer shall be accepted.


cldhms:
Want to verify if the unittest catches this (as it is compiler dependent). Will upload fix after.

cldhms:
My unit test doesn't catch this as realloc for all tested platforms seem to return a non-NULL pointer if the requested size is 0. Is the test still useful?

jpfr:
Yes, still useful!
Merged. Thanks.

PR #4838, opened 2021-12-10
Fix SimpleAttributeOperand handling for empty path

basyskom-jvoe:
An empty path should operate on the instance, not on the type node.


jpfr:
Please correct the comment just above the change you made.
Can you reference the section of the standard where this comes from?

basyskom-jvoe:
> Please correct the comment just above the change you made. Can you reference the section of the standard where this comes from?

@jpfr This was already the correct quote from the specification. I have added a reference to the section of the standard.

PR #4839, opened 2021-12-10
fix(server): Monitoring out of bounds index range fails on CTT

jonasgreen88:
The following test cases of Monitored Item Services fails:
* Monitor Basic - Err-014-0.js
* Monitor Basic - Err-041.js
* Monitor Value Change - 016.js
* Monitor Value Change - 031.js
* Monitor Value Change - Err-001.js


jpfr:
👍

PR #4843, opened 2021-12-14
[REVIEW] fix(server): Continuation point for browse refs breaks too early 

jonasgreen88:
The foor loop looking for a reference kind match breaks
after the expected reference type index is found. But there
may be one more reference list of that type index but with a
different direction.

Found when running CTT.


mlang-de:
Seems to be similar to PR https://github.com/open62541/open62541/pull/4724

jpfr:
I backported #4724 to the 1.2 branch via #4844.
That is a cleaner solution than having the same fix twice.

Anyhow, thanks for reporting/fixing on the 1.2 branch, @jonasgreen88 !

Issue #4852, opened 2021-12-15
Invalid default initialization during UA_Server_addNode_finish


mlgiraud:
## Description
The current behavior of UA_Server_addNode_finish when processing a variable node with an array seems to be incorrect.
If the valueRank is set to one, and the ArrayDimensions is set to 0, the expected behavior (as far as i know) would be to have an array that is initialized to length zero. However, currently the length is initialised to one, and an empty element is inserted.

## Background Information / Reproduction Steps
Comment lines 1021 to 2026 in `ua_server_ns0.c`:
```
//    UA_Server_deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_ESTIMATEDRETURNTIME), true);
//    UA_Server_deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_LOCALTIME), true);
//    UA_Server_deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_REQUESTSERVERSTATECHANGE), true);
//    UA_Server_deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_RESENDDATA), true);
//    UA_Server_deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION), true);
//    UA_Server_deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVER_SETSUBSCRIPTIONDURABLE), true);
```
Then view the ServerCapabilities Variable in UAExpert. The Array will have one element with a null String. Expected behavior would be to get an empty list with no elements in them.

The following code is responsible for initializing the node:
```
/* ServerCapabilities - ns=0;i=12710 */

static UA_StatusCode function_namespace0_generated_3248_begin(UA_Server *server, UA_UInt16* ns) {
UA_StatusCode retVal = UA_STATUSCODE_GOOD;
UA_VariableAttributes attr = UA_VariableAttributes_default;
attr.minimumSamplingInterval = 0.000000;
attr.userAccessLevel = 1;
attr.accessLevel = 1;
attr.valueRank = 1;
attr.arrayDimensionsSize = 1;
UA_UInt32 arrayDimensions[1];
arrayDimensions[0] = 0;
attr.arrayDimensions = &arrayDimensions[0];
attr.dataType = UA_NODEID_NUMERIC(ns[0], 12LU);
[...]

PR #4853, opened 2021-12-15
fix(server): Fix default initialization for arrays (#4852)

mlgiraud:
This fixes the issue described in #4852


jpfr:
Looks good.
Let's use this approach everywhere.

I think this code might generate values that fail our own type-checker.
See here: https://github.com/open62541/open62541/blob/ac3abb1846dec89f04b70e50f5dbc895df62dac7/src/server/ua_services_attribute.c#L991

Please also add the special case of always-zero arraydimensions over there.

And bolt it down with a unit test. :-)

mlgiraud:
@jpfr Please check and merge if you are satisfied.

src/server/ua_services_attribute.c
```diff
@@ -992,12 +992,17 @@ UA_Boolean
...
     UA_UInt32 tempArrayDimensions;
     if(!valueArrayDimensions && !UA_Variant_isScalar(value)) {
-        valueArrayDimensionsSize = 1;
-        tempArrayDimensions = (UA_UInt32)value->arrayLength;
-        valueArrayDimensions = &tempArrayDimensions;
+        if(value->arrayLength > 0) {
+            valueArrayDimensionsSize = 1;
+            tempArrayDimensions = (UA_UInt32)value->arrayLength;
+            valueArrayDimensions = &tempArrayDimensions;
+        } else {
```
jpfr, line 1002: This always accepts arrays of length zero with no array dimensions.
Here, we should always check that `dimSum==0` as well.
Otherwise we can slip in an array of length zero even if the value rank requires something else.

jpfr (commented):
Looks good. One minor comment.
Can be merged then!

PR #4856, opened 2021-12-16
fix(encoding): Check arrayLength against arrayDimensions

mlgiraud:
Before the fix it was possible to have inconsistent arrayLength and arrayDimensions values resulting in a decoding error on the client side.


andreasebner:
@jpfr @mlgiraud Is this PR up-to-date? Can we merge the changes?

PR #4857, opened 2021-12-16
fix(core): SendServiceFault incorrectly sends empty response.

mlgiraud:
Expected behavior would be that the server sends the dedicated ServiceFault Service Message.
This is now correctly done, and encoding errors during sending of a response are now also caught and sent as serviceFault instead of falling through resulting in the channel being closed.

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -40,11 +40,10 @@ void UA_debug_dumpCompleteChunk(UA_Server *const server, UA_Connection *const co
 /********************/
 
 UA_StatusCode
-sendServiceFault(UA_SecureChannel *channel, UA_UInt32 requestId, UA_UInt32 requestHandle,
-                 const UA_DataType *responseType, UA_StatusCode statusCode) {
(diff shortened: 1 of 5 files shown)


jpfr:
The general idea is very good.

I'm a bit hesitant when looking at `ua_services_subscription.c`.
That could be intentional to return a `UA_PublishResponse`.
Please check with the CTT if that is okay.

mlgiraud:
I don't think the PublishResponse is required here, because it will just be initialized to zero everywhere. But i will check with the ctt (if it runs now)

jpfr:
Part if this is now in 1.3. But not everything.
Was the recursion avoidance a problem you actually enountered or mostly "defensive programming"?

mlgiraud:
I think i encountered a problem but i don't remember how i triggered it. However, it is possible that the encoding will fail because of some underlying issue that caused the original service result to fail. In this case it would be unwise to keep trying to send a service fault. I would keep the check, because crashing from a stack overflow should be avoided.

andreasebner:
@jpfr @mlgiraud Should we close this PR since parts are contained in 1.3 or are there remaining changes which are still worth a merge?

PR #4862, opened 2021-12-19
fix: use `#!/usr/bin/env python3` instead of `#!/usr/bin/env python`

Pro:
The newer Ubuntu since 20.04 do not have the `python` binary.
There one explicitly needs to choose between `python2` or `python3`.

See also:
https://askubuntu.com/q/1296790/141958


jpfr:
Thanks @Pro !

PR #4865, opened 2021-12-21
[REVIEW] Fix the isDynamic flag for nested child variables

basyskom-jvoe:
If a node with Variable children is added, the variable children are never marked as dynamic.
This prevents the usage of the server and source timestamps in the DataValue.


jpfr:
Nice!

PR #4870, opened 2021-12-25
(fix)PubSub: Update mqtt.c

jackybek:
correct typo error : useTSL change to useTLS


jpfr:
Huh?
That should have been prevented by our tooling automatically.

It appears that the latest mqtt TLS changes never compiled properly.

Please provide unit tests that make use of mqtt encryption.
When the build flags for encryption are activated.

That is now a prerequisite before we can make any new changes to the mqtt subsystem.

jpfr:
Merged.

PR #4871, opened 2021-12-25
(fix:PubSub): add fields for TLS MQTT

jackybek:

PR #4876, opened 2022-01-02
fix:(PubSub): missing type cast to TLS fields

jackybek:


jpfr:
You opened a lot of PR for MQTT with TLS.
Since these PR build upon each other, please combine them into a single PR with multiple commits.
That way, the overall aim becomes visible.

andreasebner:
Thank you for the contribution @jackybek! We are currently working on a complete MQTT overhaul and we will remove the copied version of mqtt-c in our dependency’s. Therefore, we also created a PR on MQTT-C to solve some warnings. @NoelGraf I think we can close this PR, since it's obsolete due to the overhaul?

NoelGraf:
@andreasebner Yes, we can close the PR. Besides, the changes are already included in the 1.3 branch.
I think we can close PR #4871 as well since the changes have become obsolete as well.

andreasebner:
Tank you @NoelGraf for your feedback and thanks @jackybek for the contribution.

PR #4877, opened 2022-01-02
fix(PubSub) : TLS related fields

jackybek:


jackybek:
refer to #4940 for the tutorial on mqtt publish and subscribe

jpfr:
You opened a lot of PR for MQTT with TLS.
Since these PR build upon each other, please combine them into a single PR with multiple commits.
That way, the overall aim becomes visible.

PR #4880, opened 2022-01-03
feat(util) : websockets connections Endpoints

jackybek:
handles new connection endpoints:
1. opc.mqtt://
2. ws://
3. wss://


jackybek:
The strange thing is that the changes i made here can be compiled and build on my machine.
But here the CI process fails.
From the error message, it does not seem to be related to the new codes.

PR #4882, opened 2022-01-04
feat(nc): expand reduced information model to support generated models

epetrevska:
The usage of generated models was not possible before with the reduced
information model. By adding couple of nodes, that is possible now.

### DIFF
--- a/tools/schema/Opc.Ua.NodeSet2.Reduced.xml
+++ b/tools/schema/Opc.Ua.NodeSet2.Reduced.xml
@@ -265,6 +265,163 @@
       <Reference ReferenceType="HasSubtype" IsForward="false">i=26</Reference>
     </References>
   </UADataType>
+  <UADataType NodeId="i=94" BrowseName="PermissionType">
+    <DisplayName>PermissionType</DisplayName>
+    <Documentation>https://reference.opcfoundation.org/v104/Core/docs/Part3/8.55</Documentation>
+    <References>
+      <Reference ReferenceType="HasProperty">i=15030</Reference>
+      <Reference ReferenceType="HasSubtype" IsForward="false">i=7</Reference>
+    </References>
+    <Definition Name="PermissionType" IsOptionSet="true">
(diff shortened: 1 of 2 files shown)


epetrevska:
I noticed now with my change that in both **datatypes_pubsub.txt** and **datatypes_minimal.txt** we have
- PermissionType
- RolePermissionType

My question is if I should remove them from the pubsub file, or it is ok if they are defined in both files?
Thanks,
Elena

jpfr:
Nice!

I’d the type is already in _minimal, then you don’t have to add it here.

Many of the additions concern loading of information models at runtime.
This looks like a feature that can be selected on/off depending the capabilities.
It would be good to put this behind a feature-flag.
Such as UA_ENABLE_INFORMATIONMODEL_LOADING.

Have a look at UA_ENABLE_DIAGNOSTICS where we add nodeset definitions and types based on the feature flag in the CMakeLists.txt.

jpfr:
Cool, can you have a look at the failing tests?

epetrevska:
@jpfr I fixed the failing tests which I understood, would you please mind checking what exactly is the error in the test "**continuous-integration/appveyor/pr**". Somehow I cannot find what exactly the problem is.

Thank you so much for your help.

epetrevska:
It seems to function after the merge commit, seems that everything is good.

epetrevska:
Hi @jpfr , would you mind reviewing it? Thanks a lot.

jpfr:
Thanks for this.

I fear that we are creating more and more our own ad-how information model “profiles”.
This is now adding a mixture of file access, role permissions, etc.
I think we need a future proof way of defining and managing these “profiles”.

Because these build flags become part of the interface we don’t want to change very often.

@andreasebner pulling you into the loop as well.

@epetrevska Sorry if this is now stalling this PR for a bit.
But we need a long-term vision for handling this stuff before we just pile on cmake flags and custom xml that we need to maintain.

PR #4888, opened 2022-01-05
[REVIEW] Feature: Eventloop server integration

JanSurft:
Integrates the server iteration for the eventloop.

### Summary:
- code for server-networklayer-tcp removed from `arch/network_tcp.c`
- define and initialize connection-managersin server config (`include/open62541/server.h`, `plugins/ua_config_default.c`)
- `UA_Server_setupEventLoop` method which is called in `UA_Server_run_startup` (`server.c`)
- replace server specific code with eventloop-run in `UA_Server_run_iterate` (`server.c`)
- stop event-sources in `UA_Server_run_shutdown` (`server.c`)
- server-connection related logic and structure in `src/ua_connection.c` & `src/ua_connection_internal.h`

### Future Work / Ideas
- networklayers (consistently integrate connection-managers/event-sources in networklayers - key-value-map ? )
- Get rid of complete arch/network_tcp server code and the corresponding logic of the callers?
- Discovery server integration next step

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -25,6 +25,142 @@
 #define MSG_NOSIGNAL 0
 #endif
 
+typedef struct ConnectionEntry {
+    UA_Connection connection;
+    LIST_ENTRY(ConnectionEntry) pointers;
+} ConnectionEntry;
+
+typedef struct {
+    const UA_Logger *logger;
+    UA_UInt16 port;
+    UA_UInt16 maxConnections;
+    UA_SOCKET serverSockets[FD_SETSIZE];
+    UA_UInt16 serverSocketsSize;
+    LIST_HEAD(, ConnectionEntry) connections;
+    UA_UInt16 connectionsSize;
+} ServerNetworkLayerTCP;
+
+/***************************/
+/* Server NetworkLayer TCP */
+/***************************/
+
+#define MAXBACKLOG     100
+#define NOHELLOTIMEOUT 120000 /* timeout in ms before close the connection
+                               * if server does not receive Hello Message */
+
+static UA_StatusCode
+ServerNetworkLayerTCP_start(UA_ServerNetworkLayer *nl, const UA_Logger *logger,
+                            const UA_String *customHostname) {
+    UA_initialize_architecture_network();
+
+    ServerNetworkLayerTCP *layer = (ServerNetworkLayerTCP *)nl->handle;
+    layer->logger = logger;
+
+    /* Get addrinfo of the server and create server sockets */
+    char hostname[512];
+    if(customHostname->length) {
+        if(customHostname->length >= sizeof(hostname))
+            return UA_STATUSCODE_BADOUTOFMEMORY;
+        memcpy(hostname, customHostname->data, customHostname->length);
+        hostname[customHostname->length] = '\0';
+    }
+    char portno[6];
+    UA_snprintf(portno, 6, "%d", layer->port);
+    struct addrinfo hints;
+    memset(&hints, 0, sizeof hints);
+
+    /* Get the discovery url from the hostname */
+    UA_String du = UA_STRING_NULL;
+    char discoveryUrlBuffer[256];
+    if(customHostname->length) {
+        du.length = (size_t)UA_snprintf(discoveryUrlBuffer, 255, "opc.tcp://%.*s:%d/",
+                                        (int)customHostname->length, customHostname->data,
+                                        layer->port);
+        du.data = (UA_Byte*)discoveryUrlBuffer;
+    } else {
+        char hostnameBuffer[256];
+        if(UA_gethostname(hostnameBuffer, 255) == 0) {
+            du.length = (size_t)UA_snprintf(discoveryUrlBuffer, 255, "opc.tcp://%s:%d/",
+                                            hostnameBuffer, layer->port);
+            du.data = (UA_Byte*)discoveryUrlBuffer;
+        } else {
+            UA_LOG_ERROR(layer->logger, UA_LOGCATEGORY_NETWORK, "Could not get the hostname");
+            return UA_STATUSCODE_BADINTERNALERROR;
+        }
+    }
+    UA_String_copy(&du, &nl->discoveryUrl);
+
+    UA_LOG_INFO(layer->logger, UA_LOGCATEGORY_NETWORK,
+                "TCP network layer listening on %.*s",
+                (int)nl->discoveryUrl.length, nl->discoveryUrl.data);
+    return UA_STATUSCODE_GOOD;
+}
+
+static UA_StatusCode
+ServerNetworkLayerTCP_listen(UA_ServerNetworkLayer *nl, UA_Server *server,
+                             UA_UInt16 timeout) {
+    return UA_STATUSCODE_GOOD;
+}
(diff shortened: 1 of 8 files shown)


src/server/ua_services_securechannel.c
```diff
@@ -34,8 +34,6 @@ removeSecureChannel(UA_Server *server, channel_entry *entry,
 
     /* Detach from the connection and close the connection */
     if(entry->channel.connection) {
-        if(entry->channel.connection->state != UA_CONNECTIONSTATE_CLOSED)
```
JanSurft, line 37: let eventloop handle "connection->close"

src/server/ua_services_securechannel.c
```diff
@@ -34,8 +34,6 @@ removeSecureChannel(UA_Server *server, channel_entry *entry,
...
-            entry->channel.connection->close(entry->channel.connection);
```
JanSurft, line 38: let eventloop handle "connection->close"

src/server/ua_server.c
```diff
@@ -673,8 +657,25 @@ UA_Server_run_iterate(UA_Server *server, UA_Boolean waitInternal) {
...
+            server->config.eventLoop->nextCyclicTime(server->config.eventLoop);
+        UA_DateTime latest = now + (UA_MAXTIMEOUT * UA_DATETIME_MSEC);
+        if(nextRepeated > latest)
+            nextRepeated = latest;
+
+
+        /* round always to upper value to avoid timeout to be set to 0
+        * if(nextRepeated - now) < (UA_DATETIME_MSEC/2) */
+        if(waitInternal)
+            timeout = (UA_UInt16)(((nextRepeated - now) + (UA_DATETIME_MSEC - 1)) / UA_DATETIME_MSEC);
```
JanSurft, line 676: understand "waitInternal" ?

include/open62541/server.h
```diff
@@ -123,6 +123,7 @@ struct UA_ServerConfig {
     /* EventLoop */
     UA_EventLoop *eventLoop;
     UA_Boolean externalEventLoop; /* The EventLoop is not deleted with the config */
+    UA_ConnectionManager *connectionManagers;
```
JanSurft, line 126: yet this just hosts a single connectionManager, namely TCP

src/server/ua_server_config.c
```diff
@@ -28,6 +28,10 @@ UA_ServerConfig_clean(UA_ServerConfig *config) {
 
     /* Custom DataTypes */
     /* nothing to do */
+    for (size_t i = 0; i < config->connectionManagersSize; i++) {
+        UA_ConnectionManager* cm = config->connectionManagers[i];
+        UA_free(cm->initialConnectionContext);
```
jpfr, line 33: Use `el->free`.
https://github.com/open62541/open62541/blob/5bac022548b0ff8167278a042a75c4fcfbdcc47d/include/open62541/plugin/eventloop.h#L98

src/server/ua_server.c
```diff
@@ -744,3 +763,19 @@ UA_Server_run(UA_Server *server, const volatile UA_Boolean *running) {
     }
     return UA_Server_run_shutdown(server);
 }
+
+void
+UA_Server_setupEventLoop(UA_Server *server) {
```
jpfr, line 786: This is used only in the same file.
Make it a static function to reduce the API surface.

src/server/ua_server.c
```diff
@@ -744,3 +763,19 @@ UA_Server_run(UA_Server *server, const volatile UA_Boolean *running) {
...
+    UA_Server_BasicConnectionContext *ctx = (UA_Server_BasicConnectionContext*) UA_malloc(sizeof(UA_Server_BasicConnectionContext));
```
jpfr, line 787: We always test if malloc is a success.
Return a StatusCode if out of memory.

src/server/ua_server.c
```diff
@@ -685,25 +685,44 @@ UA_Server_run_iterate(UA_Server *server, UA_Boolean waitInternal) {
...
+    */
+    for (size_t i = 0; i < server->config.connectionManagersSize; ++i) {
+        UA_ConnectionManager *cm = server->config.connectionManagers[i];
+        if(cm) {
+
+            if (cm->eventSource.eventLoop != NULL) {
+                cm->eventSource.stop(&cm->eventSource);
+                UA_EventLoop *el = server->config.eventLoop;
+                while(cm->eventSource.state != UA_EVENTSOURCESTATE_STOPPED) {
+                    el->run(el, 0);
```
jpfr, line 733: Use a timeout here to prevent spinning.
This can even be UINT32_MAX. Since we don't quite anyway until this is not stopped.

src/server/ua_server.c
```diff
@@ -685,25 +685,44 @@ UA_Server_run_iterate(UA_Server *server, UA_Boolean waitInternal) {
...
+                }
+            }
+        }
+    }
+
+    /* TODO: check why not run el here, will free and close connections thus
```
jpfr, line 739: First set **all** eventsources to `stopping`.
Only then run `el->run` until stopped.

arch/network_tcp.c
```diff
@@ -25,6 +25,142 @@
...
+            nl->statistics->currentConnectionCount--;
+        }
+    }
+
+    /* Free the layer */
+    UA_free(layer);
+}
+
+UA_ServerNetworkLayer
+UA_ServerNetworkLayerTCP(UA_ConnectionConfig config, UA_UInt16 port,
```
jpfr, line 133: This is no longer used, right?
So it can be deleted.

jpfr (commented):
Looking good!
Close the the finish line.

src/server/ua_server_config.c
JanSurft: Does the el->free also free the initialConnectionContext?
And if it does we should maybe overthink that, because it is the responsibility of the caller to allocate memory and assign the initial connection context on a connection manager. The caller may even assign a statically allocated address in the program-text segment which should not be freed at all. Or am I missing something?

arch/network_tcp.c
JanSurft: It is still used, unfortunately the layers and their internal structures that is filled with information are not that loosely coupled to some of the internal functionalities, so for breaking as little as possible of some layer related code (and for changing as little as possible), they will still be created and removed serving as pure storage of information.

src/server/ua_server_config.c
jpfr: You are right.
The initialContext is not touched.
Only the EventSource itself is freed.

arch/network_tcp.c
JanSurft: It is used in the function `UA_ServerConfig_addNetworkLayerTCP` in `plugins/ua_config_default.c` ll. 331 ff when the networklayers get set up:

```c
    conf->networkLayers[conf->networkLayersSize] =
        UA_ServerNetworkLayerTCP(config, portNumber, 0);
    if (!conf->networkLayers[conf->networkLayersSize].handle)
    conf->networkLayersSize++;

```

jpfr: Yes. But you never use it.
It just sits there doing nothing.

jpfr:
Cool!

PR #4906, opened 2022-01-20
fix(services): Null pointer dereference in history service.

mlgiraud:
Enabling the History functionality but leaving it unconfigured leads to a null pointer dereference when calling Service_HistoryRead.


jpfr:
Nice!
I moved the PR to target 1.3.
Can you removed all the commits left over from master?

Can be merged right away then.

mlgiraud:
Done

jpfr:
👍

PR #4908, opened 2022-01-21
Some const qualifiers which causes build with older openssl (pre 1.1) to fail.

kjeldflarup:
While updating to latest master, I encountered passing of some const variables has changed.  
In the older tool chains I use, the functions in openssl does not have a const in the declaration, which causes a violation.

I'm actually in the process of migrating to openssl 1.1.1, so this will be less important for me soon.


jpfr:
Nice!

PR #4909, opened 2022-01-21
error: ‘PTHREAD_MUTEX_RECURSIVE’ undeclared (first use in this functi…

kjeldflarup:
Got this error when compiling on plain Ubuntu 20.04
```
error: ‘PTHREAD_MUTEX_RECURSIVE’ undeclared (first use in this function); did you mean PTHREAD_MUTEX_RECURSIVE_NP’
```

PTHREAD_MUTEX_RECURSIVE requires some "Linux 98" define, which is not standard anymore

Problem may however be old build chains, which do not know PTHREAD_MUTEX_RECURSIVE_NP


jpfr:
I think we can remove the line for the recursive mutex setting entirely.
We have a separate assert that the mutex shall not be taken twice anyway.

kjeldflarup:
I made a lookup of PTHREAD_MUTEX_RECURSIVE_NP:
> As the NP suffix indicates, this is a nonportable extension to the POSIX standard and should not be employed in portable programs.

It sound like a very good idea to get rid of this line. 

I have however not worked with mutex'es in open62541, so I cannot say whether it is necessary or not, and it is not possible to test our way out of this. It requires knowlegde.

This name change just makes it clearer, that there is a problem that needs to be addressed.

PR #4916, opened 2022-01-31
[Review] feat(server) automatic nodeset injector

andreasebner:
This WIP-PR development branch is targeting a new mechanism for automatic nodeset/companion spec injection into the server during the server-init. This feature will the user allow to specify the required information models in the cmake configuration and the server will integrate them automatically.

Current Status:
- [x] Extend code generation with nodeset load logic
- [x] Extend cmake nodeset compiler macros with a new option for autoload
- [ ] Add companion spec specific cmake falgs (e.g. UA_INFORMATION_MODEL_IO_LINK ON)
- [ ] Use mechansim to compile alle ua-schema companion specifications in our unit tests


jpfr:
The injector idea is cool!
On the negative side, I don’t like the ever increasing number of cmake flags.
We are at a point that makes testing all combinations of interacting cmake settings impossible.

How about we use a semi-colon separated list of companion specs? Like we do for the architectures. Then we can keep the injector logic in one place. Instead of splattering it all across many sections of the cmake files.

GoetzGoerisch:
@andreasebner is there a way that the injector could automatically resolve the nodeset dependencies? Because often one wants to use one companion spec but needs the required as well.

If this could be handled automatically, this would be great.
Should I write a user story?

doc/nodeset_compiler.rst
```diff
@@ -476,3 +476,29 @@ Finally you need to include all these files in your build process and call the c
     }
 
     retval = UA_Server_run(server, &running);
+
+
+Automatic Nodesetinjection
```
andreasebner, line 481: Space needed.

doc/nodeset_compiler.rst
```diff
@@ -476,3 +476,29 @@ Finally you need to include all these files in your build process and call the c
...
+................................
+
+The nodesetinjector is a mechanism for automatically loading nodeset/companion specifications during server initialization.
+It provides a fast and easy way to load nodesets in all applications, focusing on the official OPCFoundation/UANodeset ( https://github.com/OPCFoundation/UA-Nodeset ).
```
andreasebner, line 485: on the official .. Repository.

doc/nodeset_compiler.rst
```diff
@@ -476,3 +476,29 @@ Finally you need to include all these files in your build process and call the c
...
+    -DCMAKE_BUILD_TYPE=Debug
+    -DUA_BUILD_EXAMPLES=ON
+    -DUA_INFORMATION_MODEL_AUTOLOAD=DI;POWERLINK;PROFINET;MachineVision
+    -DUA_NAMESPACE_ZERO=FULL
+
+The order of nodesets is important! Nodesets that build on other nodesets must be placed after them in the list.
+The following nodesets are currently supported.
+
+* DI, ISA95-JOBCONTROL, OpenSCS, DEXPI, AMB, AutoID, POWERLINK, IA, Machinery, PackML, PNEM, PLCopen, MachineTool, PROFINET, MachineVision, FDT, CommercialKitchenEquipment,
+PNRIO, Scales, Weihenstephan, Pumps, CAS, TMC, IJT
```
andreasebner, line 504: I think the list is currently not handled in the cmake.

examples/nodeset/server_nodeset_plcopen.c
```diff
@@ -29,18 +29,30 @@ int main(int argc, char** argv) {
...
 
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+
     /* create nodes from nodeset */
-    UA_StatusCode retval = namespace_di_generated(server);
-    if(retval != UA_STATUSCODE_GOOD) {
-        UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "Adding the DI namespace failed. Please check previous error output.");
-        UA_Server_delete(server);
-        return EXIT_FAILURE;
+    size_t idx = LONG_MAX;
```
andreasebner, line 35: You should Use the opc ua max and min macros.

examples/nodeset/server_nodeset_plcopen.c
```diff
@@ -29,18 +29,30 @@ int main(int argc, char** argv) {
...
+    UA_Server_getNamespaceByName(server, UA_STRING("http://opcfoundation.org/UA/DI/"), &idx);
+    if(idx == LONG_MAX) {
```
andreasebner, line 37: Very complex. Avoid the variable and use getNamespaceByName directly in the if condition. (same to other examples).

Issue #3964, opened 2020-10-19
Bug(NodeSet Compiler): Values in NodeSet (enums & optionset) 


Fischer-Thomas:
## Description
Currently there is a bug when trying to compile a nodeset which includes nodes with values of custom OptionSets and enumerations. 
For enumerations the code generator tries to assign the string from the nodeset (for example **NanoSecond_0** instead of just using **0 as value** or the generated enumValue **UA_CUSTOMUNITENUM_NANOSECOND**).
And for the OptionSet it uses the wrong structure. (it doesn't use the members **Value** or **ValidBits** for assigning the values) 
Also the two values that should be assigned are generated as global variables inside the file with **the same name**.
I attached all relevant files i used to the issue.
 
[customNodeSet.zip](https://github.com/open62541/open62541/files/5401185/customNodeSet.zip)

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag):master
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [x] Self-contained code example attached
 - [ ] Critical issue

NoelGraf:
Does the problem still exist? Have you tested it with the current 1.2 version?

Fischer-Thomas:
The problem still exists. I tested it with the current master **and** 1.2 version.

Here are some screenshots:

The added object using the custom DataType.
![image](https://user-images.githubusercontent.com/45487937/115723856-ac7dd080-a380-11eb-87bb-780d566a0fe8.png)

Generated code: (OptionSet errorr)
![image](https://user-images.githubusercontent.com/45487937/115724886-a0464300-a381-11eb-8f20-64074f0214d9.png)
![image](https://user-images.githubusercontent.com/45487937/115724959-b5bb6d00-a381-11eb-9708-1d7edbe192fd.png)

Generated code: (Enumeration error)
![image](https://user-images.githubusercontent.com/45487937/115725008-bfdd6b80-a381-11eb-858c-1a833cf127c6.png)

jpfr:
Is this done with the latest changes to the nodeset compiler @andreasebner ?

andreasebner:
Should be solved by the merged PR's or the open #4715 @NoelGraf can you confirm?

andreasebner:
[...]

Issue #4901, opened 2022-01-17
Nodeset Compiler Issue: Value of a Union.


MurzynJ:
## Description
The generated code doesn't initialize the value according to the content of the XML.

Test code (including a sample nodeset):
https://github.com/MurzynJ/open62541/tree/NcIssue_UnionValue

Example:
![image](https://user-images.githubusercontent.com/70644594/149770021-891e7921-d7f5-4c1b-8092-a55bd6a067cf.png)
![image](https://user-images.githubusercontent.com/70644594/149770088-e23ac9c4-152a-48a1-8bec-09c4267633dc.png)

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): Latest master: b061554990ab38cd9062212aaa69c322f1415f33
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

MurzynJ:
This issue is still open, even after the latest commits by @NoelGraf. Value of a Union is not set in the generated code.

![image](https://user-images.githubusercontent.com/70644594/152380414-5e619a2e-65a6-468d-b93a-5e48eb0bf84d.png)

FYI @Fischer-Thomas

andreasebner:
@MurzynJ The issue should be solved, can we close it?

MurzynJ:
Sorry, I haven't had time to check the fix yet. But I trust that it's working. If I find anything in future, I will just create another issue.

PR #4934, opened 2022-02-07
[WIP] Fix Nodeset-Compiler code generation

NoelGraf:
This pull reguest fixes the issues #3964 and #4901


jpfr:
Nice.
The nodeset compiler update was also merged to 1.3.
You can make this a PR against 1.3 and we "go down the chain" instead of backporting.

PR #4940, opened 2022-02-10
feat(pubsub) Create tutorial_pubsub_mqtt_publish_subscribe_with_encryption.c

jackybek:


jpfr:
Wow, this is a huge commit.
But bigger is not always better.
4k lines are a bit much.

Can you reduce to show only the essentials of pubsub?
For example, we don't need to publish more than 1-2 variables to showcase the essence.

andreasebner:
Hi @jackybek, thanks for this contribution. We are currently on a overhaul of the MQTT-C integration, see #5086 Your example code is quite long and we will try to make it smaller and integrate it in our stack after #5086 is merged.

PR #4948, opened 2022-02-14
feat(pubsub): Recognize the PDS name

MurzynJ:
Part 14 requires that a PDS has a unique name within the Publisher. Make the corresponding check before creating a new PDS.
image
image

The new function `UA_PublishedDataSet_findPDSbyName` may be also used in future by new applications e.g. OPC UA FX.

By the way, added utility function `UA_String_isEmpty` to the core library. If such an addition is not appropriate, I can replace the check in ua_pubsub_manager.c with:
`UA_String_equal(&publishedDataSetConfig->name, &UA_STRING_NULL)`, or `(publishedDataSetConfig->name.length == 0)` like it's done elsewhere.

### DIFF
--- a/src/pubsub/ua_pubsub_manager.c
+++ b/src/pubsub/ua_pubsub_manager.c
@@ -160,12 +160,16 @@ UA_Server_addPublishedDataSet(UA_Server *server,
     if(UA_String_isEmpty(&publishedDataSetConfig->name))
     {
         // DataSet has to have a valid name
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                     "PublishedDataSet creation failed. Invalid name.");
         return result;
     }
 
     if(UA_PublishedDataSet_findPDSbyName(server, publishedDataSetConfig->name))
     {
         // DataSet name has to be unique in the publisher
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                     "PublishedDataSet creation failed. DataSet with the same name already exists.");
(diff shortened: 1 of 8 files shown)


src/pubsub/ua_pubsub_manager.c
```diff
@@ -157,6 +157,19 @@ UA_Server_addPublishedDataSet(UA_Server *server,
         return result;
     }
 
+    if(UA_String_isEmpty(&publishedDataSetConfig->name))
+    {
+        // DataSet has to have a valid name
```
jpfr, line 162: Please add a logging message here.
Otherwise users will take a long time to debug.

src/pubsub/ua_pubsub_manager.c
```diff
@@ -157,6 +157,19 @@ UA_Server_addPublishedDataSet(UA_Server *server,
...
+        return result;
+    }
+
+    if(UA_PublishedDataSet_findPDSbyName(server, publishedDataSetConfig->name))
+    {
+        // DataSet name has to be unique in the publisher
```
jpfr, line 168: Please add a logging message here.
Otherwise users will take a long time to debug.

jpfr (commented):
Looks great. 👍 
One request: We should an error if adding the PDS fails with the new checks.
Otherwise people will wonder why their old code suddenly breaks.

src/pubsub/ua_pubsub_manager.c
MurzynJ: Done.

jpfr:
Nice! Merged.

PR #4955, opened 2022-02-16
feat(server): Allow invoking ObjectType's method on its instance

MichalTyndel:
Allow situation, where the objectType's method is invoked on its instance (or instance of its subtype) - the object's objectType (or it's supertype) is the source of HasComponent reference.

Zrzut ekranu 2022-02-16 092540

### DIFF
--- a/src/server/ua_services_method.c
+++ b/src/server/ua_services_method.c
@@ -125,6 +125,7 @@ static const UA_NodeId hasComponentNodeId = {0, UA_NODEIDTYPE_NUMERIC, {UA_NS0ID
 static const UA_NodeId organizedByNodeId = {0, UA_NODEIDTYPE_NUMERIC, {UA_NS0ID_ORGANIZES}};
 static const UA_String namespaceDiModel = UA_STRING_STATIC("http://opcfoundation.org/UA/DI/");
 static const UA_NodeId hasTypeDefinitionNodeId = {0, UA_NODEIDTYPE_NUMERIC, {UA_NS0ID_HASTYPEDEFINITION}};
+static const UA_NodeId hasSubTypeNodeId = {0, UA_NODEIDTYPE_NUMERIC, {UA_NS0ID_HASSUBTYPE}};
 // ns=0 will be replace dynamically. DI-Spec. 1.01: <UAObjectType NodeId="ns=1;i=1005" BrowseName="1:FunctionalGroupType">
 static UA_NodeId functionGroupNodeId = {0, UA_NODEIDTYPE_NUMERIC, {1005}};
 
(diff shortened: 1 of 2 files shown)


Fischer-Thomas:
Really cool feature!
@jpfr @andreasebner

mlang-de:
Imho this should be part of v1.3 since for instance the UaExpert uses this for the CertificateManagement Push methods.

tests/server/check_services_call.c
```diff
@@ -230,6 +235,31 @@ START_TEST(callMethodWithEmptyArgument) {
 #endif
 } END_TEST
 
+START_TEST(callObjectTypeMethodOnInstance) {
+/* Minimal nodeset does not add any method nodes we may call here */
+#if defined(UA_GENERATED_NAMESPACE_ZERO) && defined(UA_ENABLE_SUBSCRIPTIONS)
```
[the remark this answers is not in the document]
MichalTyndel: You are right, thanks for help!

src/server/ua_services_method.c
```diff
@@ -178,6 +179,95 @@ callWithMethodAndObject(UA_Server *server, UA_Session *session,
             }
         }
     }
+    if(!found) {
+        /* If the object doesn't have a hasComponent reference to the method node,
+         * check its objectType (and its supertypes). Invoked method can be a component
+         * of objectType and be invoked on this objectType's instance (or on a instance
+         * of one of its subtypes). */
+
+        UA_ReferenceTypeSet hasTypeDefinitionRefs;
```
jpfr, line 188: You can use the existing
```c
const UA_Node *
getNodeType(UA_Server *server, const UA_NodeHead *head);
```

To get the node type. This will be the ObjectType if it is an object node.

src/server/ua_services_method.c
```diff
@@ -178,6 +179,95 @@ callWithMethodAndObject(UA_Server *server, UA_Session *session,
...
+
+        UA_ReferenceTypeSet hasSubTypeRefs;
+        result->statusCode =
+            referenceTypeIndices(server, &hasSubTypeNodeId,
+                                 &hasSubTypeRefs, true);
+        if(result->statusCode != UA_STATUSCODE_GOOD)
+            return;
+
+
+        while(objectType && !found) {
```
jpfr, line 228: Use `UA_NodeReferenceKind_findTarget` for this.
This can be much faster for nodes with many children.
Because we directly search in the binary tree instead of iterating.

src/server/ua_services_method.c
```diff
@@ -178,6 +179,95 @@ callWithMethodAndObject(UA_Server *server, UA_Session *session,
...
+                        found = true;
+                        break;
+                    }
+                }
+            }
+            if(found)
+                break;
+            
+            /* Reference was not found - next iteration check supertype */
+            const UA_Node *superType = NULL;
```
jpfr, line 248: You can use the same `getNodeType` to get the super type.

jpfr (commented):
Looks great!
There is some potential to reuse internal functions.
Have a look at the detailed comments.

Looking forward to this.

src/server/ua_services_method.c
MichalTyndel: Thanks, somehow I missed this function

src/server/ua_services_method.c
MichalTyndel: Done as suggested

MichalTyndel:
@jpfr  Thanks for tips! neweset version uses the functions you suggested.

andreasebner:
I reviewed your changes and it looks great now. Thank you for the contribution.

PR #4957, opened 2022-02-16
fix(nc): check namespace when searching for HasTypeDefinition reference

MichalTyndel:
References with i=40 can exist in other namespaces, so the namespace index needs to be checked when searching for HasTypeDefinition reference.


NoelGraf (commented):
Adjustment of the check when searching for HasTypeDefinition references.

andreasebner:
Thank you for the contribution. This PR (using also your original commit) is updated in #5091.

PR #4963, opened 2022-02-17
Extended Event-Filter select clause validation

andreasebner:
The PR introduces some additional checks for the select clause, defined in Part 4 Table 144.

### DIFF
--- a/src/server/ua_services_view.c
+++ b/src/server/ua_services_view.c
@@ -1250,10 +1250,7 @@ browseSimplifiedBrowsePath(UA_Server *server, const UA_NodeId origin,
     bp.relativePath.elementsSize = browsePathSize;
 
     /* Browse */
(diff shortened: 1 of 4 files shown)


src/server/ua_services_view.c
```diff
@@ -1250,7 +1250,7 @@ browseSimplifiedBrowsePath(UA_Server *server, const UA_NodeId origin,
     bp.relativePath.elementsSize = browsePathSize;
 
     /* Browse */
-    UA_UInt32 nodeClassMask = UA_NODECLASS_OBJECT | UA_NODECLASS_VARIABLE;
```
jpfr, line 1253: The is is somewhat arbitrary... We could simplfy allow any node class here.
So just `nodeClassMask = ~0;`.

andreasebner: I'am fine with taken all Nodeclasses into account. I will update the PR now.

PR #4965, opened 2022-02-18
fix(server): Decrease logging severity in isConditionOrBranch

ccvca:
Do not log `"Entry not found in list!"` each time when a event is triggered:

As `isConditionOrBranch` is called for all events here:
https://github.com/open62541/open62541/blob/5f5214427b62e96e066270f4c2798f68b1375213/src/server/ua_subscription_events.c#L285-L295

There should be no UA_LOG_ERROR each time a event is triggered.


jpfr:
👍

PR #4976, opened 2022-02-21
Fix for socket resource leak (out of socket)

DirkHeinrich:
On our test system we detected a socket resource leak if the following preconditions are met:
1. OPC UA server is behind a gateway.
2. OPC UA server is down.
3. OPC UA client tries to connect to OPC UA server.

Observed behavior:
After each connect operation an additional file descriptor is being added to the list of file descriptors (/proc/self/fd).

The provided patch fixes the problem.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -1107,8 +1107,12 @@ connectSync(UA_Client *client) {
         if(client->noSession && client->channel.state == UA_SECURECHANNELSTATE_OPEN)
             break;
         now = UA_DateTime_nowMonotonic();
-        if(maxDate < now)
-            return UA_STATUSCODE_BADTIMEOUT;
+        if(maxDate < now) {
+            /* set maxDate to now so timeout will be detected when calling
+	     * UA_Client_run_iterate
+	     */
+            maxDate = now;
+        }
         retval = UA_Client_run_iterate(client,
                                        (UA_UInt32)((maxDate - now) / UA_DATETIME_MSEC));
     }
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -1109,8 +1109,8 @@ connectSync(UA_Client *client) {
         now = UA_DateTime_nowMonotonic();
         if(maxDate < now) {
             /* set maxDate to now so timeout will be detected when calling
-	     * UA_Client_run_iterate
-	     */
+             * UA_Client_run_iterate
+             */
             maxDate = now;
         }
         retval = UA_Client_run_iterate(client,


jpfr:
Thanks for this.
That patch would work. But it makes the logic very indirect and dependent on different places in the code.
It is not directly visible what would happen next.

Here's a code snippet adapted from `initConnect` that would handle the error locally and clean up the resources.

```c
    if(maxDate < now) {
        UA_LOG_WARNING(&client->config.logger, UA_LOGCATEGORY_CLIENT,
                       "Could not open a SecureChannel to %.*s within the delay",
                       (int)client->endpointUrl.length, client->endpointUrl.data);
        client->connectStatus = UA_STATUSCODE_BADCONNECTIONCLOSED;
        closeSecureChannel(client);
        return client->connectStatus;
    }   
```
(That code is untested.)

But please check whether this the intended behavior.
Or whether the return code is not an "abort" for the connection and we are expected to retry the `connectSync` function again with an additional timeout.

DirkHeinrich:
> Thanks for this. That patch would work. But it makes the logic very indirect and dependent on different places in the code. It is not directly visible what would happen next.
> 
> Here's a code snippet adapted from `initConnect` that would handle the error locally and clean up the resources.
> 
> ```c
>     if(maxDate < now) {
>         UA_LOG_WARNING(&client->config.logger, UA_LOGCATEGORY_CLIENT,
>                        "Could not open a SecureChannel to %.*s within the delay",
>                        (int)client->endpointUrl.length, client->endpointUrl.data);
>         client->connectStatus = UA_STATUSCODE_BADCONNECTIONCLOSED;
>         closeSecureChannel(client);
>         return client->connectStatus;
>     }   
> ```
> 
> (That code is untested.)
> 
> But please check whether this the intended behavior. Or whether the return code is not an "abort" for the connection and we are expected to retry the `connectSync` function again with an additional timeout.

Your patch was my first idea to solve the problem but it did not work. The major issue of this problem is that even calling UA_disconnect does not work (which also calls closeSecureChannel), so applications run also into trouble if they react to this failure by calling UA_disconnect.

The reason why the patch does not work is that the channel does not get attached to the connection object (of the client object) until we left the state UA_CONNECTIONSTATE_OPENING (see code snippet: https://github.com/open62541/open62541/blob/ba379f66d98166a19e2374319c739780472f652a/src/client/ua_client_connect.c).

line: 904ff.
```
    /* Poll the connection status */
    if(client->connection.state == UA_CONNECTIONSTATE_OPENING) {
        client->connectStatus =
            client->config.pollConnectionFunc(&client->connection, timeout,
                                              &client->config.logger);
        return client->connectStatus;
    }

    /* Attach the connection to the SecureChannel */
    if(!client->channel.connection)
        UA_Connection_attachSecureChannel(&client->connection, &client->channel);
```

jpfr:
Fixed via #5006.

PR #4978, opened 2022-02-23
feat(server) Add internal API function for Namespace Lookup

andreasebner:
This PR introduces the counterpart function of the existing "getNamespaceByName" function. The Namespace String can be obtained by the namespace id.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -47,7 +47,7 @@
  * Application URI.
  *
  * This is done as soon as the Namespace Array is read or written via node value
- * read / write services, or UA_Server_addNamespace,
+ * read / write services, or UA_Server_addNamespace, or UA_Server_getNamespaceByIndex
  * UA_Server_getNamespaceByName or UA_Server_run_startup is called.
  *
  * Therefore one has to set the custom NS1 URI before one of the previously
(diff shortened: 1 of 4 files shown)


src/server/ua_server.c
```diff
@@ -128,6 +141,15 @@ UA_Server_getNamespaceByName(UA_Server *server, const UA_String namespaceUri,
     return res;
 }
 
+UA_StatusCode
+UA_Server_getNamespaceByIndex(UA_Server *server, const size_t namespaceIndex,
```
Fischer-Thomas, line 145: The signature of this new method looks like it should be in the external api (include/open62541/server.h).
However it isn't added there or am i missing something?

src/server/ua_server.c
```diff
@@ -119,6 +119,19 @@ getNamespaceByName(UA_Server *server, const UA_String namespaceUri,
...
 
+UA_StatusCode
+getNamespaceByIndex(UA_Server *server, const size_t namespaceIndex,
+                   UA_String *foundUri) {
+    /* ensure that the uri for ns1 is set up from the app description */
+    setupNs1Uri(server);
+    UA_StatusCode res = UA_STATUSCODE_BADNOTFOUND;
+    if(namespaceIndex > server->namespacesSize)
+        return res;
+    UA_String_init(foundUri);
```
ccvca, line 130: Is "_init" the way to go? In my opinion,  `_clear` would make more sense:
```
UA_String str = UA_STRING_NULL;
for(uint16_t i = 0 ; i  < nsMax; ++i)
{
getNamespaceByIndex(pServer, i, &str);
/// Do seomething with str
}
UA_String_clear(&str);
```

Issue #4964, opened 2022-02-18
Win32 Shutdown takes 120s when (unresponsive?) clients connected


ccvca:
## Description
When `running` is set to false, `UA_Server_run` takes ~120s to return, when a Client is still connected.

## Background Information / Reproduction Steps
The error was discovered here, after upgrading this test to the new open62541-Version with the event-loop:
https://github.com/umati/Sample-Server/blob/690cb0e590eeaaf3ee4c1e085995944b1b785fb4/tests/Test_BindWrongDatatype.cpp

- Start Server
- Connect Client
- Do some work
- Do not disconnect the client, do not call run_iterate (do not use the client anymore)
- Set running of `UA_Server_run` to false
- Wait ~120s until the Server is shut down

## Debugging

https://github.com/open62541/open62541/blob/5f5214427b62e96e066270f4c2798f68b1375213/arch/eventloop_posix_tcp.c#L814-L819
- The listening-Sockets are closed fast, as `TCP_close` is used 
- The client connections take the `TCP_shutdownConnection` path.

https://github.com/open62541/open62541/blob/5f5214427b62e96e066270f4c2798f68b1375213/arch/eventloop_posix_tcp.c#L503

- `UA_shutdown((UA_FD)connectionId, SD_BOTH)` (win32 `shutdown` - function is used)

(120s) gap

https://github.com/open62541/open62541/blob/5f5214427b62e96e066270f4c2798f68b1375213/arch/eventloop_posix_select.c#L162

 -  A new receive-event is issued and as there is an error, the connection is finally closed

https://github.com/open62541/open62541/blob/5f5214427b62e96e066270f4c2798f68b1375213/arch/eventloop_posix_tcp.c#L503

According to https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown the shutdown function is not used as expected in the library. As `SD_BOTH` is passed to  `shutdown`, any subsequent call to `recv` is disallowed (-1 is returned).

The proposed sequence is instead (taken from the [doc](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown)):

1. Call shutdown with how=SD_SEND.
[...]

PR #5003, opened 2022-03-11
WIP: fix(nc): add empty Extention Object if datatype is not in namespace

Kantiran91:
Hi,

I implement an adaption of the  Release Candidate of the OPC 40001-101 - UA CS for Machinery Part 101 - Result Transfer.
My modification makes the nodeset v1.04 compatible (UriString -> String, Change Required Nodeset). The nodesets can be found [here](https://github.com/umati/Sample-Server/tree/dev_gms/model/GMS)

My second nodeset use the Result Transfer and crashed because nodeset generation process. I could fix this by returning an empty Extension Object if the enc is None. This solves the problem for me.

### DIFF
--- a/tools/nodeset_compiler/datatypes.py
+++ b/tools/nodeset_compiler/datatypes.py
@@ -242,6 +242,10 @@ class Value(object):
 
                 extobj.value = []
                 members = enc.members
+                if enc is None:
+                    logger.warning("DataType is not found in nodeset")
+                    return ExtensionObject()
+                for e in enc.members:
 
                 # The EncodingMask must be skipped.
                 if ebodypart.localName == "EncodingMask":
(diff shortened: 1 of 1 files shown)


GoetzGoerisch:
@jpfr any chance we can get this merged shortly?

jpfr:
Hey @NoelGraf , please review this. Thanks.

NoelGraf:
Hi @Kantiran91, i have tried to recreate the problem. Are there problems when generating the nodeset or only when using it? Because I could generate your gms nodeset without any issues.

GoetzGoerisch:
@NoelGraf when using it.
@Kantiran91 I just made a branch copy without the fix but with an updated open62541, which works.
Beside the failing test, which are due to #4964 


Build with fix: https://github.com/umati/Sample-Server/commits/dev_gms
Build without fix: https://github.com/umati/Sample-Server/tree/goetz/gms

Kantiran91:
Sorry, I was on holiday. 
@GoetzGoerisch: This error should be fixed with the commit 0f5bc8fa0ea0603798b66bd896f5091a5020cc6b
If run this current master db19f990d16e3ecac3934d67200d86d88bf49031 and I also cannot recreate the issue. So,  In my opinion, this PR can be closed.

Issue #3221, opened 2019-10-29
TCP client socket file descriptors leaking


mtenberge:
Probably related to #3218.

## Description
In certain situations, the TCP socket file descriptor is not closed. If the application tries to (re)connect in a loop (probably with a sleep delay), the application will eventually run out of resources, as a normal application is limited to max 1024 open files.

## Background Information / Reproduction Steps
In function `UA_Client_connectTCPSecureChannel`, if the call to `client->config.connectionFunc` succeeds, but function `HelAckHandshake` fails (see #3218 for an example situation in which this can occur), `UA_Client_disconnect` is called (both in `HelAckHandshake` and in the cleanup section of `UA_Client_connectTCPSecureChannel`).

`client->connection.state` is however still set to UA_CONNECTION_OPENING, so `UA_Client_disconnect` does not call the 'client->connection.close' function.

As a result, the socket fd is not closed (as can be seen using e.g. `lsof`).

When an error occurs, my application deletes the client (`UA_Client_delete`), sleeps a while, and tries again with a new client instance (`UA_Client_new`, `UA_ClientConfig_setDefault` and `UA_Client_connect`).

Used CMake options:
See #3218

## Logs
(note: log entries that occur at (almost) the same time may be reordered by my current logging plugin)
```
INFO[0039] Connecting to endpoint opc.tcp://10.128.49.101:4840  category="OPC-UA client"
INFO[0039] SecurityPolicy not specified -> use default #None  category="OPC-UA client"
WARN[0039] Security policy None is used to create SecureChannel. Accepting all certificates  category="OPC-UA security policy"
INFO[0039] TCP connection established                    category="OPC-UA client"
ERRO[0044] Couldn't connect the client to a TCP secure channel  category="OPC-UA client"
ERRO[0044] Receiving ACK message failed with GoodNonCriticalTimeout  category="OPC-UA network"
ERRO[0044] HEL/ACK handshake failed                      category="OPC-UA client"
<sleep delay>
WARN[0049] Security policy None is used to create SecureChannel. Accepting all certificates  category="OPC-UA security policy"
INFO[0049] Connecting to endpoint opc.tcp://10.128.49.101:4840  category="OPC-UA client"
INFO[0049] SecurityPolicy not specified -> use default #None  category="OPC-UA client"
INFO[0049] TCP connection established                    category="OPC-UA client"
ERRO[0054] Couldn't connect the client to a TCP secure channel  category="OPC-UA client"
[...]

PR #5006, opened 2022-03-13
fix(client): handle timeout when connecting

informatikr:
This change adds code to handle a timeout of UA_select when the client tries to connect.

Previously, the return code for a timeout was unchecked, and instead handled as if successful. This caused issue #3221 .

Merging this change would also make pull request #4976 obsolete.


jpfr:
Nice! Switched to the 1.3 branch.
Will merge when ready.

PR #5008, opened 2022-03-14
[REVIEW] fix(el): win32 shutdown takes 120 s

### LINKED ISSUE #4964: Win32 Shutdown takes 120s when (unresponsive?) clients connected

JanSurft:
resolves #4964 

This fix resolves 120 sek timeout for win32 shutdown of tcp connections.

### HOW:

1. calls shutdown "SD_SEND" for connections to be shut down
2. adds a delayed callback that closes a tcp connection after a peek to recv showed no data on the socket
3. if there is peeked data on the socket the delayed callback adds itself again and waits another cycle of the eventloop to process the data that arrived on the socket

### WHY:

As mentioned in the according issue by @ccvca, according to win32 documentation this is the correct way of shutting down a socket gracefully:

> 1. Call shutdown with how=SD_SEND.
> 2. Call [recv](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-recv) or [WSARecv](https://docs.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-wsarecv) until the function completes with success and indicates zero bytes were received. If SOCKET_ERROR is returned, then the graceful disconnect is not possible.
> 3. Call [closesocket](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-closesocket).

https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown#remarks

### TODO:

- [x] Check if VScode CI executes unit tests
- [x] write test(s) for code coverage pass

### DIFF
--- a/arch/eventloop_posix_tcp.c
+++ b/arch/eventloop_posix_tcp.c
@@ -490,6 +490,69 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
     return UA_STATUSCODE_GOOD;
 }
 
+#ifdef _WIN32
+static UA_RegisteredFD *
+findRegisteredFD(TCPConnectionManager *cm, uintptr_t connectionId) {
+    UA_RegisteredFD *rfd;
+    LIST_FOREACH(rfd, &cm->fds, es_pointers) {
+        if(rfd->fd == connectionId) return rfd;
+    }
+    return NULL;
+}
+
+static UA_DelayedCallback
+*createExplicitlyDelayedShutdownCallback(UA_ConnectionManager *cm, UA_RegisteredFD *rfd, UA_Callback callback) {
+    UA_DelayedCallback *explicitlyDelayedShutdownCallback = (UA_DelayedCallback*) UA_malloc(sizeof(UA_DelayedCallback));
+    if(!explicitlyDelayedShutdownCallback) {
+        return NULL;
+    }
+    explicitlyDelayedShutdownCallback->callback = callback;
+    explicitlyDelayedShutdownCallback->application = cm;
+    explicitlyDelayedShutdownCallback->data = (void*) rfd->fd;
+
+    return explicitlyDelayedShutdownCallback;
+}
+
+static UA_INLINE void
+createAndAddDelayedShutdownCallback(UA_ConnectionManager *cm, UA_RegisteredFD *rfd, UA_Callback callback) {
+    UA_DelayedCallback *explicitlyDelayedShutdownCallback = createExplicitlyDelayedShutdownCallback(cm, rfd, callback);
+    if(!explicitlyDelayedShutdownCallback) {
+        UA_LOG_SOCKET_ERRNO_WRAP(
+            UA_LOG_WARNING(cm->eventSource.eventLoop->logger,
+                           UA_LOGCATEGORY_NETWORK,
+                           "TCP %u\t| Warning: Out of memory for creating delayed shutdown callback (%s)",
+                           (unsigned)rfd->fd, errno_str));
+        return;
+    }
+    UA_EventLoop *el = cm->eventSource.eventLoop;
+    el->addDelayedCallback(el, explicitlyDelayedShutdownCallback);
+}
+
+static void
+delayedShutdownCallback(void* application, void* data) {
+
+    UA_ConnectionManager *cm = (UA_ConnectionManager*) application;
+    TCPConnectionManager *tcm = (TCPConnectionManager*)cm;
+    UA_FD fd = (UA_FD) data;
+
+    UA_RegisteredFD* rfd = findRegisteredFD(tcm, fd);
+    if(rfd == NULL) {
+        return;
+    }
+    UA_ByteString response = tcm->rxBuffer;
+    int ret = UA_recv(fd, (char *)response.data, response.length, MSG_DONTWAIT | MSG_PEEK);
+    if(ret == 0 || ret == -1) {
+        cm->connectionCallback(cm, (uintptr_t)fd, &rfd->context,
+                               UA_STATUSCODE_BADCONNECTIONCLOSED,
+                               0, NULL, UA_BYTESTRING_NULL);
+        TCP_close(tcm, rfd);
+    } else {
+        /* add the delayed shutdown callback again because the peeked data will be processed in the next el cycle */
+        createAndAddDelayedShutdownCallback(cm, rfd, delayedShutdownCallback);
+    }
+}
+#endif
+
 static UA_StatusCode
 TCP_shutdownConnection(UA_ConnectionManager *cm, uintptr_t connectionId) {
     UA_LOG_DEBUG(cm->eventSource.eventLoop->logger,
@@ -500,7 +563,11 @@ TCP_shutdownConnection(UA_ConnectionManager *cm, uintptr_t connectionId) {
 #ifndef _WIN32
     int res = UA_shutdown((UA_FD)connectionId, SHUT_RDWR);
 #else
-    int res = UA_shutdown((UA_FD)connectionId, SD_BOTH);
+    int res = UA_shutdown((UA_FD)connectionId, SD_SEND);
+    TCPConnectionManager *tcm = (TCPConnectionManager*)cm;
+    UA_RegisteredFD* rfd = findRegisteredFD(tcm, (UA_FD) connectionId);
+
+    createAndAddDelayedShutdownCallback(cm, rfd, delayedShutdownCallback);
 #endif
     if(res != 0) {
         UA_LOG_SOCKET_ERRNO_WRAP(


GoetzGoerisch:
Thanks @JanSurft!
Working on our end: https://github.com/umati/Sample-Server/actions/runs/1979790663

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,66 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
     return UA_STATUSCODE_GOOD;
 }
 
+#ifdef _WIN32
+static
+UA_RegisteredFD *
```
jpfr, line 495: In all the other functions, we say
```c
static UA_RegisteredFD *
findRegisteredFD(TCPConnectionManager *cm, uintptr_t connectionId) {
```

Please keep with the established code formatting.

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,66 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
...
+findRegisteredFD(TCPConnectionManager *cm, uintptr_t connectionId) {
+    UA_RegisteredFD *rfd, *rfd_tmp;
+    LIST_FOREACH_SAFE(rfd, &cm->fds, es_pointers, rfd_tmp) {
+
+        if (rfd->fd == connectionId) {
```
jpfr, line 500: No space after `if`.
For simple single-line `if`, leave out the brackets.

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,66 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
...
+    int ret = UA_recv(rfd->fd, (char *)response.data, response.length, MSG_DONTWAIT | MSG_PEEK);
+    if(ret == 0 || ret == -1) {
+        cm->connectionCallback(cm, (uintptr_t)rfd->fd, &rfd->context,
+                               UA_STATUSCODE_BADCONNECTIONCLOSED,
+                               0, NULL, UA_BYTESTRING_NULL);
+        TCP_close(tcm, rfd);
+    } else {
+
+        /* add the delayed shutdown callback again because the peeked data will be processed in the next el cycle */
+        UA_DelayedCallback *explicitlyDelayedShutdownCallback = createExplicitlyDelayedShutdownCallback(cm, rfd, delayedShutdownCallback);
```
jpfr, line 536: Move logging and `addDelayedCallback` into `createExplicitlyDelayedShutdownCallback`.
Currently you have duplicate code after both usage points of `createExplicitlyDelayedShutdownCallback`.
At this time, you don't need the callback argument.
Because you use the same callback in both cases.

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,66 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
...
+        if (explicitlyDelayedShutdownCallback == NULL) {
+            UA_LOG_SOCKET_ERRNO_WRAP(
+                UA_LOG_ERROR(cm->eventSource.eventLoop->logger,
```
jpfr, line 539: This is rather a warning.
Because we just get some additional delay (120s).
But there is no major disruption.

Also add a comment that if `createExplicitlyDelayedShutdownCallback` fails, the connection will still be closed eventually.

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,66 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
...
+    TCPConnectionManager *tcm = (TCPConnectionManager*)cm;
+    UA_RegisteredFD* rfd = (UA_RegisteredFD *) data;
+
+    UA_ByteString response = tcm->rxBuffer;
+    int ret = UA_recv(rfd->fd, (char *)response.data, response.length, MSG_DONTWAIT | MSG_PEEK);
+    if(ret == 0 || ret == -1) {
+        cm->connectionCallback(cm, (uintptr_t)rfd->fd, &rfd->context,
+                               UA_STATUSCODE_BADCONNECTIONCLOSED,
+                               0, NULL, UA_BYTESTRING_NULL);
+        TCP_close(tcm, rfd);
```
jpfr, line 532: You can early-return here.
So you don't need the else-case.

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,66 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
...
+                UA_LOG_ERROR(cm->eventSource.eventLoop->logger,
+                             UA_LOGCATEGORY_NETWORK,
+                             "TCP %u\t| Error Out of memory for creating delayed shutdown callback (%s)",
+                             (unsigned)rfd->fd, errno_str));
+            return;
+        }
+
+        UA_EventLoop *el = cm->eventSource.eventLoop;
+        el->addDelayedCallback(el, explicitlyDelayedShutdownCallback);
+        return;
```
jpfr, line 548: No return required here.

jpfr (commented):
Looks good!
Just one CI error on Windows
`54 : ~~Dr.M~~ error #1: UNADDRESSABLE ACCESS of freed memory: reading 0x01dc5e90-0x01dc5e94 4 byte(s) [D:\a\1\s\build\tests\test-verbose.vcxproj]`
and a few cosmetic comments.

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,66 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
...
 
+#ifdef _WIN32
+static
+UA_RegisteredFD *
+findRegisteredFD(TCPConnectionManager *cm, uintptr_t connectionId) {
+    UA_RegisteredFD *rfd, *rfd_tmp;
+    LIST_FOREACH_SAFE(rfd, &cm->fds, es_pointers, rfd_tmp) {
+
+        if (rfd->fd == connectionId) {
+            return rfd;
```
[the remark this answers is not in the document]
JanSurft: I would say "explicit is better than implicit". Because it is opaque to the reader of the code that `LIST_FOREACH` assigns to `rfd` in every loop iteration. And that this assignment is still valid after exiting the loop.

Leaving it like this, it is more clear in my opinion, it also saves no lines of code.

arch/eventloop_posix_tcp.c
JanSurft: It was meant to be like this.  In `delayedShutdownCallback` the logging needs to be inline because I cannot return an error-code. Whereas in the other place I can return `UA_STATUSCODE_BADOUTOFMEMORY` and propagate the error and handle logging somewhere else.

As for the two lines of code that get the eventloop and add the callback, I would again say "explicit is better than implicit". The callback is created via `create...` and then added to the eventloop. If possible I like functions that serve as little purposes as possible.

arch/eventloop_posix_tcp.c
JanSurft: I would prefer the if/else way because it carries the semantic of what is to be done there:

"if we receive nothing or if there is a receive-error, then close the socket. Else recreate the delayed shutdown callback and add it to the loop again."

arch/eventloop_posix_tcp.c
jpfr: Still todo.
Ideally set your editor to not do this automatically.

arch/eventloop_posix_tcp.c
jpfr: The calling function cannot handle an out-of-memory any better.
So we should handle it internally and not propagate the error outside of the current scope.

Normally we would reserve the memory already when the connection is created, so that the cleanup always works.
This pattern is also used in other places in the SDK already.

However, here the worst that can happen is a 120 second wait for the server shutdown.
And only on win32 where we are typically not memory constrained.
Hence, log the error with a warning and do not return an error code.

And with this, the two code paths can be merged as in the original review comment.

arch/eventloop_posix_tcp.c
```diff
@@ -490,6 +490,70 @@ TCP_registerListenSocketDomainName(UA_ConnectionManager *cm, const char *hostnam
...
+
+static UA_DelayedCallback
+*createExplicitlyDelayedShutdownCallback(UA_ConnectionManager *cm, UA_RegisteredFD *rfd, UA_Callback callback) {
+    UA_DelayedCallback *explicitlyDelayedShutdownCallback = (UA_DelayedCallback*) UA_malloc(sizeof(UA_DelayedCallback));
+    if (!explicitlyDelayedShutdownCallback) {
+        return NULL;
+    }
+    explicitlyDelayedShutdownCallback->callback = callback;
+    explicitlyDelayedShutdownCallback->application = cm;
+    UA_FD *fd = (UA_FD *) UA_malloc(sizeof(UA_FD));
```
jpfr, line 513: The FD always fits into the size of a pointer.
So you don't have to allocate this memory here.
Just cast the pointer itself.

arch/eventloop_posix_tcp.c
jpfr: An early return has the same semantics.
With the pointer-trick mentioned in another comment you get rid of the final `UA_free(data)`.

Ideally, the "happy case" of the control flow is linear and does not flow through nested blocks with `{ }` braces.
That can be achieved by branching out the edge cases with an early return.

As often, it is a question of taste.
Here I would prefer an early return.

Which case is the "happy case", I let you decide. 😄

jpfr (commented):
Another small improvement with this review...

jpfr:
Nice! Squash-merged.

PR #5034, opened 2022-03-23
feat(server) Support for sequence numbers in pubsub rt path

andreasebner:


jpfr:
Nice!
This can even go into 1.3.

keba-estr:
This commit resolves the update of the sequence number and some memory leaks, but the sequence number still remains unhandled in the function UA_NetworkMessage_updateBufferedNwMessage (case UA_PUBSUB_OFFSETTYPE_DATASETMESSAGE_SEQUENCENUMBER).

andreasebner:
@jpfr reopened as #5042 with 1.3 as the base branch
@keba-estr added in #5042 the subscriber configuration and the decode case is added

Issue #5000, opened 2022-03-10
Build size fluctuations for the same build settings


epetrevska:
## Description

When building the stack using the regular procedure as explained here: https://open62541.org/doc/current/building.html#building-with-cmake-on-ubuntu-or-debian we noticed that the size of the binary file (libopen62541.a) fluctuates even without any CMake options being changed.

This behavior can be pretty dangerous for production firmware, especially if the fluctuations are happening on smaller controllers which anyhows work with critically low amounts of memory. The differences which we observed here were in Bytes and KBytes sizes, some examples are:
- 7637500 bytes
- 7637468 bytes
- 7637620 bytes
![BuildSizeOPCUAStack](https://user-images.githubusercontent.com/77783092/157707361-a9e5dd0d-0069-478a-87e0-5fb2f7ff5f4d.png)

After debugging it a little, we think the problem is in the NodeSet Compiler. Every time we let it generate the .h and .c files, it randomly shifts the order of the nodes for whichever reason... Also, in case further information models are attached, and not just the base information model, the size fluctuations increases as well... 

We wanted to ask you if this behavior is already known to you and you have some solution against it?

Thank you and many greetings.

## Background Information / Reproduction Steps
Happens during any normal "cmake .. && make" build.

Used CMake options:

```bash
cmake .. && make
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): we tried it with various branches, but as an example tag 1.2.3
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Ubuntu 20.04
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [x] Critical issue

jpfr:
Hmm.
That’s a good observation.
The problem might be the use of dictionaries in Python.
We can switch them to OrderedDict to be order-stable.

PR #5041, opened 2022-03-29
fix(nc): fix the determinism of the generated sources of the nodeset compiler

tobiajung:
Currently, the nodeset compiler creates source file including the nodes of the information models
in a different order every time a rebuild is done. This is due to the use of sets which change the
order of there elements every time an element gets added. With this change the nodeset compiler
uses lists where order is relevant.

### DIFF
--- a/tools/nodeset_compiler/nodes.py
+++ b/tools/nodeset_compiler/nodes.py
@@ -69,7 +69,7 @@ class Node(object):
         self.symbolicName = None
         self.writeMask = None
         self.userWriteMask = None
-        self.references = set()
+        self.references = list()
         self.hidden = False
         self.modelUri = None
         self.parent = None
@@ -137,7 +137,7 @@ class Node(object):
                     reftype = RefOrAlias(av)
                 elif at == "IsForward":
                     forward = not "false" in av.lower()
-            self.references.add(Reference(source, reftype, target, forward))
+            self.references.append(Reference(source, reftype, target, forward))
 
     def getParentReference(self, parentreftypes):
         # HasSubtype has precedence
@@ -162,7 +162,6 @@ class Node(object):
         if isinstance(self, VariableNode) or isinstance(self, VariableTypeNode):
             if str(self.dataType) in aliases:
                 self.dataType = NodeId(aliases[self.dataType])
-        new_refs = set()
         for ref in self.references:
             if str(ref.source) in aliases:
                 ref.source = NodeId(aliases[ref.source])
@@ -170,21 +169,16 @@ class Node(object):
                 ref.target = NodeId(aliases[ref.target])
             if str(ref.referenceType) in aliases:
                 ref.referenceType = NodeId(aliases[ref.referenceType])
-            new_refs.add(ref)
-        self.references = new_refs
 
     def replaceNamespaces(self, nsMapping):
         self.id.ns = nsMapping[self.id.ns]
         self.browseName.ns = nsMapping[self.browseName.ns]
         if hasattr(self, 'dataType') and isinstance(self.dataType, NodeId):
             self.dataType.ns = nsMapping[self.dataType.ns]
-        new_refs = set()
         for ref in self.references:
             ref.source.ns = nsMapping[ref.source.ns]
             ref.target.ns = nsMapping[ref.target.ns]
             ref.referenceType.ns = nsMapping[ref.referenceType.ns]
-            new_refs.add(ref)
-        self.references = new_refs
         self.namespaceMapping = nsMapping
 
 class ReferenceTypeNode(Node):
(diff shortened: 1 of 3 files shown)


ccvca:
Related to #5000

jpfr:
Cool! Does this have a significant impact on timings?
I ask because the lookup to test membership becomes O(n) instead of O(1).

tobiajung:
> Does this have a significant impact on timings?

Unfortunately, generating the nodeset with this change can take up to 4x longer on my machine.
Generating the full nodeset on open62541/master takes 4.3 seconds while with my requested change it takes 16.4 seconds.
The increase in time is mainly due to the change to list() in the references member in the Node class.

tobiajung:
OrderedDicts are key value pairs. So I don't think that they can replace the sets for the references of nodes.
Sets and lists only contain one or more elements which can be accessed through an index and not through a key.

jpfr:
A 4x performance decrease is not possible.
So we need a smart solution.

OrderedDict can be used to create a stable set.
For example by putting the element as the key that point to a `None`.

This can also be wrapped in a small class for reuse.
See here for a more complete example:
https://github.com/rspeer/ordered-set/blob/457a1ec89660f6491ea36b05b44e55dc42f6a873/ordered_set/__init__.py#L55

tobiajung:
Thank you very much for your detailed information! I will take a look at it next week

tools/nodeset_compiler/nodes.py
```diff
@@ -69,7 +69,7 @@ def __init__(self):
         self.symbolicName = None
         self.writeMask = None
         self.userWriteMask = None
-        self.references = set()
+        self.references = dict()
```
jpfr, line 72: Add a short comment to document the “special” use of dict and the motivation for it.
So future generations can understand what you did.

tools/nodeset_compiler/nodeset.py
```diff
@@ -227,16 +227,16 @@ def filterRef(r, rt):
             if r.target == node.id:
                 if r.source not in self.nodes:
                     continue
-                self.nodes[r.source].references = set(filter(
-                    lambda rt: filterRef(r, rt),
-                    self.nodes[r.source].references
+                self.nodes[r.source].references = dict(filter(
+                    lambda rt: filterRef(r, rt[0]),
```
jpfr, line 231: Why rt[0]?

tools/nodeset_compiler/nodeset.py
```diff
@@ -227,16 +227,16 @@ def filterRef(r, rt):
...
+                    self.nodes[r.source].references.items()
                 ))
             elif r.source == node.id:
                 if r.target not in self.nodes:
                     continue
-                self.nodes[r.target].references = set(filter(
-                    lambda rt: filterRef(r, rt),
-                    self.nodes[r.target].references
+                self.nodes[r.target].references = dict(filter(
+                    lambda rt: filterRef(r, rt[0]),
```
jpfr, line 238: Why rt[0]?
Add a comment to document.

jpfr (commented):
Looking good!
Good idea to just use dict.

Some minor comments. Otherwise good to go.

tools/nodeset_compiler/nodeset.py
tobiajung: Only the key is passed through the filter. The `dict.items` statement returns all key value pairs as views. In order to get the key, the first item must be selected.
I added a comment.

tools/nodeset_compiler/nodeset.py
tobiajung: only the key is passed through the filter. The dict.items statement returns all key, value pairs as views. In order to get the key, the first item must be selected.
I added a comment.

tools/nodeset_compiler/nodes.py
tobiajung: Done

tobiajung:
Seems like the CI failed.
I have no idea why. I only added some comments.
The log says that it has something to do with a missing mbedtls header:
```
[ 65%] Building C object CMakeFiles/open62541-object.dir/open62541.c.obj
D:\a\1\s\build\open62541.c:66008:10: fatal error: mbedtls/entropy_poll.h: No such file or directory
66008 | #include <mbedtls/entropy_poll.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
https://dev.azure.com/open62541/open62541/_build/results?buildId=5541&view=logs&j=8ae1b360-14bf-521d-a907-7a80f3a3a1f8&t=77a3e05d-c2b1-5f33-230d-a41770d4a825&l=299

GoetzGoerisch:
@tobiajung I do have. It is now using mbetls 3.x

`Packages (1) mingw-w64-x86_64-mbedtls-3.1.0-1
`
https://dev.azure.com/open62541/open62541/_build/results?buildId=5541&view=logs&j=ad9e6d9e-6e5f-5424-e835-306608605e5b&t=2d5aafe1-b9f7-56bd-1478-b8e51f9d0a78&l=326

The mbedtls API was changed with v3.x, Needed changes are described in #4372

jpfr:
Nice! Squash-merged.

PR #5054, opened 2022-04-05
fix(deps): Fix missing conversions and zero-terminate the strTLS value.

NoelGraf:
Fix the errors that occur in the build process with the following CMake flags ([build.log](https://github.com/open62541/open62541/files/8419291/build.log)):
-DCMAKE_BUILD_TYPE=Debug
-DUA_BUILD_EXAMPLES=ON
-DUA_BUILD_UNIT_TESTS=ON
-DUA_ENABLE_PUBSUB=ON
-DUA_ENABLE_PUBSUB_MQTT=ON
-DUA_ENABLE_MQTT_TLS=ON
-DUA_ENABLE_MQTT_TLS_OPENSSL=ON
-DUA_NAMESPACE_ZERO=FULL

Fix missing conversions and zero-terminate the strTLS value to avoid the warning when the result is not expected to be NUL-terminated.

Fixed a bug in Cmake where the UA_ENABLE_MQTT_TLS_OPENSSL flag is overwritten again after it is set to ON in the UA_ENABLE_MQTT_TLS check.

### DIFF
--- a/deps/mqtt-c/mqtt.c
+++ b/deps/mqtt-c/mqtt.c
@@ -1205,10 +1205,11 @@ ssize_t mqtt_pack_connection_request(uint8_t* buf, size_t bufsz,
             connect_flags &= (uint8_t)~MQTT_CONNECT_CLIENTKEYPATH;
         }
 
+        /* Call the strncpy function with a size of two to zero-terminate the strTLS value */
         if (useTLS) /* useTLS = true */
-            memcpy(strTLS, "1", 1);
+            strncpy(strTLS, "1", 2);
         else
-            memcpy(strTLS, "0", 1);
+            strncpy(strTLS, "0", 2);
         connect_flags |= (uint8_t)~MQTT_CONNECT_USETLS;
         remaining_length += (uint32_t)__mqtt_packed_cstrlen(strTLS);
     }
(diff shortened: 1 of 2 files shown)


deps/mqtt-c/mqtt.c
```diff
@@ -1175,41 +1175,41 @@ ssize_t mqtt_pack_connection_request(uint8_t* buf, size_t bufsz,
...
 
         if (useTLS) /* useTLS = true */
-            strncpy(strTLS, "1", 1);
+            memcpy(strTLS, "1", 1);
         else
-            strncpy(strTLS, "0", 1);
-        connect_flags |= MQTT_CONNECT_USETLS;
+            memcpy(strTLS, "0", 1);
+        connect_flags |= (uint8_t)~MQTT_CONNECT_USETLS;
         remaining_length += (uint32_t)__mqtt_packed_cstrlen(strTLS);
```
jpfr, line 1213: This line calls strlen.
You need to zero-terminate the strTLS value.

jpfr (commented):
Good find!
Just a quick remark inline.

deps/mqtt-c/mqtt.c
NoelGraf: I have changed the call of strncpy so that the strTLS value is zero-terminated.

jpfr:
Nice!

PR #5056, opened 2022-04-08
feat(pubsub): Implement a heartbeat message

MurzynJ:
This still needs additional work:

- UA_DataSetWriter_generateDataSetMessageHeartbeat function is based on the old version of UA_DataSetWriter_generateDataSetMessage. One of the effects is, that it won't pass the current static code analyzer.  Generally this function needs to be updated or reworked in any manner.
- There is no unit test
- There is no example

### DIFF
--- a/src/pubsub/ua_pubsub_networkmessage.c
+++ b/src/pubsub/ua_pubsub_networkmessage.c
@@ -1430,6 +1430,11 @@ UA_DataSetMessage_encodeBinary(const UA_DataSetMessage* src, UA_Byte **bufPos,
     UA_StatusCode rv = UA_DataSetMessageHeader_encodeBinary(&src->header, bufPos, bufEnd);
     UA_CHECK_STATUS(rv, return rv);
 
+    if(src->data.keyFrameData.fieldCount == 0) {
+        // That's a heartbeat - nothing more to encode.
+        return rv;
+    }
+
     if(src->header.dataSetMessageType == UA_DATASETMESSAGE_DATAKEYFRAME) {
         if(src->header.fieldEncoding != UA_FIELDENCODING_RAWDATA) {
             rv = UA_UInt16_encodeBinary(&(src->data.keyFrameData.fieldCount), bufPos, bufEnd);
@@ -1494,6 +1499,10 @@ UA_DataSetMessage_decodeBinary(const UA_ByteString *src, size_t *offset, UA_Data
     UA_CHECK_STATUS(rv, return rv);
 
     if(dst->header.dataSetMessageType == UA_DATASETMESSAGE_DATAKEYFRAME) {
+        if(*offset == src->length) {
+            // Message ends after the header, that's a heartbeat
+            return rv;
+        }
         switch(dst->header.fieldEncoding) {
             case UA_FIELDENCODING_VARIANT:
             {
@@ -1638,7 +1647,7 @@ UA_DataSetMessage_calcSizeBinary(UA_DataSetMessage* p, UA_NetworkMessageOffsetBu
     if(p->header.configVersionMinorVersionEnabled)
         size += UA_UInt32_calcSizeBinary(&p->header.configVersionMinorVersion);
 
-    if(p->header.dataSetMessageType == UA_DATASETMESSAGE_DATAKEYFRAME) {
+    if(p->header.dataSetMessageType == UA_DATASETMESSAGE_DATAKEYFRAME && p->data.keyFrameData.fieldCount != 0) {
         if(p->header.fieldEncoding != UA_FIELDENCODING_RAWDATA){
             size += UA_calcSizeBinary(&p->data.keyFrameData.fieldCount, &UA_TYPES[UA_TYPES_UINT16]);
         }
@@ -1694,7 +1703,7 @@ UA_DataSetMessage_calcSizeBinary(UA_DataSetMessage* p, UA_NetworkMessageOffsetBu
                 size += UA_calcSizeBinary(&p->data.keyFrameData.dataSetFields[i], &UA_TYPES[UA_TYPES_DATAVALUE]);
             }
         }
-    } else if(p->header.dataSetMessageType == UA_DATASETMESSAGE_DATADELTAFRAME) {
+    } else if(p->header.dataSetMessageType == UA_DATASETMESSAGE_DATADELTAFRAME && p->data.deltaFrameData.fieldCount != 0) {
         //TODO clarify how to handle DATADELTAFRAME messages with RT
         if(p->header.fieldEncoding != UA_FIELDENCODING_RAWDATA)
             size += UA_calcSizeBinary(&p->data.deltaFrameData.fieldCount, &UA_TYPES[UA_TYPES_UINT16]);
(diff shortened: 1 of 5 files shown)


src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -1430,6 +1430,11 @@ UA_DataSetMessage_encodeBinary(const UA_DataSetMessage* src, UA_Byte **bufPos,
     UA_StatusCode rv = UA_DataSetMessageHeader_encodeBinary(&src->header, bufPos, bufEnd);
     UA_CHECK_STATUS(rv, return rv);
 
+    if(src->data.keyFrameData.fieldCount == 0) {
```
JanSurft, line 1433: What is the relation of **6.2.4.3 KeyFrameCount** and the `keyFrameData.fieldCount`?

src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -1494,6 +1499,10 @@ UA_DataSetMessage_decodeBinary(const UA_ByteString *src, size_t *offset, UA_Data
     UA_CHECK_STATUS(rv, return rv);
 
     if(dst->header.dataSetMessageType == UA_DATASETMESSAGE_DATAKEYFRAME) {
+        if(*offset == src->length) {
```
JanSurft, line 1502: Can use the local variable `initialOffset` here.

src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -188,12 +188,14 @@ UA_Server_freezeWriterGroupConfiguration(UA_Server *server,
...
-        TAILQ_FOREACH(dataSetField, &publishedDataSet->fields, listEntry) {
-            dataSetField->configurationFrozen = true;
+        if(publishedDataSet) {
+            publishedDataSet->configurationFreezeCounter++;
+            publishedDataSet->configurationFrozen = true;
+            /* DataSetFields freeze */
+            UA_DataSetField *dataSetField;
+            TAILQ_FOREACH(dataSetField, &publishedDataSet->fields, listEntry) {
+                dataSetField->configurationFrozen = true;
+            }
```
JanSurft, line 198: Better to pull out pure refactoring out of the functional pull request to comprehend all changes related to the newly introduced feature better.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -1355,4 +1367,165 @@ UA_DataSetWriter_generateDataSetMessage(UA_Server *server,
...
+    /* Set the sequence count. Automatically rolls over to zero */
+    dataSetWriter->actualDataSetMessageSequenceCount++;
+    /* Prepare DataSetMessageContent */
+    dataSetMessage->header.dataSetMessageValid = true;
+    dataSetMessage->header.dataSetMessageType = UA_DATASETMESSAGE_DATAKEYFRAME;
+    dataSetMessage->data.keyFrameData.fieldCount = 0; // Heartbeat
+
+    return UA_STATUSCODE_GOOD;
+}
+
```
JanSurft, line 1530: I need for some semantically meaningful functions that are called from this function to reduce its cognitive complexity.

src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -1096,6 +1112,20 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
         UA_PublishedDataSet *pds =
             UA_PublishedDataSet_findPDSbyId(server, dsw->connectedDataSet);
         if(!pds) {
+            if (UA_NodeId_isNull(&dsw->connectedDataSet)) {
```
JanSurft, line 1115: Check for code formatting adhere to the rules of the stack.

e.g. `if` statements do not have a following space.

src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -1096,6 +1112,20 @@ UA_WriterGroup_publishCallback(UA_Server *server, UA_WriterGroup *writerGroup) {
...
+                if (res1 != UA_STATUSCODE_GOOD) {
+                    UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                            "PubSub Publish: Heartbeat DataSetMessage creation failed");
+                    continue;
+                }
+                dsWriterIds[dsmCount] = dsw->config.dataSetWriterId;
+                dsmCount++;
+                continue;
+            }
+
```
JanSurft, line 1128: Is it possible to check whether its a hearbeat at the same place where we check whether it possible to batch the message?

JanSurft (commented):
Thanks for the great pull request. Some comments and questions in this review. Other than that some open points:

- Where do we ensure the KeyFrameCount to be 1?
- Where do we ensure that the SubscribedDataSet is of base type SubscribedDataSetType for Heartbeat msgs?
- dataSetName if NULL or empty for Hearbeat DataSetWriter?
- Is including version fields mandatory for Heartbeat messages?

src/pubsub/ua_pubsub_networkmessage.c
MurzynJ: There is no relation.
KeyFrameCount is a parameter of the Writer, which controls the frequency of Key frames in the Key/Delta frame sequence.
keyFrameData.fieldCount is a field in the message 7.2.2.5.5 Data Key Frame DataSetMessage

MurzynJ:
>     * Where do we ensure that the SubscribedDataSet is of base type SubscribedDataSetType for Heartbeat msgs?

Currently we don't fill in address space representation of the DataSetReader with meaningful data, so in our case the SubscribedDataSet is always set to base type SubscribedDataSetType.

In the internal logic, currently we represent the SubscribedDataSet as a C union:
image

To be precise, in the above structure we should have a way to specify that the SubscribedDataSet is null, to reflect the following point in specification:
image
But in practice, the internal logic would work, regardless what is set for the SubscribedDataSet, since the code in UA_DataSetReader_process returns before the SubscribedDataSet is even checked.

MurzynJ:
> * dataSetName if NULL or empty for Heartbeat DataSetWriter?

In my PR, there is no place where checking the dataSetName makes any sense.
But if we want to support creating heartbeat writers also through the address space methods API, then some additional checks would be necessary. I found a place where the current code would return an error, instead of creating a heartbeat writer:
image

MurzynJ:
> * Where do we ensure the KeyFrameCount to be 1?

You're right, we should check it. I pushed a new version that includes the check in UA_Server_addDataSetWriter.

src/pubsub/ua_pubsub_writergroup.c
MurzynJ: That's not just refactoring. In case of heartbeat, the publishedDataSet pointer would be actually NULL.

src/pubsub/ua_pubsub_writer.c
MurzynJ: Generally this function needs refactoring, at least to bring it to the state equal to UA_DataSetWriter_generateDataSetMessage.

src/pubsub/ua_pubsub_writergroup.c
MurzynJ: You're right. Fixed by ec25477.

src/pubsub/ua_pubsub_writergroup.c
MurzynJ: You're right. Fixed by ec25477.

andreasebner:
Merged with: #5081

PR #5020, opened 2022-03-18
[Review] build(deps): Switches branch for `deps/ua-nodeset` -> UA v1.05

GoetzGoerisch:
* The `latest `branch is a merge between v1.05 base spec and all CS

Base UA Spec is v1.05 now.

PR #5062, opened 2022-04-13
fix(server) adjust regex for latest nodeset flag

andreasebner:
The open62541 can be compiled using the 1.05.01 nodeset. Due to missing specifications in the deps/ua-nodeset repository, a special flag must be set for the nodeset-option: LATEST_1_05. Due to a change between 1.05.00 and 1.05.01 a updated regex for the file version check is needed.


GoetzGoerisch:
Is this relating to #5020 
The latest branch should contain all `1.05`base files and companion specs and such from the `1.04` branch

PR #5071, opened 2022-04-20
fix(sec) Remove use of mbedtls internal headers

andreasebner:
Background: The mbedtls version from the pacman installer was updated to mbedtls 3.1.0. We used in our stack definitions from the entropy_poll.h, which is meant to be mbedtls internal only. Therefore, the header is not available anymore, see the mbedtls changelog:

   * Move internal headers.
     Header files that were only meant for the library's internal use and
     were not meant to be used in application code have been moved out of
     the include/ directory. The headers concerned are bn_mul.h, aesni.h,
     padlock.h, entropy_poll.h and *_internal.h.

The entropy sources are now configured within the mbedtls_entropy_init only and custom configurations must be done directly in the mbedtls_config.

### DIFF
--- a/plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
+++ b/plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
@@ -9,7 +9,6 @@
 #include <mbedtls/aes.h>
 #include <mbedtls/ctr_drbg.h>
 #include <mbedtls/entropy.h>
-#include <mbedtls/entropy_poll.h>
 #include <mbedtls/error.h>
 #include <mbedtls/md.h>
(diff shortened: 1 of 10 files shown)


GoetzGoerisch:
Fixes #4372

jpfr:
Great! Now we see other errors. 😄 
```
D:\a\1\s\build\open62541.c:65570:64: error: no member named 'len' in 'struct mbedtls_rsa_context'
    return mbedtls_pk_rsa(cc->policyContext->localPrivateKey)->len;
```

gokhalehemanth:
> mbedtls_pk_rsa

Getting same error 
error C2039: 'len' : is not a member of 'mbedtls_rsa_context', what is the solution

PR #5095, opened 2022-05-03
fix(arch): add missing defines for poll on arch freertosLWIP

KnoerleMan:
Commit #4704 add polling to check if output connection is ready before write. This commit add missing defines for arch freertosLWIP and fix corresponding compiler errors.


jpfr:
Nice.
It would be great to integrate freertos into our CI infrastructure.
@KnoerleMan do you have a hunch on how to approach that?

Issue #5084, opened 2022-04-27
How can I disable CPack in Open62541 when I try to use it in my project as a git submodule?


xydan83:
## Description
I will try to use the SDK in my project and include it as a git submodule. But when I try to use CPack to build a Debian package, I see in the config file (CPackConfig.cmake) that it's not my variables that are being used. How can I disable the use of CPack in SDK?

## Checklist

Please provide the following information:

 - [ ] open62541 Version (release number or git tag): v1.3
 - [ ] Other OPC UA SDKs used (client or server): client and server
 - [ ] Operating system: Ubuntu 20.04.1
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

PR #5103, opened 2022-05-05
fix(cmake): Added a new option UA_EXPECTED_BUILD_PACKAGE which can di…

xydan83:
…sable CPack so that there are no conflicts with another project.

#5084

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1739,7 +1739,9 @@ else()
     install(FILES ${PROJECT_BINARY_DIR}/open62541.h DESTINATION include)
 endif()
 
-add_subdirectory(tools/packaging)
+if (UA_PACK_DEBIAN)
+    add_subdirectory(tools/packaging)
+endif()
 
 ##################################
 # Visual studio solution folders #


jpfr:
Good idea.
Just call it UA_BUILD_PACKAGE to stick with established prefixes.

jpfr:
Good idea.
Just call it UA_BUILD_PACKAGE to stick with established prefixes.

andreasebner:
Your PR looks great. There is only one quesiton remaining: Why could we not use the "UA_PACK_DEBIAN" flag itself and make the include conditional on this flag?

xydan83:
> Your PR looks great. There is only one quesiton remaining: Why could we not use the "UA_PACK_DEBIAN" flag itself and make the include conditional on this flag?

Good point. We can indeed bind CPACK work to this flag without making a new one. The main thing is that it does not have other side effects. I will update the PR.

xydan83:
> Good idea. Just call it UA_BUILD_PACKAGE to stick with established prefixes.

As @andreasebner suggested, I'm using the already existing UA_PACK_DEBIAN flag, if I understand correctly.

PR #5104, opened 2022-05-05
[REVIEW] refactor(pubsub): cleanup add-connection call

JanSurft:
clean up and simplify pubsub-manager add-connection.

also add documentation for add-connection call

### DIFF
--- a/src/pubsub/ua_pubsub_manager.c
+++ b/src/pubsub/ua_pubsub_manager.c
@@ -16,76 +16,106 @@
 
 #define UA_DATETIMESTAMP_2000 125911584000000000
 
-UA_StatusCode
-UA_Server_addPubSubConnection(UA_Server *server,
-                              const UA_PubSubConnectionConfig *connectionConfig,
-                              UA_NodeId *connectionIdentifier) {
+static UA_PubSubTransportLayer *
+getTransportProtocolLayer(const UA_Server *server,
+                          const UA_String *transportProfileUri) {
     /* Find the matching UA_PubSubTransportLayers */
     UA_PubSubTransportLayer *tl = NULL;
     for(size_t i = 0; i < server->config.pubSubConfig.transportLayersSize; i++) {
-        if(connectionConfig &&
-           UA_String_equal(&server->config.pubSubConfig.transportLayers[i].transportProfileUri,
-                           &connectionConfig->transportProfileUri)) {
+        if(UA_String_equal(&server->config.pubSubConfig.transportLayers[i].transportProfileUri,
+                           transportProfileUri)) {
             tl = &server->config.pubSubConfig.transportLayers[i];
         }
     }
     if(!tl) {
         UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
                      "PubSub Connection creation failed. Requested transport layer not found.");
-        return UA_STATUSCODE_BADNOTFOUND;
+        return NULL;
     }
+    return tl;
+}
 
+static void
+UA_PubSubConfig_delete(UA_PubSubConnectionConfig *tmpConnectionConfig) {
+    UA_PubSubConnectionConfig_clear(tmpConnectionConfig);
+    UA_free(tmpConnectionConfig);
+}
+
+static UA_StatusCode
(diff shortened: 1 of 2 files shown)


src/pubsub/ua_pubsub_manager.c
```diff
@@ -13,79 +13,107 @@
 
 #include "server/ua_server_internal.h"
 #include "ua_pubsub_ns0.h"
+#include "open62541/plugin/log_stdout.h"
```
jpfr, line 17: Don’t depend on specific plugins from the main library.

src/pubsub/ua_pubsub_manager.c
```diff
@@ -13,79 +13,107 @@
...
     if(!tl) {
         UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
                      "PubSub Connection creation failed. Requested transport layer not found.");
-        return UA_STATUSCODE_BADNOTFOUND;
+        return NULL;
     }
+    return tl;
+}
 
+static UA_INLINE
```
jpfr, line 39: Static is enough.
Static inline is for header files.

src/pubsub/ua_pubsub_manager.c
```diff
@@ -13,79 +13,107 @@
...
+                       logger, UA_LOGCATEGORY_SERVER,
+                       "PubSub Connection creation failed. Out of Memory.");
 
-    UA_StatusCode retval = UA_PubSubConnectionConfig_copy(connectionConfig, tmpConnectionConfig);
-    if(retval != UA_STATUSCODE_GOOD){
-        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
-                     "PubSub Connection creation failed. Could not copy the config.");
-        return retval;
+    UA_StatusCode retval = UA_PubSubConnectionConfig_copy(srcConfig, *dstConfig);
+    UA_CHECK_STATUS_ERROR(retval, return retval, logger, UA_LOGCATEGORY_SERVER,
```
jpfr, line 54: You should clean up the malloc from above here.

jpfr (commented):
Looks good! A few small comments…

src/pubsub/ua_pubsub_manager.c
JanSurft: You mean there should be <,> around this instead of " ?

Because this plugin is also used in e.g. "src/pubsub/ua_pubsub_config.c" for logging without the server context.

JanSurft: Also we use all kinds of plugins in the main library, maybe I did not get the comment:

```shell
:open62541/src& ag --cc "#include <open62541/plugin"
ua_connection.c
21:#include <open62541/plugin/eventloop.h>

pubsub/ua_pubsub.h
15:#include <open62541/plugin/pubsub.h>

pubsub/ua_pubsub_networkmessage.h
14:#include <open62541/plugin/securitypolicy.h>

pubsub/ua_pubsub_config.c
9:#include <open62541/plugin/log_stdout.h>
13:#include <open62541/plugin/pubsub_ethernet.h>
15:#include <open62541/plugin/pubsub_udp.h>

pubsub/ua_pubsub_manager.c
16:#include <open62541/plugin/log_stdout.h>

ua_connection_internal.h
14:#include <open62541/plugin/network.h>
15:#include <open62541/plugin/eventloop.h>

server/ua_subscription.h
23:#include <open62541/plugin/nodestore.h>

server/ua_server_internal.h
22:#include <open62541/plugin/nodestore.h>

ua_securechannel.h
16:#include <open62541/plugin/log.h>
17:#include <open62541/plugin/securitypolicy.h>

server/ua_services_attribute.c
28:#include <open62541/plugin/historydatabase.h>
```

jpfr: Most of these are from /include/open62541/plugin.
These define the plug-in api itself.

a few of them are from /plugins/include/open62541/…
These are the plug-in implementations.
The plug-in implementations should be plugged into the config and used from there.
Hence the name…

JanSurft: That clarifies it, so we should never be able to log something if there is no log context given. Only if it happens inside a plugin itself like e.g.

`plugins/ua_pubsub_ethernet.c`

PR #5105, opened 2022-05-05
Fixes and documentation for the updated server port config

jpfr:


jpfr:
Due to public request I am merging this now.
There is still a memleak lurking in some cases.
This will be fixed in a next PR ASAP.

PR #5119, opened 2022-05-10
[WIP] Feature/testcase pubsub udp unicast

JanSurft:
Add a testcase  for unicast pubsub

```
Publisher 
Connection(address=opc.udp://localhost:4800)                                                 
WriterGroup(transportSettings.address=127.0.0.1:4801)

                 |
                 v

Subscriber
Connection(address=opc.udp://localhost:4801)
```

## Info

- changed transportlayer name from `UA_PubSubTransportLayerUDPMP` to `UA_PubSubTransportLayerUDP`

## Todo

- [x] Make `UA_DatagramWriterGroupTransport2DataType` available from NodeSet (added in _UA-1.05.01-2022-02-24 Release_)
- [x] Integrate eventloop logic for pubsub
- [x] Integrate UDP Unicast logic for pubsub eventloop
- [ ] Fix Issues below

## Issues

- Eventloop needs to be started before the first connection is opened (UA_Server_addPubSubConnection error)
  - e.g. when a publisher connection is added an set operational before the server is started it fails because for adding an operational connection on the eventloop, the eventloop must be started before.
  - possible workarounds:
    - just in time opening of pubsub connections during the first server->run
    - if the eventloop is not started already (e.g. external el) start it as soon as its is added for the server
- Error event on socket if removing subscriber from pubsub communication
- el->sendWithConnection expects the buffer to be allocated on the heap, but sometimes it might be allocated on the stack


src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -860,8 +911,17 @@ sendNetworkMessageJson(UA_PubSubConnection *connection, UA_DataSetMessage *dsm,
...
     UA_assert(bufPos == bufEnd);
 
+    /* Choose the channel */
+    UA_PubSubChannel *channel = NULL;
+
+    if(transportSettings->content.decoded.data) {
+        channel = writerGroup->channel;
+    } else {
+        channel = connection->channel;
+    }
```
Fischer-Thomas, line 921: This can result in a SegFault when the transportSettings->type is DatagramWriterGroupTransportDataType (not DatagramWriterGroupTransport**2**DataType). This can happen when configuring a multicast publisher and setting the transportsettings

src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -102,6 +102,36 @@ UA_Server_addWriterGroup(UA_Server *server, const UA_NodeId connection,
...
+                                   UA_LOGCATEGORY_SERVER, "PubSub Connection creation failed. Requested transport layer not found.");
+                UA_TransportLayerContext ctx;
+                ctx.writerGroupAddress = address;
+                ctx.connection = currentConnectionContext;
+                ctx.connectionConfig = currentConnectionContext->config;
+                ctx.decodeAndProcessNetworkMessage = UA_decodeAndProcessNetworkMessage;
+                ctx.server = server;
+
+                newWriterGroup->channel = tl->createPubSubChannel(tl, &ctx);
+            }
```
Fischer-Thomas, line 132: There should also be a check if the Connection->address is opc.udp://localhost:[port] in case the configuration is invalid

src/pubsub/ua_pubsub_writergroup.c
```diff
@@ -1010,9 +1070,16 @@ sendNetworkMessageUADP(UA_PubSubConnection *connection, UA_WriterGroup *wg,
     rv = encodeNetworkMessage(wg, &nm, &buf);
     UA_CHECK_STATUS(rv, goto cleanup_with_msg_size);
 
+    /* Choose the channel */
+    UA_PubSubChannel *channel = NULL;
+    if(transportSettings->content.decoded.data) {
+        channel = wg->channel;
+    } else {
+        channel = connection->channel;
+    }
```
Fischer-Thomas, line 1079: This can result in a SegFault when the transportSettings->type is DatagramWriterGroupTransportDataType (not DatagramWriterGroupTransport2DataType). This can happen when configuring a multicast publisher and setting the transportsettings

Fischer-Thomas (changes requested):

PR #5121, opened 2022-05-11
[REVIEW] LDS-ME IPV6 support in Open62541

Muddasir-Shakil:
This PR adds the support for the creation of mdnsd socket for IPV6/IPV4 socket and adding IPV6 records (QTYPE_AAAA) to publish. The unit tests for selecting ipv4/ipv6 interfaces are also provided in this PR. This PR also adds the functionality for the clients to do mdns lookup on .local domain if the multicast discovery is enabled. This is performed additionally to the dns lookup which will fail on link local or in subnets where no dns server is running.

### DIFF
--- a/arch/eventloop_posix_tcp.c
+++ b/arch/eventloop_posix_tcp.c
@@ -4,6 +4,7 @@
  *
  *    Copyright 2021-2022 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
  *    Copyright 2021 (c) Fraunhofer IOSB (Author: Jan Hermes)
+ *    Copyright 2022 (c) Linutronix GmbH (Author: Muddasir Shakil)
  */
 
 #include "eventloop_posix.h"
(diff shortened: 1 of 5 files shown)


jpfr:
Hey @Muddasir-Shakil ,

Cool!

A few questions:
- I can contact @pro to merge your mdnds PR in his repo with priority.
  In addition, how much effort would it be to switch to the upstream repository at https://github.com/troglobit/mdnsd ?
  This will happen eventually. And it would be a pity to lose IPv6 support then.
- The code between IPv4 and IPv6 is separated. Would it be possible to harmonize them more using the `struct sockaddr_storage` everywhere?
- In the current config, an interface name is defined. Would it make sense to have a hostname instead of an interface name?
  I'm asking because we use the hostname in other places of the config. So it could be uniform.

Eventually we will move discovery to use the same EventLoop "UDP ConnectionManager" once that has also gained multicast support. Then we can move the dependency on system headers out of the main library. Because we also have non-Linux and embedded targets that currently can't do multicast discovery.

Muddasir-Shakil:
Hey @jpfr 

>     * I can contact @Pro to merge your mdnds PR in his repo with priority.
Thank you, that will be awesome. 

>       In addition, how much effort would it be to switch to the upstream repository at https://github.com/troglobit/mdnsd ?
Upstream has refactored the code, so I have to adopt my changes accordingly and also we would have to adopt our CMake to integrate upstream with out buildsystem. They use Autotools but it is possible by ExternalProject function in Cmake. Cannot really say how much effort. 

>     * The code between IPv4 and IPv6 is separated. Would it be possible to harmonize them more using the `struct sockaddr_storage` everywhere?
> 
I tried to use sockaddr_storage as much as possible but I can look at it again where I can improve the code. 

>     * In the current config, an interface name is defined. Would it make sense to have a hostname instead of an interface name?
>       I'm asking because we use the hostname in other places of the config. So it could be uniform.
> 
I would keep the interface config option there, because this way application will be restricted to use single interface for multicast outgoing. 
I do not know why stefan opted to Ip address for interface selection, I would suggest for Interface name option like "eth0", it just looks cleaner and easier for user prespective. But in this PR I kept it as original may be a design choice that I did not understand.

Muddasir-Shakil:
Hey @jpfr 
I updated the code to use sockaddr_storage wherever I could see possible. Now waiting on @Pro feedback.
>     * The code between IPv4 and IPv6 is separated. Would it be possible to harmonize them more using the `struct sockaddr_storage` everywhere?

jpfr:
Hey @Muddasir-Shakil ,

The CI is still failing and we have a merge conflict.
We will review once this is resolved.

We already have an agreement on the feature in general from our call.

arch/eventloop_posix_tcp.c
```diff
@@ -778,7 +779,42 @@ TCP_openActiveConnection(UA_ConnectionManager *cm, const UA_KeyValueMap *params,
                        "TCP\t| Lookup of %s failed (%s)",
                        hostname, gai_strerror(error));
 #endif
+
```
jpfr, line 782: The EventLoop encapsulates architecture- and OS-specific differences.
This file does TCP and nothing else.

please handle the domain-name rewriting business one layer up before you call into the EventLoop layer.

I thought mdns was udp based. Why changes in the TCP layer?

arch/eventloop_posix_tcp.c
```diff
@@ -778,7 +779,42 @@ TCP_openActiveConnection(UA_ConnectionManager *cm, const UA_KeyValueMap *params,
...
+#if defined(UA_ENABLE_DISCOVERY) && defined(UA_ENABLE_DISCOVERY_MULTICAST)
+        char *domain = ".local";
+        size_t len = strlen(domain);
```
jpfr, line 785: It might be simpler to use a static 256-byte buffer for the domain rewriting.
And test for the length instead of mallocing.

src/server/ua_services_discovery_multicast.c
```diff
@@ -101,83 +108,269 @@ addMdnsRecordForNetworkLayer(UA_Server *server, const UA_String *appName,
...
+                     sizeof(struct in_addr)) < 0) {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                     "Multicast DNS: failed setting IP_MULTICAST_IF %s", strerror(errno));
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
+    return UA_STATUSCODE_GOOD;
+}
+
+static UA_StatusCode
+setMulticastInterface(UA_Server *server, UA_SOCKET sock,
```
jpfr, line 247: The near-identical logic was implemented for the UDP EventSource for the EventLoop.
You could at runtime
- get the EventSource by its „udp“ protocol identifier
- Open a connection to the UDP multicast group. This is detected automatically from the IP range.
- Cast the generated connection-identifier to a socket when needed for the mdns library (only for Linux, Win32 sockets are pointers).

I really want to avoid duplicate logic for the same functionality.
If you cannot follow the above steps, please document why. So we can fix the underlying issue in the future.

jpfr (commented):
Good code as usual.
Comments inline.

PR #5126, opened 2022-05-13
refactor(core): replace isGood with UA_StatusCode_isGood

mlgiraud:
Use existing UA_StatuCode_isGood function that is more accurate instead of custom internal function.

### DIFF
--- a/src/ua_securechannel.c
+++ b/src/ua_securechannel.c
@@ -911,8 +911,8 @@ UA_SecureChannel_receive(UA_SecureChannel *channel, void *application,
     /* Listen for messages to arrive */
     UA_ByteString buffer = UA_BYTESTRING_NULL;
     UA_StatusCode res = connection->recv(connection, &buffer, timeout);
-    UA_CHECK_STATUS(res, return res);
-
+    if(res != UA_STATUSCODE_GOOD)
+        return res;
     /* Try to process one complete chunk */
     res = UA_SecureChannel_processBuffer(channel, application, callback, &buffer);
     connection->releaseRecvBuffer(connection, &buffer);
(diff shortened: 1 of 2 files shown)


jpfr:
I like the simplification.
Any simplification is good.

But I think this could lead to problems in the future.
Currently, `if(retval == 0)` is a very common pattern in the core SDK.
Where of course GOOD==0.

`grep` shows me that we use
- `UA_STATUSCODE_GOODSUBSCRIPTIONTRANSFERRED` 
- `UA_STATUSCODE_GOODNONCRITICALTIMEOUT`
- `UA_STATUSCODE_GOODCALLAGAIN`
in the library.

With the new macro, we might have hidden regressions for code that uses these status codes.
And developers can introduce bugs working on the wrong assumption that the UA_CHECK macro follows the `if(res == 0)` pattern.

Why don't we rename the `isGood` method for the macro to avoid the name clash and make the intent clear?

mlgiraud:
The regression is an argument i agree with, but this should mostly be covered by the tests, and if not, we can add new tests and quickly fix the regression.

I disagree with the assumptions of developers. I for instance assumed that the UA_CHECK_STATUS macro would perform a check with the UA_StatusCoode_isGood function because that is part of the public api, and my code then did not work, because it returned on a good statuscode. In the few cases where we actually want to fail on a good statuscode, we should do so explicitly.

jpfr:
> In the few cases where we actually want to fail on a good statuscode, we should do so explicitly.

There is a lot of code that expects "exactly good".
```
$ grep -r "!= UA_STATUSCODE_GOOD" | wc -l
585
```

I would go so far that most developers don't even know there are funny-but-still-good status.
It is much better to handle the exception (funny-but-still-good) explicitly and not change the base case.

It would be great to improve the situation in a PR that documents and makes the code more clear.
But the switch to UA_StatusCode_isGood everywhere is WONTFIX.

Independent of that, I applaud the effort to have consistency!

PR #5135, opened 2022-05-20
test(core): fix compiler errors

bluhm:
OpenBSD clang version 13.0.0 complains about -Wsign-conversion
errors.  Use correct types to avoid sign conversion and bad comparison.


jpfr:
Nice

PR #5138, opened 2022-05-23
feat(cmake): Allow UA_NODESET_DIR to be custom defined.

NoelGraf:


CMakeLists.txt
```diff
@@ -23,7 +23,9 @@ include(GNUInstallDirs)
 
 # Set when installed via make install
 set(open62541_TOOLS_DIR ${PROJECT_SOURCE_DIR}/tools)
-set(open62541_NODESET_DIR ${PROJECT_SOURCE_DIR}/deps/ua-nodeset)
+if(NOT UA_NODESET_DIR)
```
jpfr, line 26: If you use `option(UA_NODESET_DIR ...)` then the configuration appears in the GUI tools for cmake as well.

jpfr (commented):
Cool!
Just one comment.

CMakeLists.txt
NoelGraf: Good point!
I have changed the code line.

PR #5148, opened 2022-05-25
[REVIEW] feat(el): udp connection-manager parameters for openConnection

JanSurft:
Integrate parameters for opening udp connections via the udp-connectionmanager.

### Logic

- retrieve address information
- create socket
- setup common options from parameters
- if multicast
  - setup ipv4 and ipv6 multicast options from parameters
- bind...

### TODO
- [x] parameter passing/processing
- [x] document parameter api
- [x] check about AI_NUMERIC

### DIFF
--- a/arch/eventloop_posix_udp.c
+++ b/arch/eventloop_posix_udp.c
@@ -8,6 +8,13 @@
 
 #include "eventloop_posix.h"
 
+#define IPV4_PREFIX_MASK 0xF0
+#define IPV4_MULTICAST_PREFIX 0xE0
+#ifdef UA_IPV6
+#   define IPV6_PREFIX_MASK 0xFF
+#   define IPV6_MULTICAST_PREFIX 0xFF
+#endif
+
 #define UA_MAXBACKLOG 100
 #define UA_MAXHOSTNAME_LENGTH 256
 #define UA_MAXPORTSTR_LENGTH 6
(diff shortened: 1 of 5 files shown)


arch/eventloop_posix_udp.c
```diff
@@ -8,8 +8,16 @@
...
+#define IPV4_PREFIX_MASK 0xF0
+#define IPV4_MULTICAST_PREFIX 0xE0
+#ifdef UA_IPV6
+#   define IPV6_PREFIX_MASK 0xFF
+#   define IPV6_MULTICAST_PREFIX 0xFF
+#endif
+
 #define UA_MAXBACKLOG 100
 #define UA_MAXHOSTNAME_LENGTH 256
+#define UA_MAXNETIFNAME_LENGTH 256
```
jpfr, line 20: The define is unused.

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+        UA_KeyValueMap_getScalar(params, paramsSize,
+                                 UA_QUALIFIEDNAME(0, "networkInterface"),
+                                 &UA_TYPES[UA_TYPES_STRING]);
+    if(!networkInterface) {
+        UA_LOG_DEBUG(logger, UA_LOGCATEGORY_NETWORK,
+                     "UDP\t| No network interface found");
+        return 0;
+    }
+    // if(networkInterface->length >= UA_MAXNETIFNAME_LENGTH) {
+    //     UA_LOG_ERROR(logger, UA_LOGCATEGORY_EVENTLOOP,
```
jpfr, line 140: Remove unused code in comments.

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+                           "Cannot set socket option IP_MULTICAST_TTL. Error: %s",
+                           errno_str));
+    }
+}
+
+static UA_INLINE void
+setReuseAddress(UA_SOCKET sockfd, UA_Boolean enableReuse, const UA_Logger *logger) {
+    /* Set reuse address -> enables sharing of the same listening address on
+    * different sockets */
+    if(enableReuse){
```
jpfr, line 244: Early return instead of wrapping everything into an if clause.
Keep the code flat.

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+                               "Cannot set socket option SO_REUSEADDR. Error: %s",
+                               errno_str));
+        }
+    }
+}
+
+#ifdef __linux__
+static UA_StatusCode
+setSocketPriority(UA_SOCKET sockfd, UA_UInt32 *socketPriority, const UA_Logger *logger) {/* Setting the socket priority to the socket */
+    if(socketPriority != NULL) {
```
jpfr, line 260: Early return instead of deep nesting.

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+                             "PubSub Connection creation problem. Priority setup failed: "
+                             "Cannot set socket option SO_PRIORITY. Error: %s", errno_str));
+            return UA_STATUSCODE_BADINTERNALERROR;
+        }
+    }
+    return UA_STATUSCODE_GOOD;
+}
+#endif
+
+static UA_StatusCode
```
jpfr, line 274: This does not need a status code it can just return a Boolean.
Below also.

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+                UA_free(socketPriority);
+            }
+#endif
+        } else if (UA_String_equal(&prop->key.name, &hostnameParam) ||
+                   UA_String_equal(&prop->key.name, &portParam) ||
+                   UA_String_equal(&prop->key.name, &listenHostnamesParam) ||
+                   UA_String_equal(&prop->key.name, &listenPortParam)) {
+            /* ignore, required args are handled elsewhere explicitly */
+        } else {
+            UA_LOG_WARNING(logger, UA_LOGCATEGORY_SERVER,
```
jpfr, line 354: Log the name of the offending parameter

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+    return UA_STATUSCODE_GOOD;
+}
+static UA_StatusCode
+configureMulticastInterfaceIPv4(UA_FD socket,
+                                socklen_t addrlen, struct sockaddr_in *addr,
+                                size_t paramsSize, const UA_KeyValuePair *params, const UA_Logger *logger){
+    struct in_addr address = addr->sin_addr;
+    UA_Byte *addressVal = (UA_Byte *) &addr->sin_addr;
+
+    if(isIPv4MulticastAddress(addressVal)) {
```
jpfr, line 384: The name of the function is misleading.
You don’t know whether it is a multicast address before calling the method.
Check this outside and only call this method of it is a multicast address.
The same applies to the below functions.

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+
+        IpMulticastRequest ipMulticastRequest;
+        ipMulticastRequest.ipv4.imr_multiaddr = address;
+        /* default outgoing interface ANY */
+        ipMulticastRequest.ipv4.imr_interface.s_addr = UA_htonl(INADDR_ANY);
+
+        UA_String netif = UA_STRING_NULL;
```
jpfr, line 432: You call getNetworkInterfaceParams two times separately.
That looks unnecessary and can be simplified.
Eg by merging the configuration and register steps.

arch/eventloop_posix_udp.c
```diff
@@ -265,6 +765,27 @@ UDP_registerListenSocket(UA_ConnectionManager *cm, UA_UInt16 port,
                 "UDP %u\t| New server socket for \"%s\" on port %u",
```
jpfr, line 765: This is not a server socket (as in TCP).

arch/eventloop_posix_udp.c
```diff
@@ -36,6 +44,524 @@ typedef struct {
...
+                return UA_STATUSCODE_BADINTERNALERROR;
+            }
+        }
+        setsockopt(socket, IPPROTO_IP, IP_MULTICAST_IF,
+                   &ipMulticastRequest.ipv4.imr_interface, sizeof(struct in_addr));
+    }
+    return UA_STATUSCODE_GOOD;
+}
+
+static UA_StatusCode
```
jpfr, line 417: It took me some time to understand the difference between “adding” and “configuring” of multicast sockets.
One is for listening and one for sending.
Call this something like “setupListenMulticast” to be immediately clear.

jpfr (commented):
Super cool!
A couple last comments in-line.

arch/eventloop_posix_udp.c
JanSurft: > Early return instead of wrapping everything into an if clause. Keep the code flat.

Good for pointing this out, there was a bug anyway: if (enableReuse) should just set the reuseValue to 1 else it is 0 and the setting of the option have to be done either way.

arch/eventloop_posix_udp.c
JanSurft: > Early return instead of deep nesting.

If condition can be left out anyway, because the caller asserts != NULL already

arch/eventloop_posix_udp.c
JanSurft: > This does not need a status code it can just return a Boolean. Below also.

Yes, good call. This was still a rudiment of an early version where a status-code was needed.

arch/eventloop_posix_udp.c
```diff
@@ -335,11 +336,15 @@ setConnectionConfigurationProperties(UA_FD socket,
...
+            char keyName[UA_MAXPARAMSTR_LENGTH];
+            size_t nonNullTerminatedLength = prop->key.name.length;
+            if (prop->key.name.length > UA_MAXPARAMSTR_LENGTH - 1) {
+                nonNullTerminatedLength = UA_MAXPARAMSTR_LENGTH - 1;
+            }
+            strncpy(keyName, (char*)prop->key.name.data, nonNullTerminatedLength);
+            keyName[nonNullTerminatedLength] = '\0';
             UA_LOG_WARNING(logger, UA_LOGCATEGORY_SERVER,
-                           "PubSub Connection creation. Unknown connection parameter: '%s'.", keyname);
+                           "PubSub Connection creation. Unknown connection parameter: '%s'.", keyName);
```
jpfr, line 347: You can use the following idiom to print non-null terminated strings.
That might save a bit of accidental complexity.

```c
printf("my string %.*s", (int)str->length, (char*)str->data);
```

jpfr (commented):
Nice! Just two small last things.

arch/eventloop_posix_udp.c
JanSurft: Perfect thanks!

jpfr:
Nice! Merged.

PR #5161, opened 2022-05-31
feat(pubsub): Extended network message content support

keba-estr:
feat(pubsub): Subscriber without payload header

Support subscriber without payload header.
If no payload header is received, only one data set is supported.
It is not possible to check the data set id in this case.

feat(pubsub): Subscriber without group header

Support subscriber without group header.
If no group header or writer group id is received, it is not possible
to check the writer group id in this case. Yet the message should
be received successfully.

### DIFF
--- a/src/pubsub/ua_pubsub_reader.c
+++ b/src/pubsub/ua_pubsub_reader.c
@@ -338,8 +338,7 @@ static UA_StatusCode
 checkReaderIdentifier(UA_Server *server, UA_NetworkMessage *msg,
                       UA_DataSetReader *reader, UA_ReaderGroupConfig readerGroupConfig) {
 if(readerGroupConfig.encodingMimeType != UA_PUBSUB_ENCODING_JSON){
-    if(!msg->groupHeaderEnabled || !msg->groupHeader.writerGroupIdEnabled ||
-       !msg->payloadHeaderEnabled) {
+    if(!msg->groupHeaderEnabled || !msg->groupHeader.writerGroupIdEnabled) {
         UA_LOG_INFO_READER(&server->config.logger, reader,
                            "Cannot process DataSetReader without WriterGroup"
                            "and DataSetWriter identifiers");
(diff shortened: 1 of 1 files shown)


keba-estr:
Is there a function where I have to check that a subscriber without payload header is not used with several data sets?

jpfr:
A subscriber is attached to just one dataset.
You mean that one connection “feeds” multiple subscribers without disambiguating by the id?

keba-estr:
> A subscriber is attached to just one dataset. You mean that one connection “feeds” multiple subscribers without disambiguating by the id?

Yes, I thought about several datasets in one connection. If the incoming network packet is also already attached to the subscriber of the dataset, no further check is necessary.

jpfr:
This makes the surface for mistakes a bit bigger.

Suppose a PubSub connection serves two readers.
If the remote side is not aware of that, they might remove the heaer and suddenly the messages get delivered to the wrong reader as well.

I would like to make this behaviour configurable.

Please add a `RuleHanding` configuration flag to the reader config.
https://github.com/open62541/open62541/blob/47223e8890fc35b83af5799a4deb3342e587082d/include/open62541/common.h#L135

The RuleHandling enum then gives the following config options:
- Allow message without the reader id
- Allow and log a warning
- Disallow
- Default (one of the above)

The default could be "allow with log warning" but I don't have a very strong opinion on the default.

keba-estr:
> This makes the surface for mistakes a bit bigger.
> 
> Suppose a PubSub connection serves two readers. If the remote side is not aware of that, they might remove the heaer and suddenly the messages get delivered to the wrong reader as well.
> 
> I would like to make this behaviour configurable.
> 
> Please add a `RuleHanding` configuration flag to the reader config.
> 
> https://github.com/open62541/open62541/blob/47223e8890fc35b83af5799a4deb3342e587082d/include/open62541/common.h#L135
> 
> The RuleHandling enum then gives the following config options:
> 
> * Allow message without the reader id
> * Allow and log a warning
> * Disallow
> * Default (one of the above)
> 
> The default could be "allow with log warning" but I don't have a very strong opinion on the default.

Should there be a separate rule for 'Subscriber without group header' and for 'Subscriber without payload header', or is the rule only necessary for one of the changes?

keba-estr:
Issue has already been resolved in the meantime by the following commit:
1c8af509fd5b2617cbd3257b6e14b49c8dd95c72 fix(pubsub): reader not receiving msgs wo grouphdr

PR #5162, opened 2022-05-31
feat(pubsub): Call statechange callback

keba-estr:
Call state change callbacks whenever a state changes

Proposal: Add parameter for error cause

Functions *_setPubSubState are extended with parameter for error cause.

-> proposal accepted (see comment below by jpfr)

### DIFF
--- a/src/pubsub/ua_pubsub.h
+++ b/src/pubsub/ua_pubsub.h
@@ -145,8 +145,10 @@ UA_DataSetWriter *
 UA_DataSetWriter_findDSWbyId(UA_Server *server, UA_NodeId identifier);
 
 UA_StatusCode
-UA_DataSetWriter_setPubSubState(UA_Server *server, UA_PubSubState state,
-                                UA_DataSetWriter *dataSetWriter);
+UA_DataSetWriter_setPubSubState(UA_Server *server,
+                                UA_DataSetWriter *dataSetWriter,
+                                UA_PubSubState state,
+                                UA_StatusCode cause);
 
 UA_StatusCode
 UA_DataSetWriter_generateDataSetMessage(UA_Server *server,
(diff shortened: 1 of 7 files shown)


keba-estr:
Is there a state change for a connection?

keba-estr:
Are the proposed status codes ok, or are there better ones / new ones needed?

src/pubsub/ua_pubsub.h
```diff
@@ -146,7 +146,8 @@ UA_DataSetWriter_findDSWbyId(UA_Server *server, UA_NodeId identifier);
 
 UA_StatusCode
 UA_DataSetWriter_setPubSubState(UA_Server *server, UA_PubSubState state,
-                                UA_DataSetWriter *dataSetWriter);
```
jpfr, line 149: Just a small nitpick.
Since we are touching the api, I prefer the following order of the arguments:

server -> writer -> state -> cause

first the context, then the change.

jpfr (commented):
Nice! Just one minor comment.
Can be merged then.

‘the status codes also look good.

src/pubsub/ua_pubsub.h
keba-estr: Order changed.

keba-estr:
What is the correct behaviour of the unit test Test_many_components in check_pubsub_subscribe_msgrcvtimeout?
When the writer groups are disabled with some server run time in between, the moment when the timeout of the reader group occurs seems to be random (according to different runs of the UNIT tests).
Is it sufficient for the UNIT test to disable the writer groups at once and then run the server routines to get a reproducible behaviour?

keba-estr:
> What is the correct behaviour of the unit test Test_many_components in check_pubsub_subscribe_msgrcvtimeout (steps C1 / C2)? When the writer groups are disabled with some server run time in between, the moment when the timeout of the reader group occurs seems to be random (according to different runs of the UNIT tests). Is it sufficient for the UNIT test to disable the writer groups at once and then run the server routines to get a reproducible behaviour?

-> I increased the runtime of the server loops in the unit test check_pubsub_subscribe_msgrcvtimeout (steps C1 / C2) such that the readers corresponding to the disabled writer group always run in a timeout before disabling the next writer group.

jpfr:
Great job!

PR #5166, opened 2022-06-01
fix(client): set connect status after disconnect

bluhm:
After calling UA_Client_disconnectAsync(), it is unclear how often
UA_Client_run_iterate() should be called.  Even worse, if it is
called to often, a new connection is established automatically.
This should only happen if another UA_Client_connectAsync() is
called.  Version 1.0 of open62541 did set client connectStatus to
UA_STATUSCODE_BADCONNECTIONCLOSED after disconnect had finished.
This status code is also returned by UA_Client_run_iterate().
Restore the behavior to use asynchronus disconnect reliably.


jpfr:
This should be in closeSecureChannel so ist works also for closing SecureChannels without a Session.

bluhm:
I came to the conclusion that this is not the correct fix for my problem.

PR #5167, opened 2022-06-02
fix(pub/sub): the socket option SO_PRIORITY should only for Linux

wenshan1:
The socket option, SO_PRIORITY is not mentioned in the posix standard.
It is a specifical implementation of Linux.

### DIFF
--- a/plugins/ua_pubsub_ethernet.c
+++ b/plugins/ua_pubsub_ethernet.c
@@ -824,6 +824,7 @@ UA_PubSubChannelEthernet_open(const UA_PubSubConnectionConfig *connectionConfig)
         return NULL;
     }
 
+#ifdef __linux__
     /* Setting the socket priority to the socket */
     if(sockOptions.socketPriority) {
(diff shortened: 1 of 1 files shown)


jpfr:
Great!
Can you add a warning if the setting is set but cannot be applied on non-Linux?

wenshan1:
I still think it should be protected by #ifdef __linux__.
If just add a warning,  OPEN62541 will fail to compile in some posix OS which don't support the sock option SO_PRIORITY.

It is protected by #ifdef __linux__ in this file, https://github.com/open62541/open62541/blob/1.3/plugins/ua_pubsub_udp.c#L378.

jpfr:
Yes, it should be protected by `#ifdef __linux__`.
Still, if somebody tries to set this configuration and he is not on Linux, then there should be a warning about an unsupported config option.

wenshan1:
This file is only used for Linux, https://github.com/open62541/open62541/blob/master/CMakeLists.txt#L382. 
Closed it.

PR #5169, opened 2022-06-02
[WIP] feature: Support for multiple PKI infrastructures

mlgiraud:
This PR adds support for multiple PKI infrastructures.
In OPC UA a PKI infrastructure corresponds to a CertificateGroup. A CertificateGroup has an issuer authority that issues certificates. This PR adds support such that a server can belong to multiple different CertificateGroups. Endpoints can be assigned to a specific CertificateGroup.

These features require changes to the public and private key loading and to some security related code.


andreasebner:
@mlgiraud Thank you for the PR! It would be great if you can add a small comment to support others to understand your changes ;)

jpfr:
This is now implemented via the SecurityGroup feature.

PR #5173, opened 2022-06-04
fix(plugin): Add default limits for chunks and message size

jpfr:
Based on a reported DoS vulnerability reported by Team82 (Claroty
Research).

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -59,13 +59,13 @@ UA_Server_new() {
 /*******************************/
 
 const UA_ConnectionConfig UA_ConnectionConfig_default = {
-    0,     /* .protocolVersion */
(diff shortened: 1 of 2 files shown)


peschuster:
Will there be a CVE for this issue? @jpfr
Is this also exploitable without a valid, trusted client certificate?
I assume systems with less than 512 MB memory will need to further reduce the max. settings or how does this config option affect memory consumption?

jpfr:
Yes, systems with less memory need to adjust.
But we do already cover malloc returning NULL when the system is out of memory.

I will post the CVE once I receive it.
Making the CVE is up to the group that reported the issue.

jpfr:
FYI, here is the CVE: https://nvd.nist.gov/vuln/detail/CVE-2022-25761

PR #5179, opened 2022-06-08
feat(pubsub): Implement StandaloneSubscribedDataSet

Fischer-Thomas:
- There is an example on how to use the StandaloneSubscribedDataSet. 
- There is no unit test. (help wanted)


Fischer-Thomas:
@andreasebner

Issue #5182, opened 2022-06-09
SegmentationFault when monitoring UnionDataTypes with SwitchField set to NONE


Fischer-Thomas:
## Description

Server crashes when trying to monitor a UnionDataType that has the SwitchField set to NONE.
Reading without a subscription is working fine.

Callstack:
> ==8340== Process terminating with default action of signal 11 (SIGSEGV)
==8340==  Access not within mapped region at address 0x18007C4330
==8340==    at 0x12608A: unionOrder (ua_types.c:1629)
==8340==    by 0x125500: variantOrder (ua_types.c:1446)
==8340==    by 0x1261BE: UA_order (ua_types.c:1684)
==8340==    by 0x681CD1: detectValueChange (ua_subscription_datachange.c:124)
==8340==    by 0x681E0C: sampleCallbackWithValue (ua_subscription_datachange.c:159)
==8340==    by 0x68209D: monitoredItem_sampleCallback (ua_subscription_datachange.c:229)
==8340==    by 0x681FBD: UA_MonitoredItem_sampleCallback (ua_subscription_datachange.c:204)
==8340==    by 0x689EBB: timerExecutionTrampoline (eventloop_posix.c:21)
==8340==    by 0x698C4E: UA_Timer_process (ua_timer.c:262)
==8340==    by 0x68A485: UA_EventLoopPOSIX_run (eventloop_posix.c:236)
==8340==    by 0x129F55: UA_Server_run_iterate (ua_server.c:855)
==8340==    by 0x12A157: UA_Server_run (ua_server.c:933)

## Background Information / Reproduction Steps

In the .zip archive i have the two examples "custom_server_types" and "client_subscription_loop" which i adjusted to recreate this issue.

[segfault_union_examples.zip](https://github.com/open62541/open62541/files/8867714/segfault_union_examples.zip)

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): current master

jpfr:
Fixed by @Fischer-Thomas via #5184.

PR #5184, opened 2022-06-09
fix(server): subscription of UnionDataTypes

Fischer-Thomas:
fix for #5182


jpfr:
Uh. Good find.
Thanks!

PR #5189, opened 2022-06-10
fix(ci) set correct release number

andreasebner:


mlang-de:
May bump direct to 1.3.2 since 1.3.1 is already released?

andreasebner:
Currently the cmake file defines: 
# The current version information. On the master branch, we take the version
# number from the latest release plus the "-undefined" label. Will be
# overwritten with more detailed information if git is available.

--> Therefore the version should be the last released version.

Issue #5191, opened 2022-06-10
Shared library linking fails if compiler is Clang and build type is debug


leoplo:
## Description

```bash
[100%] Linking C shared library bin/libopen62541.so
/usr/bin/ld: __sancov_guards has both ordered [`__sancov_guards[UA_Array_appendCopy]' in /tmp/lto-llvm-14c7a9.o] and unordered [`__sancov_guards' in /tmp/lto-llvm-14c7a9.o] sections
/usr/bin/ld: final link failed: bad value
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [CMakeFiles/open62541.dir/build.make:236: bin/libopen62541.so.1.3.1] Error 1
make[2]: Leaving directory '/home/olimex/open62541/build'
make[1]: *** [CMakeFiles/Makefile2:465: CMakeFiles/open62541.dir/all] Error 2
make[1]: Leaving directory '/home/olimex/open62541/build'
make: *** [Makefile:149: all] Error 2
```

## Background Information / Reproduction Steps

I wanted to compile open62541 with ASAN and I noticed that [`-fsanitize=address` is only added](https://github.com/open62541/open62541/blob/master/CMakeLists.txt#L767) for `Debug` build type if [`CMAKE_C_COMPILER_ID` is `Clang`](https://github.com/open62541/open62541/blob/master/CMakeLists.txt#L763).

So I defined `clang` as my default compiler.

By the way I do not get why it is not possible to use `-fsanitize=address` with `gcc` . (maybe because it is not compatible with `-fsanitize=leak` ?)

Configure Clang as default compiler:

```bash
sudo update-alternatives --install /usr/bin/cc cc /usr/bin/clang 100
sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++ 100
```

Used CMake options:

```bash
cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug ..
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): commit 90ea2f6d (master branch)
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Linux (Debian 5.10.113-1 (2022-04-29) x86_64)
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

[...]

PR #5192, opened 2022-06-10
fix(cmake): check if CMAKE_CXX_COMPILER_VERSION is defined

leoplo:
fix https://github.com/open62541/open62541/issues/5191


jpfr:
Thanks!

PR #5193, opened 2022-06-10
fix(tests): network is not instant

bluhm:
The client itereate call without timeout in the tests works only
if the TCP connect is instant.  That seems to be the case for github
Linux continuous integration.  But it may fail on other operating
systems or in real live.  Set the iterate timeout after conntect
to 1 second.  That should be enough for loopback interface.

Also sending an async browse request may not reach the server before
iterating it.  A sleep(0) to reschedule the process is enough for
my setup.  Then the server can read the request from its socket.


jpfr:
Thanks.

PR #5204, opened 2022-06-15
fix(client) connect iterate without timeout

bluhm:
Commit ef4394b1144e845df93760b951ef0f6bef63d053 may have fixed the
connect timeout, but it broke async connect without timeout.  That
made it impossible to continue processing during TCP handshake.
While the connect operation is in progress, the select system call
will return immediately if the timeout is 0.  Do not close the
connection in this case, but allow the application to call run_iterate()
again.  For the tests in the CI this makes no difference as connect
over Linux loopback interface succeeds immediately.


jpfr:
Thanks for this.
I rebased your PR on the 1.3 branch for the next patch release.
From there it will go to master.

bluhm:
I have forgotten a UA_Server_run_iterate() in the test.  I will push a fixed commit to avoid a hang in check_client_async_connect.

jpfr:
Nice!

Issue #5205, opened 2022-06-16
Issue when reading a range when using ValueBackend External


jonasgreen88:
Hi,

I'm evaluating to change from the legacy DataSource callbacks to the new ValueBackend external. However, I belive there is a fault in the implementation that extracts the requested range. `external.value` is a pointer to an UA_DataValue pointer. But the implementation takes the reference of `external.value` and casts it to a UA_Variant as shown below.

https://github.com/open62541/open62541/blob/7db79624d09b7c1e2176d31d41e1232d3ba38a44/src/server/ua_services_attribute.c#L258

Shouldn't it be `(const UA_Variant *)&(*vn->valueBackend.backend.external.value)->value` to get the variant pointer?

If you agree I can make a PR.

PR #5206, opened 2022-06-16
fix(server): Reading a range when using ValueBackend External failed

jonasgreen88:
Changed to first read the complete DataValue to get timestamps and
status. Then if a range is requested the requested range of
the variant is copied into the DataValue.

#5205

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -253,10 +253,10 @@ readValueAttributeComplete(UA_Server *server, UA_Session *session,
             }
             /* Set the result */
             if(rangeptr)
-                return UA_Variant_copyRange(
-                    (const UA_Variant *)&vn->valueBackend.backend.external.value,
-                    &v->value, *rangeptr);
-            retval = UA_DataValue_copy(*vn->valueBackend.backend.external.value, v);
+                retval = UA_DataValue_copyVariantRange(
+                    *vn->valueBackend.backend.external.value, v, *rangeptr);
+            else
+                retval = UA_DataValue_copy(*vn->valueBackend.backend.external.value, v);
             break;
         case UA_VALUEBACKENDTYPE_NONE:
             /* Read the value */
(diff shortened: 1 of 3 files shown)


src/server/ua_services_attribute.c
```diff
@@ -253,11 +253,14 @@ readValueAttributeComplete(UA_Server *server, UA_Session *session,
...
-            if(rangeptr)
-                return UA_Variant_copyRange(
-                    (const UA_Variant *)&vn->valueBackend.backend.external.value,
-                    &v->value, *rangeptr);
             retval = UA_DataValue_copy(*vn->valueBackend.backend.external.value, v);
+
+            if(rangeptr) {
+                UA_Variant_clear(&v->value);
+                retval |= UA_Variant_copyRange(
+                    (const UA_Variant *)&(*vn->valueBackend.backend.external.value)->value,
```
jpfr, line 261: No need to cast to a variant.

src/server/ua_services_attribute.c
```diff
@@ -253,11 +253,14 @@ readValueAttributeComplete(UA_Server *server, UA_Session *session,
                 break;
             }
             /* Set the result */
-            if(rangeptr)
-                return UA_Variant_copyRange(
-                    (const UA_Variant *)&vn->valueBackend.backend.external.value,
-                    &v->value, *rangeptr);
             retval = UA_DataValue_copy(*vn->valueBackend.backend.external.value, v);
```
jpfr, line 256: This can be inefficient.
Making a full copy just to clear the variant.
You could instead make a “shallow” copy of the DataValue and _init the variant portion before calling copyrange.

jpfr (commented):
Nice!
Two minor comments inline.
Please make this a fix for the 1.3 branch.
We then pull everything over to master.

src/server/ua_services_attribute.c
jonasgreen88: Thanks for the comments. Now pushed on 1.3 branch and made a copyVariantRange function for the DataValue type.

jpfr:
Nice!

PR #5207, opened 2022-06-18
build(pack): add option UA_ENABLE_SANITIZER

theirix:
Added an option `UA_ENABLE_SANITIZER` that allows to disable address sanitizers if needed. By default it is still on.


jpfr:
Good idea.
Please call it UA_ENABLE_DEBUG_SANITIZER.
So we can in the future prefix/group debug related options.
Also mark it as an advanced option.

PR #5213, opened 2022-06-21
feat(pubsub): Pass the server pointer to the PubSub StateChangeCallback

MurzynJ:
Allow applications to access the server data inside the callback. Since the server config can store also an application context, also the application data can be accessed from within the callback.


jpfr:
Looks good!

PR #5216, opened 2022-06-23
fix(server): Fix array dimensions when copying range of variant

jonasgreen88:
The array dimensions of the destionation variant was set to the
dimensions of the source variant. Now the block size calculated
of the requested range is used.

### DIFF
--- a/src/ua_types.c
+++ b/src/ua_types.c
@@ -852,8 +852,7 @@ UA_Variant_copyRange(const UA_Variant *src, UA_Variant * UA_RESTRICT dst,
         }
         dst->arrayDimensionsSize = thisrange.dimensionsSize;
         for(size_t k = 0; k < thisrange.dimensionsSize; ++k)
-            dst->arrayDimensions[k] =
-                thisrange.dimensions[k].max - thisrange.dimensions[k].min + 1;
+            dst->arrayDimensions[k] = (UA_UInt32)block;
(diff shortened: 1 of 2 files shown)


jpfr:
Um. The previous code looks correct.
The length of the dimension is range.max - range.min + 1.
The `block` variable is a bit more difficult to understand.

Does one of the new tests fail with the old code?

jonasgreen88:
> Um. The previous code looks correct. The length of the dimension is range.max - range.min + 1. The `block` variable is a bit more difficult to understand.
> 
> Does one of the new tests fail with the old code?

Yes. The new test I added `copyArrayRangeUpperBoundOutOfRange()` fails without this change implemented. range.max - range.min + 1 expects the requested range to be completely within range. If the upper bound is out of range this is not considered. I think block contains the actual number of elements of the requested range that is within range of the actual value.

jpfr:
I made a version that
- makes a copy of the range
- adjusts the range to match the actual array dimensions
- uses the adjusted range to compute the array dimensons of the result

See the commit added on top of this.

jonasgreen88:
> Ah. I get what you mean. However block is a scalar. So we'd have the same size for every dimension. That can't work...

Yes, I realized that yesterday as well when discussing this issue. Thanks for developing a fix that works for all dimensions!

PR #5220, opened 2022-06-24
test(pubsub): CI - activate UA_ENABLE_PUBSUB_FILE_CONFIG

keba-estr:
Add build option UA_ENABLE_PUBSUB_FILE_CONFIG to CI unit-tests.

Fixes have been moved to pull request #5241.


keba-estr:
To discuss:
This commit is part of pull request #4268.
Yet I don't know whether only the errors should be fixed or also the defines should be set, as the coverage decreases.

keba-estr:
> To discuss: This commit is part of pull request #4268. Yet I don't know whether only the errors should be fixed or also the defines should be set, as the coverage decreases.

The activation does not seem to be necessary.

PR #5221, opened 2022-06-24
fix: remove trailing spaces in code

keba-estr:
Some changes in pull request #4268 contained removed trailing spaces, so I removed them in internal code (not in /deps) for all files with a separate commit.


jpfr:
👍

PR #5234, opened 2022-07-05
[Review] feat(pubsub): Implement MQTT subscriber

NoelGraf:
Implement an MQTT subscriber that subscribes to the 'customTopic' topic of the published example.

- [x] Implement MQTT subscriber
- [x] Implement JSON support for subscription
- [x] Adding a test case for MQTT subscription
- [x] Adding a test case for MQTT publishing

Note:
The decodeJsonInternal method decode whether the publisherID is a number or a string. If the PublisherID is a number, the type is set to UInt64 so that the largest possible number can be stored. The problem is that with a UInt64, a string is expected in the decoding.
The current solution is to set the type to UInt32 for a number.


examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,389 @@
...
+
+#define CONNECTION_NAME              "MQTT Subscriber Connection"
+#define TRANSPORT_PROFILE_URI        "http://opcfoundation.org/UA-Profile/Transport/pubsub-mqtt"
+#define MQTT_CLIENT_ID               "TESTCLIENTPUBSUBMQTTSUBSCRIBE"
+#define CONNECTIONOPTION_NAME        "mqttClientId"
+#define SUBSCRIBER_TOPIC              "customTopic"
+#define SUBSCRIBER_METADATAQUEUENAME  "MetaDataTopic"
+#define SUBSCRIBER_METADATAUPDATETIME 0
+#define BROKER_ADDRESS_URL           "opc.mqtt://127.0.0.1:1883"
+#define SUBSCRIBE_INTERVAL             500
```
andreasebner, line 21: Please align the fields ;)

examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
```diff
@@ -0,0 +1,389 @@
...
+
+    /* configure options, set mqtt client id */
+    const int connectionOptionsCount = 1
+#ifdef EXAMPLE_USE_MQTT_LOGIN
+    + 2
+#endif
+#ifdef EXAMPLE_USE_MQTT_TLS
+    + 2
+#endif
+    ;
```
andreasebner, line 85: Could we add a define here? Seems like a magic number.

src/pubsub/ua_pubsub.h
```diff
@@ -562,6 +562,14 @@ UA_StatusCode
 receiveBufferedNetworkMessage(UA_Server *server, UA_ReaderGroup *readerGroup,
                               UA_PubSubConnection *connection);
 
+void process_callback(UA_Server *server, UA_ReaderGroup *readerGroup,
+                      UA_PubSubConnection *connection, UA_ByteString *msg,
+                      UA_ByteString *topic);
```
andreasebner, line 567: If it's only used for mqtt, than we should include "mqtt" or "broker" in the name.

src/pubsub/ua_pubsub_manager.h
```diff
@@ -28,11 +34,17 @@ typedef struct UA_PubSubManager {
...
+    TAILQ_HEAD(UA_ListOfTopicAssign, UA_TopicAssign) topicAssign;
+
 #ifndef UA_ENABLE_PUBSUB_INFORMATIONMODEL
     UA_UInt32 uniqueIdCount;
 #endif
 } UA_PubSubManager;
 
+UA_StatusCode
+UA_Server_addPubSubTopicAssign(UA_Server *server, UA_ReaderGroup *readerGroup, UA_String topic);
+
```
andreasebner, line 47: Maybe UA_PubSubManager_ ... would be more precise.

tests/pubsub/check_pubsub_publish_mqtt.c
```diff
@@ -0,0 +1,174 @@
...
+    Suite *s = suite_create("PubSub publishing via mqtt");
+    suite_add_tcase(s, tc_pubsub_publish_mqtt);
+
+    SRunner *sr = srunner_create(s);
+    srunner_set_fork_status(sr, CK_NOFORK);
+    srunner_run_all(sr,CK_NORMAL);
+    int number_failed = srunner_ntests_failed(sr);
+    srunner_free(sr);
+    return (number_failed == 0) ? EXIT_SUCCESS : EXIT_FAILURE;
+}
```
andreasebner, line 174: Is there an aspect which is not covered by the check_pubsub_subscribe test? Maybe we should combine it to one unit test.

tests/CMakeLists.txt
```diff
@@ -343,6 +343,8 @@ if(UA_ENABLE_PUBSUB)
     if(UA_ENABLE_PUBSUB_MQTT)
         if(NOT WIN32)
             ua_add_test(pubsub/check_pubsub_connection_mqtt.c)
+            #ua_add_test(pubsub/check_pubsub_publish_mqtt.c)
+            #ua_add_test(pubsub/check_pubsub_subscribe_mqtt.c)
```
andreasebner, line 347: Is that intentional?

include/open62541/server_pubsub.h
```diff
@@ -174,6 +174,7 @@ struct UA_PubSubConnectionConfig {
     size_t connectionPropertiesSize;
     UA_KeyValuePair *connectionProperties;
     UA_Variant connectionTransportSettings;
+    UA_Server *server;
```
andreasebner, line 177: An alterntive could be to remove this flag here and store it internaly during the UA_Server_addPubSubConnection call.

src/pubsub/ua_pubsub_manager.h
andreasebner: Another question: Is the assign done by the user or implied by the PubSub-config? -> Why is this function needed?

plugins/mqtt/ua_mqtt-c_adapter.c
```diff
@@ -328,37 +329,42 @@ disconnectMqtt(UA_PubSubChannelDataMQTT* channelData){
...
+        UA_ByteString_delete(topic);
+        UA_free(msg);
+        return;
+    }
+    /* copy topic and msg, call the cb */
+    memcpy(topic->data, published->topic_name, published->topic_name_size);
+    memcpy(msg->data, published->application_message, published->application_message_size);
+
+    UA_TopicAssign *topicAssign1, *topicAssign2;
+    TAILQ_FOREACH_SAFE(topicAssign1,  &pubSubManager.topicAssign,
```
andreasebner, line 362: Is the "safe" version needed here?

Issue #5236, opened 2022-07-05
FreeBSD 13.1: Implicit declaration of pipe2


herrhotzenplotz:
## Description

```console
$ git show | head -1
commit 8c64fabafd6f279bff77e7a9a4261d82189e816c
$ uname -a
FreeBSD hades.herrhotzenplotz.de 13.1-RELEASE FreeBSD 13.1-RELEASE releng/13.1-n250148-fc952ac2212 GENERIC amd64
$ freebsd-version -kru
13.1-RELEASE
13.1-RELEASE
13.1-RELEASE
$ 
```

Fails with:
```console
open62541/arch/eventloop_posix_interrupt.c:364:15: error: implicit declaration of function 'pipe2' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    int err = pipe2(pipefd, O_NONBLOCK);
              ^
open62541/arch/eventloop_posix_interrupt.c:364:15: note: did you mean 'pipe'?
/usr/include/unistd.h:358:6: note: 'pipe' declared here
int      pipe(int *);
         ^
```

Maybe not as C99-ish as claimed :-)

## Background Information / Reproduction Steps

Used CMake options:

```console
$ cmake CC=/usr/bin/cc CXX=/usr/bin/c++ -S. -Bbuild-freebsd131-amd64 -GUnix\ Makefiles -DCMAKE_INSTALL_PREFIX=/opt/sn
$ cpuset -l2-3 make -Cbuild-freebsd131-amd64 -j 2 VERBOSE=1 2>&1 | tee ../open62541.build-boom.log
```

## Notes

Sort of a hack, but it fixes the build issues:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index af36586b..07c97854 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -688,6 +688,9 @@ endfunction()
 
 if(NOT UA_FORCE_CPP AND (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang"))
     check_add_cc_flag("-std=c99")   # C99 mode
+    if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+        check_add_cc_flag("-D__BSD_VISIBLE=1")
+    endif()
     check_add_cc_flag("-pipe")      # Avoid writing temporary files (for compiler speed)
     check_add_cc_flag("-Wall")      # Warnings
[...]

PR #5238, opened 2022-07-06
(#5236) Fix FreeBSD builds

herrhotzenplotz:
As discussed in #5236 this is a workaround for the hidden pipe2 syscall on FreeBSD in pure C99 environments.


herrhotzenplotz:
WIP notes:

```console
$ /usr/bin/cc -E -H -dM -DUA_ARCHITECTURE_POSIX -DUA_DYNAMIC_LINKING_EXPORT <snip> open62541/arch/eventloop_posix_interrupt.c
. open62541/arch/eventloop_posix.h
.. open62541/build-freebsd131-amd64/src_generated/open62541/config.h
... open62541/arch/posix/ua_architecture.h
<snip>
#define _XOPEN_SOURCE 600
```

Definitely `config.h` is being included first and `_XOPEN_SOURCE` is defined to `600`. There is something else very wrong here.

herrhotzenplotz:
After some more digging I found out that my build success with `_XOPEN_SOURCE=600` was a spurious cmake hiccup. 

The main problem is that `pipe2` is not defined by POSIX and thus not available in compilation environments that enforce strict compliance by defining previously mentioned macro.

The two clean solutions are therefore:
- remove the `#define _XOPEN_SOURCE 600` (which I don't favour)
- use `pipe()` and then mark it as non-blocking using `fcntl` (better)

The hack would be to define the C std library internal preprocessor macro `__BSD_VISIBLE`. 

I am unsure....

jpfr:
I like the pure posix solution.
Can we reuse the code to mark a socket non blocking also for posix udp and tcp?
There is a posix header where we could declare the function.

herrhotzenplotz:
Sure, I will shuffle around the commit and try adapting other places to this later today. I am not sure whether `fcntl` is a safe way to mark sockets as non-blocking on all platforms though since the usual way to do this with sockets is to use `setsockopt`. The man page on FreeBSD doesn't say anything regarding that  - I will see what happens.

herrhotzenplotz:
After rebasing on the current master I see that in the meantime 961ac08b introduced even more unportable changes:


```
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:47:20: error: field has incomplete type 'struct ip_mreq'
    struct ip_mreq ipv4;
                   ^
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:47:12: note: forward declaration of 'struct ip_mreq'
    struct ip_mreq ipv4;
           ^
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:187:68: error: use of undeclared identifier 'IP_MULTICAST_LOOP'
                     ai_family == AF_INET6 ? IPV6_MULTICAST_LOOP : IP_MULTICAST_LOOP,
                                                                   ^
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:212:35: error: use of undeclared identifier 'PF_INET6'
                     ai_family == PF_INET6 ? IPPROTO_IPV6 : IPPROTO_IP,
                                  ^
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:213:35: error: use of undeclared identifier 'PF_INET6'
                     ai_family == PF_INET6 ? IPV6_MULTICAST_HOPS : IP_MULTICAST_TTL,
                                  ^
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:213:68: error: use of undeclared identifier 'IP_MULTICAST_TTL'
                     ai_family == PF_INET6 ? IPV6_MULTICAST_HOPS : IP_MULTICAST_TTL,
                                                                   ^
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:387:36: error: use of undeclared identifier 'IP_MULTICAST_IF'
    setsockopt(socket, IPPROTO_IP, IP_MULTICAST_IF,
                                   ^
/usr/home/nico/src/open62541/arch/eventloop_posix_udp.c:426:42: error: use of undeclared identifier 'IP_ADD_MEMBERSHIP'
    if(UA_setsockopt(socket, IPPROTO_IP, IP_ADD_MEMBERSHIP,
                                         ^
7 errors generated.
*** Error code 1

Stop.
make[2]: stopped in /usr/home/nico/src/open62541/build-freebsd131-amd64
*** Error code 1

Stop.
make[1]: stopped in /usr/home/nico/src/open62541/build-freebsd131-amd64
*** Error code 1

Stop.
make: stopped in /usr/home/nico/src/open62541/build-freebsd131-amd64
```

`struct ip_mreq` seems to only visible in BSD mode... furthermore
it seems that POSIX only mentions `struct ipv6_mreq`...or am I
missing something here? In any case the patch is flawed.

jpfr:
FreeBSD has
```
struct vifctl vc;
struct mif6ctl mc;
```
for multicast. These appear to be available on Linux as well.
But not on Windows.

jpfr:
We should add the BSDs to our CI.
There will be regular breakage without it.

jpfr:
Thanks!

Issue #5239, opened 2022-07-07
Missing nullptr check for channel


keba-estr:
## Description

The action 'Debug Build & Unit Tests (clang)' fails sporadically (check_pubsub_informationmodel_methods, see https://github.com/open62541/open62541/runs/7227268011?check_suite_focus=true).

## Background Information

Part of message from action:

> 68: /home/runner/work/open62541/open62541/src/server/ua_server.c:848:31: runtime error: member access within null pointer of type 'UA_PubSubChannel' (aka 'struct UA_PubSubChannel')

In ua_server.c line 848 the variable ps->channel should be checked for nullptr before accessing it.
Is there a special handling/error message necessary if the channel is nullptr?

Maybe this issue is not relevant any more after the open integration into the EventLoop.

PR #5240, opened 2022-07-07
fix(server): check for valid channel before access

keba-estr:
Unit test check_pubsub_informationmodel_methods failed
sporadically because of a nullptr access (see issue #5239).

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -845,7 +845,7 @@ UA_Server_run_iterate(UA_Server *server, UA_Boolean waitInternal) {
     UA_PubSubConnection *connection;
     TAILQ_FOREACH(connection, &server->pubSubManager.connections, listEntry){
         UA_PubSubConnection *ps = connection;
-        if(ps && ps->channel->yield){
+        if(ps && ps->channel && ps->channel->yield){
             ps->channel->yield(ps->channel, 0);
         }
     }


jpfr:
👍

ivan-ushakov:
We faced the same issue when we create/destroy and create again PubSub connection:

```
Thread 1 "opcua-server" received signal SIGSEGV, Segmentation fault.
0x00007ff185043873 in UA_Server_run_iterate (server=0x55ae84545800, waitInternal=true) at /home/ivan/opcua-server/third-party/open62541/src/server/ua_server.c:848
848	        if(ps && ps->channel->yield){
```

But I have another question. When we disconnect with `UA_Server_removePubSubConnection` we use mutex:
```
UA_StatusCode
UA_Server_removePubSubConnection(UA_Server *server, const UA_NodeId connection) {
    UA_LOCK(&server->serviceMutex);
    UA_StatusCode res = removePubSubConnection(server, connection);
    UA_UNLOCK(&server->serviceMutex);
    return res;
}
```

but when we iterate through connections, we don't use mutex:
```
#if defined(UA_ENABLE_PUBSUB_MQTT)
    UA_PubSubConnection *connection;
    TAILQ_FOREACH(connection, &server->pubSubManager.connections, listEntry){
        UA_PubSubConnection *ps = connection;
        if(ps && ps->channel->yield){
            ps->channel->yield(ps->channel, 0);
        }
    }
#endif
```
Is it safe?

PR #5241, opened 2022-07-07
fix(pubsub): correct errors in file config

keba-estr:
Correct usage of server mutex in ua_pubsub_config.c.

Fix wrong path to test .bin files.

Add a length check after loadFile(), so it's
easier to identify an error (e.g. test failed because
file path is wrong or decoding/applying pubsub configuration
does not work)


keba-estr:
This pull request contains fixes needed for #5220 (activate UA_ENABLE_PUBSUB_FILE_CONFIG) without enabling the compile option, as this would decrease the test coverage.

jpfr:
nice!

PR #5242, opened 2022-07-07
feat(server): add ReferenceTypes HasArgumentDescription...

dapaulid:
…and HasOptionalInputArgumentDescription

This allows applications to add variable nodes for method input arguments,
e.g. to provide default values and engineering units.
(see OPC 10000-3, 5.7.2 - 5.7.3)

### DIFF
--- a/src/server/ua_server_ns0.c
+++ b/src/server/ua_server_ns0.c
@@ -160,6 +160,12 @@ UA_Server_createNS0_base(UA_Server *server) {
     ret |= addReferenceTypeNode(server, "HasInterface", "InterfaceOf",
                          UA_NS0ID_HASINTERFACE, false, false, UA_NS0ID_NONHIERARCHICALREFERENCES);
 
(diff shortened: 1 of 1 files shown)


dapaulid:
Given the failing test, it seems like my change introduces a conflict with the `UA_NAMESPACE_ZERO=FULL` option...

Is there any rule for how to decide whether a ReferenceType is to be "hardcoded" in `UA_Server_createNS0_base` (like "HasComponent") or rather generated according to the `UA_NAMESPACE_ZERO` option?

jpfr:
There is a list of the bootstrapped nodes in the nodeset compiler folder.
https://github.com/open62541/open62541/blob/e7d26d483b3511fe33125e00c39b6ea5d16f702d/tools/nodeset_compiler/NodeID_NS0_Base.txt#L1

andreasebner:
Thanks for sharing—landed via #8264.

PR #3255, opened 2019-11-07
[Draft] feature(sks): SKS and pubsub integration

hzw53:
Added an implementation of the security key service for the end-to-end
encryption of pubsub communication.

Added SKS implementation under src/pubsub/ua_pubsub_sks.c.
Functions common to sks and publisher / subscribers are implemented in src/pubsub/ua_pubsub_security.c.
Added examples to examples/sks for SKS server, publisher and subscriber
Added unit tests to tests/sks.
Added AES128 and AES256 CTR security policies.

What is supported at the moment:
* All SKS methods
* Publisher and subscriber will connect to the SKS, call GetSecurityKeys
  when necessary and use the returned keys for encryption/decryption
* Publisher and subscriber provide a SetSecurityKeys method. The user
  has to call this method by himself, the SKS won't do this at the moment
* Encryption and decryption are not supported for json

PR #5243, opened 2022-07-07
[Review]feat(pubsub): KeyStorage and SetSecurityKeys

Muddasir-Shakil:
This PR adds the basic support for SKS feature. It adds a Keystorage to store and maintain the life cycle of the keys used to secure the message and implementation of setSecurityKeys method node to push the keys to Publishers and Subscribers.

Thanks to the author of #3255, I based my work on this PR.

The ua_pubsub_keystorage provides the keystorage structure and interface to interact with the keystorage. The setSecurityKeys
method callback is added to ua_pubsub_ns0. The unit tests are added to tests/pubsub/check_pubsub_keystorage.c and tests/pubsub/check_pubsub_sks_push.c

## What is supported?
- [x] Keystorage to store the keyLists (current and future keys)
- [x] Storing/updating keys in the keyStorage
- [x] Automatic rollover to next key from local keysList
- [x] Setting keys for encryption/decryption and singing/verification in the channel contexts of the associated PubSub Groups 
- [x] SetSecurityKeys method node callback to push the keys
- [x] unit tests for keystorage and setSecurityKey Method Node
- [x] Integration of sks build and keystorage in the CI pipeline
## What is missing ?

- Checking user credentials against SecurityGroup user access rights when calling setSecurityKeys
- AddSecurityGroup Method Node 
- RemoveSecurityGroup Method Node
- GetSecurityGroup Method Node
- GetSecurityKeys Method Node
- Write SecurityGroupId and SecurityKeyServices Url in PubSub Group Object Node


.github/workflows/build_linux.yml
```diff
@@ -41,6 +41,9 @@ jobs:
           - build_name: "PubSub Encryption (MbedTLS) Build & Unit Tests (gcc)"
             cmd_deps: sudo apt-get install -y -qq libmbedtls-dev
             cmd_action: unit_tests_encryption_mbedtls_pubsub
+          - build_name: "PubSub SKS Build & Unit Tests (gcc)"
+            cmd_deps: sudo apt-get install -y -qq libmbedtls-dev
+            cmd_action: unit_tests_pubsub_sks
```
andreasebner, line 46: First of all, great work! Is there a special reason to have a dedicated github runner? I would preffer to integrate these tests into one of the exisitng encryption tests.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,465 @@
...
+            if(iterator->key.data) {
+                UA_free(iterator->key.data);
+                head = iterator;
+                iterator = iterator->next;
+                UA_free(head);
+            } else
+                break;
+        }
+        keyStorage->keyListSize = 0;
+
```
andreasebner, line 85: Please avoid this empty lines (check it for the whole PR). Sometimes the seperation make sense to make the code more readable, but this should be the exception.

src/server/ua_server_internal.h
```diff
@@ -172,6 +173,10 @@ struct UA_Server {
     UA_PubSubManager pubSubManager;
 #endif
 
+#if defined(UA_ENABLE_PUBSUB_ENCRYPTION) && defined(UA_ENABLE_PUBSUB_SKS)
+    LIST_HEAD(PubSubKeyList, UA_PubSubKeyStorage) pubSubKeyList;
+#endif
+
```
andreasebner, line 179: Could you please move the "pubSubKeyList" into the "pubSubManager"?

andreasebner (commented):
Creat work! Added some minor comments. Could you also please make an interactive rebase to sync your version with the latest master and reduce the number of commits?

.github/workflows/build_linux.yml
Muddasir-Shakil: Thank you for the review. With SKS support we will have two mechanims to set the security keys. If Encryption is ON then we have no problem but if Encryption and SKS is ON then the keys has to be set from the keystorage. Also in Pull model, which will be the next PR, it requires the keys to be fetched from SKS before writer/reader group can be set operational. When they are operational, check the current key in key storage and set the encrypting keys.

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: Thank you for pointing it out. I will check the empty lines, but in most cases I add to keep the context of the code clear.

src/server/ua_server_internal.h
Muddasir-Shakil: Yes, I also realized it later that it is better suited in PubSubManager. I did the same thing for SecurityGroups  managed by the PubSub Manager ( not yet Pushed)

Muddasir-Shakil:
> Great work! Added some minor comments. 

Thank you @andreasebner  for the comments

>Could you also please make an interactive rebase to sync your version with the latest master and 

Yes, I will do. While I was working on SecurityGroups and SKS Pull, I also found some minor improvements that I will try to merge here.
>reduce the number of commits?

Now that is reviewed I can squash relatable parts like all the unit tests into one commit for a related feature. This should shorten the commit numbers

Muddasir-Shakil:
Hi @andreasebner 
I update the PR according to your comments and merge the commits to shorten the number of commits. Some additional minor improvements are also done which involve. 

1. importing key to channel context of the select PubSub Group or all the PubSubGroups with matching SecurityGroupId
2. fixed the initialize key storage method to set the maximum key size from maximum future and past key size.
3. fixed memory leak issues when adding an existing keyStorage to the PubSubGroup.
4. switched to fakesleep for UA_Sleep_ms in unit tests.
5. Synced with upstream master

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,403 @@
...
+ * @param server the server object
+ * @param securityGroupId the identifier of the SecurityGroup
+ * @param securityPolicyUri the security policy assocaited with the security algorithm
+ * @param maxPastKeyCount maximum number of past keys a keystorage is allowed to store
+ * @param maxFutureKeyCount maximum number of future keys a keystorage is allowed to store
+ * @param keyStorage pointer to the keystorage to be initialized
+ * @return UA_StatusCode return status code
+ */
+UA_StatusCode
+UA_PubSubKeyStorage_initialize(UA_Server *server, const UA_String *securityGroupId,
```
jpfr, line 271: Shorten to _init to stay within the nomenclature of the SDK.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,403 @@
...
+
+    /**
+     *  the current key used to secure the messages
+     */
+    UA_PubSubKeyListItem *currentItem;
+
+    /**
+     * the last key item in the storage
+     */
+    UA_PubSubKeyListItem *lastItem;
```
jpfr, line 178: If you need a list with first and last element known, use the TAILQ structure from /deps/open61541_queue.h.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,530 @@
...
+}
+
+static void
+UA_PubSubKeyStorage_clearKeyList(UA_PubSubKeyStorage *keyStorage) {
+    if(keyStorage->keyList) {
+        UA_PubSubKeyListItem *iterator = keyStorage->firstItem;
+        UA_PubSubKeyListItem *head = iterator;
+        while(iterator != NULL) {
+            if(iterator->key.data) {
+                UA_free(iterator->key.data);
```
jpfr, line 78: UA_ByteString_clear.
That covers the difference between "empty string" and "null string" in the SDK.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,530 @@
...
+                head = iterator;
+                iterator = iterator->next;
+                UA_free(head);
+            } else
+                break;
```
jpfr, line 83: Is there a special semantics to a listitem with empty key?
Add a comment to explain why you break.
Or don't break...

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,530 @@
...
+        }
+        keyStorage->keyListSize = 0;
+    }
+}
+
+void
+UA_PubSubKeyStorage_deleteMembers(UA_Server *server, UA_PubSubKeyStorage *keyStorage) {
```
jpfr, line 90: Name this _clear.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,530 @@
...
+
+    if(!keyStorage)
+        return;
+
+    /*Remove callback*/
+    if(server) {
```
jpfr, line 96: You can assume this is only called with the server.
It is internal API where we can prevent misuse.
You can add a UA_assert(server) toe help the analysis in the CI.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,530 @@
...
+    memset(keyStorage, 0, sizeof(UA_PubSubKeyStorage));
+}
+
+void
+UA_PubSubKeyStorage_delete(UA_Server *server, UA_PubSubKeyStorage *keyStorage) {
+
+    if(!keyStorage)
+        return;
+
+    UA_PubSubKeyStorage_deleteMembers(server, keyStorage);
```
jpfr, line 117: If this is the only place that uses UA_PubSubKeyStorage_deleteMembers, then the indirection can be removed.
Just define UA_PubSubKeyStorage_delete in that case.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,530 @@
...
+    return retval;
+error:
+    if(keyStorage) {
+        UA_PubSubKeyStorage_clearKeyList(keyStorage);
+    }
+    return retval;
+}
+
+UA_StatusCode
+UA_PubSubKeyStorage_getKeyByKeyID(const UA_UInt32 keyId, UA_PubSubKeyListItem *firstItem,
```
jpfr, line 255: In the sense of a "quasi object-oriented" API, forward a UA_PubSubKeyStorage pointer to this method.
Instead of forwarding only elements from a UA_PubSubKeyStorage.
That also allows us the change the internal definitions of UA_PubSubKeyStorage without changing this API.

This applies to all methods with the UA_PubSubKeyStorage_ prefix.

jpfr (commented):
This is already veyr high quality.
Some comments inline.

Happy to see this merged soon.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,530 @@
...
+
+    for(size_t i = 0; i < config->pubSubConfig.securityPoliciesSize; i++) {
+        if(UA_String_equal(securityPolicyUri,
+                           &config->pubSubConfig.securityPolicies[i].policyUri)) {
+            *policy = &config->pubSubConfig.securityPolicies[i];
+            retval = UA_STATUSCODE_GOOD;
+            break;
+        }
+    }
+    return retval;
```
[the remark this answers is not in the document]
Muddasir-Shakil: Thanks for the comments.
In this case, it returns the bad statuscode. Now I added a unit test to verify that and changed it to BADNOTFOUND.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,403 @@
...
+                                                 UA_PubSubSecurityPolicy **policy);
+
+/**
+ * @brief Deletes all member of a key storage.
+ *
+ * @param server server from which to remove the key storage
+ * @param keyStorage keyStorage to delete
+ */
+void
+UA_PubSubKeyStorage_deleteMembers(UA_Server *server, UA_PubSubKeyStorage *keyStorage);
```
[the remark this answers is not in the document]
Muddasir-Shakil: done

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: done

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: Not needed anymore after changing to TAILQ list

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: done

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: added UA_Assert(server) at the start of function

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: deleteMembers is now changed to clear and delete is kept because clear does not free the keystorage pointer.

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: Whereever possible changed to passing PubSubKeyStorage pointer.

Muddasir-Shakil:
Hello @jpfr,  
I updated the PR according to your comments. The major change involve is switching to TAILQ for keyList in the keyStorage. Also whereever possible, I change the internal API to forward PubSubkeyStorage pointer. Meanwhile, I have readied the Patch queue for the securityGroups, sks pull and sks client. I will rebase and create PR once this PR gets into upstream. 

Looking forward to hear from you.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+
+/**
+ * @brief Allocates memory for a UA_PubSubKeyStorage and returns a pointer of the
+ * PubSubKeyStorage. The memory must be freed by calling UA_PubSubKeyStorage_delete or
+ * UA_PubSubKeyStorage_removeKeyStorage
+ *
+ * @return UA_PubSubKeyStorage*
+ */
+UA_PubSubKeyStorage *
+UA_PubSubKeyStorage_new(void);
```
jpfr, line 204: A separate _new is unnecessary complexity.
All it does is malloc and zero out the memory.
You can just use UA_calloc in the few places that create a new storage.

src/pubsub/ua_pubsub_ns0.c
```diff
@@ -1633,6 +1634,101 @@ removeDataSetWriterAction(UA_Server *server,
...
+
+    /* TODO: Check user credentials*/
+
+    if(inputSize < 7)
+        return UA_STATUSCODE_BADARGUMENTSMISSING;
+    if(inputSize > 7 || outputSize > 0)
+        return UA_STATUSCODE_BADTOOMANYARGUMENTS;
+
+    /*check for types*/
+    if(input[0].type != &UA_TYPES[UA_TYPES_STRING] ||     /*SecurityGroupId*/
```
jpfr, line 1665: Also check whether the variants contain a scalar.
If they contain an empty array (of that type), then oob memory is accessed.
We have `UA_Variant_isScalarType` that does everything in one go.

src/pubsub/ua_pubsub_ns0.c
```diff
@@ -1633,6 +1634,101 @@ removeDataSetWriterAction(UA_Server *server,
...
+            if(session_entry->session.header.channel->securityMode !=
+               UA_MESSAGESECURITYMODE_SIGNANDENCRYPT)
+                return UA_STATUSCODE_BADSECURITYMODEINSUFFICIENT;
+        }
+    }
+
+    if(!server || !input)
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+
+    /* TODO: Check user credentials*/
```
jpfr, line 1657: This comment can leave people confused whether this is inherently insecure as an implementation.
Rather write that the user credentials are checked in the AccessControl plugin *before* this callback is executed.
It is up to the user to configure his AccessControl plugin in a way that suits the use case.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+ * @brief KeyStorage must be referenced by atleast one PubSubGroup. This method checks if
+ * KeyStorage is referenced by more then 1 PubSubGroup, then it decreases the
+ * referenceCount by one. If the referenceCount is exactly one then removes the KeyStorage
+ * from the server and PubSubGroup.
+ *
+ * @param server that holds the keystorage and PubSubGroups0
+ * @param groupId of the PubSubGroup
+ */
+void
+UA_PubSubKeyStorage_removeKeyStorage(UA_Server *server, UA_String securityGroupId);
```
jpfr, line 390: In nearly all cases where this is used, you first do a lookup, then check if `ks != NULL`, then call removeKeyStorage.
Just forward the ks pointer to `removeKeyStorage` instead of doing a duplicate lookup.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+
+/**
+ * @brief The callback function to update the current key from keystorage in the server
+ * and import the key into channel context of the associated PubSub Group
+ *
+ * @param server the server object
+ * @param keyStorage the pointer to the keystorage
+ */
+void
+UA_moveToNextkeyCallback(UA_Server *server, UA_PubSubKeyStorage *keyStorage);
```
jpfr, line 357: Why no `PubSubKeyStorage` in the prefix?

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+ *
+ * @param server the server object
+ * @param keyStorage the pointer to the existing keystorage in the server
+ * @param callback the callback function to be added to the server
+ * @param timeToNextMs time in milli seconds to trigger the callback function
+ * @param callbackID the returned callbackId of the added callback function
+ * @return UA_StatusCode the return status
+ */
+UA_StatusCode
+UA_PubSubKeyStorage_addMoveToNextKeyCallback(UA_Server *server,
```
jpfr, line 326: It is imho clearer to write `Rollover` instead of `MoveToNextKey`.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+ * @brief retreives the security policy pointer from the PubSub configuration by
+ * SecurityPolicyUri
+ *
+ * @param server the server object
+ * @param securityPolicyUri the URI of the security policy
+ * @param policy the pointer to the security policy
+ * @return UA_StatusCode return status code
+ */
+UA_StatusCode
+UA_PubSubKeyStorage_getPubSubSecurityPolicyByUri(UA_Server *server,
```
jpfr, line 229: `UA_Server_findPubSubSecurityPolicy`

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+ * @brief Find the Keystorage from the Server KeyStorageList and returns the pointer to
+ * the keystorage
+ *
+ * @param server holds the keystoragelist
+ * @param securityGroupId of the keystorage to be found
+ * @param keyStorage is the pointer to returned keystorage, if not found null is returned
+ * @return Pointer to the keystorage on success, null pointer on failure
+ */
+UA_PubSubKeyStorage *
+UA_PubSubKeyStorage_getKeyStorageWithSecurityGroupId(UA_Server *server,
```
jpfr, line 216: `UA_Server_findKeyStorage`

src/pubsub/ua_pubsub_manager.c
```diff
@@ -773,6 +777,14 @@ UA_PubSubManager_delete(UA_Server *server, UA_PubSubManager *pubSubManager) {
     TAILQ_FOREACH_SAFE(tmpSDS1, &server->pubSubManager.subscribedDataSets, listEntry, tmpSDS2){
         UA_Server_removeStandaloneSubscribedDataSet(server, tmpSDS1->identifier);
     }
+
+#if defined(UA_ENABLE_PUBSUB_ENCRYPTION) && defined(UA_ENABLE_PUBSUB_SKS)
+    UA_PubSubKeyStorage *keyStorage, *keyStorageTemp;
```
jpfr, line 782: If you call the variables `ks, ksTmp`, then the foreach fits into a single line.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+ */
+UA_StatusCode
+UA_PubSubKeyStorage_addMoveToNextKeyCallback(UA_Server *server,
+                                          UA_PubSubKeyStorage *keyStorage,
+                                          UA_ServerCallback callback,
+                                          UA_Duration timeToNextMs,
+                                          UA_UInt64 *callbackID);
+
+/**
+ * @brief It takes the current Key data, divide it into singing key, encrypting key and
```
jpfr, line 333: singing key

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+ * SecurityGroupId.
+ *
+ * @param server The server object
+ * @param pubSubGroupId the nodeId of the Reader/WirterGroup whose channel context to be
+ * updated
+ * @param securityGroupId The identifier for the SecurityGroup
+ * @return UA_StatusCode return status code
+ */
+UA_StatusCode
+UA_PubSubKeyStorage_importKeyToChannelContext(UA_Server *server, const UA_NodeId pubSubGroupId,
```
jpfr, line 346: This is used in only one place and does not have to be exposed.
It makes sense that the keys are loaded when the rollover happens (and only then).
Make the function `static`, i.e. defined only in the compilation unit.
Also, the only place that uses it calls with NODEID_NULL.

src/pubsub/ua_pubsub_keystorage.h
jpfr: This can then be reused in other places as well.

CMakeLists.txt
```diff
@@ -388,6 +388,9 @@ mark_as_advanced(UA_ENABLE_PUBSUB_MONITORING)
 option(UA_ENABLE_PUBSUB_BUFMALLOC "Enable allocation with static memory buffer for time critical PubSub parts" OFF)
 mark_as_advanced(UA_ENABLE_PUBSUB_BUFMALLOC)
 
+option(UA_ENABLE_PUBSUB_SKS "Enable Security Key Service support for publisher and subscriber" OFF)
```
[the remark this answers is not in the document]
Muddasir-Shakil: should we also mark it Experimental until all the patches are merged related to SKS?

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -0,0 +1,498 @@
...
+ *
+ * Copyright (c) 2019 ifak e.V. Magdeburg (Holger Zipper)
+ * Copyright (c) 2022 Linutronix GmbH (Author: Muddasir Shakil)
+ */
+
+#include "ua_pubsub_keystorage.h"
+
+#if defined(UA_ENABLE_PUBSUB_ENCRYPTION) && defined(UA_ENABLE_PUBSUB_SKS) /* conditional compilation */
+
+#define UA_CURRENTKEY_COUNT 1
```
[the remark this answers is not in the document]
jpfr: Yeah. A variable in the local scope is preferrable.
Global defines are a good idea in general.
But in that case it is just a headache to do the scrolling, just to find that the current count is always "1".

CMakeLists.txt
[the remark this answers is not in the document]
jpfr: I'd go so far to add a warning message in cmake.
That the feature is currently merged but not yet ready.

src/pubsub/ua_pubsub_ns0.c
Muddasir-Shakil: I added a @note before this callback and removed the comment.

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: should it also be public? UA_EXPORT?

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: should it also be public? UA_EXPORT?

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: There are technically three scenarios for this method here

1. When the SetSecurityKeys Method is called, it passes the current and future key. The current key is immediately imported into channel context with this method.
2. When keylifetime expires, the Rollover moves the currentItem pointer to the next key and then imports that key to the channelcontext. The first iteration happens after the first keylifetime period expires, which can take any amount of time.
3. When keys are fetched using Pull Model (GeSecurityKeys Method) and stored in the keyStorage and the currentItem points to the current key. Then Writer/Reader groups can be into operational at random times. Then the current key is imported from KeyStorage. It is also possible that the Rollover has already done multiple iterations before setting into Operational state, in this Channel Context will be update with same key material.

These are the reasons I set it non static.

NODEID_NULL, is used to set the keys for all the related Writer/Reader Groups. However, when a nodeId is given only that Reader/Writer Group is update. It is used in the next patches related to SKS client.

src/pubsub/ua_pubsub_keystorage.h
jpfr: Since the key storage structure is not public, no.

src/pubsub/ua_pubsub_keystorage.h
jpfr: Ok. So there is additional usage that will only come with the future PR.
I think the name _activateKey makes more sense.
Import sounds like importing from a file.

src/pubsub/ua_pubsub_ns0.c
Muddasir-Shakil: changed to hasScalarType

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: added the PubSubKeyStorage prefix

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: replaced with _keyRolloverCallback()

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: Changed it to UA_Server_findPubSubSecurityPolicy

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: Changed it to UA_Server_findKeyStorage

src/pubsub/ua_pubsub_manager.c
Muddasir-Shakil: shorten the pointer names

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: changed it to _activateKeyToChannelContext()

src/pubsub/ua_pubsub_keystorage.h
Muddasir-Shakil: removed and replaced with UA_calloc everywhere

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -0,0 +1,396 @@
...
+                                                 UA_PubSubSecurityPolicy **policy);
+
+/**
+ * @brief Deletes all member of a key storage.
+ *
+ * @param server server from which to remove the key storage
+ * @param keyStorage keyStorage to delete
+ */
+void
+UA_PubSubKeyStorage_clear(UA_Server *server, UA_PubSubKeyStorage *keyStorage);
```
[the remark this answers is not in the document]
Muddasir-Shakil: removed the indirection.

CMakeLists.txt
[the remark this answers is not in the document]
Muddasir-Shakil: Added the Check for UA_PUBSUB_ENCRYPTION and warning text.

src/pubsub/ua_pubsub_keystorage.c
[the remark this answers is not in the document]
Muddasir-Shakil: Removed from global scope to the local method scope

jpfr:
Nice! Merged.
Looking forward to the next PR.

Issue #5247, opened 2022-07-11
Use of compont volatile operations (util.h, architecture_definitions.h) throws warnings in C++20


marcmutz:
Said headers use what's called compound operations on volatile lvalue:

    volatile T  *x = ~~~;
    *x += 10;
    *x -= 10;
    *x |= 10;

which have been deprecated in C++20, because they look like a single access, but in fact are spec'ed to be a load followed by a store.

GCC and Clang in C++20 mode warns about these operations being deprecated, so projects which use open62541 with -Werror will fail to compile.

The fix is only use volatile loads and store, not compounds:

    volatile T *x = ~~~;
    *x = *x + 10;
    *x = *x - 10;
    *x = *x | 10;

PR #5248, opened 2022-07-11
fix(core): use of compound volatile operations (deprecated in C++20)

marcmutz:
C++20 deprecated compound volatile operations, so this code fails to
compile in C++20 mode due to -Werror.

Fix by separating loading and saving of the value, which, in the case
of architecture_definitions.h, even reduces the per-operation loads
from two to one.

Fixes: #5247


jpfr:
Looks good. Please also sign the cla.

jpfr:
See below in the ci test list.

GoetzGoerisch:
@marcmutz could you please sign the CLA?

marcmutz:
> @marcmutz could you please sign the CLA?

Done.

jpfr:
Nice. Merged.

PR #5256, opened 2022-07-12
test(core): fix sign-conversion compiler errors

bluhm:
These fixes to compile with clang 13 are already in master branch.
Having them also in 1.3 branch also, make testing easier for me.
If you fear too much fallout when merging, just drop this pull request.


jpfr:
Looks good. Merged.

PR #5258, opened 2022-07-13
feat(pubsub) add pre-operational state for DataSetReaders

andreasebner:
Adding support for the Pre-Operational PubSub-State. In addition, this PR introduces the functionality to access the pubsub state via the information model.


andreasebner:
Is now contained in #5456

PR #5259, opened 2022-07-14
[Review] feat (server): add call for retrieving session information

keba-ebr:
Session information can now be retrieved from the server object.
The information contains client application description and additional
session related values defined by UA_SessionInformation.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -519,6 +519,23 @@ UA_Server_getSessionParameter_scalar(UA_Server *server,
                                      const UA_DataType *type,
                                      void *outValue);
 
+typedef struct {
+    UA_ApplicationDescription clientDescription;
+    UA_String    sessionName;
+    UA_Boolean   activated;
+    void        *sessionHandle; /* pointer assigned in userland-callback */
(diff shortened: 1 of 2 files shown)


jpfr:
Hey there,

since we last spoke, a subsystem for localization (multiple languages) was added by Basyskom.
As part of that we needed to provide access to session attributes.

Every session has a key-value list of attributes that can be attached locally.
We now use the same API to also get "system-defined" attributes from a session.

https://github.com/open62541/open62541/blob/70b9170d54b1168dfbbcb1bcfce2d475f8a95f5e/include/open62541/server.h#L492
https://github.com/open62541/open62541/blob/70b9170d54b1168dfbbcb1bcfce2d475f8a95f5e/src/server/ua_session.c#L303

Sorry for the back-and-for. But we should keep a unified approach.

If you just add your additional attributes individually to `getSessionAttribute` and document the attributres in the server.h this is done without an additional API.

keba-ebr:
Hi,

thanks for the info.
Since the client description is available and that's my needed information I will close this PR.
If I need the locale list that we discussed previously I will create a new.

Thanks!

PR #5260, opened 2022-07-14
[REVIEW] Put rejected certificates to rejected folder

robertns:
Add optional automatic saving of unknown client certificates to the specified folder (e.g. PKI/CA/rejected). 

After that, user (manually or with some custom tool) can store the certificate to the folder with trusted certificates.

This involves additional parameter to `UA_CertificateVerification_CertFolders()` - a path to folder where the rejected certificate would be stored.


plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -471,6 +477,29 @@ certificateVerification_verify(void *verificationContext,
         } else {
             retval = UA_STATUSCODE_BADSECURITYCHECKSFAILED;
         }
+
+#ifdef UA_ENABLE_CERT_REJECTED_DIR
+        if(ci->rejectedListFolder.length !=0) {
+            char rejectedFileName[256] = {0};
+            UA_ByteString thumbprint;
+            UA_ByteString_allocBuffer(&thumbprint, UA_SHA1_LENGTH);
```
keba-estr, line 485: Memory must be cleared after usage (i.e. copying) with UA_ByteString_clear.

robertns: Thanks, I've added the code.

andreasebner:
Please enable your work in at least in one unit test.

andreasebner:
It's nice to see the progress with the tests. A small hint: It is sufficient to enable the new tests or the flag in one of the existing Github CI-runners (so that a runner is not needed for each feature).

robertns:
> It's nice to see the progress with the tests. A small hint: It is sufficient to enable the new tests or the flag in one of the existing Github CI-runners (so that a runner is not needed for each feature).

plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -471,6 +477,35 @@ certificateVerification_verify(void *verificationContext,
...
+            if(mbedtls_thumbprint_sha1(certificate, &thumbprint) == UA_STATUSCODE_GOOD) {
+                static const char hex2char[] = "0123456789ABCDEF";
+                for (size_t pos = 0, namePos = 0; pos < thumbprint.length; pos++) {
+                    rejectedFileName[namePos++] = hex2char[(thumbprint.data[pos] & 0xf0) >> 4];
+                    rejectedFileName[namePos++] = hex2char[thumbprint.data[pos] & 0x0f];
+                }
+                strcat(rejectedFileName, ".der");
+            } else {
+                UA_UInt32 seconds = (UA_UInt32) time(NULL);
+                sprintf(rejectedFileName, "cert%d.der", seconds);
```
andreasebner, line 495: I think you could use here the UA_DateTime_now() and therefore avoid the timer include.

andreasebner:
@jpfr The feature is only availbe with mbedtls and conditional (by an cmake flag). Currently there is a change in a API-function (new paramter). An alternative would be, to add this paramter based on the cmake flag. What do you think?

plugins/crypto/mbedtls/ua_pki_mbedtls.c
robertns: I changed it, thanks!

plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -471,6 +475,35 @@ certificateVerification_verify(void *verificationContext,
...
+            UA_ByteString_allocBuffer(&thumbprint, UA_SHA1_LENGTH);
+            if(mbedtls_thumbprint_sha1(certificate, &thumbprint) == UA_STATUSCODE_GOOD) {
+                static const char hex2char[] = "0123456789ABCDEF";
+                for (size_t pos = 0, namePos = 0; pos < thumbprint.length; pos++) {
+                    rejectedFileName[namePos++] = hex2char[(thumbprint.data[pos] & 0xf0) >> 4];
+                    rejectedFileName[namePos++] = hex2char[thumbprint.data[pos] & 0x0f];
+                }
+                strcat(rejectedFileName, ".der");
+            } else {
+                UA_UInt32 dt = (UA_UInt32)UA_DateTime_now();
```
keba-estr, line 492: UA_DateTime_now returns the 64 bit signed integer which represents the number of 100 nanosecond intervals since January 1, 1601. If you cast this value to a 32 bit unsigned integer, you get the same value every 7,15 minutes. Should the UA_DateTime_now be converted to seconds with UA_DATETIME_SEC or should the 64-bit integer be used for the file name?

plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -471,6 +475,35 @@ certificateVerification_verify(void *verificationContext,
...
+                sprintf(rejectedFileName, "cert%d.der", dt);
```
keba-estr, line 493: If an unsigned integer is printed to the file name, %u should be used instead of %d.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
robertns: I changed it to use 64-bit unsigned integer. I'm not sure if we should save the rejected certificate in this case at all.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -471,6 +475,35 @@ certificateVerification_verify(void *verificationContext,
...
+            if(mbedtls_thumbprint_sha1(certificate, &thumbprint) == UA_STATUSCODE_GOOD) {
+                static const char hex2char[] = "0123456789ABCDEF";
+                for (size_t pos = 0, namePos = 0; pos < thumbprint.length; pos++) {
+                    rejectedFileName[namePos++] = hex2char[(thumbprint.data[pos] & 0xf0) >> 4];
+                    rejectedFileName[namePos++] = hex2char[thumbprint.data[pos] & 0x0f];
+                }
+                strcat(rejectedFileName, ".der");
+            } else {
+                UA_UInt64 dt = (UA_UInt64) UA_DateTime_now();
+                sprintf(rejectedFileName, "cert_%lu.der", dt);
```
keba-estr, line 493: I don't think %lu is portable on all platforms.
Use
#include <inttypes.h>
 and 
sprintf(rejectedFileName, "cert_%" PRIu64 ".der", dt);

robertns: Thanks!

robertns:
> @jpfr The feature is only availbe with mbedtls and conditional (by an cmake flag). Currently there is a change in a API-function (new paramter). An alternative would be, to add this paramter based on the cmake flag. What do you think?

@andreasebner  @jpfr Thanks for reviewing! Should I also implement conditional compiling for the additional parameter?

PR #5268, opened 2022-07-20
fix(arch): obsolete ipv6 define

bluhm:
According to Linux /usr/include/bits/in.h IPV6_ADD_MEMBERSHIP is
obsolete.  It does not exist on OpenBSD.  Use IPV6_JOIN_GROUP
instead.


jpfr:
Nice, good find.

PR #5269, opened 2022-07-20
test(core): fix more sign-conversion compiler errors

bluhm:


jpfr:
Thanks for this.

We had a lot of PR like this. We shouldn't even need them.
I added a commit to the PR so we don't remove sign-conversion errors in the unit tests.

That will enforce that all future changes don't have these errors. Neither on OpenBSD nor elsewhere.

PR #5270, opened 2022-07-20
fix(client): use after free in asyncServiceTimeoutCheck

bluhm:
Commit de9d691906547c9fc99e0a77198a80fe5bba54b1 fixed a use after
free in UA_Client_AsyncService_cancel().  The same bug is present
in asyncServiceTimeoutCheck(), so use the same fix.  After timeout,
remove the async service calls from the global list and store them
in a local one.  Call UA_Client_AsyncService_cancel() and with it
the UA_free() while traversing this local list.  Otherwise the
callback could free a global list element that asyncServiceTimeoutCheck()
is referencing.


jpfr:
Nice, thanks!

PR #5278, opened 2022-07-28
fix(server): fix to compile server without pubsub sources when pubsub…

SchutzSeb:
The fix allow to compile the server without pubsub folders when pubsub feature is disabled.
Before this fix, an include of ua_pubsub_manager.h was done into ua_server.c without conditional compilation on UA_ENABLE_PUBSUB. This led to compilation crash if path to this file was not provided.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -25,7 +25,9 @@
 #include "ua_pubsub_ns0.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


jpfr:
Cool! Please click the CLA in the list so we can merge.

SchutzSeb:
I have signed it but it is still stated as pending. Is there a delay before it is taken into account ?

image

PR #5289, opened 2022-08-04
fix(pubsub infomodel): remove deadlock datasetreader cleanup

Fischer-Thomas:
There is a deadlock when trying to cleanup a DataSetReader that is connected to a StandaloneSubscribedDataSet.
We don't need to delete the references ourself as this is already handled by the deleteNode()-methodcall.


jpfr:
👍

PR #5291, opened 2022-08-04
Fix(pubsub/infomodel) memleak when reading PDS/SDS children

Fischer-Thomas:
There is a memleak when reading the value of PDS->PublishedData/DataSetMetaData/ConfigurationVersion.

Adjusted the unit test for the PDS to also read the value of the children to verify the memleak is fixed.


andreasebner:
Tank your for the great contribution!

PR #5296, opened 2022-08-09
[REVIEW] feat(pubsub): Implementation of the reserveIds method

NoelGraf:
This PR implements the ReserveIds method. This Method reserves unique WriterGroupIds and DataSetWriterIds to allow PubSub configuration applications to apply unique Ids to new PubSub configuration elements when preparing a update to the PubSub configuration.

- [x] Internal representation
- [x] Internal API function (external available)
- [x] Information model reserveId
- [x] Session handling
- [x] Test case
- [ ] Return the correct DefaultPublisherId.

More Information [here](https://reference.opcfoundation.org/v105/Core/docs/Part14/9.1.3/#9.1.3.7.5)


MurzynJ:
Hi @NoelGraf
I already mentioned that in conversation with @andreasebner last week, but I will leave a comment also here.

Handling of the DefaultPublisherId argument is broken in this implementation.

DefaultPublisherId is a value that is bound to the Server instance. It can be optionally exposed in the Address Space 
[here](https://reference.opcfoundation.org/v105/Core/Objects/PublishSubscribe/)

Each PubSub transport layer may have its DefaultPublisherId defined, however, at the moment only the DefaultDatagramPublisherId is specified. Typically it would be constant during the Server lifetime so it can be created on server  startup (PubSub module startup) using the recommended formula (MAC + port no.) specified [here](https://reference.opcfoundation.org/Core/Part14/6.2.7/#6.2.7.1). This formula ensures that the PublisherId is globally unique (which is more than enough to meet the actual requirement that all publishers connected to the same "Message Oriented Middleware" have unique PublishersId).

The ReservedIds Method should just output that DefaultPublisherId based on TransportProfileUri, In case of transport profiles other then UDP, I think the Method can return error for now.

MurzynJ:
> Hi @NoelGraf I already mentioned that in conversation with @andreasebner last week, but I will leave a comment also here.
> 
> Handling of the DefaultPublisherId argument is broken in this implementation.
> 
> DefaultPublisherId is a value that is bound to the Server instance. It can be optionally exposed in the Address Space [here](https://reference.opcfoundation.org/v105/Core/Objects/PublishSubscribe/)
> 
> Each PubSub transport layer may have its DefaultPublisherId defined, however, at the moment only the DefaultDatagramPublisherId is specified. Typically it would be constant during the Server lifetime so it can be created on server startup (PubSub module startup) using the recommended formula (MAC + port no.) specified [here](https://reference.opcfoundation.org/Core/Part14/6.2.7/#6.2.7.1). This formula ensures that the PublisherId is globally unique (which is more than enough to meet the actual requirement that all publishers connected to the same "Message Oriented Middleware" have unique PublishersId).
> 
> The ReservedIds Method should just output that DefaultPublisherId based on TransportProfileUri, In case of transport profiles other then UDP, I think the Method can return error for now.

After talking with the editor of Part 14 I have better understanding.
We have 2 types of transport profiles: Datagram and Broker.
Datagram: UDP, Ethernet
Broker: MQTT, AMQP

For Datagram type transport profile, the returned PublisherId is the DefaultDatagramPublisherId.
For Broker type, the returned PublisherId is an ApplicationUri

image

src/pubsub/ua_pubsub_manager.c
```diff
@@ -110,9 +110,158 @@ UA_PubSubManager_addPubSubTopicAssign(UA_Server *server, UA_ReaderGroup *readerG
...
+            if(UA_String_equal(&reserveId1->transportProfileUri, &transportProfileUri) && reserveId1->reserveIdType == reserveIdType && reserveId1->id == next_id) {
+                is_free = false;
+                next_id++;
+                break;
+            }
+        }
+        if(!is_free)
+            continue;
+
+        UA_PubSubConnection *tmpConnectoin;
```
MichaelTatarski, line 152: minor typo

src/pubsub/ua_pubsub_manager.c
```diff
@@ -110,9 +110,157 @@ UA_PubSubManager_addPubSubTopicAssign(UA_Server *server, UA_ReaderGroup *readerG
...
 }
 
+static UA_ReserveId *
+UA_ReserveId_new(UA_Server *server, UA_UInt16 id, UA_String transportProfileUri,
+                 UA_ReserveIdType reserveIdType, UA_NodeId sessionId) {
+    UA_ReserveId *reserveId = (UA_ReserveId *) calloc(1, sizeof(UA_ReserveId));
+    if(!reserveId) {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                     "PubSub ReserveId creation failed. Out of Memory.");
+        return 0;
```
andreasebner, line 120: Please return NULL for pointers.

PR #5299, opened 2022-08-12
fix usage() mistake

alongL:
In the usage() , the PCP should be 3 to meet the semantic.
program output show as follw:
```
 -pubUri     [name]   Publisher Mac address (default opc.eth://01-00-5E-00-00-01:8.7 - where 8 is the VLAN ID and 3 is the PCP)
 -subUri     [name]   Subscriber Mac address (default opc.eth://01-00-5E-7F-00-01:8.7 - where 8 is the VLAN ID and 3 is the PCP)
```


andreasebner:
I think this change would break the symmetrie of the pubsub real time examples (the would not work together anymore). Therefore i would preffer to adjust the help texts in all of the pubsub rt examples.

Issue #3248, opened 2019-11-05
opc ua client example code compile error


kusumy:
## Description
Compile one of example file (client_connect.c) with following command:

```bash
gcc -std=c99 -lopen62541 -DUA_ARCHITECTURE_POSIX client_connect.c -o client_connect
```

shows error message:

/tmp/ccBlCKcm.o: In function `UA_ByteString_clear':
client_connect.c:(.text+0x241): undefined reference to `UA_TYPES'
client_connect.c:(.text+0x249): undefined reference to `UA_clear'
/tmp/ccBlCKcm.o: In function `UA_Variant_clear':
client_connect.c:(.text+0x288): undefined reference to `UA_TYPES'
client_connect.c:(.text+0x290): undefined reference to `UA_clear'
/tmp/ccBlCKcm.o: In function `UA_Client_readValueAttribute':
client_connect.c:(.text+0x2b2): undefined reference to `UA_TYPES'
client_connect.c:(.text+0x2c6): undefined reference to `__UA_Client_readAttribute'
/tmp/ccBlCKcm.o: In function `loadFile':
client_connect.c:(.text+0x2ea): undefined reference to `UA_STRING_NULL'
client_connect.c:(.text+0x2f1): undefined reference to `UA_STRING_NULL'
/tmp/ccBlCKcm.o: In function `main':
client_connect.c:(.text+0x44e): undefined reference to `UA_STRING_NULL'
client_connect.c:(.text+0x455): undefined reference to `UA_STRING_NULL'
client_connect.c:(.text+0x809): undefined reference to `UA_Client_new'
client_connect.c:(.text+0x81f): undefined reference to `UA_Client_getConfig'
client_connect.c:(.text+0x890): undefined reference to `UA_ClientConfig_setDefaultEncryption'
client_connect.c:(.text+0x8bd): undefined reference to `UA_ClientConfig_setDefault'
client_connect.c:(.text+0x925): undefined reference to `UA_Client_connect_username'
client_connect.c:(.text+0x946): undefined reference to `UA_Client_connect'
client_connect.c:(.text+0x964): undefined reference to `UA_Client_delete'
client_connect.c:(.text+0x975): undefined reference to `UA_Log_Stdout'
client_connect.c:(.text+0x9e2): undefined reference to `UA_TYPES'
client_connect.c:(.text+0xa04): undefined reference to `UA_DateTime_toStruct'
client_connect.c:(.text+0xa3f): undefined reference to `UA_Log_Stdout'
[...]

PR #5303, opened 2022-08-16
fix command in the examples-section of the README

linderd:
fix the gcc linking argument in the examples-section of the README, see https://github.com/open62541/open62541/issues/3248#issuecomment-568370301


jpfr:
Thanks

PR #5307, opened 2022-08-22
when use UDP address, don't need an interface.

alongL:
when use UDP address as pubUri and subUri, we don't need to set the interface .
just set the route of OS.
If we set the interface, the program will crash. see below:

```
    //If you are running for UDP provide ip address as input. Connection creation
    //Failed if we provide the interface name
```

### DIFF
--- a/examples/pubsub_realtime/pubsub_TSN_publisher_multiple_thread.c
+++ b/examples/pubsub_realtime/pubsub_TSN_publisher_multiple_thread.c
@@ -1352,11 +1352,13 @@ int main(int argc, char **argv) {
         }
     }
 
+#ifdef UA_ENABLE_PUBSUB_ETH_UADP
     if (!interface) {
         UA_LOG_ERROR(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "Need a network interface to run");
         usage(progname);
         return -1;
     }
+#endif
 
     if (cycleTimeInMsec < 0.125) {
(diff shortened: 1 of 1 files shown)


JanSurft:
Could you add further information about the problems you are encountering? An error log/output of the running server and highlighting the problematic parts would be helpful.

alongL:
> Could you add further information about the problems you are encountering? An error log/output of the running server and highlighting or describing the problematic parts in that output would be helpful.

when I use udp broadcast address to pub and sub. It doesn't need to set the interface eth0. 
but if not set eth0, the program will exit.
If I set the eth0 as interface , It will crash.  So, the best method is only use  this in UADP mod.

```
./pubsub_TSN_multithread  \
-puburi  udp://224.0.0.22/  \
-suburi  udp://224.0.0.23/    
```

alongL:
> Could you add further information about the problems you are encountering? An error log/output of the running server and highlighting or describing the problematic parts in that output would be helpful.

Now I know the right method to use this program is like this:
192.168.8.101 is the IP of VLAN network card.
```
./bin/examples/pubsub_TSN_publisher_multiple_thread  -enableBlockingSocket \
-interface 192.168.8.101 \
-enableCsvLog  \
-disableSoTxtime \
-pubUri opc.udp://224.0.0.22:4840/  \
-subUri opc.udp://224.0.0.32:4840/  \
-pubAppCore 3 -subAppCore 4 \
-socketPriority 3 \
-cycleTimeInMsec 1000 \
-enableconsolePrint 
```

alongL:
mistake use. Not need this patch.

PR #5312, opened 2022-08-29
Find Python 3 in open62541Config.cmake.in

akoww:
On some systems open62541Config.cmake finds python2 instead of python3. 
This leads to fails on re.fullmatch() in: https://github.com/open62541/open62541/blob/4d08612120eadea386af1e45a78a6d458b7f2331/tools/nodeset_compiler/nodeset.py#L431

Solution:
Force cmake to find python3


jpfr:
Yeah. It is high time to switch to python3

PR #5315, opened 2022-08-31
Feature/eventloop udp unicast multicast

JanSurft:
### Mandatory
- [x] rebase on master (done)
- [x] fix TransportLayer2Type possible segfault (done)
- [x] invalidate pubsub connection when unicast and not "localhost:..." in address of pubsub connection (done)
- [x] integrate all m-fixes (done)
- [ ] - [ ] fix failing nodeset tests because of new datatype
  - adding `DatagramWriterGroupTransport2DataType` and `DatagramConnectionTransport2DataType` to `tools/schema/datatypes_pubsub.txt` leads to errors.
  - E.g.: `check_nodeset_loader_di: open62541/src/Value.c:266: setStructure: Assertion value->type == DATATYPE_COMPLEX' failed.`
  - Does nodesetLoader or another dependency need to be updated to 1.05?

### Can be done in a subsequent PR
- [ ] adapt remaining testcases to new UDP layer
  - multiple testcases have problems and errors in themselves that are currently hidden
  - this PR can reveal those errors, when using the new UDP layer, thus multiple testcases would need to be fixed
  - check strategy for that (e.g. it's possible to leave those testcases use the legacy UDP layer for now).
  - examples:
    - `check_pubsub_publish.c`
      - failing with SEGFAULT because attr value cannot be cleared after now immediately appearing publishCallback
      - return values for addDataSetField is actually an ignored fail code that is never checked against (this appears to be the case in other test-cases as well)
    - `check_pubsub_publish_rt_levels.c` and others
      - uses explicit receive call for receiving messages and process them with a custom callback
      - check how this can be mapped to the current eventloop structure


arch/eventloop_posix_udp.c
```diff
@@ -1115,25 +1332,24 @@ UDP_openSendConnection(UA_ConnectionManager *cm,
...
     }
 
+    /* Create a socket and register the destination address from the provided parameters */
+    UA_FD newSock = UA_INVALID_FD;
+    UA_StatusCode res =
+        registerSocketAndDestinationForSend(paramsSize, params, hostname, info,
+                                            error, newudpfd, &newSock, el->eventLoop.logger);
+    UA_freeaddrinfo(info);
+    if(res != UA_STATUSCODE_GOOD)
+        return res;
```
keba-estr, line 1351: free newudpfd  in case of error

include/open62541/plugin/pubsub.h
```diff
@@ -97,11 +107,27 @@ struct UA_PubSubChannel {
...
  * information about the TransportLayer handling. */
 
-typedef struct {
+typedef struct UA_PubSubTransportLayer {
     UA_String transportProfileUri;
-    UA_PubSubChannel *(*createPubSubChannel)(UA_PubSubConnectionConfig *connectionConfig);
+    void *connectionManager;
+    // UA_Server *server;
+    UA_PubSubChannel *(*createPubSubChannel)(struct UA_PubSubTransportLayer *tl, void *ctx);
+    UA_StatusCode (*createWriterGroupPubSubChannel)(UA_PubSubChannel** outChannel, struct UA_PubSubTransportLayer *tl, const UA_ExtensionObject *writerGroupTransportSettings, void *ctx);
```
JanSurft, line 115: place `outChannel` to the end of the arguments and `tl` at the beginning

plugins/ua_pubsub_ethernet.c
```diff
@@ -1028,7 +1029,7 @@ sendWithTxTime(UA_PubSubChannel *channel, UA_ExtensionObject *transportSettings,
 static UA_StatusCode
 UA_PubSubChannelEthernet_send(UA_PubSubChannel *channel,
                               UA_ExtensionObject *transportSettings,
-                              const UA_ByteString *buf) {
+                              UA_ByteString *buf) {
```
JanSurft, line 1032: constness needs to be removed from all the `UA_ByteString` that will pass through the eventloop for sending, because the eventloop is expecting it.

PR #5317, opened 2022-08-31
[REVIEW] fix(server): Publish more notifications with delayed callback

jonasgreen88:
If a DataChangeSubsription publish event needs to publish many
notifications using multiple PublishRequests the event can take a
significant amount of time. This can cause delays for other timer
events that are scheduled in near future of the publish event.

An example, a monitored item samples a variable every 10 ms and the
subscription publishes the notifications every 1 second. The
QueueSize is set to 100 and 10 notifications per publish response.
The publish event takes >10 ms and delays the next sampling event
of the monitored item.

By removing the recursive call to UA_Subscription_publish() the
stack usage is reduced as well.

### DIFF
--- a/arch/common/ua_timer.c
+++ b/arch/common/ua_timer.c
@@ -66,17 +66,6 @@ UA_Timer_init(UA_Timer *t) {
     UA_LOCK_INIT(&t->timerMutex);
 }
 
-void
-UA_Timer_addTimerEntry(UA_Timer *t, UA_TimerEntry *te, UA_UInt64 *callbackId) {
-    UA_LOCK(&t->timerMutex);
-    te->id = ++t->idCounter;
-    if(callbackId)
-        *callbackId = te->id;
-    aa_insert(&t->root, te);
-    aa_insert(&t->idRoot, te);
-    UA_UNLOCK(&t->timerMutex);
-}
-
 static UA_StatusCode
 addCallback(UA_Timer *t, UA_ApplicationCallback callback, void *application,
             void *data, UA_DateTime nextTime, UA_UInt64 interval,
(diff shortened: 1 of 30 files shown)


jpfr:
That is a good initiative.
Instead of a timed callback we should resort to a "Delayed Callback" (cf. plugin/eventloop.h).
A delayed callback is immediately executed after the network events have been processed.

We don't expose this feature from the eventloop via the server API right now.
We should add it.

jonasgreen88:
> That is a good initiative. Instead of a timed callback we should resort to a "Delayed Callback" (cf. plugin/eventloop.h). A delayed callback is immediately executed after the network events have been processed.
> 
> We don't expose this feature from the eventloop via the server API right now. We should add it.

Thanks for the feedback. I'm not familiar with the eventloop yet as we are still using v1.2.x. Should I expose the API in this PR or do you do that in another commit?

jonasgreen88:
Implemented UA_Server_addDelayedCallback() which exposes the delayed callback feature of the EventLoop and uses that API in UA_Subscription_publish() to schedule another publication of more notifications.

jpfr:
Nice!

1. I added removeDelayedCallback on master.
This might be needed during shutdown or when the subscription is removed.

2. Don’t expose delayed callbacks publicly. If people really need it they can access the event loop directly.

3. if you embed a UA_DelayedCallback inside UA_Subscription then you don’t need to malloc. The master was changed so that the delayed callback is no longer auto-freed.

Keep up the good work!

jpfr:
This is coming together quite nicely.

We now have two "entry-points" for the Publish.
Timed callbacks and delayed callbacks.
It could happen that a timed callback is triggered when a delayed callback has already been registered. That could lead to a double-registration of the delayed callback and memory corruption.

Please add a layer of protection with a Boolean flag in the subscription whether the Delayed Callback is currently already registered. Check that before adding a new Delayed Callback and also check it before removing the Delayed Callback.

We can then merge.

jpfr:
Merged

jonasgreen88:
Thanks for the reviews and good cooperation!

PR #5320, opened 2022-09-05
fix(pubsub): fix double free in openssl secured mqtt connection

zimmermannt-tq:
(for UA_ENABLE_MQTT_TLS_OPENSSL=ON) client->socketfd is the same pointer as
channeldata->sockfd. client->sockfd is freed in disconnectMqtt() and again
afterwards in freeTLS(). Therefore set it to NULL in disconnectMqtt and free
it in freeTLS only.

Signed-off-by: Tobias Zimmermann <Tobias.Zimmermann@tq-group.com>


jpfr:
Good find, thanks.

PR #5321, opened 2022-09-05
fix(examples): parsing argument frequency in tutorial_pubsub_mqtt_pub…

zimmermannt-tq:
…lish

argpos needs to be incremented before reading the value of the interval.

Signed-off-by: Tobias Zimmermann <Tobias.Zimmermann@tq-group.com>


jpfr:
Nice, thanks!

PR #5331, opened 2022-09-09
[Review] Add WebAssembly architecture support

cleitner:
WebAssembly allows a browser to use the open62541 client to connect to an OPC-UA server with WebSocket (and potentially HTTP) support. I've successfully done this with the 1.3 release, but had bad timing with the advent of event loop support 😄.

To be able to use the open62541 client, additional code is necessary to implement an event loop for the callback based WebSocket API of a browser and marshalling data between JavaScript and wasm. The object model of open62541 can be used to implement this quite elegantly 👍.

This initial patch set allows compilation of open62451 as a wasm32 static library. To compile this, you'll need the WASI sysroot from [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) and set the environment variable `WASI_SYSROOT`. The default `clang` package from Debian supports `wasm32` compilation but lacks some necessary support libs which can be obtained from the wasi-sdk releases aswell.

    export WASI_SYSROOT=...
    mkdir build
    cd build
    cmake -DCMAKE_TOOLCHAIN_FILE=../tools/cmake/Toolchain-wasm32-wasi.cmake -DCMAKE_BUILD_TYPE=Release -DUA_ARCHITECTURE=wasm -DUA_ENABLE_HARDENING=OFF ..

I stumbled over a couple of minor issues that might need a better solution:
 - The assumption that every architecture supports TCP connections wouldn't hold true anymore.
 - The assumption that every architecture is either POSIX or win32 compatible wouldn't hold true anymore. I guarded the POSIX event loop sources to only be included in `posix` and `win32` archs, but this might break some of the other architectures. Not sure how the project would prefer to handle this.

Things that need more work
 - Implementing the client event loop for a callback based WebSocket API (I implemented the old `init/pollConnectionFunc` API and haven't done the event loop yet to see if it is complete enough. First impression is pretty good. The old API needed one minor change to allow the callback style network implementation to iterate the client for async operation)
 - It's possible to use `mbedtls` to support crypto, but none of the integrated entropy sources can be used. A custom architecture specific entropy source is needed, e.g. based on `window.crypt.getRandomValues`. The current implementation uses the default context (and runs self-tests that are not context specific and will never succeed in this scenario).
 - With the old network stack the WebSocket server performs very poorly if the TCP network stack is also enabled. Hopefully this can be avoided by adding a open62541 event loop wrapper for libwebsockets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,15 +119,17 @@ GET_PROPERTY(ua_architecture_headers_beginning GLOBAL PROPERTY UA_ARCHITECTURE_H
 
 GET_PROPERTY(ua_architecture_sources GLOBAL PROPERTY UA_ARCHITECTURE_SOURCES)
 
-set(ua_architecture_sources ${ua_architecture_sources}
-    ${PROJECT_SOURCE_DIR}/arch/eventloop_posix.h
-    ${PROJECT_SOURCE_DIR}/arch/eventloop_posix.c
-    ${PROJECT_SOURCE_DIR}/arch/eventloop_posix_select.c
-    ${PROJECT_SOURCE_DIR}/arch/eventloop_posix_epoll.c
-    ${PROJECT_SOURCE_DIR}/arch/eventloop_posix_tcp.c
-    ${PROJECT_SOURCE_DIR}/arch/eventloop_posix_udp.c
-    ${PROJECT_SOURCE_DIR}/arch/eventloop_posix_interrupt.c
-)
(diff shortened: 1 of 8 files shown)


jpfr:
This is great!
Yes, we want have websocket support for the event loop.
If the wasm websocket provides the same API there will be zero difference needed within the core library. Differences are all abstracted out into the arch code.

With the event loop model websockets and tcp can coexist nicely. That is one of the original motivations.

andreasebner:
Thanks for the contribution. Unfortunately, it never made it into mainline. This PR is based on a very old version and would need a full overhaul. If there is intrest, feel free to repoen a PR.

PR #5340, opened 2022-09-19
[Review] feat(pubsub): Adding payload encryption in mqtt pubsub examples

NoelGraf:
The MQTT publish-and-subscribe example is extended to be able to encrypt the payload. The example demonstrates the possibility of encrypting the payload without a TLS connection.


jpfr:
Looks good!
Is this enabled in one of the CI runs?

NoelGraf:
Yes, there is a test in the CI where payload encryption is enabled. But TLS is also enabled in this test.

One thing that is still missing is the encryption of the payload once JSON encoding is enabled. Currently, JSON messages cannot be encrypted without TLS.

CMakeLists.txt
```diff
@@ -1159,6 +1159,9 @@ if(UA_ENABLE_PUBSUB)
 endif()
 
 if(UA_ENABLE_JSON_ENCODING)
+    if(UA_ENABLE_PUBSUB_ENCRYPTION)
+        MESSAGE(WARNING "Payload encryption with JSON encoding is not possible.")
```
jpfr, line 1163: Why is this not possible?
Not supported in the standard or not by us?
If we treat the message as a mere byte-string, then the encryption code for binary and json would be identical.

jpfr (commented):
The change is good.
But I don’t like the ifdeffery for json.
Maybe it’s easy to add encryption for json and reduce the configuration complexity.

CMakeLists.txt
NoelGraf: As far as I understand, this is not supported by the standard. In the specification (part 14) i found the following:

5.4.5.3.2:
*Message confidentiality and integrity with the Broker based communication model can be
ensured at two levels:
• transport security between Publishers or Subscribers and the Broker, or
• message security as end-to-end security between Publisher and Subscriber.
The Broker level security requires all Publishers and Subscribers to have credentials that grant
them access to the necessary queue or topic. In addition, all communication with the Broker
uses transport level security to ensure confidentiality. The security parameters are specified on
the connection and group.
The message security provided by the Publisher is only defined for the UADP message
mapping.*

7.3.5.1:
*Message security is primarily provided by a TLS connection between the Publisher or Subscriber and the MQTT server; however, this requires that the MQTT server be trusted. For that reason, it may be necessary to provide end-to-end message security. Applications that require end-to-end message security with MQTT need to use the UADP NetworkMessages and binary message encoding defined in 7.2.2. JSON encoded message bodies need to rely on the security mechanisms provided by MQTT and the MQTT broker.*

andreasebner: I was also suprised about the restriction but 7.3.5.1 sounds very clear to me.

jpfr: The user might think that he misconfigured something when reading the current warning.
We should clarify that this is not possible due to the standard.
Also there should be a similar warning when the users tries to configure encryption for JSON at runtime.

PR #5344, opened 2022-09-20
updates connection timeout counter

Kallelele:
There seem to be an error with the connection timeout counter.
Should be increased incase of timeout.


jpfr:
Good find!
Please sign the CLA so the PR becomes all green.

PR #5346, opened 2022-09-20
[Review] fix(pubsub): Fix generation of the defaultDatagramPublisherId

NoelGraf:
Correct generation of the defaultDatagramPublisherId when starting the server. According to the specification (6.2.7.1), it is recommended to assign the MAC address of one of the network interfaces to the first 6 bytes and the OPC UA Server Port of the OPC UA application to the remaining two bytes.

- [x] Include DefaultDatagramPublisherId in NS0
- [x] Get the MAC address
- [x] Generate DefaultDatagramPublisherId from MAC address and port.

### DIFF
--- a/src/pubsub/ua_pubsub_manager.c
+++ b/src/pubsub/ua_pubsub_manager.c
@@ -183,7 +183,7 @@ UA_ReserveId_createId(UA_Server *server,  UA_NodeId sessionId, UA_String transpo
         next_id = next_id_writer;
 
     for(;numberOfIds > 0;numberOfIds--) {
-#ifndef UA_ENABLE_REDUCED_ITERATIONS_FOR_TESTING        
+#ifndef UA_ENABLE_REDUCED_ITERATIONS_FOR_TESTING
         if(next_id < UA_RESERVEID_FIRST_ID)
             next_id = UA_RESERVEID_FIRST_ID;
 #else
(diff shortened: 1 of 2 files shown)


src/pubsub/ua_pubsub_manager.c
```diff
@@ -707,20 +707,55 @@ UA_PubSubManager_generateUniqueGuid(UA_Server *server) {
...
+    UA_UInt16 port = 4840; /* default */
+    UA_StatusCode res = UA_parseEndpointUrl(&server->config.serverUrls[0], &hostname, &port, &path);
+    if(res != UA_STATUSCODE_GOOD) {
+        /* If the port cannot be extracted, the default port is used to generate the defaultPublisherId */
+        UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                     "PubSub generate defaultPublisherId failed. Could not extract the port from the url");
+    }
+    UA_UInt64 id = 0;
+    struct ifreq s;
+    strcpy(s.ifr_name, "eth0");
```
jpfr, line 744: We are not going to use platform-specific ioctl here.
It makes porting to embedded targets difficult.
In the future this can be retrieved from the eventloop.
For the time, comment this out and use random as default.

jpfr (commented):
The change is good in general.
But we cannot use platform specific API in the main library.

src/pubsub/ua_pubsub_manager.c
NoelGraf: That makes sense! An alternative would be to move the system call to the eventloop_posix_tcp and store the MAC address from there in the Connection Manager, for example.
Would that work?

jpfr:
Please make the proposed change.
We should include what we have and do a proper solution for the MAC in second PR.

src/pubsub/ua_pubsub_manager.c
```diff
@@ -719,21 +719,30 @@ UA_PubSubManager_generateUniqueGuid(UA_Server *server) {
...
+    UA_UInt64 id = 0;
     UA_Guid *ident = NULL;
     ident = UA_Guid_new();
     if(!ident) {
         UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
                      "PubSub generate defaultPublisherId failed.");
-        return;
+        return 0;
     }
     *ident = UA_Guid_random();
```
jpfr, line 732: A small nitpick that was already in the code:
UA_Guid_random returns a scalar.
So you can stack-allocate `ident` without doing the malloc/free dance.

jpfr:
👍

PR #5348, opened 2022-09-22
fix: Check for successful file creation

keba-estr:
fclose should only be called on valid FILE pointers, as fclose(0) is undefined.


jpfr:
Looks good. Merged. 👍

PR #5350, opened 2022-09-22
[WIP] feat(nodeset-loader): Integration of nodesetLoader repository

keba-uso:
Integrated the [nodesetLoader](https://github.com/open62541/nodesetLoader) functionality to the open62541.
The nodesetLoader is an XML-based API that loads the information model at runtime. This is an alternative
approach from the one with statically compiled information model (Python nodeset compiler).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,7 +77,7 @@ Makefile
 /exampleClient
 /exampleClient_legacy
 /src_generated/
-/build
+**/build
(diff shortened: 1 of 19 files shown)


jpfr:
Please integrate the NodeSetLoader as a git submodule instead of moving over the code.
As @andreasebner discussed with you.

PR #5355, opened 2022-09-28
[REVIEW] Feature/eventloop cached address udp

JanSurft:
Cache address of UDP connection for using sendto instead of connect + send

### DIFF
--- a/arch/eventloop_posix_udp.c
+++ b/arch/eventloop_posix_udp.c
@@ -962,7 +962,6 @@ UDP_shutdownConnection(UA_ConnectionManager *cm, uintptr_t connectionId) {
     return UA_STATUSCODE_GOOD;
 }
 
-
 static UA_StatusCode
 UDP_sendWithConnection(UA_ConnectionManager *cm, uintptr_t connectionId,
                        size_t paramsSize, const UA_KeyValuePair *params,
@@ -1088,12 +1087,11 @@ registerSocketAndDestinationForSend(size_t paramsSize, const UA_KeyValuePair *pa
                      "UDP\t| Configuring send multicast failed");
         return UA_STATUSCODE_BADINTERNALERROR;
     }
-    memcpy(&ufd->sendAddr, info->ai_addr, sizeof(struct sockaddr));
+    memcpy(&ufd->sendAddr, info->ai_addr, info->ai_addrlen);
     ufd->sendAddrLength = info->ai_addrlen;
 
     return res;
 }
-
 
 static UA_StatusCode
 UDP_openSendConnection(UA_ConnectionManager *cm,
--- a/arch/eventloop_posix_udp.c
+++ b/arch/eventloop_posix_udp.c
@@ -30,6 +30,8 @@
 /* A registered file descriptor with an additional method pointer */
 typedef struct {
     UA_RegisteredFD fd;
+    struct sockaddr_storage sendAddr;
+    socklen_t sendAddrLength;
     UA_ConnectionManager_connectionCallback connectionCallback;
 } UDP_FD;
 
@@ -127,7 +129,7 @@ getNetworkInterfaceFromParams(size_t paramsSize, const UA_KeyValuePair *params,
     /* Prepare the networkinterface string */
     const UA_String *networkInterface = (const UA_String*)
         UA_KeyValueMap_getScalar(params, paramsSize,
-                                 UA_QUALIFIEDNAME(0, "networkInterface"),
+                                 UA_QUALIFIEDNAME(0, "network-interface"),
                                  &UA_TYPES[UA_TYPES_STRING]);
     if(!networkInterface) {
         UA_LOG_DEBUG(logger, UA_LOGCATEGORY_NETWORK,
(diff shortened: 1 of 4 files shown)


arch/eventloop_posix_udp.c
```diff
@@ -30,6 +30,8 @@
 /* A registered file descriptor with an additional method pointer */
 typedef struct {
     UA_RegisteredFD fd;
+    struct sockaddr* sendAddr;
```
jpfr, line 33: Use `sockaddr_storage` directly instead of a pointer.
That is guaranteed to be large enough for all addresses. And you don't have to malloc.
https://pubs.opengroup.org/onlinepubs/000095399/basedefs/sys/socket.h.html

include/open62541/plugin/eventloop.h
```diff
@@ -311,6 +311,15 @@ struct UA_ConnectionManager {
                       void *application, void *context,
                       UA_ConnectionManager_connectionCallback connectionCallback);
 
+    /* Validate a Connection
+     * ~~~~~~~~~~~~~~~~~
+     * Validate a connection before it is opened. Same parameters are applicable
+     * as for `openConnection`.
+     */
+    UA_StatusCode
+    (*validateConnection)(UA_ConnectionManager *cm,
```
jpfr, line 320: What is the advantage of this compared to opening a "real" connection and closing it immediately?
Because you do open an actual connection internally.

Note that DNS lookups can take a lot of time.
So this validation should eventually become asynchronous with a callback - if we not only check the parameters syntactically.
Async DNS is something general for all protocols, not just UDP though...

jpfr (commented):
Nice!
Storing the sockaddr is a big advantage.
Comments inline.

include/open62541/plugin/eventloop.h
JanSurft: I create a socket internally but I do not bind it and it will not be registered anywhere in the eventloop.
The socket opening is only a way to check that the given connection parameters are valid.

Maybe the function should be called "validateParameters" instead of "validateConnection"?

Anyway it will be put into a separate PR

arch/eventloop_posix_udp.c
```diff
@@ -77,6 +79,12 @@ UDP_setNoSigPipe(UA_FD sockfd) {
     return UA_STATUSCODE_GOOD;
 }
```
jpfr, line 81: A normal free does the same with less indirection.

arch/eventloop_posix_udp.c
```diff
@@ -290,7 +298,8 @@ setConnectionConfig(UA_FD socket, const UA_KeyValuePair *connectionProperties,
     UA_String hostnameParam = UA_STRING("hostname");
     UA_String portParam = UA_STRING("port");
     UA_String listenHostnamesParam = UA_STRING("listen-hostnames");
-    UA_String listenPortParam= UA_STRING("listen-port");
+    UA_String listenPortParam = UA_STRING("listen-port");
+    UA_String networkInterfaceParam = UA_STRING("networkInterface");
```
jpfr, line 302: Keep a similar style for parameters in the public interface.
This is the only place with camlCase.
network-interface or just interface?

arch/eventloop_posix_udp.c
```diff
@@ -377,6 +387,8 @@ setupSendMulticastIPv4(UA_FD socket, struct sockaddr_in *addr, size_t paramsSize
         memcpy(interfaceAsChar, netif.data, netif.length);
         interfaceAsChar[netif.length] = 0;
 
+        UA_Array_delete(netif.data, netif.length, &UA_TYPES[UA_TYPES_BYTE]);
```
jpfr, line 390: UA_String_clear

arch/eventloop_posix_udp.c
```diff
@@ -416,6 +428,8 @@ setupListenMulticastIPv4(UA_FD socket, size_t paramsSize, const UA_KeyValuePair
...
```
jpfr, line 431: UA_String_clear

arch/eventloop_posix_udp.c
```diff
@@ -965,16 +979,17 @@ UDP_sendWithConnection(UA_ConnectionManager *cm, uintptr_t connectionId,
...
         do {
             UA_LOG_DEBUG(cm->eventSource.eventLoop->logger,
                          UA_LOGCATEGORY_NETWORK,
                          "UDP %u\t| Attempting to send", (unsigned)connectionId);
             size_t bytes_to_send = buf->length - nWritten;
-            n = UA_send((UA_FD)connectionId,
+            n = UA_sendto((UA_FD)connectionId,
                         (const char*)buf->data + nWritten,
-                        bytes_to_send, flags);
+                          bytes_to_send, flags, (struct sockaddr*) &ufd->sendAddr, ufd->sendAddrLength);
```
jpfr, line 992: Nice!

jpfr (commented):
Nearly there.
Comments inline.

arch/eventloop_posix_udp.c
```diff
@@ -1075,21 +1088,13 @@ registerSocketAndDestinationForSend(size_t paramsSize, const UA_KeyValuePair *pa
...
-       UA_ERRNO != UA_INPROGRESS &&
-       UA_ERRNO != UA_WOULDBLOCK) {
-        UA_LOG_SOCKET_ERRNO_WRAP(
-            UA_LOG_WARNING(logger, UA_LOGCATEGORY_NETWORK,
-                           "UDP\t| Connecting the socket to %s failed (%s)",
-                           hostname, errno_str));
-        UA_close(newSock);
-        return UA_STATUSCODE_BADDISCONNECT;
-    }
+    memcpy(&ufd->sendAddr, info->ai_addr, sizeof(struct sockaddr));
```
jpfr, line 1091: Use `info->ai_addrlen` instead of `sizeof(struct sockaddr)`.
The current code might not memcpy over everything.
Is it possible that struct sockaddr ist just a "basetype" with additional stuff depending on the actual protocol?

jpfr (commented):
Looks very good now.
Just one last comment.

jpfr:
👍

PR #5358, opened 2022-09-30
adds missing counters

Kallelele:
Update with some counters I think is missing.


andreasebner:
Thank's for your contribution. The proposed extension is fixed/obsolete in newer versions.

PR #5360, opened 2022-10-03
fix(core): fix issues with checking method input arguments.

MurzynJ:
Checking method input arguments was failing with BadInvalidArgument in valid use cases:

1) Input argument is an empty array of structures (ExtensionObjects).
2) Input argument is a structure of a non-NS0 datatype, or an abstract data type.

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -1037,7 +1037,7 @@ compatibleValue(UA_Server *server, UA_Session *session, const UA_NodeId *targetD
                 const UA_UInt32 *targetArrayDimensions, const UA_Variant *value,
                 const UA_NumericRange *range, const char **reason) {
     /* Empty value */
-    if(!value->type) {
+    if(UA_Variant_isEmpty(value)) {
         /* Empty value is allowed for BaseDataType */
         if(UA_NodeId_equal(targetDataTypeId, &UA_TYPES[UA_TYPES_VARIANT].typeId) ||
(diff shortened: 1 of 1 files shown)


src/server/ua_services_attribute.c
```diff
@@ -1150,13 +1158,15 @@ adjustValueType(UA_Server *server, UA_Variant *value,
...
         return;
 
-    const UA_DataType *targetDataType = UA_findDataType(targetDataTypeId);
-    if(!targetDataType)
-        return;
-
     /* Unwrap ExtensionObject arrays if they all contain the same DataType */
     unwrapEOArray(server, value);
 
+    const UA_DataType *targetDataType = UA_findDataType(targetDataTypeId);
```
jpfr, line 1164: We should use UA_findDataTypeWithCustom here.
So we can take the custom types loaded into the server into account.

jpfr (commented):
Yeah, this is an important point we need to get right.

Let's think about the type-checking from the perspective of a user that writes callbacks.
Either for a DataSource in a Variable or a method callback in a MethodNode.

The user might think that defining a datatype is enough to guarantee that only values for that type are forwarded to the callback. And so many users don't add explicit type checks in their callback.

If we now forward an ExtensionObject to the user ("leave it for the upper layers"), then code without explicit checks on the user level will break.

What are the guarantees that we make for the user on which kind of values he well receive in the callbacks after they have passed the type-checking?
That should be a very simple rule that we document and never ever break.

For the empty-extensionobject array I'm even willing to auto-cast that to the correct type (if non-abstract and found).

More comments inline.

PR #5361, opened 2022-10-03
fix(pubsub): Set the messageReceiveTimout when pubsub monitoring is enabled

NoelGraf:
Bug fix in mqtt subscriber example. Set the messageReceiveTimout when pubsub monitoring is enabled. Also, change the port to allow the publisher and subscriber examples to run simultaneously.


jpfr:
Thx

PR #5362, opened 2022-10-03
fix(build): do not install git files

UVV-gh:
Otherwise git-related directories in a submodule end up to be installed


jpfr:
Nice, thx

UVV-gh:
@jpfr that was quick, thanks :) FTR, the patch was needed for yocto build. I submitted a yocto recipe for open62541 to meta-opemembedded repository (in review ATM)

PR #5363, opened 2022-10-04
[REVIEW]: Implement reverse connect on the server side

basyskom-jvoe:
This PR contains the first steps to support reverse connect on the server side.

Open questions:
- Do we need a callback on connect / disconnect if there is an automatic reconnect mechanism in place?
- Should this be always available or do we need to make it a configurable feature?


src/server/ua_server_binary.c
```diff
@@ -996,3 +997,202 @@ UA_Server_networkCallback(UA_ConnectionManager *cm, uintptr_t connectionId,
...
+        if(server->serverConnectionsSize >= UA_MAXSERVERCONNECTIONS) {
+            UA_LOG_WARNING(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                           "Cannot register server socket - too many already open");
+            cm->closeConnection(cm, connectionId);
+            context->status = REVERSE_CONNECT_STATUS_DISCONNECTED;
+            return;
+        }
+
+        /* Find and use a free connection slot */
+        server->serverConnectionsSize++;
```
jpfr, line 1122: The ServerConnections are for listening sockets.
Not connections to individual clients.

src/server/ua_server_internal.h
```diff
@@ -87,12 +87,29 @@ typedef enum {
 /* Maximum numbers of sockets to listen on */
 #define UA_MAXSERVERCONNECTIONS 16
 
+typedef enum {
```
jpfr, line 90: The reverse connect “spawns” a SecureChannel.
Does the reverse connect structure disappear when the channel opens?
Could we just have reverse connect as a special state in the SecureChannel itself.
So we don’t need duplicate management structures.

src/server/ua_server.c
```diff
@@ -620,6 +620,108 @@ UA_Server_createServerConnection(UA_Server *server, const UA_String *serverUrl)
...
+            context->status = REVERSE_CONNECT_STATUS_DISCONNECTED;
+        } else {
+            context->status = REVERSE_CONNECT_STATUS_CONNECTING;
+        }
+    }
+
+    return res;
+}
+
+UA_StatusCode UA_Server_addReverseConnect(UA_Server *server, UA_String url, UA_UInt64 *handle) {
```
jpfr, line 664: I miss the cleanup of pending reverse connects during server shutdown…
Is this via the closing tcp socket callback?

jpfr (commented):
Nice!
Comments inline.

src/server/ua_server.c
basyskom-jvoe: Yes, this is done automatically when the socket state changes on shutdown.

src/server/ua_server_binary.c
jpfr: We already distinguish between listen-sockets and "active connections".
Let's not add a third kind. To me, reverse connections are also "active connections".
You can immediately create a SecureChannel to store the state. And add new status in the SecureChannelState enum.

src/server/ua_server_internal.h
jpfr: You could achieve the same going directly to a SecureChannel without the additional indirection.
And define the repeated check on the SecureChannel.

src/server/ua_server_binary.c
basyskom-jvoe: @jpfr If I understand you correctly, I should add a secure channel to server->channels and extend the UA_SecureChannel struct with the necessary information (isReverseConnect flag, URL for reverse connect, etc...)?

src/server/ua_server_internal.h
basyskom-jvoe: Is it possible to create a secure channel without having a connection id?

icozzz:
Yes, I hope that the callback function can notify me about the connection and disconnection events.

basyskom-jvoe:
> Yes, I hope that the callback function can notify me about the connection and disconnection events.

Good Idea. @jpfr Do you know any precedence for such a callback in the open62541 server so we can keep the API consistent?

jpfr:
We have quite a few callback methods.
For example the state callback for the client:
https://github.com/open62541/open62541/blob/87798df12914e1f56925e99f85186af6d8725d35/include/open62541/client.h#L155

basyskom-jvoe:
> We have quite a few callback methods. For example the state callback for the client:
> 
> https://github.com/open62541/open62541/blob/87798df12914e1f56925e99f85186af6d8725d35/include/open62541/client.h#L155

@jpfr Thanks for the reply. In our case, i'd suggest to include a pointer to the server, the secure channel state and the reverse connect handle that was returned in the UA_Server_addReverseConnect() function.
Are my other assumptions (see my replies to your other comments) correct?

basyskom-jvoe:
@jpfr I have updated my implementation.
A reverse connect no longer blocks a server connection slot and a state callback has been added

include/open62541/server.h
```diff
@@ -1886,6 +1891,22 @@ typedef struct {
 UA_ServerStatistics UA_EXPORT
 UA_Server_getStatistics(UA_Server *server);
 
+/* Reverse connect */
+typedef enum {
+    REVERSE_CONNECT_STATUS_INITIAL = 0,
+    REVERSE_CONNECT_STATUS_CONNECTING,
+    REVERSE_CONNECT_STATUS_CONNECTED,
+    REVERSE_CONNECT_STATUS_DISCONNECTED
```
jpfr, line 1899: What happens if the TCP connection opens, but the connection SecureChannel fails to be created.
For example due to a problem with certificates.
Will there still be a _CONNECTED or just _DISCONNECTED?

It should be documented how the callback relates to the SecureChannel state.

jpfr (commented):
Looks good!
I only reviewed the public API.
With a few comments.
After resolving these, the implementation review follows.

include/open62541/server.h
```diff
@@ -1886,6 +1891,22 @@ typedef struct {
 UA_ServerStatistics UA_EXPORT
 UA_Server_getStatistics(UA_Server *server);
 
+/* Reverse connect */
+typedef enum {
+    REVERSE_CONNECT_STATUS_INITIAL = 0,
+    REVERSE_CONNECT_STATUS_CONNECTING,
+    REVERSE_CONNECT_STATUS_CONNECTED,
```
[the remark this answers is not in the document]
basyskom-jvoe: This refers to the secure channel.
Each retry will start with a callback for CONNECTING followed by a callback for CONNECTED or DISCONNECTED.

I'd personally prefer to keep reconnecting as long as the reverse connect is not removed by the user.
Otherwise users would need to keep book about which handle is for which URL even if they just care about the client being able to communicate with the server.

include/open62541/server.h
```diff
@@ -1886,6 +1891,22 @@ typedef struct {
...
+    REVERSE_CONNECT_STATUS_DISCONNECTED
+} UA_ReverseConnectState;
+
+typedef void (*UA_Server_ReverseConnectStateCallback)(UA_Server *server, UA_UInt64 handle,
+                                                      UA_ReverseConnectState state, void *context);
+
```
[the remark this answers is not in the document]
basyskom-jvoe: This already done in UA_Server_addReverseConnect.()

include/open62541/server.h
basyskom-jvoe: I'd say it's just a DISCONNECTED. The user can't really do anything about it on the client side, so I see no point in setting the state to CONNECTED. But we could split the CONNECTED state into two states (connection has been opened vs. secure channel is up). This could permit debugging connection issues, for example firewall problems vs. wrong security configuration in client or server.

include/open62541/server.h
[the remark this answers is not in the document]
jpfr: So there is no real difference between a fresh and a disconnected reverse connect state.

include/open62541/server.h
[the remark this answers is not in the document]
jpfr: Good.
I didn’t look at the implementation.
Please document as part of the public api.

include/open62541/server.h
jpfr: Debugging should be possible from the log output.
Having just _connected works for me as well.
But it should be clearly documented what _connected means.

we should use _disconnected for fresh reverse connections as well. See above.

basyskom-jvoe:
There is a special case when the client is configured to use an authentication token with wrong credentials.
With the current implementation, this leads to a state change to CONNECTED followed by DISCONNECTED once the activateSession request has failed.

If we really want to distinguish, there are three states: TCP connected, secure channel established, session activated.

Is there a way to find out if there is an activated session on a secure channel?

jpfr:
> There is a special case when the client is configured to use an authentication token with wrong credentials.
With the current implementation, this leads to a state change to CONNECTED followed by DISCONNECTED once the activateSession request has failed.

> If we really want to distinguish, there are three states: TCP connected, secure channel established, session activated.

There can be more than one session bound to a SecureChannel.
It is enough to have the state value of reverse connect reflect the SecureChannel.
It gets too complicated otherwise.

> Is there a way to find out if there is an activated session on a secure channel?

Of course. There is a linked list of the bound sessions in the SecureChannel struct.

basyskom-jvoe:
> It is enough to have the state value of reverse connect reflect the SecureChannel. It gets too complicated otherwise.
In this case, the implementation should be complete now.
@jpfr If you're fine with the API, I will extend the PR with some basic unit tests.

include/open62541/server.h
```diff
@@ -1886,6 +1891,57 @@ typedef struct {
...
+/**
+ * Registers a reverse connect in the server.
+ * The server periodically attempts to establish a connection if the initial connect fails
+ * or if the connection breaks.
+ *
+ * @param server The server object
+ * @param url The URL of the remote client
+ * @param stateCallback The callback which will be called on state changes
+ * @param callbackContext The context for the state callback
+ * @param handle The handle of the reverse connect
```
jpfr, line 1929: Can the handle be null?

include/open62541/server.h
```diff
@@ -1886,6 +1891,57 @@ typedef struct {
...
+typedef enum {
+    UA_REVERSECONNECTSTATE_DISCONNECTED = 0,
+    UA_REVERSECONNECTSTATE_CONNECTING,
+    UA_REVERSECONNECTSTATE_CONNECTED,
+} UA_ReverseConnectState;
+
+/**
+ * The reverse connect state change callback is called whenever the state of a reverse connect
+ * is modified by a connection attempt, a successful connection or a connection loss.
+ *
```
jpfr, line 1910: Break lines at about 90 characters.
Otherwise the generated pdf will look bad.

include/open62541/server.h
```diff
@@ -1886,6 +1891,57 @@ typedef struct {
 UA_ServerStatistics UA_EXPORT
 UA_Server_getStatistics(UA_Server *server);
 
+/**
+  * Reverse Connect
+  * ---------------
+  *
```
jpfr, line 1897: Make explicit that the connection status is for a SecureChannel.
Does connected mean we have the HEL/ACK handshake or does it mean the SecureChannel is open?
Not the session on top.

jpfr (commented):
The API is good.
It only needs more clarification to avoid misunderstandings.
Looking forward to this!

include/open62541/server.h
```diff
@@ -1886,6 +1891,57 @@ typedef struct {
...
+
+/**
+ * The reverse connect state change callback is called whenever the state of a reverse connect
+ * is modified by a connection attempt, a successful connection or a connection loss.
+ *
+ * Each attempt to establish a connection will lead to a call with UA_REVERSECONNECTSTATE_CONNECTING
+ * followed by either UA_REVERSECONNECTSTATE_CONNECTED if a secure channel was successfully established
+ * or UA_REVERSECONNECTSTATE_DISCONNECTED if the connection attempt was not successful.
+ * A connection break leads to a call with UA_REVERSECONNECTSTATE_DISCONNECTED followed by
+ * UA_REVERSECONNECTSTATE_CONNECTING when the next reconnect attempt is made.
```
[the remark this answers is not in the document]
basyskom-jvoe: From my point of view, the handle refers to a reverse connect configuration and not to a connection.

I guess most users will use reverse connect to get around strange firewall configurations and just want to enable a client to talk to the server. Not doing reconnect attempts automatically on connection  loss would mean that the user has to implement the state callback and keep tabs on the association between handle and remote client URL to add a new reverse connect once the connection has been lost.

Doing reconnects by default would cater to the "lazy" user and anyone who would want control could implement the state change callback and remove the reverse connect configuration from inside the callback by using the handle.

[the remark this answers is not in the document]
jpfr: That works as well.
The importance is to have no ambiguity in the documentation.
So that people can rely on a behavior.

That means a connection is never ‚disconnected‘.
If a reconnect attempt is made immediately.
So the. We only have ‚connecting‘ and ‚connected‘ as status.

basyskom-jvoe:
@jpfr Apart from our discussion on the reconnect behavior, the API should be done now.

include/open62541/server.h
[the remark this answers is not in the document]
basyskom-jvoe: @jpfr The attempt is made periodically by the repeated callback. Depending on when the connection breaks, this could be right away or after a few seconds.
What about a WaitingForReconnect state?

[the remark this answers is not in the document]
basyskom-jvoe: @jpfr Sounds good, I have updated the PR accordingly

src/server/ua_server_binary.c
```diff
@@ -996,3 +997,206 @@ UA_Server_networkCallback(UA_ConnectionManager *cm, uintptr_t connectionId,
...
+        context->currentConnection.connectionManager = cm;
+        return;
+    }
+
+    /* The connection is closing. This is the last callback for it. */
+    if(state == UA_CONNECTIONSTATE_CLOSING && context->channel) {
+        deleteServerSecureChannel(server, context->channel);
+        context->channel = NULL;
+        setReverseConnectState(server, context, UA_SECURECHANNELSTATE_CONNECTING);
+        context->currentConnection.connectionId = 0;
```
jpfr, line 1135: If the connection is lost, try to establish a new connection immediately.
So don’t wait for the interval.

include/open62541/server.h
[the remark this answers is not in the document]
jpfr: Great!
The API is good for me.
Just trigger us when we should take a look before merging.

basyskom-jvoe:
@jpfr I have added some tests but they have not been pushed yet because I have problems with a rebase to the current master.
My previous implementation was based on commit 0fb4efd617 and it was sufficient to do one iteration of the event loop to fully close a connection so the connection context could be deleted:
```
server->config.eventLoop->run(server->config.eventLoop, 0)
```
After the rebase to the current master, I get a heap-use-after-free on UA_Server_run_shutdown() with the same code because the connection callback is called again during the final iterations here:
```
while(el->state != UA_EVENTLOOPSTATE_STOPPED &&
      el->state != UA_EVENTLOOPSTATE_FRESH &&
      res == UA_STATUSCODE_GOOD)
      res = el->run(el, 100); /* Iterate until stopped */
```

Have there been any changes in how the closing of a TCP connection is handled?

jpfr:
> Have there been any changes in how the closing of a TCP connection is handled?

Not that I am aware.

The documented behavior is that the network callback is called a single time with `UA_CONNECTIONSTATE_CLOSING` before the connection is deleted.

Maybe for reverse connect callback you forward a different connection state to the underlying callback?

Please add logging of the connection state to the top of the
- normal connection callback
- reverse connect connection callback

And paste the log output here.

basyskom-jvoe:
@jpfr I have found the issue.
After the rebase, an openConnection() attempt was accepted by the connection manager before UA_Server_run_startup(), so my code in UA_Server_run_startup() was starting a second connection for the same context.

I have updated the PR with the rebased commits.

src/server/ua_server.c
```diff
@@ -616,6 +616,120 @@ UA_Server_createServerConnection(UA_Server *server, const UA_String *serverUrl)
...
+        UA_LOG_WARNING(&config->logger, UA_LOGCATEGORY_SERVER,
+                       "OPC UA URL is invalid: %.*s",
+                       (int)url.length, url.data);
+        return res;
+    }
+
+    if (SLIST_EMPTY(&server->reverseConnects))
+        UA_Server_setReverseConnectRetryCallback(server, true);
+
+    reverse_connect_context *newContext = (reverse_connect_context *)calloc(1, sizeof(reverse_connect_context));
```
jpfr, line 683: always check for malloc fails.

src/server/ua_server.c
```diff
@@ -807,6 +921,19 @@ UA_Server_run_shutdown(UA_Server *server) {
...
 
+    /* Mark all reverse connects as destroying */
+    reverse_connect_context *rev = NULL;
+    SLIST_FOREACH(rev, &server->reverseConnects, next) {
+        rev->destruction = true;
+        if (rev->currentConnection.connectionId) {
+            rev->currentConnection.connectionManager->closeConnection(rev->currentConnection.connectionManager,
+                                                                      rev->currentConnection.connectionId);
+        } else if (rev->stateCallback) {
+            rev->stateCallback(server, rev->handle, UA_SECURECHANNELSTATE_CLOSED,
```
jpfr, line 932: You can use `setReverseConnectState` in every case (not just for else-if) to signal state changes.

jpfr (commented):
Nice!
A few last comments inline.
Can be merged then.

src/server/ua_server.c
```diff
@@ -616,6 +616,120 @@ UA_Server_createServerConnection(UA_Server *server, const UA_String *serverUrl)
...
+            SLIST_REMOVE(&server->reverseConnects, rev, reverse_connect_context, next);
+
+            if (rev->currentConnection.connectionId) {
+                rev->destruction = true;
+                /* Request disconnect and run the event loop for one iteration */
+                if (rev->currentConnection.connectionId) {
+                    rev->currentConnection.connectionManager->closeConnection(
+                                rev->currentConnection.connectionManager,
+                                rev->currentConnection.connectionId);
+                    server->config.eventLoop->run(server->config.eventLoop, 0);
```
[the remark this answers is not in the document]
basyskom-jvoe: @jpfr  IMO the user should also be able to kick out a connected client by removing the associated reverse connect.
Would it be possible to add a deferred callback to delete the context or would that be run in the next event loop iteration before the socket has been closed?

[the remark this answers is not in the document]
jpfr: Yes!
What can be done is to remove the reverse connect from the linked list.
And add a "delayed callback" to the eventloop.
That will be automatically executed in the next iteration of the eventloop.
So that nobody else uses the pointer.

jpfr:
Nice work! Merged.

PR #5364, opened 2022-10-04
fix(client): Possible memory leak when deleting client.

dapaulid:
Cleanup of EventLoop and Logger was skipped if there are no SecurityPolicies specified.


jpfr:
Nice, thx.

PR #5370, opened 2022-10-07
fix(examples): Fix usage of publisherId in pubsub_interrupt_publish.c

JanAltenberg:
This file was missed in commit fd76d7faaea38843eb74f578f6b82bf8252b98b7, which breaks build for a few examples.


jpfr:
Nice!
Thanks @JanAltenberg

PR #5372, opened 2022-10-07
[REVIEW] feat(core): more utilities for keyvaluemaps

JanSurft:
Add some utilities for KeyValueMaps:

- copy
- contains
- merge

## TODO
- [x] Implement functional logic
- [x] Unit tests
- [x] Check restructure of KeyValueMap API


include/open62541/util.h
```diff
@@ -68,6 +68,57 @@ UA_EXPORT UA_StatusCode
...
+ *
+ * @param dst The KeyValueMap to store the resulting merge in. The size can either be an upper bound (rhsCout + lhsCount)
+ *            or can be calculated beforehand by using UA_KeyValueMap_countMergedMembers
+ * @param lhs The left hand side operand of the KeyValueMap merge
+ * @param lhsCount The count of entries in the lhs KeyValueMap
+ * @param rhs The right hand side operand of the KeyValueMap merge
+ * @param rhsCount The count of entries in the rhs KeyValueMap
+ */
+UA_EXPORT void
+UA_KeyValueMap_merge(UA_KeyValuePair *dst, UA_KeyValuePair *lhs, size_t lhsCount, UA_KeyValuePair *rhs, size_t rhsCount);
```
jpfr, line 120: This getting a bit too funky and complicated as an API.
There's a ton of edge cases that have to be handled when calling into _merge.

The KeyValueMap in general is still very new and not used in a release.
So we can still correct suboptimal design.

I would define
```c
struct UA_KeyValueMap {
    size_t size;
    UA_KeyValuePair  *map;
}
```
and then simplify to
```
UA_EXPORT UA_StatusCode
UA_KeyValueMap_merge(UA_KeyValueMap *lhs, const UA_KeyValuePair *rhs);
```

Note that the _set method already has "upsert" semantics.
So this is just a faster version of looping over the rhs and inserting everything.

jpfr (commented):
Pretty cool.
See inline. Let's think about the entire API..

include/open62541/util.h
JanSurft: `KeyValuePair[]` is used throughout the standard whenever a function is accepting a variadic number of input parameters for specific things. E.g. the `ConnectionProperties` in the `PubSubConnectionType`.
I think the current ratio is, to have an array of KeyValuePair being a KeyValueMap and being able to use this KeyValueMap as input to the functional equivalents of such properties or parameters in the code.
If we on the other hand change the KeyValueMap to be its own actual type it will not have an equivalent in the OPCUA Type System (or will it?).
Anyway, the proposed way of merging via the `set` method sounds good. If one needs to keep an original `lhs` untouched, one can just use `copy` before.

The API of `countMerged` and `merge` was initially designed like that, because KeyValuePair arrays often where allocated on the stack, which would prevent an inplace enlargement of a KeyValuePair array.

JanSurft: And of course, I also generally think it would be better to have an actual `UA_KeyValueMap` structure like you proposed, it would simplify a lot of places where we use pointer to KeyValuePair array and its size.

jpfr: Great. Then let's add the `UA_KeyValueMap` structure before the current API gets used in the wild.

> But with your proposed API, one would just have to make sure that the input arguments are heap allocated.

I think this information is generally known in places that would call _merge.

src/ua_util.c
```diff
@@ -248,38 +256,63 @@ UA_KeyValueMap_set(UA_KeyValuePair **map, size_t *mapSize,
...
+    return UA_Array_appendCopy((void**) &map->map, &map->size, &pair,
                                &UA_TYPES[UA_TYPES_KEYVALUEPAIR]);
 }
 
+UA_StatusCode
+UA_KeyValueMap_setScalar(UA_KeyValueMap *map,
+                         const UA_QualifiedName key,
+                         void * UA_RESTRICT p,
+                         const UA_DataType *type) {
+    UA_Variant *v = UA_Variant_new();
```
jpfr, line 268: This can be stack-allocated.

include/open62541/util.h
```diff
@@ -36,6 +36,11 @@ typedef enum {
     UA_TIMER_HANDLE_CYCLEMISS_WITH_BASETIME
 } UA_TimerPolicy;
 
+typedef struct {
+    size_t size;
```
jpfr, line 40: mapSize

include/open62541/util.h
```diff
@@ -47,27 +52,77 @@ typedef enum {
  * invalidates pointers into the previous array. If the key exists already, the
  * value is overwritten. */
 UA_EXPORT UA_StatusCode
-UA_KeyValueMap_set(UA_KeyValuePair **map, size_t *mapSize,
+UA_KeyValueMap_set(UA_KeyValueMap *map,
                    const UA_QualifiedName key,
                    const UA_Variant *value);
+UA_EXPORT UA_StatusCode
+UA_KeyValueMap_setScalar(UA_KeyValueMap *map,
```
jpfr, line 59: indent spaces

src/ua_util.c
```diff
@@ -248,38 +256,63 @@ UA_KeyValueMap_set(UA_KeyValuePair **map, size_t *mapSize,
...
 
-UA_StatusCode
-UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
+void
+UA_KeyValueMap_delete(UA_KeyValueMap *map) {
+    UA_Array_delete(map->map, map->size, &UA_TYPES[UA_TYPES_KEYVALUEPAIR]);
+    UA_free(map);
+}
+
+                      UA_StatusCode
```
jpfr, line 311: indent space

src/ua_util.c
```diff
@@ -294,14 +327,48 @@ UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
...
+        map->size = s - 1; /* In case resize fails, keep the longer original
                            * array around. Resize never fails when reducing
                            * the size to zero. Reduce the size integer in
                            * any case. */
         return UA_STATUSCODE_GOOD;
     }
     return UA_STATUSCODE_BADNOTFOUND;
 }
+
+UA_KeyValueMap*
```
jpfr, line 342: space before star

src/ua_util.c
```diff
@@ -294,14 +327,48 @@ UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
...
+UA_KeyValueMap_copy(UA_KeyValueMap *map) {
+
+    UA_KeyValuePair *copy = (UA_KeyValuePair *) UA_calloc(sizeof(UA_KeyValuePair), map->size);
+    if(!copy) {
+        return NULL;
+    }
+
+    for(size_t i = 0; i < map->size; i++) {
+        UA_KeyValuePair_copy(&map->map[i], &copy[i]);
```
jpfr, line 351: handle failure -> free all and return NULL

jpfr (commented):
a few details

include/open62541/util.h
```diff
@@ -41,6 +41,8 @@ typedef struct {
     UA_KeyValuePair *map;
 } UA_KeyValueMap;
```
jpfr, line 43: I see why you want to do this.
But no, please continue to use NULL instead of a sentinel pointer.
This is a public API so we have to protect against NULL anyway.
And it might happen easily that by mistake we load elements into the “const” empty value.

JanSurft: What do you think about defining a constant and use that instead (see latest commit). Like this we still have the idiomatic use of that KEYVALUEMAP literal in places where we want to pass NULL. It makes the code easier to read in my opinion. 

User code may or may not use that constant.
```c
#define UA_KEYVALUEMAP_NULL NULL
```

include/open62541/util.h
```diff
@@ -47,27 +54,84 @@ typedef enum {
  * invalidates pointers into the previous array. If the key exists already, the
  * value is overwritten. */
 UA_EXPORT UA_StatusCode
-UA_KeyValueMap_set(UA_KeyValuePair **map, size_t *mapSize,
+UA_KeyValueMap_set(UA_KeyValueMap *map,
                    const UA_QualifiedName key,
                    const UA_Variant *value);
+UA_EXPORT UA_StatusCode
```
jpfr, line 60: newline between declarations

include/open62541/util.h
```diff
@@ -47,27 +54,84 @@ typedef enum {
...
 UA_EXPORT UA_StatusCode
-UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
+UA_KeyValueMap_remove(UA_KeyValueMap *map,
                       const UA_QualifiedName key);
 
+UA_EXPORT UA_KeyValueMap *
+UA_KeyValueMap_new(size_t size_hint);
+
+UA_EXPORT void
+UA_KeyValueMap_delete(UA_KeyValueMap *map);
```
jpfr, line 94: Good idea to stay in the nomenclature of the remaining SDK!
In that case you can add UA_KeyValueMap_clear.
There are already use cases for that in this code.

include/open62541/util.h
```diff
@@ -47,27 +54,84 @@ typedef enum {
...
+/**
+ * Create a deep copy of the given KeyValueMap. Function will allocate memory for
+ * the newly create KeyValueMap which needs to be freed by the user of the function
+ * accordingly.
+ * @param map The KeyValueMap to be copeid
+ * @param mapSize The size of the KeyValueMap
+ * @return A pointer to the copied KeyValueMap, NULL if memory allocation failed.
+ */
+UA_EXPORT UA_KeyValueMap*
+UA_KeyValueMap_copy(UA_KeyValueMap *map);
```
jpfr, line 105: take two input pointers and return a statuscode.
The target map could be a stack-allocated struct.
Keep the API similar to `UA_copy` et al.
The input map can be const.

include/open62541/util.h
```diff
@@ -47,27 +54,84 @@ typedef enum {
...
+ * Check KeyValueMap for containment
+ * ---------------------------------
+ * Check if a given KeyValueMap contains an entry mapped by the provided key.
+ * @param map The KeyValueMap to check against
+ * @param mapSize The size of the KeyValueMap
+ * @param key The key to check for existence in the KeyValueMap
+ * @return true if an entry with the given key exists, false otherwise
+ */
+UA_EXPORT UA_Boolean
+UA_KeyValueMap_contains(UA_KeyValueMap *map, UA_QualifiedName key);
```
jpfr, line 117: input can be const

include/open62541/util.h
```diff
@@ -47,27 +54,84 @@ typedef enum {
...
                          const UA_DataType *type);
 
 /* Remove a single entry. To delete the entire map, use UA_Array_delete. */
 UA_EXPORT UA_StatusCode
-UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
+UA_KeyValueMap_remove(UA_KeyValueMap *map,
                       const UA_QualifiedName key);
 
+UA_EXPORT UA_KeyValueMap *
+UA_KeyValueMap_new(size_t size_hint);
```
jpfr, line 91: Remove the hint and return an empty map.
The hint is for users who "know what they are doing" (tm) when it comes to memory usage.
Those users know enough to calloc the map pointer themselves anyway.

The would make sense if we were tracking "size vs. count" separately in the struct.
Which we don't do.

include/open62541/util.h
```diff
@@ -47,27 +54,84 @@ typedef enum {
...
+ *
+ * @param dst The KeyValueMap to store the resulting merge in. The size can either be an upper bound (rhsCout + lhsCount)
+ *            or can be calculated beforehand by using UA_KeyValueMap_countMergedMembers
+ * @param lhs The left hand side operand of the KeyValueMap merge
+ * @param lhsCount The count of entries in the lhs KeyValueMap
+ * @param rhs The right hand side operand of the KeyValueMap merge
+ * @param rhsCount The count of entries in the rhs KeyValueMap
+ */
+UA_EXPORT void
+UA_KeyValueMap_merge(UA_KeyValueMap *lhs, UA_KeyValueMap *rhs);
```
jpfr, line 133: Should return a statuscode.
Make sure that - if the statuscode is bad - the input structs come back untouched.
The rhs can be const.

src/ua_util.c
```diff
@@ -225,17 +225,27 @@ UA_ByteString_fromBase64(UA_ByteString *bs,
     return UA_STATUSCODE_GOOD;
 }
 
+const UA_KeyValueMap UA_KEYVALUEMAP_NULL = {0};
+
+UA_KeyValueMap *
+UA_KeyValueMap_new(size_t sizeHint) {
+    UA_KeyValueMap *kvm = (UA_KeyValueMap*) UA_calloc(sizeof(UA_KeyValueMap), 1);
```
jpfr, line 232: Always check for malloc failure.
That happens on embedded and people actively ask for it.
Doesn't matter now if _new is a naked calloc.

src/ua_util.c
```diff
@@ -248,38 +258,70 @@ UA_KeyValueMap_set(UA_KeyValuePair **map, size_t *mapSize,
     UA_KeyValuePair pair;
     pair.key = key;
     pair.value = *value;
-    return UA_Array_appendCopy((void**)map, mapSize, &pair,
+    return UA_Array_appendCopy((void**) &map->map, &map->mapSize, &pair,
```
jpfr, line 261: Suppose map is NULL.
_get should still work.
But the appendCopy segfaults.

src/ua_util.c
```diff
@@ -248,38 +258,70 @@ UA_KeyValueMap_set(UA_KeyValuePair **map, size_t *mapSize,
...
+
 const UA_Variant *
-UA_KeyValueMap_get(const UA_KeyValuePair *map, size_t mapSize,
+UA_KeyValueMap_get(const UA_KeyValueMap *map,
                    const UA_QualifiedName key) {
-    for(size_t i = 0; i < mapSize; i++) {
-        if(map[i].key.namespaceIndex == key.namespaceIndex &&
-           UA_String_equal(&map[i].key.name, &key.name))
-            return &map[i].value;
+    for(size_t i = 0; i < map->mapSize; i++) {
```
jpfr, line 282: Check for map==NULL.

src/ua_util.c
```diff
@@ -248,38 +258,70 @@ UA_KeyValueMap_set(UA_KeyValuePair **map, size_t *mapSize,
...
+}
+
 UA_StatusCode
-UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
+UA_KeyValueMap_remove(UA_KeyValueMap *map,
                       const UA_QualifiedName key) {
-    UA_KeyValuePair *m = *map;
-    size_t s = *mapSize;
+    UA_KeyValuePair *m = map->map;
+    size_t s = map->mapSize;
```
jpfr, line 324: check for map==NULL

src/ua_util.c
```diff
@@ -294,14 +336,55 @@ UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
...
                            * the size to zero. Reduce the size integer in
                            * any case. */
         return UA_STATUSCODE_GOOD;
     }
     return UA_STATUSCODE_BADNOTFOUND;
 }
+
+UA_KeyValueMap *
+UA_KeyValueMap_copy(UA_KeyValueMap *map) {
+
```
jpfr, line 353: check for map==NULL

src/ua_util.c
```diff
@@ -294,14 +336,55 @@ UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
...
+        return NULL;
+    }
+    newMap->map = copy;
+    newMap->mapSize = map->mapSize;
+    return newMap;
+}
+
+UA_Boolean
+UA_KeyValueMap_contains(UA_KeyValueMap *map, UA_QualifiedName key) {
+    for(size_t i = 0; i < map->mapSize; ++i) {
```
jpfr, line 377: check for map==NULL

src/ua_util.c
```diff
@@ -294,14 +336,55 @@ UA_KeyValueMap_delete(UA_KeyValuePair **map, size_t *mapSize,
...
+        if(UA_QualifiedName_equal(&map->map[i].key, &key)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+void
+UA_KeyValueMap_merge(UA_KeyValueMap *lhs, UA_KeyValueMap *rhs) {
```
jpfr, line 386: check for map==NULL

jpfr (commented):
Looks good!
I did a last pass in more detail. A few more comments still came up. 😏 
In general, for the public API, the "usage contract" must be 100% clear.
Either document that the methods must never ever be called with NULL-ptrs or document that the behavior for map==NULL is the same as with UA_KEYVALUEMAP_NULL.
Either works for me, but be consistent.

jpfr:
Rebased and merged. Nice!

PR #5385, opened 2022-10-17
ReserveIds improvements

MurzynJ:
Some bugs were found when testing the recently added ReservedIds method. This PR is proposing fixes. At first, the tests were extended and improved to reveal the bugs:

1) Memory leak - freeing the reserved id structure was incomplete in the case when reservation shall be cancelled due to deactivated session.
2) Broken ID comparison - the same DataSetWriterId was reserved again, even if the previously reserved one is already in use.


MurzynJ:
Clang Static Analyzer seems to be wrong reporting the error.

jpfr (commented):
A few general house-keeping observations.
It might be that the false positive in the static analysis is removed as a side-effect of solving them.

The logic for detecting free ReserveIds is awfully complicated.
Please pull the _isFree detection into a separate method that returns just a Boolean (free or not).

I don't get why next_id++ is done in the middle of the list iteration.
It seems we would want to restart the list-iteration when next_id is increased.
This is much easier then handle when _isFree is a dedicated method.

Is it actually a good id to always start the search with `first_id = 0x8000`?
We could store the last id that was added and start from that with a rollover at 0xFFFF.
That will reduce the number of trial-lookups.

Last, if we move `LIST_INIT(&pubSubManager->reserveIds)` to the pubsubmanager_init (which it is, because that just zeroes out), then we don't need the case distinction for `LIST_INSERT_HEAD`.

It's very good to see the very thorough tests that were added. 👍

MurzynJ:
> The logic for detecting free ReserveIds is awfully complicated. Please pull the _isFree detection into a separate method that returns just a Boolean (free or not).
> 
> I don't get why next_id++ is done in the middle of the list iteration. It seems we would want to restart the list-iteration when next_id is increased. This is much easier then handle when _isFree is a dedicated method.
> 
> Is it actually a good id to always start the search with `first_id = 0x8000`? We could store the last id that was added and start from that with a rollover at 0xFFFF. That will reduce the number of trial-lookups.
> 
> Last, if we move `LIST_INIT(&pubSubManager->reserveIds)` to the pubsubmanager_init (which it is, because it just zeroes out), then we don't need the case distinction for `LIST_INSERT_HEAD`.

@NoelGraf would you like to address Julius comments above, as they actually concern your code?

jpfr:
@NoelGraf made the improvements. And they are already merged.
Please rebased (there is a conflict). Hopefully the false positive are already gone with that.

MurzynJ:
> @NoelGraf made the improvements. And they are already merged. Please rebased (there is a conflict). Hopefully the false positive are already gone with that.

I rebased it and adapted the tests to the new logic (always incrementing the ID counter until it rolls over).

There is no longer false positive by static analyzer :+1:

But testing the case with counter rolling over is almost killing Valgrind... The Valgrind enabled tests take 1 hour. What is your recommendation @jpfr? Leave is as it is? Remove this test case? Refactor the test or implementation?

jpfr:
> But testing the case with counter rolling over is almost killing Valgrind... The Valgrind enabled tests take 1 hour. What is your recommendation @jpfr? Leave is as it is? Remove this test case? Refactor the test or implementation?

You could add a flag to reduce the number of iterations for the rollover. Just for testing purposes.
As an example we have a macro `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` that disables some security consistency validations for easier fuzzing.

MurzynJ:
Thanks for suggestion @jpfr . See how I implemented it.
BTW. there is a new Mantis issue #8415 which is asking for clarification in Part 14 about error handling in ReserveIds.

jpfr:
Looks great. Merged.
We can add the correct status codes once Mantis #8415 is clear.

PR #5387, opened 2022-10-18
[Review] feat (pubsub): SKS SecurityGroups

Muddasir-Shakil:
This PR is the second in series of SKS support. It adds the support for managing SecurityGroups on SKS server side. A securitygroup is an abstraction that represents the message security settings and security keys for a subset of NetworkMessages exchanged between Publishers and Subscribers. A SKS manages the SecurityGroups and maintains the mapping
between Roles and their access Permissions.

In the implementation two server API methods are added to create and remove the SecurityGroup in the server. With a securitygroup a keystorage with initial keys is also created in the server. The keystorage is updated periodically with new keys
by a callback function. The period of update cycle is set by KeyLifeTime parameter in SecurityGroup configuration. The conditional compilation with information model support is also added for SecurityGroup.

### DIFF
--- a/src/pubsub/ua_pubsub_manager.c
+++ b/src/pubsub/ua_pubsub_manager.c
@@ -779,6 +779,14 @@ UA_PubSubManager_delete(UA_Server *server, UA_PubSubManager *pubSubManager) {
     }
 
 #ifdef UA_ENABLE_PUBSUB_SKS
+    /* Remove the SecurityGroups */
+    UA_SecurityGroup *tmpSG1, *tmpSG2;
+    TAILQ_FOREACH_SAFE(tmpSG1, &server->pubSubManager.securityGroups,
+                       listEntry, tmpSG2)
+        removeSecurityGroup(server, tmpSG1->securityGroupNodeId);
+#endif
+
+#ifdef UA_ENABLE_PUBSUB_SKS
     /* Remove the keyStorages */
     UA_PubSubKeyStorage *ks, *ksTmp;
     LIST_FOREACH_SAFE(ks, &server->pubSubManager.pubSubKeyList, keyStorageList, ksTmp)
--- a/src/pubsub/ua_pubsub_manager.c
+++ b/src/pubsub/ua_pubsub_manager.c
@@ -781,9 +781,9 @@ UA_PubSubManager_delete(UA_Server *server, UA_PubSubManager *pubSubManager) {
 #ifdef UA_ENABLE_PUBSUB_SKS
     /* Remove the SecurityGroups */
     UA_SecurityGroup *tmpSG1, *tmpSG2;
-    TAILQ_FOREACH_SAFE(tmpSG1, &server->pubSubManager.securityGroups,
-                       listEntry, tmpSG2)
-        removeSecurityGroup(server, tmpSG1->securityGroupNodeId);
+    TAILQ_FOREACH_SAFE(tmpSG1, &server->pubSubManager.securityGroups, listEntry, tmpSG2) {
+        removeSecurityGroup(server, tmpSG1);
+    }
 #endif
 
 #ifdef UA_ENABLE_PUBSUB_SKS
--- a/src/pubsub/ua_pubsub_securitygroup.c
+++ b/src/pubsub/ua_pubsub_securitygroup.c
@@ -142,6 +142,7 @@ updateSKSKeyStorage(UA_Server *server, UA_SecurityGroup *securityGroup){
 
 static UA_StatusCode
 initializeKeyStorageWithKeys(UA_Server *server, UA_SecurityGroup *securityGroup) {
+    UA_LOCK_ASSERT(&server->serviceMutex, 1);
     size_t keyLength;
     UA_StatusCode retval = UA_STATUSCODE_BAD;
 
@@ -177,9 +178,9 @@ initializeKeyStorageWithKeys(UA_Server *server, UA_SecurityGroup *securityGroup)
 
     securityGroup->baseTime = UA_DateTime_nowMonotonic();
 
-    retval = UA_Server_addRepeatedCallback(server, (UA_ServerCallback)updateSKSKeyStorage,
-                                           securityGroup, securityGroup->config.keyLifeTime,
-                                           &securityGroup->callbackId);
+    retval = addRepeatedCallback(server, (UA_ServerCallback)updateSKSKeyStorage,
+                                 securityGroup, securityGroup->config.keyLifeTime,
+                                 &securityGroup->callbackId);
     if(retval != UA_STATUSCODE_GOOD)
         goto error;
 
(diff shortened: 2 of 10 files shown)


src/pubsub/ua_pubsub_manager.c
```diff
@@ -778,6 +778,14 @@ UA_PubSubManager_delete(UA_Server *server, UA_PubSubManager *pubSubManager) {
         UA_Server_removeStandaloneSubscribedDataSet(server, tmpSDS1->identifier);
     }
 
+#ifdef UA_ENABLE_PUBSUB_SKS
+    /* Remove the SecurityGroups */
+    UA_SecurityGroup *tmpSG1, *tmpSG2;
+    TAILQ_FOREACH_SAFE(tmpSG1, &server->pubSubManager.securityGroups,
+                       listEntry, tmpSG2)
+        removeSecurityGroup(server, tmpSG1->securityGroupNodeId);
```
jpfr, line 786: You already have the SG. No need to forward the id and do a second lookup.
removeSecurityGroup should take a pointer to the SG.
Any user of it can do a UA_SecurityGroup_findSGbyId before.
removeSecurityGroup then also does not have to return a statuscode. It isn't evaluated anyhow.

src/pubsub/ua_pubsub_securitygroup.c
```diff
@@ -0,0 +1,322 @@
...
+#endif
+
+    if(securityGroupNodeId)
+        UA_NodeId_copy(&newSecurityGroup->securityGroupNodeId, securityGroupNodeId);
+
+    if(server->pubSubManager.securityGroupsSize != 0) {
+        TAILQ_INSERT_TAIL(&server->pubSubManager.securityGroups, newSecurityGroup,
+                          listEntry);
+    } else {
+        TAILQ_INIT(&server->pubSubManager.securityGroups);
```
jpfr, line 252: You should call TAILQ_INIT in the _init of the PubSubManager.
Then you don't need to differentiate between cases here at all. Just do a INSERT_TAIL.

src/pubsub/ua_pubsub_securitygroup.c
```diff
@@ -0,0 +1,322 @@
...
+    retVal |= policy->symmetricModule.generateNonce(policy->policyContext, &seed);
+    if(retVal != UA_STATUSCODE_GOOD)
+        return retVal;
+
+    retVal = policy->symmetricModule.generateKey(policy->policyContext, &secret, &seed, key);
+    return retVal;
+}
+
+static void
+updateSKSKeyStorage(UA_Server *server, UA_SecurityGroup *securityGroup){
```
jpfr, line 75: Silent failures shouldn't happen.
If you cannot return a status code here, then use the logging from the server to warn users.
The goto error is only used once. That should be move inline to reduce the code flow complexity.

jpfr (commented):
Looks really good.
A few detail comments inline.
If you want us to preserve the commit history, then use the conventions from CONTRIBUTING.md.
Or see the examples in the master history.
Otherwise I would squash the PR into a single "feat(pubsub):" commit.

src/pubsub/ua_pubsub_securitygroup.c
```diff
@@ -0,0 +1,322 @@
...
+    if(keyStorage->keyListSize >= keyStorage->maxKeyListSize) {
+        /*the first item has the oldest key*/
+        UA_PubSubKeyListItem *oldestKey = TAILQ_FIRST(&keyStorage->keyList);
+        /*shift head to second oldest*/
+        TAILQ_INSERT_HEAD(&keyStorage->keyList, oldestKey->keyListEntry.tqe_next, keyListEntry);
+        /*move oldest to the tail*/
+        TAILQ_INSERT_TAIL(&keyStorage->keyList, oldestKey, keyListEntry);
+        /*copy new key over the oldest key locatio*/
+        oldestKey->keyID = newKeyID;
+        UA_ByteString_copy(&newKey, &oldestKey->key);
```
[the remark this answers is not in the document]
Muddasir-Shakil: Each security Group on SKS is configured with MaxFutureKeyCount and MaxPastKeyCount to define the maximum number of keys SKS can store with additional CurrentKey. So the oldest key is the maximum key we can store in the keystorage. So when we reach the maximum key size, we reuse the oldest key pointer for the new key and move it to the tail of the list.

src/pubsub/ua_pubsub_securitygroup.c
Muddasir-Shakil: I anyhow have to check both cases because if list is empty then TAILQ requires INSERT_HEAD and in rest of the cases INSERT_TAIL is used.
Also I checkout other places where TAILQ_INIT is used such as shown in the following. TAILQ is used in this manner. So I simply used as it was used for other cases like connections
https://github.com/open62541/open62541/blob/779e6b7902ff223b27d402c952fc5f06e967dd23/src/pubsub/ua_pubsub_manager.c#L78

also if calling TAILQ_INIT in the _init of the PubSubManager, would result in initializing securitygroups list every time whether the Security Groups will be used or not. Therefore I did the initialization when the first SecurityGroup is added to the server only then the list should be initialized.

src/pubsub/ua_pubsub_securitygroup.c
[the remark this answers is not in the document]
Muddasir-Shakil: Thanks for pointing out, I found a possible memleak. So after copying the new key data on the oldest key, I simply frees the pointer used to store the newkey data. This way we can avoid mem leak. In second case we anyhow use the newKey pointer in the keyStorage list.

src/pubsub/ua_pubsub_securitygroup.c
jpfr: >  if list is empty then TAILQ requires INSERT_HEAD and in rest of the cases INSERT_TAIL is used.

What gives you that impression?
Reading the code in queue.h, INSERT_TAIL always works once the structure is correctly initialized.

Muddasir-Shakil: > > if list is empty then TAILQ requires INSERT_HEAD and in rest of the cases INSERT_TAIL is used.
> 
> What gives you that impression? Reading the code in queue.h, INSERT_TAIL always works once the structure is correctly initialized.

I tested it and now it works with INSERT_TAIL. However, at other places as mentioned above it is used as the same way I did. May be it should be improved at other places like `UA_PubSubManager_addConnection`, and  `UA_PubSubManager_addTopic`

Regarding INIT, It added to pubsub UA_PubSubManager_init `UA_PubSubManager_init` with `#UA_ENABLE_SKS`. and I think this can done for other PubSub related lists.?

Muddasir-Shakil:
hi @jpfr 
I updated the PR for thread safe adding of securityGroups, fix a memory leak, and passing securitygroup object. More importantly I fixed the commit messages according to the conventions
Looking forward to your feedback

src/pubsub/ua_pubsub_securitygroup.c
jpfr: Ok. let's improve the other places then as well.
But not in this PR.

PR #5402, opened 2022-10-24
feat(pubsub): add GetSecurityKeys method node callback

Muddasir-Shakil:
The GetSecurityKeys Method Node is implemented to support the sks pull model.
The GetSecurityKeys checks whether the channel is encrypted and the caller of the
method has access rights, calculates the TimeToNextKey time and returns the requested
keys. It supports the following uses cases for requesting keys.
  - request current and future key
  - request current key only
  - request past keys
  - request oldest keys

Note: Role Permissions are not supported by open62541. However, the access control
      plugin can be used to create one for the application. An example of such is shown in unit tests for
      GetSecurityKeys method Node.


src/pubsub/ua_pubsub_ns0.c
```diff
@@ -1849,6 +1849,149 @@ setSecurityKeysAction(UA_Server *server, const UA_NodeId *sessionId, void *sessi
...
+     * the SKS shall return the maximum it allows.*/
+    if(requestedKeyCount > sg->config.maxFutureKeyCount)
+        requestedKeyCount =(UA_UInt32) sg->keyStorage->keyListSize;
+    else
+        requestedKeyCount = requestedKeyCount + currentKeyCount; /* Add Current keyCount */
+
+    /*The current token is requested by passing 0.*/
+    UA_PubSubKeyListItem *startingItem = NULL;
+    if(startingTokenId == 0) {
+        startingItem = sg->keyStorage->currentItem;
```
jpfr, line 1924: Can the currentItem be NULL?
If not, add an assert and document why this can never be the case.

src/pubsub/ua_pubsub_ns0.c
```diff
@@ -1849,6 +1849,149 @@ setSecurityKeysAction(UA_Server *server, const UA_NodeId *sessionId, void *sessi
...
+            if(session_entry->session.header.channel->securityMode !=
+               UA_MESSAGESECURITYMODE_SIGNANDENCRYPT)
+                return UA_STATUSCODE_BADSECURITYMODEINSUFFICIENT;
+        }
+    }
+
+    if(!server || !input)
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+
+    if(inputSize < 3)
```
jpfr, line 1871: Check the outputsize as well.
This is set in the information model.
A badly configured AccessControl policy might allow a user to modify the output size.
Just to make sure...

src/pubsub/ua_pubsub_ns0.c
```diff
@@ -1849,6 +1849,149 @@ setSecurityKeysAction(UA_Server *server, const UA_NodeId *sessionId, void *sessi
...
+        return retval;
+
+    /*KeyLifeTime*/
+    retval = UA_Variant_setScalarCopy(&output[4], &sg->config.keyLifeTime,
+                         &UA_TYPES[UA_TYPES_DURATION]);
+    if(retval != UA_STATUSCODE_GOOD)
+        return retval;
+
+    /*Keys*/
+    UA_PubSubKeyListItem *iterator = startingItem;
```
jpfr, line 1965: Creating the array and then freeing it, that's a bit too much overhead.
What you can do instead is calloc the array directly into output[2].
Just have a look at UA_Variant_setArray how to set up an array variant.
And then copy over piecewise the elements.
If an element copy fails you can immediately drop out.
And output[2] will be cleaned up by the stack after sending the response.
The not-yet-set elements will be zeroed (initialized) already.

jpfr (commented):
Looks good!
Just a few detail comments.

Muddasir-Shakil:
hi @jpfr 
Thanks for the reviews, I updated the commits with the fixes you proposed. One major change is to remove the additional array while copying requested keys onto output.
looking forward to your feedback
BR
Muddasir Shakil

jpfr:
Looks great 👍 
Merged.

PR #5407, opened 2022-10-28
[Review] feat(pubsub): add SKS Client support and SKS examples

Muddasir-Shakil:
An SKSClient is required by the Publisher and Subscriber applications to fetch the security Keys from the SKS server. The client is required to fetch the keys at regular interval. The next period to call the GetSecurityKeys Method is calculated from KeyLifeTime when the KeyStorage switches to the LastKey in the existing list. Then a timed callback is added to the server, which uses the sks client to call the GetSecurityKeys Method. The returned results are stored in the KeyStorage.

When the sks client has fetched the initial set of keys, in UA_Server_setSKSClient() function, The Reader/WriterGroup can be set to operational. The call to setOperational function will set the encrypting key from keystorage->currentItem to the channel context of the Writer/ReaderGroup and then set it into operational state.

### DIFF
--- a/include/open62541/client.h
+++ b/include/open62541/client.h
@@ -198,6 +198,12 @@ typedef struct {
 UA_EXPORT UA_StatusCode
 UA_ClientConfig_copy(UA_ClientConfig const *src, UA_ClientConfig *dst);
 
+/**
+ * @brief It cleans the client config and frees the pointer.
+ */
+UA_EXPORT void
+UA_ClientConfig_delete(UA_ClientConfig *config);
+
  /**
  * Client Lifecycle
  * ---------------- */
--- a/include/open62541/client.h
+++ b/include/open62541/client.h
@@ -204,6 +204,12 @@ UA_ClientConfig_copy(UA_ClientConfig const *src, UA_ClientConfig *dst);
 UA_EXPORT void
 UA_ClientConfig_delete(UA_ClientConfig *config);
 
+/**
+ * @brief It cleans the client config and deletes the plugins, whereas
+ * _copy makes a shallow copy of the plugins.
+ */
+UA_EXPORT void
+UA_ClientConfig_clear(UA_ClientConfig *config);
  /**
  * Client Lifecycle
  * ---------------- */
--- a/include/open62541/client.h
+++ b/include/open62541/client.h
@@ -13,6 +13,7 @@
  *    Copyright 2018 (c) Thomas Stalder, Blue Time Concept SA
  *    Copyright 2018 (c) Kalycito Infotech Private Limited
  *    Copyright 2020 (c) Christian von Arnim, ISW University of Stuttgart
+ *    Copyright 2022 (c) Linutronix GmbH (Author: Muddasir Shakil)
  */
 
 #ifndef UA_CLIENT_H_
@@ -185,6 +186,17 @@ typedef struct {
     UA_LocaleId *sessionLocaleIds;
     size_t sessionLocaleIdsSize;
 } UA_ClientConfig;
+
+/**
+ * @brief It makes a partial deep copy of the clientconfig. It makes a shallow
+ * copies of the plugins (logger, eventloop, securitypolicy).
+ *
+ * NOTE: It makes a shallow copy of all the plugins from source to destination.
+ * Therefore calling _clear on the dst object will also delete the plugins in src
+ * object.
+ */
+UA_EXPORT UA_StatusCode
+UA_ClientConfig_copy(UA_ClientConfig const *src, UA_ClientConfig *dst);
 
  /**
  * Client Lifecycle
--- a/include/open62541/server_pubsub.h
+++ b/include/open62541/server_pubsub.h
@@ -13,6 +13,8 @@
 #define UA_SERVER_PUBSUB_H
 
 #include <open62541/common.h>
+#include <open62541/util.h>
+#include <open62541/client.h>
 #include <open62541/plugin/pubsub.h>
 #include <open62541/plugin/securitypolicy.h>
 
@@ -935,6 +937,51 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
  */
 UA_EXPORT UA_StatusCode UA_THREADSAFE
 UA_Server_removeSecurityGroup(UA_Server *server, const UA_NodeId securityGroup);
+
+/**
+ * @brief This is a repeated callback which is triggered on each iteration of SKS Pull request.
+ * The server uses this callback to notify user about the status of current Pull request iteration.
+ * The period is calculated based on the KeylifeTime of specified in the SecurityGroup object node on
+ * the SKS server.
+ *
+ * @param server The server instance managing the publisher/subscriber.
+ * @param sksPullRequestStatus The current status of sks pull request.
+ * @param context The pointer to user defined data passed to this callback.
+ */
+typedef void
+(*UA_Server_sksPullRequestCallback)(UA_Server *server, UA_StatusCode sksPullRequestStatus, void* context);
+
+/**
+ * @brief Sets the SKS client config used to call the GetSecurityKeys Method on SKS and get the
+ * initial set of keys for a SecurityGroupId and adds timedCallback for the next GetSecurityKeys
+ * method Call. This uses async Client API for SKS Pull request. The SKS Client instance is created and destroyed at
+ * runtime on each iteration of SKS Pull request by the server. The key Rollover mechanism will check if the new
+ * keys are needed then it will call the getSecurityKeys Method on SKS Server. At the end of SKS Pull request
+ * iteration, the sks client will be deleted by a delayed callback (in next server iteration).
+ *
+ * @note It is be called before setting Reader/Writer Group into Operational because this also allocates
+ * a channel context for the pubsub security policy.
+ *
+ * @note the stateCallback of sksClientConfig will be overwritten by an internal callback.
+ *
+ * @param server the server instance
+ * @param clientConfig holds the required configuration to make encrypted connection with
+ * SKS Server. The input client config takes the lifecycle as long as SKS request are made.
+ * It is deleted with its plugins when the server is deleted or the last Reader/Writer
+ * Group of the securityGroupId is deleted. The input config is copied to an internal
+ * config object and the content of input config object will be reset to zero.
+ * @param endpointUrl holds the endpointUrl of the SKS server
+ * @param securityGroupId the SecurityGroupId of the securityGroup on SKS and
+ * reader/writergroups
+ * @param callback the user defined callback to notify the user about the status of SKS
+ * Pull request.
+ * @param context passed to the callback function
+ * @return UA_StatusCode the retuned status
+ */
+UA_StatusCode UA_EXPORT
+UA_Server_setSksClient(UA_Server *server, UA_String securityGroupId,
+                       UA_ClientConfig *clientConfig, const char *endpointUrl,
+                       UA_Server_sksPullRequestCallback callback, void *context);
 
 #endif /* UA_ENABLE_PUBSUB_SKS */
 
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -194,6 +194,13 @@ UA_ClientConfig_clear(UA_ClientConfig *config) {
     UA_cleanupDataTypeWithCustom(config->customDataTypes);
 }
 
+void
+UA_ClientConfig_delete(UA_ClientConfig *config){
+    UA_assert(config != NULL);
+    UA_ClientConfig_clear(config);
+    UA_free(config);
+}
+
 static void
 UA_Client_clear(UA_Client *client) {
     /* Delete the async service calls with BADHSUTDOWN */
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -17,6 +17,7 @@
  *    Copyright 2018 (c) Kalycito Infotech Private Limited
  *    Copyright 2020 (c) Christian von Arnim, ISW University of Stuttgart
  *    Copyright 2021 (c) Fraunhofer IOSB (Author: Jan Hermes)
+ *    Copyright 2022 (c) Linutronix GmbH (Author: Muddasir Shakil)
  */
 
 #include <open62541/transport_generated.h>
@@ -30,6 +31,75 @@ clientHouseKeeping(UA_Client *client, void *_);
 /********************/
 /* Client Lifecycle */
 /********************/
+
+UA_StatusCode
+UA_ClientConfig_copy(UA_ClientConfig const *src, UA_ClientConfig *dst){
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+
+    retval = UA_ApplicationDescription_copy(&src->clientDescription, &dst->clientDescription);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
+
+    retval = UA_ExtensionObject_copy(&src->userIdentityToken, &dst->userIdentityToken);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
+
+    retval = UA_String_copy(&src->securityPolicyUri, &dst->securityPolicyUri);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
+
+    retval = UA_EndpointDescription_copy(&src->endpoint, &dst->endpoint);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
+
+    retval = UA_UserTokenPolicy_copy(&src->userTokenPolicy, &dst->userTokenPolicy);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
+
+    retval = UA_Array_copy(src->sessionLocaleIds, src->sessionLocaleIdsSize,
+                           (void **)&dst->sessionLocaleIds, &UA_TYPES[UA_TYPES_LOCALEID]);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
+
+    dst->sessionLocaleIdsSize = src->sessionLocaleIdsSize;
+    dst->connectivityCheckInterval = src->connectivityCheckInterval;
+    dst->certificateVerification = src->certificateVerification;
+    dst->clientContext = src->clientContext;
+    dst->customDataTypes = src->customDataTypes;
+    dst->eventLoop = src->eventLoop;
+    dst->externalEventLoop = src->externalEventLoop;
+    dst->inactivityCallback = src->inactivityCallback;
+    dst->localConnectionConfig = src->localConnectionConfig;
+    dst->logger = src->logger;
+#ifdef UA_ENABLE_SUBSCRIPTIONS
+    dst->outStandingPublishRequests = src->outStandingPublishRequests;
+#endif
+    dst->requestedSessionTimeout = src->requestedSessionTimeout;
+    dst->secureChannelLifeTime = src->secureChannelLifeTime;
+    dst->securityMode = src->securityMode;
+    dst->stateCallback = src->stateCallback;
+#ifdef UA_ENABLE_SUBSCRIPTIONS
+    dst->subscriptionInactivityCallback = src->subscriptionInactivityCallback;
+#endif
+    dst->timeout = src->timeout;
+    dst->userTokenPolicy = src->userTokenPolicy;
+    dst->securityPolicies = src->securityPolicies;
+    dst->securityPoliciesSize = src->securityPoliciesSize;
+    dst->authSecurityPolicies = src->authSecurityPolicies;
+    dst->authSecurityPoliciesSize = src->authSecurityPoliciesSize;
+
+cleanup:
+    if(retval != UA_STATUSCODE_GOOD) {
+        /* _clear will free the plugins in dst that are a shallow copy from src. */
+        dst->authSecurityPolicies = NULL;
+        dst->certificateVerification.context = NULL;
+        dst->eventLoop = NULL;
+        dst->logger.context = NULL;
+        dst->securityPolicies = NULL;
+        UA_ClientConfig_clear(dst);
+    }
+    return retval;
+}
 
 UA_Client *
 UA_Client_newWithConfig(const UA_ClientConfig *config) {
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -133,7 +133,7 @@ UA_Client_newWithConfig(const UA_ClientConfig *config) {
     return client;
 }
 
-static void
+void
 UA_ClientConfig_clear(UA_ClientConfig *config) {
     UA_ApplicationDescription_clear(&config->clientDescription);
 
--- a/src/pubsub/ua_pubsub_keystorage.c
+++ b/src/pubsub/ua_pubsub_keystorage.c
@@ -354,6 +354,22 @@ UA_PubSubKeyStorage_activateKeyToChannelContext(UA_Server *server, UA_NodeId pub
     return retval;
 }
 
+static void
+nextGetSecuritykeysCallback(UA_Server *server, UA_PubSubKeyStorage *keyStorage) {
+    UA_StatusCode retval = UA_STATUSCODE_BAD;
+    if(!keyStorage) {
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                     "GetSecurityKeysCall Failed with error: KeyStorage does not exist "
+                     "in the server");
+        return;
+    }
+    retval = getSecurityKeysAndStoreFetchedKeys(server, keyStorage);
+    if(retval != UA_STATUSCODE_GOOD)
+        UA_LOG_ERROR(&server->config.logger, UA_LOGCATEGORY_SERVER,
+                     "GetSecurityKeysCall Failed with error: %s ",
+                     UA_StatusCode_name(retval));
+}
+
 void
 UA_PubSubKeyStorage_keyRolloverCallback(UA_Server *server, UA_PubSubKeyStorage *keyStorage) {
     /* Callbacks from the EventLoop are initially unlocked */
@@ -380,6 +396,15 @@ UA_PubSubKeyStorage_keyRolloverCallback(UA_Server *server, UA_PubSubKeyStorage *
                          (int)keyStorage->securityGroupID.length, keyStorage->securityGroupID.data,
                          UA_StatusCode_name(retval));
         }
+    } else if(keyStorage->sksConfig.endpointUrl) {
+        UA_DateTime now = UA_DateTime_nowMonotonic();
+        /*Publishers using a central SKS shall call GetSecurityKeys at a period of half the KeyLifetime */
+        UA_Duration msTimeToNextGetSecurityKeys = keyStorage->keyLifeTime / 2;
+        UA_DateTime dateTimeToNextGetSecurityKeys =
+            now + (UA_DateTime)(UA_DATETIME_MSEC * msTimeToNextGetSecurityKeys);
+        retval = server->config.eventLoop->addTimedCallback(
+            server->config.eventLoop, (UA_Callback)nextGetSecuritykeysCallback, server,
+            keyStorage, dateTimeToNextGetSecurityKeys, NULL);
     }
     UA_UNLOCK(&server->serviceMutex);
 }
--- a/src/pubsub/ua_pubsub_keystorage.c
+++ b/src/pubsub/ua_pubsub_keystorage.c
@@ -10,7 +10,10 @@
 
 #ifdef UA_ENABLE_PUBSUB_SKS /* conditional compilation */
 
+#define UA_REQ_CURRENT_TOKEN 0
+
 #include "server/ua_server_internal.h"
+#include "client/ua_client_internal.h"
 
 UA_PubSubKeyStorage *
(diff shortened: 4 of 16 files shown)


jpfr (commented):
As usual, good work.

I have a wish for the API though.
The current approach requires that the client is set up and connected outside the server.
Furthermore the keys are only retrieved once.
Ideally the SKS client can be spun up automatically whenever the keys are running out.

The following API would greatly reduce the burden of the operator to babysit the the SKS client:
Only set a client configuration for the SKS.
The client is created at runtime and its lifecycle is fully controlled by the server.
By setting the same EventLoop for server and client, both of them can nicely interact.
But only the async client API must be used for connecting and method calling.
Otherwise the entire server blocks.
A new client can then be created whenever the keys are running out in long-term operations.

Muddasir-Shakil:
> As usual, good work.
> 

Thank you @jpfr  for the review

> The following API would greatly reduce the burden of the operator to babysit the the SKS client: Only set a client configuration for the SKS. The client is created at runtime and its lifecycle is fully controlled by the server.
By setting the same EventLoop for server and client, both of them can nicely interact. But only the async client API must be used for connecting and method calling. Otherwise the entire server blocks. A new client can then be created whenever the keys are running out in long-term operations.

I think it is possible in this way that I add an SKS config object to the PubSubkeyStorage 
` UA_SkSConfig {
        UA_ClientConfig,
        SKSEndpointUrl
    }`

Update `UA_Server_setSksClient` to now take the `UA_ClientConfig` instead of client object and `SKS server endpointUrl`. Using this information Server can create Client and connect to the sks server and fetch the keys and afterwards it can delete it. However, the user identity token must be set in the client config before passing to this API method. I think there is no other API method to set user identity token on the client config other then `UA_Client_connectUsername`. 

The keys are fetched first time and then periodically with this commit [7f64fe4](https://github.com/open62541/open62541/pull/5407/commits/7f64fe4bdb1a9ab3b8322330d1688d59ca6a02bd)

I am also in agreement in using async API for client connect and method call. 

Looking forward to your reply.
BR

jpfr:
Sounds great.
Looking forward to the feature.

Muddasir-Shakil:
@jpfr 
I have a question regarding the eventloop, do Client and Server share the same eventloop or they have separate eventloops? 
When both are started from same application for a Client-Server, I see that there has to be separate calls for run_iterate for client and server both, is this correct? 
if not how is it possible to register the client instances to use the same eventloop of the server. So that only server_run_iterate would control the lifecycle of Client-Server application. May be it is an stupid idea. 
When I am moving to async API for the client, I have to use client_run_iterate in the internal API to connect, and call service, until the connection is established and call returns with results. I am not sure if that blocks the Server from any execution like accepting other connections and so on.

Muddasir-Shakil:
Hi @jpfr ,
I have updated the PR as per your comments. The major change is now I move all the SKS client calls to Async Client APIs and the SKS client API now takes the sks client config and SKS endpointUrl. The SKS client Config has all the information required to connect SKS Server.
Looking forward to your feedback.
BR
Muddasir Shakil

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -468,4 +501,195 @@ UA_PubSubKeyStorage_removeKeyStorage(UA_Server *server, UA_PubSubKeyStorage *key
...
+
+UA_StatusCode
+getSecurityKeysAndStoreFetchedKeys(UA_Server *server, UA_PubSubKeyStorage *keyStorage) {
+    UA_StatusCode retval = UA_STATUSCODE_BAD;
+    UA_UInt32 startingTokenId = UA_REQ_CURRENT_TOKEN;
+    UA_UInt32 requestKeyCount = UA_UINT32_MAX;
+
+    UA_ClientConfig *cc = keyStorage->sksConfig.sksClientConfig;
+    cc->stateCallback = onConnect;
+    UA_Client *client = UA_Client_newWithConfig(cc);
```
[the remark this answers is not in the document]
Muddasir-Shakil: > This moves the config into the client. As such the config you give this will be cleaned up automatically. That should be documented.
> 

ACK 
> Or we need some UA_ClientConfig_copy. There already exists UA_ClientConfig_clear which can also be exposed as public API together with _copy.

I was also thinking along the same lines. However, I would like to keep the scope of the PR clear. I will create a separate PR for these APIs and use them  here.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -468,4 +501,195 @@ UA_PubSubKeyStorage_removeKeyStorage(UA_Server *server, UA_PubSubKeyStorage *key
...
+    if(!client)
+         return retval;
+
+    /* connect to sks server */
+    retval = UA_Client_connectAsync(client, keyStorage->sksConfig.sksEndpointUrl);
```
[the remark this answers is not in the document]
Muddasir-Shakil: > Before you do that you should set the EventLoop in the client config to the server eventloop.
> It can be shared between the two
So when client_run_iterate is called here, it will also process server requests as well. right?

include/open62541/server_pubsub.h
```diff
@@ -923,6 +924,23 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
...
+ * a channel context for the pubsub security policy.
+ *
+ * @param server the server instance
+ * @param sksClientConfig holds the required configuration to make encrypted connection with SKS Server
+ * @param sksEndpointUrl holds the endpointUrl of the SKS server
+ * @param securityGroupId the SecurityGroupId of the securityGroup on SKS and reader/writergroups
+ * @return UA_StatusCode the retuned status
+ */
+UA_StatusCode UA_EXPORT
+UA_Server_setSksClient(UA_Server *server, UA_ClientConfig *sksClientConfig, const char * sksEndpointUrl, UA_String securityGroupId);
```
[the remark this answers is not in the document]
Muddasir-Shakil: ACK

Muddasir-Shakil:
Thanks for the review @jpfr ,
I have updated the PR with the changes requested. I have now implemented the support for shared eventloop for Server and sks client. Another thing I had to add was a user callback to notify user about the status of sks pull request. In the examples, I have shown one use case of it that when pull request completes successfully, then writer group can be set into operational.

However, this opens up a new issue about having internal API for a shared eventloop. which can also be used for LDS Client in discovery system when registering the server. The periodic call of register server is still sync and I think it creates a new eventloop for registering the server. I think we should also think about shared eventloop API as well now.

include/open62541/server_pubsub.h
```diff
@@ -923,6 +924,36 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
 UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Server_removeSecurityGroup(UA_Server *server, const UA_NodeId securityGroup);
 
+/**
+ * @brief This user defined callback to notify the user about the current status of the periodic SKS Pull request.
+ * @param server instance used to make sks pull request
+ * @param sksClientStatus
+ */
+typedef void
+(*UA_Server_sksPullRequestCallback)(UA_Server *server, UA_StatusCode sksPullRequestStatus, UA_Boolean sksPullSuccess, void* userData);
```
jpfr, line 933: The success argument is not described in the comments.
Would it be enough to have status==good the same as success=true?
Then we don’t need that argument.

include/open62541/server_pubsub.h
```diff
@@ -923,6 +924,36 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
...
+
+/**
+ * @brief Sets the SKS client config used to call the GetSecurityKeys Method on SKS and get the
```
jpfr, line 936: Document if this is just for a one-time SKS pull request or whether it will repeat automatically once a new key is required.

Is there a way to remove or modify the SKS client once set?

if there is just one SKS client per security group, then we could just use the security group ID to remove.

jpfr (commented):
This getting really good.

yes, we can deal with the LDS in a similar way afterwards.

include/open62541/server_pubsub.h
Muddasir-Shakil: > Is there a way to remove or modify the SKS client once set?

the sks client is an internal instance user will never has access to it. However, the user can update the client config using `setSksClient` API for the next iteration of key fetch. The key rollover mechanism will check if new keys are needed then it will call `getSecurityKeysAndStoreFetchedKeys` method, which will take care of the new client lifecycle. After each SKS Pull request iteration the sks client will be deleted by a delayed callback `sksClientCleanupCb`.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -468,4 +501,273 @@ UA_PubSubKeyStorage_removeKeyStorage(UA_Server *server, UA_PubSubKeyStorage *key
...
+         /* Delete the async service calls with BADHSHUTDOWN */
+         __Client_AsyncService_removeAll(client, UA_STATUSCODE_BADSHUTDOWN);
+         /* Remove the internal regular callback */
+         UA_Client_removeCallback(client, client->houseKeepingCallbackId);
+
+         client->houseKeepingCallbackId = 0;
+
+         UA_free(client->config.clientContext);
+         /* the config will be reused for next call */
+         UA_free(client);
```
jpfr, line 741: Use UA_Client_delete.
This might work now. But in the future the client may carry additional dynamically allocated members.
So UA_Client_delete is the only "official" way of cleaning up and deleting the client.

But that also cleans up the client config.

Add UA_ClientConfig_copy (and taking a const config pointer for the SKS config) should be straightforward though.

include/open62541/server_pubsub.h
jpfr: Sounds good. Please add what you described to the public API documentation.
1-2 sentences ought to be good enough.

include/open62541/server_pubsub.h
```diff
@@ -925,6 +927,41 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
...
+ * @param server the server instance
+ * @param sksClientConfig holds the required configuration to make encrypted connection with SKS Server
+ * @param sksEndpointUrl holds the endpointUrl of the SKS server
+ * @param securityGroupId the SecurityGroupId of the securityGroup on SKS and reader/writergroups
+ * @param callback the user defined callback to notify the user about the status of SKS Pull request.
+ * @param userData passed to the callback function
+ * @return UA_StatusCode the retuned status
+ */
+UA_StatusCode UA_EXPORT
+UA_Server_setSksClient(UA_Server *server, UA_ClientConfig *sksClientConfig, const char * sksEndpointUrl, UA_String securityGroupId,
```
jpfr, line 962: Please move the "securityGroupId" to be the second argument.
We typically first indicate the context or scope of the operation (here server->securitygroup) and then the arguments of the operation.

jpfr (commented):
A very good PR.
Comments inline.

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: UA_ClientConfig_copy should do a deep copy otherwise, delete client removes references to logger, eventloop and other pointers.  It requires more works then I estimated, shallow copy will not work.

jpfr: The only problematic part is the handling of security policies.
Those are stateful when added to the configuration.

you could special-case them and reset those fields to NULL before cleaning up.

a generic configuration copy would be even better of course.

Muddasir-Shakil: > a generic configuration copy would be even better of course.

I agree. A suggestion I would like to make is to add "copy" function pointer to the all the plugin interface for security policy, eventloop etc. They all have void pointers which are casted to internal structs and make it hard to deep copy onto other location. We already have clear function pointer, so having copy will help.

Muddasir-Shakil:
Hi @jpfr ,
Thank you for making the keystorage more thread safe. I have also rebased this PR on your changes and fixed all the issues raised after those fixes. I tried to made it thread safe as mush as possible and I think it is ready for another review.
Looking forward to your review.
BR
Muddasir Shakil

include/open62541/server_pubsub.h
```diff
@@ -931,6 +933,44 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
...
+ * @param server the server instance
+ * @param sksClientConfig holds the required configuration to make encrypted connection with SKS Server
+ * @param sksEndpointUrl holds the endpointUrl of the SKS server
+ * @param securityGroupId the SecurityGroupId of the securityGroup on SKS and reader/writergroups
+ * @param callback the user defined callback to notify the user about the status of SKS Pull request.
+ * @param userData passed to the callback function
+ * @return UA_StatusCode the retuned status
+ */
+UA_StatusCode UA_EXPORT
+UA_Server_setSksClient(UA_Server *server, UA_String securityGroupId, UA_ClientConfig *sksClientConfig, const char * sksEndpointUrl,
```
jpfr, line 971: I would prefer to have a const client config.
So it is clear that we make a copy of the config internally.
So we have no coupling with the lifetime of the original config pointer.

also document that a copy of the config is made initially.

src/client/ua_client.c
```diff
@@ -31,6 +32,36 @@ clientHouseKeeping(UA_Client *client, void *_);
 /* Client Lifecycle */
 /********************/
 
+UA_StatusCode
+UA_ClientConfig_copy(UA_ClientConfig const *src, UA_ClientConfig *dst){
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
```
jpfr, line 37: Run clientconfig_clear when copying fails.
Attention! This might clear strings that are also in the original config.

Also make ClientConfig_clear public.
So we can work with stack-allocated configs.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -75,6 +78,13 @@ UA_PubSubKeyStorage_delete(UA_Server *server, UA_PubSubKeyStorage *keyStorage) {
     if(keyStorage->securityGroupID.data)
         UA_String_clear(&keyStorage->securityGroupID);
 
+    /* delete SKS client config from keystorage */
+    if(keyStorage->sksConfig.sksClientConfig) {
+        UA_ClientConfig_delete(keyStorage->sksConfig.sksClientConfig);
+    }
+
+    memset(keyStorage, 0, sizeof(UA_PubSubKeyStorage));
```
jpfr, line 86: No need to memset zero if the memory is freed right afterwards.
The paranoid can configure their malloc / free to overwrite memory once freed in the background.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -430,4 +465,268 @@ UA_PubSubKeyStorage_detachKeyStorage(UA_Server *server, UA_PubSubKeyStorage *key
...
+    UA_UInt32 startingTokenId;
+    UA_UInt32 requestedKeyCount;
+} sksClientContext;
+
+void
+sksClientCleanupCb(void *client, void *context);
+
+static void
+addDelayedSksClientCleanupCb(UA_Client *client){
+ UA_DelayedCallback *dc = (UA_DelayedCallback *)UA_calloc(1, sizeof(UA_DelayedCallback));
```
jpfr, line 484: Always check malloc results.
Often you can avoid malloc for delayed cleanups.
By adding the delayed callback structure to the object that shall be cleaned up.

we sometimes have embedded targets that run out of memory and still want to shut down gracefully.
So it’s good to require no memory on the cleanup path.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -430,4 +465,268 @@ UA_PubSubKeyStorage_detachKeyStorage(UA_Server *server, UA_PubSubKeyStorage *key
...
+                       UA_ClientConfig *sksClientConfig, const char *sksEndpointUrl,
+                       UA_Server_sksPullRequestCallback callback, void *userData) {
+    if(!server || !sksClientConfig || !sksEndpointUrl)
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+
+    UA_StatusCode retval = UA_STATUSCODE_BADNOTFOUND;
+    UA_LOCK(&server->serviceMutex);
+    UA_PubSubKeyStorage *ks = UA_PubSubKeyStorage_findKeyStorage(server, securityGroupId);
+    if(!ks)
+        return retval;
```
jpfr, line 717: Release the lock here.

src/pubsub/ua_pubsub_readergroup.c
```diff
@@ -490,9 +490,18 @@ UA_Server_setReaderGroupOperational(UA_Server *server, const UA_NodeId readerGro
...
     UA_StatusCode ret = UA_STATUSCODE_BADNOTFOUND;
     UA_ReaderGroup *rg = UA_ReaderGroup_findRGbyId(server, readerGroupId);
-    if(rg)
+    if(rg) {
+#ifdef UA_ENABLE_PUBSUB_SKS
+        if(rg->keyStorage && rg->keyStorage->currentItem) {
+            UA_StatusCode retval = UA_PubSubKeyStorage_activateKeyToChannelContext(
+                server, rg->identifier, rg->config.securityGroupId);
+            if(retval != UA_STATUSCODE_GOOD)
+            return retval;
```
jpfr, line 499: Release the lock.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -430,4 +465,268 @@ UA_PubSubKeyStorage_detachKeyStorage(UA_Server *server, UA_PubSubKeyStorage *key
...
+        UA_PubSubKeyStorage *ks = ctx->ks;
+        if(ks->sksConfig.sksPullRequestCallback)
+            ks->sksConfig.sksPullRequestCallback(ctx->server, connectStatus,
+                                            ks->sksConfig.userData);
+        addDelayedSksClientCleanupCb(client);
+    }
+    return;
+}
+
+static void
```
jpfr, line 638: Add t to eventloop

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -430,4 +465,268 @@ UA_PubSubKeyStorage_detachKeyStorage(UA_Server *server, UA_PubSubKeyStorage *key
...
+setExternalEvenloopOnSkSClient(UA_ClientConfig *cc, UA_EventLoop *externalEventloop) {
+    UA_assert(externalEventloop != NULL);
+
+    /* Stop and delete the client EventLoop */
+    UA_EventLoop *el = cc->eventLoop;
```
jpfr, line 643: Stopping the event loop can have unintended side effects.
As it could be configured for multiple applications at once.

Why is this needed? SKS should be able to initially set an already running EventLoop.
Switching the EventLoop mid-flight of the SKS opens a can of worms. I wouldn’t support that.

If the EventLoop has to be stopped, log an error and return.
But don’t stop the EventLoop on your own.
Developers will not be aware of this without having read the code.

src/pubsub/ua_pubsub_keystorage.h
```diff
@@ -112,6 +115,17 @@ typedef struct UA_PubSubKeyListItem {
 /* Queue Definition*/
 typedef TAILQ_HEAD(keyListItems, UA_PubSubKeyListItem) keyListItems;
 
+/**
+ * @brief It is used to hold configuration information required to connect an SKS server
+ * and fetch the security keys
+ */
+typedef struct UA_PubSubSKSConfig {
```
jpfr, line 122: No need to prefix the members with sks.
The structure is already prefixed.

I do prefer to have the clientconfig as a direct member, not a pointer.
Less moving parts. Less malloc error handling.

src/pubsub/ua_pubsub_securitygroup.c
```diff
@@ -108,7 +108,7 @@ updateSKSKeyStorage(UA_Server *server, UA_SecurityGroup *securityGroup){
         /* the first item has the oldest key */
         UA_PubSubKeyListItem *oldestKey = TAILQ_FIRST(&keyStorage->keyList);
         /* shift head to second oldest */
-        TAILQ_INSERT_HEAD(&keyStorage->keyList, oldestKey->keyListEntry.tqe_next, keyListEntry);
```
jpfr, line 111: You can just put an overall comment ‘move oldest to the tail’.
Some of the line-comments no longer apply.
I.e. shift head to second oldest.

jpfr (commented):
Good work!
A few last detail comments. Can be merged then.

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: > Stopping the event loop can have unintended side effects. As it could be configured for multiple applications at once.
> 
> Why is this needed? SKS should be able to initially set an already running EventLoop. Switching the EventLoop mid-flight of the SKS opens a can of worms. I wouldn’t support that.
> 
The eventloop is not switched in mid flight but immediately after creating the sks client instance. This allows sks client to use the server's eventloop and the SKS Client lifecycle is managed in UA_server_run_iterate calls. This is all happening in `getSecurityKeysAndStoreFetchedKeys`

> If the EventLoop has to be stopped, log an error and return. But don’t stop the EventLoop on your own. Developers will not be aware of this without having read the code.

Before stopping the eventloop there is a check for externalEventLoop flag. If it is set then eventloop is not stopped and simply replaced. But if the eventloop is not shared/external then stopping is important, otherwise the eventloop pointer will be lost and cannot be freed.

Muddasir-Shakil:
Hello @jpfr,
Thanks for the comments. I have updated the PR according to your comments. The major changes include 

- removed stopping and deleting of eventloop.
- made ClientConfig_clear public
- clear destination config on failure in copy method

Looking forward to your comments.
BR
Muddasir Shakil

include/open62541/client.h
```diff
@@ -186,6 +187,22 @@ typedef struct {
     size_t sessionLocaleIdsSize;
 } UA_ClientConfig;
 
+/**
+ * @brief It makes a partial deep copy of the clientconfig. It makes the shallow
```
jpfr, line 191: s/the shallow/a shallow

src/client/ua_client.c
```diff
@@ -31,6 +32,75 @@ clientHouseKeeping(UA_Client *client, void *_);
...
+        goto cleanup;
+
+    retval = UA_UserTokenPolicy_copy(&src->userTokenPolicy, &dst->userTokenPolicy);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
+
+    retval = UA_Array_copy(src->sessionLocaleIds, src->sessionLocaleIdsSize,
+                           (void **)&dst->sessionLocaleIds, &UA_TYPES[UA_TYPES_LOCALEID]);
+    if(retval != UA_STATUSCODE_GOOD)
+        goto cleanup;
```
jpfr, line 62: Set sessionLocaleIdsSize right after the retval test.

src/client/ua_client.c
jpfr: 👍

src/client/ua_client.c
```diff
@@ -31,6 +32,75 @@ clientHouseKeeping(UA_Client *client, void *_);
...
+
+    retval =
+        UA_UInt32_copy(&src->connectivityCheckInterval, &dst->connectivityCheckInterval);
```
jpfr, line 65: This is overkill.
If a member has no allocated memory, just to `dst->interval = src->interval`.

src/client/ua_client.c
```diff
@@ -31,6 +32,75 @@ clientHouseKeeping(UA_Client *client, void *_);
...
+    dst->requestedSessionTimeout = src->requestedSessionTimeout;
+    dst->secureChannelLifeTime = src->secureChannelLifeTime;
+    dst->securityMode = src->securityMode;
+    dst->stateCallback = src->stateCallback;
+#ifdef UA_ENABLE_SUBSCRIPTIONS
+    dst->subscriptionInactivityCallback = src->subscriptionInactivityCallback;
+#endif
+    dst->timeout = src->timeout;
+    dst->userTokenPolicy = src->userTokenPolicy;
+    if(src->securityPolicies) {
```
jpfr, line 89: Yout don't need the if-clause.
This works even better without.
As `dst->securityPolicies` might be uninitialized.

src/client/ua_client.c
```diff
@@ -31,6 +32,75 @@ clientHouseKeeping(UA_Client *client, void *_);
...
+        dst->securityPoliciesSize = src->securityPoliciesSize;
+    }
+    if(src->authSecurityPolicies){
+        dst->authSecurityPolicies = src->authSecurityPolicies;
+        dst->authSecurityPoliciesSize = src->authSecurityPoliciesSize;
+    }
+
+cleanup:
+    if(retval != UA_STATUSCODE_GOOD)
+        UA_ClientConfig_clear(dst);
```
jpfr, line 100: This is dangerous.
It will free the plugins in `dst` that are a shallow copy from `src`.
NULL out the shallow copies before _clear.

include/open62541/client.h
```diff
@@ -186,6 +187,22 @@ typedef struct {
...
+
+/**
+ * @brief It cleans the client config and frees the pointer.
+ */
+UA_EXPORT void
+UA_ClientConfig_delete(UA_ClientConfig *config);
+
+
+UA_EXPORT void
+UA_ClientConfig_clear(UA_ClientConfig *config);
```
jpfr, line 205: Document for _clear that it **will delete the plugins** whereas _copy makes a shallow copy of the plugins.

include/open62541/client.h
```diff
@@ -186,6 +187,22 @@ typedef struct {
     size_t sessionLocaleIdsSize;
 } UA_ClientConfig;
 
+/**
+ * @brief It makes a partial deep copy of the clientconfig. It makes the shallow
+ * copies of the plugin (logger, eventloop, securitypolicy).
```
jpfr, line 192: plugin**s** in plural.

include/open62541/server_pubsub.h
```diff
@@ -936,6 +938,45 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
...
+ * The server uses this callback to notify user about the status of current Pull request iteration.
+ * The period is calculated based on the KeylifeTime of specified in the SecurityGroup object node on
+ * the SKS server.
+ *
+ * @param server The server instance managing the publisher/subscriber.
+ * @param sksPullRequestStatus The current status of sks pull request.
+ * @param userData The pointer to user defined data passed to this callback.
+ */
+typedef void
+(*UA_Server_sksPullRequestCallback)(UA_Server *server, UA_StatusCode sksPullRequestStatus, void* userData);
```
jpfr, line 952: Rename `userData` to `context`.
Thus we are consistent with the rest of the SDK nomenclature.

include/open62541/server_pubsub.h
```diff
@@ -936,6 +938,45 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
...
+ * a channel context for the pubsub security policy.
+ *
+ * @note the stateCallback of sksClientConfig will be overwritten by an internal callback.
+ *
+ * @param server the server instance
+ * @param clientConfig holds the required configuration to make encrypted connection with SKS Server. A copy of the config is made internally.
+ * @param endpointUrl holds the endpointUrl of the SKS server
+ * @param securityGroupId the SecurityGroupId of the securityGroup on SKS and reader/writergroups
+ * @param callback the user defined callback to notify the user about the status of SKS Pull request.
+ * @param userData passed to the callback function
```
jpfr, line 972: userData -> context

include/open62541/server_pubsub.h
```diff
@@ -936,6 +938,45 @@ UA_Server_addSecurityGroup(UA_Server *server, UA_NodeId securityGroupFolderNodeI
...
+ * iteration, the sks client will be deleted by a delayed callback (in next server iteration).
+ *
+ * @note It is be called before setting Reader/Writer Group into Operational because this also allocates
+ * a channel context for the pubsub security policy.
+ *
+ * @note the stateCallback of sksClientConfig will be overwritten by an internal callback.
+ *
+ * @param server the server instance
+ * @param clientConfig holds the required configuration to make encrypted connection with SKS Server. A copy of the config is made internally.
+ * @param endpointUrl holds the endpointUrl of the SKS server
```
jpfr, line 969: More information needed for the client config.
If there are plugins defined in the input config (which is const), does the const-input-config have to "survive" as long the SKS requests are made or do we *move* the plugins to the internal config and also clean up the plugins eventually?

jpfr (commented):
We are getting there.
Much improved already.
A few comments remain

include/open62541/server_pubsub.h
Muddasir-Shakil: > If there are plugins defined in the input config (which is const), does the const-input-config have to "survive" as long the SKS requests are made

Since we cannot make the complete deep copy of the clientconfig due to plugins and the user can configure different plugins for the sks client config, It is safer to add the comment that 
**the const input config takes the lifecycle as long as SKS request are made. It is deleted with it plugins when the server is deleted or the last Reader/Writer Group of the SecurityGroup is deleted.**

jpfr:
Congratulations. Perfect PR, merged.
Thanks a lot for the good work and the sustained effort.

Muddasir-Shakil:
> Congratulations. Perfect PR, merged. Thanks a lot for the good work and the sustained effort.

Thank you for the great feedback.

PR #5410, opened 2022-10-31
fix(pubsub): fix generateDeltaFrameMessage issue

applea9:
don't clear dataSetMessage in UA_PubSubDataSetWriter_generateDeltaFrameMessage(), since some fields of dataSetMessage have been set in the UA_DataSetWriter_generateDataSetMessage().

### DIFF
--- a/src/pubsub/ua_pubsub_writer.c
+++ b/src/pubsub/ua_pubsub_writer.c
@@ -1191,6 +1191,8 @@ UA_PubSubDataSetWriter_generateKeyFrameMessage(UA_Server *server,
     return UA_STATUSCODE_GOOD;
 }
 
+/* the input message is already initialized and that the method 
+ * must not be called twice for the same message */
 #ifdef UA_ENABLE_PUBSUB_DELTAFRAMES
 static UA_StatusCode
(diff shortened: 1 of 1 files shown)


jpfr:
I find this highly unintuitive.
If we call `UA_PubSubDataSetWriter_generateDeltaFrameMessage` several times, there are counters that are never reset and just increased to infinity.

There should be a comment with a list of the fields that need to be updated and a list of things we want to explicitly retain. Otherwise we miss stuff. Or the next developer working on this misses stuff.

applea9:
> 
do you mean counter of dataSetMessage->data.deltaFrameData.fieldCount? 

But from the code I see `UA_PubSubDataSetWriter_generateDeltaFrameMessage` will only be called once in the `UA_DataSetWriter_generateDataSetMessage()`.

jpfr:
Ah good.
Then please add a comment that the input message is already initialized and that the method must not be called twice for the same message.
Can be merged then.

applea9:
> Ah good. Then please add a comment that the input message is already initialized and that the method must not be called twice for the same message. Can be merged then.

I have added the comments.

jpfr:
Nice, thanks.
I adjusted the commit title to conform to our CONTRIBUTING.md guideline.

PR #5412, opened 2022-10-31
Add an option to set a write or read context to UA_ValueCallback and UA_DataSource

ja4nm:
Example usage:

```
class EventManager{
public:

	WriteArgs
	{
	    const UA_NodeId* nodeId;
	    const UA_DataValue* value;
	};

	void setOnWrite(std::function<UA_StatusCode(WriteArgs args)>){
		UA_ValueCallback valueCallback;
	       valueCallback.onWrite = StaticOnWrite;
	       valueCallback.writeContext = this;
	       UA_Server_setVariableNode_valueCallback(server->getUaServer(), *nodeId, valueCallback);
	}

	static void StaticOnWrite(UA_Server* server,
                              const UA_NodeId* sessionId,
                              void* sessionContext,
                              const UA_NodeId* nodeId,
                              void* nodeContext,
                              const UA_NumericRange* range,
                              const UA_DataValue* value,
                              void* callbackContext)
	{
	    auto manager = (EventManager*) callbackContext;

	    WriteArgs args;
	    args.nodeId = nodeId;
	    args.value = value;

	    manager->writeCallback(args);
	}
}

int main(){
	int counter = 0;

	auto manager = EventManager(nodeId);
	manager.setOnWrite([counter](EventManager::WriteArgs args){
		std::cout << "We can now set callback as lamda functions with a capture clause." << std::endl;
		coutnetr++;
		return UA_STATUSCODE_GOOD;
	});
}
```

### DIFF
--- a/examples/discovery/server_register.c
+++ b/examples/discovery/server_register.c
@@ -30,7 +30,8 @@ static UA_StatusCode
 readInteger(UA_Server *server, const UA_NodeId *sessionId,
             void *sessionContext, const UA_NodeId *nodeId,
             void *nodeContext, UA_Boolean includeSourceTimeStamp,
-            const UA_NumericRange *range, UA_DataValue *value) {
+            const UA_NumericRange *range, UA_DataValue *value,
+            void *callbackContext) {
     UA_Int32 *myInteger = (UA_Int32*)nodeContext;
     value->hasValue = true;
     UA_Variant_setScalarCopy(&value->value, myInteger, &UA_TYPES[UA_TYPES_INT32]);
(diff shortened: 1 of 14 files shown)


jpfr:
Hey @ja4nm ,

I looked only at the public header file.
Once this looks good to us, the implementation easily follows.

Changing an existing API always requires a lot of thought.
Because there is a real downside to breaking existing code that uses the library.

There's something I don't get. Let me understand your requirement here.
Why do you need a separate node context and callback context?
If you have additional data that is relevant for the callback, you could integrate that in the node context and pull it out in the callback.

ja4nm:
The main issue for me was, that the nodeContext was always NULL. Both in the dataSourceCallback and the valueCallback.

This is how I have added a node:

```
UA_Server_addObjectNode(server->getUaServer(),
                            UA_NODEID_NULL,
                            *parentNodeId,
                            UA_NODEID_NUMERIC(0, UA_NS0ID_ORGANIZES),
                            *browseName,
                            *typeNodeId,
                            *attr,
                            contex,
                            &newNodeId);

    UA_Server_setNodeContext(server->getUaServer(), newNodeId, context);
```

_Why do you need a separate node context and callback context?_

I thought it was convenient. Different callbacks have different implementation and require different object. The same as each lamda function in c++ can have different variables in its capture clause. Also, I was afraid that someone will overwrite my node context needing it for something else.

In any case. We either have to fix a bug with nodeContext beegin NULL or we can keep the current implementation with multiple contexts (or we can do both). Please let me know, what you think is best.

PR #5416, opened 2022-11-02
fix(pubsub): Create a copy of SecurityGroupId string when copying WG and RG config

MichalTyndel:
We're doing deep copy of the config, so the string should be copied as well.


jpfr:
👍

PR #5417, opened 2022-11-02
refactor(pubsub): Removed unused field from DSR and RG config

MichalTyndel:
SecurityParameters member of reader and readerGroup config was not used anywhere and was a single instance of UA_PubSubSecurityParameters type. ReaderGroupConfig already contains this info in other memebers.


jpfr:
@Muddasir-Shakil ,

since you are working on the SKS in #5407 .
Okay to remove the config options as done in this PR?

PR #5418, opened 2022-11-02
fix(pubsub): Set correct block counter for aes128 encryption

MichalTyndel:
According to part 14 the initial value of AES128 encryption block counter should be set to 1, not 0.
obraz


jpfr:
Good find!
It appears the AES256 implementation suffers from the same problem.
Please add the fix for AES256 and make this a PR against the 1.3 branch.
We will then prepare a patch release to fix.

MichalTyndel:
> 

I only worked with aes128 and forgot about the others. In new version aes256 uses correct counter as well.

jpfr (approved):

jpfr:
Nice, thanks!

PR #5422, opened 2022-11-03
[REVIEW]feat(client): Enable clients to securely authenticate to a server using an X.509 certificate

NoelGraf:
- [x] Provide API function[s] to enable clients to securely authenticate to a server using an X.509 certificate as per OPC-UA specification. E.g. client side to use the “X509IdentityToken” to generate the “userTokenSignature” 
- [x] Documentation for the new API function[s] and its functionality
- [x] Example where a client authenticates itself to the server with a certificate.
- [x] Automated CI pipeline tests

Note:
There are currently some issues with the LibreSSL integration. For this reason, client authentication with certificates with LibreSSL is not supported for the time being.


include/open62541/client.h
```diff
@@ -280,6 +280,25 @@ UA_Client_connectUsername(UA_Client *client, const char *endpointUrl,
...
+/* Connect to the server and create+activate a Session with the given X.509 certificate.
+ * This first set the X509IdentityToken in the client config and
+ * then calls the regular connect method. */
+static UA_INLINE UA_StatusCode
+UA_Client_connectCertificate(UA_Client *client, const char *endpointUrl,
+                             const char *policyId, const char *certificateData) {
+    UA_X509IdentityToken* identityToken = UA_X509IdentityToken_new();
+    if(!identityToken)
+        return UA_STATUSCODE_BADOUTOFMEMORY;
+    identityToken->policyId = UA_STRING_ALLOC(policyId);
```
basyskom-meerkoetter, line 292: UA_STRING_ALLOC could also fail due to out of memory. does this also need to be checked?

include/open62541/client.h
```diff
@@ -28,6 +28,9 @@
 #include <open62541/plugin/network.h>
 #include <open62541/plugin/securitypolicy.h>
 #include <open62541/plugin/eventloop.h>
+#ifdef UA_ENABLE_ENCRYPTION
+#include <open62541/plugin/securitypolicy_default.h>
```
jpfr, line 32: Don't include a particular plugin implementation in the core library.

include/open62541/client.h
```diff
@@ -147,6 +150,9 @@ typedef struct {
     /* Certificate Verification Plugin */
     UA_CertificateVerification certificateVerification;
 
+    /* Security Policy for Authentication. It contains the given private key */
+    UA_SecurityPolicy *securityPolicyAuth;
```
jpfr, line 154: I prefer with a prefix `authSecurityPolicy`.
Security-Policy-Authentication sounds like authenticating a SecurityPolicy.

include/open62541/client.h
```diff
@@ -279,6 +285,14 @@ UA_Client_connectUsername(UA_Client *client, const char *endpointUrl,
     cc->userIdentityToken.content.decoded.data = identityToken;
     return UA_Client_connect(client, endpointUrl);
 }
+#ifdef UA_ENABLE_ENCRYPTION
+/* Connect to the server and create+activate a Session with the given X.509 certificate.
+ * This first set the X509IdentityToken in the client config and
+ * then calls the regular connect method. */
+UA_StatusCode UA_EXPORT
+UA_Client_connectCertificate(UA_Client *client, const char *endpointUrl,
```
jpfr, line 293: For logging in with username/pw we have two options:
1. Use UA_Client_connectUsername
2. Set up the userIdentityToken in the config and connect with UA_Client_connect

The same should possible for certificate authentication.
In the future we will have parsing of the config from some .ini file (JSON).

So I would prefer a public convenience function to set up the client config with cert and key.
And have UA_Client_connectCertificate first set up the config and then call into UA_Client_connect.

src/client/ua_client_connect.c
```diff
@@ -1372,6 +1421,62 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+    UA_ClientConfig *cc = UA_Client_getConfig(client);
+    cc->securityPolicyAuth = (UA_SecurityPolicy*) UA_malloc(sizeof(UA_SecurityPolicy));
+    UA_String aes128_Sha256_RsaOaep = UA_STRING("http://opcfoundation.org/UA/SecurityPolicy#Aes128_Sha256_RsaOaep");
+    UA_String basic128Rsa15  = UA_STRING("http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15");
+    UA_String basic256  = UA_STRING("http://opcfoundation.org/UA/SecurityPolicy#Basic256");
+    UA_String basic256Sha256 = UA_STRING("http://opcfoundation.org/UA/SecurityPolicy#Basic256Sha256");
+
+    UA_StatusCode retval = UA_STATUSCODE_BAD;
+    if(UA_String_equal(&cc->securityPolicyUri, &aes128_Sha256_RsaOaep)) {
+        retval =
```
jpfr, line 1439: Don't use any specific plugin in the core library.
The SecurityPolicies are already set up for the SecureChannels in the client config.
You can juse use the name to look up the correct policy.

src/server/ua_services_session.c
```diff
@@ -424,7 +424,14 @@ checkSignature(const UA_Server *server, const UA_SecurityPolicy *securityPolicy,
     if(!securityPolicy)
         return UA_STATUSCODE_BADINTERNALERROR;
 
-    const UA_ByteString *localCertificate = &securityPolicy->localCertificate;
+    const UA_ByteString *localCertificate;
+    if(userToken != NULL) {
+        UA_X509IdentityToken *userIdentityToken = (UA_X509IdentityToken*)
```
jpfr, line 429: Is it ensured that this cast is always safe?

jpfr (commented):
Looks good!
Comments inline.
A unit test would be nice as well.

include/open62541/client.h
```diff
@@ -280,6 +284,28 @@ UA_Client_connectUsername(UA_Client *client, const char *endpointUrl,
...
     return UA_Client_connect(client, endpointUrl);
 }
+#ifdef UA_ENABLE_ENCRYPTION
+/* Connect to the server and create+activate a Session with the given X.509 certificate.
+ * This first set the X509IdentityToken in the client config and
+ * then calls the regular connect method. */
+static UA_INLINE UA_StatusCode
+UA_Client_connectCertificate(UA_Client *client, const char *endpointUrl,
+                             UA_ByteString certificate, UA_ByteString privateKey) {
+    UA_X509IdentityToken* identityToken = UA_X509IdentityToken_new();
```
jpfr, line 294: `UA_Client_connectCertificate` is incomplete.
It does nothing with the private key.

Maybe we cannot make this work without a hard link to some specific plugin.
But we will not change the rule that the core library has no dependency on specific plugins.

In that case we could just document how the `authSecurityPolicies` field is to be set up before calling the normal _connect.
And reference `UA_ClientConfig_setCertAuthentication` in a comment.

include/open62541/client.h
```diff
@@ -147,6 +147,10 @@ typedef struct {
     /* Certificate Verification Plugin */
     UA_CertificateVerification certificateVerification;
 
+    /* Available SecurityPolicies for Authentication. */
+    size_t authSecurityPoliciesSize;
+    UA_SecurityPolicy *authSecurityPolicies;
```
jpfr, line 152: Please document when a certificate authentication is used.
If there is a (matching) authSecurityPolicy, do we automatically use that for authentication?
Or does that have to be configured in addition in the userTokenPolicy?

src/server/ua_services_session.c
```diff
@@ -415,16 +415,26 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
...
+               const UA_ExtensionObject *userToken, const UA_SignatureData *signature) {
     /* Check for zero signature length */
     if(signature->signature.length == 0)
         return UA_STATUSCODE_BADAPPLICATIONSIGNATUREINVALID;
 
     if(!securityPolicy)
         return UA_STATUSCODE_BADINTERNALERROR;
 
-    const UA_ByteString *localCertificate = &securityPolicy->localCertificate;
+    const UA_ByteString *localCertificate;
```
jpfr, line 427: It is easier to read if we don't break up declaration and assignment of the variable.

include/open62541/client.h
```diff
@@ -147,6 +147,10 @@ typedef struct {
     /* Certificate Verification Plugin */
     UA_CertificateVerification certificateVerification;
 
+    /* Available SecurityPolicies for Authentication. */
+    size_t authSecurityPoliciesSize;
```
jpfr, line 151: I think @mlgiraud said that the same certificates aught to be used for SecureChannel and authentication.
Please check with him.
If that is the case, then we don't need to configure separate authSecurityPoliciesSize.

jpfr (commented):
Nice, this is making progress.
Comments inline.

jpfr:
Nice!
The code is good.
There is a last issue for the LibreSSL build.
Can be merged once all is green.

src/server/ua_services_session.c
```diff
@@ -784,8 +795,8 @@ Service_ActivateSession(UA_Server *server, UA_SecureChannel *channel,
 
         /* Check the user token signature */
         response->responseHeader.serviceResult =
-            checkSignature(server, channel->securityPolicy, tempChannelContext,
-                           &session->serverNonce, &request->userTokenSignature);
+            checkSignature(server, securityPolicy, tempChannelContext,
```
mlgiraud, line 798: Rename securityPolicy to utpSecurityPolicy or something else, indicating that it is the securitypolicy associated with the usertoken.

src/client/ua_client_connect.c
```diff
@@ -75,14 +86,66 @@ signActivateSessionRequest(UA_Client *client, UA_SecureChannel *channel,
...
+    if(dataToSignSize > MAX_DATA_SIZE)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    UA_ByteString userTokenSign;
+    retval = UA_ByteString_allocBuffer(&userTokenSign, userTokenSignSize);
+    if(retval != UA_STATUSCODE_GOOD)
+        return retval; /* sd->signature is cleaned up with the response */
 
+    /* Create the userTokenSignature */
+    memcpy(userTokenSign.data, userIdentityToken->certificateData.data,
```
mlgiraud, line 139: The standard specifies: The data to sign is created by appending the last [serverNonce](https://reference.opcfoundation.org/search/17?t=serverNonce)to the [serverCertificate](https://reference.opcfoundation.org/search/17?t=serverCertificate)specified in the [CreateSession](https://reference.opcfoundation.org/search/17?t=CreateSession)response.

So this should be the remote certificate that is used as data to sign.

src/server/ua_services_session.c
```diff
@@ -415,16 +415,27 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
 
 static UA_StatusCode
 checkSignature(const UA_Server *server, const UA_SecurityPolicy *securityPolicy,
-               void *channelContext, const UA_ByteString *serverNonce,
-               const UA_SignatureData *signature) {
+               void *channelContext, const UA_ByteString *serverNonce,const UA_UserTokenPolicy *userTokenPolicy,
+               const UA_ExtensionObject *userToken, const UA_SignatureData *signature) {
```
mlgiraud, line 419: Remove these new parameters since they are not needed here.

src/server/ua_services_session.c
```diff
@@ -415,16 +415,27 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
...
+
+    if(userTokenPolicy != NULL && userToken != NULL) {
+        if(userTokenPolicy->tokenType == UA_USERTOKENTYPE_CERTIFICATE) {
+            UA_X509IdentityToken *userIdentityToken = (UA_X509IdentityToken *)
+                userToken->content.decoded.data;
+            /* Client certificate */
+            localCertificate = &userIdentityToken->certificateData;
+        }
+    }
+
```
mlgiraud, line 438: Remove. The data that is signed is always the server certificate plus the last server nonce.
See: https://reference.opcfoundation.org/Core/Part4/v104/docs/5.6.3
The data to sign is created by appending the last [serverNonce](https://reference.opcfoundation.org/search/17?t=serverNonce)to the [serverCertificate](https://reference.opcfoundation.org/search/17?t=serverCertificate)specified in the [CreateSession](https://reference.opcfoundation.org/search/17?t=CreateSession)response.

mlgiraud (changes requested):

PR #5425, opened 2022-11-05
Add parentNodeContext to the createOptionalChildCallback

ja4nm:


ja4nm:
Connected issue: https://github.com/open62541/open62541/issues/1733

jpfr:
The nodeid of the parent is passed.
You can use that to load the parents context from inside the callback.
So this breaking change doesn’t give us functionality  we don’t already have.

Please close if the issue is resolved.

ja4nm:
The issue can be resolved the way you have suggested.
However, all the other callbacks contain both the nodeId and context parameter, so I thought it would be nice to be consistent.

jpfr:
There are different node contexts to consider.
The parent, the new source node, etc.
If this was a new API, yes having the context is a good idea.
But since this API is already used in the wild, let's keep it stable if it is just a little effort to get the context manually.

Btw, `UA_NODESTORE_GET` needs to be matched with `UA_NODESTORE_RELEASE`.

It's good to see so many suggestions for the stack from you.

ja4nm:
Okay, than I am closing the PR. Thanks for your input.

PR #5430, opened 2022-11-07
[REVIEW] refactor(el): make sure eventloop starts correctly

JanSurft:
- only start if its not started yet (external el)
- start in the configuration state (ready to be used)


jpfr:
The change looks good.
But somehow fuzzing fails.

I think we might have to check whether `config->eventLoop == NULL`.
Maybe there are cases where fuzzing causes the config generation to abort.

PR #5437, opened 2022-11-10
feature(core): XML data encoding for numeric + string built-in types

keba-uso:
Initial PR for XML data encoding. Implemented encoding for the following data types:

- Boolean
- SByte
- Byte
- Int16
- UInt16
- Int32
- UInt32
- Int64
- UInt64
- Float
- Double
- String
- DateTime
- Guid
- NodeId
- ExpandedNodeId

Unsupported data types when trying to en-/decode shall throw an _UA_STATUSCODE_BADNOTIMPLEMENTED_ error. Also, included unit tests to validate functionality of both encoding and decoding of data values.

**NOTE**: Added TODO comments within decoding section, which note the restrictions of certain data types.


jpfr:
Looks great!
A few additional datatypes might be easy to add.
For example we already have UA_NodeId_parse, UA_ExpandedNodeId_Parse and UA_Guid_parse that take the same string-input also used for XML.
The DateTime parsing might be very similar to the JSON parsing.

jpfr:
And we also have the matching _print for these.

PR #5448, opened 2022-11-16
Respond with status code when no memory can be allocated to the response buffer

Kallelele:
Hi,
I think that it would be good to respond with an error when we fail to allocate memory in 
`UA_StatusCode retval = connection_recv(&e->connection, &buf, 0); ` to notify the client instead of not responding.

What do you think?

### DIFF
--- a/arch/network_tcp.c
+++ b/arch/network_tcp.c
@@ -579,7 +579,7 @@ ServerNetworkLayerTCP_listen(UA_ServerNetworkLayer *nl, UA_Server *server,
         } else if(retval == UA_STATUSCODE_BADOUTOFMEMORY) {
             UA_TcpErrorMessage error;
             
-            error.error = retval;
+            error.error = UA_STATUSCODE_BADTCPNOTENOUGHRESOURCES;
             error.reason = UA_STRING_NULL;
             UA_Connection_sendError(&e->connection, &error);
             (&e->connection)->close(&e->connection);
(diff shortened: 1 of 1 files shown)


jpfr:
Hey @Kallelele ,

I like the idea of sending diagnostics int he error case.
This patch however can *never* be ported to the master branch and only works for 1.3.
Because the networking part was revamped to a large extend.
We need something for the long-term future.

On master the server uses a TCP EventSource that is configured within its EventLoop.
The TCP EventSource also allocates the memory for networking.

By design, at any time there can be only one receiving and one sending operation.
We already have a reusable receive buffer (rxBuffer) in the TCP EventSource.
So sending never allocates new memory.
https://github.com/open62541/open62541/blob/0f208d36a721f7f60a8a6ebd6653000547f8cc88/arch/eventloop_posix_tcp.c#L33

If you additionaly add a transmit Buffer (txBuffer) that is returned from `TCP_allocNetworkBuffer`, then the entire problem of potentially running out of memory for networking has been completely eliminated.

The configuration options for the TCP EventSource are documented here:
https://github.com/open62541/open62541/blob/0f208d36a721f7f60a8a6ebd6653000547f8cc88/include/open62541/plugin/eventloop.h#L411

andreasebner:
Thanks for the proposal! Since the code paths in 1.4 and 1.5 have been completely rewritten, I’m going to close this PR.

PR #5449, opened 2022-11-16
fix(client): UA_Client_Subscriptions_create no longer adds subscripti…

dpazj:
…on to internal subscription list if the request fails


dpazj:
Resolves: https://github.com/open62541/open62541/issues/5450

jpfr:
Thanks. Good find.

PR #5456, opened 2022-11-17
PubSub State machine improvements

gauravks:


MurzynJ:
As it was mentioned by @andreasebner in the meeting. Latest Part 14 v. 1.05.02 has additional clarification of using pub sub states with regards to security keys. Since we now have in opne6541 the SKS methods implemented by @Muddasir-Shakil this becomes relevant.

image

Currently, the behaviour is that if you created and enabled a WriterGroup with enabled security, but you haven't pushed the keys yet, there is a segfault when it tries to publish data. When the state machine is fully implemented, we could recognize that the WriterGroup is not yet operational and skip publishing.

image
image

include/open62541/server_pubsub.h
```diff
@@ -282,6 +282,16 @@ UA_Server_getPubSubConnectionConfig(UA_Server *server,
 UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Server_removePubSubConnection(UA_Server *server, const UA_NodeId connection);
 
+UA_StatusCode
```
jpfr, line 285: I don't recognize the "enabled" nomenclature.
Does this come from the standard?

The PubSubState that underlies the state machine has five possible states:

```
    UA_PUBSUBSTATE_DISABLED = 0,
    UA_PUBSUBSTATE_PAUSED = 1,
    UA_PUBSUBSTATE_OPERATIONAL = 2,
    UA_PUBSUBSTATE_ERROR = 3,
    UA_PUBSUBSTATE_PREOPERATIONAL = 4,
```
I suggest a single method to interact with the state machine:
```c
UA_Server_setPubSubConnectionState(UA_Server *Server, const UA_NodeId connectionIdent
                  UA_PubSubState state, UA_StatusCode cause);
```
The same for Reader/Writer/ReaderGroup/WriterGroup.
This also very well maps to the internal methods we already have.
Such as 
```c
UA_StatusCode
UA_DataSetWriter_setPubSubState(UA_Server *server,
                                UA_DataSetWriter *dataSetWriter,
                                UA_PubSubState state,
                                UA_StatusCode cause);```

The old public API (like setOperational) can be kept for a little while as a lightweight wrapper.
Before eventually marking it as deprecated.

File: plugins/include/open62541/server_config_default.h [hunk 2]
```diff
@@ -78,7 +78,7 @@ UA_ServerConfig_setDefaultWithSecurityPolicies(UA_ServerConfig *conf,
  * certificate. */
 static UA_INLINE UA_StatusCode
 UA_ServerConfig_setDefault(UA_ServerConfig *config) {
-    return UA_ServerConfig_setMinimal(config, 4840, NULL);
+    return UA_ServerConfig_setMinimal(config, 4841, NULL);
```
jpfr, line 81: Don't change this.
It will break people's projects that rely on that port.

src/pubsub/ua_pubsub_connection.c
```diff
@@ -200,6 +200,11 @@ UA_PubSubConnection_create(UA_Server *server,
     if(newConnectionsField->channel->openSubscriber) {
         newConnectionsField->channel->openSubscriber(newConnectionsField->channel);
     }
+
+    if (connectionConfig->enabled)
+    {
+        UA_Server_enablePubSubConnection(server, *connectionIdentifier);
```
jpfr, line 206: This could lead to a dead-lock.
Methods starting with `UA_Server_` are in the public API and take a lock to protect the server being accessed from parallel threads. Once you have the lock you must not take it a second time.

This method looks like you are already within the server-lock-context.
So don't call methods starting with UA_Server_ from here.
Ideally add UA_LOCK_ASSERT at the beginning of the method to make the assumptions explicit.

src/pubsub/ua_pubsub_connection.c
```diff
@@ -298,4 +303,234 @@ UA_PubSubConnection_regist(UA_Server *server, UA_NodeId *connectionIdentifier,
...
+                                            UA_StatusCode cause) {
+    switch(connection->state) {
+        case UA_PUBSUBSTATE_DISABLED:
+        case UA_PUBSUBSTATE_PAUSED:
+            connection->state = UA_PUBSUBSTATE_PREOPERATIONAL;
+            return UA_STATUSCODE_GOOD;
+        case UA_PUBSUBSTATE_PREOPERATIONAL:
+            break;
+        case UA_PUBSUBSTATE_OPERATIONAL:
+            return UA_STATUSCODE_GOOD;
```
jpfr, line 379: If you go from operational to pre-operational, wouldn't you have to close the underlying connection?
(e.g. MQTT...)

src/pubsub/ua_pubsub_connection.c
```diff
@@ -298,4 +303,234 @@ UA_PubSubConnection_regist(UA_Server *server, UA_NodeId *connectionIdentifier,
...
+                stateChangeCallback(server, &connection->identifier, state, cause);
+        }
+    }
+    return ret;
+}
+
+UA_StatusCode
+UA_Server_enablePubSubConnection(UA_Server *server,
+                                    const UA_NodeId connectionIdent) {
+    //UA_LOCK(&server->serviceMutex);
```
jpfr, line 500: You need to take the locks coming from the public API.
Create an "inner version" of the methods that assumes that locks are already taken.
If you need to call it also from within other server methods.

jpfr (commented):
Good work.
A few comments inline.
Looking forward to this feature.

gauravks:
Split into #5705 and #5706

Issue #5435, opened 2022-11-09
Unhelpful behaviour of UA_ClientConfig_setDefault and UA_Client_new


ukaea-chah:
## Description

The documentation for [Client](http://www.open62541.org/doc/master/client.html) says:

> The usual usage is as follows:
> 1. Create a client configuration with default settings as a starting point
> 2. Modifiy the configuration, e.g. modifying the timeout 
> 3. Instantiate a client with it 
> 4. After shutdown of the client, clean up the configuration (free memory)

There are some problems with this:

- UA_ClientConfig_setDefault sets a certificate validation of AcceptAll, and logs a warning about it, _even if_ you subsequently set a different certificate validation.
- If you don't want that spurious warning to appear on stdout, or at all, you would have to construct and set e.g. a null logger _before_ calling UA_ClientConfig_setDefault, then fix it up to a real logger later.
- If you want to set parts of the configuration different from the defaults, in some cases you have to undo / remove existing settings (taking care to call free() and clear() methods where needed).

Also, UA_Client_new() does not create a functional client: some settings such as timeout and localConnectionConfig.recvBufferSize are left as zero, preventing it establishing a connection to a server.  The user is forced to call UA_Client_getConfig and UA_ClientConfig_setDefault themselves, so there is little if any advantage over doing:
```C
 UA_ClientConfig config;
 memset(&config, 0, sizeof(config));
 UA_ClientConfig_setDefault(&config);
 client = UA_Client_newWithConfig(&config);
```
UA_Client_new also duplicates the code to set up a default event loop.

I would have hoped that UA_Client_new would have returned a usable client, to reduce the amount of boilerplate code needed by a newcomer trying out the library, working through tutorials etc.

I would suggest that:

-  UA_ClientConfig_setDefault should set only those fields which are unset (zero);
-  UA_Client_new should call UA_ClientConfig_setDefault rather than doing a half-hearted job itself;
- The recommended sequence for the general case would then be:
  1. Create an empty client configuration (memset to zero)
  2. Set any fields for which the default is not appropriate e.g. a different timeout or logger
  3. Call UA_ClientConfig_setDefault to set usable values for the remainder
  4. Instantiate a client with UA_Client_newWithConfig

## Background Information / Reproduction Steps

Used CMake options:

```bash
cmake ..
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): bdd3d62
 - [x] Operating system: GNU/Linux

[...]

PR #5463, opened 2022-11-23
[Review] Update handling of ClientConfig

ukaea-chah:
Make a number of related changes intended to improve the handling of UA_ClientConfig and its use with setting up a UA_Client.  Related to #5435.

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -800,8 +800,26 @@ UA_Client * UA_Client_new(void) {
 
 UA_StatusCode
 UA_ClientConfig_setDefault(UA_ClientConfig *config) {
-    config->timeout = 5000;
-    config->secureChannelLifeTime = 10 * 60 * 1000; /* 10 minutes */
+    /* The following fields are untouched and OK to leave as NULL or 0:
+     *  clientContext
+     *  userIdentityToken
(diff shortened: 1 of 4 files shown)


plugins/ua_config_default.c
```diff
@@ -823,57 +841,45 @@ UA_ClientConfig_setDefault(UA_ClientConfig *config) {
         config->eventLoop->registerEventSource(config->eventLoop, (UA_EventSource *)udpCM);
     }
 
-    if(config->sessionLocaleIdsSize > 0 && config->sessionLocaleIds) {
```
ukaea-chah, line 835: It's not clear to me in what circumstances a config with sessionLocaleIds would be presented.  And it seemed heavy-handed to actively delete them in a "setDefault" function.

jpfr: True.
We don't need that.

jpfr:
Many CI pipelines fail to run / run out of time.
Please have a look. Otherwise the proposed changes are great.

ukaea-chah:
> Many CI pipelines fail to run / run out of time.

CI now green.  So now I shall squash / rebase.

jpfr:
Perfect PR. 👍

PR #5464, opened 2022-11-23
fix(server) Refine fixup of logger pointers

ukaea-chah:
Only fix up the loggers of the event loop and security policies if they pointed into the supplied config.  E.g. a shared external event loop may use a logger with a different context from that of the server instance.


jpfr:
Thanks!

PR #5478, opened 2022-11-26
Fix the build error in pubsub_interrupt_publish_tsn.c

wenshan1:
For OPEN62541 upgraded, some variables and functions are removed or changed. So change the file,
pubsub_interrupt_publish_tsn.c to meet the upgradation.


jpfr:
👍

Issue #5462, opened 2022-11-22
[-Werror,-Wmissing-braces] while packge for Android on Unreal Engine 5


ffreality:
## Description

Hi, I am working on an Unreal Engine 5 OPC UA Client plugin based on open62541. Window side is finished and I would like to extend it to also Android.

I can create necessary ".c" and ".h" files on Ubuntu with CMake while selecting POSIX as my target platform and integrate them to Unreal Engine on Windows 11. But when I package my project (Creating an APK), I have these errors.

Is there any suggestion for that ?

```
open62541.h(17401,29): error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
LogPlayLevel:     UA_ExpandedNodeId id = {0};
LogPlayLevel:                                                 ^
LogPlayLevel:                                                 {}

open62541.c(5785,51): error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
LogPlayLevel: const UA_ExpandedNodeId UA_EXPANDEDNODEID_NULL = {0};

LogPlayLevel: Warning:
open62541.c(8097,9): warning: No native IEEE 754 format detected. Use slow generic encoding. [-W#pragma-messages]

open62541.c(212823,52): error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
LogPlayLevel:     UA_PubSubChannelDataUDPMC defaultValues = {0, {0}, 0, {0}, 255, true, true, true};
LogPlayLevel:                                                                                          ^
LogPlayLevel:                                                                                          {}

open62541.c(212823,60): error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
LogPlayLevel:     UA_PubSubChannelDataUDPMC defaultValues = {0, {0}, 0, {0}, 255, true, true, true};
LogPlayLevel:                                                                                                    ^
LogPlayLevel:                                                                                                    {}

open62541.c(213202,37): error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
LogPlayLevel:         struct ipv6_mreq groupV6 = {0};
LogPlayLevel:                                                        ^
LogPlayLevel:                                                         {}

open62541.c(213250,35): error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
LogPlayLevel:         struct ip_mreq groupV4 = {0};
LogPlayLevel:                                                    ^
LogPlayLevel:                                                     {}

[...]

PR #5480, opened 2022-11-26
Use memset to zero out instead of '= {0};' (Android NDK problems)

jpfr:


ffreality:
https://github.com/open62541/open62541/issues/5462#issuecomment-1328125933

there are additional {0} to memset errors, too. I changed them and it works.

Issue #5447, opened 2022-11-15
Client do not respond appropriately to PublishResponse with status Bad_NoSubscription


nokem61:
## Description

The client does not react in any way to the lack of subscription on the server. There may be a situation when the subscription "disappears", in which case PublishResponse will come with Bad_NoSubscription status. In this case, we should execute UA_Client_Subscription_deleteInternal for this subscription because it no longer exists. So when we have callback for delete subscription we can react correctly.

## Background Information / Reproduction Steps

When the client reconnects to the server that has restarted, there are situations when the connection is successful, and there are subscriptions on the client side, but there are no longer any on the server.

Used CMake options:

```bash
cmake -G "Visual Studio 15 2017 Win64" ..\.. -DCMAKE_INSTALL_PREFIX:PATH="../install_64_release" -DCMAKE_BUILD_TYPE:STRING="RelWithDebInfo" -DUA_ARCHITECTURE:STRING="win32" -DUA_NAMESPACE_ZERO:STRING="FULL" -DUA_ENABLE_SUBSCRIPTIONS_EVENTS:BOOL="1" -DUA_ENABLE_ENCRYPTION:BOOL="1" -DUA_ENABLE_ENCRYPTION_OPENSSL:BOOL="1" -DOPENSSL_INCLUDE_DIR:PATH="OpenSSL/include64" -DSSL_EAY_RELEASE:FILEPATH="../../OpenSSL/bin/x64/Release/libssl" -DLIB_EAY_DEBUG:FILEPATH="../../OpenSSL/bin/x64/Debug/libcrypto" -DSSL_EAY_DEBUG:FILEPATH="../../OpenSSL/bin/x64/Debug/libssl" -DLIB_EAY_RELEASE:FILEPATH="../../OpenSSL/bin/x64/Release/libcrypto" -DUA_ENABLE_HISTORIZING:BOOL="1" -DBUILD_SHARED_LIBS:BOOL="1" -DUA_LOGLEVEL:STRING="100" -DUA_ENABLE_DISCOVERY:BOOL="1"  -DUA_DEBUG:BOOL="0" -DUA_ENABLE_DISCOVERY_MULTICAST:BOOL="1" -DCMAKE_SYSTEM_VERSION="10.0.17763.0"
```

## Modified function
```C
static void
UA_Client_Subscriptions_processPublishResponse(UA_Client *client, UA_PublishRequest *request,
                                               UA_PublishResponse *response) {
[...]

PR #5489, opened 2022-11-30
Fix response on Publish status: BAD_NOSUBSCRIPTION

nokem61:
Changes for https://github.com/open62541/open62541/issues/5447


jpfr:
Nice!

PR #5496, opened 2022-12-04
WIP feat(server) overhaul access control examples and fix check of method execution right with async methods

andreasebner:
The PR contains the following changes:

- Restructure the async tutorials structure.
- Fix currently crashing async example.
- Fix the test of user method execution rights in case of async methods (server side)
- Extend server with util function for generating a c-string (null terminated) from a UA_String

ToDo:
From my point of view, the current implementation of the client_method_async makes no sense. There is a call of a methods in the method callback, which results in an endless loop.

### DIFF
--- a/include/open62541/types.h
+++ b/include/open62541/types.h
@@ -11,6 +11,8 @@
  *    Copyright 2015-2016 (c) Oleksiy Vasylyev
  *    Copyright 2017 (c) Stefan Profanter, fortiss GmbH
  *    Copyright 2017 (c) Thomas Stalder, Blue Time Concept SA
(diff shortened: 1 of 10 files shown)


include/open62541/types.h
```diff
@@ -196,6 +198,9 @@ typedef struct {
 UA_String UA_EXPORT
 UA_String_fromChars(const char *src) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
 
+UA_StatusCode UA_EXPORT
```
jpfr, line 201: Is this zero-terminated?
Does c have to be pre-allocated?
Document.

jpfr (commented):
Looks good. Small nitpick.

andreasebner:
The issues are solved and the PR outdated.

PR #5498, opened 2022-12-05
[WIP] Switch PubSub to EventLoop model

jpfr:


jpfr:
> With the eventloop architecture, do we still have the possibility of switching mqtt library from MQTT-C to other libraries?

Yes we do. The MQTT implementation just has to implement the "ConnectionManager" interface from the `eventloop.h`.

keba-estr:
@jpfr: Is this merge request still relevant or has it already been resolved with other merge requests?

jpfr:
You are right. This is already resolved.

jpfr:
@applea9, yes you can still use a different mqtt library.

Just add a second implementation of this
https://github.com/open62541/open62541/blob/master/arch/eventloop_mqtt.c
and add your custom "ConnectionManager" to the server configuration.

PR #5507, opened 2022-12-09
[REVIEW]: Implement the client side of reverse connect

basyskom-jvoe:
This is the first attempt at server-side reverse connect.
The proposed API is already used in the server-side reverse connect tests.


jpfr:
The log looks good.
But I don't see any log information for the client socket.
Also no information that the client socket creation failed.
At least that should be visible.

basyskom-jvoe:
@jpfr The builds in the CI are passing now, the PR is ready for review.

src/client/ua_client_internal.h
```diff
@@ -177,6 +177,12 @@ struct UA_Client {
 #if UA_MULTITHREADING >= 100
     UA_Lock clientMutex;
 #endif
+
+    UA_Boolean isReverseConnect;
```
jpfr, line 181: I think we should keep all the connection state information in one place: in the SecureChannel.
It is easier to reason about a single state-machine than many coupled state machines.
Also we notify state changes to the user application in a callback that has the SecureChannel state.
And in the SecureChannel you already have the the ConnectionManager+uintptr-id combination.

The following would be similar to the existing logic and do all in one state-machine:

1) Add two additional states to the `UA_SecureChannelState` enum. Something like
- `UA_SECURECHANNELSTATE_REVERSE_LISTENING` and
- `UA_SECURECHANNELSTATE_REVERSE_CONNECTED` (before the RHE is received)

2) Store the "listening socket" in the SecureChannel when the listening state is set.

3) Once the server does a reverse connect you get called in the "connectionCallback" set for the listen connection.
From within this callback
- shutdown the listening socket (if required set a new context pointer for the current connection, as the connectionCallback is called again when the listening socket has fully closed)
- Replace the socket stored in the client's SecureChannel with the new "active connection"
- Set the state to `UA_SECURECHANNELSTATE_REVERSE_CONNECTED` or similar.

src/client/ua_client_connect.c
```diff
@@ -1665,7 +1835,10 @@ UA_Client_disconnectSecureChannel(UA_Client *client) {
 
 UA_StatusCode
 UA_Client_disconnect(UA_Client *client) {
-    UA_LOCK(&client->clientMutex);
+    closeListeningReverseConnectSocket(client);
```
jpfr, line 1838: This is also called from within disconnectSecureChannel that is called just below.
So shouldn't be necessary here.

src/client/ua_client_connect.c
```diff
@@ -1601,6 +1767,8 @@ closeSessionCallback(UA_Client *client, void *userdata,
 
 UA_StatusCode
 UA_Client_disconnectAsync(UA_Client *client) {
+    closeListeningReverseConnectSocket(client);
```
jpfr, line 1770: The below async approach will call disconnectSecureChannel eventually.
So closing the reverse socket here is redundant.

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1557,128 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+        return;
+    }
+
+    UA_UNLOCK(&client->clientMutex);
+}
+
+UA_StatusCode UA_Client_startListeningForReverseConnect(UA_Client *client, const UA_String *listenHostnames,
+                                                        size_t listenHostnamesLength,
+                                                        UA_UInt16 port) {
+    UA_Client_disconnect(client);
```
jpfr, line 1600: Don't disconnect here.
Rather check if the client is already connected.
And abort in that case.
Similar to the "normal" connect.

src/client/ua_client_connect.c
```diff
@@ -409,6 +409,25 @@ sendHELMessage(UA_Client *client) {
...
+                                &UA_TRANSPORT[UA_TRANSPORT_TCPREVERSEHELLOMESSAGE], NULL);
+
+    if(retval == UA_STATUSCODE_GOOD) {
+        UA_String_clear(&client->endpointUrl);
+        UA_String_copy(&rheMessage.endpointUrl, &client->endpointUrl);
+    }
+
+    UA_TcpReverseHelloMessage_clear(&rheMessage);
+
+    sendHELMessage(client);
```
jpfr, line 428: sendHELMessage will set the state to `UA_SECURECHANNELSTATE_HEL_SENT`.
To me that is a different state. As in "we have sent HEL, but have not receive an ACK.

I'd rather set to `UA_SECURECHANNELSTATE_ACK_RECEIVED` after this sendHelMessage succeeds.
As in "the initial handshake is done".

Maybe I don't understand the logic.
Isn't the client supposed to send the following OPN request to open a SecureChannel?
That is only done by the client if the state is `UA_SECURECHANNELSTATE_ACK_RECEIVED`.

jpfr (commented):
Good to see you got the CI all green.
Comments inline.

Looking forward getting this great feature in. 👍

src/client/ua_client_connect.c
basyskom-jvoe: @jpfr The server's RHE message is the first message that is sent over the TCP connection and must be answered with a HEL message by the client (see https://reference.opcfoundation.org/Core/Part6/v104/docs/7.1.3#_Ref456138622). After this, the sequence is the same as for a normal connect.
I like the idea of having this intermediate UA_SECURECHANNELSTATE_HEL_SENT state so I can see where the process got stuck if the ACK message is not sent due to some error.

basyskom-jvoe:
@jpfr The PR is ready for another round of review.

src/client/ua_client_connect.c
```diff
@@ -1050,11 +1069,14 @@ connectActivity(UA_Client *client) {
         /* Nothing to do if the connection has not opened fully */
     case UA_SECURECHANNELSTATE_CONNECTING:
     case UA_SECURECHANNELSTATE_CLOSING:
+    case UA_SECURECHANNELSTATE_HEL_SENT:
         return;
 
         /* Send HEL */
     case UA_SECURECHANNELSTATE_CONNECTED:
-        client->connectStatus = sendHELMessage(client);
+        if (!client->channel.isReverseConnect) {
```
jpfr, line 1077: You don't need to test `isReverseConnect`.
That should only be set for the `UA_SECURECHANNELSTATE_REVERSE_CONNECTED` state.
The original code should work fine.

src/client/ua_client_connect.c
```diff
@@ -1197,8 +1219,12 @@ __Client_networkCallback(UA_ConnectionManager *cm, uintptr_t connectionId,
          * before sending the HEL message. */
         if(state == UA_CONNECTIONSTATE_OPENING)
             client->channel.state = UA_SECURECHANNELSTATE_CONNECTING;
-        else /* state == UA_CONNECTIONSTATE_ESTABLISHED */
-            client->channel.state = UA_SECURECHANNELSTATE_CONNECTED;
+        else {
+            if (client->channel.isReverseConnect)
```
jpfr, line 1223: Test for `UA_SECURECHANNELSTATE_REVERSE_LISTENING` instead of the `isReverseConnect` flag.

src/client/ua_client_connect.c
```diff
@@ -1291,12 +1317,40 @@ __Client_networkCallback(UA_ConnectionManager *cm, uintptr_t connectionId,
     UA_UNLOCK(&client->clientMutex);
 }
 
+static void closeListeningReverseConnectSocket(UA_Client *client) {
+    UA_LOCK(&client->clientMutex);
```
jpfr, line 1321: You are unlocking below just to lock again here.
Not needed.

src/client/ua_client_connect.c
```diff
@@ -1291,12 +1317,40 @@ __Client_networkCallback(UA_ConnectionManager *cm, uintptr_t connectionId,
...
+    if (client->channel.state != UA_SECURECHANNELSTATE_REVERSE_LISTENING) {
+        UA_UNLOCK(&client->clientMutex);
+        return;
+    }
+
+    client->channel.listenerDisconnected = true;
+    client->channel.connectionManager->closeConnection(client->channel.connectionManager, client->channel.connectionId);
+
+    UA_EventLoop *el = client->config.eventLoop;
```
jpfr, line 1330: The cleanup of the listen socket can be done asynchronous in the next EventLoop iteration.
No need to run the EventLoop here.

Just reset `client->channel.connectionId` (to zero or to the id of the new "active socket").
Then, at some later point you will drop into the networkCallback with a closing socket with `connectionId != client->channel.connectionId`.
That tells you that this is the cleanup for an old connection and there is nothing left to do in the networkCallback (no SecureChannel to close).

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1567,145 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+        return;
+    }
+
+    UA_UNLOCK(&client->clientMutex);
+}
+
+UA_StatusCode UA_Client_startListeningForReverseConnect(UA_Client *client, const UA_String *listenHostnames,
+                                                        size_t listenHostnamesLength,
+                                                        UA_UInt16 port) {
+    if (client->channel.state > UA_SECURECHANNELSTATE_FRESH && client->channel.state < UA_SECURECHANNELSTATE_CLOSED)
```
jpfr, line 1626: Add a log warning here.
Otherwise users without access direct to the source will be confused on what is going on.

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1567,145 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+        return UA_STATUSCODE_BADINVALIDSTATE;
+
+    UA_LOCK(&client->clientMutex);
```
jpfr, line 1629: Move the lock to the beginning.
There can be a race condition already when you access client->channel.state.

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1567,145 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+
+    const UA_String tcpString = UA_STRING_STATIC("tcp");
+    UA_StatusCode res = UA_STATUSCODE_BADINTERNALERROR;
+
+    // FIXME: Are these good default values?
+    client->noReconnect = true;
+    client->channel.config.recvBufferSize = 8192;
```
jpfr, line 1636: Take the default values from the client config.
The same as the normal SecureChannel logic.

src/client/ua_client_connect.c
```diff
@@ -1636,6 +1829,7 @@ UA_Client_disconnectAsync(UA_Client *client) {
 
 UA_StatusCode
 UA_Client_disconnectSecureChannel(UA_Client *client) {
+    closeListeningReverseConnectSocket(client);
```
jpfr, line 1832: You don't need `closeListeningReverseConnectSocket`.
The normal `closeSecureChannel` calls UA_SecureChannel_shutdown which call connectionManager->closeConnection.
That works just the same for the listening socket.

src/ua_securechannel.h
```diff
@@ -105,6 +105,11 @@ struct UA_SecureChannel {
     UA_ConnectionManager *connectionManager;
     uintptr_t connectionId;
 
+    /* Reverse connect */
+    uintptr_t reverseConnectListenConnectionId;
```
jpfr, line 109: You don't any of these three.
Simply reuse the `connectionId` also for the listen socket.
And replace with the "active socket" once it opens.

The boolean flags are not needed either.
The `UA_SECURECHANNELSTATE_REVERSE_LISTENING` and `UA_SECURECHANNELSTATE_REVERSE_CONNECTED` flags ought to be enough.

include/open62541/common.h
```diff
@@ -175,8 +175,10 @@ typedef enum {
 
 typedef enum {
     UA_SECURECHANNELSTATE_FRESH = 0,
+    UA_SECURECHANNELSTATE_REVERSE_LISTENING,
     UA_SECURECHANNELSTATE_CONNECTING,
     UA_SECURECHANNELSTATE_CONNECTED,
+    UA_SECURECHANNELSTATE_REVERSE_CONNECTED,
```
jpfr, line 181: The new states work great!
Please adjust UA_SecureChannel_isConnected so that a reverse-listening SecureChannel is also considered "connected".
(So socket->shutdown is called on client->channel.connectionId.)

```c
UA_Boolean
UA_SecureChannel_isConnected(UA_SecureChannel *channel) {
    return (channel->state >= UA_SECURECHANNELSTATE_CONNECTING &&
            channel->state < UA_SECURECHANNELSTATE_CLOSING);
}
```

jpfr (commented):
This is coming together nicely!
Who would have though we can have this feature in <200 new lines.

There is some more potential for simplification.
I'm pretty sure it will work to reuse the `channel.connectionId` for both the listen- and the active-socket. And switch them out (+ shutdown the listen-socket) when the active connection opens.

See inline.

Keep up the good work!

src/client/ua_client_connect.c
basyskom-jvoe: The suggested change makes the state callback miss the return to UA_SECURECHANNELSTATE_CLOSED if UA_Client_disconnect() is called while in state UA_SECURECHANNELSTATE_REVERSE_LISTENING. I'd like to keep the event loop run calls here if that's fine with you.

src/client/ua_client_connect.c
basyskom-jvoe: Wouldn't this also apply to https://github.com/open62541/open62541/pull/5507#discussion_r1111092912?

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

include/open62541/common.h
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1538,141 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+    UA_Client *client = (UA_Client*)application;
+
+    UA_LOCK(&client->clientMutex);
+
+    if (!client->channel.connectionId) {
+        client->channel.connectionId = connectionId;
+        *connectionContext = reverseConnectIndicator;
+    }
+
+    if (*connectionContext == reverseConnectIndicator && state == UA_CONNECTIONSTATE_CLOSING) {
```
jpfr, line 1557: Add a small comment to avoid misunderstandings.

This is what I think is happening. Right?
So in the outer if the listen-socket is closing.
And in the inner if you check whether an active connection has been added since.
If there is an active connection (and we are shutting the listening socket down ourselves), do nothing.
Otherwise set the channel to "closed".

For the inner if `client->channel.connectionId == connectionId` should be enough.
As we immediately reset the connectionId when the active socket opens.

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1538,141 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+        notifyClientState(client);
+    }
+
+    /* This is a connection initiated by a server, disconnect the listener */
+    if (client->channel.connectionId != connectionId) {
+        cm->closeConnection(cm, client->channel.connectionId);
+    }
+
+    /* First call with the new connection from the server, reset secure channel information */
+    if (client->channel.connectionId != connectionId && *connectionContext == reverseConnectIndicator) {
```
jpfr, line 1578: This can be merged with the if-condition just above.
Because the the new active connection will inherit the reverseConnectIndicator.
And we only reset it during the first callback.

jpfr (commented):
Nice! This is really coming together now.
A few last comments inline. Mostly cosmetic.

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
basyskom-jvoe: Done

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1538,141 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+                         void *application, void **connectionContext,
+                         UA_ConnectionState state, const UA_KeyValueMap *params,
+                         UA_ByteString msg) {
+    void *reverseConnectIndicator = (void *)(uintptr_t)0xFFFFFFFF;
+
+    UA_Client *client = (UA_Client*)application;
+
+    UA_LOCK(&client->clientMutex);
+
+    if (!client->channel.connectionId) {
```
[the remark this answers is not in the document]
basyskom-jvoe: Done

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1538,141 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
     return client->connectStatus;
 }
 
+static void
+__Client_reverseConnectCallback(UA_ConnectionManager *cm, uintptr_t connectionId,
+                         void *application, void **connectionContext,
+                         UA_ConnectionState state, const UA_KeyValueMap *params,
+                         UA_ByteString msg) {
+    void *reverseConnectIndicator = (void *)(uintptr_t)0xFFFFFFFF;
```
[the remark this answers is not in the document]
basyskom-jvoe: Done

src/client/ua_client_connect.c
```diff
@@ -409,6 +409,25 @@ sendHELMessage(UA_Client *client) {
     return UA_STATUSCODE_GOOD;
 }
 
+void processRHEMessage(UA_Client *client, const UA_ByteString *chunk) {
+    UA_LOG_INFO(&client->config.logger, UA_LOGCATEGORY_CLIENT, "RHE received");
+
+    size_t offset = 0; /* Go to the beginning of the TcpHelloMessage */
+    UA_TcpReverseHelloMessage rheMessage;
+    UA_StatusCode retval =
```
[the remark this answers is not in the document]
basyskom-jvoe: Done

src/client/ua_client_connect.c
```diff
@@ -1513,6 +1538,141 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+        UA_UNLOCK(&client->clientMutex);
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
+
+    if(el->state != UA_EVENTLOOPSTATE_STARTED) {
+        res = el->start(el);
+        UA_CHECK_STATUS(res, UA_UNLOCK(&client->clientMutex); return res);
+    }
+
+    for(UA_EventSource *es = client->config.eventLoop->eventSources; es != NULL; es = es->next) {
```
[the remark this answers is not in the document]
basyskom-jvoe: Done

basyskom-jvoe:
@jpfr Many thanks for your reviews!

Issue #218, opened 2015-04-10
Remove the "open62541" namespace and application name


jpfr:
Currently, there is no way to add namespaces and to change the standard application name from outside the library.

Both should be provided for in the 0.1 release.

Stasik0:
do you mean this branch?
https://github.com/acplt/open62541/blob/f1ce7710ac3e997e96fc621622c5af326330d026/src/server/ua_server.c#L102

Stasik0:
relates to #151, is this a duplicate?

Stasik0:
is now set via server config: https://github.com/acplt/open62541/blob/master/src/server/ua_server.c#L18
still hardcoded for certificate generation, we need to doc it somewhere

PR #5511, opened 2022-12-12
feat(tests): Node ordering test case for open62541-nodeset-loader

keba-uso:
Within this PR, the ordering of the nodes within a nodeset XML file is tested for the [open62541-nodeset-loader](https://github.com/open62541/open62541-nodeset-loader). A modified version of the [DI](https://github.com/OPCFoundation/UA-Nodeset/blob/latest/DI/Opc.Ua.Di.NodeSet2.xml) companion specification is used as an example, where the parent node with id **ns=1;i=15001** is defined after its child nodes:

- ns=1;i=15002
- ns=1;i=15003
- ns=1;i=15004
- ns=1;i=15005
- ns=1;i=15006
- ns=1;i=15007
- ns=1;i=15008
- ns=1;i=15031
- ns=1;i=15032
- ns=1;i=15033

**NOTE**: Before this PR can be considered as completed, the dependency PR [#218](https://github.com/open62541/open62541-nodeset-loader/pull/218) must be merged.


keba-uso:
@jpf, or anyone else from the maintainers side, I would like to inform you that this PR is ready for review.

PR #5513, opened 2022-12-14
fix(core): Fix encoding of the structure with optional fields

anzeskerjanc:


jpfr:
Uh. Good find.

1. This should have broken the unit tests in the first place.
Please extend the unit test for structs with optional fields in a way that would have broken before and now works.
3. Make this a PR for the 1.3 branch. It warrants a patch release for 1.3.

anzeskerjanc:
I changed base branch and added missing tests.

jpfr:
Muchas gracias 👍

PR #5515, opened 2022-12-15
fix(client): Check existence of event loop in UA_Client_newWithConfig

keba-estr:
Only copy event loop logger in UA_Client_newWithConfig if an event loop has been set in the config.


jpfr:
👍

Issue #5527, opened 2022-12-20
Bug in getNamespaceByIndex


chhuebner:
There is a bug in getNamespaceByIndex:

https://github.com/open62541/open62541/blob/8eeb6c33da78a0d020aeafcf20eab6cbf84a3c0a/src/server/ua_server.c#L131 

It should be: 

if(namespaceIndex >= server->namespacesSize)

keba-estr:
Yes, this is an error. Could you please create a pull request with the fix?

chhuebner:
I created the pull request: #5528

chhuebner:
Created new PR #5534 for 1.3 branch

jpfr:
Merged.
Will go to master from there.

PR #5528, opened 2022-12-21
fix(server): Range check in getNamespaceByIndex (#5527)

chhuebner:


basyskom-meerkoetter:
Could you please also extend the unit test to cover this boundary condition?
See https://github.com/open62541/open62541/blob/8eeb6c33da78a0d020aeafcf20eab6cbf84a3c0a/tests/server/check_server.c#L49

jpfr:
If this can be backported I’d to have this as a pr for the 1.3 branch.

jpfr:
I switched the target branch to 1.3.
But this pulled in hundreds of commits.
Please rebased on top of  1.3.

chhuebner:
I will create a new PR for 1.3 branch

PR #5535, opened 2022-12-23
feat(core): Add XML en/decode support for ExtensionObject and Variant

keba-uso:
- [x] Include depth tracking and `prettyPrint` support for XML en/decode
- [x] Add support for XML en/decode of `ExtensionObject`
- [ ] Add support for XML en/decode of `Variant`


src/ua_types_encoding_xml.c
```diff
@@ -48,6 +226,60 @@ xmlEncodeWriteChars(CtxXml *ctx, const char *c, size_t len) {
...
+}
+
+static status UA_FUNC_ATTR_WARN_UNUSED_RESULT
+writeXmlElement(CtxXml *ctx, const char *name,
+                const void *value, const UA_DataType *type) {
+    status ret = UA_STATUSCODE_GOOD;
+    UA_Boolean prevPrintVal = ctx->printValOnly;
+    ctx->printValOnly = false;
+    ret |= writeXmlElemNameBegin(ctx, name);
+    ctx->printValOnly = XML_TYPE_IS_PRIMITIVE;
```
jpfr, line 275: printvalonly is boolean.
Don't understand this...

src/ua_types_encoding_xml.c
```diff
@@ -806,29 +1715,229 @@ const decodeXmlSignature decodeXmlJumpTable[UA_DATATYPEKINDS] = {
...
+
+    ParseCtxXml *ctxt = (ParseCtxXml*)ctx;
+    XmlParsingCtx *pctxt = ctxt->parseCtx;
+    XmlValue *val = ctxt->value;
+    const char *localNameChr = (const char*)localname;
+
+    // printf("%s %d: %s\n", __FILE__, __LINE__, localNameChr);
+
+    if(!strncmp(localNameChr, "ListOf", strlen("ListOf")))
+        val->isArray = UA_TRUE;
```
jpfr, line 1778: Use just `true`.
Les visual noise.

jpfr (commented):
This is already promising.
Comments inline.

The biggest point I have concerns the structure-decoding logic.
XML parsers come in two flavors:
- DOM: First parse a tree-representation of the XML. Then walk over the tree
- SAX: Use a stream-processing approach with callbacks. At no point is the entire XML-tree loaded into memory.

Here you somehow mix both.
Use SAX, but then build up a home-grown tree-representation of the XML document.
Then do a second pass that walks over the tree.

In order to be fast and also able to process big XML documents, I prefer a 100% SAX implementation.

That is, from a SAX callback only look at the "current" element.
And keep the necessary state to process the "current" element in the context data structure.

The only big hurdle is recursion. And that can be handled quite well by introducing a "stack" in the context data structure. So you can keep track which member of which structure you are decoding (recursively). Like this:

```c
#define UA_XML_MAX_RECURSION 16

struct {
     void *data;
     const UA_DataType *type;
} stack [UA_XML_MAX_RECURSION];
UA_Byte depth;
```

andreasebner:
Thank you for your contribution back then. The feature has since been implemented and is now supported in the latest master and in version 1.5.

Issue #5388, opened 2022-10-19
Building lib fpr WEC7: undeclared identifier UA_POLLOUT/__BitScanForward


fraxh22:
I have downloaded today from github master branch 1.3. -  Git-Revision: v1.3.3-undefined

According "Building with cmake on Windows" I have downloaded Python 3.1 and cmake.
I want to use OPCUA library on Windows Embedded Compact 7 with ARM processor, Compiler has to be Visual Studio 2008.
So I started cmake with line:

cmake -DUA_ARCHITECTURE=wec7  -DUA_ENABLE_AMALGAMATION=ON  -DUA_ENABLE_DISCOVERY_SEMAPHORE=OFF .. -G "Visual Studio 9 2008"

Then I could build open6251-amalgamation-header and open62541_amalgamation-source without errors.

Now using the two files open62541.h and open62541.c in my project I get error messages, mainly
error C2065: UA_POLLOUT: undeclared identifier
error C3861: _BitScanForward: identifier not found
and some other errors related to missing UA_POLLOUT

Note: Above procedure did work with revision 1.2 two years ago

Any ideas ? Thank you !

matiashr:
how could you build using vs2008, the library requires c99, is that really supported by vs2008 ?

fraxh22:
Please find attached amalgamated files that could be compiled with VS2000. Please use compile option /TP



Mit freundlichen Grüßen / best regards

i.A. Franz Hofer
Entwicklung - R. & D.
________________________________
[cid:at_logo_2d3bfbc8-d5ed-4bc5-b025-bdee46831ae9.png]

ATOMA Gesellschaft für automatische Waagen GmbH
Traunreuter Str. 2-4 | D-84478 Waldkraiburg | Tel. +49 8638 961-214 | Fax +49 8638 961-111
***@***.******@***.***> | www.multipond.com<http://www.multipond.com>

Follow us for the latest updates
[...]

PR #5536, opened 2022-12-23
Disable polling for Windows CE

beffe:
Windows CE does not support socket status determination via WSAPoll, so that CPU-saving measure needs to be disabled.

This is an attempt to solve issue #5388.

Please note that I am actually working on Windows Embedded Compact 2013 (aka wec8) and not wec7. I got it running there but I have no way to test this on wec7, so it would be great if @fraxh22 could test if this solves the problem.

I am also totally open for input since the changes feel hacky.
If this becomes an agreeable fix eventually, I am also provide the changes to get it running on wec2013, i.e., as a new architecture if the maintainers agree.

Cheers!

### DIFF
--- a/deps/aa_tree.c
+++ b/deps/aa_tree.c
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
  *    Copyright 2020 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
  */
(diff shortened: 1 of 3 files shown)


deps/aa_tree.c
```diff
@@ -1,21 +1,23 @@
...
 
 #include "aa_tree.h"
 #include <stddef.h>
 
-#if !defined(_MSC_VER) || _MSC_VER >= 1800
-# include <inttypes.h>
-#elif !defined(uintptr_t)
-  /* Workaround missing standard includes in older Visual Studio */
-# if !defined(__midl) && (defined(_X86_) || defined(_M_IX86)) && _MSC_VER >= 1300
+#ifndef UNDER_CE
```
jpfr, line 11: That's a deep nesting of #ifdefs and a bit confusing.
I think it could be a flat list. As in
``` 
#ifdef UNDER_CE

#elif !defined(_MSC_VER) ....

#else

#endif
```

jpfr (commented):
Looks good.
One of the CI tests broke.
Have a look at the inline comments and resubmit for the CI run.

deps/aa_tree.c
beffe: Changed it as you suggested and added a comment for `UNDER_CE` since there's nothing to define/include in that case.

PR #5558, opened 2023-01-11
fix(server): Fix ConditionRefresh events with multiple server instances

teijokinnunen:
Refresh event node ids were stored in a static variable; this failed if there were more than one server instance in the same address space. The node ids are now moved to UA_Server.


jpfr:
Nice!

PR #5559, opened 2023-01-11
fix(core) remove special case for enum handling in type compatible check

andreasebner:
During the write of a value, we check if the node datatype and the incoming value are compatible. In some cases a automatic adjustment can be done. With changes in the master, the adjust behavior for enums was not triggered anymore. This PR changes the compatible type check and therefore enables the enum-type transformation (from int32 to dedicated enum type).


keba-estr:
Real server also does not correctly with this change.

keba-estr:
@andreasebner: Is there any working solution for this problem?

PR #5563, opened 2023-01-12
fix(pubsub): Fix memory leak in udp transport layer

keba-estr:


keba-estr:
Saved pointer in handle of other structure had to be reset to satisfy clang-analyzer.

keba-estr:
I had overlooked something, see #5564.

Issue #5421, opened 2022-11-02
open62541Targets.cmake uses absolute paths in the INTERFACE_LINK_LIBRARIES


carl-cimon:
## Description

In the generated open62541Targets.cmake file, the INTERFACE_LINK_LIBRARIES uses the absolute paths of any linked libraries, which causes the library to not be relocatable. This is affecting our cross compile toolchain, as the path of the OpenSSL library when we build the toolchain is different than location after the toolchain is installed.

I'm going to be honest, I'm not super knowledgeable about the toolchain build process (as that is handled by another member of my team), but looking into how the INTERFACE_LINK_LIBRARIES is generated in cmake, it seems like the absolute path is being provided to the target_link_libraries() call in the CMakeLists. According to the [cmake docs](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages), it seems that using the imported target for an external library is better practice.

I was able to fix this on our end by modifying the CMakeLists to do just that, replacing:
```
list(APPEND open62541_LIBRARIES ${OPENSSL_LIBRARIES})
````
with
```
list(APPEND open62541_LIBRARIES OpenSSL::SSL)
list(APPEND open62541_LIBRARIES OpenSSL::Crypto)
```

I do not know if this is the best method to fix the issue, but figured I'd share a fix in case anyone else runs into a similar problem. It did require us to find_package() OpenSSL separately in our project which uses open62541 though, likely to be able to resolve where the libraries are, but this should be acceptable in most cases. I assume a similar fix would work for the other external libraries as well.

## Background Information / Reproduction Steps

Here's the steps I used on a windows build.  This occurs for both and static shared builds. I believe this would happen with any build which links an external library, not just OpenSSL. 

```
git clone https://github.com/open62541/open62541.git
cd open62541
git checkout ff73268
git submodule update --init --recursive

mkdir build && cd build

[...]

PR #5565, opened 2023-01-13
fix(build): replace absolute paths of OpenSSL with imported target

zimmermannt-tq:
${OPENSSL_LIBRARIES} contains the absolute paths to the OpenSSL library. Instead, the imported targets are used. This is important when using exported targets e.g. while cross compiling. This is fixing #5421.


jpfr:
Nice

Issue #4345, opened 2021-04-16
pubsub_TSN_publisher.c:70:10: fatal error: sys/io.h: No such file or directory


razr:
## Description

Native compilation of the real-time PubSub example fails on arm64 Ubuntu 20.04
```bash
/work/tmp/open62541/examples/pubsub_realtime/pubsub_TSN_publisher.c:70:10: fatal error: sys/io.h: No such file or directory
   70 | #include <sys/io.h>
      |          ^~~~~~~~~~
compilation terminated.
```

## Background Information / Reproduction Steps

```
# pull arm64 Ubuntu
$ docker pull arm64v8/ubuntu:focal

# clone open62541
$ cd ~/tmp
$ git clone https://github.com/open62541/open62541.git

# run docker and install a development environment
$ docker run -h open62541 -it -v ~/tmp:/root/tmp arm64v8/ubuntu:focal
root@open62541:/# apt-get update && apt-get install -q -y build-essential cmake lsb-release software-properties-common git vim

# build open62541 real-time PubSub example as described in https://github.com/open62541/open62541/blob/master/examples/pubsub_realtime/README_pubsub_TSN.txt#L57 
$ cd /root/tmp/open62541 
$ mkdir build && cd build
$ cmake -DUA_BUILD_EXAMPLES=ON -DUA_ENABLE_PUBSUB=ON -DUA_ENABLE_PUBSUB_ETH_UADP=ON ..
$ make
...
[ 49%] Building C object examples/CMakeFiles/pubsub_TSN_publisher.dir/pubsub_realtime/pubsub_TSN_publisher.c.o
/work/tmp/open62541/examples/pubsub_realtime/pubsub_TSN_publisher.c:70:10: fatal error: sys/io.h: No such file or directory
   70 | #include <sys/io.h>
      |          ^~~~~~~~~~
compilation terminated.
make[2]: *** [examples/CMakeFiles/pubsub_TSN_publisher.dir/build.make:63: examples/CMakeFiles/pubsub_TSN_publisher.dir/pubsub_realtime/pubsub_TSN_publisher.c.o] Error 1
...
```
```sys/io.h``` header does not exist on arm64 Ubuntu 20.04, see https://packages.ubuntu.com/search?suite=focal&arch=arm64&mode=exactfilename&searchon=contents&keywords=sys%2Fio.h

This trivial patch fixes the problem
```bash
root@2754b3d0e63a:/work/tmp/open62541/build# git diff
[...]

PR #5571, opened 2023-01-16
fix(examples): remove sys/io.h

zimmermannt-tq:
For Arm processors sys/io.h does not exist, which
results in a build error. On x86 the header is not needed. Therefore it can be removed. This is also a fix for #4345

### DIFF
--- a/examples/pubsub_realtime/pubsub_TSN_loopback.c
+++ b/examples/pubsub_realtime/pubsub_TSN_loopback.c
@@ -77,7 +77,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <linux/types.h>
-#include <sys/io.h>
 #include <getopt.h>
 
 /* For thread operations */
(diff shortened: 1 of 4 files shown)


jpfr:
Thanks for this.

A small heads-up.
We have reworked the PubSub configuration to now allow the definition of dedicated EventLoops for PubSub:
https://github.com/open62541/open62541/blob/master/include/open62541/server_pubsub.h#L182

So can already have critical PubSub parts run in a different thread from the main server application.

For hard realtime I'm thinking about architecture-specific EventLoop implementations.
XDP on Linux doesn't use the standard POSIX API. And some embedded targets are non-POSIX as well.
So that is difficult to mix with the standard "select" that mainly drives the EventLoop.

With that the RT examples should work
- using only the public PubSub API
- allowing the configuration of multiple RT publishers and subscribers in a clean way.

But that will need a proper push to get done.

Issue #5576, opened 2023-01-17
CI build Windows (msys64) sporadically runs into timeout


keba-estr:
The CI build job Windows (msys64) sporadically runs into timeout (60 min.)

Good case:
https://dev.azure.com/open62541/open62541/_build/results?buildId=7182&view=logs&j=8ae1b360-14bf-521d-a907-7a80f3a3a1f8&t=8ae1b360-14bf-521d-a907-7a80f3a3a1f8

Timeout cases:
https://dev.azure.com/open62541/open62541/_build/results?buildId=7178&view=results
https://dev.azure.com/open62541/open62541/_build/results?buildId=7153&view=results

Overview:
https://dev.azure.com/open62541/open62541/_build?definitionId=1

The timeout should be increased to have an always running CI build.

keba-estr:
See PR #5577

keba-estr:
Reopened according to discussion in PR #5577

keba-estr:
At the moment the msys64 job runs for half an hour, so no problem here any more.

PR #5577, opened 2023-01-17
WIP: fix(ci): Increase win_msys64 job timeout to 120 minutes

keba-estr:
The job timed out sporadically with the previous timeout of 60 minutes.

### DIFF
--- a/.azure-pipelines/azure-pipelines-win.yml
+++ b/.azure-pipelines/azure-pipelines-win.yml
@@ -33,6 +33,7 @@ jobs:
     displayName: "Publish Artifact: open62541-$(CC_SHORTNAME)"
 
 - job: 'win_msys64'
+  timeoutInMinutes: 120
   displayName: 'Windows (msys64)'
   pool:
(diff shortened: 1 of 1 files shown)


jpfr:
Hmm.
I would prefer that we rework the unit tests that take a long time.
We have a few that take many seconds mostly sleeping / waiting for things to happen.

My fear is that if we increase the timeout, that will lead to more sloppy unit tests.
The speed of the CI allows faster iteration cycles. Which are fundamentally a good thing.

Do you have an insight where the msys64 job is taking particularly long?
We might split it up. For examples parallel runs with and without encryption.

keba-estr:
Unfortunately I do not have any insights in these tests, I was just facing the long run times.
Maybe the compilation time is high, too.

keba-estr:
Should we close this PR and reopen issue #5576?

jpfr:
> Should we close this PR and reopen issue https://github.com/open62541/open62541/issues/5576?

Yes, please.

PR #5583, opened 2023-01-18
refactor(plugin): Use pointer to pointer to logger for pki plugins

keba-estr:
Also the logger in the callback context must be updated when the client or server configuration is moved.

Fixes commit b0da4178d23b4bbb882da49339879a8dc0fdbd9a.

### DIFF
--- a/include/open62541/plugin/pki.h
+++ b/include/open62541/plugin/pki.h
@@ -37,11 +37,11 @@ struct UA_CertificateVerification {
     void *context;
 
     /* Verify the certificate against the configured policies and trust chain. */
-    UA_StatusCode (*verifyCertificate)(void *verificationContext,
+    UA_StatusCode (*verifyCertificate)(const UA_CertificateVerification *cv,
                                        const UA_ByteString *certificate);
 
     /* Verify that the certificate has the applicationURI in the subject name. */
-    UA_StatusCode (*verifyApplicationURI)(void *verificationContext,
+    UA_StatusCode (*verifyApplicationURI)(const UA_CertificateVerification *cv,
                                           const UA_ByteString *certificate,
                                           const UA_String *applicationURI);
 
@@ -52,7 +52,10 @@ struct UA_CertificateVerification {
     /* Delete the certificate verification context */
     void (*clear)(UA_CertificateVerification *cv);
 
-    const UA_Logger *logger;
+    /* Pointer to logging pointer in the server/client configuration.
+       If the logging pointer is changed outside of the plugin, the new
+       logger is used automatically*/
+    UA_Logger **logging;
 };
 
 _UA_END_DECLS
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -81,7 +81,9 @@ typedef struct {
  * The :ref:`tutorials` provide a good starting point for this. */
 
 struct UA_ServerConfig {
-    UA_Logger logger;
+    UA_Logger logger;   /* logger is deprecated but still supported at this time.
+                           Use logging pointer instead. */
+    UA_Logger *logging; /* If NULL and "logger" is set, make this point to "logger" */
     void *context; /* Used to attach custom data to a server config. This can
                     * then be retrieved e.g. in a callback that forwards a
                     * pointer to the server. */
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -395,8 +395,13 @@ UA_Server_newWithConfig(UA_ServerConfig *config) {
     if(server->config.eventLoop->logger == &config->logger)
         server->config.eventLoop->logger = &server->config.logger;
 
-    if(server->config.certificateVerification.logger == &config->logger)
-        server->config.certificateVerification.logger = &server->config.logger;
+    if((server->config.logging == NULL) ||
+       (server->config.logging == &config->logger)) {
+        /* re-set the logger pointer */
+        server->config.logging = &server->config.logger;
+    }
+    if(server->config.certificateVerification.logging == &config->logging)
+        server->config.certificateVerification.logging = &server->config.logging;
 
     /* Reset the old config */
     memset(config, 0, sizeof(UA_ServerConfig));
@@ -557,7 +562,7 @@ verifyServerApplicationURI(const UA_Server *server) {
         if(UA_String_equal(&sp->policyUri, &securityPolicyNoneUri) && (sp->localCertificate.length == 0))
             continue;
         UA_StatusCode retval = server->config.certificateVerification.
-            verifyApplicationURI(server->config.certificateVerification.context,
+            verifyApplicationURI(&server->config.certificateVerification,
                                  &sp->localCertificate,
                                  &server->config.applicationDescription.applicationUri);
 
--- a/src/server/ua_server_config.c
+++ b/src/server/ua_server_config.c
@@ -85,6 +85,13 @@ UA_ServerConfig_clean(UA_ServerConfig *config) {
 #endif
 
     /* Logger */
+    if(config->logging != NULL) {
+        if((config->logging != &config->logger) &&
+           (config->logging->clear != NULL)) {
+            config->logging->clear(config->logging->context);
+        }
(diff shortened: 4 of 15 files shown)


jpfr:
This is pretty messy.
I think going forward we should use a pointer to a logger instead of having it directly embedded in the config.
That way we are independent of "moving" the config.

But that I would do after branching out the next stable release.
Because it is a pretty invasive change.

keba-estr:
> This is pretty messy. I think going forward we should use a pointer to a logger instead of having it directly embedded in the config. That way we are independent of "moving" the config.
> 
> But that I would do after branching out the next stable release. Because it is a pretty invasive change.

Future refactoring of logger storage might be done in context of issue #5435

keba-estr:
> This is pretty messy. I think going forward we should use a pointer to a logger instead of having it directly embedded in the config. That way we are independent of "moving" the config.
> 
> But that I would do after branching out the next stable release. Because it is a pretty invasive change.

@jpfr: Is it still possible to integrate this PR with this release, otherwise a revert of commit b0da4178d23b4bbb882da49339879a8dc0fdbd9a must be done.

jpfr:
The plugin API is public API.
I wouldn't add temporary fixes to the public API that we don't want to keep over the long term.

My recommendation is then that we switch to using a pointer in the client and server config.
That will change the access pattern from `UA_LOG(&server->config.logger, ...)` to `UA_LOG(server->config.logger, ...)`.
But that can be mostly solved with a global regex-replace.

That will break many existing PR, but so be it.

keba-estr:
What would have to be done for a user who wants to replace the logger after calling e.g. UA_Client_new() and UA_Client_getConfig()? Would there still be a UA_Logger variable in the config where the other pointer variable config.logger and the loggers in other structs point to in order to enable a global replacement?

keba-estr:
Would it be sufficient for this release to pass the logger to the pki plugin not as a pointer, but as a pointer to a pointer? So there would exist a config.logger and config.pLogger (= &config.logger), the pki plugin would receive &config.pLogger at initialization and could dereference it for usage. In the functions where the current logger is re-set to the new config structure, only the pointer pLogger would have to be reset to &newConfig.logger and the pki plugin could use it as before. Then not all calls to the logger would have to be changed.

jpfr:
> Would there still be a UA_Logger variable in the config where the other pointer variable config.logger and the loggers in other structs point to in order to enable a global replacement?

Good idea.
So having `UA_Logger *logger` in the server config and `UA_Logger **logger` in the plugins?

keba-estr:
> > Would there still be a UA_Logger variable in the config where the other pointer variable config.logger and the loggers in other structs point to in order to enable a global replacement?
> 
> Good idea. So having `UA_Logger *logger` in the server config and `UA_Logger **logger` in the plugins?

Not quite. I would suggest having a UA_Logger logger in the server config (as before), additionaly a UA_Logger *pLogger in the server config, which is updated in the New-function, and UA_Logger **ppLogger in the plugins.
Is it ok to add "p" for pointer variables in the open62541 stack, or is there a different naming style?

keba-estr:
With this change the pki plugins automatically use an updated logger
in the server and client configuration.

keba-estr:
@jpfr: Is there a possibility to find out where the segmentation fault occurs? I can not find it with source code study.

keba-estr:
> Hey there,
> 
> the only resonable way to trace down the crash is with a debugger. But before we do that, I have thought about the problem and have a preferred solution.
> 
> In the end we should have only a pointer in the config. But the transition should be piece-by-piece and not in one big commit.
> 
> For the transition the server config looks as follows:
> 
> ```c
> struct ServerConfig {
>     // ...
>     UA_Logger logger;
>     UA_Logger *logging; /* If NULL and "logger" is set, make this point to "logger"
>     // ...
> };
> ```
> 
> During startup and _newWithConfig the `logging` pointer can be updated. And we can move the code piece-by-piece to use the `logging` pointer.
> 
> Once `logging` is used everywhere internally, we can add the `UA_DEPRECATED` flag to the old logger. And eventually remove the old `logger` structure entirely.
> 
> What are your thoughts on this?

That was also my plan. The current commit would be the first step of this.

src/server/ua_server.c
```diff
@@ -392,8 +392,8 @@ UA_Server_newWithConfig(UA_ServerConfig *config) {
     if(server->config.eventLoop->logger == &config->logger)
         server->config.eventLoop->logger = &server->config.logger;
 
-    if(server->config.certificateVerification.logger == &config->logger)
-        server->config.certificateVerification.logger = &server->config.logger;
+    /* re-set the logger pointer */
+    server->config.pLogger = &server->config.logger;
```
jpfr, line 396: Only reset the pointer if
- pLogger is NULL
- config->pLogger == &config->logger (for the original config)

That way it is already possible to use the pLogger without the "old" logger.

src/server/ua_server_config.c
```diff
@@ -89,6 +89,7 @@ UA_ServerConfig_clean(UA_ServerConfig *config) {
         config->logger.clear(config->logger.context);
     config->logger.log = NULL;
     config->logger.clear = NULL;
+    config->pLogger = NULL;
```
jpfr, line 92: Use `config->pLogger->clear` for the cleanup.
Because some users will not set the normal config->logger at all.

include/open62541/plugin/pki.h
```diff
@@ -52,7 +52,10 @@ struct UA_CertificateVerification {
     /* Delete the certificate verification context */
     void (*clear)(UA_CertificateVerification *cv);
 
-    const UA_Logger *logger;
+    /* Pointer to logger pointer
```
jpfr, line 55: Rename to `logging`. Describe as Pointer to logging pointer in the server/client configuration.

include/open62541/server.h
```diff
@@ -82,6 +82,7 @@ typedef struct {
 
 struct UA_ServerConfig {
     UA_Logger logger;
+    const UA_Logger *pLogger; /* pointer to logger for easier substitution of logger */
```
jpfr, line 85: Rename pLogger to `logging`.
We don't use the p-prefix anywhere else in the SDK.
And we should keep the API stable and not have to rename `pLogger` in the future.

Add a comment to the `logger` member that it is deprecated but still supported at this time.

jpfr (commented):
Nice. Just a few cosmetic requests.

src/server/ua_server_config.c
keba-estr: Call pLogger->clear is pLogger is not NULL and pLogger != &config->logger. config->logger.clear must still be called as pLogger might be different from &config->logger. Is that right?

jpfr: I'd say yes if pLogger != logger and logger.clear != NULL.

src/server/ua_server.c
keba-estr: Added check

src/server/ua_server_config.c
keba-estr: Added clear call and check

include/open62541/plugin/pki.h
keba-estr: renamed

include/open62541/server.h
keba-estr: renamed, added comment

jpfr:
Looks good.
Can be merged once green.
You want to get rid of the changes to the /deps/ua-nodeset submodule though.

keba-estr:
The concept with pointer to pointer does not work, as the logging in the certificateVerification still points to the old config.logging pointer that becomes invalid after copying the config structure.

keba-estr:
Added pointer to certificate structure to context instead of logger for easy update of logger pointer and use of updated logger in context functions.

keba-estr:
@jpfr: saving a pointer to the certificateVerification structure in the context does not work either, as the certificateVerification structure is also changed when copying a client config.

jpfr:
Nice! Merged.

PR #5585, opened 2023-01-18
fix(test): use correct timeout in check_client_securechannel test

keba-estr:
The maximum server timeout has been increased from 50 to 200 with
commit 384a67276457223d505c2060c3224e567ce1930d.

fix(test): fix pointer parameter in check_save_rejected_cert


jpfr:
Thanks

PR #5586, opened 2023-01-18
fix(tests): add check UA_ENABLE_SUBSCRIPTIONS_EVENTS in CMakeLists

Muddasir-Shakil:
When UA_ENABLE_SUBSCRIPTIONS is enabled and UA_ENABLE_SUBSCRIPTIONS_EVENTS is not enabled, then events related unit tests are also added to tests. This causes compilation errors due to missing declarations from Events system. Therefore a check is added in CMakeLists, to if the UA_ENABLE_SUBSCRIPTIONS_EVENTS is enabled, then add events related tests.

Signed-off-by: Muddasir Shakil <muddasir.shakil@linutronix.de>


jpfr:
👍

PR #5588, opened 2023-01-20
prevent NULL pointer assinment error in UA_ClientConnectionTCP_poll()

joerg555:
With  VS2019 i had a NULL pointer assinment error in my client.c example delivered from yours.
I had subscribed 3 variables from an S7-1500 CPU with OPC-UA and it was working fine. 
I have droped the ethernet cable many times to watch reconnect, and than i got this error, and fixed it.
I cant find this in master and 1.3 branch.
The reconnect dose not work, in your example, but that's no proplem here.

Regards Jörg


jpfr:
Thanks. Please sign the CLA also.

jpfr:
Cool, merged.

PR #5593, opened 2023-01-25
fix(pubsub): fixed string field length calculations for RawData DatasetMessages

MichalTyndel:
We can't use the size of UA_String structure for field length calculation in dataSetMessages in form of rawData. It contains pointer to allocated memory, so the size is the same for all strings. Also the size of UA_String varies between platforms, since sizeof(size_t) and sizeof(byte*) are 4 bytes on 32-bit systems and 8 bytes on 64-bit systems. Length of binary encoded string = 4 (string length encoded as uint32) + N bytes (actual string data). For fixed message layout N equals maxStringlength.

### DIFF
--- a/src/pubsub/ua_pubsub_reader.c
+++ b/src/pubsub/ua_pubsub_reader.c
@@ -801,7 +801,19 @@ DataSetReader_processRaw(UA_Server *server, UA_ReaderGroup *rg,
         const UA_DataType *type =
             UA_findDataTypeWithCustom(&dsr->config.dataSetMetaData.fields[i].dataType,
                                       server->config.customDataTypes);
-        msg->data.keyFrameData.rawFields.length += type->memSize;
+        size_t fieldLength = 0;
+        if(type->typeKind == UA_DATATYPEKIND_STRING || type->typeKind == UA_DATATYPEKIND_BYTESTRING)
+        {
+            UA_assert(dsr->config.dataSetMetaData.fields[i].maxStringLength != 0);
+            // Length of binary encoded string = 4 (string length encoded as uint32) + N (actual string data) bytes.
+            // For fixed message layout N equals maxStringlength.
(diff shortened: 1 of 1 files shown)


jpfr:
Looks good.
However we have to handle the case where maxStringLength == 0.
We could then decode the int32 string length to move the position.

Is it ensure that we use no other "complex" types other than string?

Fischer-Thomas:
> However we have to handle the case where maxStringLength == 0.

I think this should be handled when setting up a DataSetReader and not when processing a NWM.
When using RawData encoding you need to have a MaxStringLength for String/ByteString/LocalizedText (as written in Part14).
MaxStringLength == 0 in the DataSetMetaData is only allowed for DataTypes that are not String/ByteString/LocalizedText or when the maximum is unknown (which is not allowed for RawData-Encoding).

src/pubsub/ua_pubsub_reader.c
```diff
@@ -778,7 +801,19 @@ DataSetReader_processRaw(UA_Server *server, UA_ReaderGroup *rg,
         const UA_DataType *type =
             UA_findDataTypeWithCustom(&dsr->config.dataSetMetaData.fields[i].dataType,
                                       server->config.customDataTypes);
-        msg->data.keyFrameData.rawFields.length += type->memSize;
+        size_t fieldLength = 0;
+        if((type->typeKind == UA_DATATYPEKIND_STRING || type->typeKind == UA_DATATYPEKIND_BYTESTRING) 
+            && (dsr->config.dataSetMetaData.fields[i].maxStringLength != 0))
+        {
+            // Length of binary encoded string = 4 (string length encoded as uint32) + N (actual string data) bytes.
```
jpfr, line 808: Add an assertion here that the max string length is always set for strings.
Because we do now check this in the config.

jpfr:
Nice, thanks.

PR #5595, opened 2023-01-25
fix(pubsub): fixed program crashing when sending too long strings

MichalTyndel:
Trying to send too long string would result in program crashing:  when the string length exceeded maxStringLength, lengthDifference would underflow and we would try to zero out gigabytes of memory. Added check to avoid that.

Would that be the correct approach? to just return with error and try sending the message later?
Is this the correct statuscode to return in such case?

### DIFF
--- a/src/pubsub/ua_pubsub_networkmessage_binary.c
+++ b/src/pubsub/ua_pubsub_networkmessage_binary.c
@@ -1404,6 +1404,9 @@ UA_DataSetMessage_keyFrame_encodeBinary(const UA_DataSetMessage* src, UA_Byte **
                 if(fmd->maxStringLength != 0 &&
                    (v->value.type->typeKind == UA_DATATYPEKIND_STRING ||
                     v->value.type->typeKind == UA_DATATYPEKIND_BYTESTRING)) {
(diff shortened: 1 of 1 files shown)


jpfr:
@andreasebner please have a look

andreasebner:
The mentioned error-case should definitely be handled. The standard itself is not very precise on how the handling should look like. In my opinion, there are 2 options:
1. Check length and return with an error code (this PR).
2. Encode the string up to the defined maxStringLength without an error. (e.g. maxStringLength = 3; content "hello" --> encode "hel").
@jpfr  Do you prefer one of these solutions? 
@MichalTyndel  I think your return code is fine. Another option would be UA_STATUSCODE_BADOUTOFRANGE.

jpfr:
returning an error case is preferred.

MichalTyndel:
Adjusted the code to work with new master.
Error is returned as discussed. This disables writerGroup, but it's better than crashing the program.
This is out of scope of this PR, but IMO it should be considered how to handle such errors without stopping the writer(or writerGroup).

jpfr:
Merged, but thanks

PR #5596, opened 2023-01-25
prevents NULL pointer assinment error in UA_ClientConnectionTCP_poll()

joerg555:
Hi this is same issue from branch 1.2 i have checkst it an ein
Jörg


jpfr:
Thanks I created a PR #5597 that merges 1.2 to 1.3.
The result is the same. In addition the merge makes visible that all fixes on 1.2 are on 1.3 as well.

PR #5599, opened 2023-01-26
[REVIEW] fix(pubsub): enable remove and add readergroups

JanSurft:
Removing and adding readergroups with pubsub lead to errors when the readergroups were set operational before. This fix opens and closes the communication channel for readergroups that are bound to a connection in the scope of the connection.

Thanks @MurzynJ for the discovery and possible fix of that error.


JanSurft:
@MurzynJ I added a test for the supposedly failing setup, but when I rand this test on the original implementation there was no failure. Did you do something differently on your test setup?

PR #5607, opened 2023-01-31
fixe(server) samplingInterval when negative number

StalderT:
According to https://reference.opcfoundation.org/Core/Part4/v104/docs/5.12.1.2, a negative Sampling interval number indicates that the default sampling interval defined by the publishing interval of the Subscription is requested.

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -11,7 +11,7 @@
  *    Copyright 2018 (c) Ari Breitkreuz, fortiss GmbH
  *    Copyright 2017 (c) Mattias Bornhager
  *    Copyright 2017 (c) Henrik Norrman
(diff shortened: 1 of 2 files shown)


StalderT:
Hi @jpfr,

Happy to discuss with you again.

Good idea. Maybe for the master branch ?

Today I have a client (PROSYS) that sends -1 as samplingInterval. The opc server does not return any change value for these monitoreditems. Maybe there is another problem with samplingInterval = 0.

I think it would be nice to merge this patch into version 1.3.

jpfr:
=0 has a different meaning for us.
If the server allows 0, then we attach the MonitoredItems to the node and use every "write" as the trigger.
Like for an event.

jpfr:
Extended in #5608.
Please have a look and comment.
Closing here.

PR #5608, opened 2023-01-31
Negative sampling interval

jpfr:


StalderT:
tested and works fine for me.

thank you very much for the great work

StalderT:
@jpfr 

I did additional tests with my server stress and I monitor 10,000 variables with a client. 

Performances are greatly reduced with this new patch. 

Indeed if a publish response cannot contain the 10,000 variables then UA_Subscription_publish is called back and then the 10,000 variables are ressampled in permanently, therefore much more often than PublishingInterval.

In the state it might be good to go back on this commit.

StalderT:
Or make sure to call:

```
    LIST_FOREACH(mon, &sub->samplingMonitoredItems, sampling.samplingListEntry) {
        monitoredItem_sampleCallback(server, mon);
    }
```

Only if the last call is older than PublishingInterval.

PR #5609, opened 2023-02-01
fix(core): Initialize entire nodeid variable

keba-estr:
If entire Nodeid variable was used, valgrind reported uninitialized memory.

### DIFF
--- a/include/open62541/types.h
+++ b/include/open62541/types.h
@@ -434,6 +434,7 @@ UA_INLINABLE(UA_NodeId
              UA_NODEID_NUMERIC(UA_UInt16 nsIndex,
                                UA_UInt32 identifier), {
     UA_NodeId id;
+    memset(&id, 0, sizeof(UA_NodeId));
     id.namespaceIndex = nsIndex;
     id.identifierType = UA_NODEIDTYPE_NUMERIC;
     id.identifier.numeric = identifier;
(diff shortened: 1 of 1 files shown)


include/open62541/types.h
```diff
@@ -424,6 +424,7 @@ UA_NodeId_parse(UA_NodeId *id, const UA_String str);
 UA_INLINABLE(UA_NodeId
              UA_NODEID(const char *chars), {
     UA_NodeId id;
+    memset(&id, 0, sizeof(UA_NodeId));
```
jpfr, line 427: This one is good.
But I'd rather move it into UA_NodeId_parse.
So it comes automatic in more cases.

jpfr (commented):
Looks good.
Just one small request.

include/open62541/types.h
keba-estr: @jpfr: In the function parse_nodeid the id is already set to UA_NODEID_NULL, so this might already be enough initalization for the function UA_NodeId_parse?

jpfr: You are right. That is enough.

keba-estr: Should I remove the memset in line 427?

jpfr: Yes.

keba-estr: Removed memset in line 427.

jpfr:
👍

Issue #5611, opened 2023-02-01
BUG: Client: Browsing or subscribing failed with latest ZipTree updates


GoetzGoerisch:
## Description

The latest ZipTree fix [a85bcee](https://github.com/open62541/open62541/pull/5580/commits/a85bcee25827707f4f2f23123b6cc14dd596bd3f) breaks our existing client implementation.

Error:
```
[DashboardOpcUaClient] INFO bool Umati::MachineObserver::MachineObserver::canBrowseMachineList(std::__cxx11::list<ModelOpcUa::BrowseResult_t>&):129 Searching for machines
Assertion failed: !UA_TimerTree_ZIP_MIN(&t->tree) || UA_TimerTree_ZIP_MIN(&t->tree)->nextTime > nowMonotonic (/src/DashboardOpcUaClient/deps/open62541/arch/common/ua_timer.c: UA_Timer_process: 261)
```
while the client starts browsing for machines.
Sometimes it only breaks later while starting to subscribe to nodes.

## Background Information / Reproduction Steps

Build container image: 
https://github.com/umati/Dashboard-OPCUA-Client/commit/dc5f2e3f1c7c395c8c3b8783467f6285d36b4eec

Use `configuration.json.example` and set the OPC UA endpoint to

```
  "OpcUa": {
    "Endpoint": "opc.tcp://opcua.umati.app:4840",
    "Username": "",
    "Password": "",
    "Security": 1
  },
```

and a available MQTT broker.
then run with
`docker run -it --rm -v /path/to/configuration.json:/app/configuration.json dashboardopcuaclient`

Used CMake options:

```bash
cmake -DUA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS:BOOL=ON -DUA_ENABLE_SUBSCRIPTIONS_EVENTS:BOOL=ON
               -DUA_NAMESPACE_ZERO:STRING=FULL -DUA_ENABLE_ENCRYPTION:STRING=LIBRESSL
               -DLIBRESSL_ROOT_DIR:PATH=${CMAKE_INSTALL_PREFIX} -DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_INSTALL_PREFIX}
```

## Checklist

Please provide the following information:

 - [X] open62541 Version (release number or git tag): [f70693b](https://github.com/open62541/open62541/tree/f70693b08d097d2a49354903034e2a2029adc821)
 - [ ] Other OPC UA SDKs used (client or server): 
[...]

PR #5618, opened 2023-02-06
[REVIEW] fix memory leaks in SKS implementation

Muddasir-Shakil:
After running all SKS unit tests with UA_ENABLE_UNIT_TESTS_MEMCHECK (valgrind), the memory issues were identified in the test and implementation code. This PR includes the patches with fixes and improvements to fix all the memory issues identified with UA_ENABLE_UNIT_TESTS_MEMCHECK.


src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -509,6 +509,7 @@ sksClientCleanupCb(void *client, void *context) {
         sksClient->config.securityPoliciesSize = 0;
         sksClient->config.certificateVerification.context = NULL;
         sksClient->config.logger.context = NULL;
+        UA_free(sksClient->config.clientContext);
```
jpfr, line 512: The client context lifecycle should be controlled by the userland.
It could even be stack-allocated. And/or contain dynamically allocated memory.

Wouldn't it be possible that the shutdown of the SKS system is signaled in the configured callback so that user can handle the context lifecycle?

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -563,6 +564,12 @@ storeFetchedKeys(UA_Client *client, void *userdata, UA_UInt32 requestId,
             goto cleanup;
     }
 
+    if(ks->callBackId != 0) {
```
jpfr, line 567: Add a short comment.
It is surprising that the callback is removed from within `storeFetchedKeys`.
When and where will it be re-added?

We could also pull this out rather into the method calling `storeFetchedKeys`.

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -592,26 +600,30 @@ callGetSecurityKeysMethod(UA_Client *client) {
     sksClientContext *ctx = (sksClientContext *)client->config.clientContext;
 
     UA_Variant *inputArguments = (UA_Variant *)UA_calloc(3, (sizeof(UA_Variant)));
```
jpfr, line 602: As this is an array of known size, allocate it statically.
So you also do not have to check for malloc failing.

jpfr (commented):
Great! Comments inline.
Please add this to the CI with memcheck enabled.
So more problems are caught automatically.

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: In this case, the client context is used to store the data to call getSecurityKeys, in the callbacks when sks async connect is successful. It is created, used and destroyed only internally by the SKS client. So user will/shall not have access to it.

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: > It is surprising that the callback is removed from within `storeFetchedKeys`.
> When and where will it be re-added?

The previous callback is removed from the queue and new callback is added with a new keylifetime. I added this information in the comments now.

src/pubsub/ua_pubsub_keystorage.c
Muddasir-Shakil: > 



> Please add this to the CI with memcheck enabled.

> Great! Comments inline. Please add this to the CI with memcheck enabled. So more problems are caught automatically.

unfortunately the sks_client test is also failing due to the similar issue as mentioned in #5611  & #5634 . That is why I am unable to add memcheck to the CI at the moment. May be when this is fixed. Can I add this in a new PR when #5611 is resolved?

`72: check_pubsub_sks_client: /home/muddasir/work/Projects/DronEdge/sks_open62541/open62541/arch/common/ua_timer.c:261: UA_Timer_process: Assertion `!UA_TimerTree_ZIP_MIN(&t->tree) || UA_TimerTree_ZIP_MIN(&t->tree)->nextTime > nowMonotonic' failed.
72: Aborted`

Muddasir-Shakil:
@jpfr license/cla tests does not start. Do you have any idea about this?

GoetzGoerisch:
Sometimes the CLA Assistant is stuck. Fix is to `Recheck PRs` in the cla-assistant.io app from the project owner.

jpfr:
Done.
Also rebased on the current master.
Merging when green.

PR #5623, opened 2023-02-09
fix(pubsub): fix issues with removing a DataSetField

MurzynJ:


jpfr:
Nice, that's a lot better. 👍

PR #5624, opened 2023-02-09
refactor(pubsub): optimize removal of a PublishedDataSet

MurzynJ:
Before this change, removeDataSetField was called in a loop for all DataSet fields. Inside removeDataSetField the DataSetMetadata is regenerated again and again and that's not necessary when you want to remove the whole DataSet anyway.


jpfr:
Makes sense.
Please add your explanation from the PR in a comment.
Otherwise another developer might want to reduce code-duplication in the future and reverses the change.

MurzynJ:
> Makes sense. Please add your explanation from the PR in a comment. Otherwise another developer might want to reduce code-duplication in the future and reverses the change.

Done.

jpfr:
Nice, merged.

PR #5625, opened 2023-02-09
fix(pubsub): set a correct builtInTypeId in FieldMetaData

MurzynJ:
By mistake the internal typeKind was set instead of a numeric identifier of the type's NodeId.

In future the special cases should be handled as specified:

image


MurzynJ:
It seems that the failing test has nothing to do with this change. @jpfr can you check it?

jpfr:
Thanks, good find.

PR #5626, opened 2023-02-09
refactor(pubsub): don't clear TargetVariables twice

MurzynJ:
TargetVariables structure was cleared twice unnecessarily. It's cleared for the first time in UA_DataSetReaderConfig_clear just above.


jpfr:
Nice one

PR #5629, opened 2023-02-10
fix(pubsub): prevent segfault on reception of a broken UADP message

MurzynJ:
UDP subscriber crashed when it received the following message.

image

Without discussing if this is even a valid message, such crash shall not happen in any case. How can it be checked in the best way that stack is resistant to any malformed message?


jpfr:
Thanks for this.

PR #5636, opened 2023-02-13
[WIP]feat: Implementation of role permission for well known roles

opcua-tsn-team-kalycito:


include/open62541/plugin/accesscontrol.h
```diff
@@ -16,6 +16,34 @@ _UA_BEGIN_DECLS
...
+typedef enum {
+    UA_ANONYMOUS_WELL_KNOWN_RULE = 0,
+    UA_AUTHENTICATEDUSER_WELL_KNOWN_RULE = 1,
+    UA_CONFIGUREADMIN_WELL_KNOWN_RULE = 2,
+    UA_ENGINEER_WELL_KNOWN_RULE = 3,
+    UA_OBSERVER_WELL_KNOWN_RULE = 4,
+    UA_OPERATOR_WELL_KNOWN_RULE = 5,
+    UA_SECURITYADMIN_WELL_KNOWN_RULE = 6,
+    UA_SUPERVISOR_WELL_KNOWN_RULE = 7,
+}UA_AccessControlGroup;
```
jpfr, line 30: 1. Please include all well known roles: https://reference.opcfoundation.org/Core/Part3/v105/docs/4.8
2. The following naming-scheme is preferred: `UA_ROLE_ANONYMOUS` and similar.
3. AccessControlGroup is a funky name for OPC UA Roles. Rename this to `UA_Role`.
4. Add a (const) array of NodeId called `UA_ROLES` where the Role-index maps to the NodeId of the role.
    Note that users can add their own roles. These will not show up in the enum or the NodeId array.
    So the Roles-Enum is only a shortcut. We should always use the NodeId role identifier to allow the extension with custom roles going forward.

include/open62541/plugin/accesscontrol.h
```diff
@@ -127,6 +164,13 @@ struct UA_AccessControl {
                                                       UA_DateTime endTimestamp,
                                                       bool isDeleteModified);
 #endif
+    UA_Boolean (*checkUserDatabase)(const UA_UserNameIdentityToken *userToken, UA_String *roleName);
+
+#ifdef UA_ENABLE_ROLE_PERMISSION
```
jpfr, line 169: The roles are attached to a Session.
It makes sense to return the roles from `activateSession` (also in the AccessControl plugin) and store the returned roles in the session. Looking up the roles (e.g. from an ActiveDirectory service) may take a lot of time.
So caching the roles in the Session is a must requirement.

We have an API for attaching attributes to a session.
https://github.com/open62541/open62541/blob/73a60b501094876d637c78878e0c90da34b42cf7/include/open62541/server.h#L524
Use the key "0:roles" for storing the roles in the session.

With that the following methods are not needed:
- checkUserDatabase
- setRoleAccessPermission
- readUserDefinedRolePermission
- checkTheRoleSessionLoggedIn

include/open62541/plugin/nodestore.h
```diff
@@ -411,6 +411,12 @@ struct UA_NodeHead {
     UA_MonitoredItem *monitoredItems; /* MonitoredItems for Events and immediate
                                        * DataChanges (no sampling interval). */
 #endif
+#ifdef UA_ENABLE_ROLE_PERMISSION
+    size_t rolePermissionsSize;
+    UA_RolePermissionType *rolePermissions;
+    size_t userRolePermissionsSize;
```
jpfr, line 417: The user roles are not static but depend on the session that is reading the node.
Do not store the userRolePermissions as an explicit field in the nodes.

Rather modify the "Read" service to compute this value at runtime.
Just return all RolePermissions where both the role is defined for the node **and** the current session has that role.

include/open62541/plugin/accesscontrol.h
```diff
@@ -80,27 +108,36 @@ struct UA_AccessControl {
...
                                   const UA_NodeId *sessionId, void *sessionContext,
-                                  const UA_NodeId *nodeId, void *nodeContext);
+                                  const UA_NodeId *nodeId, void *nodeContext, UA_RolePermissionType *userRolePermission,size_t userRoleSize);
+
+    /* Check access to Node */
+    UA_Boolean (*hasAccessToNode)(UA_Server *server, UA_AccessControl *ac,
+                                  const UA_NodeId *sessionId, void *sessionContext,
+                                  const UA_NodeId *nodeId, void *nodeContext, UA_Byte* serviceAccessLevel);
+    /* Check access to Node */
+    UA_Boolean (*hasAccessToMethod)(UA_Server *server, UA_AccessControl *ac,
```
jpfr, line 138: Is there a difference between `hasAccessToMethod` and `getUserExecutable`?
This seems to be a duplicate.

include/open62541/plugin/accesscontrol.h
```diff
@@ -80,27 +108,36 @@ struct UA_AccessControl {
...
+                                       const UA_DeleteReferencesItem *item, UA_RolePermissionType *userRolePermission, size_t userRoleSize);
 
     /* Allow browsing a node */
     UA_Boolean (*allowBrowseNode)(UA_Server *server, UA_AccessControl *ac,
                                   const UA_NodeId *sessionId, void *sessionContext,
-                                  const UA_NodeId *nodeId, void *nodeContext);
+                                  const UA_NodeId *nodeId, void *nodeContext, UA_RolePermissionType *userRolePermission,size_t userRoleSize);
+
+    /* Check access to Node */
+    UA_Boolean (*hasAccessToNode)(UA_Server *server, UA_AccessControl *ac,
```
jpfr, line 134: What is the difference between `hasAccessToNode` and `getUserRightsMask`?
This seems to be a duplicate.

include/open62541/plugin/accesscontrol.h
```diff
@@ -80,27 +108,36 @@ struct UA_AccessControl {
     /* Allow adding a node */
     UA_Boolean (*allowAddNode)(UA_Server *server, UA_AccessControl *ac,
                                const UA_NodeId *sessionId, void *sessionContext,
-                               const UA_AddNodesItem *item);
+                               const UA_AddNodesItem *item, UA_RolePermissionType *userRolePermission, size_t userRoleSize);
```
jpfr, line 111: Don't modify the API for allowAddNode, etc.

I see the RolePermissions as an alternative way for controlling the access.
If there are RolePermissions defined for a node (and not all nodes will have them), then we check the RolePermissions internally.

No need to query the AccessControl plugin for this.
Because the server internally has access to the RolePermissions of the node and the roles configured for each session.

If however there are no RolePermissions defined for a node, then we continue to query to AccessControl plugin as before.

include/open62541/plugin/nodestore.h
```diff
@@ -606,6 +612,7 @@ typedef struct {
     UA_NodeHead head;
     UA_NODE_VARIABLEATTRIBUTES
     UA_Byte accessLevel;
+    UA_Byte userAccessLevel;
```
jpfr, line 615: The UserAccessLevel is computed dynamically for each session. Don't store it in the node.

include/open62541/plugin/nodestore.h
```diff
@@ -645,6 +652,7 @@ typedef UA_StatusCode
 typedef struct {
     UA_NodeHead head;
     UA_Boolean executable;
+    UA_Boolean userExecutable;
```
jpfr, line 655: The UserExecutable attribute is computed dynamically for each session. Don't store it in the node.

include/open62541/server.h
```diff
@@ -711,6 +711,13 @@ UA_Server_readAccessLevel(UA_Server *server, const UA_NodeId nodeId,
                             outAccessLevel);
 }
 
+static UA_INLINE UA_THREADSAFE UA_StatusCode
+UA_Server_readUserAccessLevel(UA_Server *server, const UA_NodeId nodeId,
```
jpfr, line 715: This method does not make any sense.
Because the UserAccessLevel depends on the session.
But if you locally read the UserAccessLevel then you do it outside the context of a session.
You would get back the exact same result as for `UA_Server_readAccessLevel`.

I would prefer if you left a comment why there is no `UA_Server_readUserAccessLevel`.

include/open62541/server.h
```diff
@@ -733,6 +740,27 @@ UA_Server_readExecutable(UA_Server *server, const UA_NodeId nodeId,
                             outExecutable);
 }
 
+static UA_INLINE UA_THREADSAFE UA_StatusCode
+UA_Server_readUserExecutable(UA_Server *server, const UA_NodeId nodeId,
```
jpfr, line 744: See the comment for `UA_Server_readUserAccessLevel`.

include/open62541/server.h
```diff
@@ -733,6 +740,27 @@ UA_Server_readExecutable(UA_Server *server, const UA_NodeId nodeId,
...
+
+static UA_INLINE UA_THREADSAFE UA_StatusCode
+UA_Server_readRolePermissions(UA_Server *server, const UA_NodeId nodeId,
+                              UA_Variant *rolePermissions) {
+    return __UA_Server_read(server, &nodeId, UA_ATTRIBUTEID_ROLEPERMISSIONS,
+                            rolePermissions);
+}
+
+static UA_INLINE UA_THREADSAFE UA_StatusCode
+UA_Server_readUserRolePermissions(UA_Server *server, const UA_NodeId nodeId,
```
jpfr, line 758: See the comment for `UA_Server_readUserAccessLevel`.

include/open62541/server.h
```diff
@@ -894,6 +922,19 @@ UA_Server_writeExecutable(UA_Server *server, const UA_NodeId nodeId,
...
 
+static UA_INLINE UA_THREADSAFE UA_StatusCode
+UA_Server_writeRolePermissions(UA_Server *server, const UA_NodeId nodeId,
+                               const UA_Variant rolePermissions) {
+    return __UA_Server_write(server, &nodeId, UA_ATTRIBUTEID_ROLEPERMISSIONS,
+                             &UA_TYPES[UA_TYPES_VARIANT], &rolePermissions);
+}
+
+static UA_INLINE UA_THREADSAFE UA_StatusCode
+UA_Server_writeUserRolePermissions(UA_Server *server, const UA_NodeId nodeId,
```
jpfr, line 933: See the comment for `UA_Server_readUserAccessLevel`.

include/open62541/server.h
```diff
@@ -1974,4 +2015,13 @@ _UA_END_DECLS
 #include <open62541/server_pubsub.h>
 #endif
 
+#ifdef UA_ENABLE_ROLE_PERMISSION
+UA_EXPORT UA_StatusCode
+setUserRole_settings(UA_Server* server, UA_String roleName,
```
jpfr, line 2020: 1. I don't understand what these methods are doing just from their naming.
    So other users won't either.
3. Use the UA_ prefix for all public API.
4. Roles are identified by a NodeId, not by a string.

jpfr (commented):
Nice!
This is a welcome addition.

I just reviewed the public API only.
Will look at the implementation code once the public API is a go.

The changes to the client are looking all good (besides small cosmetic comments).
If you pull the client changes into an extra PR we can merge them quickly.

andreasebner:
Thank's for shareing this ideas. A RBAC implementation is part of the latest master.

PR #5654, opened 2023-02-20
examples: CMakeLists.txt: fix multithreading check

JohannesKauffmann:
UA_MULTITHREADING can be greater than 100, in which case these examples won't be built.


jpfr:
Nice thx!

Issue #5614, opened 2023-02-01
feat(plugin):  plugin design to move plugin objects


Muddasir-Shakil:
hello everyone,
This issue is created to start the discussion on topic of making the deep copy/moving plugin objects. Recently UA_ClientConfig_copy is added to upstream, which makes  a partial deep copy of it members. The problem is caused by the plugins pointers such as security pointers, eventloop, loggers etc. The implementation of these plugins contains internal objects whose structs are not exposed outside of the plugin implementation, which makes it difficult (impossible) to make the deep copy of such objects also including serverconfig. Therefore, some rethinking is required to fix the issues in the plugins, so that they take care of making deep copies of their internally contained objects. There are various approaches can be taken to solve the above mentioned problem. I would like to suggest the following solutions and invite you to propose something else or comment on which approach is better. We can then create a PR to fix this issues in our plugins.

**suggestion**

1. Copy/move the pointers and keep record of reference counting technique inside the plugin implementation [(reference counting technique)](https://mortoray.com/what-is-reference-counting/)
2. Marking the object of the plugin by a flag as external, when moved/copied as done in the Eventloop implementation (ExternalEventloop)
3. Adding a new "copy" function to the Plugin API. The copy function takes care of the making the deep copy of the internal objects to the destination.

I suggest to use solution 1. The benefit of it as single object can be linked to multiple objects which means less complex logic to allocate/deallocate the memory. if the more number of copies are required which can be simply created without allocating any new memory, rather reuse the pointer to the original object.

Looking forward to your suggestions.
BR
Muddasir Shakil

jpfr:
I prefer option 2.
It allows more „manual control“ that is deterministic on when a plug-in is cleaned up.

In nearly all cases the plugins will „live“ for the duration of the entire application.
Either because they „live“ in a master-server/client.
Or because they are initialized in a global variable from which runtime-server/client-instances can access them during initialization.
And when the plug-ins are held in a global variable then it is easy to clean them up right before the application shuts down.

jpfr:
Option 2 also allows the re-linking you describe.
But there is one master-pointer that has a lifecycle longer than the other users.

Any other arguments we forgot?

Muddasir-Shakil:
the problem with option 2 is that the when we make a copy of the object the copied object can survive until the original survives. It will cause issues if the user decides to free the original object at runtime. If we also mark original object as external then it is difficult to differentiate between original and copied object when object is being cleaned up. 
[...]

PR #5672, opened 2023-03-07
[WIP] feat(plugin):  plugin design to move plugin objects

Muddasir-Shakil:
This PR contains targets the issues described in #5614. As discussed in the thread, plugins are added marked by a flag. This flag is used to inform the plugin user that the plugin object is shared or it has multiple owners.

### DIFF
--- a/include/open62541/client.h
+++ b/include/open62541/client.h
@@ -149,6 +149,7 @@ typedef struct {
     /* Available SecurityPolicies */
     size_t securityPoliciesSize;
     UA_SecurityPolicy *securityPolicies;
+    UA_Boolean externalPolicy; /* The securityPolicies are not deleted with the config */
 
     /* Certificate Verification Plugin */
     UA_CertificateVerification certificateVerification;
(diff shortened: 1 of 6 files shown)


Muddasir-Shakil:
@jpfr I opened this PR to check with you if, we both had same understanding. if you think the approach is right then I will continue to solve for other plugins.

include/open62541/client.h
```diff
@@ -149,6 +149,7 @@ typedef struct {
     /* Available SecurityPolicies */
     size_t securityPoliciesSize;
     UA_SecurityPolicy *securityPolicies;
+    UA_Boolean externalPolicy; /* The securityPolicies are not deleted with the config */
```
jpfr, line 152: -> externalPolicies

include/open62541/plugin/log.h
```diff
@@ -63,6 +63,8 @@ typedef struct {
     void *context; /* Logger state */
 
     void (*clear)(void *context); /* Clean up the logger plugin */
+
+    bool externalLogger; /* The logger is not deleted with the config */
```
jpfr, line 67: Don't add the flag to the plugin itself.
Put it next to the plugin in the config.

jpfr (commented):
Looks good!
Detailed comments inline.

You can make several PR out of the overall push.
Whenever you have finished some (coherent set of) plugins.

Looking forward to this. 👍

jpfr:
Converting to draft because of the WIP title.

PR #5673, opened 2023-03-08
fix(pubsub): fix memory issue after reading publishedData property

matkonnerth:
Copy nodeId when reading UA_NS0ID_PUBLISHEDDATAITEMSTYPE_PUBLISHEDDATA.


jpfr:
Thanks

PR #5682, opened 2023-03-14
fix(nodeset-loader): Minor bugfixes

MurzynJ:


MurzynJ:
@keba-uso
UA_Server_loadNodeset is declared to return UA_StatusCode, so I changed the implementation to return UA_STATUSCODE_GOOD or BAD.

But in all of the tests you treat the result as bool, so they fail now. Shall I go on and update the tests?

keba-uso:
> @keba-uso UA_Server_loadNodeset is declared to return UA_StatusCode, so I changed the implementation to return UA_STATUSCODE_GOOD or BAD.
> 
> But in all of the tests you treat the result as bool, so they fail now. Shall I go on and update the tests?

@MurzynJ since the function returns `UA_StatusCode`, it is much cleaner to apply this fix and update the tests. Thank you!

jpfr:
Thanks for this.
Suddely we have failing tests. E.g. check_nodeset_loader_di.

MurzynJ:
> Thanks for this. Suddely we have failing tests. E.g. check_nodeset_loader_di.

Yes, I knew it, but wanted first to confirm with Srdjan if it's okay to update that many tests.
Now the tests are updated and the CI is green.

jpfr:
Ok, merged.

PR #5683, opened 2023-03-14
[REVIEW] Integrate libwebsockets library

NoelGraf:
This PR integrates libwebsockets library into the open62541 stack. The integration is done via the adapter in arch/lws. This adapter acts as a custom "event library" interface layer between the lws event lib support and the open62541 event loop. By integrating in this way, the libwebsockets library makes use of the open62541 event loop, which allows for a perfect integration into the stack.


.gitmodules
```diff
@@ -13,3 +13,6 @@
 	path = deps/nodesetLoader
 	url = https://github.com/open62541/nodesetLoader
 	branch = master
+[submodule "deps/libwebsockets"]
```
jpfr, line 16: Don't use a git submodule.
99% of users should install libwebsockets via a package manager.
And the remaining 1% know how to do it themselves.

CMakeLists.txt
```diff
@@ -689,14 +701,35 @@ if(UA_ENABLE_TPM2_SECURITY)
...
+if(UA_ENABLE_LWS)
     find_package(libwebsockets REQUIRED)
-    list(APPEND open62541_LIBRARIES ${LIBWEBSOCKETS_LIBRARIES})
+    if(LIBWEBSOCKETS_FOUND)
+        if(${LIBWEBSOCKETS_VERSION} VERSION_LESS "4.3")
+            message(STATUS "The version of libwebsockets is too low (required v4.3). Instead, the library is used in the dependencies.")
+            set(LIBWEBSOCKETS_FOUND 0)
+        endif()
+    endif()
+    if(NOT LIBWEBSOCKETS_FOUND)
```
jpfr, line 718: Throw an exception when not found.
The `find_package` thing allows users to manually define the folders.
That should be the only way of doing things.
Like for OpenSSL/mbedTLS.

CMakeLists.txt
```diff
@@ -1596,6 +1629,13 @@ endif()
 
 add_dependencies(open62541-object open62541-code-generation)
 
+#Libwebsockets
+if(UA_ENABLE_LWS)
+    if(NOT LIBWEBSOCKETS_FOUND)
+        add_dependencies(open62541-object websockets_shared)
```
jpfr, line 1635: Not needed. LIBWEBSOCKETS_FOUND has to be true.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+*
+*   Copyright 2023 (c) Fraunhofer IOSB (Author: Noel Graf)
+*/
+
+/* The HTTP ConnectionManager is architecture agnostic. It selects an existing
+* TCP ConnectionManager and uses it for the underlying TCP connections. */
+
+/* Suppress the warning 'redefinition of typedef' in the libwebsockets library.
+ * Removed as soon as the patch is included in the main. */
+#pragma GCC diagnostic ignored "-Wpedantic"
```
jpfr, line 13: Which patch included in the main?
The main of libwebsockets?
Refer to the patch so the status can be tracked.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+typedef struct HTTPConnectionManager HTTPConnectionManager;
+
+#define HTTP_PARAMETERSSIZE 5
+
+static const struct {
+    UA_QualifiedName name;
+    const UA_DataType *type;
+    UA_Boolean required;
+} HTTPSendParameters[HTTP_PARAMETERSSIZE] = {
+    {{0, UA_STRING_STATIC("url")}, &UA_TYPES[UA_TYPES_STRING], true},
```
jpfr, line 30: You mean address instead of url.
The url would include the port and path as well.
How would this be extended for https?
You can already add this as a comment.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+        /* because we are protocols[0] ... */
+    case LWS_CALLBACK_CLIENT_CONNECTION_ERROR:
+        lwsl_err("CLIENT_CONNECTION_ERROR: %s\n",
+                 in ? (char *)in : "(null)");
+        break;
+
+        /* chunks of chunked content, with header removed */
+    case LWS_CALLBACK_RECEIVE_CLIENT_HTTP_READ:
+        lwsl_user("RECEIVE_CLIENT_HTTP_READ: read %d\n", (int)len);
+        lwsl_hexdump_info(in, len);
```
jpfr, line 56: The hexdump seems to be just for debugging.
Remove if not needed.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+        UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_NETWORK,
```
jpfr, line 57: The LOG_INFO should be a LOG_DEBUG.
The message seems to be quite frequent.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+    {{0, UA_STRING_STATIC("data")}, &UA_TYPES[UA_TYPES_BYTESTRING], false}, // data for post request
+};
+
+static int
+callback_http(struct lws *wsi, enum lws_callback_reasons reason,
+              void *user, void *in, size_t len) {
+    switch (reason) {
+
+    case LWS_CALLBACK_ESTABLISHED_CLIENT_HTTP:
+        lwsl_user("LWS_CALLBACK_ESTABLISHED_CLIENT_HTTP: resp %u\n",
```
jpfr, line 43: Can we make lwsl to use our logging facilities?
Some people are configuring open62541 to log into a file or dedicated logging service.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+static void
+HTTP_eventSourceStop(UA_ConnectionManager *cm) {
+    if(cm->eventSource.state == UA_EVENTSOURCESTATE_STOPPING ||
+       cm->eventSource.state == UA_EVENTSOURCESTATE_STOPPED)
+        return;
+
+    UA_LOG_INFO(cm->eventSource.eventLoop->logger, UA_LOGCATEGORY_NETWORK,
+                "HTTP\t| Shutting down the ConnectionManager");
+
+    cm->eventSource.state = UA_EVENTSOURCESTATE_STOPPING;
```
jpfr, line 125: At this point you should forcefully close the sockets.
That also means you need to keep a list of the open sockets.
See how _stop is done for TCP and UDP.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+    HTTPConnectionManager *mcm = (HTTPConnectionManager*)cm;
+    if(mcm->cm.eventSource.state == UA_EVENTSOURCESTATE_STOPPING) {
+        mcm->cm.eventSource.state = UA_EVENTSOURCESTATE_STOPPED;
+        return;
+    }
+}
+
+static UA_StatusCode
+HTTP_eventSourceDelete(UA_ConnectionManager *cm) {
+    UA_String_clear(&cm->eventSource.name);
```
jpfr, line 137: Check if closed.
See how UDP and TCP do this.

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+    HTTPConnectionManager *hcm = (HTTPConnectionManager*)cm;
+
+    protocols[0].user = (void*)cm;
+
+    /* for the purpose of debugging */
+    //int logs = LLL_USER | LLL_ERR | LLL_WARN | LLL_NOTICE;
+    lws_set_log_level(0, NULL);
+
+    struct lws_context_creation_info info;
+    void *foreign_loops[1];
```
jpfr, line 157: I have not understood this construct.
Why not just say `info.foreign_loops = hcm->cm.eventSource.eventLoop;`?

arch/eventloop_lws_http.c
```diff
@@ -0,0 +1,259 @@
...
+    info.foreign_loops = foreign_loops;
+
+    hcm->context = lws_create_context(&info);
+    if (!hcm->context)
+        return UA_STATUSCODE_BADINTERNALERROR;
+    return UA_STATUSCODE_GOOD;
+}
+
+static UA_StatusCode
+HTTP_sendWithConnection(UA_ConnectionManager *cm, uintptr_t connectionId,
```
jpfr, line 177: HTTP is synchronous.
So you have to wait for the previous request to receive a response before sending the next request.
Ensure there can be no misuse.

So either
- block sending requests while still waiting for a response
- Add the next requests to a list and send them out automatically later on.

arch/lws/ua_lws_adapter.c
```diff
@@ -0,0 +1,172 @@
...
+
+    UA_EventLoopPOSIX_deregisterFD((UA_EventLoopPOSIX*)priv->io_loop, &conn->rfd);
+    ZIP_REMOVE(UA_FDTree, &priv->fds, &conn->rfd);
+    priv->fdsSize--;
+
+    return (int)UA_STATUSCODE_GOOD;
+}
+
+static const struct lws_event_loop_ops event_loop_ops_custom = {
+    .name				= "custom",
```
jpfr, line 154: don't use tabs

examples/lws/server_lws_http_client.c
```diff
@@ -0,0 +1,48 @@
...
+    params[1].key = UA_QUALIFIEDNAME(0, "port");
+    UA_Variant_setScalar(&params[1].value, &port, &UA_TYPES[UA_TYPES_UINT16]);
+    params[2].key = UA_QUALIFIEDNAME(0, "path");
+    UA_Variant_setScalar(&params[2].value, &path, &UA_TYPES[UA_TYPES_STRING]);
+    params[3].key = UA_QUALIFIEDNAME(0, "method");
+    UA_Variant_setScalar(&params[3].value, &method, &UA_TYPES[UA_TYPES_STRING]);
+
+    UA_KeyValueMap map = {paramsSize, params};
+
+    cm->openConnection(cm, NULL, NULL, NULL, NULL);
```
jpfr, line 41: Add a callback to log the received result.
Also add the handling of failure cases where the remote side does not exist.

How are HTTP status codes lik 404 handled?
Given to the callback as part of the parameter map?

include/open62541/util.h
```diff
@@ -30,6 +30,13 @@ UA_UInt32_random(void); /* no cryptographic entropy */
 UA_Guid UA_EXPORT
 UA_Guid_random(void);   /* no cryptographic entropy */
 
+/**
+ * Memory Handling
+ * --------------- */
+
+/* Force casting a pointer. Attention, unsafe! */
+#define UA_FORCE_CAST_PTR(t, p) ((t)(uintptr_t)p)
```
jpfr, line 38: Nice!

tools/cmake/Findlibwebsockets.cmake
```diff
@@ -6,28 +6,10 @@
...
 #
 # This currently works probably only for Linux
+set(FPHSA_NAME_MISMATCHED 1) # Suppress warnings, see https://cmake.org/cmake/help/v3.17/module/FindPackageHandleStandardArgs.html
+include(FindPkgConfig)
 
-FIND_PATH ( LIBWEBSOCKETS_INCLUDE_DIR libwebsockets.h
-    /usr/local/include
-    /usr/include
-)
+pkg_search_module(LIBWEBSOCKETS REQUIRED libwebsockets)
```
jpfr, line 12: This requires pkg-config and is only available for Linux.
The previous cmake file was multi-platform.
And it should stay that way.

jpfr (commented):
Very nice!
Comments inline.

jpfr:
On more point.
The current Debian and Ubuntu releases are at lws version 4.0 and 4.1.
We should be compatible out-of-the-box with these.

Try to get as low as possible in the version requirements.
But stop before the #ifdef jungle gets too messy.

.gitmodules
jpfr: Remove the submodule. The users should manually install or use a distorted package.

CMakeLists.txt
jpfr: Do not modify settings.
Abort with a warning when you cannot build.

arch/eventloop_lws_http.c
jpfr: Still unresolved

include/open62541/plugin/eventloop.h
```diff
@@ -555,6 +555,25 @@ UA_ConnectionManager_new_POSIX_UDP(const UA_String eventSourceName);
...
+/**
+ * HTTP Connection Manager
+ * ~~~~~~~~~~~~~~~~~~~~~~~
+ *
+ * The HTTP ConnectionManager uses the libwebsockets library to send http requests.
+ *
+ * Open Connection Parameters:
+ * - 0:hostname [string]: Hostname (or IPv4/v6 address) of the connection (required).
+ * - 0:port [uint16]: Port of the target host (default: 443).
+ * - 0:path [string]: Path is the string of information that comes after the top level domain name (default: /).
```
jpfr, line 568: The way I understand it, the connection will stay open for multiple requests.
The path, method and data should be parameters for `sendWithConnection`, not here.

jpfr (commented):
Looks good. A few comments were unresolved from last time.
Let’s merge this for 1.4!

jpfr:
Rebased on master.
The examples segfault...

CMakeLists.txt
```diff
@@ -536,6 +539,15 @@ if(UA_ENABLE_TPM2_SECURITY)
     list(APPEND open62541_LIBRARIES ${TPM2_LIB})
 endif()
 
+if(UA_ENABLE_LWS)
+    find_package(libwebsockets REQUIRED)
+#    message(LWS Version: ${LIBWEBSOCKETS_VERSION})
```
jpfr, line 544: If you need LWS >= 4.3, then enable the fatal error.
Otherwise remove the comment.

arch/lws/eventloop_lws_http.c
```diff
@@ -0,0 +1,611 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ *    Copyright 2023 (c) Fraunhofer IOSB (Author: Noel Graf)
+ */
+
+#include <open62541/plugin/eventloop.h>
```
jpfr, line 8: LWS is not an architecture.
It currently depends on POSIX.
Hence I would move it to /arch/posix/lws.

arch/lws/eventloop_lws_posix.h
```diff
@@ -0,0 +1,26 @@
...
+#define _XOPEN_SOURCE 700
+#include <libwebsockets.h>
+#include "../posix/eventloop_posix.h"
+
+/* one of these is appended to each pt for our use */
+struct pt_eventlibs_custom {
+    UA_EventLoopPOSIX *io_loop;
+    struct lws_context *context;
+    size_t fdsSize;
+    UA_FDTree fds;
```
jpfr, line 20: For the first entry, you can use a `UA_POSIXConnectionManager`.
Then you get the "local" FDTree for free.
And potentially some more helper functions.

jpfr (commented):
1) Having examples is good, having unit tests is even better.
Lws can also act as a server. But that would require adding that.

Or you open a dummy server (e.g. using `python -m http.server`) and test against that.

2) How much different would it be to have websockets (client+server)?
The websockets interface should be quasi-identical to the TCP interface of the EventLoop.

diplfranzhoepfinger:
@NoelGraf i guess you must again rebase.

PR #5686, opened 2023-03-16
[Review] PubSub v1.05 add optional Padding for DataSetMessages

basyskom-dege:
Accoding to the spec ([7.2.2.5.6](https://reference.opcfoundation.org/Core/Part14/v105/docs/7.2.2.5.6#Table144)):
"A [DataSetMessage](https://reference.opcfoundation.org/search/307?t=DataSetMessage) should be padded with 0 if the size is smaller than the [ConfiguredSize](https://reference.opcfoundation.org/Core/Part14/v105/docs/6.3.1.3.3). If the  [ConfiguredSize](https://reference.opcfoundation.org/Core/Part14/v105/docs/6.3.1.3.3) is 0 the message can have dynamic length. If the [DataSetMessage](https://reference.opcfoundation.org/search/307?t=DataSetMessage) size is bigger than the [ConfiguredSize](https://reference.opcfoundation.org/Core/Part14/v105/docs/6.3.1.3.3), the bit 0 of the [DataSetFlags1](https://reference.opcfoundation.org/search/307?t=DataSetFlags1) should be set to false to indicate that the [DataSetMessage](https://reference.opcfoundation.org/search/307?t=DataSetMessage) is not valid."

Since the configuredSize not exposed to the `encodeBianary` and `calcSizeBinary` functions, I added a member in the `dataSetMessage` struct.

This might be wrong but it was the easiest way to achieve this. 

Any input on how to do this properly is highly appreciated. 

The changes on the receiver side are still WIP.

### DIFF
--- a/src/pubsub/ua_pubsub_networkmessage.c
+++ b/src/pubsub/ua_pubsub_networkmessage.c
@@ -1334,13 +1334,6 @@ UA_DataSetMessage_encodeBinary(const UA_DataSetMessage* src, UA_Byte **bufPos,
 
             if(src->header.fieldEncoding == UA_FIELDENCODING_VARIANT) {
                 rv = UA_Variant_encodeBinary(&v->value, bufPos, bufEnd);
-                /* if the configuredSize is set -> padd the message with 0. */
-                if(src->configuredSize > 0 && src->header.dataSetMessageValid) {
-                    /* Set the bytes to 0*/
-                    memset(*bufPos, 0, src->configuredSize);
-                    /* move the bufpos accordingly*/
-                    *bufPos += src->configuredSize;
-                }
             } else if(src->header.fieldEncoding == UA_FIELDENCODING_RAWDATA) {
                 UA_FieldMetaData *fmd =
                     &src->data.keyFrameData.dataSetMetaDataType->fields[i];
@@ -1361,6 +1354,7 @@ UA_DataSetMessage_encodeBinary(const UA_DataSetMessage* src, UA_Byte **bufPos,
             } else if(src->header.fieldEncoding == UA_FIELDENCODING_DATAVALUE) {
                 rv = UA_DataValue_encodeBinary(v, bufPos, bufEnd);
             }
+
             UA_CHECK_STATUS(rv, return rv);
         }
     } else if(src->header.dataSetMessageType == UA_DATASETMESSAGE_DATADELTAFRAME) {
@@ -1392,6 +1386,14 @@ UA_DataSetMessage_encodeBinary(const UA_DataSetMessage* src, UA_Byte **bufPos,
         return UA_STATUSCODE_BADNOTIMPLEMENTED;
     }
 
+    /* if the configuredSize is set -> padd the message with 0. */
+    if(src->configuredSize > 0 && src->header.dataSetMessageValid) {
+        size_t padding = (size_t)(bufEnd - *bufPos);
+        /* Set the bytes to 0*/
+        memset(*bufPos, 0, padding);
+        /* move the bufpos accordingly*/
+        *bufPos += padding;
+    }
     /* Keep-Alive Message contains no Payload Data */
     return UA_STATUSCODE_GOOD;
 }
--- a/src/pubsub/ua_pubsub_networkmessage.c
+++ b/src/pubsub/ua_pubsub_networkmessage.c
@@ -1635,9 +1635,9 @@ UA_DataSetMessage_calcSizeBinary(UA_DataSetMessage* p,
     if(p->configuredSize > 0) {
         /* If the message is larger than the configuredSize, it shall be set to not valid */
         if(p->configuredSize < size) 
-            p->header.dataSetMessageValid = 0;
+            p->header.dataSetMessageValid = UA_FALSE;
         
-        size += (p->configuredSize - size);
+        size = p->configuredSize;
     }
     
     /* KeepAlive-Message contains no Payload Data */
--- a/src/pubsub/ua_pubsub_networkmessage.c
+++ b/src/pubsub/ua_pubsub_networkmessage.c
@@ -1334,6 +1334,13 @@ UA_DataSetMessage_encodeBinary(const UA_DataSetMessage* src, UA_Byte **bufPos,
 
             if(src->header.fieldEncoding == UA_FIELDENCODING_VARIANT) {
                 rv = UA_Variant_encodeBinary(&v->value, bufPos, bufEnd);
+                /* if the configuredSize is set -> padd the message with 0. */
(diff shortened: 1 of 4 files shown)


src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -1753,6 +1762,15 @@ UA_DataSetMessage_calcSizeBinary(UA_DataSetMessage* p, UA_NetworkMessageOffsetBu
             }
         }
     }
+
+    if(p->configuredSize > 0) {
+        /* If the message is larger than the configuredSize, it shall be set to not valid */
+        if(p->configuredSize < size) 
+            p->header.dataSetMessageValid = 0;
```
andreasebner, line 1769: Please use true/false for better readability.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -690,6 +690,9 @@ UA_DataSetWriter_generateDataSetMessage(UA_Server *server,
             dsm->configuredSize = 0;
         }
 
+        /* setting configured size in the dataSetMessage to add padding later on */
+        dataSetMessage->configuredSize = dsm->configuredSize;
```
andreasebner, line 694: Where does the value come from? UA_UadpDataSetWriterMessageDataType ?

src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -1753,6 +1762,15 @@ UA_DataSetMessage_calcSizeBinary(UA_DataSetMessage* p, UA_NetworkMessageOffsetBu
             }
         }
     }
+
+    if(p->configuredSize > 0) {
+        /* If the message is larger than the configuredSize, it shall be set to not valid */
+        if(p->configuredSize < size)
```
andreasebner, line 1768: Can be combined in one statement.

jpfr:
Cool! Just remove the wip flag when we shall review.

src/pubsub/ua_pubsub_writer.c
basyskom-dege: Yes. See the added testcase.

src/pubsub/ua_pubsub_writer.c
```diff
@@ -784,9 +784,12 @@ UA_DataSetWriter_generateDataSetMessage(UA_Server *server,
                                   "Static DSM configuration not supported, using defaults");
             dsm->networkMessageNumber = 0;
             dsm->dataSetOffset = 0;
-            dsm->configuredSize = 0;
+          //  dsm->configuredSize = 0;
```
basyskom-dege, line 787: For testing purpose, I just removed this line to prevent using the default value for the configuredSize. 
How should this be done? Should we allow a static DSM configuration? How should the configuration be set otherwise?

basyskom-dege:
There are still some issues with this. 
When I use `checkReceived()` function, the test fails because the subscribed integer is 0 (the published one is 42). Using the configuredSize in the `tutorial_pubsub_publish` and `tutorial_pubsub_subscribe`, the correct integer is received and added to the server. (see screenshots from Wireshark to see the actual padding to 80 bytes and from UA_Expert, showing the Integer that has been set.)

Bildschirmfoto vom 2023-03-30 13-51-42

 
Bildschirmfoto vom 2023-03-30 13-58-54

src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -1623,6 +1632,15 @@ UA_DataSetMessage_calcSizeBinary(UA_DataSetMessage* p,
         return 0;
     }
 
+    if(p->configuredSize > 0) {
+        /* If the message is larger than the configuredSize, it shall be set to not valid */
+        if(p->configuredSize < size) 
+            p->header.dataSetMessageValid = 0;
```
jpfr, line 1638: false instead of 0.

src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -1623,6 +1632,15 @@ UA_DataSetMessage_calcSizeBinary(UA_DataSetMessage* p,
...
+        
+        size += (p->configuredSize - size);
```
jpfr, line 1640: just set size = p->configuredSize. That is more clear.

src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -1385,6 +1386,14 @@ UA_DataSetMessage_encodeBinary(const UA_DataSetMessage* src, UA_Byte **bufPos,
         return UA_STATUSCODE_BADNOTIMPLEMENTED;
     }
 
+    /* if the configuredSize is set -> padd the message with 0. */
+    if(src->configuredSize > 0 && src->header.dataSetMessageValid) {
+        size_t padding = (size_t)(bufEnd - *bufPos);
+        /* Set the bytes to 0*/
+        memset(*bufPos, 0, padding);
+        /* move the bufpos accordingly*/
+        *bufPos += padding;
```
jpfr, line 1395: *bufPos = bufEnd

src/pubsub/ua_pubsub_writer.c
jpfr: What do you mean by static DSM configuration?
The `dsm` variable contains *the* configuration.

The fastpath with static precomputed offsets should already work as well.
Since the normal encoding path is used for the initial message generation.

Either remove the commented-out line.
Or add additional explanation to the comment.

jpfr (commented):
Looks good!
Comments inline.

src/pubsub/ua_pubsub_writer.c
basyskom-dege: Yes the `dsm` variable contains the configuration but it is overwritten with 0 when you set it with `UA_UadpDataSetWriterMessageDataType` in the `UA_DataSetWriterConfig`.

How can one set the `configuredSize` so it wont be overwritten here?

src/pubsub/ua_pubsub_networkmessage.c
basyskom-dege: I used `UA_Byte_copy(bufEnd, *bufPos);` since bufEnd is const and `*bufPos = bufEnd` results in `error: assignment discards ‘const’ qualifier `

basyskom-dege:
FYI: After talking with @andreasebner about the padding, he told me it would be good to check if other SDKs are compatible with this approach. 
I used the `UA-.NETStandard` project and sent messages with a padding. After adjusting the configuration of the dotnet example subscriber, I was able to receive the messages just fine with the padding to 256 byte:

Bildschirmfoto vom 2023-04-06 11-11-21

jpfr:
Nice, merged. 👍

PR #5688, opened 2023-03-16
fix(el): Check event loop pointer before access

keba-estr:
fix(el): Check event loop pointer before access

In case of an initialization error of event loop an invalid pointer was accessed in function UA_server_delete.

fix(el): do not stop external event loop at exit

fix (el): Stop server event loop after cleaning admin session

Cleanup of admin session adds a delayed callback to the event loop.
If the event loop is stopped before, this callback is not executed any more.

fix (el): Remove duplicate stop of server eventloop

Eventloop is already stopped in UA_ServerConfig_clean.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -224,7 +224,7 @@ void UA_Server_delete(UA_Server *server) {
     UA_AsyncManager_clear(&server->asyncManager, server);
 #endif
 
-    if(server->config.eventLoop) {
+    if(server->config.eventLoop && !server->config.externalEventLoop) {
         /* Stop the EventLoop and iterate until stopped or an error occurs */
         if(server->config.eventLoop->state == UA_EVENTLOOPSTATE_STARTED)
             server->config.eventLoop->stop(server->config.eventLoop);
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -227,21 +227,6 @@ void UA_Server_delete(UA_Server *server) {
     /* Clean up the Admin Session */
     UA_Session_clear(&server->adminSession, server);
 
-    if(server->config.eventLoop && !server->config.externalEventLoop) {
-        /* Stop the EventLoop and iterate until stopped or an error occurs */
-        if(server->config.eventLoop->state == UA_EVENTLOOPSTATE_STARTED)
-            server->config.eventLoop->stop(server->config.eventLoop);
-        UA_StatusCode res = UA_STATUSCODE_GOOD;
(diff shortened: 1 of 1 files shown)


jpfr:
Looks good.
This should be the same as for the client.
https://github.com/open62541/open62541/blob/153adb15039c6cea594a54802b51794ce873cfe3/src/client/ua_client.c#L184

There we check if an "external eventloop" is configured.
In that case we don't stop it and just remove the connections related to client/server.

keba-estr:
> Looks good. This should be the same as for the client.
> 
> https://github.com/open62541/open62541/blob/153adb15039c6cea594a54802b51794ce873cfe3/src/client/ua_client.c#L184
> 
> There we check if an "external eventloop" is configured. In that case we don't stop it and just remove the connections related to client/server.

This is a separate issue. The same check for the external event loop appears in the function UA_ServerConfig_clean(), where the event loop is also stopped.
In my opinion there are two additional issues to check/resolve:
- Why is an external eventloop not stopped at server/client exit? (It is clear that an external eventloop must not be freed.)
- Why is the server eventloop stopped twice in UA_Server_delete()? (in line 227 and in the following function UA_ServerConfig_clean.

keba-estr:
Additional question: Is it safe to stop the server eventloop before calling UA_Session_clear(&server->adminSession, server);?
In the function UA_Session_clear there is a comment "Remove all Subscriptions. This may send out remaining publish responses." Without eventloop no responses will be sent.

jpfr:
We might have attached two servers to the same eventloop.
I think the following should happen:

- If we "own" the EventLoop, stop it and loop until everything has shut down.
- If do not "own" the EventLoop, shutdown the server sockets and open SecureChannels.
   Loop until the connection-counters for the server are at zero and then finalize the cleanup without deleting (or stopping) the EventLoop.

keba-estr:
> 

These

> We might have attached two servers to the same eventloop. I think the following should happen:
> 
> * If we "own" the EventLoop, stop it and loop until everything has shut down.
> * If do not "own" the EventLoop, shutdown the server sockets and open SecureChannels.
>   Loop until the connection-counters for the server are at zero and then finalize the cleanup without deleting (or stopping) the EventLoop.

All these checks are already implemented in the function UA_Server_run_shutdown, which is called when UA_Server_run is finished. Is it really necessary to do that again in UA_server_delete?

keba-estr:
I deleted the branch because of a wrong merge to newer master version. With this action the PR was closed unintentionally.

PR #5692, opened 2023-03-18
Explicitly set encoding to UTF-8 when reading file

mojca:
I would like to request an explicit switch to UTF-8 when reading the file.
The rest of the files (a few lines below) are already opened as UTF-8.
(If the patch gets accepted, it would be great to have it end up both in 1.3 and master branch.)


jpfr:
Cool.
Will be merged to master as well.

mojca:
Thank you very much for the super quick merge.

PR #5697, opened 2023-03-20
WIP: fix(el): Check event loop pointer before access

keba-estr:
fix(el): Check event loop pointer before access

In case of an initialization error of event loop an invalid pointer was accessed in function UA_server_delete.

fix(el): do not stop external event loop at exit

fix (el): Stop server event loop after cleaning admin session

Cleanup of admin session adds a delayed callback to the event loop.
If the event loop is stopped before, this callback is not executed any more.

fix (el): Remove duplicate stop of server eventloop

Eventloop is already stopped in UA_ServerConfig_clean.

Substitutes unintentionally closed PR #5688.

### DIFF
--- a/examples/pubsub/pubsub_subscribe_standalone.c
+++ b/examples/pubsub/pubsub_subscribe_standalone.c
@@ -61,7 +61,7 @@ customDecodeAndProcessCallback(UA_PubSubChannel *psc, void *ctx, const UA_ByteSt
     UA_NetworkMessage networkMessage;
     memset(&networkMessage, 0, sizeof(UA_NetworkMessage));
     size_t currentPosition = 0;
-    UA_NetworkMessage_decodeBinary(buffer, &currentPosition, &networkMessage);
+    UA_NetworkMessage_decodeBinary(buffer, &currentPosition, &networkMessage, NULL);
 
     /* Is this the correct message type? */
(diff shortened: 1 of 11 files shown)


keba-estr:
> You are right. The sockets are already closed in `UA_Server_run_shutdown`.
> 
> This leaves open the possible bug where users call `_delete` without calling `UA_Server_run_shutdown` before. Please add a check that the server is actually shut down before calling `_delete`. Or you call `UA_Server_run_shutdown` from within `_delete`.

Is there a standard way in the open62541 to check if another function has been called or is it sufficient to add the following line at the start of UA_server_delete?

`UA_assert(server->houseKeepingCallbackId` == 0);`

jpfr:
Good idea!
But don’t add it as an assert.
Make it a check with a warning log that returns an error status code.

jpfr:
The test pubsub_encoding_custom failed.
Can be merged once this is done.

keba-estr:
> The test pubsub_encoding_custom failed. Can be merged once this is done.

Seems to be because of PR #5681. I don't see how my changes affect this UNIT test. I created an issue #5707.

keba-estr:
@jpfr: Every time I reset the feature branch the PR is closed. Is there any way to reopen the PR?

keba-estr:
Pipeline fails from first commit on, where only the event loop pointer is checked -> failure because of PR #5681

PR #5705, opened 2023-03-22
State handling updates for Reader/Writer Groups and DataSetReader/Writer

gauravks:
- Add support for Pre-operational states
- Add API to enable the objects

Replaces #5456

PR #5706, opened 2023-03-22
Add state handling for PubSub Connection

gauravks:
- Implement State handling methods and include state information in Address space for PubSub connection

I have based this on the #5705 branch as this depends on some changes from that pull request.

Depends on #5705 

Replaces #5456


jpfr:
No longer current.

PR #5712, opened 2023-03-28
Openssl runtime server cert update

Maddin-619:
[Update the Server Certificate at Runtime](https://www.open62541.org/doc/master/server.html#update-the-server-certificate-at-runtime) only works for the mbedTLS implementation. For openssl the updateCertificateAndPrivateKey function pointers are missing. This leads to an segmentation fault at runtime. So I added those functions for all 4 secure endpoints.


jpfr:
Nice!
I see a lot of repetition in the new code.
Do you think we can pull out a common implementation to share between the security policies?
We can also add parameters to the „context struct“. Such as key lenghts.

jpfr:
Cool.
Yes, refactoring for more code reuse would be great.

PR #5713, opened 2023-03-28
Make the Certificate expiration configurable

Maddin-619:
The expiration time for certificates generated by the UA_CreateCertificate function where hard coded to 1 year. For more flexability I added the ExpirationInDays as a function argument. This is a public API breaking change.


jpfr:
There are many many more parameters this might take.
Make it a UA_KeyValueMap that takes all parameters.
So we can extend without future breakage.

jpfr:
Nice! Merged.

PR #5715, opened 2023-03-29
Allow dependency to libmbedtls14 for debian bookworm

keba-estr:
For information about package see https://packages.debian.org/search?keywords=libmbedtls14


jpfr:
Nice

PR #5723, opened 2023-04-05
fix(PubSub) Add StatusCode checks.

basyskom-dege:
After replacing the optional `UA_NS0ID_SERVER_LOCALTIME` with the `UA_NS0ID_SERVER_SERVERSTATUS_CURRENTTIME` in #5716, I saw that this was used in some other places without checks for the StatusCode. As a result, some tests didn't do what they intended because the Node is not available and no DataSet is being published.

I removed the LocalTime in all Pubsub tests and added missing StatusChecks all over the tests.

This fixes some issues with the encryption or publishspeed tests that were not detected because of the missing StatusCode checks.


jpfr:
Thanks!
I slightly adjusted the commit title to fit our convention.

PR #5744, opened 2023-04-17
Add activate session

engel-bichlerh:
Add global UA_Client_ActivateSession method to change preferred locale without disconnecting.
All subscriptions of the already activated session remain valid.


jpfr:
Looks good!
Please also add an async version.
So we have the entire API in sync and async versions.

src/client/ua_client_connect.c
```diff
@@ -1709,6 +1709,72 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
+    client->sessionState = UA_SESSIONSTATE_CREATED;
+    activateSessionAsync(client);
+    notifyClientState(client);
+
+    /* EventLoop is started. Otherwise activateSessionSync would have failed. */
+    UA_EventLoop *el = client->config.eventLoop;
+    UA_assert(el);
+    while(true) {
+
+        if(client->connectStatus != UA_STATUSCODE_GOOD &&
```
jpfr, line 1729: Just abort if the connectStatus is not good.
We cannot recover from that. Irrespective of the channel state.

src/client/ua_client_connect.c
```diff
@@ -1709,6 +1709,72 @@ UA_Client_connectSecureChannel(UA_Client *client, const char *endpointUrl) {
...
 
+static void
+activateSessionSync(UA_Client *client) {
+    UA_LOCK_ASSERT(&client->clientMutex, 1);
+    
+    UA_DateTime now = UA_DateTime_nowMonotonic();
+    UA_DateTime maxDate = now + ((UA_DateTime)client->config.timeout * UA_DATETIME_MSEC);
+
+    // reset state to created
+    client->sessionState = UA_SESSIONSTATE_CREATED;
```
jpfr, line 1720: Check whether the session is either already created or already activated.
Because you could call this method when there is no session at all.
(Connection with only a SecureChannel.)

You could do the test within `activateSessionAsync`.
With logging and a return statuscode.
This statuscode (if bad) should get returned from UA_Client_activateSession and UA_Client_activateSessionAsync.

Note that it does not impact the connectStatus.
As the SecureChannel might still be good.

engel-bichlerh:
Now I removed the no longer existing constand UA_SECURECHANNELSTATE_FRESH.
Was this the problem with the CI error at check_activateSession?

jpfr:
#5789 was merged.
Please verify if all works for you or if you need something in addition.

jpfr:
You are the author of the main commit in the merged PR.
Thanks for the work. 👍

Issue #4219, opened 2021-03-03
How can the server get the client url / ip / hostname in activateSession() callback ?


mpfj:
## Description
My server accepts connections without problem and I have a custom activateSession() callback.
Inside this, I'd like to be able to log the IP address / hostname / etc of the connecting client.

If supported on the target, I can see the hostname is reported in ServerNetworkLayerTCP_add()
This works as expected on my hardware and I can see the correct IP address in the open62541 logging.

```
[2021-03-03 11:17:22.648 (UTC+0000)] info/network	TCP network layer listening on opc.tcp://10.225.1.233:4840/
[2021-03-03 11:17:30.906 (UTC+0000)] info/network	Connection 19 | New connection over TCP from 10.225.1.165
[2021-03-03 11:17:30.909 (UTC+0000)] info/channel	Connection 19 | SecureChannel 1 | SecureChannel opened with SecurityPolicy http://opcfoundation.org/UA/SecurityPolicy#None and a revised lifetime of 300.00s
```

But I would like to get access to the *same* information within the activateSession() callback.
I've had a quick look but can't see anything obvious.

Can anyone tell me if this is possible ?

Cheers

## Checklist

Please provide the following information:

 - [ x ] open62541 Version (release number or git tag): 1.2
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

johanneskopf:
I try to achieve the same as @mpfj in open62541 version 1.36. I also implemented a custom activateSession() callback on the server and now need to find a way to access the client's hostname / ip address / URL (the info that is printed out as info/session) when a new client connects. In addition, I need the same information also in the closeSession() callback to also have access to that information when a client disconnects.
`info/session       SecureChannel 5 | Session "urn:[hostname]:UnifiedAutomation:UaExpert" | Session created`

Has anything changed over the past 2 years that makes this possible now?

PR #5746, opened 2023-04-18
(feat)server: Allow server to access client ip address (github/open62…

mpfj:
…541#3303)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -721,11 +721,13 @@ endif()
 
 set(exported_headers ${exported_headers}
                      ${PROJECT_SOURCE_DIR}/deps/ms_stdint.h
+                     ${PROJECT_SOURCE_DIR}/deps/open62541_queue.h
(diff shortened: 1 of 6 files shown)


mpfj:
I'm happy for any "generic" solution.
This custom fix worked for me and (apologies) I think I pushed the commit upstream by accident.

johanneskopf:
An implementation of this feature would be much appreciated. Thanks for the work so far on this. I just commented to on the related issue here: #4219 . In my case, I ideally would need access to the IP Address as well as the hostname and URL (which is printed to the console, e.g. "urn:PC-NAME:UnifiedAutomation:UaExpert").

johanneskopf:
@jpfr Thank you for your dedication to the Open62541 library. We're really interested in this feature and its integration timeline. Is it planned to implement this feature soon? Knowing the approximate timeframe would help us to plan accordingly.

andreasebner:
Thanks again for sharing your code back then. With the latest release and the introduction of the event loop, the entire networking layer has been rewritten.

PR #5747, opened 2023-04-19
fix(CI): Enable sks and encryption examples. Fix memory leak in sks example

basyskom-dege:
This fixes two small memory leaks and enables the following examples in the valgrind CI:

```
pubsub_publish_encrypted 
pubsub_subscribe_encrypted 
pubsub_publish_encrypted_sks
pubsub_subscribe_encrypted_sks 
server_pubsub_central_sks
```


.github/workflows/build_linux.yml
```diff
@@ -101,7 +101,7 @@ jobs:
             cmd_deps: sudo apt-get install -y -qq valgrind libmbedtls-dev mosquitto
             cmd_action: examples_valgrind MBEDTLS
           - build_name: "Valgrind Examples  with OpenSSL (gcc)"
-            cmd_deps: sudo apt-get install -y -qq valgrind openssl mosquitto
+            cmd_deps: sudo apt-get install -y -qq valgrind openssl libmbedtls-dev mosquitto
```
basyskom-dege, line 104: Adding mbedTLS as a dependency  because mbedTLS is included when `UA_PUBSUB_ENCRYPTION` is enabled even if `UA_ENABLE_ENCRYPTION_OPENSSL` is set. Is there a reason behind this? 
 > eg. [CMakeLists.txt line 1088](https://github.com/open62541/open62541/blob/master/CMakeLists.txt#L1088)

examples/pubsub/sks/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 ########################################
 # PubSub Security Key Service Examples #
 ########################################
-
-add_example(server_pubsub_central_sks server_pubsub_central_sks.c)
+if(UA_NAMESPACE_ZERO STREQUAL "FULL")
+    add_example(server_pubsub_central_sks server_pubsub_central_sks.c)  
+endif()
```
basyskom-dege, line 6: central sks example only works with the full ns0. Should we add the full namespace to the linux build workflow or will this take too much time?

.github/workflows/build_linux.yml
jpfr: We currently only implement the pubsub security policies for mbedtls.
They need to be ported.
Either do that or exclude the pubsub / openssl combination.

examples/pubsub/sks/CMakeLists.txt
jpfr: Try it out.

jpfr (commented):
Looks good. Comments inline.

.github/workflows/build_linux.yml
basyskom-dege: Okay. For now we could just do it like this  and add libmbedtls-dev as a dep?

We need to have Pubsub and OpenSSL as Cmake arguments in order for the examples to be built and executed.

basyskom-dege:
I added a stophandler to the inheritance example because the CI would get stuck there and not receive the SIGINT same thing at #5750

basyskom-dege:
I added the extra logic to the CI that handles a stuck process. With the full namespace0, the CI got stuck at some examples and the SIGINT wouldnt kill the process in time. I added the same on #5750 #5755 and #5757

jpfr:
> With the full namespace0, the CI got stuck at some examples and the SIGINT wouldnt kill the process in time.

This allows examples that hang (maybe due to a mutex) to "slip through the cracks".
I would prefer an approach where we don't "make ourselves blind".

For example by sending only a "first SIGINT" after 10 seconds.
And in cases where the server needs longer to start, wait for some additional "goodwill period", send SIGINT again and then kill the process.

basyskom-dege:
>For example by sending only a "first SIGINT" after 10 seconds.
And in cases where the server needs longer to start, wait for some additional "goodwill period", send SIGINT again and then kill the process.

I added something like this. Is this better? I wont add this to #5750 and #5755 for now.

jpfr:
> I added something like this. Is this better?

Looks good!
But please only wait the additional time if the process was not stopped the first time around.

Is this code already doing it?
```bash
# using a 20 second timeout with SIGTERM to kill the process if it is still running
timeout 20s bash -c 'wait $pid || kill -TERM $pid' ; EXIT_CODE=$?
```

basyskom-dege:
> Looks good! But please only wait the additional time if the process was not stopped the first time around.
> 
> Is this code already doing it?
> 
> ```shell
> # using a 20 second timeout with SIGTERM to kill the process if it is still running
> timeout 20s bash -c 'wait $pid || kill -TERM $pid' ; EXIT_CODE=$?
> ```

The timeout will just call "wait" but it will only wait for 20 seconds. If the process terminates before it will just act like it used to before my changes.

I think we should at least give 5 seconds for the first SIGINT to be processed. Then check if the process is still running with the `ps | grep "$pid";`

If it still runs afterwards, send anoter SIGINT and then use the timeout command with a 20 seconds timeout. If the timout is triggerd before the example processed the SIGINT, we kill it with SIGTERM:

```
    # wait for the process to terminate and check if the process is still running
    sleep 5
    if ps | grep "$pid"; then
        echo $pid is still in the ps output. Must still be running.
        # send the SIGINT signal again       
        kill -INT $pid || true
    fi
    
   # using a 20 second timeout with SIGTERM to kill the process if it is still running
   timeout 20s bash -c 'wait $pid || kill -TERM $pid' ; EXIT_CODE=$?
```

PR #5750, opened 2023-04-20
fix(CI):Enabled nodeset examples in CI. Fixed small memory leak

basyskom-dege:
This enables the nodeset examples and fixes a small memory leak. 

EDIT: figuring out why the CI stops and the SIGINT is not received


basyskom-dege:
@jpfr Do you know why AppVeyor build fails?

PR #5752, opened 2023-04-21
fix(plugin,cmake): Failure when build in MacOSX

Pastor:
Included fixes for:
• Incorrect way of creation an empty library for macOS. 
• Incorrect macros for macOS.


jpfr:
Nice, thx.

PR #5757, opened 2023-04-24
fix(CI): Enabled pubsub realtime examples

basyskom-dege:
This enables the realtime examples. 

NOTE: Since the CI just runs the examples without any arguments, I just fixed memory leaks that occur when those examples are just called without any arguments. I also returned 0 instead of -1 since this would break the CI.

How should this be done? Should we provide a way to run the examples with arguments?

### DIFF
--- a/examples/pubsub_realtime/pubsub_interrupt_publish.c
+++ b/examples/pubsub_realtime/pubsub_interrupt_publish.c
@@ -305,7 +305,6 @@ addPubSubConfiguration(UA_Server* server) {
 #endif
     UA_Server_addDataSetField(server, publishedDataSetIdent, &counterValue,
                               &dataSetFieldIdentCounter);
-
     UA_WriterGroupConfig writerGroupConfig;
     memset(&writerGroupConfig, 0, sizeof(UA_WriterGroupConfig));
     writerGroupConfig.name = UA_STRING("Demo WriterGroup");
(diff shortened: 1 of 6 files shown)


examples/pubsub_realtime/pubsub_interrupt_publish.c
```diff
@@ -373,6 +372,7 @@ int main(void) {
     UA_StatusCode retval = UA_Server_run(server, &running);
 #if defined(PUBSUB_CONFIG_FASTPATH_FIXED_OFFSETS) || (PUBSUB_CONFIG_FASTPATH_STATIC_VALUES)
     if(staticValueSource != NULL) {
+        UA_DataValue_init(staticValueSource);
         UA_DataValue_delete(staticValueSource);
```
basyskom-dege, line 376: Invalid free()... copied this line from [server_pubsub_publisher_rt_level.c#L210](https://github.com/open62541/open62541/blob/master/examples/pubsub/server_pubsub_publisher_rt_level.c#L210)

basyskom-dege:
> Looks good. Needs to be rebased after I merged the last PR.

Did the rebase. I also enabled the `rt_publisher` example

jpfr:
@andreasebner Do we really need

          -DUA_ENABLE_MALLOC_SINGLETON=ON \
          -DUA_ENABLE_IMMUTABLE_NODES=ON \

for the RT unit tests?
I thought we just use the `externalDataSource` for synchronization nowadays...

jpfr:
@basyskom-dege after a sync with andreas, please deactivate

          -DUA_ENABLE_MALLOC_SINGLETON=ON \
          -DUA_ENABLE_IMMUTABLE_NODES=ON \

The tests should run just as well.
If the CI works fine, let's remove it from the docs afterwards.

PR #5760, opened 2023-04-25
[REVIEW] CI: Added support for arguments in the examples workflow

basyskom-dege:
This allows to set arguments for the examples in the CI script.

The execution of the examples is outsourced to a python script since the associative arrays in bash didn't work as expected in the CI environment. 

You can set a client or server that needs to be running for each examples. Servers are started before the example, clients are started 10 seconds afterwards. 

You can also set arguments for the examples like a server address or certificates and keys that will be generated as well.

Some examples just run the client examples as well and might need a better client but this provides an environment where things can be adjusted easily. 
I added another example `ci_server.c` that is used for a lot of examples. 

There is also a blacklist for examples that dont work right now. Its mostly because they exit with a non zero return value. This results in a failing CI workflow.

There are also some memory leak fixes that have been undetected with the old workflow without args/client/server.

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -50,7 +50,8 @@ int main(int argc, char *argv[]) {
                endpointArray[i].endpointUrl.data);
     }
     UA_Array_delete(endpointArray,endpointArraySize, &UA_TYPES[UA_TYPES_ENDPOINTDESCRIPTION]);
-
+    UA_Client_delete(client);
+    
     /* Create a client and connect */
     client = UA_Client_new();
     UA_ClientConfig_setDefault(UA_Client_getConfig(client));
--- a/examples/client.c
+++ b/examples/client.c
@@ -50,7 +50,11 @@ int main(int argc, char *argv[]) {
                endpointArray[i].endpointUrl.data);
     }
     UA_Array_delete(endpointArray,endpointArraySize, &UA_TYPES[UA_TYPES_ENDPOINTDESCRIPTION]);
+    UA_Client_delete(client);
 
+    /* Create a client and connect */
+    client = UA_Client_new();
+    UA_ClientConfig_setDefault(UA_Client_getConfig(client));
     /* Connect to a server */
     /* anonymous connect would be: retval = UA_Client_connect(client, "opc.tcp://localhost:4840"); */
     retval = UA_Client_connectUsername(client, "opc.tcp://localhost:4840", "user1", "password");
--- a/examples/client.c
+++ b/examples/client.c
@@ -50,7 +50,6 @@ int main(int argc, char *argv[]) {
                endpointArray[i].endpointUrl.data);
     }
     UA_Array_delete(endpointArray,endpointArraySize, &UA_TYPES[UA_TYPES_ENDPOINTDESCRIPTION]);
-    UA_Client_delete(client);
 
     /* Create a client and connect */
     client = UA_Client_new();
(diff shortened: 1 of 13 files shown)


basyskom-dege:
@jpfr

tools/examples_with_valgrind.py
```diff
@@ -0,0 +1,183 @@
+import os
+import subprocess
+import time
+from collections import defaultdict
+import shlex
+
```
jpfr, line 6: import shlex will probably fail if we don't add this via pip-install, etc.
do we really need it?

tools/examples_with_valgrind.py
```diff
@@ -0,0 +1,183 @@
...
+example_args = {
+        "client_encryption":"opc.tcp://localhost:4840 client_cert.der client_key.der server_cert.der",
```
jpfr, line 8: you have added a description for all examples.
It would be more convenient to only describe the examples that need command-line arguments.
All other examples can be used without additional arguments by default.

That would be better because we have less breakage when a new example is added.

jpfr (commented):
Thanks for the work. Comments inline.

examples/client.c
```diff
@@ -50,7 +50,11 @@ int main(int argc, char *argv[]) {
                endpointArray[i].endpointUrl.data);
     }
     UA_Array_delete(endpointArray,endpointArraySize, &UA_TYPES[UA_TYPES_ENDPOINTDESCRIPTION]);
+    UA_Client_delete(client);
```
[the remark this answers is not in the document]
basyskom-dege: There is no purpose. Removed.

tools/examples_with_valgrind.py
```diff
@@ -0,0 +1,183 @@
...
+        "tutorial_server_alarms_conditions":"tutorial_client_events opc.tcp://localhost:4840",
+        "tutorial_server_datasource":"tutorial_client_events opc.tcp://localhost:4840",
+        "tutorial_server_firststeps":"tutorial_client_firststeps",
+        "tutorial_server_historicaldata":"client_historical",
+        "tutorial_server_historicaldata_circular":"client_historical",
+        "tutorial_server_method":"client_method_async",
+        "tutorial_server_method_async":"client_method_async"
+}
+
+blacklist = {
```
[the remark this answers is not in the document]
basyskom-dege: The "blacklist" was done before by disabling some features in CMake. Now all features are enabled but there was not enough time to check and fix all the examples that didn't run through. Some of them got stuck or had leaks in them. So I put a blacklist for only those and not disable a feature that might disable an example that works.

tools/examples_with_valgrind.py
basyskom-dege: I changed it to not use shlex.

tools/examples_with_valgrind.py
basyskom-dege: There is no need to add any description for an example.
Only if it needs a client/server or arguments. 
I am using the get() method of the dictionary. If the key is not present, this returns None and since None evaluates to False in a Boolean context, no additional arguments or a client/server will be added.

examples/client.c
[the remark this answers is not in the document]
basyskom-dege: EDIT: There was a memory leak if were not throwing out the previous client.

jpfr:
Nice! Merged.

PR #5763, opened 2023-04-28
[WIP] fix(pubsub): application crashing on connection

JanSurft:
Fixes a connection crashing when setting up and closing PubSubConnections, thanks a lot @MurzynJ ! I added the reasoning also:

## Reasoning

- iterate the eventloop -> several network events occur
  - one is a method call, requesting to close a `PubSubConnection`
  - ... some other events
  - another is an ordinary UDP message to be received on the same `PubSubConnection` that was requested to be closed

This leads to the following erroneous chain of events:

1. Process the close request (`UA_Server_removePubSubConnection`)
  - `UA_PubSubChannelUDP_closeSubscriber`
  - `UDP_shutdownConnection` (here the delayed callback is added to close the socket in the next eventloop iteration)
  - `UA_PubSubConnection_clear`
  - `UA_PubSubChannelUDP_close` (here the `UA_UDPConnectionContext` is freed, but the eventloop still keeps a pointer to it)
2. Process other network events, and eventually allocate new memory and overwrite the area previously held by `UD_UDPConnectionContext`
3. Continue to process the UDP message receive event:
  - `UA_PubSub_udpCallbackSubscribe` (here we read the context which has already been freed - and later overwritten)

## TODO

- [x] include the patch by @MurzynJ
- [ ] add a test case that tests the given behavior (any suggestions on a minimal test setup for that?)

### DIFF
--- a/plugins/ua_pubsub_udp.c
+++ b/plugins/ua_pubsub_udp.c
@@ -72,14 +72,29 @@ UA_PubSub_udpCallbackSubscribe(UA_ConnectionManager *cm, uintptr_t connectionId,
                                void *application, void **connectionContext,
                                UA_ConnectionState state, const UA_KeyValueMap *params,
                                UA_ByteString msg) {
+    UA_UDPConnectionContext *ctx = (UA_UDPConnectionContext*) *connectionContext;
+
     if (state == UA_CONNECTIONSTATE_CLOSED || state == UA_CONNECTIONSTATE_CLOSING) {
+        // We are the last users of this context, clean it up now.
+        if(ctx)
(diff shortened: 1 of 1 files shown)


jpfr:
The offending code was reworked for the EventLoop model.
Is this fix still relevant for 1.3?

MurzynJ:
> The offending code was reworked for the EventLoop model. Is this fix still relevant for 1.3?

This was a fix for an issue introduced after 1.3. We can safely close this PR.

jpfr:
Thanks @MurzynJ

PR #5771, opened 2023-05-04
fix(plugin): free old thumbprint befor creating the new one

Maddin-619:
Recently I added the update server cert at runtime function for the openssl builds with this PR https://github.com/open62541/open62541/pull/5712 .

I forgot to clear the thumbprint befor creating the new one so there is a memory leak. This PR should fix that.


jpfr:
Thanks. Merged.

PR #5779, opened 2023-05-12
fix(nc): Fixed bugs in nodeset compiler that prevented building some nodesets correctly.

NoelGraf:
Fixes code generation for enum values


jpfr:
Nice

PR #5798, opened 2023-05-25
feat(core): Allow parsing of opc.eth:// in UA_parseEndpointUrl

keba-estr:
fix(core): fix string length error with mqtt in UA_parseEndpointUrl

refactor(arch): Don't listen for events on UDP send-connections


keba-estr:
@jprf: This is the communicated PR about your two bug fixes including additional mqtt bug fixes.

PR #5807, opened 2023-06-06
fix(HDA): Replace HDA dataEncoding with empty string

douwefaber:
Fetching HDA did not work while UAExpert did. Used Wireshark to analyze the difference.


jpfr:
Thanks!

PR #5832, opened 2023-06-23
[Review] fix(nodesetinjector): Extend test for Scales v1.01 specification

GoetzGoerisch:
* Updates ua-nodeset
* Fixes Scales test, as new CS requirements where added with v1.01

Replace #5826


jpfr:
Nie

PR #5833, opened 2023-06-23
Simplify architecture includes

jpfr:


mojca:
Is this backportable to 1.3?

jpfr:
> Is this backportable to 1.3?

Not overall. It took a long time to refactor the architecture-dependent networking stuff out of the core library.

But some aspects like the custom printf could be backported.

PR #5839, opened 2023-06-26
Replace intrinsics to fix the build with multithreading support for MSVC

mojca:
This is a draft pull request since I would first like to get some guidance from you and I'm not sure if these are the absolutely correct fixes.

For some reason that I cannot fully explain I don't seem to be able to compile the code without these changes. With these changes the code compiles fine.

We were initially not sure how to upstream the changes, but I noticed now that this has already been addressed in the `master` branch. The only issue is that we would like to use a stable branch, and would thus ideally like to see a fix in `1.3`. I'm not sure if the changes from `master` could easily be cherry-picked since there were multiple non-trivial commits involved, the last one being 76bb7dc11f0302c1dddc52ddf72d20d5a5bf52bb by @jpfr.

These functions were mostly removed by 108b8c537b2b7edb28038162ffe109c08715dabe
```
commit 108b8c537b2b7edb28038162ffe109c08715dabe
Author: Julius Pfrommer <julius.pfrommer@web.de>
Date:   Fri May 20 10:54:52 2022 +0200

    refactor(core): Remove UA_atomic_ functions (not needed)

    We currently protect against race conditions with mutexes instead of
    atomic operations.
```
and the last one was added in a clean way by a77ea9a80c8ca031a5052b08e6a49fc7e198229a
```
commit a77ea9a80c8ca031a5052b08e6a49fc7e198229a
Author: Julius Pfrommer <julius.pfrommer@web.de>
Date:   Tue Jan 24 22:42:19 2023 +0100

    feat(core): Add atomic compare-and-switch operation
```

See also #5836, #5838.

@anzeskerjancDS

### DIFF
--- a/include/open62541/architecture_definitions.h
+++ b/include/open62541/architecture_definitions.h
@@ -415,11 +415,24 @@ UA_STATIC_ASSERT(sizeof(bool) == 1, cannot_overlay_integers_with_large_bool);
     #define UA_atomic_sync()
 #endif
 
+/* Intrinsic atomic operations are not available everywhere for MSVC.
+ * Use the win32 API. Prevent duplicate definitions by winsock2. */
+#if UA_MULTITHREADING >= 100 && defined(_WIN32)
+#ifndef _WINSOCKAPI_
+#define _NO_WINSOCKAPI_
+#endif
+#define _WINSOCKAPI_
+#include <windows.h>
+#ifdef _NO_WINSOCKAPI_
+#undef _WINSOCKAPI_
+#endif
+#endif
+
 static UA_INLINE void *
 UA_atomic_xchg(void * volatile * addr, void *newptr) {
 #if UA_MULTITHREADING >= 100
 #ifdef _MSC_VER /* Visual Studio */
-    return _InterlockedExchangePointer(addr, newptr);
+    return InterlockedExchangePointer(addr, newptr);
 #else /* GCC/Clang */
     return __sync_lock_test_and_set(addr, newptr);
 #endif
@@ -434,7 +447,7 @@ static UA_INLINE void *
 UA_atomic_cmpxchg(void * volatile * addr, void *expected, void *newptr) {
 #if UA_MULTITHREADING >= 100
 #ifdef _MSC_VER /* Visual Studio */
-    return _InterlockedCompareExchangePointer(addr, newptr, expected);
+    return InterlockedCompareExchangePointer(addr, newptr, expected);
 #else /* GCC/Clang */
     return __sync_val_compare_and_swap(addr, expected, newptr);
 #endif
@@ -451,7 +464,7 @@ static UA_INLINE uint32_t
 UA_atomic_addUInt32(volatile uint32_t *addr, uint32_t increase) {
 #if UA_MULTITHREADING >= 100
 #ifdef _MSC_VER /* Visual Studio */
-    return _InterlockedExchangeAdd(addr, increase) + increase;
+    return InterlockedExchangeAdd(addr, increase) + increase;
 #else /* GCC/Clang */
     return __sync_add_and_fetch(addr, increase);
 #endif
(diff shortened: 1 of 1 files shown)


jpfr:
Are you sure that winsock2 is the correct include for this?
It does work. But it pulls in quite a bit more...

mojca:
> Are you sure that winsock2 is the correct include for this? It does work. But it pulls in quite a bit more...

No, I'm definitely not sure, and I would be grateful if you have a better suggestion. The site https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchangepointer points to `winnt.h`.

I would also double-check all the casts. I'm sure that some were needed, but maybe not all.

But the bigger question is: if we find the proper include, clean up the casts, ... would you consider a change similar to this one acceptable for the 1.3 branch? Would you perhaps prefer to do it in a different way?

include/open62541/architecture_definitions.h
```diff
@@ -415,11 +415,17 @@ UA_STATIC_ASSERT(sizeof(bool) == 1, cannot_overlay_integers_with_large_bool);
     #define UA_atomic_sync()
 #endif
 
+#if UA_MULTITHREADING >= 100
+#ifdef _MSC_VER // Visual Studio
+    #include <winsock2.h>
+#endif
+#endif
```
mojca, line 422: Your master branch contains the following code which also seems to work here (at least for a short test, I might want to check a bit more thoroughly):
```suggestion
#if UA_MULTITHREADING >= 100 && defined(_WIN32)
#ifndef _WINSOCKAPI_
#define _NO_WINSOCKAPI_
#endif
#define _WINSOCKAPI_
#include <windows.h>
#ifdef _NO_WINSOCKAPI_
#undef _WINSOCKAPI_
#endif
#endif
```

include/open62541/architecture_definitions.h
```diff
@@ -415,11 +415,17 @@ UA_STATIC_ASSERT(sizeof(bool) == 1, cannot_overlay_integers_with_large_bool);
...
+
 static UA_INLINE void *
 UA_atomic_xchg(void * volatile * addr, void *newptr) {
 #if UA_MULTITHREADING >= 100
 #ifdef _MSC_VER /* Visual Studio */
-    return _InterlockedExchangePointer(addr, newptr);
+    return InterlockedExchangePointer((PVOID*) addr, newptr);
```
mojca, line 428: I suspect that `(PVOID*)` is not necessary here.

include/open62541/architecture_definitions.h
```diff
@@ -499,7 +505,7 @@ static UA_INLINE size_t
 UA_atomic_subSize(volatile size_t *addr, size_t decrease) {
 #if UA_MULTITHREADING >= 100
 #ifdef _MSC_VER /* Visual Studio */
-    return _InterlockedExchangeSub(addr, decrease) - decrease;
+    return InterlockedExchangeAdd((LONG*) addr, - (LONG) decrease) - decrease;
```
mojca, line 508: The second `(LONG)` is required.

The first one is not required for 64-bit build, I would have to check the 32-bit.

include/open62541/architecture_definitions.h
jpfr: Yes, by via is a typo.

jpfr:
Looks good.
Remove the draft flag when you are happy.
Then we merge.

mojca:
Thank you very much. I squashed the commits and removed the Draft label. Please let me know in case I need to do anything else.

mojca:
(PS: it probably makes sense to merge (by discarding changes) the 1.3 branch into `master` soon after merging this, to avoid headaches with merge conflicts later.)

jpfr:
Nice!

mojca:
@jpfr: thank you very much

PR #5842, opened 2023-06-27
Request for cherry-picking some commits from master into 1.3

mojca:
I'm not sure about the desired way to handle cherry-picking in your project.

Here are two commits from master that we would greatly appreciate in the 1.3 branch.

Thanks a lot.

@gusentom @jpfr


jpfr:
Done

mojca:
Thank you.

PR #5849, opened 2023-07-01
fix(cmake): Add missing compile definitions to open62541-object/plugins

mojca:
UA_ARCHITECTURE_* compile definition was added only to open62541 target, but it should be also added to open62541-plugins and open62541-object targets.

An additional somewhat related question: why do we have a standalone
```cmake
add_definitions(-DUA_ARCHITECTURE_${UA_ARCHITECTURE_UPPER})
```
that is not bound to any target? @Pro 

@janm-dw / @ja4nm


janm-dw (approved):

ja4nm (approved):

jpfr:
We currently simplify the architecture definitions in #5833.

When that is done, the only architecture-dependencies will be in /arch and /plugins.

The core library the only takes some definitions from /include/open62541/config.h.

Then we will have a much smaller header surface. For example for distribution packages.

let’s revisit this when #5833 has landed.

Issue #5814, opened 2023-06-12
CS PADIM could not be compiled


ccvca:
## Description

Using the code generator for the PADIM-Specification, the following code is generated:

```c++
#define UA_TYPES_PADIM_COUNT 0
#define UA_TYPES_PADIM NULL
```

```c++
/* Change namespaceIndex from current namespace */
for(int i = 0; i < UA_TYPES_PADIM_COUNT; i++) {
UA_TYPES_PADIM[i].typeId.namespaceIndex = ns[2];
UA_TYPES_PADIM[i].binaryEncodingId.namespaceIndex = ns[2];
}
```
As UA_TYPES_PADIM is "void* NULL", this results in the follwing error:
```c++
/[...]/src_generated/namespace_padim_generated.c:4809:15: warning: dereferencing ‘void *’ pointer
 4809 | UA_TYPES_PADIM[i].typeId.namespaceIndex = ns[2];
      |               ^
/[...]/src_generated/namespace_padim_generated.c:4809:18: error: request for member ‘typeId’ in something not a structure or union
 4809 | UA_TYPES_PADIM[i].typeId.namespaceIndex = ns[2];
      |                  ^
/[...]/src_generated/namespace_padim_generated.c:4810:15: warning: dereferencing ‘void *’ pointer
 4810 | UA_TYPES_PADIM[i].binaryEncodingId.namespaceIndex = ns[2];
      |               ^
/[...]/src_generated/namespace_padim_generated.c:4810:18: error: request for member ‘binaryEncodingId’ in something not a structure or union
 4810 | UA_TYPES_PADIM[i].binaryEncodingId.namespaceIndex = ns[2];
      |                  ^

```

The code is generated by these lines:

https://github.com/open62541/open62541/blob/f481959ca3330fb24f81e23383a5e007f68e9fe3/tools/nodeset_compiler/backend_open62541.py#L311-L319

This condition is not sufficiant: `if len(typesArray) > 0:` But I don't known how to get the real length of the types.

 - [x] open62541 Version (release number or git tag): master

GoetzGoerisch:
@ccvca thank you for the report.

@NoelGraf could you please look into this?

NoelGraf:
@ccvca can you provide your CMake file in which the PADIM specification is generated?

However, I already suspect what the problem is. PADIM defines its own data types but does not provide a bsd file. Also the xml file does not contain a base64 decoded block from which the bsd file can be generated.
[...]

PR #5851, opened 2023-07-02
fix(nodesetinjector):  Bug fixing when adjusting the type array.

NoelGraf:
The type array of the respective nodeset is not adjusted if an empty bsd file is specified. If no bsd is specified or the xml file does not contain a base64 block, an empty bsd file is automatically generated.
PADIM specification included in the nodesetinjector.
Fix for #5814


jpfr:
Changed this against the 1.4 branch.
Requires resolving though...

jpfr:
👍

PR #5859, opened 2023-07-04
fix(client): fix UA_Client_connect responseFindServers incorrectly skipping Discovery or ClientServer application types

dpazj:


jpfr:
Thanks!

Issue #5813, opened 2023-06-12
Modification of global variables in generated code for multi-server applications


ccvca:
## Description

Before #4916, it should have been possible to instantiate mutliple OPC UA Servers, as all information about the server was inside the UA_Server structure.

But the current code generator does introduce this for all companion specifications, which adapts the global variable to the current namespace layout:
```C++
/* Change namespaceIndex from current namespace */
for(int i = 0; i < UA_TYPES_ROBOTICS_COUNT; i++) {
  UA_TYPES_ROBOTICS[i].typeId.namespaceIndex = ns[2];
  UA_TYPES_ROBOTICS[i].binaryEncodingId.namespaceIndex = ns[2];
}
```
I do understand why this is necessesry, but in case multiple servers with different namespaces layouts are instantiated, this will not work.

## Background Information / Reproduction Steps

https://github.com/open62541/open62541/pull/4916/files#diff-5460518878d90fbfcd62d617629d9aecc0b290c6ef7b60841f724468c710a0a7R309-R318

Resolving this issue, may result in an API/ usage Change of the open62541-library.

I see the following possibilities to resolve this:

- `Idea 1:  Do a copy of the type description to UA_Server and adjust only the copy. Use `UA_Server_findDataType()` in all places, where a direct access to the `UA_TYPES_` is used at the moment.
- Idea 2: Do a copy of the type description to UA_Server and adjust only the copy. Return a pointer to the copy in `namespace_XXX_generated`
- Idea 3: Do a copy of the type description to UA_Server and adjust only the copy. Return a struct with a pointer to each defined type.

I'm unsure if `Idea 1` will degrade the performance too much. `Idea 2`should provide a similar performance to the existing implementation and allows a straight forward migration path. `Idea 3` does allows to store the types internaly in a different data structure (e.g. zip-tree) for an increased performance for all cases that require `UA_Server_findDataType()`, but this would also result in a new API.

Due to the copy, all proposed ideas result in an increased memory usage of around 60 bytes per type entry.

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): master
 - [x] Self-contained code example attached
 - [x] Critical issue

PR #5864, opened 2023-07-07
[WIP]feat(server): Include the TypeArrays in the server structure.

NoelGraf:
Let's consider a scenario where we have a program that starts two servers with identical nodesets, but the nodesets are loaded in different order.
At startup, the first server modifies the global variable TypesArray and changes the Indices of the nodesets according to the order in which they are loaded.
Similarly, the second server is started and modifies the TypesArray variable, changing the Indices of the nodesets due to their different loading order. As a result, the Indices no longer match for the first server, leading to a crash.

To solve this problem, include the TypeArrays in the server structure, ensuring that each server has its own independent TypesArrays.

Fix for #5813


jpfr:
There already is a type array.
Defined in types.h just above.
Better modify the existing structure.

Further, use a linked list with a simple forward pointer. So you don’t have to pull the queue.h into the public headers.

Currently we don’t need to make a system-wide install of the queue.h for those who just install the binary library + headers with a Debian package.

include/open62541/server.h
```diff
@@ -1839,6 +1839,10 @@ UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Server_getNamespaceByIndex(UA_Server *server, const size_t namespaceIndex,
                               UA_String *foundUri);
 
+/* Add a new typeArray to the server. */
```
jpfr, line 1842: This needs documentation whether we make a deep copy of the array.
Otherwise we can add a flag to indicate that an array is always accessible and does not need to be copied. That means:

- the array outlives the lifetime of the server
- the array can be accessed from any thread (not on the stack)
- the array is not cleaned up automatically

Given all that it could be safer to always make a deep copy.
What do you think?

The custom types are part of the visible server config.
Do we add to the server or to the config? The api should be clear about that.
Will serverconfig _clear delete the array?

how do we handle server-specific namespace-index?
Where will the rewriting of the datatype namespace index take place?

src/server/ua_server_internal.h
```diff
@@ -148,6 +148,9 @@ struct UA_Server {
     size_t namespacesSize;
     UA_String *namespaces;
 
+    /* Type Arrays */
+    const UA_DataTypeArray *typeArrays;
```
jpfr, line 152: Now we have a type array in the server config and internal.
Why both and how are they related?

PR #5870, opened 2023-07-12
feat (server): add asserts for serviceMutex

keba-estr:
Call UA_LOCK_ASSERT in all functions where unlock is done.


jpfr:
Thanks!

Issue #5871, opened 2023-07-12
Double lock with UA_Client_Service_browse


keba-estr:
## Description
Double lock if user callback calls function which locks serviceMutex

## Background Information / Reproduction Steps

User call at client side: UA_Client_Service_browse

Server callstack:
-> open62541 code
processMSGDecoded -> UA_LOCK(serviceMutex)
service (Service_Browse)
UA_Server_processServiceOperations
operationCallback (Operation_Browse)
browse
allowBrowseNode
-> user code
   allowBrowseNodeAad
   ...
   -> calls open62541 code
      UA_Server_browse
      -> UA_LOCK(serviceMutex)

## Checklist

Please provide the following information:

 - [ ] open62541 Version (release number or git tag): master

keba-estr:
Will be resolved with PR #5872

PR #5872, opened 2023-07-12
fix(server): unlock serviceMutex before calling accessControl callbacks

keba-estr:
See issue #5871


jpfr:
You need to take that lock again before returning.
For example, store the result in a Boolean, take the lock again, and only then evaluate the Boolean.

Otherwise great as always!

keba-estr:
> You need to take that lock again before returning. For example, store the result in a Boolean, take the lock again, and only then evaluate the Boolean.
> 
> Otherwise great as always!

Does the commited simple UA_UNLOCK suffice or do I really have to save some parameters or the return value?

Issue #5316, opened 2022-08-31
UA_Variant_copy ignores storageType


keba-estr:
## Description

If the function UA_Variant_copy is called for a variant which has set the storageType to UA_VARIANT_DATA_NODELETE, the data array pointer should not be allocated and deep copied, but only used from the old data array pointer.

If you confirm that this is the correct usage, I can create a pull request for the change.

The new function would look like this:

`static UA_StatusCode`
`Variant_copy(UA_Variant const *src, UA_Variant *dst, const UA_DataType *_) {`
`    size_t length = src->arrayLength;`
`    if(UA_Variant_isScalar(src))`
`        length = 1;`
`    UA_StatusCode retval;`
`    if(p->storageType == UA_VARIANT_DATA_NODELETE) {`
`       dst->data = src->data;`
`       retval = UA_STATUSCODE_GOOD;`
`    } else {`
`       retval = UA_Array_copy(src->data, length,`
`                              &dst->data, src->type);`
`    }`
`    if(retval != UA_STATUSCODE_GOOD)`
`        return retval;`
`    dst->arrayLength = src->arrayLength;`
`    dst->type = src->type;`
`    if(src->arrayDimensions) {`
`        retval = UA_Array_copy(src->arrayDimensions, src->arrayDimensionsSize,`
`            (void**)&dst->arrayDimensions, &UA_TYPES[UA_TYPES_INT32]);`
`        if(retval != UA_STATUSCODE_GOOD)`
`            return retval;`
`        dst->arrayDimensionsSize = src->arrayDimensionsSize;`
`    }`
`    return UA_STATUSCODE_GOOD;`
`}`

## Checklist

 - [ ] open62541 Version (release number or git tag): 1.3.3 (also in latest master with commit e03be588a2a55fcfa0c1f2a00ef0b305c3d0de96)

keba-estr:
Create PR #5875 for this issue.

jpfr:
> The way UA_VARIANT_DATA_NODELETE is used now is that it only affects the variable data, not the variable arrayDimensions which must be deep copied in both cases.

No. This is what the documentation says:
```c
/* The data is "borrowed" by the variant and is
 * not deleted when the variant is cleared up.
 * The array dimensions also borrowed. */
 ```

In the current code, a normal copy is a deep copy that also copies the array dimensions.
If you make a shallow copy, the value and the array dimensions are re-used, but both are not freed in UA_Variant_clear if the NoDelete flag is still set.

[...]

PR #5875, opened 2023-07-13
WIP: fix(core): Check for UA_VARIANT_DATA_NODELETE when copying variant

keba-estr:
Resolves issue #5316


keba-estr:
Wrong solution

PR #5876, opened 2023-07-13
fix(ci): Fix git version matching

keba-estr:
Fix parsing of version_minor and version_patch from git description.

For version v1.3.5-... the version_minor was set to 5 and version_patch
was empty with the old regex.

Error was introduced in PR #4854.


jpfr:
Thx

PR #5884, opened 2023-07-18
fix(tests): Correct type of publisherId in tests

keba-estr:
Correct type of publisherId in tests to match publisherId type in connection and readergroup configuration.


jpfr:
👍

PR #5891, opened 2023-07-19
[WIP] refactor(el): Provide common EventLoop functions as external interface

keba-estr:
Provide common EventLoop functions as external interface. These functions provide common handling to be used in user EventLoop callbacks.


jpfr:
What is the advantage of moving the lock to the public api?

keba-estr:
> What is the advantage of moving the lock to the public api?

Functions like UA_EventLoopPOSIX_registerEventSource or others, that only access the EventLoop type and no posix specific extensions, could be added to the public EventLoop API to be reused easily in user implemented EventLoops.

keba-estr:
Closed, as no specific event loop is needed at KEBA side at the moment.

PR #5894, opened 2023-07-21
Allow sampling interval 0 for monitored items

ja4nm:
I am pretty sure, this is not the correct solution, but it was the only thing that have worked for me.

Since changes on 31 Jan 2023, I was no longer able to set the monitored item to "event notifier" node. When calling `UA_Server_writeValue()`, the method `triggerImmediateDataChange()` was triggered. However the variable `node->head.monitoredItems` was always NULL, no matter how I configured the monitored item. I have tried the following. Non of which worked before my change.

```
UA_MonitoredItemCreateRequest& monRequest = monRequests.itemsToCreate[i];
monRequest = UA_MonitoredItemCreateRequest_default(nodeId);
monRequest.requestedParameters.queueSize = 1000;
monRequest.requestedParameters.discardOldest = true;
monRequest.itemToMonitor.attributeId = UA_ATTRIBUTEID_EVENTNOTIFIER;
monRequest.monitoringMode = UA_MONITORINGMODE_REPORTING;
```

```
UA_MonitoredItemCreateRequest& monRequest = monRequests.itemsToCreate[i];
monRequest = UA_MonitoredItemCreateRequest_default(nodeId);
monRequest.requestedParameters.samplingInterval = 0;
monRequest.requestedParameters.queueSize = 1000;
monRequest.requestedParameters.discardOldest = true;
```


janm-dw:
The correct solution probably is to change the sampling interval limits on the server config:

```
UA_ServerConfig* config = UA_Server_getConfig(server);
config->samplingIntervalLimits.min = 0;
```

This PR can be closed.

PR #5898, opened 2023-07-25
Add a simple CI job for Windows VS 2019

mojca:
There is a lot that could be improved in this pull request, but I wanted to start with a MWE that shows a build issue with Visual Studio in Release configuration. I would be grateful for further feedback and can do minor improvement, but probably not super extensive ones.

I'll open a separate PR to fix this particular build issue, but I wanted to open two separate ones because they can be merged (and improved) independent from each other.


janm-dw (approved):

jpfr:
I wouldn't muck the CI for "historical" branches.
Let's add this to 1.4 or master.

mojca:
The target can easily be changed. Is there something simple I could change?

andreasebner:
@marwinglaser should we add this change to our mainline?

marwinglaser:
GitHub is useless for reviewing this PR. Checking locally, this PR actually only adds the workflow in a single commit. Considering we already have Windows CI using azure pipelines (VS2019) and appveyor (VS2013) and from what I can tell already had back when this PR was created, I'm not sure it's a good idea to add this separately. The question should rather be why the existing CI didn't catch the build failure and update it to do so (likely multithreading or VS2022).

I have played with the thought of porting the windows CI (at least azure) to github actions before to make it easier to work with, but this would be a separate and probably somewhat larger undertaking.

andreasebner:
The project now has a comprehensive GitHub Actions Windows workflow with four configurations: MSVC (VS 2022), MSVC+OpenSSL, MinGW, and Clang-MinGW.

PR #5905, opened 2023-07-27
fix(examples): Use synchronous control flow in client_subscription_lo…

jpfr:
…op.c


hdabis1234:
But what happens if the opc server goes down then up?

thomvet:
Could this be merged? I just fell into this trap and it took me a while to figure it out... (arrived at the same workaround as in this PR).

17steen:
Still getting into this issue when running 1.4 examples.

PR #5906, opened 2023-07-28
Change in nodemanagement to fix problem with structures which are not

ml-bb:
created with Variable Type BaseDataVariableType.

If creating a structure with BaseDataVariableType as type definition, the attribute read before has no value. Because of this writeAttribute() is not called.

If creating a structure with another variable type as type definition, the attribute has a value. Strange it tells, that it is an extension object and not a structure anymore.
Afterwards writeAttribute() in the NodeSet changes the structre to an extension object.

This leads to the problem that we can not describe and handle the node properly after construction.

Here we fix the problem by not overwriting the attribute if it of type extension object.


ml-bb:
Since we are not completely confident that this really solves the problem, we would really be thankful to have your feedback about this.

andreasebner:
Has been rewritten on master. The old `readWithReadValue` / `if(orig.type)` pattern no longer exists. BadWaitingForInitialData for empty  non-nullable variables is returned now. Thank's for your input to our project.

PR #5913, opened 2023-08-01
[WIP] feat(server): Add GetFirstSession, GetNextSession

keba-lbm:
Add two functions to get the sessions that are in the session list. The function GetFirstSession gets the head of the list and GetNextSession returns the next session that is in the list.

This allows to get access to the sessions and their fields. This makes it possible to get the information that are in the clientDescription field of the session.

### DIFF
--- a/src/server/ua_server_internal.h
+++ b/src/server/ua_server_internal.h
@@ -265,6 +265,12 @@ getSessionByToken(UA_Server *server, const UA_NodeId *token);
 UA_Session *
 getSessionById(UA_Server *server, const UA_NodeId *sessionId);
 
+const UA_Session *
+UA_Server_GetFirstSession(UA_Server *server);
+
+const UA_Session *
+UA_Server_GetNextSession(UA_Server *server, const UA_Session *session);
+
 /*****************/
 /* Node Handling */
 /*****************/
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -178,6 +178,32 @@ getSessionById(UA_Server *server, const UA_NodeId *sessionId) {
     return NULL;
 }
 
+const UA_Session *
+UA_Server_GetFirstSession(UA_Server *server) {
+    UA_LOCK_ASSERT(&server->serviceMutex, 1);
+    if (server->sessions.lh_first != NULL) {
+        return &server->sessions.lh_first->session;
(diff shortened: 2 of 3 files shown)


keba-lbm:
Should the UA_LOCK(&server->serviceMutex) and UA_UNLOCK(&server->serviceMutex) be set in the check_session.c or in the functions UA_Server_GetFirstSession and UA_Server_GetNextSession itself?

jpfr:
This is already exposed via the public API.
https://www.open62541.org/doc/master/server.html#session-handling

Just use `UA_Server_getSessionAttribute`.

keba-lbm:
Thanks for the quick answer.

I have two questions regarding the identification of a client.

1) Is it possible to get the IP-Address of a connected client?
2) What is the common way to identify a client? 
    Can I use the applicationUri as unique identifier for the client like it is described here [https://reference.opcfoundation.org/Core/Part4/v105/docs/7.2](https://reference.opcfoundation.org/Core/Part4/v105/docs/7.2)

   At the moment, I can create two clients and set the same applicationUri and both get connected to the server. So it is not 
   possible to distinguish between those two clients. 
  
Here is a code snipped how I connect the clients to the server and configure the applicationUri.

```
    // create first client
    UA_Client* client1 = UA_Client_new();
    UA_ClientConfig_setDefault(UA_Client_getConfig(client1));
    
    // set applicationUri
    UA_String_clear(&client1->config.clientDescription.applicationUri);
    client1->config.clientDescription.applicationUri = UA_String_fromChars("First_Application");

    // connect first client
    UA_StatusCode retval1 = UA_Client_connect(client1, "opc.tcp://localhost:4840");
    ck_assert_uint_eq(retval1, UA_STATUSCODE_GOOD);

    // create second client
    UA_Client* client2 = UA_Client_new();
    UA_ClientConfig_setDefault(UA_Client_getConfig(client2));

    // set applicationUri
    UA_String_clear(&client2->config.clientDescription.applicationUri);
    client2->config.clientDescription.applicationUri = UA_String_fromChars("First_Application");

    // connect second client
    UA_StatusCode retval2 = UA_Client_connect(client2, "opc.tcp://localhost:4840");
    ck_assert_uint_eq(retval2, UA_STATUSCODE_GOOD);
```

PR #5923, opened 2023-08-03
Feature/fix segmentation fault

keba-estr:
Segmentation fault occurred on linux system with UA_ENABLE_PUBSUB_BUFMALLOC enabled


jpfr:
Good find

PR #5924, opened 2023-08-03
fix(pubsub): Support all DSR publisher id types for xml encoding

keba-estr:


keba-estr:
At least the old check is wrong.
If only 32 bit publisher id was allowed, there should have been a '||' between the if statements.
The old check only checked that if the publisher id type was not UINT32, the publisherId value converted to 32Bit would need to match.

jpfr:
👍

PR #5926, opened 2023-08-04
[WIP] Move all relevant public "static inline"  API functions to UA_INLINABLE 

pieter-scholtz:
Some  "static inline" functions need to be moved to UA_INLINABLE so that the library can be comfortably used as a dynamically linked library. 

**The following needs to be done :** 

- [x] Generated type handling functions.
- [x] Client (raw) functions.
- [ ] High level client functions.
- [ ] Async high level client functions.
- [ ] Client subscription functions
- [ ] Server (raw) functions.
- [ ] Data type functions (include/open62541/types.h) and Utils
- [ ] Logging functions

**Unsure about moving these :**
- Node store 
- Event loop
- Architechture

Maybe I missed something, but I am busy writing dart/flutter bindings , and the automated tooling that I am using complains about "static inline" functions, so I these will come to the surface later on.

### DIFF
--- a/include/open62541/client.h
+++ b/include/open62541/client.h
@@ -240,10 +240,10 @@ UA_ClientConfig_clear(UA_ClientConfig *config);
  * UA_ClientConfig_setAuthenticationCert for x509-based authentication, which is
  * implemented as a plugin (as it can be based on different crypto
  * libraries). */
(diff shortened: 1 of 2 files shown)


jpfr:
Looks good.
I’m already merging this. Better to have several smaller pr that are easier to review.

PR #5934, opened 2023-08-08
fix(plugin): Declare log constants static

bluhm:
The constant string arrays logLevelNames and logCategoryNames are not exported neither in header include nor in dynamic library.  But they appear in static libopen62541.a library and my create conflicts. Declare these symbols static to keep them local in ua_log_stdout.c .


jpfr:
Looks good. Merged.

PR #5935, opened 2023-08-08
feat(plugin): Allow custom password validation

bluhm:
Username password validation in access control plugin allows only passwords stored in plaintext.  Add a callback to implement custom validation.  This allows salted password hashes and timing safe comparison in the server.  Furthermore remote validation could be implemented.  The new function UA_AccessControl_setCallback() is used to install a UA_UsernamePasswordLoginCallback with context.

The function to set the callback is added to the server config.  This reduces the API change.  Adding a parameter to UA_AccessControl_default() would also work, but this would require users to change the callers.  If this change is accepted, I can write some tests.  It looks like there are not any tests for username password login.

### DIFF
--- a/plugins/include/open62541/plugin/accesscontrol_default.h
+++ b/plugins/include/open62541/plugin/accesscontrol_default.h
@@ -18,6 +18,11 @@ typedef struct {
     UA_String password;
 } UA_UsernamePasswordLogin;
 
+typedef UA_StatusCode (*UA_UsernamePasswordLoginCallback)
+    (const UA_String *userName, const UA_ByteString *password,
+    size_t usernamePasswordLoginSize, const UA_UsernamePasswordLogin
+    *usernamePasswordLogin, void **sessionContext, void *loginContext);
+
 /* Default access control. The log-in can be anonymous or username-password. A
  * logged-in user has all access rights.
  *
@@ -30,6 +35,16 @@ UA_AccessControl_default(UA_ServerConfig *config,
                          const UA_ByteString *userTokenPolicyUri,
                          size_t usernamePasswordLoginSize,
                          const UA_UsernamePasswordLogin *usernamePasswordLogin);
+
+UA_EXPORT UA_StatusCode
+UA_AccessControl_defaultWithLoginCallback(UA_ServerConfig *config,
+                         UA_Boolean allowAnonymous,
+                         UA_CertificateVerification *verifyX509,
+                         const UA_ByteString *userTokenPolicyUri,
+                         size_t usernamePasswordLoginSize,
+                         const UA_UsernamePasswordLogin *usernamePasswordLogin,
+                         UA_UsernamePasswordLoginCallback loginCallback,
+                         void *loginContext);
 
 _UA_END_DECLS
 
--- a/plugins/ua_accesscontrol_default.c
+++ b/plugins/ua_accesscontrol_default.c
@@ -21,6 +21,8 @@ typedef struct {
     UA_Boolean allowAnonymous;
     size_t usernamePasswordLoginSize;
     UA_UsernamePasswordLogin *usernamePasswordLogin;
+    UA_UsernamePasswordLoginCallback loginCallback;
+    void *loginContext;
     UA_CertificateVerification verifyX509;
 } AccessControlContext;
 
@@ -95,11 +97,18 @@ activateSession_default(UA_Server *server, UA_AccessControl *ac,
 
         /* Try to match username/pw */
         UA_Boolean match = false;
-        for(size_t i = 0; i < context->usernamePasswordLoginSize; i++) {
-            if(UA_String_equal(&userToken->userName, &context->usernamePasswordLogin[i].username) &&
-               UA_String_equal(&userToken->password, &context->usernamePasswordLogin[i].password)) {
+        if(context->loginCallback) {
+            if(context->loginCallback(&userToken->userName, &userToken->password,
+               context->usernamePasswordLoginSize, context->usernamePasswordLogin,
+               sessionContext, context->loginContext) == UA_STATUSCODE_GOOD)
                 match = true;
-                break;
+        } else {
+            for(size_t i = 0; i < context->usernamePasswordLoginSize; i++) {
+                if(UA_String_equal(&userToken->userName, &context->usernamePasswordLogin[i].username) &&
+                   UA_String_equal(&userToken->password, &context->usernamePasswordLogin[i].password)) {
+                    match = true;
+                    break;
+                }
             }
         }
         if(!match)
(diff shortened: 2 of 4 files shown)


plugins/include/open62541/plugin/accesscontrol_default.h
```diff
@@ -18,6 +18,11 @@ typedef struct {
     UA_String password;
 } UA_UsernamePasswordLogin;
 
+typedef UA_StatusCode (*UA_UsernamePasswordLoginCallback)
+    (const UA_String *userName, const UA_ByteString *password,
+    size_t usernamePasswordLoginSize, const UA_UsernamePasswordLogin
+    *usernamePasswordLogin, void *loginContext);
```
jpfr, line 24: The login context could be a double-pointer.
So the context can be set (replaced) from within the login context.

plugins/include/open62541/plugin/accesscontrol_default.h
```diff
@@ -31,6 +36,11 @@ UA_AccessControl_default(UA_ServerConfig *config,
                          size_t usernamePasswordLoginSize,
                          const UA_UsernamePasswordLogin *usernamePasswordLogin);
 
+UA_EXPORT UA_StatusCode
+UA_AccessControl_setCallback(UA_ServerConfig *config,
```
jpfr, line 40: Why not have `UA_AccessControl_defaultWithLoginCallback` in parallel to `UA_AccessControl_default`?
Then we don't have bugs in the interaction of _default and _setCallback.

jpfr (commented):
Nice, looks good.
Comments inline.

plugins/include/open62541/plugin/accesscontrol_default.h
bluhm: Most contexts are simple void* which are easy to use.  I only encountered double void pointers in UA_GlobalNodeLifecycle and for me this feature makes correct memory management hard.  For my use case, single pointer is enough and standard.  Double pointer feels over engineered, but this impression could change when I would need it.

plugins/include/open62541/plugin/accesscontrol_default.h
bluhm: True.  I remember that I had problems in my tests calling them in the correct order.  I will create a diff, test it and push to this pull request if it works better.

plugins/include/open62541/plugin/accesscontrol_default.h
jpfr: The activateSession API for the plugin is as follows:
```
UA_StatusCode
activateSession(UA_Server *server, UA_AccessControl *ac,
                        const UA_EndpointDescription *endpointDescription,
                        const UA_ByteString *secureChannelRemoteCertificate,
                        const UA_NodeId *sessionId,
                        const UA_ExtensionObject *userIdentityToken,
                        void **sessionContext);
```
The idea is that activateSession adds a context pointer to the session.
I think many users will want to attach the context pointer in the place where the username is evaluated.
So please keep the double-pointer.

plugins/include/open62541/plugin/accesscontrol_default.h
```diff
@@ -31,6 +36,16 @@ UA_AccessControl_default(UA_ServerConfig *config,
                          size_t usernamePasswordLoginSize,
                          const UA_UsernamePasswordLogin *usernamePasswordLogin);
 
+UA_EXPORT UA_StatusCode
+UA_AccessControl_defaultWithLoginCallback(UA_ServerConfig *config,
```
jpfr, line 40: I would either have the loginCallback or the usernamePasswordList, but not both.
Please remove the userNamePasswordList here.

jpfr (commented):
Cool. See the additional comments.

plugins/include/open62541/plugin/accesscontrol_default.h
bluhm: To pass a void pointer to a callback is common pattern in C.  The usual case, where the callback gets some memory to operate on or link to external data structures, is covered.  It is uncommon that some context is needed or globally changed during username and password validation.  But it is still possible with a single void pointer.  Just install a pointer to a struct where the callback has a field which it can change.  A single void pointer makes common things easy and complicated things possible.  So I think single pointer is a good API in this case.

plugins/include/open62541/plugin/accesscontrol_default.h
bluhm: Actually I use the usernamePasswordList to store usernames and salted password hashes.  I just change the algorithm in the callback to compare them.  Also there are some conditions (usernamePasswordLoginSize > 0) that would need extension to reflect a possible callback.

What i really dislike in the current design are the default users that are built in the code.  We got trouble during a security audit with the user1/password and user2/password1 built in your software.  By setting only the callback I do not get rid of there, although they are not used anymore.

In PR #5951 I show what API I would really like.  No build in username/password at all.  But of course that would be an API break.

jpfr: Good point.
There was a PR right now that removed the default username/pw in the default accesscontrol plugin.
Now username/pw has to be added explicitly by renewing the default AccessControl plugin instance.

plugins/include/open62541/plugin/accesscontrol_default.h
jpfr: We should be consistent with our overall API.
This is how the AccessControl plugin API works:
https://github.com/open62541/open62541/blob/bf3b5945da2d744706d626fdf8ca42b39e5ee7ed/include/open62541/plugin/accesscontrol.h#L47

We need the double-pointer, because we want to set a void-pointer that is stored in the session.
But we don't define the session struct publicly.

I think users who use your username callback will want to do the same: Attach some custom data to the session.

The `sessionContext` double-ptr is in addition to the login context.
I would prefer that you forward the server-struct as a pointer instead of the login context.
If the login context is stored in the accesscontrol plugin, then we can get it via the server pointer anyway.

jpfr (commented):
See my comments inline.
Can be merged when they are resoved.

bluhm:
Adding the sessionContext double-ptr is easy.  I do not need it, but if it help others why not.

Accessing loginContext through server config and access control makes it hard to use.  The structures are then part of the API and not opaque to the user anymore.

I have added a server password test to show how I think this new API could be used.

bluhm:
I have rebased my changes to current 1.3 branch.  Some of your suggestions are implemented now.  For others I am not sure if they are useful or if I completely misunderstood what you want.  Could you explain again how my current API differs from what you have in mind?

bluhm:
I only added one public function to the API, data structures that are part of the ABI did not change.  So it is a backwards compatible ABI change.  I have seen more severe ABI breakage in the 1.3 branch.  But if you don't like it for 1.3, I have a 1.4 commit ready.  Which version do you prefer?

jpfr:
Looks good! Merged.

PR #5950, opened 2023-08-16
fix(pubsub): No WriterGroup address with old transportSettings type

keba-estr:
If the old type UA_TYPES_DATAGRAMWRITERGROUPTRANSPORTDATATYPE is used for transportSettings, no address can be used for unicast configuration. Yet a multicast configuration with this type must be possible.


jpfr:
👍

PR #5951, opened 2023-08-16
feat(plugin): use callback for password validation

bluhm:
Username password validation in access control plugin allows only passwords stored in plaintext.  Use a callback to implement custom validation.  This allows salted password hashes and timing safe comparison in the server.  Furthermore remote validation could be implemented.  The function UA_AccessControl_default() is changed to a callback parameter, username password arrays are not accepted anymore.

This is a API break.  I doubt it should be merged.  I just want to show how a clean interface would look like instead of PR #5935 .


bluhm:
This PR was never intended for merge due to API break.

PR #5952, opened 2023-08-17
fix(pubsub): Set reader group to operational

keba-estr:
If the function UA_PubSubConnection_setPubSubState is used with UA_PUBSUBSTATE_OPERATIONAL in state UA_PUBSUBSTATE_PREOPERATIONAL, The state is reset to UA_PUBSUBSTATE_PREOPERATIONAL internally. Thus the state has to be set manually (like in ua_pubsub_eventloop.c) and the user callback has to be called.

fix(tests): Unlock mutex in state callbacks before throwing assert

fix(tests): Reset expected cb counter in check_pubsub_subscribe_msgrcvtimeoout

### DIFF
--- a/src/pubsub/ua_pubsub_reader.c
+++ b/src/pubsub/ua_pubsub_reader.c
@@ -1099,9 +1099,12 @@ UA_ReaderGroup_process(UA_Server *server, UA_ReaderGroup *readerGroup,
     /* Received a (first) message for the ReaderGroup.
      * Transition from PreOperational to Operational. */
     if(readerGroup->state == UA_PUBSUBSTATE_PREOPERATIONAL) {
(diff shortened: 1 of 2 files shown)


keba-estr:
@jpfr: Are the extremely high runtimes caused by my change or due to some CI server issue?

jpfr:
There seems to be an endless loop with the callback during the deardown of check_pubsub_subscribe_msgrcvtimeout.

keba-estr:
> There seems to be an endless loop with the callback during the deardown of check_pubsub_subscribe_msgrcvtimeout.

@jpfr: Might there be an issue if a ck_assert fails in the PubSubStateChangeCallback that the server access mutex is not unlocked?

jpfr:
Nice 👍

PR #5965, opened 2023-08-21
feat(ns): Add a script that checks if the CSV needs to be repaired.

NoelGraf:
The Siome tool generates CSV files which the Nodeset compiler cannot process. The problem occurs when the user generates custom data types in Siome. The DefaultBinary and DefaultXML objects in the csv file do not contain the data type name.

For example:

Encoding_DefaultBinary,40,Object -> cannot be processed by the nodeset compiler

PowerlinkPDOMappingEntryDataType_Encoding_DefaultBinary,40,Object -> can be processed by the nodeset compiler


NoelGraf:
Exports properly in the current SIOME release

PR #5968, opened 2023-08-22
style(server): add missing space in error message

dns13:
Not sure if this is of "style" type as it may affect code by requiring more memory.


jpfr:
Renamed to fit our conventions and merged. Thanks!

Issue #5978, opened 2023-08-24
Wrong order of SecurityHeader and ExtendedNetworkMessageHeader


keba-estr:
In the PubSub specification (chapter 7.2.2.5 DataSetMessage) the Extended Network MessageHeader is displayed before the Security Header. Hence in the function UA_NetworkMessage_decodeHeaders the function UA_SecurityHeader_decodeBinary is called before UA_ExtendedNetworkMessageHeader_decodeBinary (which both increase the offset variable).

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): 1.4
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

keba-estr:
Resolved with PR #5979

PR #5979, opened 2023-08-24
fix(PubSub): fix order of SecurityHeader and ExtendedNetworkMessageHe…

keba-estr:
…ader

In the PubSub specification the SecurityHeader is before the ExtendedNetworkMessageHeader, thus the headers must be decoded in this order.


keba-estr:
Resolves issue #5978

jpfr:
Is this also valid for 1.3? If yes, please rebase and we merge upwards from there.

keba-estr:
> Is this also valid for 1.3? If yes, please rebase and we merge upwards from there.

Yes, this also was a problem on the 1.3 branch. I rebased the feature branch to 1.3.

keba-estr:
@jpfr: CI on 1.3 fails. Is this the correct branch?

Issue #5975, opened 2023-08-23
DataSetReader: data transfer with UA_PUBSUB_RT_FIXED_SIZE not working for raw data encoding


keba-estr:
If a PubSub connection is transferred with UA_PUBSUB_RT_FIXED_SIZE and raw data encoding, the process data is not evaluated correctly.

In the error case the contentType of the cyclic data is UA_PUBSUB_OFFSETTYPE_NETWORKMESSAGE_FIELDENCDODING.
This is because for raw data encoding the keyFrameData.fieldCount is not available. The function UA_DataSetReader_process seems to handle raw data encoding correctly, but the functions prepareOffsetBuffer or UA_NetworkMessage_updateBufferedNwMessage not.

The function prepareOffsetBuffer seems to calculate a wrong byte size, as the process data is not calculated correctly (p->data.keyFrameData.fieldCount is 0 in UA_DataSetMessage_calcSizeBinary)

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): 1.4
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

keba-estr:
Possible solution: (?)
UA_DataSetMessage_calcSizeBinary must consider configured field count in case of UA_FIELDENCODING_RAWDATA. (compare DataSetReader_processRaw)

keba-estr:
Resolved with PR #5980

PR #5980, opened 2023-08-24
fix(PubSub): Fix decoding of rawdata message

keba-estr:
Fix decoding of rawdata message with rtLevel UA_PUBSUB_RT_FIXED_SIZE.

Resolves issue #5975

### DIFF
--- a/src/pubsub/ua_pubsub_connection.c
+++ b/src/pubsub/ua_pubsub_connection.c
@@ -87,7 +87,7 @@ loops_exit:
     }
 #endif
 
-    rv = UA_NetworkMessage_decodePayload(buffer, pos, nm, server->config.customDataTypes);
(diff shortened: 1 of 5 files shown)


keba-estr:
Does not work for multiple connections yet.

jpfr:
Oh nice!
That’s a good contribution.

two wishes:
- extend the unit tests to prevent a future regression of the feature. Somehow this wasn’t caught by the pubsub j it tests that we have already.
- the networkmessage code deliberately didn‘t include ua_pubsub.h. In the future we might even make the message decoding part of the public API. And hence should not depend on internals in the invocation. You only need the dataSetMetaData from the reader. Please only forward that. And not the entire reader.

Keep up the good work!

keba-estr:
Changed interface to use only dataSetMetaData.

keba-estr:
For the UNIT tests the test PublishSubscribeWithWriteCallback in check_pubsub_subscribe_rt_levels.c should be extended with a second transferred variable.

PR #5982, opened 2023-08-25
fix(server): StartTime is missing in minimal NS0

engycz:
StartTime is set for all data models in ua_server.c, but the node is missing in the minimal NS0.


jpfr:
👍

PR #5983, opened 2023-08-25
fix(PubSub): Only check PublisherId if PublisherId is enabled

keba-estr:


jpfr:
Please early-return instead of deep nesting.

keba-estr:
> Please early-return instead of deep nesting.

Done

jpfr:
Cool, merged.

Issue #5981, opened 2023-08-25
PubSub: Invalid check for reader group when receiving multiple messages with RT_FIXED_SIZE


keba-estr:
If multiple connections are transmitted, the function decodeAndProcessRT must check if the reader is matching before updating the buffered network message. The function UA_DataSetReader_checkIdentifier is called for the reader-specific saved network message and not for the actually received data.

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): 1.4
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

keba-estr:
Resolved with PR #5980

keba-estr:
Resolved with PR #5980 and PR #5994.

PR #5994, opened 2023-09-04
fix(PubSub): Check DataSetReader before using saved NetworkMessage

keba-estr:
If a received message is not intended for a DataSetReader, the saved
NetworkMessage must not be used.

### DIFF
--- a/src/pubsub/ua_pubsub_networkmessage.c
+++ b/src/pubsub/ua_pubsub_networkmessage.c
@@ -772,6 +772,26 @@ UA_NetworkMessage_decodeHeaders(const UA_ByteString *src, size_t *offset, UA_Net
 }
 
 UA_StatusCode
+UA_NetworkMessage_decodeHeadersRT(const UA_ByteString *src, UA_NetworkMessage *dst) {
+
+    size_t offset = 0;
+    UA_StatusCode rv = UA_NetworkMessageHeader_decodeBinary(src, &offset, dst);
+    UA_CHECK_STATUS(rv, return rv);
+
+    if(dst->groupHeaderEnabled) {
+        rv = UA_GroupHeader_decodeBinary(src, &offset, dst);
+        UA_CHECK_STATUS(rv, return rv);
+    }
+
+    if(dst->payloadHeaderEnabled) {
+        rv = UA_PayloadHeader_decodeBinary(src, &offset, dst);
+        UA_CHECK_STATUS(rv, return rv);
+    }
+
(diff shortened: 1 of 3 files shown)


keba-estr:
Resolves issue #5981

jpfr:
The change is good - and even reduces the code size!

However it breaks an assumption:
The header decoding is now also done in the RT fastpath.
The header decoding may use malloc - which is why we require to _clear the structure in the end.

I would prefer if the header decoding was also done with the offset buffer.
Then the entire fastpath decoding is a single loop over the offset buffer fields.

We should still check for the identifier of course!

keba-estr:
The header decoding should only be done once for each received packet and not for each reader group. There is no "the offset buffer", but a separate offset buffer for each reader group. The header decoding is necessary to get the publisherId of the current packet, which is needed for comparison with the reader group configuration.

keba-estr:
> How do you suggest getting rid of the potential malloc in the RT path?

For verifyAndDecryptNetworkMessage only the securityHeader is relevant, for decodeAndProcessRT (i.e. UA_DataSetReader_checkIdentifier) only the publisherId, publisherIdType, groupHeader, and the payloadHeader is needed. So there is no need to call UA_ExtendedNetworkMessageHeader_decodeBinary (where the buffer is allocated) in this specific case. I would suggest to define a new function which does not decode the ExtendedNetworkMessageHeader (can also be called by UA_NetworkMessage_decodeHeaders) and call this function in the RT path.
Or am I missing some other allocation?

keba-estr:
The interpretation of ExtendedNetworkMessageHeader is only needed in UA_NetworkMessage_calcSizeBinary, therefore UA_NetworkMessage_decodeHeaders is called in prepareOffsetBuffer, which already allocates the networkMessage.

keba-estr:
Changed implementation according to discussion in PR #5999.

keba-estr:
@jpfr: There would be a possible optimization to save the intermediate offset from UA_NetworkMessage_decodeHeadersRT and use it for a variant of UA_NetworkMessage_decodeHeaders that only interprets the two remaining headers. with UA_ExtendedNetworkMessageHeader_decodeBinary and UA_SecurityHeader_decodeBinary.
Also, if the function UA_NetworkMessage_decodeHeadersRT gets the offset parameter again, it could be called in UA_NetworkMessage_decodeHeaders.
What do you think about this?

keba-estr:
Memory for dataSetWriterIds is still allocated in UA_NetworkMessage_decodeHeadersRT

PR #5996, opened 2023-09-06
Use macros for valueRank

keba-estr:


jpfr:
Thanks!
FYI, I had to add the `refactor(server):` prefix to the commit message.

PR #5999, opened 2023-09-07
WIP: Test different proposal for PR #5994

keba-estr:


src/pubsub/ua_pubsub_networkmessage.c
```diff
@@ -771,6 +771,25 @@ UA_NetworkMessage_decodeHeaders(const UA_ByteString *src, size_t *offset, UA_Net
     return UA_STATUSCODE_GOOD;
 }
 
+UA_StatusCode
+UA_NetworkMessage_decodeHeadersRT(const UA_ByteString *src, size_t *offset, UA_NetworkMessage *dst) {
```
jpfr, line 775: I wouldn't forward the offset as a pointer.
You cannot use it afterwards anyway. Because it will point to the middle of the headers.
So not usable for the decoding afterwards.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1242,50 +1220,63 @@ UA_ReaderGroup_decodeAndProcessRT(UA_Server *server, UA_ReaderGroup *readerGroup
...
+               nm = &currentNetworkMessage;
+            }
+            rv = UA_NetworkMessage_decodeHeaders(buf, &payLoadPosition, nm);
+            if(rv != UA_STATUSCODE_GOOD) {
+                UA_LOG_WARNING_READERGROUP(&server->config.logger, readerGroup,
+                                      "PubSub receive. decoding headers failed");
+                goto done;
+            }
+            /* Decrypt */
+            rv = verifyAndDecryptNetworkMessage(&server->config.logger, buf, &payLoadPosition,
```
jpfr, line 1256: It seems inefficient to decrypt separately for every reader.

jpfr (commented):
A few comments inline.

src/pubsub/ua_pubsub_networkmessage.c
keba-estr: ok

src/pubsub/ua_pubsub_reader.c
keba-estr: The header was already decoded for every packet in the original 1.4 implementation. Now the decoding is only done if the reader group matches. Can there be several reader groups subscribing to the same message?
If yes, then the decryption (and decoding) should only be done for the first matching reader. After this the message is decoded and can be used by all readers.

jpfr: There can be several ReaderGroups sitting on top of the same PubSub-Connection.
Then it depends on their configuration which messages they are processing.
So yes, it would be best to decrypt only when there is at least one match.

You could do a first loop to check for a match, then decode and match again to process.

keba-estr: I would prefer a boolean flag to avoid double loops.

src/pubsub/ua_pubsub_reader.c
```diff
@@ -1242,50 +1220,67 @@ UA_ReaderGroup_decodeAndProcessRT(UA_Server *server, UA_ReaderGroup *readerGroup
...
+        if(rv == UA_STATUSCODE_GOOD) {
+#ifdef UA_ENABLE_PUBSUB_ENCRYPTION
+            if(!decrypted) {
+                UA_NetworkMessage *nm = dsr->bufferedMessage.nm;
+                size_t payLoadPosition = 0;
+                if(!nm) {
+                   /* first call */
+                   nm = &currentNetworkMessage;
+                }
+                rv = UA_NetworkMessage_decodeHeaders(buf, &payLoadPosition, nm);
```
keba-estr, line 1251: If the header is decoded to the buffered message, is the memory for the ext header still allocated in each cycle? Must this memory be freed in between?

keba-estr: According to the valgrind build the decoding of the payload header allocates memory which is not freed. Thus UA_NetworkMessage_decodeHeaders can not be called on the buffered network message without freeing the memory in every cycle.

keba-estr:
#5994 resolved

PR #6000, opened 2023-09-08
Fix CI build 'Debug Build & Unit Tests (gcc, 32bit)'

keba-estr:


keba-estr:
@jpft: The failed test case 'Linux Build & Test / Debug Build & Unit Tests (gcc, 32bit) (pull_request)' works again with this revert of part of the commit 0e2cbabb7f8a70e2971f1821655e6f4b8f06b9ea.
Yet I don't understand the cause for the error.

PR #6007, opened 2023-09-11
fix(core): Include stdlib.h for usage of malloc, free etc.

keba-estr:
Fixes commit 928931ec36d4441fc12d27b11470ea96f7dc9b9b


jpfr:
👍

PR #6012, opened 2023-09-12
Fix building examples

koalo:
The pubsub_realtime did not build properly anymore as described in https://github.com/open62541/open62541/blob/master/examples/pubsub_realtime/README.md
In particular, 928931ec36d4441fc12d27b11470ea96f7dc9b9b has removed including standard headers from the core headers and that now seems to ripple through the code basis.

Also added UA_ENABLE_PUBSUB_ETH_UADP that is referred to in the README to the CMakeLists.txt. I am not sure if that was missing on purpose or if there is any policy which flags should be included?


jpfr:
Thanks.
Will merge when green.

the real-time examples will be moved over to this soonish: #5992

then we have proper support for sending with txtime and don’t have to reinvent the wheel for every example individually.

koalo:
Thanks for the hint! Currently, I am mostly looking into TAPRIO alone, so TXTIME from the middleware is not essential for me at the moment, but I will definitely have a look again later.

jpfr:
👍

PR #6014, opened 2023-09-19
fix(server): Return publishing interval if sampling interval == -1.0

cldhms:
cherry-picked from 1.4 but kept the check that verifies that the publishing interval is a valid sampling interval.
Without this fix several CTT test cases in "Monitored Item Services" failed.

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  *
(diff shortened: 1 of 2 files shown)


cldhms:
I kept the validation that the publishing interval is a valid sampling interval in checkAdjustMonitoredItemParams(). This check has been removed in the 1.4 branch and the master branch for some reason. I think this is a potential bug in those branches as the publishing interval ending up being used may not be a valid sampling interval

jpfr:
Cool, merged.
This will be part of the next patch release.

jpfr:
I cannot think of a reason why a valid publish interval should not be a valid sampling interval.
We can re-add the test when users report problems…

cldhms:
It's probably not likely but the specification mentions that it should be considered.

Part 4 1.05 Table 144 - MonitoringParameters:

> The value -1 indicates that the default sampling interval defined by the publishing interval of the Subscription is requested. A different sampling interval is used if the publishing interval is not a supported sampling interval.
> 

But then again, it's probably not likely and may not be an issue for anyone. But the check was in the code originally and has now been removed in the 1.4/master branch.

PR #6023, opened 2023-09-25
build(arch): disable compiler error unused variable

bluhm:
If the macro UA_LOCK_ASSERT() is not generating code, the variable UA_EventLoopPOSIX *el is not used.  OpenBSD 7.4 clang version 13.0.0 generates a warning in this case which aborts the build due to -Werror.  Compile with -Wno-unused-variable instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -298,9 +298,6 @@ option(UA_FORCE_32BIT "Force compilation as 32-bit executable" OFF)
 mark_as_advanced(UA_FORCE_32BIT)
 
 option(UA_FORCE_WERROR "Force compilation with -Werror (or /WX on MSVC)" OFF)
-if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-    set(UA_FORCE_WERROR ON)
-endif()
 
 option(UA_ENABLE_DEBUG_SANITIZER "Use sanitizer in debug mode" ON)
 mark_as_advanced(UA_ENABLE_DEBUG_SANITIZER)
--- a/plugins/ua_config_json.c
+++ b/plugins/ua_config_json.c
@@ -790,6 +790,7 @@ PARSE_JSON(SecurityPkiField) {
     }
 #ifndef __linux__
     /* Currently not supported! */
+    (void)field;
     return UA_STATUSCODE_GOOD;
 #else
     /* set server config field */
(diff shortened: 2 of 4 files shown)


jpfr:
In general we want the warning for unused varaibles.
And also the build failing.

We could do the following:
Check if the build-type is "Release" and if so disable the `UA_FORCE_WERROR` flag by default.

So we can have the benefits in our build pipeline.
And you can build packages with release settings that are not affected by compiler changes.

bluhm:
It seems that none of your CI platforms checks the unused variable by default.
I cannot build open62541 git checkout 1.4 on my OpenBSD.
Setting this in a release build does not fix my problem, I want to compile it on OpenBSD with and without debugging.

You could also add -Wunused-variable to your build pipeline and fix issues with that.
Currently it fails here, but I expect more problems.
[  2%] Building C object CMakeFiles/open62541-plugins.dir/arch/eventloop_posix_udp.c.o
/home/bluhm/github/open62541/arch/eventloop_posix_interrupt.c:208:24: error: unused variable 'el' [-Werror,-Wunused-variable]
    UA_EventLoopPOSIX *el = (UA_EventLoopPOSIX*)singletonIM->im.eventSource.eventLoop;
                       ^

Running compiler with -Werror makes porting C code hard, as the porter has to deal with various issues that are valid C code, but some compiler considers it suboptimal.

1. Is -Wunused-variable and  -Werror a goal of our project?
2. If yes, can you enable and fix it for you CI?
3. Or is it just a corner case and I have to deal with it on OpenBSD somehow?
4. Then disabling it would be fine for me.  That's what my PR does.  I don't think an if (openbsd) makes it better.

Building OpenBSD packages is not a problem at all.  There can patch the build system in many ways.
https://cvsweb.openbsd.org/ports/misc/open62541/patches/?hideattic=1
I try to merge things upstream to keep maintenance burden low.  Most of the patches are in some version, but not in 1.3.

jpfr:
In that case, let's do the following:
Disable `UA_FORCE_WERROR` by default everywhere, but add it to all CI builds.
The CI builds are defined in `/tools/ci.sh` and in `/tools/azure-devops/win/build.sh`.

jpfr:
Please do not remove UA_FORCE_WERROR, but change the default to OFF.
Otherwise the CI changes will have no effect either…

bluhm:
Like this?

jpfr:
👍

the windows build seems to fail…

jpfr:
Nice. Merged.

PR #6029, opened 2023-09-26
[REVIEW]feat(server): Added support for the AccessLevelEx attribute.

NoelGraf:
For servers implemented according to specification 1.04, this attribute is required for variables. The attribute is an extension of the AccessLevel attribute and must therefore have at least the same bits set as the AccessLevel attribute.

The server returns BadNotSupported for some attributes like RolePermissions or AccessRestrictions. Unsupported/not allowed attributes must be answered with BadAttributeIdInvalid.

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -522,6 +522,16 @@ ReadWithNode(const UA_Node *node, UA_Server *server, UA_Session *session,
         retval = UA_Variant_setScalarCopy(&v->value, &node->variableNode.accessLevel,
                                           &UA_TYPES[UA_TYPES_BYTE]);
         break;
+    case UA_ATTRIBUTEID_ACCESSLEVELEX:
+        CHECK_NODECLASS(UA_NODECLASS_VARIABLE);
+        /* The normal AccessLevelEx contains the lowest 8 bits from the normal AccessLevel.
+         * In our case, all other bits are zero. */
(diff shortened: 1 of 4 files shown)


jpfr:
Split the PR into two PR for client and server. The client is good and can go in as-is.

For the server, do not add a new attribute to the nodes. The AccessLevelEx contains the lowest 8 bits from the normal AccessLevel.

in our case, all other bits are zero.
See here for the definition: https://reference.opcfoundation.org/Core/Part3/v104/docs/8.58

src/server/ua_services_attribute.c
```diff
@@ -522,6 +522,17 @@ ReadWithNode(const UA_Node *node, UA_Server *server, UA_Session *session,
         retval = UA_Variant_setScalarCopy(&v->value, &node->variableNode.accessLevel,
                                           &UA_TYPES[UA_TYPES_BYTE]);
         break;
+    case UA_ATTRIBUTEID_ACCESSLEVELEX:
+        CHECK_NODECLASS(UA_NODECLASS_VARIABLE);
+        /* The normal AccessLevelEx contains the lowest 8 bits from the normal AccessLevel.
+         * In our case, all other bits are zero. */
+        UA_Byte value = (UA_Byte)(*(UA_UInt32*)v->value.data & 0xFF);
```
jpfr, line 529: This code looks fishy.
You want to wrap `node->variableNode.accessLevel` in a UInt32 and return that.
But you return a Byte. Wut?

jpfr (commented):
See inline.

PR #6033, opened 2023-09-30
Simplify PubSub state machine

jpfr:


keba-estr:
@jpfr: Are these changes planned to also go into the 1.4 branch? At least bugfixes (e.g. signal timeout errors if DataSetReader is not enabled) would also be relevant for 1.4 branch.

jpfr:
We can try to backport.
But this builds on previous PR that are only on master as well.

PR #6039, opened 2023-10-03
feat(server) enable automatic datatype adjust logic

andreasebner:
This PR solves an issue with compatible value types. Currently, the server checks if a value type is compatible and tries afterwards to adjust the type if needed. For example, the types duration and double are seen as equal. This results in an unexpected behavior because the adjust logic is not executed in some cases.

Scenario: 
Server with a method containing a duration array as input argument. 
A client calls the method. -> The Method-callback receives a double-array since the adjust logic is not called and the type is transmitted as a double-array.

This PR solves the issue by changing the default behavior to an automatic type adjust and a single compatible check afterwards.

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -1443,14 +1443,11 @@ writeNodeValueAttribute(UA_Server *server, UA_Session *session,
 
     /* Type checking. May change the type of editableValue */
     const char *reason;
-    if(value->hasValue && value->value.type &&
-       !compatibleValue(server, session, &node->dataType, node->valueRank,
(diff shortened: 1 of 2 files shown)


jpfr:
Great, the logic is much cleaner now.

In addition we can make a speed improvement.
`adjustDataType` first performs a linear search with `UA_findDataTypeWithCustom`.
That does hundreds of NodeId comparisons.

Since we no longer check for compatibility before adjusting, we can add a "fast exit" to the adjust method.
That is, if the target datatype was already achieved, quit.
That is much less than the previos compat check, but will speed up 99% of cases...

jpfr:
Cherry-picked into PR #6098 and merged.

PR #6041, opened 2023-10-04
[REVIEW] Added ability to configure server with secure policies only

NoelGraf:
Added start parameters for disabling Aes policies in the server_ctt.
Added start parameters for enabling None and deprecated policies in the server_ctt.
Added new function that configures the server with secure security policies only.


jpfr:
👍

Issue #6040, opened 2023-10-04
Deadlock calling UA_Server_addPubSubConnection from low priority task


keba-estr:
If UA_Server_addPubSubConnection is called in a (real-time) linux system from a task with lower priority than the task that calls UA_Server_run, a deadlock occurs.

## Background Information / Reproduction Steps

The function UA_Server_run waits for a event with UA_EventLoopPOSIX_pollFDs (epoll_wait) with the locked event loop mutex, while the function UA_Server_addPubSubConnection calls UA_PubSubConnection_connectUDP and UDP_openConnection which tries to lock the event loop mutex.

Adding the mutex option PTHREAD_PRIO_INHERIT does not help, either.
pthread_mutexattr_setprotocol(&mutexName##_attr, PTHREAD_PRIO_INHERIT);

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

Please provide the following information:

 - [ X ] open62541 Version (release number or git tag): 1.4
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system: linux
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

keba-estr:
Is it possible/safe to call UA_EventLoopPOSIX_pollFDs with unlocked eventloop mutex?

keba-estr:
Priority inheritance does not solve the problem, because only the priority of the task that locks the mutex is increased to a task waiting for the lock, not vice verse.

keba-estr:
@jpfr: Reguarding your latest refactoring PRs, did you already face this problem that the event loop waiting condition is done in locked state?
Is there a safe point where the lock can be opened before waiting inside UA_EventLoopPOSIX_pollFDs (or the 'select' use case)?

keba-estr:
For the 'select' use case, the call of UA_select without locked EventLoop mutex should suffice to resolve the priority problem, as all prameters of UA_select are local variables. Is it possible that the EventLoop itself gets invalid after calling UA_select (i.e. when locking again)?

keba-estr:
For the 'epoll' use case, the epollfd must be stored in a local variable to call epoll_wait without locked mutex. Save question as above about invalid EventLoop after call raises.

[...]

PR #6042, opened 2023-10-05
[REVIEW] fix(el): Unlock EventLoop mutex before wait condition

keba-estr:
If UA_Server_addPubSubConnection was called in a (real-time) linux system from a task with lower priority than the task that called UA_Server_run, a deadlock occured.

### DIFF
--- a/arch/eventloop_posix_epoll.c
+++ b/arch/eventloop_posix_epoll.c
@@ -70,7 +70,9 @@ UA_EventLoopPOSIX_pollFDs(UA_EventLoopPOSIX *el, UA_DateTime listenTimeout) {
 
     /* Poll the registered sockets */
     struct epoll_event epoll_events[64];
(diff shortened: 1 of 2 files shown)


keba-estr:
Resolves issue #6040

jpfr:
The lock has to go around `rfd->eventSourceCB` for both select and epoll.
I will have a look later how we can protect connections being closed from within a neighbor callback...

keba-estr:
> eventSourceCB

As after epoll_wait and UA_select the lock is set again, also 'rfd->eventSourceCB' is locked.

keba-estr:
For opening an additional connection, see notes on epoll_wait man page:

       While one thread is blocked in a call to epoll_wait(), it is
       possible for another thread to add a file descriptor to the
       waited-upon epoll instance.  If the new file descriptor becomes
       ready, it will cause the epoll_wait() call to unblock.

Yet, I did not find any notes about deleting a connection.

PR #6052, opened 2023-10-10
fix(core): LibreSSL build on win32

ccvca:
As a recent mingw version is required for the build, the libressl hack will cause problems and is no longer required for recent libressl and mingw version. (MinGW with gcc 8 was not able to build open62541 anymore, upgrade to recent mingw with gcc 13 requires these fixes)


jpfr:
Nice!

PR #6054, opened 2023-10-11
feat(client): Allow setting a custom SessionName in the client config

jpfr:


RonnieFTI:
I'm having trouble with a session being created with ignition using open 62541.

PR #6064, opened 2023-10-16
Added startup parameters to disable x509 or username/password authentication in server_ctt example.

NoelGraf:


jpfr:
There is a new valgrind failure in the CI.
Looks good otherwise!

Issue #3178, opened 2019-10-15
UA_ENABLE_AMALGAMATION for .c and .h data generated by XML


prrw:
## Description
The header file generated from the XML using the Nodeset Compilor has the following code inside:
```
#ifdef UA_ENABLE_AMALGAMATION
# include "open62541.h"
#else
# include <open62541/server.h>
#endif
```
While compiling, since UA_ENABLE_AMALGAMATION is not defined for the .c and .h data, it always enters the else loop and asks for `<open62541/server.h>`. Hence it has to be explicitly defined before using
```
#define UA_ENABLE_AMALGAMATION
```
 But this is not the right way since, these files are generated form XML and it means this has to be editted every time it is generated. Is there any other way to solve this problem ?

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_AMALGAMATION=TRUE
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag):  https://github.com/open62541/open62541/commit/6574a46bd11d1c933930b2f638b9efd5ac4c10a0
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system:  Windows 10 with Microsoft Visual Studio 2015
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

prrw:
okay. I think this problem is because of the added .c and .h file due to XML. Because we compile the XML files first and it does not know if we have used UA_ENABLE_AMALGAMATION. Thats why it assumes it as undefined.

teccheck:
The merge request was merged, so I assume this issue to be done

PR #6067, opened 2023-10-17
build(pkgconfig): export amalgamation flag

panicgh:
While a UA_ENABLE_AMALGAMATION macro is defined in the amalgamated open62541.h file and undefined in the non-amalgamated open62541/config.h file, it is not usable in downstream projects as they would need a priori knowledge which of the header files to include. They may try a test compilation in their build system's configure phase with one of the headers and set UA_ENABLE_AMALGAMATION if amalgamation is detected. However, this causes some code duplication in each downstream project.

Exporting the flag in the pkg-config file enables downstream projects to check for the open62541 dependency as usual and automatically get the necessary flag.

This is slightly related to #3178, although just concerning an import of open62541 via pkg-config.
In case of cmake, it would probably (I'm not very familiar with cmake) be done via `target_compile_definitions(open62541 INTERFACE UA_ENABLE_AMALGAMATION)` ([manual](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html)).

It would also be nice if this patch could be ported to the 1.4 branch, so that the fixed pc file becomes available with the 1.4.0 release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -732,6 +732,9 @@ file(MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/src_generated")
 configure_file(include/open62541/config.h.in ${PROJECT_BINARY_DIR}/src_generated/open62541/config.h)
 
 # Generate a .pc file for pkg-config
+if(UA_ENABLE_AMALGAMATION)
+    set(PC_EXTRA_CFLAGS "-DUA_ENABLE_AMALGAMATION")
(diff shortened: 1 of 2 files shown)


jpfr:
Looks good.
I just changed the target of this PR to the 1.4 branch.
Please rebase to get rid of the extra commits.

jpfr:
Cool. Merged with a slightly adjust commit message.

PR #6068, opened 2023-10-18
Support defining UA_LOGLEVEL in an external CMakefile

bisbastuner:
Allows to compile the library as a amalgamated, and then change the UA_LOGLEVEL in the CMakefile of the project that's using the amalgamated files.


jpfr:
Cool.
Slightly adjusted the commit title.

bisbastuner:
Could this be included in an official release? I thought this would be in the next major release, but 1.4.0 is out and it doesn't have this PR

PR #6076, opened 2023-10-20
[Review] Fix zero sampling interval handling

irostagno:
Solves sampling interval zero handled as positive sampling interval.

Zero sampling interval was introduced in 46a12f16a6357dc692a30e07c7e16611fa3e99a7 and negative sampling interval was introduced later in 05951d59732238414fd9ddbb6d44e0c84fd7ab3e. Second commit breaks zero sampling interval.


jpfr:
Your commit was merged as part of #5946.
Thanks again for the fix!

PR #6086, opened 2023-10-23
fix(server): Bug fixes in the subscription transfer service.

NoelGraf:
* [x] Fixed a bug where after transferring a subscription and deleting the old session, the publish callback was no longer called.
* [x] Subscriptions that were transferred to another Session must be deleted by the Client that owns the Session.
* [x] Implementation of the allowTransferSubscription function. The function must check if a transfer is allowed. Currently it always returns true.

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -384,12 +384,14 @@ static void
 sendStatusChangeDelete(UA_Server *server, UA_Subscription *sub,
                        UA_PublishResponseEntry *pre) {
     /* Cannot send out the StatusChange because no response is queued.
-     * Delete the Subscription without sending the StatusChange. */
+     * Delete the Subscription without sending the StatusChange, if the statusChange is Bad*/
     if(!pre) {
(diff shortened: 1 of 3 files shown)


src/server/ua_subscription.c
```diff
@@ -384,12 +384,14 @@ static void
...
     if(!pre) {
         UA_LOG_DEBUG_SUBSCRIPTION(&server->config.logger, sub,
-                                  "Cannot send the StatusChange notification. "
-                                  "Removing the subscription.");
-        UA_Subscription_delete(server, sub);
+                                  "Cannot send the StatusChange notification because no response is queued.");
+        if(UA_StatusCode_isBad(sub->statusChange)) {
+            UA_LOG_DEBUG_SUBSCRIPTION(&server->config.logger, sub, "Removing the subscription.");
+            UA_Subscription_delete(server, sub);
+        }
```
jpfr, line 394: In case where the subscription is not deleted we still should disable it.
Otherwise it continues to send out keepalives, etc.

jpfr: Hmm k.

PR #6108, opened 2023-11-02
feat(client): add UA_Client_sendAsyncBrowseNextRequest

bluhm:
The function UA_Client_sendAsyncBrowseNextRequest() was missing in the API.  In older open62541 versions the user of the library could implement it using UA_Client_sendAsyncRequest().  But as this function is gone in version 1.4 the client highlevel async API should be complete.


jpfr:
Cool

PR #6113, opened 2023-11-07
Porting to ARM Ertec with eCos

Isarhamster:
Porting to eCos with GCC C89 compiler
Generate symbolic defines so that the application does not have to use magic numbers

### DIFF
--- a/arch/eCos/ua_clock.c
+++ b/arch/eCos/ua_clock.c
@@ -6,38 +6,28 @@
 
 #ifdef UA_ARCHITECTURE_ECOS
 
-#include <open62541/types.h>
-#include <pkgconf/kernel.h> /* Needed for HAL timer resolution */
-
-UA_DateTime UA_DateTime_now(void) {
-  return UA_DateTime_nowMonotonic();
+UA_DateTime UA_DateTime_now(void) 
+{
+	unsigned long long t_ms = ertec_localtime();	
+	return (t_ms * UA_DATETIME_MSEC) + UA_DATETIME_UNIX_EPOCH;
 }
 
 /* Credit to https://stackoverflow.com/questions/13804095/get-the-time-zone-gmt-offset-in-c */
 UA_Int64 UA_DateTime_localTimeUtcOffset(void){
-  time_t gmt, rawtime = time(NULL);
+  time64_t gmt, rawtime = ertec_time_s(NULL);
   struct tm *ptm;
   struct tm gbuf;
-  ptm = gmtime_r(&rawtime, &gbuf);
+  ptm = gmtime64_r(&rawtime, &gbuf);
   // Request that mktime() looksup dst in timezone database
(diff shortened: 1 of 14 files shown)


tools/nodeset_compiler/backend_open62541.py
```diff
@@ -134,8 +135,19 @@ def writeh(line):
 
     def writec(line):
         print(unicode(line), end='\n', file=outfilec)
+        
+    def parentNodeString(node,L):
```
jpfr, line 139: This can be simplified

```python
    def parentNodeString(node,L):
        if node is None or node.parent is None:
            return ""
        return str(parentNodeString(node.parent,L)) + "_" + str(node.displayName).upper()
```

arch/eCos/ua_clock.c
```diff
@@ -24,9 +25,12 @@ UA_Int64 UA_DateTime_localTimeUtcOffset(void){
   return (UA_Int64) (difftime(rawtime, gmt) * UA_DATETIME_SEC);
 }
 
+#define TICKS_TO_MS(Ticks)  \
```
jpfr, line 28: This file only exists in the 1.3 branch and before.

We refactored quite a lot of the architecture integration API.
We will bring this back when we can.
But we didn't have the tooling to compile against eCos during the time when the API was revised.

jpfr (commented):
Looks good!

Please make this PR against the 1.3 branch.
From there we will forward-port to the 1.4 branch and then to master.

arch/network_tcp.c
```diff
@@ -50,6 +50,7 @@ connection_releaserecvbuffer(UA_Connection *connection,
     UA_ByteString_clear(buf);
 }
 
+#ifdef poll
```
JohannesKauffmann, line 53: Isn't it possible to disable the poll bits of the function by ifdef's, so that only 3 + 4 lines are disabled instead of completely duplicating the function?

include/open62541/plugin/log.h
```diff
@@ -52,103 +52,109 @@ typedef struct {
...
 
     void *context; /* Logger state */
 
     void (*clear)(void *context); /* Clean up the logger plugin */
 } UA_Logger;
 
-static UA_INLINE UA_FORMAT(3,4) void
-UA_LOG_TRACE(const UA_Logger *logger, UA_LogCategory category, const char *msg, ...) {
+static UA_INLINE UA_FORMAT(4,5) void
+UA_LOG_TRACE2(const UA_Logger *logger, UA_LogCategory category, unsigned int line, const char *msg, ...) {
```
JohannesKauffmann, line 64: When adding an argument here, you should probably also add `(void) line;` in case UA_LOGLEVEL > 100. Ditto for the functions below.

andreasebner:
Thanks for the contribution. Since the eCos arch was never brought back after the eventloop integration this PR is stall.

PR #6116, opened 2023-11-08
Fix monitoreditem with negative sampling interval

jpfr:


basyskom-jvoe:
@jpfr As a (side ?) effect of [d3ac911](https://github.com/open62541/open62541/pull/6116/commits/d3ac911c08c32e6d5ea7074e3b91ab23fe1fcd25), values stored in the in-memory history database via the UA_HISTORIZINGUPDATESTRATEGY_VALUESET strategy no longer get a default source timestamp.
Is this the intended behavior or should we implement this in the history plugins?

PR #6142, opened 2023-11-24
[REVIEW] EventFilter Query language

FlorianDue:
Query Language to create OPC UA EventFilter from string queries.


jpfr:
Wow. More than 4.000 lines of code.
At a first glance:
- Add a license header to each file that is not auto-generated.
- Look at the CI pipeline results to make every compiler happy.

Looking forward to this. 👍

jpfr:
Some builds still throw warnings.

grafik

FlorianDue:
windows checks are now finally successful

plugins/eventfilter_parser/eventfilter_parser_grammar.leg
```diff
@@ -0,0 +1,199 @@
...
+  } else {                                     \
+      result = yyctx->input.length - yyctx->pos;\
+      if(result > max_size)                     \
+          result = max_size;                   \
+      memcpy(buf, &yyctx->input.data[yyctx->pos], result); \
+      yyctx->pos += result;                    \
+  }                                            \
+}
+
+UA_Counters ctr;
```
jpfr, line 30: Global variables are dangerous.
It could be that a user of the open62541 uses parallel threads.
Then we would have to use locks within the parsing to protect against parallel access of the parsing.
It would be much more elegant, if these global variables were part of some individual "context" that is passed into the internal parsing.

jpfr (changes requested):
Thanks for the hard work. Looking nice!

Inline are my comments.
Just contact me if you need further clarification.

src/util/eventfilter_parser/eventfilter_parser_remove_compiler_warnings.txt
```diff
@@ -0,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jpfr, line 1: What is the intended purpose of this file?

src/util/eventfilter_parser/example_queries/case_0.txt
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jpfr, line 1: This should be in the /examples folder.
Also, use the CC0 license for this.
Otherwise people might wonder if they are allowed to build upon the examples you provide for their own queries.

src/util/eventfilter_parser/eventfilter_parser.h
```diff
@@ -0,0 +1,133 @@
...
+    UA_LiteralOperand literal;
+} UA_Local_Operand;
+
+typedef struct counters{
+    size_t branch_element_number;
+    size_t for_operator_reference;
+    size_t operand_ctr;
+}UA_Counters;
+
+UA_EXPORT void save_string(char *str, char **local_str);
```
jpfr, line 85: This is now all internal.
You can remove the UA_EXPORT prefix.

jpfr (commented):
Cool!
A few more observations inline.

src/util/eventfilter_parser/eventfilter_parser_remove_compiler_warnings.txt
FlorianDue: The compiler throws some errors when using the autogenerated eventfilter_parser_grammar.c file from the eventfilter_parser_grammar.leg. This file should give some hints on how to correct the compiler errors.

FlorianDue:
Unit tests are still missing. I will add them by the beginning of next week.

FlorianDue:
Test are finally passed. @jpfr do you want to do a final review? otherwise ill close this PR and create one with a clean commit history that can be merged

FlorianDue:
replaced with #6327

PR #6149, opened 2023-11-29
[WIP]feat: OPC UA Role Permissions

AsishGanesh1:
Attach well known roles to session

 - [X] Add 8 well known role details to session instance
         UA_Server_getSessionAttribute API can be made use of for accessing the role data
 - [X] Remove redundant API's (like hasAccessToNode, hasAccessToMethod)
 - [X] Remove read user AccessLevel/Executable/RolePermissions related methods
         To avoid read outside the context of a session

 **TODO:**
 - [ ] Add CMake flag for usage of role based permission
 - [ ] Add test application
 - [ ] Address remaining review comments
 - [ ] Rebase to current master head

<asish@eclatron.com>


AsishGanesh1:
This pull request is build on top of #5636.
Track it here for further updates.

src/server/ua_session.h
```diff
@@ -57,6 +57,9 @@ typedef struct {
     size_t localeIdsSize;
     UA_String *localeIds;
 
+    /* Role information */
+    UA_NodeId role;
```
jpfr, line 61: Does a session have only a single role?
I would guess that for fine-grained roles we would like to have several roles per session.

CMakeLists.txt
```diff
@@ -186,7 +186,7 @@ endif()
 set(UA_MULTITHREADING ${MULTITHREADING_DEFAULT} CACHE STRING
     "Multithreading support (<100: No multithreading support, >=100: Thread-safety enabled (internal mutexes)")
 
-option(UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS "Enable the use of A&C. (EXPERIMENTAL)" OFF)
```
jpfr, line 189: A&C doesn't have much to do with the roles.
Let's look at the status of A&C in a second step.

include/open62541/client_highlevel.h
```diff
@@ -454,6 +464,16 @@ UA_Client_writeUserExecutableAttribute(UA_Client *client, const UA_NodeId nodeId
                                       &UA_TYPES[UA_TYPES_BOOLEAN]);
 }
 
+UA_StatusCode UA_EXPORT
+UA_Client_writeRolePermissionsAttribute(UA_Client *client, const UA_NodeId nodeId,
+                                       size_t newRolePermissionSize,
+                                       const UA_RolePermissionType *rolePermission);
+
+UA_StatusCode UA_EXPORT
```
jpfr, line 472: My guess is that the UserRolePermissions are auto-computed from the roles of the user and the permissions of these roles.
Similar to AccessRights vs UserAccessRights.
So this is not something that can be written, only read.
Check if `writeUserRolePermissionsAttribute` makes sense or should be removed.

include/open62541/plugin/accesscontrol.h
```diff
@@ -80,27 +81,36 @@ struct UA_AccessControl {
     /* Allow adding a node */
     UA_Boolean (*allowAddNode)(UA_Server *server, UA_AccessControl *ac,
                                const UA_NodeId *sessionId, void *sessionContext,
-                               const UA_AddNodesItem *item);
+                               const UA_AddNodesItem *item, UA_RolePermissionType *userRolePermission, size_t userRoleSize);
```
jpfr, line 84: I think we can keep the current API intact.
By adding a lightweight method where we can look up the userRolePermissions from the session id.
So `allowNode` can read the permissions only when needed.

include/open62541/plugin/accesscontrol.h
```diff
@@ -80,27 +81,36 @@ struct UA_AccessControl {
...
-                                       const UA_DeleteReferencesItem *item);
+                                       const UA_DeleteReferencesItem *item, UA_RolePermissionType *userRolePermission, size_t userRoleSize);
 
     /* Allow browsing a node */
     UA_Boolean (*allowBrowseNode)(UA_Server *server, UA_AccessControl *ac,
                                   const UA_NodeId *sessionId, void *sessionContext,
-                                  const UA_NodeId *nodeId, void *nodeContext);
+                                  const UA_NodeId *nodeId, void *nodeContext, UA_RolePermissionType *userRolePermission,size_t userRoleSize);
+
+    /* Check access to Node */
```
jpfr, line 106: Add more documentation.
How is `access` different from allowRead and allowBrowse?

include/open62541/plugin/accesscontrol.h
```diff
@@ -127,6 +137,8 @@ struct UA_AccessControl {
                                                       UA_DateTime endTimestamp,
                                                       bool isDeleteModified);
 #endif
+    UA_Boolean (*checkUserDatabase)(const UA_UserNameIdentityToken *userToken, UA_String *roleName);
```
jpfr, line 140: What does this do?
Validate that the user has a certain role?

include/open62541/plugin/nodestore.h
```diff
@@ -411,6 +412,11 @@ struct UA_NodeHead {
     UA_MonitoredItem *monitoredItems; /* MonitoredItems for Events and immediate
                                        * DataChanges (no sampling interval). */
 #endif
+
+    size_t rolePermissionsSize;
```
jpfr, line 416: It seems to be a lot of overhead to manage all role-permissions for each node.
Does the standard define a way how permissions propagate along the hierarchical references?

The best solution would be to inherit standard role-permissions for every new node based on the parents.
And require manual role definitions only as override to the inherited permissions.

include/open62541/plugin/nodestore.h
```diff
@@ -411,6 +412,11 @@ struct UA_NodeHead {
...
+    UA_RolePermissionType *rolePermissions;
+    size_t userRolePermissionsSize;
```
jpfr, line 418: The user-role-permissions are the result of compating the userroles and the role-permissions.
They are different for every user.
So it doesn't make sense to store them in the node for all users at once.

jpfr (commented):
Nice! Good to see a contribution from your end.
You have to help us a bit getting our head around this part of the spec.
Comments inline.

src/server/ua_session.h
AsishGanesh1: I've incorporated managing multiple roles per session.

include/open62541/client_highlevel.h
AsishGanesh1: `writeUserRolePermissionsAttribute` is removed

include/open62541/plugin/accesscontrol.h
AsishGanesh1: These API's are removed.

include/open62541/plugin/accesscontrol.h
AsishGanesh1: `checkUserDatabase` is not required and is removed.

include/open62541/server.h
```diff
@@ -711,6 +713,13 @@ UA_Server_readAccessLevel(UA_Server *server, const UA_NodeId nodeId,
                             outAccessLevel);
 }
 
+static UA_INLINE UA_THREADSAFE UA_StatusCode
+UA_Server_readUserAccessLevel(UA_Server *server, const UA_NodeId nodeId,
```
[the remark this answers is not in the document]
AsishGanesh1: This method is removed.

plugins/ua_accesscontrol_default.c
```diff
@@ -7,6 +7,7 @@
  */
 
 #include <open62541/plugin/accesscontrol_default.h>
+#include <server/ua_server_internal.h>
```
jpfr, line 10: The server_internal.h is not public for the plugins to use…

src/server/ua_session.h
jpfr: Nice.
Since every role gets a power of two 2,4,8,16,32,… we can use a 64Bit integer as bitmap for all the roles. That is more efficient also for comparing inside the nodes..

src/server/ua_services.h
```diff
@@ -50,6 +50,27 @@ typedef void (*UA_Service)(UA_Server*, UA_Session*,
 typedef void (*UA_ChannelService)(UA_Server*, UA_SecureChannel*,
                                   const void *request, void *response);
 
+typedef enum {
```
jpfr, line 53: Where are custom roles assigned to a number?

jpfr (commented):
Very nice.
I am still missing how the access control plugin can access the roles via the API.
That should be possible as well.

i would like to split this into two PR.
The first adds roles to the sessions and exposes them to accesscontrol.
The second one adds roles to the nodes as well to match.

andreasebner:
Thank's for shareing this ideas. A RBAC implementation is part of the latest master.

PR #6158, opened 2023-12-04
fix(pubsub): Evaluate status code of addCustomCallback callback

keba-estr:
Evaluate status code of addCustomCallback callback and UA_WriterGroup_addPublishCallback

With commit f462b9f80788b748a499f321e72a029a20fb15ac the custom cyclic callbacks have been called again, but not evaluated.


jpfr:
Good find

PR #6171, opened 2023-12-09
fix(client): honor endpoint security policies

### LINKED ISSUE #6167: Client Does Not Honor SecurityLevel in Endpoint Selection
The client matches the first suitable endpoint in the list returned by the server without reference to the `SecurityLevel`. In the attached example, the server returns three endpoints, with security policies of None, Basic256, and Basic256Sha256, all of which are available to the client. These have increasing security levels, which means the client should pick Basic256Sha256, but it instead picks None. The only way to override this is to force an encrypted connection via `securityMode` in the config structure, but this requires _a priori_ knowledge of the fact that the server will indeed support this. The fix is pretty trivial: iterate through the entire list, maintaining a record of the best policy found to date and allowing it to be overridden by better ones. I've coded a fix locally, and will submit if you like.

## Attachments

[files.zip](https://github.com/open62541/open62541/files/13583661/files.zip)

## Checklist

 - [X] Other OPC UA SDKs used (client or server): Ignition 8.1.4 OPC UA Server
 - [X] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [X] Wireshark network dump attached
 - [ ] Critical issue

MikeGranby:
Pick endpoint based on highest security level rather than using first

Closes #6167

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -808,14 +808,20 @@ responseGetEndpoints(UA_Client *client, void *userdata,
     /* Warn if the Endpoints look incomplete / don't match the EndpointUrl */
     Client_warnEndpointsResult(client, resp, &client->discoveryUrl);
 
-    UA_Boolean endpointFound = false;
-    UA_Boolean tokenFound = false;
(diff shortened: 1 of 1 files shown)


jpfr:
Cool!
Rebased on the 1.4 branch.

PR #6172, opened 2023-12-09
fix(client): failure to switch from discovery url

### LINKED ISSUE #6165: Creating Session on Discovery Channel Fails on Ignition Server
When connecting to an Ignition OPC UA server, the client first opens a secure channel to the supplied URL and invokes FindServers. This returns the discovery URL, which is then used to open a new secure channel on which the FindEndpoints service is executed. This returns the (non-discovery) URL of the endpoint, but the client continues with the existing channel, attempting to create a session. This fails with BadServiceUnsupported, presumably because sessions cannot be created on channels that are opened on the discovery endpoint. This causes the client to assert, as shown in the attached logs, but either way, no connection is established. The whole sequence is shown in the attached "fail" capture. The "good" capture is from UAExpert, which performs the same steps, but closes the second channel and opens a third, non-discovery one before calling CreateSession. Is Ignition being too picky here? Or are we missing a step in the connection establishment dance?

## Attachments

[files.zip](https://github.com/open62541/open62541/files/13575640/files.zip)

## Checklist

 - [X] Other OPC UA SDKs used (client or server): Ignition 8.1.4 OPC UA Server
 - [X] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [X] Wireshark network dump attached
 - [ ] Critical issue

MikeGranby:
Client does not switch from discovery to full-feature endpoint if security policy does not change

Closes #6165

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -413,8 +413,12 @@ sendHELMessage(UA_Client *client) {
     hello.maxMessageSize = client->config.localConnectionConfig.localMaxMessageSize;
     hello.maxChunkCount = client->config.localConnectionConfig.localMaxChunkCount;
 
-    /* Use the DiscoveryUrl -- if already known via FindServers */
-    if(client->discoveryUrl.length > 0) {
+    /* We may open difference SecureChannels with different EndpointUrls. The
+     * response of the FindServers and GetEndpoints services can depend on the
+     * EndpointUrl used in the HEL message. */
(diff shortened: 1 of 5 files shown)


jpfr:
Rebased on 1.4.
Some assorted improvements on top...

MikeGranby:
I see some fixes of now vs. monotonic. I had some issues with this on Alpine that I had to work around, so I'll pull this and see if that addresses those...

PR #6202, opened 2024-01-03
[REVIEW] Handling of symbolic names during data type generation.

NoelGraf:
The nodeset compiler had trouble resolving the NodeIds of certain data types correctly because the SymbolicName was not processed properly. The reason for this is that the SymbolicName can also be present in the CSV file, which means that the corresponding data type is not found. This in turn affects the correct setting of the NodeId for that data type.

This Pull Request (PR) addresses this issue. A lookup table is created that replaces the possible SymbolicName with the corresponding BrowseName. To make this possible, it is necessary to additionally pass the XML file when generating the datatypes.


jpfr:
Looks good. If I understand correctly, we will have the Nodeset-XML as the description of the datatypes anyway. As the bsd file import is deprecated...

PR #6214, opened 2024-01-05
[REVIEW] Mark the NamespaceMap attribute as deprecated

NoelGraf:
The NAMESPACE_MAP argument is deprecated and no longer has any effect. The index of the nodeset is set automatically.


jpfr:
Nice. I merged this.
Once this has landed one master, please remove any reference to the deprecated argument.
So we give a warning on 1.4 and clean upon master.
So we don’t accumulate cruft…

PR #6220, opened 2024-01-08
feat(server) Extended Asynchronous Operations (Read + Write of Variables)

andreasebner:
This PR introduces an extension of asynchronous operations. Currently, only methods can be executed asynchronously. The PR extends the mechanism and allows the asynchronous execution of write and read requests. Execution is thread-based in the user code. 

With this change, long running DataSource callbacks, for example, can no longer block the server.


examples/tutorial_server_variable_async.c
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * Async Read Write of Variables
+ * -------------------------
+*/
+
+#include <open62541/client_config_default.h>
+#include <open62541/plugin/log_stdout.h>
+#include <open62541/server.h>
+
+#ifndef WIN32
```
JohannesKauffmann, line 17: You should use `UA_ARCHITECtURE_WIN32` or `_WIN32`, but not `WIN32`. The latter works because it is defined by default by CMake, but you shouldn't rely on it.

(Note that existing code does violate this principle as well)

src/server/ua_server_async.c
```diff
@@ -431,11 +532,14 @@ UA_Server_processServiceOperationsAsync(UA_Server *server, UA_Session *session,
 
     /* Finish / dispatch the operations. This may allocate a new AsyncResponse internally */
     uintptr_t respOp = (uintptr_t)*respPos;
-    uintptr_t reqOp = *(uintptr_t*)((uintptr_t)requestOperations + sizeof(size_t));
+    //uintptr_t reqOp = *(uintptr_t*)((uintptr_t)requestOperations + sizeof(size_t));
```
V-Strassheim, line 535: Should the commented-out code be included?

src/server/ua_services_attribute.c
```diff
@@ -1809,6 +1913,91 @@ copyAttributeIntoNode(UA_Server *server, UA_Session *session,
...
+        *opResult =
+            UA_AsyncManager_createAsyncResponse(&server->asyncManager, server,
+                                                &session->sessionId, requestId, requestHandle,
+                                                UA_ASYNCOPERATIONTYPE_WRITE, ar);
+        if(*opResult != UA_STATUSCODE_GOOD)
+        goto cleanup;
+    }
+
+    //Create the Async Request to be taken by workers
+    //TODO make "UA_AsyncManager_createAsyncOp" generic for read/write
```
V-Strassheim, line 1953: I wonder if the todo is still valid or is in the wrong place?
If UA_AsyncManager_createAsyncOp should somehow become more generic, then I would expect the todo comment to be in the function.

examples/tutorial_server_variable_async.c
```diff
@@ -0,0 +1,262 @@
...
+                    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "AsyncMethod_Testing: Call done: OKAY");
+                    UA_CallMethodResult_clear(&response);
+                    break;
+                case UA_ASYNCOPERATIONTYPE_READ:
+                    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "AsyncRead_Testing: Got entry: OKAY");
+                    UA_DataValue readResponse;
+                    //ToDo check timestamp to return logic
+                    //TODO public API uses admin sesseion in the next call level therefore e.g. access level are wrong
+                    readResponse = UA_Server_read(globalServer, &request->readValueId, UA_TIMESTAMPSTORETURN_BOTH);
+                    //UA_Server_readValue(globalServer, request->readValueId.nodeId, &readResponse.value);
```
V-Strassheim, line 208: Can the comment be removed or extended with a note regarding the purpose?

V-Strassheim (commented):
Thank you very much @andreasebner , I find the async functionality quite important for many applications.
I hope it is ok that I have added a few questions as a review to better understand what is still missing or how it should be used.

examples/tutorial_server_variable_async.c
JohannesKauffmann: Correction: since this is an example, `_WIN32` should be used. `UA_ARCHITECTURE_WIN32` is only used for core library code.

MurzynJ:
We tested this PR in our application and it works fine with one exception. With Read service, the application doesn't get the TimestampsToReturn hint, so it doesn't know what to fill inside the DataValue. This is a difference from the sync API.

Another missing point in both sync and async Read API is that the MaxAge hint is not propagated via callbacks to application. But this is probably a topic for another PR.

examples/tutorial_server_variable_async.c
andreasebner: Thank you for the hint. Will be added in the next commit.

src/server/ua_server_async.c
andreasebner: This change is intended. The commented out line will be removed in the next commit. In the past there was no requests parameter and the function was only used for method calls. With this in mind, the pointer to methodsToCallSize within the UA_CallRequest was handed in. Adding the size of size_t pointed then to the *methodsToCall. With the new requests parameter this step is not needed any more.

src/server/ua_services_attribute.c
andreasebner: Right, old artefact. Removed.

andreasebner:
> We tested this PR in our application and it works fine with one exception. With Read service, the application doesn't get the TimestampsToReturn hint, so it doesn't know what to fill inside the DataValue. This is a difference from the sync API.
> 
> Another missing point in both sync and async Read API is that the MaxAge hint is not propagated via callbacks to application. But this is probably a topic for another PR.

You're right, the last version was not able to pass through TimestampsToReturn and MaxAge. Since Asyc-Operations handle only a "single read" or a "single call" etc. the structure of the requests differs a little from the standard service calls. In the latest code-base a passthrough of this information’s is added. The API-design may need some discussion. 

Regarding the MaxAge in general -> You're right. This is a great hint but will not be part of this PR. We will discuss this internally.

andreasebner:
This PR is replaced by #7328

andreasebner:
@basyskom-jvoe Thanks for reporting, the issue is solved by #7424

PR #6221, opened 2024-01-08
fix(server): memory handling in the server mdns implementation.

nogryy:
Fix memcpy issue (maybe "copy&paste" error from fix in 1.3 ->  47239db132de3db30b975d65853a5a2ee295124c).


jpfr:
*ouch*
Thanks for the find!

PR #6227, opened 2024-01-12
Fix heap corruption on Windows when loading a PEM encoded private key

basyskom-jvoe:
If the data buffer created by i2d_PrivateKey() is returned directly, freeing the ByteString causes a heap corruption on Windows.


jpfr:
👍

PR #6229, opened 2024-01-13
Remove duplicate data source definitions

jpfr:


andreasebner:
The code path has been overhauled in recent versions, and the change is no longer applicable.

PR #6233, opened 2024-01-16
fix(plugin): fix hashing of certificate chains with Mbed tls

### LINKED ISSUE #6232: Mbed Thumbprint Calculation Incorrect with Certificate Chains
## Description

I have a server that is providing a certificate chain as the server certificate in the endpoint definition, this comprising the actual server certificate and the root certificate used to generate it. I am not sure whether this is the correct behavior, but I am assuming that it could occur in some cases, e.g., when there is an intermediate certificate involved. The server is calculating the certificate thumbprint based on the leaf certificate, but the Mbed crypto support just calculates a SHA1 hash over the entire data structure, which includes any other certs in the chain. This causes the connection to fail with a thumbprint mismatch. The fix would be to read the DER element length from the first few bytes of the certificate structure and limit the hash to that, but I am unsure as to whether this ought to be necessary. I'm going to code it up simply as an aid to interoperability, and you can decide from there.

## Checklist

 - [X] Other OPC UA SDKs used (client or server): Red Lion Crimson 3.2
 - [ ] Critical issue

MikeGranby:
Include only first certificate of a chain in SHA1 thumbprint hash.

Closes #6232

### DIFF
--- a/plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
+++ b/plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
@@ -159,6 +159,34 @@ mbedtls_sign_sha1(mbedtls_pk_context *localPrivateKey,
     return UA_STATUSCODE_GOOD;
 }
 
+static size_t
+get_der_size(const UA_ByteString *data) {
+
+    size_t p = 0;
+
+    if(data->length > p && data->data[p++] == 0x30) {
+        if(data->length > p) {
+            size_t cert = 0;
+            if(data->data[p] & 0x80) {
+                size_t c = data->data[p++] & 0x7F;
+                if(data->length > p + c) {
+                    for(size_t n = 0; n < c; n++) {
+                        cert = (cert << 8) + data->data[p++];
+                    }
+                }
+            } else {
+                cert = data->data[p++];
+            }
+
+            if(cert && data->length > (cert += p)) {
+                return cert;
+            }
+        }
+    }
+
+    return data->length;
+}
+
 UA_StatusCode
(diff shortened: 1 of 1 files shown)


jpfr:
Thanks for this.

We recently added something very similar:
https://github.com/open62541/open62541/blob/5e76d257fd0ab4bc2558bf5e24ba4d5f0bab0aec/src/ua_util.c#L502

I'd like to see the following:

- Use the existing `getLeafCertificate` internally. So we only pass the leaf certificate into the plugin. There is only a small number of places where we call the `thumbprint` method callback. That solves the problem for OpenSSL at the same time.
- Do the change for the 1.4 branch. From this it gets merged to master.

Keep up the good work!

MikeGranby:
A couple of observations:

* The OpenSSL path uses `x509_digest` which I believe already handles the DER decoding to ensure it gets the right data size, although I don't have an openssl build running here so I'm not 100% sure. We could certainly handle trimming the cert list earlier in the process, but I did it in mbed since that seemed to be where the problem is; and

* The `getLeafCertificate` implementation assumes a two byte length encoding of the certificate, which is probably reasonable in canonical form when considering the likely size of a certificate, but in theory, the length could be an arbitrary number of bytes.

I'll rework this per your suggestions, but I will also improve `getLeafCertificate` to handle arbitrary lengths.

jpfr:
Sounds great!

MikeGranby:
Oh, and can you email me at mikeg@mikeg.net? Want to discuss any plans to add this library to Alpine as a package.

andreasebner:
Thank's for your input! On current master, `mbedtls_x509_crt_parse` parses the entire concatenated DER chain into a linked list and this PR seems outdated.

PR #6240, opened 2024-01-19
fix(examples,tests): correct Windows define

JohannesKauffmann:
Checking for `WIN32` only works by virtue of CMake defining `WIN32`, by itself. The `_WIN32` define actually comes from the compiler, so use that instead.
    
Since these are examples and tests, don't use `UA_ARCHITECTURE_WIN32`.


jpfr:
Nice, thanks

PR #6254, opened 2024-01-28
fix(nodesetinjector): Fix issues with automatic NodeSet loading

MurzynJ:
- Allow both autoload and NodesetLoader at the same time.
That's an obvious fix. These 2 features are independent from each other, so using NodesetLoader should not exclude autoload.

- Make the autoload usable for projects where open62541 is not a root project.
Currently if you include open62541 as a CMake subproject (subdirectory), the autoload won't work. My proposed changes work in combination with the following sample CMakeLists.txt at the parent project level.

image

image


NoelGraf:
@MurzynJ Thanks for the contribution. I will take a look at it.

jpfr:
Nice. Merged.

PR #6255, opened 2024-01-30
[REVIEW] Disable the reuseaddr socket option by default

NoelGraf:
The reuseaddr socket option is now disabled by default.
Server does not start if Binary Protocol Manager fails to start.

### DIFF
--- a/arch/eventloop_posix_tcp.c
+++ b/arch/eventloop_posix_tcp.c
@@ -11,19 +11,21 @@
 #include "eventloop_common.h"
 
 /* Configuration parameters */
-#define TCP_PARAMETERSSIZE 5
+#define TCP_PARAMETERSSIZE 6
 #define TCP_PARAMINDEX_RECVBUF 0
 #define TCP_PARAMINDEX_ADDR 1
 #define TCP_PARAMINDEX_PORT 2
 #define TCP_PARAMINDEX_LISTEN 3
 #define TCP_PARAMINDEX_VALIDATE 4
+#define TCP_PARAMINDEX_REUSE 5
 
 static UA_KeyValueRestriction TCPConfigParameters[TCP_PARAMETERSSIZE] = {
     {{0, UA_STRING_STATIC("recv-bufsize")}, &UA_TYPES[UA_TYPES_UINT32], false, true, false},
     {{0, UA_STRING_STATIC("address")}, &UA_TYPES[UA_TYPES_STRING], false, true, true},
     {{0, UA_STRING_STATIC("port")}, &UA_TYPES[UA_TYPES_UINT16], true, true, false},
     {{0, UA_STRING_STATIC("listen")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false},
-    {{0, UA_STRING_STATIC("validate")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false}
+    {{0, UA_STRING_STATIC("validate")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false},
+    {{0, UA_STRING_STATIC("reuse")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false}
 };
 
 typedef struct {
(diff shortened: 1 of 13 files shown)


src/server/ua_server_binary.c
```diff
@@ -1760,6 +1764,7 @@ UA_BinaryProtocolManager_start(UA_Server *server,
     if(!haveServerSocket) {
         UA_LOG_ERROR(config->logging, UA_LOGCATEGORY_SERVER,
                      "The server has no server socket");
+        return UA_STATUSCODE_BADINTERNALERROR;
     }
 
     /* Update the application description to include the server urls for
```
NoelGraf, line 1770: @jpfr How do we deal with this?
Because we use only the wildcard option by default, no URL is added to the DiscoveryUrl list.

tests/server/check_server.c
```diff
@@ -23,7 +23,9 @@ static UA_Server *server = NULL;
 static void setup(void) {
     server = UA_Server_new();
     ck_assert(server != NULL);
-    UA_ServerConfig_setDefault(UA_Server_getConfig(server));
+    UA_ServerConfig *config = UA_Server_getConfig(server);
```
NoelGraf, line 26: @jpfr Because the REUSEADDR option is disabled by default, some tests fail with an 'address already in use' message.
My suggestion is to activate the REUSEADDR option for all tests. What do you think?

jpfr: I don't see why a test should fail.
Do we spin up parallel servers with the same port?
Then we should change the port for one of these...

src/server/ua_server_binary.c
jpfr: Is this a problem?
I think we mirror back the DiscoveryUrl used to open the connection.
When somebody calls GetEndpoints...

jpfr (commented):
Looks good! A last "beautify" comment inline.

arch/eventloop_posix_tcp.c
```diff
@@ -11,19 +11,21 @@
...
+#define TCP_PARAMINDEX_REUSEADDR 5
 
 static UA_KeyValueRestriction TCPConfigParameters[TCP_PARAMETERSSIZE] = {
     {{0, UA_STRING_STATIC("recv-bufsize")}, &UA_TYPES[UA_TYPES_UINT32], false, true, false},
     {{0, UA_STRING_STATIC("address")}, &UA_TYPES[UA_TYPES_STRING], false, true, true},
     {{0, UA_STRING_STATIC("port")}, &UA_TYPES[UA_TYPES_UINT16], true, true, false},
     {{0, UA_STRING_STATIC("listen")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false},
-    {{0, UA_STRING_STATIC("validate")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false}
+    {{0, UA_STRING_STATIC("validate")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false},
+    {{0, UA_STRING_STATIC("reuseaddr")}, &UA_TYPES[UA_TYPES_BOOLEAN], false, true, false}
```
jpfr, line 28: The UDP connection manager has a pre-rexisting option called `reuse`.
Let's do it identical everywhere. --> reuse.

include/open62541/server.h
```diff
@@ -176,6 +176,7 @@ struct UA_ServerConfig {
                               * (default: 0 -> unbounded) */
     UA_UInt32 tcpMaxChunks;  /* Max number of chunks per message
                               * (default: 0 -> unbounded) */
+    UA_Boolean reuseAddr;
```
jpfr, line 179: Rename to tcpReuseAddr.

src/server/ua_server_binary.c
```diff
@@ -23,6 +23,13 @@
 #include "ua_server_internal.h"
 #include "ua_types_encoding_binary.h"
 #include "ua_services.h"
+#include "mp_printf.h"
+
+#ifdef _WIN32
+# include <winsock2.h>
```
jpfr, line 29: We don't include system libraries from the /src folder. Only the plugins do it.
Move the DiscoveryUrl generation and `UA_Array_appendCopy` into the `serverNetworkCallback`.
There the eventloop will give you the hostnames for listen sockets when they open.

jpfr (commented):
More comments inline.
Keep up the good work!

jpfr:
Merged. Great!

For the future we should
- document the new `reuse` option for TCP
- get rid of the system header include in the core lib

Let's do that next week.

PR #6274, opened 2024-02-14
[REVIEW] Interface definition for gds push management

NoelGraf:


src/server/ua_gds_pushmanagement_ns0.c
```diff
@@ -0,0 +1,161 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Copyright (c) 2023 Fraunhofer IOSB (Author: Noel Graf)
+ */
+
+#include "ua_gds_pushmanagement_ns0.h"
```
jpfr, line 8: Move the information model to a second PR when you have some usable functionality.

Also this should be behind another build flag.
Not just UA_ENABLE_ENCRYPTION.

jpfr (commented):
Nice! Comments inline.

jpfr:
The code is perfect.

One minor request: Use `git-mv` to rename files instead of deleting them and creating a new file. So git knows how to adjust commits that we continue to merge from 1.4 to master.

NoelGraf:
I tried, but Git automatically does a delete/add from the actual rename/modify. I guess because over 50 percent of the file has been changed. Otherwise two commits would probably have to be made.

jpfr:
Nice. Merged.

PR #6275, opened 2024-02-15
fix(plugins): remove dead code.

Emilvks:
The UA_ByteString_clear(&outSegment) call will never happen. bufferAllocated is decleard to false and will only be true if retval equals UA_STATUSCODE_GOOD. Retval don't change between the the two if statements so retval will still be in the same state (UA_STATUSCODE_GOOD) and it will not reach the if(bufferAllocated) check. If retval don't equals UA_STATUSCODE_GOOD the retavl will be return in the previous check and bufferAllocated will never be true.


jpfr:
Good find.
I think we can even simplify more.
The entire if-block can be removed.

And the definition of `UA_StatusCode retval` can be moved inside the for-loop as well.
So retval only exists at the location where the `outval` memory is allocated.

jpfr:
👍

PR #6280, opened 2024-02-15
fix(pubsub): Fixed DSR not coming back from error state after communication resumes

MichalTyndel:
Reader which MessgaeReceiveTimeout elapsed was ignored when parsing incoming messages, so it couldn't get back to 'operational' state even after communication resumed.


MichalTyndel:
Updated pubsub test (wrong timeout). Removed part of it, where the state toggles between error and operational. Now that it actully toggles checking expected state in callback will not work this way anymore.

PR #6281, opened 2024-02-15
feat(pubsub): Support arrays in messages with RawData field encoding

MichalTyndel:
Until now only scalars were handled correctly in messages with layout 'periodic fixed'


jpfr:
Cool, thanks

Issue #6205, opened 2024-01-03
Different data types for namespace indexes used.


sebschwinn:
Hello all,

I stumbled upon the following unexpected implementation regarding the public namemspace functions.

The namespace index in the open62541 is defined as two byte unsigned integers (uint16_t). In file [open62541/include/open62541/server.h beginning at line 1856](https://github.com/open62541/open62541/blob/26d7d15442a7d94b67958035c53baacf93f67eca/include/open62541/server.h#L1856C1-L1856C11) the prototypes of functions `UA_Server_addNamespace`, `UA_Server_getNamespaceByName` and `UA_Server_getNamespaceByIndex` are defined. The function `UA_Server_addNamespace` returns the added namespace index as expected as two byte unsigned integer. The other two functions `UA_Server_getNamespaceByName` and `UA_Server_getNamespaceByIndex` instead use the plattform-dependent type `size_t` as call by reference. I expected the usage of type `uint16_t` instead of `size_t`. This implementation requires unnecessary casting and could be a source of errors due to overflows. Is there a deeper reason I'm missing?

Thanks and greetings,
Sebastian Schwinn

jpfr:
Sounds good.
If you can do so, feel free to send a pull request on Github to make the change.

PR #6289, opened 2024-02-19
Fix for issue open62541/open62541#6205: Different data types for namespace indexes used.

sebschwinn:
Fix for issue open62541/open62541#6205 Different data types for namespace indexes used.
Changed prototypes and functions for namespace methods from `size_t` to `UA_UInt16`. Also updated the affected tests.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -1831,11 +1831,11 @@ UA_Server_addNamespace(UA_Server *server, const char* name);
 /* Get namespace by name from the server. */
 UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Server_getNamespaceByName(UA_Server *server, const UA_String namespaceUri,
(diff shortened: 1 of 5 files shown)


sebschwinn:
Hello all,
Unfortunately I am not able to evaluate the error of the CI-processes of msys64 (and clang) on my own. Can anybody support me?  Or has anybody else seen a fix for this? The command includes the -NonInteractive flag, but the process still wants to read credentials, which is (as of my knowledge) not possible.
Greetings,
Sebastian

jpfr:
Thanks we'll look into it.

mcpat:
Is there a reason why you do not change the prototypes and implementations of the internal functions `getNamespaceByIndex` and `getNamespaceByName`? It would be great if the public and internal prototypes remain identical.

Issue #6291, opened 2024-02-20
Wrong pointer calculation in readSessionDiagnostics


the-om:
## Description
The readSessionDiagnostics function in ua_server_ns0_diagnostics.c calculates the wrong pointer address when accessing array members of the SessionDiagnostics or SessionSecurityDiagnostics nodes.

https://github.com/open62541/open62541/blob/9f0abab2336e0c66779d7373615740915457cb62/src/server/ua_server_ns0_diagnostics.c#L437

This assigns the **address** of the array data pointer to `content`. It should actually copy the array data pointer to content.

## Background Information / Reproduction Steps
Opening a Server Diagnostics View in UA Expert causes a open62541 server to crash on my system (Windows 10, msvcrt143).

Alternatively navigate to Root/Objects/Server/ServerDiagnostics/SessionsDiagnosticsSummary/\<your session\>/SessionSecurityDiagnostics/ClientUserIdHistory

open62541 needs to be compiled with UA_ENABLE_DIAGNOSTICS.

## Checklist
 - [x] open62541 Version (release number or git tag): 1.3, 1.4
 - [x] Other OPC UA SDKs used (client or server): UA Expert client
 - [x] Operating system: tested on Windows 10, but applies to any
 - [x] Logs (with `UA_LOGLEVEL` set as low as necessary) attached: N/A
 - [x] Wireshark network dump attached: N/A
 - [x] Self-contained code example attached: No, because this applies to any server
 - [x] Critical issue: maybe?

PR #6292, opened 2024-02-20
Use correct address when reading array members of session diagnostic types

the-om:
Fix for #6291


jpfr:
Nice. Please change the target branch to 1.3.
We will forward-port from there.

jpfr:
I pulled your commit into #6294.

PR #6293, opened 2024-02-20
Use TRACE as the compiled-in log level and INFO as the configuration default

jpfr:


keba-estr:
Clear logging at the end of the function UA_ServerConfig_clean to avoid usage of invalid pointer in securityPolicies.

jpfr:
@keba-estr thanks!

PR #6298, opened 2024-02-21
fix(server): Problems with event subscription in reduced namespace

nogryy:
Hi,

we are currently trying to integrate version 1.4.0-rc2 (from 1.3.6). If we subscribe to the EventSource in the UAExpert (v1.7.1) now, the Tool returns an error "Item 0 failed with error BadTypeDefinitionInvalid." and cancel the event subscription.
This also happens in the example project (tutorial_server_events). 
We use the REDUCED namespace, if we switch to the FULL namespace the event subscription works but this is not an option for us.

If we adjust the order of the clause validation in the ua_subscription_events_filter.c (like in 1.3.6) the event subscription works again.

### DIFF
--- a/src/server/ua_subscription_events_filter.c
+++ b/src/server/ua_subscription_events_filter.c
@@ -1125,6 +1125,11 @@ UA_SimpleAttributeOperandValidation(UA_Server *server,
     if(UA_NodeId_isNull(&sao->typeDefinitionId))
         return UA_STATUSCODE_BADTYPEDEFINITIONINVALID;
 
+    /* If the BrowsePath is empty, the Node is the instance of the
+     * TypeDefinition. (Part 4, 7.4.4.5) */
+    if(sao->browsePathSize == 0)
+        return UA_STATUSCODE_GOOD;
+
     /* EventType is a subtype of BaseEventType? */
     UA_NodeId baseEventTypeId = UA_NODEID_NUMERIC(0, UA_NS0ID_BASEEVENTTYPE);
     if(!isNodeInTree_singleRef(server, &sao->typeDefinitionId,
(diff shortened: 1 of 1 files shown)


jpfr:
With your change, if the BrowsePath is empty, then the TypeDefinitionId is not checked.
That doesn't sound right.
Is it possible that your EventFilter uses an EventType that does not exist in the reduced information model?
Then we can try to add just that EventType...

nogryy:
Hi @jpfr  ,

thanks for the fast reply.

I guess we have exactly the same problem, described in this comment (1.3.6):

https://github.com/open62541/open62541/blob/fe2069550076304e03a2d99480a9a6017d337775/src/server/ua_subscription_events_filter.c#L1067-L1071

In 1.4.0 the TypeDefinitionId check is before the BrowsePath check, without the "Workaround" (see comment above):

https://github.com/open62541/open62541/blob/e1bd2f5478ba2f75e769f000b0ead73d0679d4f5/src/server/ua_subscription_events_filter.c#L1129-L1141

In 1.3.6 the TypeDefinitionId check is after the BrowsePath check: 

https://github.com/open62541/open62541/blob/fe2069550076304e03a2d99480a9a6017d337775/src/server/ua_subscription_events_filter.c#L1076-L1087

So, our solution was to return to the check order like in 1.3.6.

Perhaps we don't fully understand this function and also the OPC Spec is not easy to understand.

At the moment, our workaround (without a open62541 fix) is to add the missing NodeId (mentioned in the comment above) when the opc server starts up. 

But I don't know if this NodeID fits into an Eventminimal nodeset...

Unfortunately, we cannot see what the best solution is here.

nogryy:
fixed in 1.4.2 with this commit:

https://github.com/open62541/open62541/commit/3888af3410b53be0c5fa15b9d762dbab5b84902e

Thx

PR #6318, opened 2024-02-29
fix(plugin): return status code from mbedtls_hmac.

Emilvks:
mbedtls_md_hmac_starts, mbedtls_md_hmac_update and mbedtls_md_hmac_finish is function calls in mbedtls_hmac which returns (int) status code.
In current implementation the status is ignored which could cause different problems.
In this PR mbedtls_hmac returns status codes instead of void, UA_STATUSCODE_GOOD (0x0) on success and
UA_STATUSCODE_BADSECURITYCHECKSFAILED (0x8013) if something went wrong. This way we can return faster when a error occur. This update also help remove static code analysis errors like "Unchecked return value"


jpfr:
Nice. Looks good.
But please remove the nested and if-statements.
We have different conventions throughout.
Flat control flow and early-return is preferred.
You can OR the status codes and check at the end.

Emilvks:
@jpfr

jpfr:
Nice! Merged.

PR #6332, opened 2024-03-05
[REVIEW] Fix potential memleak in sks examples

NoelGraf:


jpfr:
We have an "official" way to set/get information attached to a session: `UA_Server_getSessionAttribute`
The clientUserId used for the last activation is always available to get.
Please check if you can use that instead of the void-pointer session context.

NoelGraf:
The session context must be set for the getUserExecutableOnObject_sks function callback.

NoelGraf:
Changes made.
However, the `UA_Server_getSessionAttribute` function cannot be used directly because the session KeyValueMap is NULL.

jpfr:
You don't have to set the attribute. `0:clientUserId ` is always available.
https://github.com/open62541/open62541/blob/dc91b22a34cdbe89311e09a036af44412c815253/include/open62541/server.h#L569C6-L569C21

jpfr:
👍

PR #6336, opened 2024-03-07
[Review] style: fix shadow var found from "gcc -Wshadow"

laroche:
Compiling with "gcc -Wshadow" shows a few shadow vars, this patch is fixing them.

I am very new to open62541, so please forgive if something is not fully complient with
how patch requests should be sent in.

open62541 compile flags are already very strict, but it seems that "-Wshadow" is still missing
and might be a good addition to the default flags. Also I have just done a standard compile,
but not looked at tests or all possible compile targets.

Comments?

best regards and thanks for open62541,

Florian La Roche


jpfr:
Nice, thanks!

PR #6345, opened 2024-03-13
Make ua2json compile under Windows

RolfKal:
This allows to compile the ua_tools under Windows. CMAKE option UA_BUILD_TOOLS


jpfr:
Thanks for this!

We have a failing CI test.
You should replace #if with #ifdef.

Also have a look at our conventions for the commit title.

Keep up the good work!

Issue #6352, opened 2024-03-18
PubSub readergroup state is PREOPERATIONAL after re-enabling readergroup for UA_PUBSUB_RT_FIXED_SIZE


keba-estr:
## Description
If a PubSub readergroup is configured with UA_PUBSUB_RT_FIXED_SIZE, the readergroup state is only set from PREOPERATIONAL to OPERATIONAL when receiving the first message.
If the readergroup is disabled with UA_Server_setReaderGroupDisabled and re-enabled with UA_Server_setReaderGroupOperational, the state remains in PREOPERATIONAL even after receiving a message, because the function prepareOffsetBuffer, which does the state change in the first step, is not called any more.

What is the correct solution?
Should we clear the prepared offset buffer when disabling the ReaderGroup and reallocate again after re-enabling, or should we check the state PREOPERATIONAL also in UA_NetworkMessage_updateBufferedNwMessage?

## Checklist

Please provide the following information:

 - [X] open62541 Version (release number or git tag): 1.4

keba-estr:
Resolved with PR #6354

keba-estr:
Resolved with PR #6354

PR #6354, opened 2024-03-19
fix(pubsub): Set readergroup state to operational after restart

keba-estr:
Set the state of a readergroup with UA_PUBSUB_RT_FIXED_SIZE to UA_PUBSUBSTATE_OPERATIONAL also after several disable and endable calls. The buffered network message is not reset when disabling a readergroup.


keba-estr:
Resolved issue #6352

jpfr:
Nice, thanks.

keba-estr:
Problem already resolved with other commits on master branch.

PR #6355, opened 2024-03-19
Nodeset compiler with guid nodeids

paddor:
Refers open62541/open62541-nodeset-loader#253

This adds support for GUID Node IDs in nodeset compiler.

### DIFF
--- a/tools/nodeset_compiler/backend_open62541_datatypes.py
+++ b/tools/nodeset_compiler/backend_open62541_datatypes.py
@@ -99,6 +99,8 @@ def generateNodeIdCode(value):
     elif value.s != None:
         v = makeCLiteral(value.s)
         return u"UA_NODEID_STRING(ns[%s], \"%s\")" % (value.ns, v)
+    elif value.g != None:
+        return u"UA_NODEID_GUID(ns[%s], %s)" % (value.ns, generateGuidCode(value.gAsStrings()))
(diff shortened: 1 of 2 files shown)


paddor:
This is not quite correct yet. Leading zeroes from the int->hex_string conversion are missing.

jpfr:
Cool! Please remove the [WIP] when you are happy.

tools/nodeset_compiler/backend_open62541_datatypes.py
```diff
@@ -99,6 +99,8 @@ def generateNodeIdCode(value):
     elif value.s != None:
         v = makeCLiteral(value.s)
         return u"UA_NODEID_STRING(ns[%s], \"%s\")" % (value.ns, v)
+    elif value.g != None:
+        return u"UA_NODEID_GUID(ns[%s], %s)" % (value.ns, generateGuidCode(value.gAsStrings()))
     raise Exception(str(value) + " no NodeID generation for bytestring and guid..")
```
basyskom-meerkoetter, line 104: the error string in the exception still needs to be adjusted?

paddor: You're right. But that's not all. See above. My python skills are bad.

paddor:
@jpfr This is ready for review.

PR #6361, opened 2024-03-25
fix(pubsub): Set deleteFlag of connection after clearing lists

keba-estr:
Inhibit double free of data set reader.

fix(pubsub): Only remove dataset reader from list if linked reader group found

Adapt implementation to PubSub dataset writer.


keba-estr:
Partially resolved with 6362

PR #6363, opened 2024-03-26
Mark applicable high-level client functions thread-safe

sgoll:
## Description

It seems that some higher-level client functions are actually thread-safe but are not marked as such. Would it be sensible to add this?

For the functions whose signatures are adjusted by this PR, I verified that they either use thread-safe functions themselves, or that their code explicitly guards access to `client` with the appropriate mutex calls.


jpfr:
Looks good!
However there is a clash with recent changes on master.

Sorry for the additional work, but the other PR we just merged existed for some weeks already.

jpfr:
Nice, merged. 👍

PR #6365, opened 2024-03-26
fix (tools): Make ua2json compile under Windows

RolfKal:
With this addition the UA_BUILD_TOOLS option in CMake can be enabled


jpfr:
Nice, thanks

PR #6370, opened 2024-03-27
All architecture-specific files isolated in single folder

RavilN:
Now custom architecture-specific source files can be located in one isolated folder. For default architectures (win32 and posix), it is the folder ``arch/eventloop_posix``.  For custom architectures, this folder could be under the arch folder for public code, or outside of the open62541 source code tree for private code. The location of the folder with architecture-specific files can be set using the new build option ``UA_ARCH_FOLDER``. Default architecture-specific files under the folder  ``arch/eventloop_posix`` are renamed, they start with "eventloop_arch_ ". This way no need to modify the list of files in the file CMakeLists.txt for custom architecture build.


jpfr:
It is a good idea to make the architecture porting easier / cleaner.

Two remarks.

The POSIX EventLoop shouldn't be renamed to generic "arch". It heavily relies on the POSIX socket API.
Everything that deviates from that (like the lwip raw API) should do a fork and not share code with the POSIX implementation.
What can be potentially shared is already in the eventloop_common folder.

We can put additional CMakeLists.txt files in the /arch/eventloop_xxx folders.
That way, from the main CMakeLists.txt we would just include the correct folder.
And all the additional complexity is encapsulated there.

RavilN:
The reason to rename ``arch\eventloop_posix\eventloop_posix.h`` to ``arch\eventloop_posix\eventloop_arch.h`` was because ``plugins\crypto\mbedtls\ua_mbedtls_create_certificate.c`` was including it with path ``arch\eventloop_posix\eventloop_posix.h``, which is architecture-specific. What I am not sure about is: whether this file must be included for all architectures. I assumed that each architecture needs to implement its version of this file, so for the ``freertos`` it would be like ``arch\eventloop_freertos\eventloop_freertos.h``.  In this case we would need to change file  ``plugins\crypto\mbedtls\ua_mbedtls_create_certificate.c`` when new architecture is added. To avoid such changes, one of the ways would be to have the file name the same for all architectures, and include it only by file name, which is architecture-neutral, like ``eventloop_arch.h``, where ``arch`` indicates it has architecture-specific code. To make the compiler find it, path to the architecture-specific files is added on include directories using the new option UA_ARCH_FOLDER.

jpfr:
The file `plugins\crypto\mbedtls\ua_mbedtls_create_certificate.c` should not require the POSIX eventloop includes.
It generates a certificate as a bytestring...

That being said, my two remarks from above still hold.
Please update the PR accordingly.

PR #6374, opened 2024-03-28
fix(pubsub): Mark some functions as exported for shared library compilation

RolfKal:
Mark some functions that are used in the ua2json tool as exported to allow linking this tool when open62541 is compiled as shared library on Windows


jpfr:
Exported functions should go into a header in the /include folder.
Please make a suggestion.

RolfKal:
Hi Julius

While I understand the logic behind that, I'm not yet familiar enough with
the library hierarchy to make a good suggestion.
Considering the location of the relevant source file, the closest existing
header that I could think of would probably be server_pubsub.h and then
adding a section for NetworkMessage to that. But is that API really
intended to be exported/public? Or is the ua2json just a weird use case?
What about the other functions in ua_pubsub_networkMessage.h? Any other
candidates that should be exported too?

Regards, Rolf

PR #6399, opened 2024-04-09
fix(pubsub): Fix compile warning and move NetworkMessage functions used in ua2json tool to an export header 

RolfKal:


jpfr:
Nice!
I put three commits on top.
The de/encoding is not required for the server-pubsub API.
On top, I made the API more similar to the de/encoding in `types.h`.

jpfr:
Good point. Added the new header in the CMakeLists.txt.

PR #6410, opened 2024-04-12
fix(pubsub): Move variable declaration into its own conditional define

RolfKal:
The retval variable is used in two different conditional compile parts of the function and declared in the first.
If UA_ENABLE_PUBSUB_INFORMATIONMODEL is not defined but UA_ENABLE_PUBSUB_MONITORING is defined, this results in a compilation error.


jpfr:
👍

PR #6416, opened 2024-04-16
fix(pubsub): export two functions used in examples\server_pubsub_file_configuration.exe

RolfKal:
This allows to compile this example under Windows with an open62541 shared library

### DIFF
--- a/include/open62541/pubsub.h
+++ b/include/open62541/pubsub.h
@@ -17,6 +17,19 @@ _UA_BEGIN_DECLS
 
 #ifdef UA_ENABLE_PUBSUB
 
+#ifdef UA_ENABLE_PUBSUB_FILE_CONFIG
+/* Decodes the information from the ByteString. If the decoded content is a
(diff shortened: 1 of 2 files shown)


jpfr:
Good initiative.

Also here we need to clean up the API before we can export it.
So we don't have to break the API in the future.

The API should go into `pubsub_server.h`.

My proposal for the API is this:

```c
UA_EXPORT UA_StatusCode
UA_Server_importPubSubConfiguration(UA_Server *server,
    const UA_PubSubConfigurationDataType *config);

UA_EXPORT UA_StatusCode
UA_Server_exportPubSubConfiguration(UA_Server *server,
    UA_PubSubConfigurationDataType *config);
```

Reading/writing the `UA_PubSubConfigurationDataType` from a UA_ByteString is a second topic.
I would split between the two steps. That way it is also possible to do a JSON config in the future.

RolfKal:
thanks a lot for your comments. I'll take a look at this. The whole open62541 library is still a bit intimidating in size and I definitely lack the overview at this point to judge about what is a clean API and what is more an internal hack.

PR #6419, opened 2024-04-17
fix(pubsub): Allow interface name for multicast

keba-estr:
In case of multicast configuration only an IP address was allowed
for the network interface. Now also network interface names can be
used.

### DIFF
--- a/arch/eventloop_posix_udp.c
+++ b/arch/eventloop_posix_udp.c
@@ -12,8 +12,8 @@
 #define IPV4_PREFIX_MASK 0xF0
 #define IPV4_MULTICAST_PREFIX 0xE0
 #if UA_IPV6
-#   define IPV6_PREFIX_MASK 0xFF
-#   define IPV6_MULTICAST_PREFIX 0xFF
+# define IPV6_PREFIX_MASK 0xFF
+# define IPV6_MULTICAST_PREFIX 0xFF
 #endif
 
 /* Configuration parameters */
@@ -64,6 +64,17 @@ typedef enum {
     MULTICASTTYPE_IPV6
 } MultiCastType;
 
+typedef union {
+#ifdef _WIN32
+    struct ip_mreq ipv4;
+#else
+    struct ip_mreqn ipv4;
+#endif
+#if UA_IPV6
+    struct ipv6_mreq ipv6;
+#endif
+} MulticastRequest;
+
 static UA_Boolean
 isMulticastAddress(const UA_Byte *address, UA_Byte mask, UA_Byte prefix) {
     return (address[0] & mask) == prefix;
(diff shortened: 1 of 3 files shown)


jpfr:
Very nice.
Two remarks:

We use our own printf internally.
That will also remove the Windows build error.
https://github.com/open62541/open62541/blob/master/deps/mp_printf.h

Please document this new capability in /include/open62541/plugins/eventloop.h

Again, nice!

keba-estr:
@jpfr: Unfortunately I do not have the possibility to test it on a windows system. Is there any other possibility to execute a test on windows?

jpfr:
We can test Windows on our end.

Please move the header documentation to the UDP section, not TCP.

Is it true that now both IP address and network interface name work?
Right now that is somewhat hidden, that we do the work in the if-condition and if that fails try a different way.
Please make that really explicit in the code for future developers (and architecture porting efforts).
E.g. with a small comment or an explicit if/else structure.

keba-estr:
> We can test Windows on our end.
> 
> Please move the header documentation to the UDP section, not TCP.
> 
> Is it true that now both IP address and network interface name work? Please make that really explicit in the code for future developers (and architecture porting efforts).

The documentation is already in the section "UDP Connection Manager", the display of the function in the comparison window is misleading.

Is the commentary in eventloop_posix_udp.c line 131ff enough or is there more needed? I extended it a bit compared to the original commit.

jpfr:
Cool! You are right, I was relying on the Github diff for the TCP vs. UDP.

I am "nearly happy". Right now there is still a mismatch between IPv4 and IPv6.
IPv6 only takes the interface name, not the IP address.

However getting the interface name from a local IP address is ugly.
The best way seems to iterate over the interface descriptions.

Overall I think we can reduce the complexity by always doing that.
Iterating over the interfaces and stopping when either the interface name or the IP address match.

`getifaddrs` for Linux
https://man7.org/linux/man-pages/man3/getifaddrs.3.html

`GetAdaptersAddresses` for Windows
https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses

keba-estr:
Maybe it was an error that an IP address was evaluated for ipv4? What does the specification say about the network interface?

jpfr:
> @jpfr: Which IP address of the GetAdaptersAddresses should be compared to the network interface?

No idea. If the documentation is unclear we could simply test if any of them fits...

keba-estr:
> > @jpfr: Which IP address of the GetAdaptersAddresses should be compared to the network interface?
> 
> No idea. If the documentation is unclear we could simply test if any of them fits...

I would start by comparing the unicast addresses. For me it makes no sense to restrict sending a multicast request to a network interface with anycast address or with multicast address itself.

jpfr:
Nice. I simplified the code a bit further and rolled it back into the original commit. Merged.

PR #6430, opened 2024-04-21
refactor(util): Separate the construction of BuildInfo and ApplicationDescription

RolfKal:
Separate this functionality into their own exported function to allow querying the build version from a client. Use these functions in the server to construct the default server configuration structure.


jpfr:
Looks good!
However, the new methods are no longer plugins.
So the implementation should live in the /src folder.

RolfKal:
Ahh, a very valid point. Would util/ua_util.c be a desirable location to put this?

jpfr:
Yes, Sounds good.

PR #6433, opened 2024-04-22
refactor(util): Separate the construction of BuildInfo and ApplicationDescription

RolfKal:
Separate this functionality into their own exported functions to allow querying the build version from a client. Use these functions in the server to construct the default server configuration structure.


jpfr:
Instead of methods I would constant global variables.
Because all the methods do is return a "static" value.

```c
extern const UA_BuildInfo UA_BuildInfo_default;
```

We have UA_STRING_STATIC, etc. to help put together compile-time instance definitions of the built-in datatypes.

jpfr:
There was an update a few days ago.
Please consider the review comment.

RolfKal:
I was going to look into it but needed to verify a few more things. One issue with using exported static variables instead of functions is that there is an additional complexity when trying to dynamically import exported data symbols. Also I wasn't sure if static allocated strings in data variables are properly ignored when a structure containing them gets passed to the standard memory_clear functions.

PR #6434, opened 2024-04-23
refactor(core): Import Python3 in public macros and use Python3_EXECUTABLE everywhere

ccvca:
The public macros use python, without included python (find_package is added to `tools/cmake/macros_public.cmake`).
The used CMAKE-Variable `Python_EXECUTABLE` does normally point to a python2 installation. (https://cmake.org/cmake/help/latest/module/FindPython.html#result-variables )
To not interfere with a possible python2 installation, all variables are updated to the python3 executable.


jpfr:
Nice 👍

PR #6436, opened 2024-04-23
fix(server): Fix string output with %*s

keba-estr:
%.*s must be used if a length of an unlimited buffer is provided and only the first 'n' bytes should be printed.


jpfr:
👍

Issue #5701, opened 2023-03-20
Single Callback for whole subscription possible? / processDataChangeNotification is slow for subscriptions with many monitored items


JulZimmermann:
Hi,

this is a follow-up on #5053 and #4318.

The current implementation of `processDataChangeNotification` in `ua_client_subscriptions.c` is not really performant when subscribing to multiple (e.g. 10000) monitored items.

#4318 already asks about the possibility to get notified via a single callback call for a subscription, instead of callbacks for every monitored item. Unfortunately, this question was never answered.

#5053 and #5068 changed the implementation of `processDataChangeNotification` to be more performant by using a hash map. However, these changes are not in the latest release and only exist in the main branch.

I have a couple of questions now:

1. Is there some (hidden) way I didn't see, to get notified only once per subscription instead of for every monitored item? This would allow bypassing `processDataChangeNotification` altogether. In my (and probably in many other) use cases, I use a hashmap for managing the monitored items anyway, already at application level. The logic in `processDataChangeNotification` (even with a more performant hash map implementation) is therefore not needed.
2. If this is currently not possible: Would it be possible to add this functionality?
3. Could #5068 be backported to the 1.3 branch?
4. Is it "save" to use the master branch?

Thanks!

JulZimmermann:
Any news on this?

Issue #5226, opened 2022-06-27
Initial data change callback is sometimes not called


basyskom-jvoe:
## Description
Sometimes when creating a monitored item using the synchronous open62541 client API, the initial data change notification is dropped with the following message:
```
Could not process a notification with clienthandle 9 on subscription 4
```
The error message is printed in https://github.com/open62541/open62541/blob/1.3/src/client/ua_client_subscriptions.c#L917 and indicates that the client's monitored item context has not yet been inserted into the list when the notification arrives.

The issue occurs with open62541 1.2 and 1.3, the server is also based on open62541 (same version).

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): v1.3.1
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system: Ubuntu 18.04
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

JulZimmermann:
I have the same problem when creating monitored items via the async api.
I create monitored items dynamically while the application is running. `UA_Client_run_iterate` runs in a loop in an own thread.

The calls to `UA_Client_Subscriptions_create_async`, `UA_Client_MonitoredItems_createDataChanges_async`, and `UA_Client_run_iterate` are protected by a mutex.

I analyzed the network communication and it seems like the flow is the fallowing:
1. CreateMonitoredItemsRequest is send by the client
2. PublishRequest is send by the client
3. PublishResponse is send by the server, already containing the initial values of the requested monitored items
4. CreateMonitoredItemsResponse is send by the server

I'm not sure if the server is allowed to send the values inside the PublishResponse before sending the CreateMonitoredItemsResponse, or if the the OPCUA standard is violated by that. Anyway, the open62541 client lib can't handle this and will discard the values.

basyskom-jvoe:
I have instrumented some places in the code with marker output and according to my analysis, the `CreateMonitoredItemsResponse` arrives before the `PublishResponse` with the initial value (as expected).

```
qt.opcua.plugins.open62541.sdk.client: Decode a message of type CreateMonitoredItemsResponse
qt.opcua.plugins.open62541.sdk.client: Decode a message of type PublishResponse
qt.opcua.plugins.open62541.sdk.client: Got a notification with clienthandle 2 on subscription 1
[...]

Issue #4318, opened 2021-04-08
Single callback for all data change notifications.


andigor:
## Background Information / Reproduction Steps

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

Please provide the following information:

 - [master] open62541 Version (release number or git tag):
 - [none] Other OPC UA SDKs used (client or server): 
 - [windows] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

Hello,

Thanks for reading this.

Background.
I am building an OPCUA gateway to another protocol. I am using ``UA_Client_MonitoredItems_createDataChange``  to subscribe and monitor data changes. As soon as notification arrives I am bypassing data further through the gateway.

The problem.
Data change notifications are delivered one by one. That means I also need to send the data one by one through my gateway which is quite a heavy operation in my case. It would be much simpler if I could receive whole bunch of data in one call.

I do see that ``dataChangeCallback`` is called in the loop from ``processDataChangeNotification`` and delivers every changed data item one by one. I do see that ``processDataChangeNotification`` receives all the data at once via ``UA_DataChangeNotification``. This is exactly what I need. However I don't see a way how I can access ``UA_DataChangeNotification`` from my code.

Question.
Basing on the description above - is there a way to receive all active data change notifications in a single callback call? (same as ``processDataChangeNotification``) ?

Regards,
Andrii

andigor:
it could also help if there was a way to determine the amount of changed data items. I could the organize a counter in the callback and act as soon as the callback is called last time.

andigor:
Another thing is that the call itself looks like an exponential complexity:
```
static void
processDataChangeNotification(UA_Client *client, UA_Client_Subscription *sub,
                              UA_DataChangeNotification *dataChangeNotification) {
    for(size_t j = 0; j < dataChangeNotification->monitoredItemsSize; ++j) {
        UA_MonitoredItemNotification *min = &dataChangeNotification->monitoredItems[j];

        /* Find the MonitoredItem */
        UA_Client_MonitoredItem *mon;
        LIST_FOREACH(mon, &sub->monitoredItems, listEntry) {
            if(mon->clientHandle == min->clientHandle)
                break;
        
}
```

[...]

PR #6443, opened 2024-04-26
Allow a single complete callback for all monitored items in one subscription

JulZimmermann:
Solution for #4318, #5701 and #5226

### DIFF
--- a/include/open62541/client_subscriptions.h
+++ b/include/open62541/client_subscriptions.h
@@ -33,6 +33,10 @@ _UA_BEGIN_DECLS
  */
 
 /* Callbacks defined for Subscriptions */
+typedef void (*UA_Client_DataChangeCallback)
+    (UA_Client *client, UA_UInt32 subId, void *subContext,
+     UA_DataChangeNotification *dataChangeNotification);
+
 typedef void (*UA_Client_DeleteSubscriptionCallback)
     (UA_Client *client, UA_UInt32 subId, void *subContext);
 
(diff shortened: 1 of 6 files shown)


include/open62541/client_subscriptions.h
```diff
@@ -78,6 +82,24 @@ UA_Client_Subscriptions_create_async(UA_Client *client,
     UA_ClientAsyncServiceCallback callback,
     void *userdata, UA_UInt32 *requestId);
 
+UA_CreateSubscriptionResponse UA_EXPORT UA_THREADSAFE
+UA_Client_Subscriptions_create_complete_data_change(UA_Client *client,
```
jpfr, line 86: We don't use snake_case_notation in the library.
Use camlCaseNotation instead.

include/open62541/client_subscriptions.h
```diff
@@ -78,6 +82,24 @@ UA_Client_Subscriptions_create_async(UA_Client *client,
     UA_ClientAsyncServiceCallback callback,
     void *userdata, UA_UInt32 *requestId);
 
+UA_CreateSubscriptionResponse UA_EXPORT UA_THREADSAFE
```
jpfr, line 85: This needs a lot more inline documentation in the header.
Please add it. What does the method do, what are intended use cases, why is it not possible to achive the same with the other API methods.

jpfr (commented):
Cool! See the inline comments.

include/open62541/client_subscriptions.h
JulZimmermann: I renamed the functions to `UA_Client_Subscriptions_createCompleteDataChange` and `UA_Client_Subscriptions_createCompleteDataChange_async` is this OK?

include/open62541/client_subscriptions.h
JulZimmermann: I added some documentation

jpfr (commented):
Ok, cool.
Understood what you want to do.

Please rename this to UA_Client_Subscriptions_createEx.

We use this naming also in other places to denote extended versions of an API.

Furthermore it should still be possible to have detailed callbacks with each MonitoredItem. And at the same time have a global callback for the subscription.

hdabis1234:
What is the status of this?

electro-logic:
This feature is a blocker for my project. I need a single callback for performance reasons and compatibility with the UA OPC Stack that provide a FastDataChangeCallback (https://github.com/OPCFoundation/UA-.NETStandard/blob/711160e84a96606931a6737a70518e1f7b6e666d/Libraries/Opc.Ua.Client/Subscription/Subscription.cs#L454)
Can this pull request be merged into the main branch?

electro-logic:
About this PR:

The clientHandle of the UA_MonitoredItemNotification is an UInt32 and is less flexible than "void* monContext" available in the UA_Client_DataChangeNotificationCallback with the signature (UA_Client* client, UA_UInt32 subId, void* subContext, UA_UInt32 monId, void* monContext, UA_DataValue* value).

For example I was using monContext to assign a structure with the NodeId, DisplayName and SamplingInterval.
Why the void* was changed to an UInt32?

PR #6451, opened 2024-04-30
[WIP] Implement memorystore backend for certificate groups

NoelGraf:
This PR adds the memorystore backend for the certificate groups.

- [x] mbedTLS backend
- [x] openSSL backend

### DIFF
--- a/include/open62541/plugin/certificategroup.h
+++ b/include/open62541/plugin/certificategroup.h
@@ -10,7 +10,6 @@
 #define UA_PLUGIN_CERTIFICATEGROUP_H
 
 #include <open62541/types.h>
-#include <open62541/types_generated.h>
 #include <open62541/plugin/log.h>
 
 _UA_BEGIN_DECLS
--- a/plugins/crypto/mbedtls/ua_certificategroup_memorystore_mbedtls.c
+++ b/plugins/crypto/mbedtls/ua_certificategroup_memorystore_mbedtls.c
@@ -1,6 +1,9 @@
 /* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
  * See http://creativecommons.org/publicdomain/zero/1.0/ for more information.
  *
+ *    Copyright 2018 (c) Mark Giraud, Fraunhofer IOSB
+ *    Copyright 2019 (c) Kalycito Infotech Private Limited
+ *    Copyright 2019 (c) Julius Pfrommer, Fraunhofer IOSB
  *    Copyright 2024 (c) Fraunhofer IOSB (Author: Noel Graf)
  */
 
@@ -9,6 +12,11 @@
 #include <open62541/plugin/log_stdout.h>
 
 #ifdef UA_ENABLE_ENCRYPTION_MBEDTLS
+
+#include <mbedtls/x509.h>
+#include <mbedtls/x509_crt.h>
+#include <mbedtls/entropy.h>
+#include <mbedtls/version.h>
 
 #include "securitypolicy_mbedtls_common.h"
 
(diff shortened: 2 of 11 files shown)


include/open62541/plugin/certificategroup.h
```diff
@@ -36,7 +36,8 @@ struct UA_CertificateGroup {
 
     UA_StatusCode (*getRejectedList)(UA_CertificateGroup *certGroup, UA_ByteString **rejectedList, size_t *rejectedListSize);
 
-    UA_StatusCode (*verifyCertificate)(UA_CertificateGroup *certGroup, const UA_ByteString *certificate);
+    UA_StatusCode (*verifyCertificate)(UA_CertificateGroup *certGroup, const UA_ByteString *certificate,
+                                       const UA_ByteString *issuerCertificates, size_t issuerCertificatesSize);
```
jpfr, line 40: The whole idea of a certificate group is that it provides context for crypto operations.
The issuer list is part of that context.
So the issuer list does not need to be provided again. Remove.

plugins/crypto/certstores/ua_certificategroup_memorystore.h
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef UA_CERTIFICATEGROUP_MEMORYSTORE_H
+#define UA_CERTIFICATEGROUP_MEMORYSTORE_H
+
+#include <open62541/plugin/certificategroup.h>
+
+struct MemoryCertStore;
+typedef struct MemoryCertStore MemoryCertStore;
+
+struct MemoryCertStore {
```
jpfr, line 16: I think most of the structure can be replaced with a UA_TrustListDataType.
Implement UA_TrustListDataType_add, UA_TrustListDataType_remove.
Both taking two trust lists as the input.
This can go into /open62541/util.h as it is generally useful.

In the end we will have a trust list and a rejected list. That is all.
This is so little code that we should fold it into the security group and not keep it separate.

plugins/ua_config_json.c
```diff
@@ -760,9 +760,7 @@ PARSE_JSON(SecurityPolciesField) {
 PARSE_JSON(SecurityPkiField) {
 #ifdef UA_ENABLE_ENCRYPTION
     UA_CertificateGroup *field = (UA_CertificateGroup*)configField;
-    UA_String trustListFolder = {.length = 0, .data = NULL};
-    UA_String issuerListFolder = {.length = 0, .data = NULL};
-    UA_String revocationListFolder = {.length = 0, .data = NULL};
+    UA_String certFolder = {.length = 0, .data = NULL};
```
jpfr, line 763: Both in memory and file based should be possible.

jpfr (commented):
This contains a lot of cool stuff.

a few requests inline.
In general:

Implement the new security groups in new files. Not in the old implementation. We will remove the old implementation once we are happy in the new world. That limits the breakage for people that have their custom config code.

Move the file based storage to a different pr. Let’s do in-memory only first.

file-based should be a separate security group. Not the same code for both.

this will unlock many simplifications for the in-memory code.

keep up the good job!

plugins/crypto/mbedtls/ua_certificategroup_memorystore_mbedtls.c
```diff
@@ -0,0 +1,494 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information.
+ *
+ *    Copyright 2024 (c) Fraunhofer IOSB (Author: Noel Graf)
```
jpfr, line 4: Also add the people who worked on the original implementation.
A lot of the code was reused. The same for the OpenSSL implementation.

jpfr (commented):
Looks great. Just a small comment for the copyright header.

For the file-based certificate group:
It might be easiest to "embed" the memory-based certificategroup inside the file-based one.
The file-based certificategroup could even be completely general and work with both the mbedTLS and the OpenSSL inner onion layer...

PR #6452, opened 2024-05-01
fix(pubsub): export two more functions

RolfKal:
This will allow compilation of examples/pubsub/server_pubsub_file_configuration.c when compiling open62541 library as shared library under Windows


jpfr:
I would split this as follows:

- Add a method for parsing the PubSubConfiguration from the binary file (as a ByteString)
- Add a method for loading the (already parsed) PubSubConfiguration(2).

That way, we are flexible to add JSON as a configuration format in the future.

As all the parts are already implemented, it is just a question of separating the concerns.

PR #6465, opened 2024-05-09
Fix dead link to certified SDK page

paskozdilar:
The documentation page https://www.open62541.org/doc/master/ contains a dead link in the introduction:

> [...] See https://open62541.org/certified-sdk for more details.

Opening the link https://open62541.org/certified-sdk returns 404 Page not found.

I assume the proper link is https://open62541.org/certification which returns a page with the title "Certified SDK", which matches the dead link path name.


jpfr:
👍

PR #6472, opened 2024-05-13
Fix stdlib includes

keba-estr:
Include stdlib.h where open62541 no longer does it

### DIFF
--- a/tests/check_types_builtin_xml.c
+++ b/tests/check_types_builtin_xml.c
@@ -11,6 +11,7 @@
 
 #include <check.h>
 #include <math.h>
+#include <stdlib.h>
 
 #if defined(_MSC_VER)
 # pragma warning(disable: 4146)
--- a/tests/encryption/check_cert_generation.c
+++ b/tests/encryption/check_cert_generation.c
@@ -10,6 +10,7 @@
 #include <open62541/plugin/create_certificate.h>
 
 #include <check.h>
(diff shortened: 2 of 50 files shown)


jpfr:
We include it here: https://github.com/open62541/open62541/blob/84ebf2509a0465a0f20025ff03ece117b67fd885/include/open62541/config.h.in#L278

keba-estr:
> We include it here:
> 
> https://github.com/open62541/open62541/blob/84ebf2509a0465a0f20025ff03ece117b67fd885/include/open62541/config.h.in#L278

@jpfr If UA_ENABLE_MALLOC_SINGLETON is set, stdlib.h is not included in the config.h file, but the test files still use definitions from stdlib.h.
Should we include stdlib.h in config.h independent of UA_ENABLE_MALLOC_SINGLETON?

jpfr:
Then I prefer moving those uses of a "naked calloc" to UA_calloc instead.
It is even good that stdlib.h is not included when the malloc singleton is used.
So we get a compiler error if a "naked calloc" is used by mistake.

keba-estr:
I removed the commit for the ua_pubsub_manager.c due to the commit "refactor(pubsub): Use UA_calloc everywhere in ua_pubsub_manager.c".
The test files need the include because of the usage of EXIT_SUCCESS.

jpfr:
Nice, merged.

PR #6473, opened 2024-05-13
fix(tests): Only build add_node_integration_test with UA_ENABLE_XML_E…

keba-estr:
…NCODING

The function UA_decodeXml is needed by the add_node_integration_test in client.c, but only available with the define UA_ENABLE_XML_ENCODING.


jpfr:
👍

PR #6475, opened 2024-05-13
Update debian packaging

jpfr:


keba-estr:
@jpfr: It seems that the shema files have been removed with the commit [refactor(build): Move the cmake files into the -tools debian package](https://github.com/open62541/open62541/pull/6475/commits/8bb1e03bb7fa69ef6c6ca2de55f8e88ecafeced1).
Did you mean 'Only install..' instead of 'Also install..'?

jpfr:
The debian package is built for the reduced information model.
So we deliver the nodeset files that are actually baked into the server.

People who want the full information model have to build from source.
And get the nodeset files directly from the OPC Foundation via the git submodule.

Is this a problem for your workflow?

keba-estr:
> The debian package is built for the reduced information model. So we deliver the nodeset files that are actually baked into the server.
> 
> People who want the full information model have to build from source. And get the nodeset files directly from the OPC Foundation via the git submodule.
> 
> Is this a problem for your workflow?

I was using the file usr/share/open62541/tools/ua-nodeset/Schema/Opc.Ua.NodeSet2.xml for generating the nodeset of namespace 0. Do I have to use a file via the git submodule now?

PR #6478, opened 2024-05-15
fix(build): Declare tools package architecture-specific

keba-estr:
Due to commit 6091b1865f87acf3285b45b06c38d740cab40cd5 the tools package now contains architecture-specific
content (usr/lib/*/cmake/open62541/*).
Thus the package must be declared with Architecture 'any' instead of
'all'.


keba-estr:
Dependency to python3 does not work with Architecture any.

keba-estr:
With Architecture 'any' the dependency from the architecture-independent package 'Python3' is not working any more.

jpfr:
Yes. Nice.

PR #6479, opened 2024-05-15
[REVIEW] Implement filestore backend for certificate groups

NoelGraf:


mcpat:
Hi Noel,
if I understand this PR correctly, the filestore backend cannot be explicitly disabled at the moment. Is a feature flag (via CMake) planned so far?

plugins/crypto/openssl/ua_certificategroup_openssl.c
```diff
@@ -839,6 +841,51 @@ UA_CertificateUtils_getSubjectName(UA_ByteString *certificate,
...
+    if(certificate == NULL || thumbprint->length != (SHA1_DIGEST_LENGTH * 2))
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+
+    BIO *certBio = BIO_new_mem_buf(certificate->data, certificate->length);
+    if(!certBio)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    X509 *cert = PEM_read_bio_X509(certBio, NULL, 0, NULL);
```
mcpat, line 856: This looks rather restrictive. Why not allowing DER here too (like MbedTLS nicely does for us).

plugins/crypto/ua_certificategroup_filestore.c
```diff
@@ -0,0 +1,717 @@
...
+    /* Memory cert store as a base */
+    UA_CertificateGroup *store;
+
+    int inotifyFd;
+
+    UA_String trustedCertFolder;
+    UA_String trustedCrlFolder;
+    UA_String issuerCertFolder;
+    UA_String issuerCrlFolder;
+    UA_String rejectedCertFolder;
```
mcpat, line 46: Persistence of rejected certificates is mainly a nice-to-have because it is not required by the spec. Nevertheless I just wanted to ask, if there are any limits applied on how many rejected certificates are kept?

CMakeLists.txt
```diff
@@ -1073,6 +1073,10 @@ list(APPEND plugin_sources
      ${PROJECT_SOURCE_DIR}/plugins/crypto/openssl/ua_certificategroup_memorystore_openssl.c)
 endif()
 
+if(UA_ENABLE_ENCRYPTION_MBEDTLS OR UA_ENABLE_ENCRYPTION_OPENSSL OR UA_ENABLE_ENCRYPTION_LIBRESSL OR UA_ENABLE_AMALGAMATION)
```
jpfr, line 1076: `UA_ENABLE_ENCRYPTION` could be enough to test.

NoelGraf:
Hi Marcel,
We do not currently plan to add a CMake flag, as we aim to avoid increasing the number of flags.

mcpat:
Hi Noel,
so the filestore backend will automatically be disabled if the target platform is neither POSIX nor Windows then? This is usually the case when building the open62541 stack for embedded platforms.

plugins/crypto/ua_certificategroup_filestore.c
```diff
@@ -0,0 +1,764 @@
...
+    if(dir == NULL) {
+        return 1;
+    }
+
+    if(!stat(dir, &sb)) {
+        /* Directory already exist */
+        return 0;
+    }
+
+    char* tmp_dir = copyStr(dir);
```
jpfr, line 73: copyStr is not needed.
It uses strlen internally - so the input to copyStr is already null-terminated.

The entire code is strange. mkpath calls itself recursively???

plugins/crypto/ua_certificategroup_filestore.c
```diff
@@ -0,0 +1,764 @@
...
+
+    retval = UA_Array_resize((void **)list, listSize, numCerts, &UA_TYPES[UA_TYPES_BYTESTRING]);
+    if (retval != UA_STATUSCODE_GOOD) {
+        return retval;
+    }
+
+    /* Read files from directory */
+    size_t numActCerts = 0;
+    dir = opendir(listPath);
+    if(dir) {
```
jpfr, line 249: You can reset dir with `rewinddir`.
That cannot fail.

plugins/crypto/ua_certificategroup_filestore.c
```diff
@@ -0,0 +1,764 @@
...
+        struct dirent *dirent;
+        while((dirent = readdir(dir)) != NULL) {
+            if(dirent->d_type == DT_REG) {
```
jpfr, line 252: Use continue to skip instead of nested ifs.

plugins/crypto/ua_certificategroup_filestore.c
```diff
@@ -0,0 +1,764 @@
...
+
+    return UA_STATUSCODE_GOOD;
+}
+
+static UA_StatusCode
+FileCertStore_getTrustList(UA_CertificateGroup *certGroup, UA_TrustListDataType *trustList) {
+    /* Check parameter */
+    if (certGroup == NULL || trustList == NULL) {
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
```
jpfr, line 559: Dont add { } guards for a single instruction-line after an if.

plugins/crypto/ua_certificategroup_filestore.c
```diff
@@ -0,0 +1,764 @@
...
+        UA_Array_delete(rejectedList, rejectedListSize, &UA_TYPES[UA_TYPES_BYTESTRING]);
+    }
+
+    return retval;
+}
+
+static void
+FileCertStore_clear(UA_CertificateGroup *certGroup) {
+    /* check parameter */
+    if (certGroup == NULL) {
```
jpfr, line 674: Do it as this:
```c
if(!certGroup || !certGroup->context)
    return;
```

Then you also don't need the if-clause below.

jpfr (commented):
Looks good! Comments inline.

plugins/crypto/ua_certificategroup_filestore.c
mcpat: Why not? It seems to be a bad habbit of not doing it and results in cleaner (here: unambiguous) code?

PR #6486, opened 2024-05-17
fix(build): Add libopen62541.so to libopen62541-dev package

keba-estr:
Has been removed with commit 9e0b766fbba73be8d703b6ccbfdeec5bfd200b48. The so file is needed for linking.


jpfr:
The so file is in the "main" package already.

keba-estr:
See discussion in issue

PR #6488, opened 2024-05-20
fix(arch): Remove assertion with invalid assumption

### LINKED ISSUE #6455: Failed mutex assertion with TCP_sendWithConnection() in 1.4

When issuing concurrent calls to both `UA_Client_run_iterate()` and `__UA_Client_AsyncService()`, there is a race condition where the following assertion at the beginning of `TCP_sendWithConnection()` may fail:

https://github.com/open62541/open62541/blob/84347820c8550b5750f2cd581c14ab201611c579/arch/eventloop_posix_tcp.c#L627-L629

This manifests itself in the following error at runtime. Both functions are marked `UA_THREADSAFE`:

```
Assertion failed: (lock->mutexCounter == num), function UA_LOCK_ASSERT, file config.h, line 426.
```

My guess is that this happens when `__UA_Client_AsyncService()` tries to send out the request while another thread holds the mutex lock while iterating the event loop from within `UA_Client_run_iterate()`. Looking at the log output preceding the failed assertion, it seems that at the same time as sending the request, the response to another request arrives at the client:

```
Iterate the EventLoop
Process delayed callbacks
TCP 5        | SC 262911     | Sending request with RequestId 17 of type BrowseRequest
Processing event 1 on fd 5
TCP 5        | Activity on the socket
TCP 5        | SC 262911     | Send from a symmetric message buffer of length 131072 a message of header+payload length of 9153
TCP 5        | Allocate receive buffer
TCP 5        | SC 262911     | Send from a symmetric message buffer of length 131072 a message of length 9153
Assertion failed: (lock->mutexCounter == num), function UA_LOCK_ASSERT, file config.h, line 426.
```

Used CMake options:

```bash
cmake ..
```

## Checklist

 - [x] open62541 Version (release number or git tag): [v1.4.0](https://github.com/open62541/open62541/releases/tag/v1.4.0)
 - [x] Operating system: macOS
 - [ ] Critical issue

sgoll:
## Description

This removes an assertion that is based on an assumption that does not hold true in certain concurrent uses of the async client on Posix platforms. See #6455 for a more detailed description.

Fixes #6455

### DIFF
--- a/arch/eventloop_posix_tcp.c
+++ b/arch/eventloop_posix_tcp.c
@@ -633,9 +633,8 @@ TCP_shutdownConnection(UA_ConnectionManager *cm, uintptr_t connectionId) {
 static UA_StatusCode
 TCP_sendWithConnection(UA_ConnectionManager *cm, uintptr_t connectionId,
                        const UA_KeyValueMap *params, UA_ByteString *buf) {
-    /* Don't have a lock and don't take a lock. As the connectionId is the fd,
-     * no need to to a lookup and access internal data strucures. */
-    UA_LOCK_ASSERT(&((UA_EventLoopPOSIX*)cm->eventSource.eventLoop)->elMutex, 0);
+    /* We may not have a lock. But we need not take it. As the connectionId is
+     * the fd, no need to do a lookup and access internal data strucures. */
 
     /* Prevent OS signals when sending to a closed socket */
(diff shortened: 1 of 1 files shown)


jpfr:
Nice. I changed the target branch to 1.4.
Please rebase and use `fix(arch)` in the commit message.

PR #6501, opened 2024-05-24
fix(server) transform to uppercase in comparing open62541_crypto_plugin

WangWeiLin-MV:
Fixed `@open62541_crypto_plugin@` to uppercase that same as `set(UA_ENCRYPTION_PLUGINS "MBEDTLS" "OPENSSL" "LIBRESSL")` in [v1.4.1/CMakeLists.txt#L144](https://github.com/open62541/open62541/blob/v1.4.1/CMakeLists.txt#L144)


jpfr:
👍

PR #6505, opened 2024-05-25
Fix CMake git working directory

OgreTransporter:
If CMake is not executed in the source code directory, there is an error because Git cannot find the repository. It is therefore necessary to specify the working directory for Git.


jpfr:
👍

Issue #6487, opened 2024-05-18
Add version resource to Windows shared library DLL


ahuj9:
## Description

Open62541 when built as a shared library on Windows could use a version resource .rc file configured from ${OPEN62541_VER_X} CMake variables.

jpfr:
Sounds good. Where should this be installed?

irostagno:
The easy way is to add [this](https://github.com/halex2005/CMakeHelpers/blob/master/generate_product_version.cmake), [this](https://github.com/halex2005/CMakeHelpers/blob/master/VersionInfo.in) and [this](https://github.com/halex2005/CMakeHelpers/blob/master/VersionResource.rc) to `tools/cmake`. Then in the `CMakeLists.txt` you should add:
```
generate_product_version(
    open62541_resource_version
    NAME "open62541"
    VERSION_MAJOR ${OPEN62541_VER_MAJOR}
    VERSION_MINOR $OPEN62541_VER_MINOR}
    VERSION_PATCH ${OPEN62541_VER_PATCH}
    VERSION_REVISION ${OPEN62541_VER_LABEL}
)
```

This generates variable `open62541_resource_version` to add into the `add_library` command:
```
add_library(open62541 ... ${open62541_resource_version})
```

For Windows, metadata is embedded into the DLL, so, no installation is needed.

OgreTransporter:
It is not necessary to make it so complicated with macros and different files. A simple template and `configure_file` is completely sufficient.

PR #6506, opened 2024-05-25
feat(build): Add Windows version resource file to DLL (issue #6487)

OgreTransporter:
Add Windows version resource file to DLL. Solve issue #6487.


ahuj9:
Looks good to me. Their msys/clang build pipeline isn't setting CMAKE_RC_COMPILE_OBJECT for windres though. My theory is they are installing msys2 instead of mingw on chocolatey, I don't see that here.

PR #6515, opened 2024-05-29
[REVIEW] Add GDS push management functionality to the information model

NoelGraf:


CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ option(UA_ENABLE_DISCOVERY "Enable Discovery Service (LDS)" ON)
 option(UA_ENABLE_JSON_ENCODING "Enable JSON encoding" ON)
 option(UA_ENABLE_XML_ENCODING "Enable XML encoding (EXPERIMENTAL)" OFF)
 option(UA_ENABLE_NODESETLOADER "Enable nodesetLoader public API" OFF)
+option(UA_ENABLE_PUSHMANAGEMENT "Enable pushManagement support" OFF)
```
jpfr, line 111: The build flag should contain the term "GDS".
For example `UA_ENABLE_GDS_PUSHMANAGEMENT`.
Can we differentiate the name for push-management between the GDS appliance and the receiving end?

src/server/ua_server_ns0_pushmanagement.c
```diff
@@ -0,0 +1,1126 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ *    Copyright 2024 (c) Fraunhofer IOSB (Author: Noel Graf)
+ */
+
+#include "ua_server_internal.h"
```
jpfr, line 8: Rename the file to `ua_server_ns0_gds.c`.

src/server/ua_server_ns0_pushmanagement.c
```diff
@@ -0,0 +1,1126 @@
...
+
+    /* check boundaries */
+    if((size_t)length >= fileContext->file->length) {
+        length = fileContext->file->length;
+    }
+    if((size_t)length >= (fileContext->file->length - fileContext->currentPos)) {
+        length = (fileContext->file->length - fileContext->currentPos);
+    }
+
+    UA_ByteString *readBuffer = UA_ByteString_new();
```
jpfr, line 620: This can be done easier once the length is correct.
```c
UA_ByteString localCopy = fileContext->file;
localCopy.length = length;
localCopy.data += fileContext->currentPos;
UA_ByteString_copy(&localCopy, readBuffer);
fileContext->currentPos += length;
```

And of course check the result of the _copy as well.

jpfr (commented):
Looks good!
Comments inline.

mcpat:
Hi Noel,
the "LastUpdateTime" property of TrustListType seems to be missing (if I'm not mistaken).

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1124 @@
...
+    UA_CertificateGroup *certGroup = getCertGroup(server, objectId);
+    if(certGroup == NULL) {
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+    }
+
+    UA_CertificateGroup *transactionCertGroup =
+        UA_Transaction_getCertificateGroup(server->transaction, certGroup);
+    if(transactionCertGroup == NULL) {
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
```
AdanRuoff, line 295: As far as I understand it, both the "AddCertificate" and "RemoveCertificate" Methods are supposed to work atomically, bypassing the transaction mechanism completely. As they do not return an "applyChanges" boolean, a user would not be able to derive if the Method started a transaction and hence requires a subsequent call to "applyChanges" later - or not.
What is more, the specification states that these functions shall abort with Bad_TransactionPending when a transaction is currently active on the specific CertificateGroup, see
https://reference.opcfoundation.org/GDS/v105/docs/7.8.2.4

That said, I think the code should rather look like this:
`    /* abort when a transaction is active on this CertGroup */
    if(UA_GDSTransaction_containsCertificateGroup(server->transaction, certGroup->certificateGroupId)) {
        return UA_STATUSCODE_BADTRANSACTIONPENDING;
    }
`

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1124 @@
...
+            }
+            break;
+        }
+    }
+
+    UA_CertificateGroup *transactionCertGroup =
+        UA_Transaction_getCertificateGroup(server->transaction, certGroup);
+    if(transactionCertGroup == NULL) {
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
```
AdanRuoff, line 376: See comment for "addCertificate".

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+                list.issuerCertificates = certificate;
+                list.issuerCertificatesSize = 1;
+            }
+            break;
+        }
+    }
+
+    if(list.specifiedLists != UA_TRUSTLISTMASKS_NONE) {
+        retval = certGroup->removeFromTrustList(certGroup, &list);
+    } else {
```
mcpat, line 374: Even though not explicitly required by the OPC UA specification it seems not be an error to validate the complete trust list again. At least if an issuer certificate is removed here, to ensure, that all remaining certificate chains in the trust list are still intact.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+    UA_Byte fileOpenMode = *(UA_Byte*)input[0].data;
+
+    UA_CertificateGroup *certGroup = getCertGroup(server, objectId);
+    if(!certGroup)
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+
+    if(fileOpenMode & (UA_OPENFILEMODE_WRITE | UA_OPENFILEMODE_ERASEEXISTING)) {
+        if(!server->transaction)
+            server->transaction = UA_GDSTransaction_new(*sessionId);
+
```
mcpat, line 407: Well, that's a deviation from the specification. A call to Open with "Write+EraseExisting" shall only return a BADTRANSACTIONPENDING if a transaction is already pending. But it shall not create a new one. This is the task for the CloseAndUpdate method (see https://reference.opcfoundation.org/GDS/v105/docs/7.8.2.3).
And further, any change during Open and CloseAndUpdate have to be cached somewhere else and are only added to a transaction, if CloseAndUpdate is successfully called. In all other cases the changes done so far are silently discarded without ever being part of a transaction.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+    if(retval != UA_STATUSCODE_GOOD)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    UA_CertificateGroup *transactionCertGroup =
+        UA_GDSTransaction_getCertificateGroup(server->transaction, certGroup);
+    if(!transactionCertGroup)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    return transactionCertGroup->setTrustList(transactionCertGroup, &trustList);
```
mcpat, line 671: As stated above, any change done through the File API must be cached elsewhere and is not part of a transaction.

A simple "Close()" will just discard those changes as will a failing "CloseAndUpdate()" call (see https://reference.opcfoundation.org/GDS/v105/docs/7.8.2.3: "If errors occur, the new TrustList is discarded. ")

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+}
+
+static UA_StatusCode
+closeAndUpdateTrustList(UA_Server *server,
+               const UA_NodeId *sessionId, void *sessionHandle,
+               const UA_NodeId *methodId, void *methodContext,
+               const UA_NodeId *objectId, void *objectContext,
+               size_t inputSize, const UA_Variant *input,
+               size_t outputSize, UA_Variant *output) {
+    UA_StatusCode retval = UA_Server_applyChanges(server);
```
mcpat, line 728: As stated above: CloseAndUpdate initiates a new transaction (or appends to an existing one) but does not end a transaction.

mcpat (commented):
I added some comments regarding the transaction management.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+    if(!certGroup)
+        return UA_STATUSCODE_BADINVALIDARGUMENT;
+
+    /* This Method cannot be called if the containing TrustList Object is open */
+    if(certGroup->fileInfoContext)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    UA_TrustListDataType trustList;
+    memset(&trustList, 0, sizeof(UA_TrustListDataType));
+    trustList.specifiedLists = UA_TRUSTLISTMASKS_ALL;
```
mcpat, line 333: It is sufficient to request either UA_TRUSTLISTMASKS_ISSUERCERTIFICATES or UA_TRUSTLISTMASKS_TRUSTEDCERTIFICATES here to somewhat reduce the overhead to a minimum.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+    size_t certificatesSize = 0;
+    certGroup->getTrustList(certGroup, &trustList);
+
+    if(isTrustedCertificate) {
+        certificates = trustList.trustedCertificates;
+        certificatesSize = trustList.trustedCertificatesSize;
+    } else {
+        certificates = trustList.issuerCertificates;
+        certificatesSize = trustList.issuerCertificatesSize;
+    }
```
mcpat, line 345: You could also take the target addresses here where to finally place the certificate pointer and size information into. This would make the for loop below even smaller.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+    if(fileContext->openFileMode & UA_OPENFILEMODE_READ)
+        return UA_STATUSCODE_BADINVALIDSTATE;
+
+    UA_TrustListDataType trustList;
+    memset(&trustList, 0, sizeof(UA_TrustListDataType));
+     UA_StatusCode retval =
+         UA_decodeBinary(&data, &trustList, &UA_TYPES[UA_TYPES_TRUSTLISTDATATYPE], NULL);
+
+    if(retval != UA_STATUSCODE_GOOD)
+        return UA_STATUSCODE_BADINTERNALERROR;
```
mcpat, line 664: The expectation, that only complete TrustListDataType objects are written in an atomic way seems not to be reasonable. The OPC UA specification states: 

> The TrustList file is a UA Binary encoded stream containing an instance of TrustListDataType (see 7.8.2.6).

Therefore I suppose that incremental writes are likely to occur.

AdanRuoff:
It looks like the OpenCount variable (though maintained in the "UA_FileInfo" struct) is not made available to the information model yet.

AdanRuoff:
Open file handles on the (virtual) TrustList file should get closed/freed when the associated Session ends. Maybe I missed something, but at least in _my_ tests it is currently possible to lock yourself out of accessing the TrustList ever again by first "Open"ing the TrustList with "Write+EraseExisting", and then closing the Session. Future sessions cannot "Open" the TrustList anymore, as there is an open "Write" handle, but they are also unable to close the handle, as the FileContext is (correctly) bound to the Session ID.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+    UA_NodeId trustListType = UA_NODEID_NUMERIC(0, UA_NS0ID_TRUSTLISTTYPE);
+    if(UA_NodeId_equal(&objectType->head.nodeId, &trustListType)) {
+        return openTrustList(server, sessionId, sessionHandle, methodId, methodContext, objectId, objectContext,
+                             inputSize, input, outputSize, output);
+    }
+
+    UA_LOG_ERROR(server->config.logging, UA_LOGCATEGORY_SERVER,
+                 "File type functions are currently only supported for TrustList types");
+    return UA_STATUSCODE_BADNOTIMPLEMENTED;
+}
```
mcpat, line 885: Valgrind tells us that this is not good. You fetch two nodes from the NodeStore here without releasing them. But it is also not a good style for the object-oriented UA information model.

May I instead suggest, that you register the doSomething**File** methods to the FileType directly and all doSomething**TrustList** to the TrustListType and below? If you do this you can remove this checking code from all doSomethingFile functions and therefore prevent leakage of the TrustListType and the TrustList-object nodes.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1111 @@
...
+    retval |= deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTUSERTOKENGROUP_TRUSTLIST_OPEN),true);
+    retval |= addRef(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTUSERTOKENGROUP_TRUSTLIST),
+                     UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT),
+                     UA_NODEID_NUMERIC(0, UA_NS0ID_FILETYPE_OPEN), true);
+
+    /* Read */
+    retval |= deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTAPPLICATIONGROUP_TRUSTLIST_READ),true);
+    retval |= addRef(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTAPPLICATIONGROUP_TRUSTLIST),
+                     UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT),
+                     UA_NODEID_NUMERIC(0, UA_NS0ID_FILETYPE_READ), true);
```
mcpat, line 1050: We stumbled upon this code several times. Locally we decided not to remove the original method nodes but to register the callbacks to them to keep their original node IDs. But doing this we now unveilled an issue with the method resolution #6740.

So is this removal and re-referencing code here just a workaround for the problem or is there also more about it?

include/open62541/plugin/certificategroup.h
```diff
@@ -21,6 +21,12 @@ struct UA_CertificateGroup {
     /* The nodeId of the certificate group this pki store is associated with */
     UA_NodeId certificateGroupId;
     void *context;
+    /* Used internally to create a FileType from the TrustList,
+     * which can then be read and written using the FileType functions.
+     * Only used when the GDS push information model is enabled (UA_ENABLE_GDS_PUSHMANAGEMENT).
+     */
+    void *fileInfoContext;
```
jpfr, line 28: This somewhat pollutes a public structure.
It would be good to avoid that.

src/server/ua_server_ns0.c
```diff
@@ -1136,7 +1136,6 @@ initNS0(UA_Server *server) {
     deleteNode(server, UA_NS0ID(SERVER_REQUESTSERVERSTATECHANGE), true);
     deleteNode(server, UA_NS0ID(SERVER_SETSUBSCRIPTIONDURABLE), true);
     deleteNode(server, UA_NS0ID(SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTHTTPSGROUP), true);
-    deleteNode(server, UA_NS0ID(SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTUSERTOKENGROUP), true);
```
jpfr, line 1139: You want to remove that always or only when the gds is enabled?

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1338 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ *    Copyright 2024 (c) Fraunhofer IOSB (Author: Noel Graf)
+ */
+
+#include "ua_server_internal.h"
+#include <open62541/plugin/log_stdout.h>
```
jpfr, line 9: Don’t include the logger.
The core library never imports a specific plugin.
Rather use the logger from the configuration.

jpfr (commented):
Comments inline.

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -937,41 +932,19 @@ UA_CertificateUtils_comparePublicKeys(const UA_ByteString *certificate1,
 UA_StatusCode
 UA_CertificateUtils_ckeckKeyPair(const UA_ByteString *certificate,
```
mcpat, line 933: c**k**eckKeyPair -> c**h**eckKeyPair

src/server/ua_server.c
```diff
@@ -701,10 +706,9 @@ secureChannel_delayedCloseTrustList(void *application, void *context) {
     UA_CertificateGroup certGroup = server->config.secureChannelPKI;
     UA_SecureChannel *channel;
     TAILQ_FOREACH(channel, &server->channels, serverEntry) {
-        const UA_SecurityPolicy *policy = channel->securityPolicy;
         if(channel->state != UA_SECURECHANNELSTATE_CLOSED && channel->state != UA_SECURECHANNELSTATE_CLOSING)
             continue;
-        if(certGroup.verifyCertificate(&certGroup, &policy->localCertificate) != UA_STATUSCODE_GOOD)
+        if(certGroup.verifyCertificate(&certGroup, &channel->remoteCertificate) != UA_STATUSCODE_GOOD)
```
jpfr, line 711: Do we need to fix that for 1.4 also?

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1559 @@
...
+#include "ua_server_internal.h"
+
+#ifdef UA_ENABLE_GDS_PUSHMANAGEMENT
+
+#define UA_SHA1_LENGTH 20
+#define CHECKACTIVESESSIONINTERVAL 10000 /* 10sec */
+
+typedef struct UA_FileContext {
+    LIST_ENTRY(UA_FileContext) listEntry;
+    UA_ByteString *file;
```
jpfr, line 17: This could be in-situ instead of a pointer to the bytestring.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1559 @@
...
+} UA_FileInfo;
+
+typedef struct UA_FileInfoContext {
+    UA_NodeId certificateGroupId;
+    UA_FileInfo fileInfo;
+} UA_FileInfoContext;
+
+static UA_StatusCode applyChangesToServer(UA_Server *server);
+
+/* There are currently only two certification groups that are supported */
```
jpfr, line 37: Which of the groups has the first slot?
Document this here.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1559 @@
...
+    }
+
+    if(removeCallback) {
+        UA_Server_removeCallback(server, checkSessionCallbackId);
+        checkSessionCallbackId = 0;
+    }
+}
+
+static UA_Boolean
+compareThumbprint(const UA_String *str1, const UA_String *str2) {
```
jpfr, line 257: You can reuse `UA_String_equal_ignorecase`
https://github.com/open62541/open62541/blob/cd0aa1a987231b1b3cdf8145c30ce958f91b9fe6/src/util/ua_util_internal.h#L320

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1559 @@
...
+        UA_ByteString_delete(fileContext->file);
+        UA_free(fileContext);
+        return retval;
+    }
+
+    fileInfo->openCount += 1;
+    UA_Variant_setScalarCopy(output, &fileContext->fileHandle, &UA_TYPES[UA_TYPES_UINT32]);
+
+    /* Updating OpenCount Varialbe in the information model */
+    UA_LOCK(&server->serviceMutex);
```
jpfr, line 704: Don't you need the lock for the entire callback?
Otherwise a parallel thread could work on the same data structures.

src/server/ua_server_ns0_gds.c
```diff
@@ -0,0 +1,1559 @@
...
+typedef struct UA_FileInfoContext {
+    UA_NodeId certificateGroupId;
+    UA_FileInfo fileInfo;
+} UA_FileInfoContext;
+
+static UA_StatusCode applyChangesToServer(UA_Server *server);
+
+/* There are currently only two certification groups that are supported */
+size_t fileInfoContextSize = 2;
+UA_FileInfoContext fileInfoContext[2];
```
jpfr, line 39: Global variables should be const.
Mutable variables should go to the server struct.
So that we can have multiple server instances in the same application.

Either put it directly into the server struct.
Or define a "GDS manager struct" which aggregates them.

jpfr (commented):
Lots of progress! Good good.
Comments inline.

src/server/ua_server.c
NoelGraf: No. This is only relevant for the master.

jpfr:
👍

PR #6519, opened 2024-05-31
[REVIEW] fix(build): Correct path for schema files of external build

### LINKED ISSUE #6508: Case mismatch for 'Schema' directories
## Description
Nodeset generation might fail on case dependent Linux systems, as the directory referenced in the file open62541Macros.cmake is '${UA_NODESET_DIR}/Schema', but in the debian tools package the folder name is 'schema' with lower case letters.

## Checklist

 - [X] open62541 Version (release number or git tag): 1.4.1
 - [X] Operating system: Linux

keba-estr:
Use UA_SCHEMA_DIR for path to schema file to be independent from the different case of the 'schema' and 'Schema' folder.
Resolves #6508


jpfr:
👍

PR #6521, opened 2024-06-01
Role-based Access Control

jpfr:


src/server/ua_nodes.c
```diff
@@ -546,6 +546,9 @@ UA_Node_copy(const UA_Node *src, UA_Node *dst) {
     UA_StatusCode retval = UA_NodeId_copy(&srchead->nodeId, &dsthead->nodeId);
     retval |= UA_QualifiedName_copy(&srchead->browseName, &dsthead->browseName);
 
+    /* Copy the RolePermissions */
+    memcpy(dsthead->rolePermissions, srchead->rolePermissions, sizeof(UA_RoleSet) * 16);
```
mcpat, line 550: Why not using sizeof(dsthead->rolePermissions) for the size or the named constant "UA_ROLEPERMISSIONS_COUNT" instead of "16" here?

plugins/ua_config_default.c
```diff
@@ -334,7 +334,10 @@ setDefaultConfig(UA_ServerConfig *conf, UA_UInt16 portNumber) {
         }
     }
 
-    conf->tcpReuseAddr = false;
+    /* If a second server is started later it can "steal" the port.
+     * Having port reuse enabled is important for development.
+     * Otherwise a long TCP TIME_WAIT is required before the port can be used again. */
+    conf->tcpReuseAddr = true;
```
mcpat, line 340: This change seems to be off-topic here.

include/open62541/util.h
```diff
@@ -175,6 +175,83 @@ UA_EXPORT extern const UA_ReferenceTypeAttributes UA_ReferenceTypeAttributes_def
...
+#define UA_ROLEINDEX_AUTHENTICATEDUSER 1
+#define UA_ROLEINDEX_OBSERVER 2
+#define UA_ROLEINDEX_OPERATOR 3
+#define UA_ROLEINDEX_ENGINEER 4
+#define UA_ROLEINDEX_SUPERVISOR 5
+#define UA_ROLEINDEX_CONFIGUREADMIN 6
+#define UA_ROLEINDEX_SECURITYADMIN 7
+
+typedef struct {
+    UA_UInt16 bits[UA_ROLESET_MAX / 16];
```
mcpat, line 195: Using "(UA_ROLESET_MAX + 15) / 16" here might relax the current restriction of UA_ROLESET_MAX being a multiple of 16.

src/server/ua_services_nodemanagement.c
```diff
@@ -1982,6 +1999,19 @@ deleteNodeOperation(UA_Server *server, UA_Session *session, void *context,
...
+    if(session != &server->adminSession &&
+       !UA_RoleSet_intersects(delNodeRoles, session->roles)) {
+        UA_LOG_NODEID_INFO(&node->head.nodeId,
+        UA_LOG_INFO_SESSION(server->config.logging, session, "DeleteNode (%.*s): "
+                            "Cannot delete the node because of missing RolePermisions",
+                            (int)nodeIdStr.length, nodeIdStr.data));
+        UA_NODESTORE_RELEASE(server, node);
+        *result = UA_STATUSCODE_BADUSERACCESSDENIED;
+        return;
+    }
```
mcpat, line 2013: If I understand this correctly, this checks, if the session has a role assigned which is required to gain the DELETENODE permission. What is the difference of this check and the one in the AccessControl Plugin (allowDeleteNode)?

mcpat:
Hi,
I just want to place an additional conceptual question here:

It seems, that the new permission checks are now hard-wired into several places instead of being encapsulated in the AccessControl plugin. Is this desired? What purpose does the AccessControl plugin have then if the permission checks are not done there?

include/open62541/server.h
```diff
@@ -607,6 +626,105 @@ UA_EXPORT UA_StatusCode UA_THREADSAFE
...
 
+/**
+ * .. _roles-access:
+ *
+ * Role Permissions and Access Control
+ * -----------------------------------
+ *
+ * OPC UA uses role-based access control (RBAC). Every session is assigned a set
+ * of roles it represents. Every node in the information contains
+ * RolePermissions that define the rights of each role.
```
mcpat, line 637: This does not seem to be correct. When looking into the [specification](https://reference.opcfoundation.org/Core/Part3/v105/docs/5.2.9) it states:

> If not specified, the value of [DefaultRolePermissions](https://reference.opcfoundation.org/search/356?t=DefaultRolePermissions) [Property](https://reference.opcfoundation.org/search/356?t=Property) from the [NamespaceMetadata](https://reference.opcfoundation.org/search/356?t=NamespaceMetadata) [Object](https://reference.opcfoundation.org/search/356?t=Object) associated with the [Node](https://reference.opcfoundation.org/search/356?t=Node) shall be used instead.

My understanding is,  that not every node in the information model contains RolePermissions,

include/open62541/plugin/nodestore.h
```diff
@@ -391,12 +388,43 @@ typedef struct UA_LocalizedTextListEntry {
...
     UA_NodeId nodeId;
     UA_NodeClass nodeClass;
     UA_QualifiedName browseName;
 
+    /* The PermissionType defines 17 different RolePermissions. The AddNode role
+     * is global and not defined for each node. For all other 16 PermissionTypes
+     * the node keeps a role-set of all the roles which have this permission.
+     * When a RolePermission is changed manually (added or removed) the changes
+     * propagate recursively to all child nodes (hierarchical reference). */
+    UA_RoleSet rolePermissions[UA_ROLEPERMISSIONS_COUNT];
```
mcpat, line 426: This looks like quite some overhead here. Beside that not every node actually has RolePermissions assigned (see comment below) it also poses an upper limit of roles that may be added at runtime.

Wouldn't it be better to keep the RolePermissions outside of the node and accessing them (or only check for access) through an abstraction? That way it would reduce the overhead here and it would also allow to deviate from the standard role mapping which is supported by the OPC UA specification?

include/open62541/server.h
jpfr: Every node can have it (optional).
We would chose to have the overhead for every node.

include/open62541/plugin/nodestore.h
jpfr: These are 32byte overhead per node.
16 permission types and 2 byte per type (assuming up to 16 roles).

src/server/ua_nodes.c
jpfr: Good find. Will change that.

plugins/ua_config_default.c
jpfr: Yes. Will be moved out of this PR.

src/server/ua_services_nodemanagement.c
jpfr: The AccessControl plugin checks go on top of the RolePermissions.
Everything the RolePermissions do can also be done in the pure AccessControl plugin.
But it will be a lot easier to move most checks into the Role-Based Access Control.
Because you don't to build all the machinery to track node access permissions individually.

jpfr:
> It seems, that the new permission checks are now hard-wired into several places instead of being encapsulated in the AccessControl plugin. Is this desired?

We will have both.
You can set the RolePermissions to "always allow" and continue to use the AccessControl plugin as before.

mcpat:
> > It seems, that the new permission checks are now hard-wired into several places instead of being encapsulated in the AccessControl plugin. Is this desired?
> 
> We will have both. You can set the RolePermissions to "always allow" and continue to use the AccessControl plugin as before.

Well that sounds good.

include/open62541/plugin/nodestore.h
mcpat: Fair enough, the current size of UA_Node is already at around 400 Bytes so this increase is only about 8%.
And also from what I see, Namespace 0 contains around 5000 nodes. So yes, this overhead seems to be negligible.

Nevertheless I want to point two things out here, that might be worth considering:

* I would expect the amount of permutations for these RolePermissions to be strictly smaller than the overall number of nodes.
  Why? Because first, it seems to be a maintenance burden to assign different role permissions to each node.
  And second, doing this does not make sense at all because of the inter-dependencies between the nodes.
* It does not feel right to have this matrix embedded inside the UA_Node structure. ;-)
  Having a pointer here for example allows to easily detect that no RolePermissions have been assigned (e.g. to fall back to the namespace defaults). And it would also allow to add roles at runtime without any limitations.

andreasebner:
The RBAC implementation has been merged into master and fully covers the functional scope defined by this PR.

xR3b0rn:
Hi everyone,

I noticed that the RBAC feature is still marked as experimental in the top-level CMakeLists.txt. I looked through the git history to understand the background, but I couldn't find any clear clues as to why it hasn't been stabilized yet. Could someone provide some context on its current status? I'd love to know what features are still missing or if there are any specific blockers preventing it from being marked as production-ready.

PR #6523, opened 2024-06-03
[REVIEW] Correct setting of the namespaceIndex

NoelGraf:
When generating the namespace code, the namespaceIndex may not be set correctly if the nodeset is based on multiple nodesets.

The code for extracting the bsd file from the nodeset has been adjusted. The check that the variable node is the node containing the bsd as a base64-encoded byte string has been extended.


jpfr:
👍

PR #6526, opened 2024-06-04
[Review] fix(build): fix OpenSSL link flags in build output files

rickboks:
Fixed CMake imported targets `OpenSSL::Crypto` and `OpenSSL::SSL` being written literally (without being expanded) to output files `open62541Targets.cmake` and `open62541.pc`.


jpfr:
Nice, thanks.

PR #6528, opened 2024-06-04
[REVIEW] Removal of the old CertificateVerification implementation

NoelGraf:
- Removing the old CertificateVerification implementation and adapted the affected examples and tests.
- Rename UA_CertificateVerification_AcceptAll to UA_CertificateGroup_AcceptAll.
- Removal of the check_save_rejected_cert test file as the test is performed in the new check_certificategroup file.


jpfr:
With the new API that takes only a single folder we should document very good how that folder has to look like.

That is still lacking…

We could even have an example with a pki folder in the repository.

CMakeLists.txt
```diff
@@ -1048,8 +1042,7 @@ list(APPEND plugin_sources
      ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/ua_securitypolicy_aes128sha256rsaoaep.c
      ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/ua_securitypolicy_aes256sha256rsapss.c
      ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/ua_mbedtls_create_certificate.c
-     ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/ua_certificategroup_mbedtls.c
-     ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/ua_certificategroup_memorystore_mbedtls.c)
```
mcpat, line 1052: Is there a reason why this is dropped?

plugins/crypto/mbedtls/ua_certificategroup_memorystore_mbedtls.c
mcpat, line 1: Same as above: why is the memory store deleted here?

mcpat: Ah ok, I got it: you merged it ino ua_certificategroup_mbedtls.c

CMakeLists.txt
mcpat: Should have checked the whole PR before asking ;-) - the sources are merged!

NoelGraf:
@jpfr I have added additional documentation for the function and also detailed the folder structure.

jpfr:
👍

PR #6530, opened 2024-06-04
Clean soname

jpfr:


plugins/ua_config_default.c
```diff
@@ -361,11 +361,7 @@ setDefaultConfig(UA_ServerConfig *conf, UA_UInt16 portNumber) {
     conf->buildInfo.softwareVersion =
         UA_STRING_ALLOC(VERSION(UA_OPEN62541_VER_MAJOR, UA_OPEN62541_VER_MINOR,
                                 UA_OPEN62541_VER_PATCH, UA_OPEN62541_VER_LABEL));
-#ifdef UA_PACK_DEBIAN
-    conf->buildInfo.buildNumber = UA_STRING_ALLOC("deb");
-#else
     conf->buildInfo.buildNumber = UA_STRING_ALLOC(__DATE__ " " __TIME__);
```
guillemj, line 364: BTW if the above line gets removed then this will guarantee this package is not reproducible anymore. I think in general it's advised that this kind of information not to be embedded in the compiled code, or if you must, then either use a stable value such a commit id or use the established [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/).

PR #6531, opened 2024-06-04
fix(server) copy mbedtls find cmake file

andreasebner:
For MbedTLS Version (<3) e.g. delivered with Ubuntu 20LTS, we need a dedicated cmake-find file (see Mbedtls 3.0 release notes "Add CMake package config generation for CMake projects consuming Mbed TLS." https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.0.0). Therefore we have to copy the file during the make install.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1523,6 +1523,11 @@ export(TARGETS open62541
        NAMESPACE open62541::
        FILE ${CMAKE_CURRENT_BINARY_DIR}/open62541Targets.cmake)
 
+if(UA_ENABLE_ENCRYPTION STREQUAL "MBEDTLS")
+    install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/FindMbedTLS.cmake"
+            DESTINATION "${cmake_configfile_install}")
+endif()
+
 # Generate and install open62541.pc
 if(UA_ENABLE_AMALGAMATION)
     set(PC_EXTRA_CFLAGS "-DUA_ENABLE_AMALGAMATION")


jpfr:
Please check the mbedtls version for this.
I would prefer not to do this for the debian package.
So we don't collide with stuff that other projects should be doing.

Or have this in a file where we aggregate different helper macros...

keba-estr:
I also get a similar error with a debian bookworm armhf build:

CMake Error at /usr/share/cmake-3.25/Modules/CMakeFindDependencyMacro.cmake:47 (find_package):
  By not providing "FindMbedTLS.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "MbedTLS", but
  CMake did not find one.
  Could not find a package configuration file provided by "MbedTLS" with any
  of the following names:
    MbedTLSConfig.cmake
    mbedtls-config.cmake
  Add the installation prefix of "MbedTLS" to CMAKE_PREFIX_PATH or set
  "MbedTLS_DIR" to a directory containing one of the above files.  If
  "MbedTLS" provides a separate development package or SDK, be sure it has
  been installed.
  /usr/lib/arm-linux-gnueabihf/cmake/open62541/open62541Config.cmake:[50](https://gitlab.keba.com/industrial-automation/pc-control-io/firmware/opcua-open62541/-/jobs/8989630#L50) (find_dependency)
  CMakeLists.txt:34 (include)

keba-estr:
The following commit has an influence on this topic:
14c1d6fb1175257b624bcd509817887f5e8a2543

andreasebner:
> The following commit has an influence on this topic: [14c1d6f](https://github.com/open62541/open62541/commit/14c1d6fb1175257b624bcd509817887f5e8a2543)

This was an issue and is solved in the latest release. But without the change in this PR, the FindMbedtls.cmake file (needed for mbedtls versions <3.0) is not copied to the install location. Since the copy process is more or less a workarround, we currently discuss on how to include it a good manner.

andreasebner:
Solved with latest changes on 1.4.

PR #6534, opened 2024-06-05
[REVIEW] Implement a transaction structure for updating trust list and server certificates

NoelGraf:


include/open62541/plugin/certificategroup.h
```diff
@@ -40,6 +40,31 @@ struct UA_CertificateGroup {
...
 };
 
+struct UA_Transaction;
+typedef struct UA_Transaction UA_Transaction;
+
+struct UA_Transaction {
+    UA_NodeId sessionId;
+
+    size_t certGroupSize;
+    UA_CertificateGroup *certGroups;
```
mcpat, line 50: Are arbitrary certificate groups supported here or is this restricted to DefaultApplicationGroup and DefaultUserTokenGroup?

include/open62541/plugin/certificategroup.h
```diff
@@ -40,6 +40,31 @@ struct UA_CertificateGroup {
...
+
+    /* The entries in the arrays are related to each other at the same index. */
+    size_t serverCertSize;
+    UA_ByteString *certificates;
+    UA_ByteString *privateKeys;
+    UA_NodeId *certificateGroups;
+    UA_NodeId *certificateTypes;
```
mcpat, line 57: Does it make sense to combine the items (certificate, key, group and type) into a new structure to make this relation clear?

src/server/ua_server_internal.h
```diff
@@ -191,6 +191,9 @@ struct UA_Server {
     /* Statistics */
     UA_SecureChannelStatistics secureChannelStatistics;
     UA_ServerDiagnosticsSummaryDataType serverDiagnosticsSummary;
+
+    /* Transaction for certificate management */
+    UA_Transaction *transaction;
```
mcpat, line 196: Is there only one kind of transaction specified for OPC UA or should this one be renamed (and also the structure)?

src/server/ua_server.c
```diff
@@ -544,65 +676,138 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
-            UA_Session *session = &current->session;
-            if(!session->channel)
+UA_Server_applyChanges(UA_Server *server) {
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    if(!server->transaction)
+        return UA_STATUSCODE_BADNOTHINGTODO;
+
+    UA_GDSTransaction *transaction = server->transaction;
+
+    /* Apply Trust list changes */
```
mcpat, line 693: The validation of the Trust list is missing.

NoelGraf: This will be done in a separate PR.

src/server/ua_server.c
```diff
@@ -544,65 +678,174 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
 }
 
 static void
-notifySecureChannelsStopped(UA_Server *server, struct UA_ServerComponent *sc,
-                            UA_LifecycleState state) {
-    if(sc->state == UA_LIFECYCLESTATE_STOPPED &&
-       server->state == UA_LIFECYCLESTATE_STARTED) {
-        sc->notifyState = NULL; /* remove the callback */
-        sc->start(server, sc);
+sercureChannel_delayedClose(void *application, void *context) {
```
mcpat, line 681: Typo: sercure -> secure

src/server/ua_server.c
```diff
@@ -544,65 +678,174 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+
+    UA_NodeId defaultApplicationGroup =
+            UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTAPPLICATIONGROUP_TRUSTLIST);
+
+    for(size_t i = 0; i < transaction->certGroupSize; i++) {
+        UA_CertificateGroup certGroup = transaction->certGroups[i];
+        if(!UA_NodeId_equal(&certGroup.certificateGroupId, &defaultApplicationGroup))
+            continue;
+        closeRsaMin = true;
+        closeRsaSha256 = true;
```
mcpat, line 707: Why do you hardcode this here? The type can be directly taken from the CertificateInfo in the transaction object.

src/server/ua_server.c
```diff
@@ -544,65 +678,174 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
-        if(UA_ByteString_equal(&ed->serverCertificate, oldCertificate)) {
-            UA_String_clear(&ed->serverCertificate);
-            UA_String_copy(newCertificate, &ed->serverCertificate);
-            UA_SecurityPolicy *sp = getSecurityPolicyByUri(server,
-                            &server->config.endpoints[i].securityPolicyUri);
-            UA_CHECK_MEM(sp, return UA_STATUSCODE_BADINTERNALERROR);
-            sp->updateCertificateAndPrivateKey(sp, *newCertificate, *newPrivateKey);
-        }
-        i++;
+    UA_StatusCode retval = UA_GDSTransaction_addCertificateInfo(server->transaction, certificateGroupId,
```
mcpat, line 843: This looks odd. As far as I know, UA_Server_updateCertificate can be invoked either via C-API or via OPC UA. For the former case I somehow understand why server->adminSession is used. But for the latter, the OPC UA session ID must be used and of course be checked before adding changes to the transaction.

Will the transaction logic be added in a later PR and does this one only introduces some architectural foundation?

NoelGraf: When the updateCertificate function is called via OPC UA, a new transaction with the appropriate sessionId is created in the information model method (see https://github.com/open62541/open62541/pull/6515/files#diff-ea8abd26b36279bc52bb0e4dc60d56ecbfc041a9180c57e70eb78bfbcd504a02).

mcpat: Oh well, I did not see that the other PR also does some transaction stuff too. It seems, that both PRs depend on each other which I did not expect to see here. So sorry for the noise again. Still I have problems understanding why the two PRs are structured that way...

src/server/ua_server.c
```diff
@@ -173,6 +178,133 @@ UA_Server_forEachChildNodeCall(UA_Server *server, UA_NodeId parentNodeId,
...
+
+    memset(&transaction->certGroups[transaction->certGroupSize-1], 0, sizeof(UA_CertificateGroup));
+
+    UA_TrustListDataType trustList;
+    UA_TrustListDataType_init(&trustList);
+    trustList.specifiedLists = UA_TRUSTLISTMASKS_ALL;
+    certGroup->getTrustList((UA_CertificateGroup*)(uintptr_t)certGroup, &trustList);
+
+    UA_CertificateGroup_Memorystore(&transaction->certGroups[transaction->certGroupSize-1],
+        (UA_NodeId*)(uintptr_t)&certGroup->certificateGroupId, &trustList, certGroup->logging, NULL);
```
mcpat, line 224: This is not good. The `maxTrustListSize` limit for the memory store is defaulted to 64 KB here which in some cases is too low. It would be better, if it could take the same limit as the targeted certificate group...

src/server/ua_server.c
```diff
@@ -544,65 +678,174 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+
+UA_StatusCode
+UA_Server_updateCertificate(UA_Server *server,
+                            const UA_NodeId certificateGroupId,
+                            const UA_NodeId certificateTypeId,
+                            const UA_ByteString *certificate,
+                            const UA_ByteString *issuerCertificates,
+                            const size_t issuerCertificatesSize,
+                            const UA_ByteString *privateKey,
+                            const UA_String *privateKeyFormat) {
```
mcpat, line 824: In case no key is provided, we have to find a matching key here by either:

- checking against the latest not-activated one (created by a CSR)
- checking against the current activated one (that is already used in the current application instance certificate)

If neither of them matches -> an error is expected.

An error must also be return, if the provided certificate and key do not belong to each other. This checks also seems to be missing.

plugins/crypto/openssl/securitypolicy_aes128sha256rsaoaep.c
```diff
@@ -121,6 +121,12 @@ updateCertificateAndPrivateKey_sp_aes128sha256rsaoaep(UA_SecurityPolicy *securit
     Policy_Context_Aes128Sha256RsaOaep *pc =
         (Policy_Context_Aes128Sha256RsaOaep *)securityPolicy->policyContext;
 
+    UA_Boolean isLocalKey = false;
+    if(newPrivateKey.length <= 0) {
+        if(UA_CertificateUtils_comparePublicKeys(&newCertificate, &securityPolicy->localCertificate) == 0)
+            isLocalKey = true;
+    }
```
mcpat, line 128: This somehow looks like a workaround because the information whether to use the current private key or the one from the latest CSR is not passed to the callback. But this information has already been gathered in UA_Server_updateCertificate().

plugins/crypto/mbedtls/securitypolicy_aes256sha256rsapss.c
```diff
@@ -703,12 +709,19 @@ updateCertificateAndPrivateKey_sp_aes256sha256rsapss(UA_SecurityPolicy *security
...
-    mbedtls_pk_init(&pc->localPrivateKey);
-    int mbedErr = UA_mbedTLS_LoadPrivateKey(&newPrivateKey, &pc->localPrivateKey, &pc->entropyContext);
-    if(mbedErr) {
-        retval = UA_STATUSCODE_BADSECURITYCHECKSFAILED;
-        goto error;
+    if(newPrivateKey.length > 0) {
+        mbedtls_pk_free(&pc->localPrivateKey);
+        mbedtls_pk_init(&pc->localPrivateKey);
+        if(UA_mbedTLS_LoadPrivateKey(&newPrivateKey, &pc->localPrivateKey, &pc->entropyContext)) {
+            retval = UA_STATUSCODE_BADSECURITYCHECKSFAILED;
```
mcpat, line 716: This error should be UA_STATUSCODE_BADNOTSUPPORTED according to https://reference.opcfoundation.org/GDS/v105/docs/7.10.4.

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -719,6 +720,175 @@ UA_CertificateUtils_getKeySize(UA_ByteString *certificate,
...
+        retval = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    /* Initialize the hashing context with SHA-256 */
+    const mbedtls_md_info_t *md_info = mbedtls_md_info_from_type(MBEDTLS_MD_SHA256);
+    if(!md_info) {
+        retval = UA_STATUSCODE_BADOUTOFMEMORY;
+        goto cleanup;
+    }
```
mcpat, line 791: Where is this context used?

src/server/ua_server.c
```diff
@@ -521,65 +688,201 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+    UA_NodeId defaultApplicationGroup =
+            UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTAPPLICATIONGROUP_TRUSTLIST);
+
+    for(size_t i = 0; i < transaction->certGroupSize; i++) {
+        UA_CertificateGroup certGroup = transaction->certGroups[i];
+        if(!UA_NodeId_equal(&certGroup.certificateGroupId, &defaultApplicationGroup))
+            continue;
+        closeRsaMin = true;
+        closeRsaSha256 = true;
     }
```
mcpat, line 718: This may be a little too offensive. When the trust list of the DefaultApplicationGroup has been modified, only the secure channels where the certificate of the remote peer is not trusted anymore shall be closed (see https://reference.opcfoundation.org/GDS/v105/docs/7.10.6). After adding certificates for example no secure channel needs to be touched.

There is also an additional issue:
While it is valid to leave any session untouched when only UpdateCertificate() has been called, a change to the TrustList of the DefaultUserTokenGroup must be accompanied by closing all Sessions with a now untrusted / revoked certificate (see https://reference.opcfoundation.org/GDS/v105/docs/7.8.2.3). As of Spec version 1.5.4 also the underyling SecureChannels have to be closed in this case.

plugins/crypto/mbedtls/certificategroup.c
NoelGraf: You're right. It's a leftover and will be removed.

src/server/ua_server.c
NoelGraf: For this reason, the SecureChannel is simply closed if the server certificate or the TrustList changes. The client can re-establish the connection with its reconnect logic and its certificate is re-validated against the new TrustList.
This makes the whole thing relatively easy to implement.

Section of the CloseAndUpdate function:
"When the TrustList changes the Server shall re-evaluate the Certificate associated with any open Sessions. Sessions with an untrusted or revoked Certificate shall be closed."

I believe this only applies when no transaction logic is in place. With transaction logic, the changes are only applied during the apply process.

mcpat: Of course it is easy to implement. Nevertheless it leads to interruption where it is not required. 
Still certificate-authenticated sessions remain intact which in case of changes to the TrustList of the DefaultUserTokenGroup is not desired. Affected sessions shall be closed as well in this case.

plugins/crypto/ua_securitypolicy_filestore.c
```diff
@@ -178,13 +178,13 @@ writeCertificateAndPrivateKeyToFilestore(const UA_String storePath,
 
     /* Create the paths to the certificates and private key folders */
     char ownCertPathDir[PATH_MAX];
-    if(mp_snprintf(ownCertPathDir, PATH_MAX, "%.*s/%s", (int)storePath.length,
-                   (char *)storePath.data, "/own/certs") < 0)
+    if(mp_snprintf(ownCertPathDir, PATH_MAX, "%.*s%s", (int)storePath.length,
+                   (char *)storePath.data, "/pki/ApplCerts/own/certs") < 0)
```
jpfr, line 182: We wanted to get rid of the first /pki.

jpfr (commented):
Small comments inline.

My big comment is this:

A requirement is the possibility to use a TPM (trusted platform module) that generates the certificate and private key pair.
This would be implemented in a dedicated SecurityPolicy implementation.
We have talked about this several times.

In that case the private key never leaves the trusted hardware.
YOU DO NOT EVER SEE THE PRIVATE KEY.

From the perspective of the server, the private key of the last CSR simply remains cached inside the SecurityPolicy. That should not be a problem, there can be only one CSR "in flight" anyway.

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -719,6 +720,168 @@ UA_CertificateUtils_getKeySize(UA_ByteString *certificate,
...
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+
+    mbedtls_x509_crt cert1;
+    mbedtls_x509_crt cert2;
+    mbedtls_x509_csr csr1;
+    mbedtls_x509_csr csr2;
+    mbedtls_mpi N1, E1;
+    mbedtls_mpi N2, E2;
+
+    UA_ByteString data1 = UA_mbedTLS_CopyDataFormatAware(certificate1);
```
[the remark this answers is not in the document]
NoelGraf: To ensure the passed certificates can be parsed correctly, the mbedtls parse function requires the certificate string to be null-terminated, and this null character must be included in the length parameter.

plugins/crypto/ua_securitypolicy_filestore.c
```diff
@@ -178,13 +178,13 @@ writeCertificateAndPrivateKeyToFilestore(const UA_String storePath,
 
     /* Create the paths to the certificates and private key folders */
     char ownCertPathDir[PATH_MAX];
-    if(mp_snprintf(ownCertPathDir, PATH_MAX, "%.*s/%s", (int)storePath.length,
-                   (char *)storePath.data, "/own/certs") < 0)
+    if(mp_snprintf(ownCertPathDir, PATH_MAX, "%.*s%s", (int)storePath.length,
+                   (char *)storePath.data, "/ApplCerts/own/certs") < 0)
```
jpfr, line 182: Why this change?
Deep nested folders are bad.
And this subfolder thing is not documented.

include/open62541/server.h
```diff
@@ -1892,11 +1892,10 @@ UA_Server_setExpirationDate(UA_Server *server, const UA_NodeId conditionId,
...
 UA_Server_updateCertificate(UA_Server *server,
-                            const UA_ByteString *oldCertificate,
-                            const UA_ByteString *newCertificate,
-                            const UA_ByteString *newPrivateKey,
-                            UA_Boolean closeSessions,
-                            UA_Boolean closeSecureChannels);
+                            const UA_NodeId certificateGroupId,
+                            const UA_NodeId certificateTypeId,
+                            const UA_ByteString certificate,
+                            const UA_ByteString *privateKey);
```
jpfr, line 1898: This needs documentation.
Why is the privateKey a pointer and the others are not?
Can we have privateKey == NULL if we don't have it?

Otherwise make it a non-pointer argument also.

jpfr (commented):
Merging this.
Please consider the inline comments to be resolved shortly in a follow up.

include/open62541/server.h
NoelGraf: I will update the documentation. If no private key is specified, the system will attempt to use either the existing key or the one generated during the CSR creation process.

plugins/crypto/ua_securitypolicy_filestore.c
NoelGraf: I will provide more documentation.
This function is triggered when the server certificate is updated using the UA_Server_updateCertificate function. It only updates the application instance certificate. The path to the PKI folder is stored in the SecurityPolicy_FilestoreContext, so you need to specify the necessary subfolder.

PR #6535, opened 2024-06-05
[REVIEW] Adding the create signing request function

NoelGraf:
Implement the createSigningRequest function to create a PKCS # 10 DER encoded certificate request signed with the server's private key.
- [x] Done for mbedTLS
- [x] Done for openSSL 
- [x] Add private key regeneration functionality


plugins/crypto/mbedtls/ua_securitypolicy_aes128sha256rsaoaep.c
```diff
@@ -642,6 +642,144 @@ updateCertificateAndPrivateKey_sp_aes128sha256rsaoaep(UA_SecurityPolicy *securit
...
+            mbedtls_x509_crt_free(&x509Cert);
+            mbedtls_x509write_csr_free(&request);
+            UA_free(subj);
+            return UA_STATUSCODE_BADINTERNALERROR;
+        }
+    }
+
+    /* Get the subject alternate names from certificate and set them in CSR context*/
+    const mbedtls_x509_sequence *san_list = &x509Cert.subject_alt_names;
+    mbedtls_x509write_csr_set_subject_alt_name(&request, san_list);
```
mcpat, line 753: Copying the subject alternative names here is a good thing so far. The question is how did they initially came into the certificate? And how to handle the case, where the subject name shall remain unchanged but any subject alternative name (e.g. IP or DNS) must be changed?

Usually, when setting up a new device some kind of self-signed certificate comes into play. After setting the device up and configuring it properly, it is now the time to create a "final" certificate. This usually involves placing IP address and/or hostname of the device into the certificate. How can this be done when creating the certificate signing request?

plugins/crypto/mbedtls/ua_securitypolicy_aes256sha256rsapss.c
```diff
@@ -734,6 +735,144 @@ updateCertificateAndPrivateKey_sp_aes256sha256rsapss(UA_SecurityPolicy *security
     return retval;
 }
 
+static UA_StatusCode
```
mcpat, line 738: Is this amount of code duplication really necessary? Some common parts for mbedTLS are already factored out into the common source file. So why not this too?

plugins/crypto/mbedtls/ua_securitypolicy_aes128sha256rsaoaep.c
mcpat: [The specification](https://reference.opcfoundation.org/Core/Part6/v105/docs/6.2.2) is a little vague about this topic:

> [Servers](https://reference.opcfoundation.org/search/359?t=Servers) shall specify a partial or a fully qualified dNSName [or a ](https://reference.opcfoundation.org/search/359?t=or%20a)static IPAddress which identifies the machine where the application [Instance](https://reference.opcfoundation.org/search/359?t=Instance) runs.

include/open62541/plugin/securitypolicy.h
```diff
@@ -305,6 +305,9 @@ struct UA_SecurityPolicy {
      * depends on the used key length. */
     UA_ByteString localCertificate;
 
+    UA_NodeId certificateGroupId;
+    UA_NodeId certificateTypeId;
```
mcpat, line 309: The member field certificateGroupId is not required here. Instead, it should be placed as parameter in any function that requires it (e.g. into createSigningRequest). This avoids having to duplicate Security Policies for each Certificate Group where it shall be used and it also avoids having to filter for Security Policies (for the DefaultApplicationGroup) when creating Endpoints. The other reasons, why I think this field is of no use, are the following:

- Security Policies are required for establishing Secure Channels and thus in the Endpoints. This use case implictly requires to use the DefaultApplicationGroup -> no need of the field here.
- Security Policies are required for validation of certificate-based user tokens. This use case implicitly requires to use the DefaultUserTokenGroup -> no need of the field here.
- Security Policies are required to validate certificates to be placed into the trust lists. The API for this always provides the target certificateGroupId which can be forwarded.
- Security Policies are required to create self-signed certificates or certificate signing request. The API for this also provides the target certificateGroupId which can be forwared.

Do I miss something?

BTW, a similiar thing is the localCertificate field, which might be the reason why the certificateGroupId field found its way here. The local certificate is not required, because it can always be read (at least now) from a CertificateGroup. And in the Endpoint case it is already a member of UA_EndpointDescription...

include/open62541/plugin/securitypolicy.h
```diff
@@ -319,6 +322,14 @@ struct UA_SecurityPolicy {
                                                     const UA_ByteString newCertificate,
                                                     const UA_ByteString newPrivateKey);
 
+    /* Creates a PKCS #10 DER encoded certificate request signed with the server's
+     * private key. */
+    UA_StatusCode (*createSigningRequest)(UA_SecurityPolicy *securityPolicy,
+                                          const UA_String *subjectName,
```
jpfr, line 328: As long as this API is not widely used, I have a few wishes.

- Const input strings as a non-pointer value.
- Add a `UA_KeyValueMap` for additional parameters (as a pointer that can be NULL).
  That way we can add future parameters via named entries of the map without changing the API.
  The same trick is used in the EventLoop.

Happy to discuss.

include/open62541/server.h
```diff
@@ -1900,6 +1900,18 @@ UA_Server_updateCertificate(UA_Server *server,
                             UA_Boolean closeSessions,
                             UA_Boolean closeSecureChannels);
 
+/**
+ * Creates a PKCS #10 DER encoded certificate request signed with the server's private key
+ * ---------------------------------------------------------------------------------------- */
+UA_StatusCode UA_EXPORT
+UA_Server_createSigningRequest(UA_Server *server,
+                               const UA_NodeId *certificateGroupId,
```
jpfr, line 1908: Input strings and nodeids as non-pointer const values.

plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
```diff
@@ -289,6 +300,252 @@ mbedtls_decrypt_rsaOaep(mbedtls_pk_context *localPrivateKey,
...
+    pc = buf + buflen;
+
+    while(cur) {
+        switch (cur->buf.tag & 0x0F) {
+            case MBEDTLS_X509_SAN_DNS_NAME:
+            case MBEDTLS_X509_SAN_RFC822_NAME:
+            case MBEDTLS_X509_SAN_UNIFORM_RESOURCE_IDENTIFIER:
+            case MBEDTLS_X509_SAN_IP_ADDRESS:
+                MBEDTLS_ASN1_CHK_CLEANUP_ADD(len,
+                                             mbedtls_asn1_write_raw_buffer(&pc, buf, (const unsigned char *)cur->buf.p,
```
jpfr, line 343: Purely cosmetic:
Put mbedtls_asn1_write_raw_buffer on its own line (e.g. with an additiona "result variable".)

plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
```diff
@@ -289,6 +300,252 @@ mbedtls_decrypt_rsaOaep(mbedtls_pk_context *localPrivateKey,
...
+                                                     MBEDTLS_X509_KU_DATA_ENCIPHERMENT |
+                                                     MBEDTLS_X509_KU_NON_REPUDIATION |
+                                                     MBEDTLS_X509_KU_KEY_ENCIPHERMENT) != 0) {
+        retval = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    /* Add entropy */
+    if(entropyContext && nonce && nonce->length > 0) {
+        if(mbedtls_entropy_update_manual(entropyContext,
```
jpfr, line 456: Assign an additional result variable that is then checked in the if-condition.
Shorter lines, easier to read.

plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
```diff
@@ -289,6 +300,252 @@ mbedtls_decrypt_rsaOaep(mbedtls_pk_context *localPrivateKey,
...
+    }
+
+    /* Set the subject in CSR context */
+    if(mbedtls_x509write_csr_set_subject_name(&request, subj) != 0) {
+        retval = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    /* Get the subject alternate names from certificate and set them in CSR context*/
+    const mbedtls_x509_sequence *san_list = &x509Cert.subject_alt_names;
```
jpfr, line 503: Some compilers complain when you `goto` over a variable declaration.
I would declare the san_list variable at the top / before the first goto.

plugins/crypto/mbedtls/securitypolicy_mbedtls_common.c
```diff
@@ -289,6 +300,252 @@ mbedtls_decrypt_rsaOaep(mbedtls_pk_context *localPrivateKey,
...
+        mbedtls_rsa_gen_key(mbedtls_pk_rsa(*csrLocalPrivateKey), mbedtls_ctr_drbg_random,
+                            drbgContext, (unsigned int)keySize, 65537);
+        mbedtls_x509write_csr_set_key(&request, csrLocalPrivateKey);
+        mbedtls_writePrivateKeyDer(csrLocalPrivateKey, newPrivateKey);
+    } else {
+        mbedtls_x509write_csr_set_key(&request, localPrivateKey);
+    }
+
+
+    unsigned char requestBuf[4096];
```
jpfr, line 522: Magic number 4096.
Can this become a problem?
You can make this a #define at the top of the file.

jpfr (commented):
Looks good!
Comments inline. Please also answer to the comments of @mcpat and resolve where possible.

The test coverage could be a bit more.
Failure cases should be easy to add with copy-pasta from what you have.

include/open62541/plugin/securitypolicy.h
NoelGraf: I added the KeyValueMap parameter.
SubjectName does not need to be passed and is otherwise extracted from the existing certificate. So I would leave it as a pointer.

jpfr:
Nice. Merged.

Issue #6520, opened 2024-05-31
Can no longer write UA_Int32 to an enumeration node


rickboks:
Hi! I have been on `open62541` version `1.3.6` for about a year, and tried to switch to version `1.4.1` today. In my application, I dynamically define enumeration types as outlined in https://github.com/open62541/open62541/issues/2032#issuecomment-474752649 . This always worked perfectly in `v1.3.6`. The type definition also works fine in `v1.4.1`, but when I try to create a node with the created enum type using `UA_Server_addVariableNode`, I get an error `BadTypeMismatch` and the following log message:
```
[2024-05-31 17:28:23.103 (UTC+0200)] warning/server	Writing the value of Node ns=2;s=myEnumInstance failed with the following reason: DataType of the value is incompatible
```
I found out that in commit 0d0b18b08bd27d5e7da17dd40bc115b541bb7993, a special case for accepting `UA_Int32` writes to Enumerations was removed from the function `compatibleDataTypes`. This is the cause of my problem; if I add the check to `open62541` version `v1.4.1`, my enums work again. Why was this special case removed? Can it be re-introduced, or do I need to instantiate enumeration variables differently?

The relevant check that was removed from `compatibleDataTypes` is:

```
  /* The constraint is an enum -> allow writing Int32 */
  if(UA_NodeId_equal(dataType, &UA_TYPES[UA_TYPES_INT32].typeId) &&
     isNodeInTree_singleRef(server, constraintDataType, &enumNodeId,
                            UA_REFERENCETYPEINDEX_HASSUBTYPE))
      return true;
```
 Thanks in advance!

jpfr:
@andreasebner Please check whether your commit can be reverted.
In any case this needs documentation and better log messages.

jpfr:
According to the unit tests this is working:
https://github.com/open62541/open62541/pull/6532/commits/e87db6c07116595dfdbf1eddecd8ec11cb4dba12

If you please extend this unit test to reproduce the problem.
Then we can fix it easily.

rickboks:
Thanks @jpfr. I think the difference is that the test operates on a predefined enum `UA_TYPES_MESSAGESECURITYMODE`, while I would like it to work on dynamically created 'custom' enums. I created a pull request #6536 to include such a custom enum in the tests. Currently the test fails. I checked that the test succeeds again if I re-introduce the check which was removed in 0d0b18b

jpfr:
There should be no difference between built in and custom types. But ok. Will check.

[...]

PR #6536, opened 2024-06-05
feat(tests): Extend tests for writing int32 into custom enum type

rickboks:
Extended tests to include creating a custom enumeration type and writing a int32 to it. As it stands, this test fails. See issue #6520

### DIFF
--- a/tests/server/check_services_attributes.c
+++ b/tests/server/check_services_attributes.c
@@ -74,6 +74,66 @@ static void setup(void) {
                                        vattr, NULL, NULL);
     ck_assert_int_eq(retval, UA_STATUSCODE_GOOD);
 
+    /* Custom enum VariableNode */
+    UA_DataTypeAttributes dattr = UA_DataTypeAttributes_default;
+    dattr.description = UA_LOCALIZEDTEXT("locale", "custom enum type description");
+    dattr.displayName = UA_LOCALIZEDTEXT("locale", "custom enum type display name");
+
+    size_t customEnumSize = 3;
+    UA_NodeId customEnumTypeNodeId;
+    retval = UA_Server_addDataTypeNode(server, UA_NODEID_NULL,
+                                       UA_NODEID_NUMERIC(0, UA_NS0ID_ENUMERATION),
+                                       UA_NODEID_NUMERIC(0, UA_NS0ID_HASSUBTYPE),
+                                       UA_QUALIFIEDNAME (1, "custom enum type"),
+                                       dattr, NULL, &customEnumTypeNodeId
+    );
+    ck_assert_int_eq(retval, UA_STATUSCODE_GOOD);
+
+    vattr = UA_VariableAttributes_default;
+    vattr.description = UA_LOCALIZEDTEXT("locale", "EnumValues");
+    vattr.displayName = UA_LOCALIZEDTEXT("locale", "EnumValues");
+    vattr.dataType = UA_TYPES[UA_TYPES_ENUMVALUETYPE].typeId;
+    vattr.valueRank = UA_VALUERANK_ONE_DIMENSION;
+    vattr.arrayDimensionsSize = 1;
+    UA_UInt32 customEnumValuesDimensions[1] = {customEnumSize};
+    vattr.arrayDimensions = customEnumValuesDimensions;
+
+    UA_EnumValueType* customEnumValues = (UA_EnumValueType*) UA_Array_new(customEnumSize, &UA_TYPES[UA_TYPES_ENUMVALUETYPE]);
+    char* customEnumDisplayNames[] = {"enum value one", "enum value two", "enum value three"};
+    for (size_t i = 0; i < customEnumSize; i++){
+        customEnumValues[i].description = UA_LOCALIZEDTEXT_ALLOC("locale", customEnumDisplayNames[i]);
+        customEnumValues[i].displayName = UA_LOCALIZEDTEXT_ALLOC("locale", customEnumDisplayNames[i]);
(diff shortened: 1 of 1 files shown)


jpfr:
I see the definition of the DataTypeNode.
But you don't add the custom UA_DataType description to the server config.
Historically the `UA_DataType` structure is used to describe datatypes for de/encoding.
And we look that up in the code path we are discussing.

Here is the example for adding custom types.

https://github.com/open62541/open62541/blob/46c5caae45e5523ac3021a33e74d3902f04d32f4/examples/custom_datatype/server_types_custom.c#L252

rickboks:
Many thanks! I got it working. One thing that is not quite clear to me yet is the `UA_DataType.binaryEncodingId`. The examples in `server_types_custom.c` have different values for `.typeId` and `.binaryEncodingID`; the `namespaceIndex` and `identifierType` are identical between the two, but the `identifier` is chosen differently (see below snippet from `custom_datatype.h`). What is the reason behind this? Can't I simply use the same `.typeId` and `.binaryEncodingId`? Also, do the `indentifierType`s need to match between the two?

```
static const UA_DataType PointType = {
        UA_TYPENAME("Point")                /* .tyspeName */
        {1, UA_NODEIDTYPE_NUMERIC, {4242}}, /* .typeId */
        {1, UA_NODEIDTYPE_NUMERIC, {Point_binary_encoding_id}}, /* .binaryEncodingId, the numeric
                                            identifier used on the wire (the
                                            namespaceindex is from .typeId) */
        sizeof(Point),                      /* .memSize */
        UA_DATATYPEKIND_STRUCTURE,          /* .typeKind */
        true,                               /* .pointerFree */
        false,                              /* .overlayable (depends on endianness and
                                            the absence of padding) */
        3,                                  /* .membersSize */
        Point_members
};
```

jpfr:
The type is being different is one of the warts of the standard. Maybe at some time we can use the typeid everywhere. But not now. And also all built in types in ns0 have a different encoding id for binary/json/xml.

rickboks:
I see. But I *can* use the same type/binary encoding ID for my custom types, right?

jpfr:
It will probably work alright.
Not sure if some other OPC UA product will complain.

rickboks:
Great, thanks again for your help.

maikrou:
Hi @rickboks, would it be possible to share the code you are using to get the custom enumerations working?

I had the exact same issue when using UA_Server_addDataSourceVariableNode, even after adding the data type to the configuration, I am probably adding it incorrectly or compiling with wrong settings? I tried with UA_NAMESPACE_ZERO "FULL" but no luck there either.

For me I got things working (not perfectly) by adding that enum to int32 check in compatibleDataTypes() at ua_services.attribute.c

Another question: does your enumerations work with all clients? I got mine working with UaExpert but not with OpcUaBrowser. Reason for the enumerations don't show up is because the data type is not visible in the OPC Binary DataTypeDictionary nor is the DataTypeDefinition attribute supported. OPC-10000-3 states that either one is mandatory for enumerations.

I am not sure if that DataTypeDictionary should get updated if the type is added properly to the stack but currently I can see that its content is compiled into: 

static const UA_Byte variablenode_ns_0_i_7617_variant_DataContents_byteArray[182253] = {...}

I can't see that it gets updated at any point, and the stack does not support the "DataTypeDefinition" attribute so I am not sure how that would work according to specification.

rickboks:
Hi @maikrou. I added the following code **before** the call to `UA_Server_addDataTypeNode`. If you add it afterwards, UAExpert will show created variables with this enum type as plain `Int32`. I have only tested with `UAExpert` so far. Is [this](https://prosysopc.com/products/opc-ua-browser/) the OpcUaBrowser you refer to?

Regarding compile options, I only use `UA_ENABLE_ENCRYPTION=OPENSSL`

```
// I added the below block of code (needs to be BEFORE the call to UA_Server_addDataTypeNode)
UA_NodeId enumTypeNodeId = UA_NODEID_STRING_ALLOC(namespaceIndex, name);
UA_DataType *newEnumDataType = (UA_DataType*)(UA_malloc(1 * sizeof(UA_DataType)));
newEnumDataType[0] = UA_TYPES[UA_TYPES_ENUMERATION]; 
newEnumDataType->typeId = enumTypeNodeId;
newEnumDataType->binaryEncodingId = enumTypeNodeId;
newEnumDataType->typeName = strdup(name);
UA_ServerConfig* serverConfig = UA_Server_getConfig(server);
UA_DataTypeArray tempCustomDataTypes {serverConfig->customDataTypes, 1, newEnumDataType, UA_TRUE};
UA_DataTypeArray* newCustomDataTypes = (UA_DataTypeArray*) UA_malloc(sizeof(UA_DataTypeArray));
memcpy(newCustomDataTypes, &tempCustomDataTypes, sizeof(UA_DataTypeArray));
serverConfig->customDataTypes = newCustomDataTypes;

// This was already here
UA_StatusCode statusCode = UA_Server_addDataTypeNode(server, enumTypeNodeId, 
  UA_NODEID_NUMERIC(0, UA_NS0ID_ENUMERATION),
  UA_NODEID_NUMERIC(0, UA_NS0ID_HASSUBTYPE),
  UA_QUALIFIEDNAME (namespaceIndex, name),
  dattr, nullptr, nullptr));

// ... define EnumValues ...
```

maikrou:
@rickboks Yes, Prosys OPC UA Browser. I managed to get the enumerations to show up like this: https://postimg.cc/gallery/d088gy9

I think I have done exactly like you but just added the data type to the config after calling UA_Server_addDataTypeNode, need to give it another try.

PR #6538, opened 2024-06-06
delete .cmake flag from .gitignore.

avcuenes:
If anyone want to update tools/cmake files , we can allow to update and push it.


jpfr:
👍

PR #6540, opened 2024-06-07
[REVIEW] Add ModelChangeEventType and ConditionType to reduced namespace

NoelGraf:
Add ModelChangeEventType and ConditionType to support subscriptions with reduced namespace


jpfr:
👍

Issue #5783, opened 2023-05-14
Cliend hangs if server rejects client's sertificate and sends error to client


Alexk12:
## Description

Conditions:
1. Client connects with security.
2. Server rejects client's certificate and sends error.

If error received on early stage when channel not opened yet (for example, if server rejects client's certificate and reports error), setting state to UA_SECURECHANNELSTATE_CLOSING prevents exit from event loop (client hangs).

Suggested fix:

    void
    processERRResponse(UA_Client *client, const UA_ByteString *chunk) {
        //client->channel.state = UA_SECURECHANNELSTATE_CLOSING; // FIX: commented because this state is set below on call to closeSecureChannel

        size_t offset = 0;
        UA_TcpErrorMessage errMessage;
        UA_StatusCode res =
            UA_decodeBinaryInternal(chunk, &offset, &errMessage,
                                    &UA_TRANSPORT[UA_TRANSPORT_TCPERRORMESSAGE], NULL);
        if(res != UA_STATUSCODE_GOOD) {
            UA_LOG_ERROR_CHANNEL(&client->config.logger, &client->channel,
                                 "Received an ERR response that could not be decoded "
                                 "with StatusCode %s", UA_StatusCode_name(res));
            client->connectStatus = res;
            closeSecureChannel(client);
            return;
        }
    
        UA_LOG_ERROR_CHANNEL(&client->config.logger, &client->channel,
                             "Received an ERR response with StatusCode %s and the following "
                             "reason: %.*s", UA_StatusCode_name(errMessage.error),
                             (int)errMessage.reason.length, errMessage.reason.data);
        client->connectStatus = errMessage.error;
        closeSecureChannel(client);
        UA_TcpErrorMessage_clear(&errMessage);
    }

## Background Information / Reproduction Steps

Used CMake options:
```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): Master
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

peschuster:
We also encountered this bug.

[...]

PR #6541, opened 2024-06-12
fix(client): channel state set to CLOSING without closing underlying connection

peschuster:
Thanks to @Alexk12 for the analysis and suggested fix (see issue #5783).

If the connection is denied by the OPC UA server with e.g. `BadSecurityChecksFailed` the subsequent call to `UA_Client_delete` / `UA_Client_disconnect` hangs, because the exit condition for the while loop in `disconnectSecureChannel` is never reached: 

https://github.com/open62541/open62541/blob/3888af3410b53be0c5fa15b9d762dbab5b84902e/src/client/ua_client_connect.c#L2212

It was probably caused by the added "recursion check" in closeSecureChannel (commit 4bd6f4023ab59fa8ba1f17f5d82dd89636fc9995) that was added in v1.4.0-rc1

`UA_SecureChannel_shutdown` (correctly) sets `UA_SECURECHANNELSTATE_CLOSING` when actually closing the TCP connection, but is never called when the state is already set to `UA_SECURECHANNELSTATE_CLOSING` before.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -396,8 +396,6 @@ checkCreateSessionSignature(UA_Client *client, const UA_SecureChannel *channel,
 
 void
 processERRResponse(UA_Client *client, const UA_ByteString *chunk) {
-    client->channel.state = UA_SECURECHANNELSTATE_CLOSING;
-
     size_t offset = 0;
     UA_TcpErrorMessage errMessage;
     client->connectStatus =
--- a/tests/client/check_client.c
+++ b/tests/client/check_client.c
@@ -440,6 +440,28 @@ START_TEST(Client_activateSessionLocaleIds) {
 }
 END_TEST
 
+START_TEST(Client_closes_on_server_error) {
+    UA_Client *client = UA_Client_new();
+    UA_ClientConfig_setDefault(UA_Client_getConfig(client));
+    UA_StatusCode retval = UA_Client_connect(client, "opc.tcp://localhost:4840");
+
+    ck_assert_uint_eq(retval, UA_STATUSCODE_GOOD);
+
+    ck_assert_uint_eq(server->sessionCount, 1);
+    UA_SecureChannel *channel = server->sessions.lh_first->session.header.channel;
+
+    // send error message from server to client
+    UA_TcpErrorMessage errMsg = {.error = UA_STATUSCODE_BADSECURITYCHECKSFAILED,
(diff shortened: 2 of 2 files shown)


jpfr:
Nice!
Please add a unit test also to prevent future regressions.

The client unit tests are here: https://github.com/open62541/open62541/blob/master/tests/client/check_client.c
You can use `UA_SecureChannel_sendError` to send the ERR message from the (only) SecureChannel the server has.

peschuster:
> Nice! Please add a unit test also to prevent future regressions.
> 
> The client unit tests are here: https://github.com/open62541/open62541/blob/master/tests/client/check_client.c You can use `UA_SecureChannel_sendError` to send the ERR message from the (only) SecureChannel the server has.

I tried, but was not able to do so, yet:

I could not find a way to get a handle for the secure channels of the server in the unit tests. References to the secure channels seem to be only available through `UA_BinaryProtocolManager`. I can get a reference to this using `getServerComponentByName`. But the definition of the struct itself is only available in `ua_server_binary.c`.

I also tried to create a test that mimics the behavior (server denying client certificate with `BadSecurityChecksFailed`).  But this did not work, because the open62541 server implementation closes the tcp connection itself and therefore the client does not hang even when it does not initiate closing the connection.

So the issue seems to arise only, when the connection is not already closed by the server (might be a bug in the server implementation).

Any recommendations how to proceed? Do you see a way to access the server-side secure channel in a client unit test?

jpfr:
Right. We might better have global SecureChannel list in the server…

For the time you can open a single session.
And then retrieve the securechannel pointer from the session.
https://github.com/open62541/open62541/blob/3d0da9ad73b49058e197913352bb0eb30235c14b/src/server/ua_session.h#L41

peschuster:
Thanks for the hint. I missed that because in 1.4 it is nested inside the session header, which I did not expect.

I added the test and checked that it fails before the fix and passes with the fix.

If the test fails, the test execution currently hangs up. libcheck has a `tcase_set_timeout` feature for this, but it has no effect, when tests are run with `CK_NOFORK`, which seems to be set for all open62541 unit tests. \
I also tested to run the tests with activated forking, but libcheck only supports this for unix systems and I only have a development environment setup on Windows, at the moment.

As this scenario is kind of a corner case and the related code is probably not touched much, it might be ok to accept this behavior for now. That way, at least it won't go unnoticed if it breaks again.

jpfr:
Nice. 👍 
Merged.

Issue #6462, opened 2024-05-08
Sender certificate shall be ignored in CreateSessionRequest for channel with SecurityPolicy None


phuong-nm:
## Description

In the "OPC UA Specification Part 4 V1.04, Section 5.6.2.2" it is stated that "If the securityPolicyUri is None, the Server shall ignore the
ApplicationInstanceCertificate".

## Background Information / Reproduction Steps

- OPC Compliance Test Tool V1.04.506, test case 012 of Session Services / Session Base conformance unit. In this test a secure channel with SecurityPolicy None is created, following by a CreateSessionRequest with an invalid sender certificate which does not match ApplicationURI of the sender. According to the specification, verification of the sender certificate shall be ignored by the server.
- See also: https://opcfoundation.org/forum/opc-ua-standard/shall-server-reject-a-createsessionrequest-with-badly-formatted-applicationinstancecertificate-for-channel-with-securitypolicynone/#p5146

## Suggested solution

```c
/*
* File: src/server/ua_services_session.c
*/

void
Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
                      const UA_CreateSessionRequest *request,
                      UA_CreateSessionResponse *response) {

    /* ... */

    if(!UA_ByteString_equal(&channel->securityPolicy->policyUri,
                            &UA_SECURITY_POLICY_NONE_URI) &&
       request->clientNonce.length < 32) {
        response->responseHeader.serviceResult = UA_STATUSCODE_BADNONCEINVALID;
        return;
    }

    /*** Hilscher ***/
    /* "If the securityPolicyUri is None, the Server shall ignore the ApplicationInstanceCertificate".
     * See: "OPC UA Specification Part 4 V1.04, Section 5.6.2.2". */
    if(channel->securityMode == UA_MESSAGESECURITYMODE_SIGN ||
       channel->securityMode == UA_MESSAGESECURITYMODE_SIGNANDENCRYPT) {
        if(request->clientCertificate.length > 0) {
            UA_CertificateVerification *cv = &server->config.certificateVerification;
            response->responseHeader.serviceResult =
                cv->verifyApplicationURI(cv->context, &request->clientCertificate,
                                        &request->clientDescription.applicationUri);
            if(response->responseHeader.serviceResult != UA_STATUSCODE_GOOD) {
                UA_LOG_WARNING_CHANNEL(&server->config.logger, channel,
                                    "The client's ApplicationURI did not match the certificate");
[...]

PR #6547, opened 2024-06-17
Fixed verification of sender certificate in CreateSessionRequest

phuong-nm:
Fix for issue #6462.

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -289,17 +289,22 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
         return;
     }
 
-    if(request->clientCertificate.length > 0) {
-        UA_CertificateVerification *cv = &server->config.certificateVerification;
-        response->responseHeader.serviceResult =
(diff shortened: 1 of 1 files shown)


jpfr:
Not sure about this.
If the client uses a certificate, then we should also validate it.
Please explain the scenario where this is needed.

We could make this a UA_RuleHandling configuration option…

phuong-nm:
@jpfr this problem is already described in details in the issue #6462. It is also confirmed in OPC UA forum that the sender certificate in CreateSessionRequest **shall** be ignored if the securityPolicyUri is None. Hence, the server shall not perform any verification of client certificate parameter of the request.

jpfr:
Then my wish is as follows:

1) If encryption is disabled and a certificate is present, then we emit a warning and run `UA_ByteString_clear(&request->clientCertificate)`.
2) We consolidate the certificate validation in one place.
   Elsewhere there surely is some logic to check that a certificate must be present for CreateSession if encryption is enabled.
   Right now the code you changed only checked the certificate if it does exist...

PR #6549, opened 2024-06-17
fix(plugins): mdnsIpAddressList only if UA_HAS_GETIFADDR is undefined

laroche:
In plugins/ua_config_json.c mdnsIpAddressList can only be set if UA_HAS_GETIFADDR is undefined.


jpfr:
Thanks for this.
Please add a comment before the #if to explain the situation why it is not needed with getifaddr activated.

laroche:
Hello Julius Pfrommer,

does this update look good?

best regards,

Florian La Roche

jpfr:
Perfect. Merged.

PR #6550, opened 2024-06-17
fix(core): ret is already defined, so reuse var

laroche:
If src/ua_types_encoding_json.c is compiled with "gcc -Wshadow", a warning is output for an already defined var. This change fixes this.


jpfr:
👍

PR #6552, opened 2024-06-18
fix(core): fix default network values in UA_ConnectionConfig_default

laroche:
Some items use 2<<y instead of 1<<y==2^y.
I've kept the buffer size in ua_pki_mbedtls.c the same, though this could also get reduced.


jpfr:
Cool.
Adjusted the commit title.
This is a good improvement but not a (critical) fix.

PR #6553, opened 2024-06-18
fix(core): increase maxSecureChannels to 100, fix check for enough ch…

laroche:
…annels

Increase maxSecureChannels to 100 and change the check for enough SecureChannels to not give a warning if maxSessions==maxSecureChannels.


jpfr:
Cool.
We will verify against the foundations CTT tools…

PR #6561, opened 2024-06-26
[REVIEW] Move the list of secure channels into the server structure

NoelGraf:


src/server/ua_server_internal.h
```diff
@@ -59,6 +59,11 @@ typedef struct {
 
 #endif /* !UA_ENABLE_SUBSCRIPTIONS */
 
+typedef struct channel_entry {
+    UA_SecureChannel channel;
+    TAILQ_ENTRY(channel_entry) pointers;
```
jpfr, line 64: You need a second entry here for the second tailq.

PR #6564, opened 2024-06-27
[REVIEW] Adding filestore backend security policies

NoelGraf:
- Adding filestore backend security policies to support writing updated certificates back to the file system.
- Make 'storePath' required for the filestore backend.
- Add test cases


plugins/crypto/ua_certificategroup_filestore.c
```diff
@@ -439,18 +439,10 @@ FileCertStore_createPkiDirectory(UA_CertificateGroup *certGroup, const UA_String
...
-    }
-    else {
-        if(getcwd(rootDirectory, PATH_MAX) == NULL) {
-            return UA_STATUSCODE_BADINTERNALERROR;
-        }
-    }
+    if(directory.length <= 0)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    memcpy(rootDirectory, directory.data, directory.length);
```
mcpat, line 445: Potential buffer overflow here. You should ensure whether the length of the given directory does not exceed PATH_MAX.

plugins/crypto/ua_securitypolicy_filestore.c
```diff
@@ -0,0 +1,411 @@
...
+
+#ifdef __linux__ /* Linux only so far */
+
+#include <stdio.h>
+#include <dirent.h>
+#include <bits/stdio_lim.h>
+
+typedef struct FileCertStore {
+    /* Memory cert store as a base */
+    UA_SecurityPolicy policy;
```
mcpat, line 26: Would it be sufficient, to just store a pointer to the surrounding security policy here?

plugins/crypto/ua_securitypolicy_filestore.c
```diff
@@ -0,0 +1,411 @@
...
+    policy->certificateGroupId = pc->policy.certificateGroupId;
+    policy->certificateTypeId = pc->policy.certificateTypeId;
+    policy->asymmetricModule = pc->policy.asymmetricModule;
+    policy->symmetricModule = pc->policy.symmetricModule;
+    policy->certificateSigningAlgorithm = pc->policy.certificateSigningAlgorithm;
+    policy->channelModule = pc->policy.channelModule;
+    policy->logger = pc->policy.logger;
+    policy->createSigningRequest = pc->policy.createSigningRequest;
+    policy->updateCertificateAndPrivateKey = updateCertificateAndPrivateKey_sp_filestore;
+    policy->clear = clear_sp_filestore;
```
mcpat, line 271: This looks like a complete clone of the policy within the context. Would it be better to just have the context pointing to the policy (instead of owning a duplicate) to avoid doing this copying stuff?

plugins/include/open62541/server_config_default.h
```diff
@@ -243,6 +239,111 @@ UA_ServerConfig_addAllSecureSecurityPolicies(UA_ServerConfig *config,
                                        const UA_ByteString *certificate,
                                        const UA_ByteString *privateKey);
 
+#ifdef __linux__ /* Linux only so far */
+
+/* Adds the security policy ``SecurityPolicy#Basic128Rsa15`` to the server. A
+ * server certificate may be supplied but is optional.
```
mcpat, line 245: Well, that's really cool! The server certificate being optional on security policy construction time is definitely a feature!
Nevertheless, the private key seems to be still required here, is that right? At least a look into the "UA_mbedTLS_LoadPrivateKey" function does not state otherwise and it should result in an error if an empty private key is provided.

plugins/crypto/ua_securitypolicy_filestore.c
```diff
@@ -0,0 +1,411 @@
...
+
+    UA_String_copy(&storePath, &pc->storePath);
+    policy->policyContext = (void *)pc;
+
+    retval = UA_SecurityPolicy_Basic256Sha256(&pc->policy, localCertificate,
+                                              localPrivateKey, logger);
+    if(retval != UA_STATUSCODE_GOOD)
+        return retval;
+
+    init_sp_filestore(policy);
```
AdanRuoff, line 350: I think I get the idea behind this approach: You basically create a second SecurityPolicy inside the SecurityPolicy_FilestoreContext, initialize this "inside" policy with UA_SecurityPolicy_Basic256Sha256 and then copy the functions pointers from the "inside" policy to the "outside" policy, replacing those functions that need to do additional work (storing an updated Application Instance Certificate in the certificate store).

However, this does not work (at least for me), as the functions behind the copied function pointers (e.g. "asymmetricModule_compareCertificateThumbprint_sp_basic256sha256") expect the policyContext to be of type "Basic256Sha256_PolicyContext", but get a "SecurityPolicy_FilestoreContext" instead.

To make this work, you will IMHO either have to provide "trampoline" functions for _each and every_ function registered in the SecurityPolicy that accesses the policyContext, which will then call the actual function with the "inside" policy (as is done by "updateCertificateAndPrivateKey_sp_filestore"), or add something like an additional "storeContext" in the UA_SecurityPolicy structure. Seeing that the "transactions" feature waiting to be included does not use the additional "inside" SecurityPolicy for storing the new certificate and private key, this would allow to get rid of the SecurityPolicy structure duplications, as far as I see.

plugins/include/open62541/server_config_default.h
NoelGraf: That was a copying mistake. A certificate and a private key are required to create a security policy.

plugins/crypto/ua_securitypolicy_filestore.c
```diff
@@ -0,0 +1,508 @@
...
+
+    UA_SecurityPolicyAsymmetricModule *const asymmetricModule = &policy->asymmetricModule;
+    memcpy(asymmetricModule, &pc->policy->asymmetricModule, sizeof(UA_SecurityPolicyAsymmetricModule));
+    asymmetricModule->makeCertificateThumbprint = asym_makeThumbprint_sp_filestore;
+    asymmetricModule->compareCertificateThumbprint = asym_compareCertificateThumbprint_sp_filestore;
+
+    UA_SecurityPolicySymmetricModule *const symmetricModule = &policy->symmetricModule;
+    memcpy(symmetricModule, &pc->policy->symmetricModule, sizeof(UA_SecurityPolicySymmetricModule));
+    symmetricModule->generateKey = sym_generateKey_sp_filestore;
+    symmetricModule->generateNonce = sym_generateNonce_sp_filestore;
```
AdanRuoff, line 341: No offense meant, but this approach of using "trampoline" functions may end up as a maintainer's nightmare. If any crypto-plugin backend function apart from these two decides to use the "securityPolicy->context" in future, you will have to provide an additional trampoline function as well, or we will again run into the same "assume wrong context type" issue. This is hard to track as the implementer of crypto-plugin functions may not be aware of these "wrap-around" SecurityPolicies, and additionally it leads to code bloat since we need to implement trampoline functions for _all_ SecurityPolicies that will support the Certificate File Store, and additionally provide a trampoline function _for each function_ encapsulated inside a SecurityPolicy that uses the securityPolicy->context.
Sorry for ranting - just my two cents...

jpfr: This is now very generic.
The "file-based" security policy is a wrapper that can wrap around any other SecurityPolicy.
So we do not have to duplicate functionality.

jpfr:
👍

PR #6565, opened 2024-06-28
[REVIEW] Don't force UA_MESSAGESECURITYMODE_SIGNANDENCRYPT if encryption is disabled.

r-sin:
Added static check for enabled encryption.
The discovery example fails (server_register can't register with the LDS) if encryption is disabled and this check is missing. It corresponds to UA_Server_register, line 475.


jpfr:
Looks good. Please make this a PR against the 1.4 branch.

PR #6568, opened 2024-06-28
remove server capability "NA" from config file

laroche:
specs say that server capability "NA" should be listed as first capability and be the only one.


jpfr:
We allow comment syntax.
Please add comments to the JSON when to add which option.

laroche:
Hello Julius Pfrommer,

Ok, added some more verbose comments on usage and a pointer to more
in-depth information as well.
Looks good now?

best regards,

Florian La Roche

jpfr:
👍

PR #6569, opened 2024-06-28
adjust compiler optimization levels

laroche:
adjust compiler optimization levels:
- Use "-g -Og" for debug builds
- Use "-g -O2" otherwise.
- Enable CMAKE_INTERPROCEDURAL_OPTIMIZATION only for non-debug builds. (This speeds up compile time for debug builds.)


laroche:
Some tests fail due to missing -fPIC now. I wonder if enabling "-fPIC" for all UNIX builds would
solve this. A bit of a gross hack, not sure there is a better fix for this.

laroche:
In my opinion this should now be ready to be merged. Not sure this one failing test is
timing dependent or dependent on compiler optimization level. Adding proper/different
compiler flags seems a good direction.

I am pushing one more patch here that enables -fPIC only if UA_ENABLE_AMALGAMATION is set.

jpfr:
What does the amalgamation has to do with fPIC?
It shouldn't be important whether the sources are combined into a single file...

laroche:
Hello Julius Pfrommer,

correct, instead of the fPIC workaround there should at some point be a real fix for this.
I've already added a comment for this into CMakeLists.txt.

I was astonished that -flto (LinkTimeOptimization) was used to compile the source, but
then no -Ox optimization flag was used at all. That was my original goal for this change
request. Maybe it would be also good to use a new issue to track this fPIC issue separately.
For UA_ENABLE_AMALGAMATION, I also don't have any experience with it until now.

best regards,

Florian La Roche

jpfr:
The -O2 comes from CMake already when the Release build is chosen.
You can check that by building with CMake in verbose mode.

Also we shouldn't add '-g' if the build is not in Debug mode.

Disabling inter-procedural optimization in Debug mode is good-to-go.

Concerning fPIC I am leaning towards not building the amalgamation at all.
That is, with the amalgamation activated we continue to generate the .c and .h files.
But we always build the library the old-fashioned way.

The only reason why we generate the amalgamation is to facilitate the inclusion of the source in other build systems without all the overhead from CMake and the build flags it brings...

ccvca:
There are more generic ways for ``-fPIC too.  https://stackoverflow.com/a/38297422

CMakeLists.txt
```diff
@@ -556,6 +556,11 @@ endfunction()
 if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
     check_add_cc_flag("-std=c99")   # C99 mode
     check_add_cc_flag("-pipe")      # Avoid writing temporary files (for compiler speed)
+    if(UA_DEBUG)
+        check_add_cc_flag("-g -Og")
+    else()
+        check_add_cc_flag("-g -O2")
```
mcpat, line 562: I also would not add "-g" to any optimising build. Please leave this to debug builds only.

laroche:
Hello Julius Pfrommer

your review sounds good for me. I'll update the changeset to only disable inter-procedural optimization in Debug mode.

best regards,

Florian La Roche

jpfr:
Nice.

Btw, the amalgamation now builds the .c/.h file but does not build with it...

PR #6581, opened 2024-07-04
refactor(examples): server_multicast.c: remove "NA" from server capabilities

laroche:
server_multicast.c: remove "NA" from server capabilities and only keep "LDS" as server capability.


laroche:
The comment above suggest that UAexpert does not list "LDS"-only servers, so maybe we have a reason to
add another server capability here. Not sure what the intenion is with this server, so please review...

best regards,

Florian La Roche

PR #6582, opened 2024-07-04
refactor(server): add local hostname if none given for LDS-ME.

laroche:
In src/server/ua_discovery_mdns.c if no local hostname is given and thus INADDR_ANY/0.0.0.0 should be used, use gethostname() as local hostname.
Until now an error is returned from UA_Discovery_addRecord().


jpfr:
👍

PR #6587, opened 2024-07-09
refactor(build): CMakeLists.txt: define UA_HAS_GETIFADDR on UNIX

laroche:
For building on UNIX, always define UA_HAS_GETIFADDR.

### DIFF
--- a/include/open62541/config.h.in
+++ b/include/open62541/config.h.in
@@ -145,11 +145,13 @@
 #  define _BSD_SOURCE
 # endif
 
-/* Define _GNU_SOURCE to get functions like ppoll. Comment this out to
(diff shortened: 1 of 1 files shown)


jpfr:
This will only work for cmake.
Not if the amalgamation is used outside of cmake.

Consider adding this to the posix header in /arch.

laroche:
Hello Julius Pfrommer,

I tried to find a suitable header in arch, but o header is included often enough. UA_HAS_GETIFADDR is
also needed e.g. in server.h.

What is possible and arch-dependent stuff is already included often there is config.h.in.

best regards

Florian La Roche

jpfr:
Looks good. 👍

PR #6588, opened 2024-07-09
refactor(examples,tests): avoid spaces for mDNS records

laroche:
Avoid spaces in mdnsServerName as this is used for mDNS records.


jpfr:
👍

PR #6589, opened 2024-07-09
feat(core): add UA_String_fromFormat

dpazj:


jpfr:
Use this: https://github.com/open62541/open62541/blob/master/deps/mp_printf.h

jpfr:
We want to have 100% identical results across all OS and libc. Including float encoding, etc.
So we have our own printf.
It is small enough..

src/ua_types.c
```diff
@@ -115,6 +117,26 @@ UA_cleanupDataTypeWithCustom(const UA_DataTypeArray *customTypes) {
...
 
+UA_String
+UA_String_fromFormat (char *format, ...)
+{
+    UA_String s;
+    s.data = NULL;
+    va_list args;
+    va_list args_copy;
+
+    va_start (args, format);
```
jpfr, line 128: I suggest you first encode into a 256 byte stack-allocated buffer first.
When that succeeds you can memcpy and don’t have to encode a second time.

include/open62541/types.h
```diff
@@ -200,6 +200,9 @@ UA_String_fromChars(const char *src) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
 UA_Boolean UA_EXPORT
 UA_String_isEmpty(const UA_String *s);
 
+UA_String UA_EXPORT
```
jpfr, line 203: Add a short comment to explain that printf format is used. Also explain that memory is allocated to match the formatted string size.

make a second version that takes a preallocated buffer as the first argument.

dpazj:
Is that along the lines of what you are looking for? I will add some tests in a bit.

include/open62541/types.h
```diff
@@ -200,6 +200,14 @@ UA_String_fromChars(const char *src) UA_FUNC_ATTR_WARN_UNUSED_RESULT;
 UA_Boolean UA_EXPORT
 UA_String_isEmpty(const UA_String *s);
 
+/* Allocates a formatted string on the heap */
+UA_String UA_EXPORT
+UA_String_fromFormat (char *format, ...);
+
+/* Creates a formatted string using memory from the provided buffer */
+UA_String UA_EXPORT
+UA_String_fromFormatWithBuffer (char *buffer, size_t size, char *format, ...);
```
jpfr, line 209: Use `UA_String` for the buffer.

src/ua_types.c
```diff
@@ -114,6 +115,43 @@ UA_cleanupDataTypeWithCustom(const UA_DataTypeArray *customTypes) {
...
+    return s;
+}
+
+UA_String
+UA_String_fromFormat (char *format, ...)
+{
+    char tmp[FORMAT_STACK_SIZE];
+    va_list args;
+    va_start (args, format);
+    UA_String s = vfromFormatWithBuffer (tmp, FORMAT_STACK_SIZE, format,args);
```
jpfr, line 148: Check whether the string length was enough.
If not, take the required size (returned from printf), malloc and print again.

Only do the memcpy if printing into the buffer succeeded.

src/ua_types.c
```diff
@@ -114,6 +115,48 @@ UA_cleanupDataTypeWithCustom(const UA_DataTypeArray *customTypes) {
...
+    va_list args;
+    va_start (args, format);
+    int ret = mp_vsnprintf(tmp, FORMAT_STACK_SIZE, format, args);
+    va_end(args);
+    if (ret <= 0) return UA_STRING_NULL;
+    UA_String s;
+    s.length = ret;
+    if (ret < FORMAT_STACK_SIZE)
+    {
+        s.data = (UA_Byte*) UA_malloc(s.length);
```
jpfr, line 146: Check for malloc fails.
This can happen on embedded targets.

Otherwise great.

andreasebner:
In recent versions, UA_String_format and UA_String_vformat have been implemented, making this PR obsolete. Thanks for sharing your ideas back then.

PR #6592, opened 2024-07-12
[REVIEW] Allow dynamic port assignment

NoelGraf:


keba-estr:
If dynamic port assignment should be fully supported with this pull request, should there really be a warning if this feature is used?

keba-estr:
Is there any suitable documentation that can be extended to document the usage of port 0 as dynamic port assignment?

NoelGraf:
Typically, dynamic port assignment is not used; it is intended primarily for test environments.
Therefore, a warning is issued.

jpfr:
👍

PR #6595, opened 2024-07-15
[REVIEW] Return all endpoints without requiring user identity token configuration

NoelGraf:
The correct user identity token is required to execute GetEndpoints().

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -929,12 +929,12 @@ responseGetEndpoints(UA_Client *client, void *userdata,
         }
 
         /* We have found a matching endpoint.
-         * But does it contain a matching user token policy? */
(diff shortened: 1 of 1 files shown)


include/open62541/client.h
```diff
@@ -109,6 +109,7 @@ struct UA_ClientConfig {
                                   * any matching SecurityPolicy. */
 
     UA_Boolean noSession;   /* Only open a SecureChannel, but no Session */
+    UA_Boolean noToken;     /* Don't check for matching user identity token */
```
jpfr, line 112: Make this an internal flag.
When we manually do a getendpoints, return all of them.
when we want to connect ourselves, of course we only consider the relevant ones..

jpfr:
:+1

jpfr:
:+1

PR #6602, opened 2024-07-19
Fix discoveryUrl initialisation

matkonnerth:
initialize newUrl
do not add path to discoveryUrl if empty, otherwise (null) will be added to the discoveryUrl


jpfr:
👍

PR #6606, opened 2024-07-23
feat(server): add local monitored event items

dpazj:


jpfr:
Nice!
I would like to extend the API as follows.
We can print the SimpleAttributeOperand that describes the event fields into a human readable string.

We should do that, cache the human readable string, and forward an (ordered) key value map to the event callback defined by the user.

The client does that already.

jpfr:
Scratch that.
The client also uses a flat list. I’ll propose something for the client. And we can reuse a similar api here.

PR #6607, opened 2024-07-24
refactor(examples): do not set an invalid IP in server_multicast.c

laroche:
Currently an IP 42.42.42.42 is set and the comment suggests that the network stack resets this to 0.0.0.0 in server_multicast.c. Just set 0.0.0.0 directly.


jpfr:
👍

PR #6608, opened 2024-07-24
refactor(server): do not set mdns IP address

laroche:
Starting LDS-ME with Linux I only get some data if changing from the mDNS IP 224.0.0.251 to 0.0.0.0.

With version 1.3 this was working ok and from the source it looks like multicast setup is now not done at all. Should this be re-added again for 1.4 to work properly on Linux?

### DIFF
--- a/src/server/ua_discovery_mdns.c
+++ b/src/server/ua_discovery_mdns.c
@@ -886,7 +886,11 @@ discovery_createMulticastSocket(UA_Server* server, UA_DiscoveryManager *dm) {
     size_t paramsSize = 5;
 
     UA_UInt16 port = 5353;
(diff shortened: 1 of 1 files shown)


jpfr:
Shouldn’t that be a config option?

laroche:
I've adjust the patch to put this under the config option UA_ARCHITECTURE_POSIX.
Does this look ok or what config option is better suited?

best regards,

Florian La Roche

jpfr:
What I meant is: Are these IP addresses fixed or should a user be able to change it in the UA_ServerConfig.

laroche:
I've reverted this change on my local machine. So I'll also close this patch request.

best regards,

Florian La Roche

Issue #6609, opened 2024-07-24
1.4.2: Build fails, if open62541 is not the top level folder in a project hierarchy.


r-sin:
## Description

Use case:
```bash
MyProject
   |-project_srcFiles
   |-CMakeLists.txt uses open62541's ua_generate_nodeset_and_datatypes
   |-open62541
        |-CMakeLists.txt
        |...  
```
In 1.4.2, in the open62541's top level CMakeLists.txt
```bash
set(UA_SCHEMA_DIR ${UA_NODESET_DIR}/Schema)
```
has been introduced in order to substitue 
```bash
'${UA_NODESET_DIR}/Schema/'
```
But UA_SCHEMA_DIR is not known in higher level projects (no cache variable).

ua_generate_nodeset_and_datatypes now uses
set(NODESET_DEPENDS "***${UA_SCHEMA_DIR}***/Opc.Ua.NodeSet2.xml") insteaf of
set(NODESET_DEPENDS "***${UA_NODESET_DIR}***/Schema/Opc.Ua.NodeSet2.xml")

UA_NODESET_DIR is a cached variable. UA_SCHEMA_DIR is not.

Potential fix: make UA_SCHEMA_DIR an internal cache variable.

## Background Information / Reproduction Steps

 - [x] open62541 Version 1.4.2
 - [x] Operating system: observed on Win10

r-sin:
Meanwhile, I don't think it has to do with the open62541 folder location but only with UA_SCHEMA_DIR being unknown.
In the standard build with enabled examples, the relevant logical branch in 'ua_generate_nodeset_and_datatypes' seems never to be entered. Our project references the open62541 as a subfolder and enters this branch due to the used parameters.

PR #6610, opened 2024-07-24
Make UA_SCHEMA_DIR an internal cache variable.

r-sin:
Fix for issue #6609 (build fails if open62541 is not the top level folder in a project hierarchy).


jpfr:
👍

PR #6616, opened 2024-07-26
[REVIEW] Check bounding properly before converting time until next scheduled callback

NoelGraf:


NoelGraf:
The client side is not affected.

jpfr (commented):
Use 0 directly

PR #6618, opened 2024-07-26
[REVIEW] refactor(server): Make first reverse connect attempt also from event loop

tomitkl:
Avoid making operations that possibly take long time (like getaddrinfo doing DNS lookup) from the calling thread when adding reverse connection. Instead of calling attemptReverseConnect directly from UA_Server_addReverseConnect, schedule immediate connect attempt from the event loop.

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -1119,11 +1119,13 @@ UA_Server_addReverseConnect(UA_Server *server, UA_String url,
     if(handle)
         *handle = newContext->handle;
 
-    /* Attempt to connect right away */
-    res = attemptReverseConnect(bpm, newContext);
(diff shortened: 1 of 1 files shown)


jpfr:
Ideally we have an async version of getaddrinfo.
So that this can be solved internally for the (particular) POSIX EventLoop.

andreasebner:
Thanks for sharing this PR. Closing as superseded — the underlying concern is valid but the approach conflicts with the current design. The proper fix for the blocking getaddrinfo is already acknowledged inside the TCP EventLoop itself: arch/posix/eventloop_posix_tcp.c line 991 carries a /* TODO: Make this non-blocking */ comment at exactly that call site.

PR #6626, opened 2024-07-30
fix(nc): Default for ValueRank is -1

### LINKED ISSUE #5217: Nodeset Compiler seems to parse ValueRank attribute incorrectly

Hello everyone,
it looks like the nodeset compiler has an issue with parsing the ValueRank Attribute when it is not available in the NodeSet2.xml file. In my understanding, when the ValueRank Attribute is not in the XML file, we should assume the default value -1 (scalar) according to the official UANodeSet.xsd file. 
grafik

However, the nodeset compiler tries apparently to inheret the ValueRank of the parent UAVariableType recursively until it reaches "BaseDataVariableType" which hase ValueRank -2 (any) and applies this ValueRank on all UAVariable nodes. For example "ServerStatus" and all its children.  "ServerStatus" should have ValueRank -1 according to the specification (Part 5, V1.04, 7.6), but instead it gets -2.
grafik

Also all the children of "ServerStatus" instance have ValueRank -2 (according to the specification, it is correct in the type address space becaue these variables has the type BaseDataVariableType). However, according to the NodeSet2.xml file, where the instances are defined, they should have -1! How to decide in this case, which ValueRank is correct?

Thank you in advance
 
Used CMake options:

```bash
cmake -DUA_BUILD_EXAMPLES:BOOL="1" -DUA_ENABLE_DA:BOOL="1" 
```

## Checklist

 - [x] open62541 Version (release number or git tag): V1.3.1
 - [x] Other OPC UA SDKs used (client or server): UaExpert (Client)
 - [x] Operating system: Windows
 - [ ] Critical issue

peschuster:
Fixes #5217

Basically reverts 1e020024e825e43269ef826fe271d95d296251f4

The previous change lead to ValueRank -2 for many variables with no explicit value in the node set xml. This also affected the ValueRank of many nodes in the "reduced" ns0.

Root cause of the original issue is a bug in the POWERLINK nodeset xml. This is reported in the OPC UA issue tracker (https://mantis.opcfoundation.org/view.php?id=9683) and a fix is proposed in the internal opc ua nodeset repository (OPCF-Members/UA-NodeSet#24).


jpfr:
This results in some breakage.
Potentially also for people already using v1.4.
Would it be okay for you to do the (breaking) change on master?

peschuster:
I rebased the commit onto the master branch.

I am open to other ideas on how to fix this or at least make the current POWERLINK nodeset work with open62541, but could not come up with another solution to the problem myself.
Looking through the XSDs, there is no ambiguity about the value rank of variables and types when no value is set in the nodeset xml.

Also the difference in the generated reduced and full ns0 is in my opinion an indicator that this is a bug/regression in open62541.

This also went unnoticed by us until recently when a customer contacted our support hotline because the OPC UA client in Beckhoff TwinCat does not support variables with value rank -2.

peschuster:
I've contacted the POWERLINK standards group to see if they can help solve the schema xml problem, but haven't heard back yet.

To solve this problem and get all builds working again, I see three options:

1. I am wrong in my analysis that the powerlink schema xml is broken and the valuerank issue needs to be fixed in the nodeset compiler in some other way. This would be the easiest option, but requires someone to point out why the powerlink schema xml is still valid.
2. The open62541 project temporarily removes all examples with and tests against the powerlink schema xml.
3. The open62541 project ships its own powerlink schema xml.

Option 3 is probably strategically bad, because the project would have to maintain its copy of the schema files, and there is no official process for proving that the schema xml files only remove bugs from the original files, and do not introduce other problems.

Option 2 seems to be the best in my opinion. At least until someone points out why option 1 is actually the right one 😉

I added another commit that removes the powerlink examples and tests.

jpfr:
Let's wait a few days for the feedback of the PowerLink group.
I'd also prefer version 2.
Even better, if we have a local patch for PowerLink that is aligned with the official channel.

jpfr:
On 1.5 the default value is -1 now.
Like mandated in the xsd schema definition for nodeset xml.

PR #6630, opened 2024-07-31
fix(plugin): search CRL in correct list in verifyCertificate (mbedTLS)

AdanRuoff:
The verifyCertificate function of the mbedTLS plugin has a bug (at least I assume so) in that when it finds the issuer of the remote certificate in the "list of **issuer** certificates", it will look for the appropriate CRL in the "list of **trusted** CRLs".


jpfr:
This looks relevant for 1,4.
Please make the pr for 1.4. we take it from there. 
@NoelGraf please have a look.

AdanRuoff:
1.4 does not seem to have this problem, as it keeps a single revocation list only (i.e. there is no distinction between "trusted" and "issuer" CRLs).

jpfr:
Got it.
It is merged now.

jpfr:
Thanks!

PR #6641, opened 2024-08-05
Fix binary encoding and decoding for DiagnosticInfo

basyskom-jvoe:
The order of the `locale` and `localizedText` fields is different between the encoding mask and the encoded values in the body: https://reference.opcfoundation.org/Core/Part6/v105/docs/5.2.2.12

The two fields have been in the wrong order since v0.1 of open62541.

### DIFF
--- a/deps/base64.c
+++ b/deps/base64.c
@@ -98,9 +98,9 @@ UA_unbase64(const unsigned char *src, size_t len, size_t *out_len) {
 		block[count] = tmp;
 		count++;
 		if(count == 4) {
-			*pos++ = (block[0] << 2) | (block[1] >> 4);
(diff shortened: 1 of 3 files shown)


jpfr:
*ouch* apparently nobody has used them so far...

Please make this a PR for 1.3.
We'll pull this forward from there.

basyskom-jvoe:
I have rebased the commit to the 1.3 branch.

I'm not sure if anybody will use DiagnosticInfo at all, I've just noticed the swap when checking our Qt OPC UA code generator against a structured type for testing purposes containing a DiagnosticInfo.

jpfr:
Merged. Thanks.

PR #6655, opened 2024-08-12
Client-side namespace management [WIP]

jpfr:
This allows the configuration of fixed namespace identifiers.
The mapping is performed automagically in the background.


andreasebner:
Solved in #6892 .

andreasebner:
Solved in #6892

PR #6662, opened 2024-08-22
fix(build): CMakeLists.txt: set BUILD_TYPE_LOWER_CASE after CMAKE_BUILD_TYPE is final

ped7g:
In case CMAKE_BUILD_TYPE is not set, it defaults to "Debug", but BUILD_TYPE_LOWER_CASE was left unset.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,8 +8,6 @@ endif()
 
 # set(CMAKE_VERBOSE_MAKEFILE ON)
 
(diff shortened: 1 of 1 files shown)


ped7g:
Just curious, unrelated to this change:
what's the point of the Azure pipeline "Dist (debian)"? Checking  the python script could produce the changelog? Seems to me like wasting 2-3minutes of azure on practically nothing.

jpfr:
Thanks. Merged.
Yes, you can send a PR to remove that azure build…

PR #6665, opened 2024-08-24
ci: comment out azure-pipeline-dist.yml

ped7g:
Looks like some remnants of Debian build, now doing nothing except producing fake test ".deb" files for artifact and wasting ~2-3min of azure servers time.

I know almost nothing about this project and its history to attempt to modify this toward functional Azure Debian build (I don't even know if that would be desired), but jpfr agreed about removing it - I can do *that*.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,4 +9,4 @@ trigger:
 
 jobs:
     - template: ./tools/azure-devops/azure-pipelines-win.yml
-    - template: ./tools/azure-devops/azure-pipelines-dist.yml
+    # - template: ./tools/azure-devops/azure-pipelines-dist.yml


ped7g:
Hmm... it says here in checks:
> open62541.open62541 (Dist (debian)) Expected — Waiting for status to be reported

But that's not defined in the source code, I guess that's something about github repo/actions configuration? If it's commented out, it can't be expected.

What now? Should be this "expected" configuration changed (by someone else, I don't think I could resolve this), or should I modify the pull request to instead of removing the build completely do the minimum-time empty busy task with expected name?

Or feel free just close/delete this PR. It's completely unimportant change, just my OCD nagging me for empty task wasting CI time.

jpfr:
👍

Issue #6663, opened 2024-08-23
pthread and rt problem on Android


ffreality:
## Background Information / Reproduction Steps

As I mention on my encryption related problem,
https://github.com/open62541/open62541/issues/6657

I made these changes to build open62541 for Android with Android NDK toolchain.
1. commented out `-lpthread` and `-lrt` libraries on `CMakeLists.txt`
2. commented out `<bits/stdio_lim.h>` in `ua_securitypolicy_filestore.c` **(stdio already has this include)**
3. Add `set(ANDROID_PLATFORM 24)` to `CMakeLists.txt` 
Because `<ifaddrs.h>` (used by `freeifaddrs` and `getifaddrs`) and `epoll_create1` came with `NDK level 24`
4. set cross compile toolchain on cmake as `NDK 25.1.8937393 android.toolchain.cmake`
5. set platform type as Posix

I can build open62541, integrate it to Unreal Engine, package and run it.
But if I use same `UA_Client*` for two operations in different thread at the same time (for example read a node value and historical access. I use call historical access in another thread because I don't want to block main thread), I get `art_sigsegv_fault` crash like below.

![Screenshot (107)](https://github.com/user-attachments/assets/32528888-114e-4b62-9215-617aee20e752)

## Used CMake options:

```
USED NDK TOOLCHAIN = NDK 25.1.8937393 on Android Studio 2024.1.1 Patch 2 Koala on Ubuntu 24.04 LTS

-DUA_ENABLE_ENCRYPTION_TPM2:STRING="OFF" 
-DCMAKE_INSTALL_LIBEXECDIR:PATH="libexec" 
-DUA_ENABLE_DISCOVERY:BOOL="1" 
-DUA_LOGLEVEL:STRING="100" 
-DUA_ARCHITECTURE:STRING="posix" 
-DUA_NAMESPACE_ZERO:STRING="FULL" 
-DCMAKE_INSTALL_SYSCONFDIR:PATH="etc" 
-DUA_ENABLE_PUBSUB_MONITORING:BOOL="1" 
-DUA_AMALGAMATION_MULTIARCH:BOOL="1" 
-DCMAKE_INSTALL_INFODIR:PATH="" 
-DUA_ENABLE_PUBSUB_ENCRYPTION:BOOL="1" 
-DUA_ENABLE_PUBSUB:BOOL="1"
-DUA_ENABLE_AMALGAMATION:BOOL="1" 
[...]

PR #6666, opened 2024-08-25
Android Compatibility Improvements

ffreality:
- `make` can't find `pthread` and `rt` libraries while building `open62541` with `android.toolchain.cmake` and they cause errors. 
But NDK already has them. So, we don't need to look them for Android.

- We need to set `Android platform` to at least `24` because `epoll_create1` came with level `21` and `ifaddrs.h` came with `24`.

- `#include <bits/stdio_lim.h>` already included in `stdio`. We don't need to add it and android can't find it.

https://github.com/open62541/open62541/issues/6663

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.13)
 
+# Default android platform of android.toolchain.cmake is "19" but we need "24" for "epoll_create1" and "<ifaddrs.h>"
+set(ANDROID_PLATFORM 24)
+
 if(UA_BUILD_FUZZING OR UA_BUILD_OSS_FUZZ OR UA_BUILD_FUZZING_CORPUS)
     project(open62541) # We need to have C++ support configured for fuzzing
 else()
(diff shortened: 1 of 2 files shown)


ped7g (approved):

CMakeLists.txt
```diff
@@ -438,10 +446,10 @@ set(open62541_LIBRARIES "")
 set(open62541_PUBLIC_LIBRARIES "")
 if("${UA_ARCHITECTURE}" STREQUAL "posix")
     list(APPEND open62541_LIBRARIES "m")
-    if(UA_MULTITHREADING GREATER_EQUAL 100 OR UA_BUILD_UNIT_TESTS)
+    if(UA_MULTITHREADING GREATER_EQUAL 100 OR UA_BUILD_UNIT_TESTS AND NOT ANDROID_NDK_TOOLCHAIN_INCLUDED)
```
ped7g, line 449: I personally (not open62541 maintainer) prefer parentheses enforcing precedence when AND and OR is compounded like this, so I don't have to remember what language I'm reading and what are the precedence rules for this language.

ffreality: thank you for the document. I prefer to use parenthesis when there are lots of calculations but if it is necessary for cmake or open5241, I can change it.

PR #6667, opened 2024-08-26
fix(plugin): remove overhead for thumbprint creation (MbedTLS)

mcpat:
This PR removes two allocations while creating the certificate thumbprint and instead does the binary to hexstring conversion in place.

### DIFF
--- a/plugins/crypto/mbedtls/certificategroup.c
+++ b/plugins/crypto/mbedtls/certificategroup.c
@@ -673,25 +673,31 @@ UA_CertificateUtils_getThumbprint(UA_ByteString *certificate,
     if(certificate == NULL || thumbprint->length != (UA_SHA1_LENGTH * 2))
         return UA_STATUSCODE_BADINTERNALERROR;
 
-    UA_ByteString thumbpr = UA_BYTESTRING_NULL;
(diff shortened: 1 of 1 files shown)


jpfr:
Yes. It is better to stack-allocate where you have a fixed size.
But please keep the magic to a minimum.
Use deps/itoa.h to print hex with base 16.

mcpat:
Using deps/itoa.h either requires to handle alignment properly (new magic of different kind) or to call the conversion function byte-wise (expensive). I do not like either choice so feel free to just close this PR.

jpfr:
The speed improvement is still a very good improvement.
What we could do:

- Use `char buf[UA_SHA1_LENGTH];` to stack-allocate the fixed size buffer.
Instead of the malloc/free pair.

- Provide a reference for the hex conversion you did.
This is tricky code. And a good reference would help to increase our confidence.
Otherwise we need to do a pretty deep review to ensure there are no bugs.

mcpat:
Would it make sense to extend deps/itoa.h to also support a function to convert an arbitrary buffer of bytes into its ASCII representation? I could then re-use existing code without the need of calling the conversion function multiple times.

jpfr:
Nice, thanks.

PR #6668, opened 2024-08-26
fix(server): Select the right sp to check userTokenSignature

Maddin-619:
If there is a userTokenPolicy use that one to verify the userTokenSignature


jpfr:
Good find!

PR #6670, opened 2024-08-28
fix(server): fix issue where namespace1 would be created prematurely

dpazj:


jpfr:
Instead of clearing the string, better set the correct ns1 (the application uri from the server config) before calling initns0.

dpazj:
> Instead of clearing the string, better set the correct ns1 (the application uri from the server config) before calling initns0.

initns0 is called in `UA_Server_init` meaning that if we use UA_Server_new then ns1 is already set before the applicationUri can be set in the config. Is this behaviour expected? I guess a work around would be to use `UA_Server_newWithConfig` but it seems the usage of UA_Server_new would be that you can set the applicationURI and ns1 to that.

jpfr:
UA_Server_new internally calls UA_Server_newWithConfig.
We could just reset the ns1 at every startup.
So we can change the config in between.

At any rate we should have ns1 before initNs0.

dpazj:
So to clarify just clear ns1 (`server->namespaces[1]`)  in `UA_Server_run_startup` ?

jpfr:
> So to clarify just clear ns1 (server->namespaces[1]) in UA_Server_run_startup ?

Yes.

jpfr:
Nice, merged 👍

PR #6671, opened 2024-08-28
send multicast messages to all valid hardware interfaces.

engel-bichlerh:


jpfr:
Thanks for the improvement.

Please fix the CI breakage and document the behavior in the public header of the eventloop.

arch/eventloop_posix/eventloop_posix_udp.c
```diff
@@ -997,6 +997,29 @@ UDP_shutdownConnection(UA_ConnectionManager *cm, uintptr_t connectionId) {
...
+#else
+isAddressValid(struct ifaddrs *ifa) {
+    char sourceAddr[NI_MAXHOST];
+    if(ifa->ifa_addr->sa_family == AF_INET) {
+        getnameinfo(ifa->ifa_addr,
+                    (ifa->ifa_addr->sa_family == AF_INET) ? sizeof(struct sockaddr_in)
+                                             : sizeof(struct sockaddr_in6),
+                sourceAddr, NI_MAXHOST, NULL, 0, NI_NUMERICHOST);
+#endif
+        return !(sourceAddr[0] == '1' && sourceAddr[1] == '6' && sourceAddr[2] == '9' &&
```
ccvca, line 1016: "169.254.0.0/16" is AutoIP and might be a totally valid IP Address. Why do you exclude it?

PS: Binary checks should be always preffered over string comparisons.

ccvca (commented):
Will all UDP-Messages be send to all interfaces now? (Or did I miss the check for multicast messages?)

PR #6673, opened 2024-08-29
fix(examples): Remove server_pubsub_subscribe_custom_monitoring from Windows compilation

RolfKal:
This example uses several Posix functions such as timer_create() and also sigevent that are not available in the Win32 API.


jpfr:
The example was recently revamped on master.
Is this still an issue today?

PR #6674, opened 2024-08-29
fix(pubsub): Move two APIs to the public header

RolfKal:
Move two APIs to the public header as they are used in examples\pubsub\server_pubsub_file_configuration.c
Also fix the prototype for UA_Server_setReaderGroupEncryptionKeys() so the parameter types match the declaration of the actual function

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.13)
 
+# Default android platform of android.toolchain.cmake is "19" but we need "24" for "epoll_create1" and "<ifaddrs.h>"
+set(ANDROID_PLATFORM 24)
+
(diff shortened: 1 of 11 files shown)


jpfr:
Looks good.
Please adjust the method names to begin with UA_Server_.
Like the rest of the public pubsub API that acts on the server.

jpfr:
In ua_pubsub_ns0.c you take the server lock before calling into the method that is now public.
The public method should be threadsafe and take the lock internally.
That is, there can be a public method that takes the lock and then calls the internal method.

To make the API naming more symmetric, you could use `UA_Server_writePubSubConfigurationToByteString` or similar.

jpfr:
Nice. :-)
Squash-merged into a single commit.

PR #6676, opened 2024-09-02
fix(core): add UA_EXPORT for UA_DataType_getStructMember

ralphlange:
When building under Linux, the function `UA_DataType_getStructMember` is a global (external) symbol when building statically and a local symbol when building a shared library.

This fix makes it a global (external) symbol for both build types.

### DIFF
--- a/src/ua_types.c
+++ b/src/ua_types.c
@@ -1882,7 +1882,7 @@ UA_Array_delete(void *p, size_t size, const UA_DataType *type) {
 }
 
 #ifdef UA_ENABLE_TYPEDESCRIPTION
-UA_Boolean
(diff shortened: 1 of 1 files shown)


jpfr:
Good find.  Thanks.

ralphlange:
Thinking... does that need a fix for Windows, too?

ralphlange:
I forgot to apply the same change to the header file.

The merged branch has another commit fixing this omission.
Sorry!

ralphlange:
@jpfr Please also merge the second commit - the header needs the same decoration.

Won't be noticed on Linux, but broke my MSVC builds. (Using a cmake-created VS project file.)

PR #6679, opened 2024-09-03
fix(server): Eliminate memcpy() undefined behavior

teijokinnunen:
Calling memcpy() with a null pointer is undefined behavior even if the length is zero. The source pointer may be null if there are no arguments so check the argument count before calling memcpy().


jpfr:
👍

PR #6688, opened 2024-09-09
fix(sec): avoid memmory read access errors

soekkle:
Check for null pointer befor passing then into a openssl function.


jpfr:
Nice, we will pull this to 1.4 and master from here.

PR #6689, opened 2024-09-09
fix(sec): avoid memmory  read access errors

soekkle:
Check for null pointer befor passing then into a openssl function.

See https://github.com/open62541/open62541/pull/6688 for 1.3 branch.


jpfr:
👍

PR #6699, opened 2024-09-12
fix(pubsub): Fix exception at PubSub update

keba-estr:
If a beforeWrite callback is used without an externalDataValue pointer, an exception occurred.


keba-estr:
@jpfr: What do I have to do to fix the CIFuzz pipeline error?

keba-estr:
@jpfr: Why does the Windows Build: Visual Studio 16 2019 fail?

keba-estr:
To be merged after #6700

jpfr:
Looks good. Merged.

PR #6705, opened 2024-09-12
[REVIEW] Proper error handling when checking parent CRL and adding test cases

NoelGraf:


mcpat:
Hi Noel,

what do you mean by "checking all CRLs"? Does it mean "all CRLs" of an (more or less arbitrary) issuer certificate? Or does it mean "all CRLs" of all (root+intermediate) issuer certificates in a verification chain?

I ask because mbedTLS treats the given CA list and CRL list as trusted during verification and will stop if the first issuer certificate is found in the given CA list. So with that in mind my second interpretation of this MR title is not that easily accomplished.

NoelGraf:
Currently, the system only verifies whether the parent certificate has a Certificate Revocation List (CRL) and whether it is accessible, but it does not recheck the status. As a result, an application certificate may be accepted even if the intermediate certificate (issuer) has been revoked by the root certificate.

mcpat:
Ok, so instead of a generic approach you stick to the CA -> Cert and CA -> interm. CA -> Cert scenarios. Deeper certificate chains will therefore still not be validated correctly.

Issue #6707, opened 2024-09-13
Worse PubSub performance without set externalDataValue


keba-estr:
## Description
Since commit 3a59e733de9d007bc1944c5f78dd5686070577d2 (fix(pubsub): Don't overwrite the ReaderGroup external value configuration) the externalDataValue of a subscribed variable in the subscribed data set is not set by open62541 in the function UA_ReaderGroup_freezeConfiguration any more.
If the externalDataValue has not been set by the user, the performance of the update decreases by a factor of 4 (on my system), as the externalDataValue is not used for update any more.

@jpfr: Why has the code been commented with this commit?
Is it now strictly necessary to set externalDataValue externally when initializing the subscribed data set, or should the externalDataValue only be set in the function UA_ReaderGroup_freezeConfiguration if it is not already set?

## Checklist

Please provide the following information:

 - [X] open62541 Version (release number or git tag): 1.4
 - [X] Operating system: Linux

keba-estr:
Proposal for solution in #6708

keba-estr:
Result of discussion with jpfr: No imcompatible change on 1.4 branch, on master branch the documentation of the write callbacks will be updated such that the new value is not provided at beforeWrite, but both functions can be used for double buffering.

keba-estr:
The externalDataValue can and must be set by the customer, if needed.

PR #6708, opened 2024-09-13
fix(pubsub): Set externalDataValue in subscribed data set if not set …

keba-estr:
…by user

With commit 3a59e733de9d007bc1944c5f78dd5686070577d2 the setting of the externalDataValue was commented in the function
UA_ReaderGroup_freezeConfiguration. If the externalDataValue is not set by the user, the performance of the update decreases drastically.

Thus the externalDataValue is now set in the function UA_ReaderGroup_freezeConfiguration, if it has not been set by the user and if the user provided the backend type UA_VALUEBACKENDTYPE_EXTERNAL in the variable node configuration.

### DIFF
--- a/src/pubsub/ua_pubsub_readergroup.c
+++ b/src/pubsub/ua_pubsub_readergroup.c
@@ -628,24 +628,21 @@ UA_ReaderGroup_freezeConfiguration(UA_Server *server, UA_ReaderGroup *rg) {
 
     size_t fieldsSize = dsr->config.dataSetMetaData.fieldsSize;
     for(size_t i = 0; i < fieldsSize; i++) {
-        /* TODO: Use the datasource from the node */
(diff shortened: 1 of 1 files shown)


jpfr:
Looks good.
Please add logging to it.
So that users are aware their configuration has been overridden.

keba-estr:
> Looks good. Please add logging to it. So that users are aware their configuration has been overridden.

The new implementation does not overwrite the user configuration. Should it do so? See #6707

keba-estr:
Add logging (info) that externalDataValue has been set by open62541, but not set by the user.

keba-estr:
Added logging message (once per data set reader)

keba-estr:
@jpfr: Is this PR ok and mergable?

keba-estr:
The externalDataValue can and must be set by the customer, if needed.

PR #6710, opened 2024-09-13
fix(pubsub): Correct documentation of beforeWrite callback

keba-estr:
Remove a misleading comment about the content of the variable
externalDataValue in the PubSub beforeWrite callback.
The beforeWrite callback is called with the data pointer containing the
data before the update. It may be used for a double buffer
implementation.
The original code change that changed the usage of the beforeWrite
callback was done with commit
06da5d07bd9a653d32e7a6ed06deae994d29953c.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,19 @@
 This changelog reports changes visible through the public API. Internal refactorings and bug
 fixes are not reported here.
 
+2025-02-18 Andreas Eckerstorfer <estr at keba.com>
+
+ * PubSub beforeWrite callback
+
+   A misleading comment about the content of the variable
+   externalDataValue in the PubSub beforeWrite callback has been
+   removed. Please review your beforeWrite callback implementation.
+   The beforeWrite callback is called with the data pointer
+   containing the data before the update. It may be used for a
+   double buffer implementation.
+   The original code change was done with commit
+   06da5d07bd9a653d32e7a6ed06deae994d29953c.
+
 2023-07-02 Jonas Green <jgr at hms.se>
 
  * Decoding variant with array of structure
--- a/include/open62541/server_pubsub.h
+++ b/include/open62541/server_pubsub.h
@@ -652,7 +652,7 @@ typedef struct {
     /* If realtime-handling is required, set this pointer non-NULL and it will be used
      * to memcpy the value instead of using the Write service.
      * If the beforeWrite method pointer is set, it will be called before a memcpy update
-     * to the value. But param externalDataValue already contains the new value.
+     * to the value.
      * If the afterWrite method pointer is set, it will be called after a memcpy update
      * to the value. */
     UA_DataValue **externalDataValue;


src/pubsub/ua_pubsub_reader.c
```diff
@@ -929,7 +933,7 @@ UA_DataSetReader_process(UA_Server *server, UA_ReaderGroup *rg,
             if(tv->beforeWrite)
                 tv->beforeWrite(server, &dsr->identifier, &dsr->linkedReaderGroup,
                                 &tv->targetVariable.targetNodeId,
-                                tv->targetVariableContext, tv->externalDataValue);
+                                tv->targetVariableContext, &field);
```
jpfr, line 936: I think, there is a misunderstanding.
We want to give `tv->externalDataValue` into beforeWrite, so that the target memory location can be switched out before writing into it.
So that we can do "double buffering" where we don't want to have second thread work on a half-updated value.

jpfr (commented):
Comment inline.

src/pubsub/ua_pubsub_reader.c
keba-estr: I tried to fix this with an updated commit similar to the first call of beforeWrite. Is this ok this way?

keba-estr: @jpfr: If the solution in this PR is not suitable for the double buffering, please provide a solution where tv->externalDataValue is provided with the new value to beforeWrite according to the documentation. The user callbacks rely on the documentation.

jpfr: Set `*tv->externalDataValue` to the new location so that `**tv->externalDataValue` gets written to it.

keba-estr: @jpfr: Set *tv->externalDataValue now.
Is the first change from line 815 on fine?

keba-estr:
@jpfr: Can you please find out why the windows msys64 builds fail? The web frontend does not load the entire log.

src/pubsub/ua_pubsub_reader.c
jpfr: I don't understand the preview thing.
If you are unsure if the data is acceptable, you can use `beforeWrite` to set up a temporary memory (and have the double-pointer point to it) and in `afterWrite` decide whether you want to to keep the value...

keba-estr: I don't understand it either. I just noticed the incompatible change in the callback usage.

keba-estr:
@jpfr:
Thinking about a different compatible solution again, maybe it would also be possible to let the external value point to the new value only, if only the beforeWrite callback is set. If both the beforeWrite and afterWrite callback is set (use case of double buffer), the pointer would not be altered. What is your opinion about that?

keba-estr:
After discussion with jpfr 18.02.2025:
- Remove documentation of new value in header file
- Add to CHANGELOG:
Misleading comment removed, please review your beforeWrite implementation
with link to old commit

keba-estr:
@jpfr: Please check if the solution is fine

PR #6713, opened 2024-09-14
Simplify Nodeset Compiler by parsing values only at runtime

jpfr:
Defer parsing the values in XML encoding.
So we can keep that complexity entirely out of the Nodeset Compiler.

This requires the full variant-parsing for XML from #5535.


jpfr:
Already resolved

PR #6719, opened 2024-09-22
fix(nc): Improve handling of QualifiedName

MurzynJ:
1) NodesetCompiler run into an error parsing the QualfiedName Value, if the Name element was empty.
We got a NodeSet with such code to compile. It's questionable whether such QualifiedName makes sense at all, but at least it's a valid XML, produced by one of the modelling tools.
image

2) NamespeceIdx of the QualifiedName was not remapped to the target scope.
I used the same pattern to remap the index as in the NodeId related function.


andreasebner:
Thanks for shareing, landed via #8265

PR #6727, opened 2024-09-24
Refactor certificate verification for mbedtls

jpfr:


plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -236,251 +238,158 @@ reloadCertificates(CertInfo *ci) {
...
+    /* Verification Step: Revocation Check */
+    if(mbedtlsCheckRevoked(ci, cert))
+        return UA_STATUSCODE_BADCERTIFICATEREVOKED;
+
+    /* Is the certificate in the trust list? If yes, then we are done. */
+    for(mbedtls_x509_crt *t = &ci->certificateTrustList; t; t = t->next) {
+        if(cert->tbs.len == t->tbs.len &&
+           memcmp(cert->tbs.p, t->tbs.p, cert->tbs.len) == 0)
+            return UA_STATUSCODE_GOOD;
     }
```
mcpat, line 295: It is understandable to shorten the verification, but two issues have to be considered here:

1. We cannot be sure that the TrustList is consistent. This is because of the RemoveCertificate() function which may eventually remove a CA certificate. Let's imagin the root CA has been removed (regardless of which sublist) and we currently look at an derived intermediate CA certificate which happens to be in the list of trusted certificates. Returning here now signals that the certificate chain is complete - but it isn't.
2. Expiration of CA certificates is not considered here. If we are not at the top of the chain here, it is possible that any CA the current certificate depends on might have been expired. Still we acknowledge the current certificate as valid - but it isn't.

To prevent point 1 one can consider to let RemoveCertificate() also remove all certificates that depend on the to-be-deleted one. Yet point 2 remains and seems to force us to completely validate the chain without a shortcut.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -236,251 +238,158 @@ reloadCertificates(CertInfo *ci) {
...
+mbedtls_verifyChain(CertInfo *ci, mbedtls_x509_crt *cert,
+                    mbedtls_x509_crt **old_issuers, int depth) {
+    /* Maxiumum chain length */
+    if(depth == UA_MBEDTLS_MAX_CHAIN_LENGTH)
+        return UA_STATUSCODE_BADCERTIFICATECHAININCOMPLETE;
+
+    /* Verification Step: Validity Period */
+    if(mbedtls_x509_time_is_future(&cert->valid_from) ||
+       mbedtls_x509_time_is_past(&cert->valid_to))
+        return UA_STATUSCODE_BADCERTIFICATETIMEINVALID;
```
mcpat, line 284: Isn't it required to check what certificate type we are actually looking at? For several error conditions there are versions for Application Instance Certificates (AICs) and for Issuer Certificates.

If we look at an AIC `UA_STATUSCODE_BADCERTIFICATETIMEINVALID` is ok here. But for an Issuer Certificate it should be `UA_STATUSCODE_BADCERTIFICATEISSUERTIMEINVALID` or not? This should apply to further conditions as well.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -236,251 +238,158 @@ reloadCertificates(CertInfo *ci) {
...
+
+    /* Verification Step: Revocation Check */
+    if(mbedtlsCheckRevoked(ci, cert))
+        return UA_STATUSCODE_BADCERTIFICATEREVOKED;
```
mcpat, line 288: What happens if no CRL is found here? Shouldn't it return `UA_STATUSCODE_BADCERTIFICATEREVOCATIONUNKNOWN` or `UA_STATUSCODE_BADCERTIFICATEISSUERREVOCATIONUNKNOWN`?

mcpat (commented):
Well this is an awesome piece of work. Good job!
I added some comments below.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
jpfr: 1. That is correct. We might want to "walk up" until we reach a self-signed certificate.
It is not enough to have a trusted intermediate certificate.

2. This is considered. We check the expiration of every certificate in the chain already.

mcpat: 2. If I understand the code correctly the expiration is only checked for each certificate that is processed. But if the shortcut applies (meaning that you do not walk up the chain any further) the expiration of all remaining certificates higher up the chain that is not checked anymore.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
```diff
@@ -236,251 +238,158 @@ reloadCertificates(CertInfo *ci) {
...
     }
 #else
-    if((remoteCertificate.private_key_usage & MBEDTLS_X509_KU_KEY_CERT_SIGN) &&
-       (remoteCertificate.private_key_usage & MBEDTLS_X509_KU_CRL_SIGN)) {
+    if((cert.private_key_usage & MBEDTLS_X509_KU_KEY_CERT_SIGN) &&
+       (cert.private_key_usage & MBEDTLS_X509_KU_CRL_SIGN)) {
+        mbedtls_x509_crt_free(&cert);
         return UA_STATUSCODE_BADCERTIFICATEUSENOTALLOWED;
     }
 #endif
```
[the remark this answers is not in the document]
jpfr: Nice. Done.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
jpfr: We now walk the entire chain until the self-signed root CA.
Then we try if a trusted certificat is in the chain on the "way down".
Still we do the correct branching to explore the "tree of possible chains".

plugins/crypto/mbedtls/ua_pki_mbedtls.c
jpfr: The standard says the following:
> Each CA [Certificate](https://reference.opcfoundation.org/search/357?t=Certificate) may have a revocation list. This check fails if this list is not available (i.e. a network interruption prevents the application from accessing the list). No error is reported if the [Administrator](https://reference.opcfoundation.org/search/357?t=Administrator) disables revocation checks for a CA [Certificate](https://reference.opcfoundation.org/search/357?t=Certificate).

To me this is something for certificates with a defined CRL distribution point (e.g. HTTP URL to download the CRL).

We have many "casual users" who just dump a list of trusted certificates into their config.
But who do not have a CRL for all CA or self-signed certificates.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
jpfr: We now distinguish between the two error codes.

plugins/crypto/mbedtls/ua_pki_mbedtls.c
mcpat: The standard is not clear about this point. Reading the description of "AddCertificate()"

> This [Method](https://reference.opcfoundation.org/search/49?t=Method) cannot provide CRLs so issuer [Certificates](https://reference.opcfoundation.org/search/49?t=Certificates) cannot be added with this [Method](https://reference.opcfoundation.org/search/49?t=Method). Instead, CA [Certificates](https://reference.opcfoundation.org/search/49?t=Certificates) and their CRLs shall be managed with the [Write Method](https://reference.opcfoundation.org/search/49?t=Write%20Method) on the containing [TrustList Object.](https://reference.opcfoundation.org/search/49?t=TrustList%20Object.)

we conclude that CRLs are mandatory for all CAs.

jpfr: The trustlist data type looks like this:
grafik
And there are a couple of enums to define whether CRLs are provided or not.

I think we should handle the CRL presence check in the GDS code. Not in the underlying crypto plugin.
So we can do it just once for OpenSSL and mbedTLS.
At least for 1.3 and 1.4 we will not make the (breaking) change.

mcpat: I don't known if I can follow. Isn't it required to parse the certificates and CRLs using the underlying crypto library to do the matching? I would assume that most code already have to be in the crypto plugin. So either you end up parsing stuff multiple times our I just do not understand your proposal.

PR #6733, opened 2024-09-27
fix(plugins): Check for both MBEDTLS_X509_KU_KEY_CERT_SIGN and MBEDTL…

jpfr:
…S_X509_KU_CRL_SIGN to check certificate usage

### DIFF
--- a/plugins/crypto/mbedtls/ua_pki_mbedtls.c
+++ b/plugins/crypto/mbedtls/ua_pki_mbedtls.c
@@ -408,12 +408,13 @@ certificateVerification_verify(void *verificationContext,
 
     /* Verification Step: Certificate Usage
      * Check whether the certificate is a User certificate or a CA certificate.
-     * If the KU_KEY_CERT_SIGN and KU_CRL_SIGN of key_usage are set, then the
+     * If the KU_KEY_CERT_SIGN or KU_CRL_SIGN of key_usage are set, then the
      * certificate shall be condidered as CA Certificate and cannot be used to
      * establish a connection. Refer the test case CTT/Security/Security
      * Certificate Validation/029.js for more details */
-    unsigned int ca_flags = MBEDTLS_X509_KU_KEY_CERT_SIGN | MBEDTLS_X509_KU_CRL_SIGN;
-    if(mbedtls_x509_crt_check_key_usage(&cert, ca_flags)) {
+    if(mbedtls_x509_crt_check_key_usage(&cert, MBEDTLS_X509_KU_KEY_CERT_SIGN) ||
+       mbedtls_x509_crt_check_key_usage(&cert, MBEDTLS_X509_KU_CRL_SIGN) ||
+       cert.MBEDTLS_PRIVATE(ca_istrue)) {
         mbedtls_x509_crt_free(&cert);
         return UA_STATUSCODE_BADCERTIFICATEUSENOTALLOWED;
     }
--- a/plugins/crypto/openssl/ua_pki_openssl.c
+++ b/plugins/crypto/openssl/ua_pki_openssl.c
@@ -589,12 +589,12 @@ UA_CertificateVerification_Verify(void *verificationContext,
 
     /* Verification Step: Certificate Usage
      * Check whether the certificate is a User certificate or a CA certificate.
-     * If the KU_KEY_CERT_SIGN and KU_CRL_SIGN of key_usage are set, then the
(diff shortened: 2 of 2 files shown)


mcpat:
Is there a reason why you change the logic from "and" to "or"? And isn't it problematic now because self-signed Application Instance Certificates will have KU_KEY_CERT_SIGN set but shall not be treated as CA certificates?

[The spec states](https://reference.opcfoundation.org/Core/Part6/v105/docs/6.2): Self-signed [Certificates](https://reference.opcfoundation.org/search/359?t=Certificates) shall also include keyCertSign.

jpfr:
The PR is no longer current.
The current proposal is this:
https://github.com/open62541/open62541/pull/6732/commits/bfe91e2b088fe3f20fd88349b3f6c16d2a53bec6

```c
static UA_Boolean
mbedtlsCheckCA(mbedtls_x509_crt *cert) {
    /* The Basic Constraints extension is set. Can the cert act as CA? */
    if(cert->MBEDTLS_PRIVATE(ext_types) & MBEDTLS_X509_EXT_BASIC_CONSTRAINTS)
        return cert->MBEDTLS_PRIVATE(ca_istrue);
    /* The Key Usage extension is set. Is a CA usage permitted? */
    if(cert->MBEDTLS_PRIVATE(ext_types) & MBEDTLS_X509_EXT_KEY_USAGE)
        return (mbedtls_x509_crt_check_key_usage(cert, MBEDTLS_X509_KU_KEY_CERT_SIGN) == 0 ||
                mbedtls_x509_crt_check_key_usage(cert, MBEDTLS_X509_KU_CRL_SIGN) == 0);
    return false;
}
```

This is very close to how OpenSSL detects CA also.
We could even simplify that and only look for `MBEDTLS_X509_KU_KEY_CERT_SIGN` to detect a CA.

The merge is currently blocked as it breaks some of the SKS unit tests.
Working on it.

mcpat:
Sadly this check will reject the certificates UAExpert issues because they have the CA flag set. This seems to be legacy issue yet it is still present in this specific UA client.

I found your original version quite intriguing where you check for the combined key usage first (CERT_SIGN | CRL_SIGN). 
In this case it shall also have the CA flag set - otherwise it is not an acceptable certificate at all.

jpfr:
What we want to protect against are end-user certificates being used to sign more child-certificates.
This is how it is done natively in OpenSSL and mbedTLS.

https://github.com/openssl/openssl/blob/12d14de641c299ec080edc521f7080acc44e366f/crypto/x509/v3_purp.c#L653
https://github.com/Mbed-TLS/mbedtls/blob/46771295f2f49d8f9b91532c84d45fde4e35a5b9/library/x509_crt.c#L2180

mbedTLS needs the CA flag in the basic constraints **and** the MBEDTLS_X509_KU_KEY_CERT_SIGN flag.
I would opt that we just require either.
So we protect against "normal certificates" being used for signing (other than self-signing) and we are more compatible with the existing OPC UA clients.

mcpat:
The normal certificate UAExpert generates has the following properties:

```
        X509v3 extensions:
            X509v3 Basic Constraints: critical
                CA:TRUE, pathlen:0
            Netscape Comment: 
                "Generated with Unified Automation UA Base Library using OpenSSL"
...
            X509v3 Key Usage: critical
                Digital Signature, Non Repudiation, Key Encipherment, Data Encipherment, Certificate Sign
            X509v3 Extended Key Usage: critical
                TLS Web Server Authentication, TLS Web Client Authentication
```

So it has the CA flag and CERT_SIGN key usage set. Worth noting is, that the pathlen is 0 (which is also required by the OPC UA spec in this case). So if the CRL_SIGN key usage is not mandatory for a CA (in terms of OPC UA!) than the only way to distinguish between CA and non-CA would be the pathlen then.

You may neither rely an MbedTLS nor on OpenSSL to do this categorisation properly in terms of the OPC UA spec. Therefore it is essential to also validate the TrustList with the same measures to really ensure, that no "normal" certificate sign more child certificates.

mcpat:
As far as I see now: CRLs are not mandatory for CAs if and only if the revocation status has to be checked online (via OCSP). For offline checking CRLs are mandatory and therefore a CA must have the CRL_SIGN key usage specified.

This somehow follows after reading https://reference.opcfoundation.org/Core/Part6/v105/docs/E.6

So as long as OCSP is not supported, any CA must have the key usage CRL_SIGN specified which would be ~~a~~**the** criterion to distinguish between CA and normal certificate.

PR #6741, opened 2024-10-01
feat: Add zephyr support

xR3b0rn:
Just a little side note for anyone trying to make this library run on small embedded devices. With the following open62541 CMake options:
```
set(UA_ENABLE_DISCOVERY_SEMAPHORE OFF CACHE STRING "" FORCE)
set(UA_ARCHITECTURE "zephyr" CACHE STRING "" FORCE)
set(UA_MULTITHREADING 0 CACHE STRING "" FORCE)
set(UA_NAMESPACE_ZERO "MINIMAL" CACHE STRING "" FORCE)
set(UA_ENABLE_PUBSUB OFF CACHE STRING "" FORCE)
set(UA_ENABLE_PUBSUB_INFORMATIONMODEL OFF CACHE STRING "" FORCE)
set(UA_ENABLE_SUBSCRIPTIONS_EVENTS OFF CACHE STRING "" FORCE)
set(UA_ENABLE_SUBSCRIPTIONS OFF CACHE STRING "" FORCE)
set(UA_ENABLE_HISTORIZING OFF CACHE STRING "" FORCE)
set(UA_ENABLE_JSON_ENCODING OFF CACHE STRING "" FORCE)
```
The foot print of the zephyr.elf is:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      671980 B         1 MB     64.09%
```

# What is included?
* Implemented TCP event loop support for Zephyr.
* Ensured consistent usage of abstraction layer macros such as `UA_getaddrinfo` and `UA_fd`.
* Standardized the use of the `UA_ARCHITECTURE_*` macros across the event loop abstraction layer. Introduced type aliases for common structures (e.g., using `UA_addrinfo` instead of `struct addrinfo`).
* Updated the top-level CMakeLists.txt to add compile options to the target instead of directly modifying `CMAKE_C_FLAGS`.

### DIFF
--- a/include/open62541/plugin/eventloop.h
+++ b/include/open62541/plugin/eventloop.h
@@ -750,6 +750,13 @@ UA_ConnectionManager_new_MQTT(const UA_String eventSourceName);
 UA_EXPORT UA_InterruptManager *
 UA_InterruptManager_new_POSIX(const UA_String eventSourceName);
 
+#elif defined(UA_ARCHITECTURE_ZEPHYR)
+
+UA_EXPORT UA_EventLoop *
+UA_EventLoop_new_Zephyr(const UA_Logger *logger);
+UA_EXPORT UA_ConnectionManager *
+UA_ConnectionManager_new_Zephyr_TCP(const UA_String eventSourceName);
+
 #endif /* defined(UA_ARCHITECTURE_POSIX) || defined(UA_ARCHITECTURE_WIN32) */
 
 _UA_END_DECLS
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ set(ANDROID_PLATFORM 24)
 if(UA_BUILD_FUZZING OR UA_BUILD_OSS_FUZZ OR UA_BUILD_FUZZING_CORPUS)
     project(open62541) # We need to have C++ support configured for fuzzing
 else()
-    project(open62541 C) # Do not look for a C++ compiler
+    project(open62541 C) # Do not look for a C++ compiler endif()
 endif()
 
 # set(CMAKE_VERBOSE_MAKEFILE ON)
@@ -59,7 +59,7 @@ MESSAGE(STATUS "open62541 Version: ${OPEN62541_VERSION}")
 # Architecture #
 ################
 
-set(UA_ARCHITECTURES "none" "posix" "win32")
+set(UA_ARCHITECTURES "none" "posix" "win32" "zephyr")
 set(UA_ARCHITECTURE "" CACHE STRING "Architecture to build open62541 for")
 SET_PROPERTY(CACHE UA_ARCHITECTURE PROPERTY STRINGS "" ${UA_ARCHITECTURES})
 
@@ -76,6 +76,8 @@ if("${UA_ARCHITECTURE}" STREQUAL "posix")
     set(UA_ARCHITECTURE_POSIX 1)
 elseif("${UA_ARCHITECTURE}" STREQUAL "win32")
     set(UA_ARCHITECTURE_WIN32 1)
+elseif("${UA_ARCHITECTURE}" STREQUAL "zephyr")
+    set(UA_ARCHITECTURE_ZEPHYR 1)
 endif()
 
 #################
@@ -512,12 +514,14 @@ function(add_cc_flag CC_FLAG)
 endfunction()
 
 if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
-    check_add_cc_flag("-std=c99")   # C99 mode
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-std=c99")   # C99 mode
+    endif()
     check_add_cc_flag("-pipe")      # Avoid writing temporary files (for compiler speed)
-    check_add_cc_flag("-Wall")      # Warnings
-    check_add_cc_flag("-Wextra")    # More warnings
-    check_add_cc_flag("-Wpedantic") # Standard compliance
-    if(UA_FORCE_WERROR)
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wall")      # Warnings
+        check_add_cc_flag("-Wextra")    # More warnings
+        check_add_cc_flag("-Wpedantic") # Standard compliance
         check_add_cc_flag("-Werror") # All warnings are errors
         check_add_cc_flag("-Wunused-command-line-argument") # Warning for command line args instead of error
     endif()
@@ -528,7 +532,10 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
     check_add_cc_flag("-Wno-maybe-uninitialized") # too many false positives
 
     # Use a strict subset of the C and C++ languages
-    check_add_cc_flag("-Wc++-compat")
+
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wc++-compat")
+    endif()
 
     # Check that format strings (printf/scanf) are sane
     check_add_cc_flag("-Wformat")
@@ -536,19 +543,35 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
     check_add_cc_flag("-Wformat-nonliteral")
 
     # Check prototype definitions
-    check_add_cc_flag("-Wmissing-prototypes")
-    check_add_cc_flag("-Wstrict-prototypes")
-    check_add_cc_flag("-Wredundant-decls")
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wmissing-prototypes")
+    endif()
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wstrict-prototypes")
+    endif()
+
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wredundant-decls")
+    endif()
 
     check_add_cc_flag("-Wuninitialized")
     check_add_cc_flag("-Winit-self")
-    check_add_cc_flag("-Wcast-qual")
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wcast-qual")
+    endif()
     check_add_cc_flag("-Wstrict-overflow")
-    check_add_cc_flag("-Wnested-externs")
+
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wnested-externs")
+    endif()
     check_add_cc_flag("-Wmultichar")
-    check_add_cc_flag("-Wundef")
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-Wundef")
+    endif()
     check_add_cc_flag("-fno-strict-aliasing") # fewer compiler assumptions about pointer types
-    check_add_cc_flag("-fexceptions") # recommended for multi-threaded C code, also in combination with C++ code
+    if(NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
+        check_add_cc_flag("-fexceptions") # recommended for multi-threaded C code, also in combination with C++ code
+    endif()
 
     # Generate position-independent code for shared libraries (adds a performance penalty)
     if(BUILD_SHARED_LIBS)
@@ -581,7 +604,7 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
             check_add_cc_flag("-fstack-protector-strong") # more performant stack protector, available since gcc 4.9
             check_add_cc_flag("-fstack-clash-protection") # increased reliability of stack overflow detection, available since gcc 8
             # future use (control flow integrity protection)
-            if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD")
+            if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD" AND NOT ${UA_ARCHITECTURE} STREQUAL "zephyr")
                 check_add_cc_flag("-mcet")
                 check_add_cc_flag("-fcf-protection")
             endif()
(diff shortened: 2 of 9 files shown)


jpfr:
That looks good.
Thanks for the contribution!

- Please accept the project CLA here in the PR.
- Is there a way to build and test against the Zephyr API in the CI?
I would like to avoid a full-blown Qemu simulation. But maybe there is something lightweight.
That would protect us from future regressions.

xR3b0rn:
@jpfr building yes. Running idk. the first thing that comes into my mind is to run the binary in a qemu environment.

xR3b0rn:
@jpfr I figured out, it is possible to build native zephyr executables, which bridge between zephyr OS and native POSIX.

jpfr:
Yes. This looks promising:

https://docs.zephyrproject.org/latest/boards/native/native_sim/doc/index.html#native-sim

xR3b0rn:
Any ideas on what and how tests should run inside the native-sim?

jpfr:
Ideally every unit test becomes a static binary against the zephyr shim and can run standalone.

xR3b0rn:
I.e. every test in the [tests](https://github.com/open62541/open62541/tree/master/tests) folder?

jpfr:
Nice. A few builds are still red.
Please also click the CLA agreement in the CI checks.

jpfr:
> Since, I don't have access to all kind of architectures, reproducing CI/CD errors on my local PC isn't always straightforward, so I rely on the pipeline.

That's okay. We all do it as well.

xR3b0rn:
I'm not asking for any high confidential or risky access, I'm only asking for not having to wait for approval of the workflows each time. Due to the nature of this PR having to wait makes developing more difficult than it should be.

jpfr:
Normally the CI launches automatically once you have a PR landed to the repo.
Let me check if I can add you to some whitelist...

kanavin:
Can I also chime in here? We (Linutronix) are super interested in Zephyr support, and were considering how to go about it, but you beat us to it, thanks! :)

What isn't clear is how was this built and tested, so that someone can try the code? Can you document that?

My understanding is this:

Typically Zephyr libraries have zephyr/KConfig that sets possible options (including a top level 'enable the whole thing' option), and a zephyr/CMakeLists.txt that builds the code according to these options using zephyr-specific cmake definitions, for example:

https://gitlab.com/systerel/S2OPC/-/tree/master/zephyr?ref_type=heads

These files can be maintained in the component tree under zephyr/, or in zephyr tree itself under modules/, but they have to exist.

xR3b0rn:
@kanavin I can share my CMakeLists.txt of my Zephyr project:
```
cmake_minimum_required(VERSION 3.20.0)

project(open62541-zephyr-test)

find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})

set(UA_ENABLE_DISCOVERY_SEMAPHORE OFF CACHE STRING "" FORCE)
set(UA_ARCHITECTURE "zephyr" CACHE STRING "" FORCE)
set(UA_MULTITHREADING 0 CACHE STRING "" FORCE)
set(UA_NAMESPACE_ZERO "MINIMAL" CACHE STRING "" FORCE)
set(UA_ENABLE_PUBSUB OFF CACHE STRING "" FORCE)
set(UA_ENABLE_PUBSUB_INFORMATIONMODEL OFF CACHE STRING "" FORCE)
set(UA_ENABLE_SUBSCRIPTIONS_EVENTS OFF CACHE STRING "" FORCE)
set(UA_ENABLE_SUBSCRIPTIONS OFF CACHE STRING "" FORCE)
set(UA_ENABLE_HISTORIZING OFF CACHE STRING "" FORCE)
set(UA_ENABLE_JSON_ENCODING OFF CACHE STRING "" FORCE)

add_subdirectory(open62541)

include(FetchContent)
FetchContent_Declare(
  open62541
  GIT_REPOSITORY https://github.com/primes-gmbh/open62541.git 
  GIT_TAG        662906b5fc4bd3de234884782ce8aaee2672014c
)
FetchContent_MakeAvailable(open62541)

target_sources(app PRIVATE src/main.c)

target_link_libraries(app PUBLIC open62541)
```
With this CMakeLists.txt you are able to build the minimal server example: https://github.com/open62541/open62541/blob/master/examples/server.cpp

@jpfr Thanks for the feedback. Of course I will revert the formattings, I didn't apply them by intention. Regarding the architectural changes, the moment I started, I saw, that for any reason you put win32 and posix support into the posix_eventloop files. Since, the Zephyr API is more similar to posix than win32, I simple sticked to that. After seeing all the problems, I am not sure if I would  go down the same path again. I found several incosistencies and problems in the CMakeLists.txt, which I tried to fix on the fly to make the zephyr port work.

From my point view, there are two kind of non zephyr related refactorings I did:
1. Changes in the top level CMakeLists.txt
2. Changes in the C code (in the `eventloop_posix_*`-files)

## Problems with the old CMakeLists.txt
1. The old version of the CMakeListst.txt added compiler flags to `CMAKE_C_FLAGS`, what will also add them to the top level project which is including the open62541 library, what is critical for flags such as `-std=c99`.
2. After adding the flags to the target instead, I ran into another problem: Your tests depend on the changes to the `CMAKE_C_FLAGS` variables as well. Therefor, I had to change the signature of the `*add_cc_flag` functions, to also take an argument for setting the visibility, to make linking targets inherit some of the compiler flags.

## Problems with the C code
1. Zephyr's API is very similar to posix, however, the POSIX API functions and types are prefixed with `zsock_`,. Since, I decided to put the zephyr code into the `eventloop_posix*`-files, I had to fix some of the incosistencies. E.g., you defined abstraction macros for `getsockopt`, but never used them.


My question is, will this PR be merged if I am able to fix the pipelines, or should I pull the code apart (e.g. creating `zephyr_evenloop*`-files) and create another PR?

I wanted to make it as clean as possible, while sticking to the projects conventions.

kanavin:
> @kanavin What is the advantage of providing a Zephyr folder with a Kconfig file if I can manage everything with CMake?

It's pretty much a standard requirement for zephyr integration and adding external code to one's zephyr app:
https://docs.zephyrproject.org/latest/develop/modules.html

You can't reasonably expect anyone to write their own custom cmake file for compiling with a 3rd party library; the expectation is that one can enable CONFIG_THINGY=y in their project file, and the rest will happen automagically.

xR3b0rn:
@jpfr this is how it looks like:
grafik
> 4 workflows awaiting approval

xR3b0rn:
@jpfr How am I able to set configRxBufSize? Setting it via 

> > @kanavin What is the advantage of providing a Zephyr folder with a Kconfig file if I can manage everything with CMake?
> 
> It's pretty much a standard requirement for zephyr integration and adding external code to one's zephyr app: https://docs.zephyrproject.org/latest/develop/modules.html
> 
> You can't reasonably expect anyone to write their own custom cmake file for compiling with a 3rd party library; the expectation is that one can enable CONFIG_THINGY=y in their project file, and the rest will happen automagically.

Well, I can make the options above default for zephyr targets, then all you have to do is FetchContent. What is harder than this?

The example you provided from Systerel's S2OPC looks like a quick and dirty solution to me. They basically copied their normal CMakeLists.txt logic into zephyr/CMakeLists.txt and removed all checks and other modifications, which are done in their main CMakeLists.txt configuration workflow. It is not possible to configure any of the cmake-variables of S2OPC in their CMakeLists.txt for zephyr ([here](https://gitlab.com/systerel/S2OPC/-/blob/master/samples/embedded/cli_pubsub_server/src/cli_pubsub_server.c?ref_type=heads). The way they use cmake for zephyr is also not very cmake idiomatic (going up and down in folder hierarchy).

xR3b0rn:
I think the picolibc library is handling zephyr much cleaner: https://github.com/picolibc/picolibc/tree/main/zephyr

kanavin:
There's also a minimal, clean example module and an app using it:

https://github.com/yashi/module-sample
https://github.com/yashi/module-app

I'm just trying to convince you that modules is the standard way to integrate 3rd party libraries and build them into apps in zephyr. Without it zephyr support in open62541 isn't complete. Whether s2opc has done it horribly is a different issue.

xR3b0rn:
Just added a `zephyr/` folder to the PR. The project now can be fetched the usual way using west.yml. The library can be fully configured using Kconfig variables. An example configuration may look like this:
```
...
CONFIG_OPEN62541=y
CONFIG_OPEN62541_ENABLE_DISCOVERY_SEMAPHORE=n
CONFIG_OPEN62541_MULTITHREADING=0
CONFIG_OPEN62541_NAMESPACE_ZERO="MINIMAL"
CONFIG_OPEN62541_ENABLE_PUBSUB=n
CONFIG_OPEN62541_ENABLE_PUBSUB_INFORMATIONMODEL=n
CONFIG_OPEN62541_ENABLE_SUBSCRIPTIONS_EVENTS=n
CONFIG_OPEN62541_ENABLE_SUBSCRIPTIONS=n
CONFIG_OPEN62541_ENABLE_HISTORIZING=n
CONFIG_OPEN62541_ENABLE_JSON_ENCODING=n
```
As for now, open62541 requires a working loopback interface for the self pipe trick (if you have no loopback interface configured in zephyr, open62541 will hang while setting up the self pipe). In theory, zephyr's kpipe could be used for this purpose. However, the problem with kpipe is, `select` won't work what makes the self pipe trick not working anymore. So, I sticked to the solution used on windows which uses a normal socket instead.

include/open62541/plugin/eventloop.h
```diff
@@ -750,6 +750,13 @@ UA_ConnectionManager_new_MQTT(const UA_String eventSourceName);
 UA_EXPORT UA_InterruptManager *
 UA_InterruptManager_new_POSIX(const UA_String eventSourceName);
 
+#elif defined(UA_ARCHITECTURE_ZEPHYR)
```
jpfr, line 753: This header is for the plugin interface definition.
Please move the plugin implementation to /plugins/include/...

zephyr/Kconfig
```diff
@@ -0,0 +1,197 @@
+menuconfig OPEN62541
```
jpfr, line 1: We try to keep the top-level directory clean.
Please move this to /arch/zephyr.
Also the eventloop implementation should go to /arch/zephyr.
We will then move the existing eventloop to /arch/posix.

arch/eventloop_zephyr/eventloop_zephyr.h
```diff
@@ -0,0 +1,243 @@
...
+#ifndef UA_EVENTLOOP_Zephyr_H_
+#define UA_EVENTLOOP_Zephyr_H_
+
+#include <open62541/config.h>
+#include <open62541/plugin/eventloop.h>
+
+#if defined(UA_ARCHITECTURE_ZEPHYR)
+
+#include <errno.h>
+
```
jpfr, line 20: This file contains a lot of internals.
I would not include it from the config_default plugin.
Rather include the header in /plugins/include/... that only defines the constructor function for the eventloop and the tcp layer.

arch/eventloop_zephyr/eventloop_zephyr.h
```diff
@@ -0,0 +1,243 @@
...
+    UA_DelayedCallback *delayedCallbacks;
+
+    /* Flag determining whether the eventloop is currently within the
+     * "run" method */
+    volatile UA_Boolean executing;
+
+    UA_RegisteredFD **fds;
+    size_t fdsSize;
+
+    /* Self-pipe to cancel blocking wait */
```
jpfr, line 179: If you only need the self-pipe to abort an (endless) call to poll/select, Zephyr has so-called event-fd.
https://docs.zephyrproject.org/latest/doxygen/html/posix_2sys_2eventfd_8h.html

They event-fd can also be listened on with poll/select.
That will have less overhead and is generally cleaner than sending bits over a localhost udp connection.

jpfr (commented):
This looks very clean and complete. I like it!

Comments inline.

It would be great to have a README.md in the /arch/zephyr folder.
With the basic steps to get this running with native_sim in a Linux environment.
This then empowers all the other developers to also test against the Zephyr target.

include/open62541/plugin/eventloop.h
jpfr: Scratch that. The /arch things are correct in this header.

zephyr/Kconfig
xR3b0rn: Zephyr requires the two files `Kconfig` and `module.yml` at this specific location to make the library an zephyr module: https://docs.zephyrproject.org/latest/develop/modules.html

arch/eventloop_zephyr/eventloop_zephyr.h
xR3b0rn: Nice hint. Will check this out.

xR3b0rn:
First, thanks for the feedback!

> It would be great to have a README.md in the /arch/zephyr folder. With the basic steps to get this running with native_sim in a Linux environment. This then empowers all the other developers to also test against the Zephyr target.

I could create a `hello-world` sample project and put it somewhere under `samples`?

jpfr:
> I could create a hello-world sample project which builds out-of-the box for native_sim and put it somewhere under samples?

Is it possible to import the config you provided into cmake and just build **all** of the examples and unit tests into static binaries?

That should be possible with this sim_native target.

Then we can integated nicely into our CI pipeline and ensure nothing breaks over time.

xR3b0rn:
> Is it possible to import the config you provided into cmake and just build all of the examples and unit tests into static binaries?
I think the problem here is, you need some boilerplate code to make basic examples and tests work with zephyr. For reference [check](https://gitlab.com/systerel/S2OPC/-/blob/master/samples/embedded/platform_dep/zephyr/src/zephyr_network_init.c?ref_type=heads).

I believe getting all the samples and tests to work will take considerable effort, and unfortunately, I’m unable to dedicate the necessary time to this task at the moment. My hope is that Zephyr's support for open62541 will attract attention from the community, allowing the code to be tested collaboratively. Perhaps we could consider merging the PR and marking Zephyr support as experimental for now? (after resolving your comments)

zephyr/Kconfig
xR3b0rn: Moving `arch/eventloop_zephyr` to `arch/zephyr` is done.

arch/eventloop_zephyr/eventloop_zephyr.h
xR3b0rn: The `eventloop_zephyr.h` defines the macro `UA_LOG_SOCKET_ERRNO_WRAP`, which relies on `errno`. In my opinion, the best approach is to either also remove this macro or stick with the current solution. Any other approach would leave the instructions unclear.

arch/eventloop_zephyr/eventloop_zephyr.h
xR3b0rn: done

kanavin:
Thanks for pushing this forward (and adding module setup :)! 

If building everything is not feasible, then a 'hello world' sample with instructions should at least be provided. Ideally, building it should also be added to existing open62541 CI.

xR3b0rn:
Just added a basic example: https://github.com/primes-gmbh/open62541/tree/master/examples/zephyr/server

@kanavin

kanavin:
> Just added a basic example: https://github.com/primes-gmbh/open62541/tree/master/examples/zephyr/server
> 

Looks good, thank you! Can the steps be implemented into a github workflow that can be added to existing ones? At least up to the 'west build' point, if setting up for running the executable and checking its output is too complicated.

xR3b0rn:
@kanavin I think so. I would base this CI on the image from S2OPC:
```
podman run -it registry.gitlab.com/systerel/s2opc/zephyr_build:v3.6.0-b@sha256:13069418dd7b0317f93f02986b01f285f0f66956670f2fecded9850ace85d734 /bin/bash
```

jpfr:
I cleaned this up a bit: https://github.com/open62541/open62541/pull/6792
You are of course still author of the commits.
Please build on top of that and add the example + Readme.
Then we can merge.

xR3b0rn:
@jpfr Am I able to commit to this fork, or should I fork and create a pull request to your fork?

kanavin:
> @jpfr Am I able to commit to this fork, or should I fork and create a pull request to your fork? The easies thing could be that you just merge this branch into yours (you are missing two commits). A basic example is already pushed to my fork.

Another option is that you rebase this PR on top of that branch, adding only the difference on top.

xR3b0rn:
@jpfr By moving the example files, you broke the example. Is there any reason why you don't want to have a new folder for the example? I named the example after the file I based my test on (server.cpp). So, in the future, additional Zephyr exampels can be added.

xR3b0rn:
@jpfr By moving the files, you broke the example.

jpfr:
Done. Thanks for the contribution!

PR #6742, opened 2024-10-02
[REVIEW] Add API functions for managing the server's trustlist

NoelGraf:


include/open62541/plugin/certificategroup.h
```diff
@@ -80,6 +80,18 @@ UA_EXPORT UA_StatusCode
...
                                  const UA_ByteString *privateKey);
 
+/* Determines if the Certificate Revocation List (CRL) provided belongs to the given certificate.
+ *
+ * @param certificate The byte string representing the certificate.
+ * @param crl The byte string representing the CRL.
+ * @return UA_STATUSCODE_GOOD if the CRL belongs to the certificate,
+ *         UA_STATUSCODE_BADNOMATCH if the CRL does not belong to the certificate,
+ *         UA_STATUSCODE_BADINTERNALERROR if an error occurs.
+ */
```
jpfr, line 90: You check whether certificate and crl have the same issuer.
That is not the same as matching against the (exact) certificate and can be misread.
Also make sure to mention that we do not check whether the certificate has been revoked by the crl.
That’s another way the intention of the method could be misinterpreted.

src/server/ua_server.c
```diff
@@ -693,6 +693,238 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+        return &server->config.secureChannelPKI;
+    }
+    if(UA_NodeId_equal(&certificateGroupId, &defaultUserTokenGroup)) {
+        return &server->config.sessionPKI;
+    }
+    return NULL;
+}
+
+static UA_StatusCode
+checkCrlMatching(const UA_ByteString *certificates, size_t certificatesSize,
```
jpfr, line 729: This crl check is performed only when we add certificates at runtime.
Not when we load them initially into the SecurityGroup (when the server config is generated).
Also the client doesn’t have these checks.

Can we unify that? So that the same rules (and overrides) are used everywhere….

src/server/ua_server.c
```diff
@@ -693,6 +693,238 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+
+    UA_TrustListDataType oldTrustList;
+    UA_TrustListDataType_init(&oldTrustList);
+    UA_ByteString *crlsOld;
+    size_t crlsOldSize = 0;
+    UA_Boolean fullMatchRequired = false;
+
+    if(isTrustedCertificates) {
+        oldTrustList.specifiedLists = UA_TRUSTLISTMASKS_TRUSTEDCRLS;
+        certGroup->getTrustList(certGroup, &oldTrustList);
```
jpfr, line 850: Duplicate logic in both cases.

src/server/ua_server.c
```diff
@@ -693,6 +693,238 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+        fullMatchRequired = true;
+    }
+
+    /* There is a maximum of one CRL for each certificate.
+     * Therefore, the maximum number of CRLs to be removed is equal to the number of certificates to be removed. */
+    UA_ByteString *crls = (UA_ByteString*)UA_Array_new(certificatesSize, &UA_TYPES[UA_TYPES_BYTESTRING]);
+    size_t crlsSize = 0;
+
+    for(size_t i = 0; i < crlsOldSize; i++) {
+        for(size_t j = 0; j < certificatesSize; j++) {
```
jpfr, line 868: Break already when equality is achieved.
Otherwise you have an overflow access below.

src/server/ua_server.c
```diff
@@ -693,6 +693,238 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+            if(crlsSize > certificatesSize)
+                break;
+            if(UA_CertificateUtils_ckeckCrlMatch(&certificates[j], &crlsOld[i]) != UA_STATUSCODE_GOOD)
+                continue;
+            crls[crlsSize] = crlsOld[i];
+            crlsSize++;
+        }
+        if(crlsSize > certificatesSize) {
```
jpfr, line 876: Add logging.
The error code is impossible to understand without using a debugger to find where it comes from.

src/server/ua_server.c
```diff
@@ -693,6 +693,238 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+            UA_TrustListDataType_clear(&oldTrustList);
+            UA_free(crls);
+            return UA_STATUSCODE_BADINTERNALERROR;
+        }
+    }
+
```
jpfr, line 882: Explain this step in a comment.
You require every certificate that is removed to also have a crl that gets removed.
If you do it in the loop above already, then you can log exactly which certificate/ issuer has a problem.

also beware that you can have multiple certificates from be same issuer.
They then share the same crl.

include/open62541/plugin/certificategroup.h
```diff
@@ -80,6 +80,22 @@ UA_EXPORT UA_StatusCode
...
+ *
+ * @param certificate The byte string representing the certificate.
+ * @param crl The byte string representing the CRL.
+ * @return UA_STATUSCODE_GOOD if the certificate is the issuer of the CRL,
+ *         UA_STATUSCODE_BADNOMATCH if the certificate is not the issuer,
+ *         UA_STATUSCODE_BADCERTIFICATEUSENOTALLOWED If the certificate is not a CA certificate,
+ *         UA_STATUSCODE_BADINTERNALERROR if an error occurs during processing.
+ */
+UA_EXPORT UA_StatusCode
+UA_CertificateUtils_ckeckCrlMatch(const UA_ByteString *certificate,
```
mcpat, line 96: Typo: "ckeckCrlMatch" -> "checkCrlMatch"

include/open62541/server.h
```diff
@@ -1909,6 +1909,45 @@ UA_Server_createSigningRequest(UA_Server *server,
...
+ */
+UA_StatusCode UA_EXPORT
+UA_Server_addCertificates(UA_Server *server,
+                          const UA_NodeId certificateGroupId,
+                          UA_ByteString *certificates,
+                          size_t certificatesSize,
+                          UA_ByteString *crls,
+                          size_t crlsSize,
+                          const UA_Boolean isTrustedCertificates,
+                          const UA_Boolean appendCertificates);
```
mcpat, line 1933: Should "isTrustedCertificates" be renamed to "areTrustedCertificates"?

And also just a note, not a change request:
The name "addCertificates" hints that certificates are added = appended to some collection. Still there is a flag to choose between appending and replacing. My initial thought was to rename the function to "setCertificates" but this feels even worse.

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -895,6 +895,60 @@ UA_CertificateUtils_ckeckKeyPair(const UA_ByteString *certificate,
...
+
+    mbedtls_x509_dn_gets(certSubject, sizeof(certSubject), &cert.subject);
+    mbedtls_x509_dn_gets(crlIssuer, sizeof(crlIssuer), &certRevocationList.issuer);
+
+    if(strncmp(certSubject, crlIssuer, MBEDTLS_X509_MAX_DN_NAME_SIZE) == 0) {
+        retval = UA_STATUSCODE_GOOD;
+    } else {
+        retval = UA_STATUSCODE_BADNOMATCH;
+    }
+
```
mcpat, line 942: The signature validation is missing, right?

src/server/ua_server.c
mcpat: > also beware that you can have multiple certificates from be same issuer. They then share the same crl.

This does not sound right. Certificates with the same issuer usually have distinct private keys. How can you tell that a CRL really belongs to a specific CA if you do not check the signature of the CRL?

src/server/ua_server.c
```diff
@@ -693,6 +693,242 @@ UA_Server_removeCallback(UA_Server *server, UA_UInt64 callbackId) {
...
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    size_t foundMatchCount = 0;
+    /* For each certificate to be removed, the corresponding CRL must also be deleted.
+     * If the certificate is not a CA certificate, the search can be skipped.
+     * However, if it is a CA certificate, all CRLs issued by that certificate must be removed.
+     * A CA certificate can also have multiple CRLs. */
+    for(size_t i = 0; i < certificatesSize; i++) {
+        UA_Boolean foundMatch = false;
+        for(size_t j = 0; j < crlsOldSize; j++) {
+            retval = UA_CertificateUtils_ckeckCrlMatch(&certificates[i], &crlsOld[j]);
```
mcpat, line 868: I understand the intention to process everything as high up in the stack as possible to do it only at one location.
But this specific check here now involves parsing the same CRLs multiple times. There are already also other cases regarding the certificate validation where a piece of work internally requires doing conversions multiple times. It feels like offloading those tasks to the specific crypto layers would not be a bad solution to at least reduce the duplication of conversion work.

I may also add that in an old discussion we also had the topic of what the removeFromTrustList callback should be responsible for. And back then it had the responsibility to also remove all CRLs as required (which basically meant that the checking part should be offloaded into the crypto layer).

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -132,6 +132,121 @@ MemoryCertStore_getRejectedList(UA_CertificateGroup *certGroup, UA_ByteString **
...
+        retval = UA_STATUSCODE_BADINTERNALERROR;
+        goto cleanup;
+    }
+
+    /* Check if the certificate is a CA certificate.
+     * Only a CA certificate can have a CRL. */
+    if(!cert.ca_istrue) {
+        retval = UA_STATUSCODE_BADCERTIFICATEUSENOTALLOWED;
+        goto cleanup;
+    }
```
mcpat, line 165: Please do this check with the mbedtlsCheckCA function below.

include/open62541/server.h
jpfr: That could be simply "isTrusted".

include/open62541/server.h
```diff
@@ -1909,6 +1909,45 @@ UA_Server_createSigningRequest(UA_Server *server,
...
+ *
+ * @param server The server object
+ * @param certificateGroupId The NodeId of the certificate group from which certificates will be removed
+ * @param certificates The certificates to be removed
+ * @param certificatesSize The number of certificates
+ * @param isTrustedCertificates Indicates whether the certificates are being removed from the trusted list or the issuer list
+ * @return ``UA_STATUSCODE_GOOD`` on success
+ */
+UA_StatusCode UA_EXPORT
+UA_Server_removeCertificates(UA_Server *server,
```
jpfr, line 1945: Will this also remove the CRL?
isTrustedCertificates --> isTrusted

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -132,6 +134,119 @@ MemoryCertStore_getRejectedList(UA_CertificateGroup *certGroup, UA_ByteString **
...
+checkCrlMatch(const UA_ByteString *certificate, const UA_ByteString *crl) {
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+
+    mbedtls_x509_crt cert;
+    mbedtls_x509_crt_init(&cert);
+
+    mbedtls_x509_crl certRevocationList;
+    mbedtls_x509_crl_init(&certRevocationList);
+
+    UA_ByteString data1 = UA_mbedTLS_CopyDataFormatAware(certificate);
```
jpfr, line 147: I don't think you need to copy.
The `crt_parse` takes the length as an extra argument.
So it does not rely on a \0 at the end.

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -132,6 +134,119 @@ MemoryCertStore_getRejectedList(UA_CertificateGroup *certGroup, UA_ByteString **
...
+}
+
+static UA_StatusCode
+checkCrlList(UA_CertificateGroup *certGroup, const UA_ByteString *certificate,
+             const UA_ByteString *crlList, const size_t crlListSize,
+             UA_ByteString **crls, size_t *crlsSize) {
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    UA_Boolean foundMatch = false;
+    for(size_t i = 0; i < crlListSize; i++) {
+        retval = checkCrlMatch(certificate, &crlList[i]);
```
jpfr, line 197: You parse and checkCA the certificate in every iteration.
You can do it only once at the beginning.

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -132,6 +134,119 @@ MemoryCertStore_getRejectedList(UA_CertificateGroup *certGroup, UA_ByteString **
...
+        goto cleanup;
+    }
+
+    char certSubject[MBEDTLS_X509_MAX_DN_NAME_SIZE];
+    char crlIssuer[MBEDTLS_X509_MAX_DN_NAME_SIZE];
+
+    mbedtls_x509_dn_gets(certSubject, sizeof(certSubject), &cert.subject);
+    mbedtls_x509_dn_gets(crlIssuer, sizeof(crlIssuer), &certRevocationList.issuer);
+
+    if(strncmp(certSubject, crlIssuer, MBEDTLS_X509_MAX_DN_NAME_SIZE) == 0) {
```
jpfr, line 175: Do  we need to check if the crl was also signed by the issuer?
If yes, then we should do that at the beginning when we load the CRL first.
So we don't have to redo taht check...

jpfr (commented):
Comments inline for mbedTLS.
Most will also hold for OpenSSL.

plugins/crypto/mbedtls/certificategroup.c
NoelGraf: The description of the function parameters includes the following.
"buflen – The size of buf, including the terminating NULL byte in case of PEM encoded data."

plugins/crypto/mbedtls/certificategroup.c
NoelGraf: Good point! We should also check that the CRL has been signed by the issuer, and it's best to include this check when adding it. This will be implemented in a separate PR as part of the validation stage implementation.

plugins/crypto/mbedtls/certificategroup.c
NoelGraf: I've added a helper function for mbedtls, similar to the one in OpenSSL, to load certificate files and CRLs. The function identifies the format of the input and only copies the byte string if it's in PEM format.

include/open62541/server.h
jpfr: Mention that in the docs.

plugins/crypto/mbedtls/certificategroup.c
```diff
@@ -132,6 +134,108 @@ MemoryCertStore_getRejectedList(UA_CertificateGroup *certGroup, UA_ByteString **
...
+MemoryCertStore_getCertificateCrls(UA_CertificateGroup *certGroup, const UA_ByteString *certificate,
+                                   UA_ByteString **crls, size_t *crlsSize) {
+    /* Check parameter */
+    if(certGroup == NULL || certificate == NULL || crls == NULL) {
+        return UA_STATUSCODE_BADINTERNALERROR;
+    }
+
+    MemoryCertStore *context = (MemoryCertStore *)certGroup->context;
+
+    UA_StatusCode retval = mbedtlsFindCrls(certGroup, certificate,
```
jpfr, line 225: Is it possible that - by mistake or by design - a CRL is both in the issuer and the revocation list?
For example when an issuer gets moved to the trustlist.
We should be concatenating the two lists.

But we can also simplify overall.
Please keep a single CRL list for both issuers and trustlist.
The functionality is identical as we map to issuer names anyhow.

jpfr (commented):
Comments inline

plugins/crypto/mbedtls/certificategroup.c
NoelGraf: Based on my understanding of the specification, it's possible for a CA certificate to appear on both the Issuer and Trusted lists. The Trusted list contains certificates that are explicitly trusted, which may include CA certificates. The Issuer list, on the other hand, includes certificates necessary for validation purposes.

For this reason, I believe it’s important to maintain a clear distinction between trusted and issuer CRLs. If a CA certificate is removed from the trusted list, only the CRLs associated with it should be removed from the trusted list, while the Issuer list should remain unaffected.

This separation wouldn’t be feasible if there were only one combined CRL list. Additionally, the TrustListDataType is designed to support this distinction.

NoelGraf: I have modified the getCertificateCrls function by introducing the isTrusted parameter, allowing you to specify which list to reference. Since a certificate may appear in both lists, this parameter ensures the correct list is selected.

Additionally, a certificate can also have multiple CRLs, and the trustedCrls list and the issuerCrls list may each contain different CRLs for the same certificate.

PR #6746, opened 2024-10-02
fix(plugins): More thorough validation that the issuer is a valid CA

jpfr:


mlang-de:
@jpfr Against which CTT Version and which CTT Scripts version this shall be compliant?

jpfr:
We completed the certification for the v1.4 branch against the v1.04 CTT at the time.
Please try with the latest v1.3.15 or v1.4.8 release.

The v1.4.8 has only a single issue remaining wrt certificates in the CTT on my box.
And that is "just" a bad statuscode that expects a different statuscode.
(And where I suspect the CTT are at fault, discussions are pending.)

mlang-de:
In combination with https://github.com/open62541/open62541/pull/6902 only one script left:
033.js (fails with BadCertificateTimeInvalid)
So I guess this is the TC in ongoing discussion.

jpfr:
Yes. When we mask that return code, then other tests break instead.

jpfr:
Also, we would like to also match the authority key identifier for the issuing certificate. But the ctt don’t let us…

PR #6774, opened 2024-10-16
fix(el): Treat some accept() errors as temporary

teijokinnunen:
If accept() fails, some of the error codes indicate failure with the incoming connection, not the server socket. Do not kill the server socket if one of those error codes are detected.


jpfr:
Nice, thanks for the fix.

PR #6788, opened 2024-10-24
fix(pubsub): Only check message timeout if message is valid

keba-estr:
Has already been resolved on the master branch with commit c4686f6351ec192f75dadb5ac570b07b3fbda5d3.


jpfr:
👍

PR #6791, opened 2024-10-25
fix(server) check if discovery url is in endpoint description

andreasebner:
Skip copying endpoint url into endpoint description within the GetEndpoints Service if the endpoint url is already contained in the discovery url config of the server.

### DIFF
--- a/src/server/ua_services_discovery.c
+++ b/src/server/ua_services_discovery.c
@@ -9,6 +9,7 @@
  *    Copyright 2016-2017 (c) Stefan Profanter, fortiss GmbH
  *    Copyright 2017 (c) frax2222
  *    Copyright 2017 (c) Mark Giraud, Fraunhofer IOSB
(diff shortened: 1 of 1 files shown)


src/server/ua_services_discovery.c
```diff
@@ -432,12 +433,23 @@ setCurrentEndPointsArray(UA_Server *server, const UA_String endpointUrl,
...
+                    if(UA_String_equal(&endpointUrl,
+                                       &ed->server.discoveryUrls[n])) {
+                        foundEndpointUrl = true;
+                        break;
+                    }
+                }
+                if(!foundEndpointUrl) {
+                    /* Mirror back the requested EndpointUrl and also add it to the
+                    * array of discovery urls */
+                    retval |= UA_String_copy(&endpointUrl, &ed->endpointUrl);
```
maikrou, line 448: `&ed->endpointUrl` is null and not set if the endpoint url is already found (should always be set)

PR #6792, opened 2024-10-25
Zephyr support

jpfr:
Cleanups from #6741.


xR3b0rn:
* Moving the files, broke the example
* By removing the zephyr folder, open62541 is no longer a zephyr module. You can check other libraries, such as [picolibc](https://github.com/picolibc/picolibc/tree/main) which provide a zephyr folder.

xR3b0rn:
@jpfr You even pushed your bugs under my name... I'm a little bit offended by that. Please make this right.

jpfr:
We can provide a shell script to manually move folders and generate symlinks before the zephyr build.

We plan in supporting multiple embedded os.
It is not possible for any embedded os to pollute the top-level folder.

This is why CI integration is necessary.
When we don’t know what will break the code, then the code will break.

PR #6795, opened 2024-10-25
fix: Fixed Zephyr example and introduced CI

xR3b0rn:
Fixed bugs introduced in the "cleaned up" [PR](https://github.com/open62541/open62541/pull/6792) by reverting some of the changes. The original PR can be found [here](https://github.com/open62541/open62541/pull/6741).


xR3b0rn:
@jpfr I now fixed the example and the Zephyr integration while keeping the adapted folder structure (without the top level zephyr folder). To avoid regressions, I added a CI for building the Zephyr server example.

jpfr:
Nice, looks good.
The Zephyr build throws a few warnings (which we would abort on for the other architectures).
But nothing scary at the moment.

If you want to keep the current commit history, please use the commit line prefixes `fix(arch)`, `feat(ci)`, etc. described here:
https://github.com/open62541/open62541/blob/master/CONTRIBUTING.md

Otherwise I can squash-merge the entire PR as `feat(ci): Add CI build for Zephyr`.

Thanks for putting in the effort.

xR3b0rn:
I don't mind if you squash.

jpfr:
👍

Issue #6805, opened 2024-10-30
Wrong source port in mDNS queries (LDS-ME)


V-Strassheim:
## Description
mDNS queries are sent from the wrong UDP port.

## Background Information / Reproduction Steps

According to mdns [RFC 6762](https://datatracker.ietf.org/doc/html/rfc6762#section-5.2), section 5.2. "Continuous Multicast DNS Querying",
_"A compliant Multicast DNS querier, which implements the rules specified in this document, MUST send its Multicast DNS queries from UDP source port 5353"_

At the moment, sockets for send connections are not bound and are therefore using a random port number. Other clients could ignore the queries. This can be reproduced e.g. with `avahi-browse -r _opcua-tcp._tcp`, the avahi [client-browse-services](https://github.com/avahi/avahi/blob/master/examples/client-browse-services.c) example, which does not show service registered by [server_multicast](https://github.com/open62541/open62541/blob/master/examples/discovery/server_multicast.c) example. Other example [core-browse-services](https://github.com/avahi/avahi/blob/master/examples/core-browse-services.c)  sees the queries and complains about the wrong source port.

Non default used CMake options:
UA_NAMESPACE_ZERO=FULL
UA_BUILD_EXAMPLES=TRUE 
UA_ENABLE_DISCOVERY_MULTICAST=TRUE

## Checklist
 - [x] open62541 Version (release number or git tag): master (8945d6e7e6fa4d831e5f452380f49367eeeaf97a)
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: debian 12
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

V-Strassheim:
Thanks for the tip!
I'm currently looking at how the mDNS library is connected and stumbled across it. Maybe the problem can be avoided if the mdns daemon is used differently, as has already been discussed [here](https://github.com/open62541/open62541/issues/4896#issue-1097662862) .
I'll get back to as soon as I have a good solution proposal.
I also would leave the issue open as a reminder so that it doesn't go down even if the mdnsd change doesn't work.

PR #6806, opened 2024-10-30
[WIP] fix(el): bind udp send connection sockets

V-Strassheim:
Closes https://github.com/open62541/open62541/issues/6805

According to mdns RFC 6762, section 5.2. "Continuous Multicast DNS Querying", "A compliant Multicast DNS querier, which implements the rules specified in this document, MUST send its Multicast DNS queries from UDP source port 5353" At the moment, sockets for send connections are using a random port. Therefore, other mdns clients may ignore the mdns queries. This was observed with the avahi client-browse-services example.

Add bind also to send direction sockets to use the provided port.
WIP, as I'm not sure if this will break other connections.


V-Strassheim:
I am going to test another solution to avoid the problem.

PR #6816, opened 2024-11-01
fix(docs): PubSub build options are now enabled by default

redstrate:
In 4ec3bd92f5761020ce546ddeec2f1de610d02acb these build options were switched on by default, but this was not reflected in the docs yet.


jpfr:
👍

PR #6819, opened 2024-11-01
fix(docs): Fix rendering of the UA_ENABLE_ENCRYPTION options

redstrate:
The list didn't render correctly in Sphinx, nor Github because it was missing a newline. Now it should look correct!


jpfr:
👍

PR #6821, opened 2024-11-02
[Review] fix(pubsub): fix MQTT subscribe

melekhove:
- fix` UA_ReaderGroupConfig` copy function
- use correct data type in example

BTW, why `UA_BrokerDataSetReaderTransportDataType ` is used for transport settings in ReaderGroup while specifications says that subtypes of ReaderGroupTransportDataType should be used?

### 6.2.8.2.1 ReaderGroupDataType
This Structure DataType is used to represent the configuration parameters for ReaderGroups.
The ReaderGroupDataType is formally defined in Table 46.
Table 46 – ReaderGroupDataType structure
|Name|Type|Description|Allow Subtypes|
|-------------------------|----------|-------------------------------------|----------------------|
|ReaderGroupDataType|Structure|Subtype of PubSubGroupDataType defined in 6.2.5.7.||
|TransportSettings|ReaderGroupTransportDataType|Transport mapping specific ReaderGroup parameters. The abstract base type is defined in 6.2.8.2.2. The concrete subtypes are defined in the subclauses for transport mapping specific parameters. If no concrete subtype is defined for the transport mapping, the field shall be null.|True|

### DIFF
--- a/examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
+++ b/examples/pubsub/tutorial_pubsub_mqtt_subscribe.c
@@ -111,8 +111,8 @@ addReaderGroup(UA_Server *server) {
         readerGroupConfig.encodingMimeType = UA_PUBSUB_ENCODING_JSON;
 
     /* configure the mqtt publish topic */
-    UA_BrokerWriterGroupTransportDataType brokerTransportSettings;
-    memset(&brokerTransportSettings, 0, sizeof(UA_BrokerWriterGroupTransportDataType));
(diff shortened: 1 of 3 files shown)


jpfr:
Uh. Thanks.

Two asks:
- Can you make the PR for the 1.4 branch? Or is this only a problem on master?
- We have an MQTT unit test (check_pubsub_mqtt.c ). That should have protected against the problem.
   Please extend the unit test to cover this fix.

melekhove:
Closing since #6850 is merged.

PR #6832, opened 2024-11-04
fix(plugins): Reenable writing rejected certificates to a reject folder

jpfr:


jpfr:
We can discuss semantics.
To me "rejected" can be any of the above.
So it becomes visible which certificates have been tried...

dismirlian:
Sure!

The good thing is that the logs will show WHY it was rejected.

Thanks!!

mcpat:
At least it is a recommendation to add "every valid but untrusted Certificate" (see https://reference.opcfoundation.org/GDS/v105/docs/7.8.3.2) to the rejected list. But it is also stated: "[Servers](https://reference.opcfoundation.org/search/49?t=Servers) only add [Certificates](https://reference.opcfoundation.org/search/49?t=Certificates) to this list that have no unsuppressed validation errors but are not trusted.

PR #6835, opened 2024-11-05
[REVIEW] Allow stream-based writing of the trust list

NoelGraf:
Currently, after a write, the data is decoded directly into a TrustListDataType and applied to the transaction. However, this means that it is not possible to write larger data using multiple write calls.
This will be fixed in this PR. The data to be written is first cached. After a CloseAndUpdate the data is decoded into a TrustListDataType and applied to the transaction.

### DIFF
--- a/src/server/ua_server_ns0_gds.c
+++ b/src/server/ua_server_ns0_gds.c
@@ -15,6 +15,9 @@
 typedef struct UA_FileContext {
     LIST_ENTRY(UA_FileContext) listEntry;
     UA_ByteString file;
+    /* Caches any data to be written using the Write method.
+     * With a CloseAndUpdate, the data to be written is applied to the transaction. */
+    UA_ByteString dataToWrite;
     UA_UInt32 fileHandle;
     UA_NodeId sessionId;
     UA_UInt64 currentPos;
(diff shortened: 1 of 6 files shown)


src/server/ua_server_ns0_gds.c
```diff
@@ -15,6 +15,7 @@
 typedef struct UA_FileContext {
     LIST_ENTRY(UA_FileContext) listEntry;
     UA_ByteString file;
+    UA_ByteString dataToWrite;
```
jpfr, line 18: I think I know what this does.
Knowing is better.
Please a comment to explain the struct member.

src/server/ua_server_ns0_gds.c
```diff
@@ -943,10 +945,34 @@ closeAndUpdateTrustList(UA_Server *server,
     if(!fileContext)
         return UA_STATUSCODE_BADINTERNALERROR;
 
+    if(fileContext->openFileMode != (UA_OPENFILEMODE_WRITE | UA_OPENFILEMODE_ERASEEXISTING))
+        return UA_STATUSCODE_BADINVALIDSTATE;
+
+    UA_CertificateGroup *transactionCertGroup =
+    UA_GDSTransaction_getCertificateGroup(transaction, certGroup);
```
jpfr, line 952: indent

src/server/ua_server_ns0_gds.c
```diff
@@ -943,10 +945,34 @@ closeAndUpdateTrustList(UA_Server *server,
...
+    if(!transactionCertGroup)
+        return UA_STATUSCODE_BADINTERNALERROR;
+
+    UA_TrustListDataType trustList;
+    memset(&trustList, 0, sizeof(UA_TrustListDataType));
```
jpfr, line 957: Decode does the memset internally.

jpfr (commented):
Looks good.
Minor comments only.

PR #6840, opened 2024-11-06
fix(build): Enable building static library on Win32

jpfr:


ernstblechaPT:
Thank you for the quick resolution!

PR #6844, opened 2024-11-07
fix(ci): fix allow reuse of TIME-WAIT sockets

GoetzGoerisch:
* `sudo echo 1 >  /proc/sys/net/ipv4/tcp_tw_reuse` not allowed in CI (permission denied)


jpfr:
Nice 👍

PR #6858, opened 2024-11-09
feature: optional types decoding

electro-logic:
- Add a new UA_ENABLE_TYPES_DECODING advanced option (ON by default)
- Variant types containing Extension Objects can be decoded or not according to the parameter

It's now possible to avoid decoding and retrieve raw Extension Objects bytes when UA_ENABLE_TYPES_DECODING is manually set to OFF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,9 @@ option(UA_ENABLE_XML_ENCODING "Enable XML encoding (EXPERIMENTAL)" OFF)
 option(UA_ENABLE_NODESETLOADER "Enable nodesetLoader public API" OFF)
 option(UA_ENABLE_DATATYPES_ALL "Generate all datatypes for namespace zero (uses more binary space)" ON)
 
+option(UA_ENABLE_TYPES_DECODING "Automatically decode known Extension Objects" ON)
+mark_as_advanced(UA_ENABLE_TYPES_DECODING)
+
 if(UA_INFORMATION_MODEL_AUTOLOAD AND NOT UA_ENABLE_NODESETLOADER AND NOT UA_BUILD_FUZZING)
     set(UA_ENABLE_NODESET_INJECTOR ON)
 endif()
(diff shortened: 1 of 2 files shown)


electro-logic:
PR Update: 

Client Subscription is processing messages in the processNotificationMessage function, silently returning undecoded extension objects. An improvement here can be Logging a warning message.

Latest commit of this PR is manually decoding the message using the new function UA_ExtensionObject_decode if the automatic types decoding is disabled (UA_ENABLE_TYPES_DECODING = 0)

jpfr:
Why? To me it is an advantage to allow “all types” even when the server cannot decide.

electro-logic:
> Why? To me it is an advantage to allow “all types” even when the server cannot decode.

Hi,

This was about the processNotificationMessage() routing specific to subscriptions only, so it's expecting known types only and returning otherwise. By disabling automatic decoding globally the routing needs to "manually" decode the ext.obj.

There are more changes than expected, maybe introducing a method to read a node without the automatic type decoding can affect less existing code than a compile option. I have to prototype a bit about this.

jpfr:
The decoding method takes an options argument.
It is less invasive to a this as a flag to the options.

electro-logic:
I pushed a commit to the PR to fix some issues on the server-side, and to learn when this automatic decoding is needed.

From an user perspective would be great to have an option in the method
`UA_Client_Service_read(UA_Client *client, const UA_ReadRequest request)`
so a specific node can be read without any automatic decoding, but the eventloop logic is making things more complex.

jpfr:
If you need the bytes, you can just reencode with UA_encodeBinary.
You will get out the same bytestring that was sent over the network.

in case that the data type is not known locally, the decoding gives you an opaque extension object. That can also be re-encoded to give you the original bytestring.

jpfr:
We have automated 24/7 fuzzing tests for the decode-encode roundtrip.

So decode -> encode -> decode -> encode.

That two encode steps must produce byte-identical results.
The encoding is unique. For the NodeId there are well-defined thresholds when to use the 2-byte encoding and when not.

Anyway, if the NodeId can be encoded in different ways, then the decoded data is still identical.

After the discussion, I still have not understood which use case might ever need this functionality.

Issue #6860, opened 2024-11-10
[BUG] Type Casting rules


electro-logic:
Hello,

The following code 

https://github.com/open62541/open62541/blob/b7ae6e60eba2d0a7fe5ba947f7d3a5f7956657f8/src/server/ua_subscription_eventfilter.c#L88

should follow the table 123 where order is:

Double, Float, Int64, UInt64, .. (Part 4 Table 123 "Data Precedence Rules")

but the current order is:

DATETIME, DOUBLE, STRING, UINT64, FLOAT, etc..

Do you know why there is a mismatch with the specifications?

My suggestions:
- use the UA_DATATYPEKIND_ constants to avoid issues
- 255 is a "magic number", why not using a constant?
- Same above apply to convertLookupIndex and convertLookup in the same file

This will also help me to better implement a PR for the [issue #6780](https://github.com/open62541/open62541/issues/6780)

Issue #6780, opened 2024-10-22
Improve UA_DATATYPEKIND with NULL value for better consistency


electro-logic:
UA_DataTypeKind doesn't include a NULL value, for example 

UA_DATATYPEKIND_NULL = 0

this make it not consistent with the UA_Variant encoding mask https://reference.opcfoundation.org/Core/Part6/v105/docs/5.1.2#_Ref83387521

Is there a reason why the null value was omitted?
Thank you

electro-logic:
What about adding also few other data types? For example UA_NS0ID_NUMBER, UA_NS0ID_INTEGER and UA_NS0ID_UINTEGER don't have an equivalent UA_DATATYPEKIND

electro-logic:
Hi @jpfr 

I created the PR #6861 to fix this issue, but I need some help in the automatic data type generation to align UA_TYPES_* to UA_DATATYPEKIND_*

Issue #6859, opened 2024-11-10
[Doc] improvement suggestion


electro-logic:
Build instructions for Windows are missing Check (for unit tests).

These are the official instructions https://libcheck.github.io/check/web/install.html#visualstudios, but cloning the Check repository, building and installing the solution seems a more modern approach.

electro-logic:
There are two parameters CHECK_INCLUDE_DIR and CHECK_LIBRARIES to configure before generating the VS project.

CHECK_LIBRARIES is a bit misleading because it's actually the .lib file and NOT the lib folder. 

Another problem is that there are two .lib files: check.lib and checkDynamic.lib

Compiling with UA_ENABLE_ENCRYPTION set to MBEDTLS leads to many errors similar to: 

```
Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	LNK2019	unresolved external symbol __imp___time64 referenced in function _x509_crt_verify_chain	check_ziptree	D:\open62541\build\tests\mbedx509.lib(x509_crt.obj)

Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	LNK2019	unresolved external symbol __imp__fseek referenced in function _mbedtls_entropy_update_seed_file	check_server_callbacks	D:\open62541\build\tests\mbedcrypto.lib(entropy.obj)

Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	LNK2019	unresolved external symbol __except_handler4_common referenced in function __except_handler4	check_ziptree	D:\open62541\build\tests\MSVCRT.lib(chandler4gs.obj)

```

seems there are conflicts with other libraries

electro-logic:
@RolfKal 

I pushed PR #6862 that fix the compilation issue under Windows

Note: CHECK_LIBRARIES should point to check.lib

RolfKal:
> @RolfKal
> 
> I pushed PR #6862 that fix the compilation issue under Windows
> 
> Notes:
> 
> * CHECK_LIBRARIES should point to check.lib
> * Firewall pop-up at every test the first time
> * Examples are fixable in a very similar way, by adding
> 
> ```
> if(MSVC)    
>     add_compile_options(
>         $<$<CONFIG:>:/MD>
>         $<$<CONFIG:Debug>:/MDd>
>         $<$<CONFIG:Release>:/MD>
>     )    
> endif()
> ```
> 
> to examples/cmakelist.txt. I will push a PR soon for examples too.

[...]

PR #6861, opened 2024-11-10
[WIP] Add UA_DATATYPEKIND_NULL to align with Variant types

electro-logic:
This PR add UA_DATATYPEKIND_NULL to align UA_DATATYPEKIND with the Variant types (https://reference.opcfoundation.org/Core/Part6/v105/docs/5.1.2#_Ref83387521) and is related to issue #6780 

Still needs some help to fix automatically generated types.

types_generated.h contains types starting from 0, while the addition of the UA_DATATYPEKIND_NULL should now make them start from 1.

```
#define UA_TYPES_COUNT 388
extern UA_EXPORT UA_DataType UA_TYPES[UA_TYPES_COUNT];

/* Boolean */
#define UA_TYPES_BOOLEAN 0

/* SByte */
#define UA_TYPES_SBYTE 1

```

should become

```
#define UA_TYPES_COUNT 389
extern UA_EXPORT UA_DataType UA_TYPES[UA_TYPES_COUNT];

/* Undefined type */
#define UA_TYPES_NULL 0

/* Boolean */
#define UA_TYPES_BOOLEAN 1

/* SByte */
#define UA_TYPES_SBYTE 2

```

some changes to generate_datatypes.py are probably required, any help is welcome.

This PR is also related to https://github.com/open62541/open62541/issues/6860


jpfr:
The data type kind is a different index from the UA_TYPES array.

the kind index is used mostly internally for jump tables.

You probably can simplify quite a bit of the proposed changes.

electro-logic:
Hi, thank you for the reply.

The issue is that a lot of code (library and user code) is often retrieving the UA_DATATYPE by doing UA_TYPES[kind] where kind is of type UA_DATATYPEKIND.

I think that would be much better to keep the internal values aligned. A simple way to align them would be to add someway a "virtual" hardcoded UA_TYPES_NULL node. What do you think about this direction?

jpfr:
Lookup of the type via the kind only works in some cases. Because one kind is “structure” which maps to many types.
The pattern of looking up the type via the kind should be avoided!!

How would the PR improve the situation?
Because you increase both the kind and the normal array index.

electro-logic:
> Lookup of the type via the kind only works in some cases. Because one kind is “structure” which maps to many types. The pattern of looking up the type via the kind should be avoided!!

Yes, the conversion is done only when the kind value is <= UA_DATATYPEKIND_DIAGNOSTICINFO, in a similar way to ENCODE_BINARY(Variant) / DECODE_BINARY(Variant)

> How would the PR improve the situation? Because you increase both the kind and the normal array index.

Aligning with the variant type avoid the +/-1 conversion in ENCODE_BINARY(Variant) / DECODE_BINARY(Variant) and makes the code cleaner. Is also safer because makes clear when UA_DATATYPEKIND is not initialized.

src/ua_types.c
```diff
@@ -1334,6 +1334,7 @@ copyNotImplemented(const void *src, void *dst, const UA_DataType *type) {
 }
 
 const UA_copySignature copyJumpTable[UA_DATATYPEKINDS] = {
+    NULL,
```
jpfr, line 1337: This should point to a function that returns an error…

src/ua_types.c
```diff
@@ -1437,6 +1438,7 @@ static void nopClear(void *p, const UA_DataType *type) { }
 
 const
 UA_clearSignature clearJumpTable[UA_DATATYPEKINDS] = {
+    NULL,
```
jpfr, line 1441: The same here.

also for binary, JSON and xml encoding.

electro-logic:
Note: I'm still trying to optimize my dev environment for open62541, but unit tests are not working under Windows / Visual Studio 2022 (see https://github.com/open62541/open62541/issues/6859). Some PR may not pass basic checks because of this.

src/ua_types.c
electro-logic: good point, for now I put a NULL to compile and trying to test the happy path as a first step.

jpfr:
If the only advantage is that we can save a +1 or -1 on the indices in the binary decoding, then this is not worth the effort.

I agree it would be an improvement.
But there is just too much risk of either breaking existing user code or overlooking an edge case, leading to bugs.

electro-logic:
Let's re-evaluate this PR for a breaking changes release (ex. v2.0).

jpfr:
:+1

PR #6863, opened 2024-11-12
[Review] feat(server): add ability to purge TCP connections

jepa-hms:
Purge when trying to establish a new connection and the maximum
number of connections has already been reached. Disabled by
default. Enable by setting maxConnections to >0

### DIFF
--- a/src/server/ua_server_binary.c
+++ b/src/server/ua_server_binary.c
@@ -198,6 +198,22 @@ decodeHeaderSendServiceFault(UA_Server *server, UA_SecureChannel *channel,
     return retval;
 }
 
+/* Remove the first (oldest) connection that has no open secureChannel */
+static UA_Boolean
+purgeFirstConnectionWithoutChannel(UA_ConnectionManager *cm,
+                                   UA_BinaryProtocolManager *bpm) {
+    UA_SecureChannel *channel;
+    TAILQ_FOREACH(channel, &bpm->channels, componentEntry) {
+        if(channel->state == UA_SECURECHANNELSTATE_OPEN) {
+            continue;
+        }
+        if(cm->closeConnection(cm, channel->connectionId) == UA_STATUSCODE_GOOD) {
+            return true;
+        }
+    }
+    return false;
+}
+
 
 /*************************/
 /* Process Message Types */
@@ -618,6 +634,9 @@ createServerSecureChannel(UA_BinaryProtocolManager *bpm, UA_ConnectionManager *c
     if(scs->currentChannelCount >= config->maxSecureChannels &&
        !purgeFirstChannelWithoutSession(bpm))
         return UA_STATUSCODE_BADOUTOFMEMORY;
+    if(config->maxConnections && scs->currentChannelCount >= config->maxConnections &&
+       !purgeFirstConnectionWithoutChannel(cm, bpm))
+        return UA_STATUSCODE_BADTCPNOTENOUGHRESOURCES;
 
     /* Allocate memory for the SecureChannel */
     UA_SecureChannel *channel = (UA_SecureChannel*)UA_calloc(1, sizeof(UA_SecureChannel));
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -243,6 +243,8 @@ struct UA_ServerConfig {
     UA_UInt16 maxSessions;
     UA_Double maxSessionTimeout; /* in ms */
 
+    UA_UInt16 maxConnections;
+
     /* Operation limits */
     UA_UInt32 maxNodesPerRead;
     UA_UInt32 maxNodesPerWrite;


jpfr:
All connections are aggregated in the SecureChannels list:
https://github.com/open62541/open62541/blob/7c4ea969590311df5b2c1f30b0a5ce219e479adf/src/server/ua_server_binary.c#L78

I would prefer to do the purging using that.

arch/posix/eventloop_posix_tcp.c
```diff
@@ -8,6 +8,7 @@
 
 #include "open62541/types.h"
 #include "eventloop_posix.h"
+#include "../../src/server/ua_server_internal.h"
```
jepa-hms, line 11: Not a fan of this include but I could not find another way to get access to the SecureChannels list from this file. Is there a better way I'm not seeing?

jpfr: Yes, that include is a no-go. It breaks abstraction boundaries.
Can't you trigger the purging from within the server?

jpfr (commented):
I would fold the two new methods into a single one.
The code will still be ~10 lines overall.

Try to emulate our coding style.
We have curly braces on the same line.

Add a comment on which connection gets selected.
Are we purging the newest or the oldest connection that is not open?
This depends on whether we add new connections to the head or to the tail of the queue.

A new connection immediately gets the state UA_SECURECHANNELSTATE_CONNECTED.
I think we should purge everything where the SecureChannel is not fully open.
Otherwise a malicious actor can just keep TCP connections open and hog all networking resources.

arch/posix/eventloop_posix_tcp.c
jepa-hms: Yes, I managed to move everything to within the server

src/server/ua_server_binary.c
```diff
@@ -618,6 +646,9 @@ createServerSecureChannel(UA_BinaryProtocolManager *bpm, UA_ConnectionManager *c
     if(scs->currentChannelCount >= config->maxSecureChannels &&
        !purgeFirstChannelWithoutSession(bpm))
         return UA_STATUSCODE_BADOUTOFMEMORY;
+    if(config->maxConnections && scs->currentChannelCount >= config->maxConnections &&
+       !purgeFirstConnectionWithoutChannel(cm, bpm))
+        return UA_STATUSCODE_BADOUTOFMEMORY;
```
jpfr, line 651: I think there is a dedicated status for that.
Something like out-of-networking-resources or similar.

jpfr:
Looking forward to the feature. This is very good to have.

src/server/ua_server_binary.c
jepa-hms: Is it t UA_STATUSCODE_BADTCPNOTENOUGHRESOURCES?
or maybe the more general UA_STATUSCODE_BADRESOURCEUNAVAILABLE?
Those were the only relevant-looking statuscodes I could find.

jepa-hms:
Thanks for the quick response! Have tried to address your comments

jpfr:
Now I am wondering to which extend `purgeFirstChannelWithoutSession` already removes the same SecureChannels.
Because for us any connection immediately becomes a SecureChannel.
Already for the HEL/ACK/OPN sequence.

jpfr:
I mean that `purgeFirstChannelWithoutSession` closes the same SecureChannels as your code.
Because all connections are added as SecureChannels already.
So the new code-path is actually never taken.

jepa-hms:
The new code-path is taken if the maximum number of connections is higher than the maximum number of channels. This would allow the number of connections and securechannels to be unrelated. In my tests I am seeing clear a difference in behavior with this change.
For clarity I am trying to replicate the functionality in this commit: 83b59054182fe0da2f641323c2f220add0636dac

jepa-hms:
Pinging this.
For more clarification, this new code get called because of this guard clause in `UA_SecureChannel_shutdown,` which is called by `purgeFirstChannelWithoutSession`.
https://github.com/open62541/open62541/blob/7c4ea969590311df5b2c1f30b0a5ce219e479adf/src/ua_securechannel.c#L151

jpfr:
`UA_SecureChannel_isConnected` covers all SecureChannels that do not have an active socket.
A new TCP connection should automatically get at least the `UA_SECURECHANNELSTATE_CONNECTING` state.

Still having troubles to see why TCP connections are not covered by the previous logic.
Can you set a breakpoint in the new logic and dump debug logs?

jepa-hms:
> A new TCP connection should automatically get at least the UA_SECURECHANNELSTATE_CONNECTING state.

That's not what I am seeing. New connections are set to `UA_SECURECHANNELSTATE_CLOSED` initially. Seems like the only place the state is set during channel creation is where this memset sets it to 0 (closed):
https://github.com/open62541/open62541/blob/7c4ea969590311df5b2c1f30b0a5ce219e479adf/src/ua_securechannel.c#L30

Here is the log. I added a debug log to `purgeFirstConnectionWithoutChannel` right before `closeConnection` is called, similar to where the log in `purgeFirstChannelWithoutSession` is. I open 5 connections and `maxConnections` is set to 4 to force a purge. `maxSecureChannels` is set to 3.
```
[1601-01-01 00:00:19.025] (null)/trace  Starting the EventLoop
[1601-01-01 00:00:21.116] (null)/network        TCP     | Listening on all interfaces
[1601-01-01 00:00:21.116] (null)/server New DiscoveryUrl added: opc.tcp://(null):4840
[1601-01-01 00:00:21.120] (null)/trace  Starting the EventLoop
[1601-01-01 00:00:21.125] (null)/client skip verifying ApplicationURI for the SecurityPolicy http://opcfoundation.org/UA/SecurityPolicy#None
[1601-01-01 00:00:21.126] (null)/network        OPC UA URL is invalid
[1601-01-01 00:00:21.126] (null)/client Client Status: ChannelState: Fresh, SessionState: Closed, ConnectStatus: 
[1601-01-01 00:00:21.209] (null)/network        TCP 6   | Connection opened from "" via the server socket 5
[1601-01-01 00:00:21.210] (null)/channel        TCP 6   | SC 1  | SecureChannel created
[1601-01-01 00:00:22.220] (null)/network        TCP 7   | Connection opened from "" via the server socket 5
[1601-01-01 00:00:22.221] (null)/channel        TCP 7   | SC 2  | SecureChannel created
[1601-01-01 00:00:23.232] (null)/network        TCP 8   | Connection opened from "" via the server socket 5
[1601-01-01 00:00:23.232] (null)/channel        TCP 8   | SC 3  | SecureChannel created
[1601-01-01 00:00:24.158] (null)/network        TCP 9   | Connection opened from "" via the server socket 5
[1601-01-01 00:00:24.159] (null)/channel        TCP 6   | SC 1  | Channel was purged since maxSecureChannels was reached and channel had no session attached
[1601-01-01 00:00:24.159] (null)/channel        TCP 9   | SC 4  | SecureChannel created
[1601-01-01 00:00:25.160] (null)/network        TCP 10  | Connection opened from "" via the server socket 5
[1601-01-01 00:00:25.160] (null)/channel        TCP 6   | SC 1  | Channel was purged since maxSecureChannels was reached and channel had no session attached
[1601-01-01 00:00:25.160] (null)/channel        TCP 6   | SC 1  | Connection was purged since maxConnections was reached and channel was not open
[1601-01-01 00:00:25.161] (null)/channel        TCP 10  | SC 5  | SecureChannel created
[1601-01-01 00:00:25.161] (null)/channel        TCP 0   | SC 0  | SecureChannel closed
[1601-01-01 00:00:25.162] (null)/network        TCP 6   | Socket closed
[1601-01-01 00:00:30.175] (null)/channel        TCP 0   | SC 0  | SecureChannel closed
[1601-01-01 00:00:30.175] (null)/network        TCP 9   | Socket closed
[1601-01-01 00:00:30.176] (null)/channel        TCP 0   | SC 0  | SecureChannel closed
[1601-01-01 00:00:30.176] (null)/network        TCP 8   | Socket closed
[1601-01-01 00:00:30.180] (null)/channel        TCP 0   | SC 0  | SecureChannel closed
[1601-01-01 00:00:30.180] (null)/network        TCP 7   | Socket closed
[1601-01-01 00:00:30.181] (null)/channel        TCP 0   | SC 0  | SecureChannel closed
[1601-01-01 00:00:30.182] (null)/network        TCP 10  | Socket closed
[1601-01-01 00:00:51.916] (null)/network        TCP 5   | Socket closed
[1601-01-01 00:00:51.916] (null)/trace  Stopping the EventLoop
[1601-01-01 00:00:51.916] (null)/network        TCP     | Shutting down the ConnectionManager
[1601-01-01 00:00:51.916] (null)/trace  The EventLoop has stopped
```
As you can see, `purgeFirstConnectionWithoutChannel` is called right after `purgeFirstChannelWithoutSession`. Keep in mind the message is logged _before_ `UA_SecureChannel_shutdown` is called. So the message doesn't tell us if the channel was already closed or not.

PR #6877, opened 2024-11-16
feat(client): Add new function to retrieve the user supplied subscription and monitored item contexts

RolfKal:
I need this when trying to integrate open62541 into an environment (LabVIEW from National Instruments) that is considerably different than most programming environments. It offers a possibility to interface to shared libraries but requires event resources that must be allocated and deallocated in the LabVIEW environment itself and from the shared library side can't be managed, only used. So I store that resource in the subscription user context.
While there is a delete callback to manage that user context, I can not call any C function from that callback to deallocate the event resource when the subscription is deleted. Instead I have to retrieve that event resource and pass it back into the environment to close it there.
Without this function I would have to somehow post the event resource from the delete callback to some queue or something to retrieve then retrieve it after the UA_Client_Subscriptions_delete() call. Rather than adding yet another asynchronous operation it would be much easier to simply retrieve the user context directly as it is already stored in the client handle.

### DIFF
--- a/include/open62541/client_subscriptions.h
+++ b/include/open62541/client_subscriptions.h
@@ -102,6 +102,10 @@ UA_Client_Subscriptions_delete_async(UA_Client *client,
 UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Client_Subscriptions_deleteSingle(UA_Client *client, UA_UInt32 subscriptionId);
 
+/* Retrieve the user supplied subscription contexts */
+UA_StatusCode UA_EXPORT UA_THREADSAFE
+UA_Client_Subscriptions_getUserContexts(UA_Client *client, UA_UInt32 subscriptionId, void **subContext, UA_UInt32 monitorId, void **monitorContext);
+
 static UA_INLINE UA_THREADSAFE UA_SetPublishingModeResponse
 UA_Client_Subscriptions_setPublishingMode(UA_Client *client,
     const UA_SetPublishingModeRequest request) {
--- a/include/open62541/client_subscriptions.h
+++ b/include/open62541/client_subscriptions.h
@@ -102,9 +102,12 @@ UA_Client_Subscriptions_delete_async(UA_Client *client,
 UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Client_Subscriptions_deleteSingle(UA_Client *client, UA_UInt32 subscriptionId);
 
-/* Retrieve the user supplied subscription contexts */
+/* Retrieve or change the user supplied subscription contexts */
 UA_StatusCode UA_EXPORT UA_THREADSAFE
-UA_Client_Subscriptions_getUserContexts(UA_Client *client, UA_UInt32 subscriptionId, void **subContext, UA_UInt32 monitorId, void **monitorContext);
+UA_Client_Subscriptions_getContext(UA_Client *client, UA_UInt32 subscriptionId, void **subContext);
+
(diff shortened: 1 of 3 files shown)


include/open62541/client_subscriptions.h
```diff
@@ -102,6 +102,10 @@ UA_Client_Subscriptions_delete_async(UA_Client *client,
 UA_StatusCode UA_EXPORT UA_THREADSAFE
 UA_Client_Subscriptions_deleteSingle(UA_Client *client, UA_UInt32 subscriptionId);
 
+/* Retrieve the user supplied subscription contexts */
+UA_StatusCode UA_EXPORT UA_THREADSAFE
+UA_Client_Subscriptions_getUserContexts(UA_Client *client, UA_UInt32 subscriptionId, void **subContext, UA_UInt32 monitorId, void **monitorContext);
```
jpfr, line 107: Split this into separate methods for Subscription and MonitoredItem.
Also provide both _get and _set methods.

Also call it _getContext instead of _getUserContext.
Then this is consistent with the existing API. Such as UA_Server_setAdminSessionContext.
https://github.com/open62541/open62541/blob/31947a93608e6e7d7132f06573ad00c263641b7b/include/open62541/server.h#L1115

src/client/ua_client_subscriptions.c
```diff
@@ -1007,6 +1007,37 @@ UA_Client_MonitoredItems_modify_async(UA_Client *client,
     return statusCode;
 }
 
+static void *
+UA_MonitoredItem_findByID(void *data, UA_Client_MonitoredItem *mon) {
```
jpfr, line 1011: I like to have `UA_MonitoredItem_findByID`.
But it should do the zip-tree-walking internally (or with another static method).
This is then also consistent with the other _findById methods. And it can be easily reused.

jpfr (commented):
Nice!
Comments inline.

RolfKal:
I believe I did address all the points you mentioned. Did not think that you wanted to have an exported function to find a monitored item but just an internal convenience function like findSubscriptionById() which I renamed as such too.

src/client/ua_client_subscriptions.c
```diff
@@ -1007,6 +1043,70 @@ UA_Client_MonitoredItems_modify_async(UA_Client *client,
     return statusCode;
 }
 
+static void *
+ua_MonitoredItem_findByID(void *data, UA_Client_MonitoredItem *mon) {
+	UA_UInt32 monitorId = *(UA_UInt32*)data;
+	if (monitorId && (mon->monitoredItemId != monitorId)) {
```
jpfr, line 1049: Here you want to check for equality. Not inequality!
But this is rather inefficient because the ziptree is iterated linearly.
The balanced tree can be searched in log time however.

The methods for that are defined here:
https://github.com/open62541/open62541/blob/31947a93608e6e7d7132f06573ad00c263641b7b/src/client/ua_client_subscriptions.c#L46

Here is an example use:
https://github.com/open62541/open62541/blob/31947a93608e6e7d7132f06573ad00c263641b7b/src/client/ua_client_subscriptions.c#L1067

You can copy / paste the code from the last example and use findMonitoredItemById in that place instead.

jpfr: Ah I see the problem.
We have both MonitoredItemId and ClientHandle.
It is okay to have the linear search for the MonitoredItemId then...

jpfr:
Nice, merged.

PR #6882, opened 2024-11-19
[Review] feat(plugins): Support certificate generation with mbedTLS [backporting]

GoetzGoerisch:
(cherry picked from commit 1f19a8e484f5e6ff3ec77793c67f433b6318810c)

Back porting #6145 from main.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1043,7 +1043,7 @@ if(UA_ENABLE_ENCRYPTION_MBEDTLS OR UA_ENABLE_PUBSUB_ENCRYPTION OR UA_ENABLE_AMAL
 endif()
 
 if(UA_ENABLE_PUBSUB_ENCRYPTION)
-  list(INSERT plugin_sources 1
+  list(APPEND plugin_sources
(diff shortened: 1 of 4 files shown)


jpfr:
Good initiative!
The linker still fails:
grafik

NoelGraf:
The problem occurs with the Windows Amalgamation build.
The reason for this is that the UA_inet_pton macro is not defined in the generated open62541.c file until it is used.
My suggestion is to move the file /arch/win32/ua_architecture.h in the CMake project to the same place as /arch/posix/ua_architecture.h. This should ensure that the definition is done in the Windows Amalgamation build before it is used.

Line 878 in cmake should look as follows:

if (NOT "${UA_ARCHITECTURE}" STREQUAL "none")
    list(PREPEND lib_headers ${PROJECT_SOURCE_DIR}/arch/posix/ua_architecture.h)
    list(PREPEND lib_headers ${PROJECT_SOURCE_DIR}/arch/win32/ua_architecture.h)
endif()

GoetzGoerisch:
Thanks, this solved it.

jpfr:
👍

PR #6887, opened 2024-11-20
Discussion: feature(client): Add an extra callback to the subscription structure to handle dataChange events as an aggregated callback

RolfKal:
Send one callback event for all items in a notification rather than individually per item

This is a proposal and I would like to get some feedback about if this is a desired feature or simply seen as unnecessary overhead. If this is seen as a desirable extension, I can add the according function exports to set and clear this callback.

It could be done by an additional UA_Client_Subscription_setDataChangeCallback() function or instead by an additional UA_Client_Subscription_createEx() function with this additional callback parameter.


electro-logic:
@RolfKal I think this should be a high-priority feature, as triggering a callback for each item is extremely inefficient for the user code logic in real-world. I would say this is the main current issue of this library.

I have updated #6443 some days ago to compile against the master branch, please see https://github.com/open62541/open62541/pull/6857, but I would have a look at this PR soon because improvements in this area are much appreciated and needed.

RolfKal:
> The main open issue with #6443 and #6857 is that is not possible to call further client functions inside the callback, because the code is still running inside the event_loop. This means that the user-code should start a new thread to have a callback that is not running inside the event_loop and able to call other functions (ex. read other nodes)

That should be solved with my proposal as I do release the lock before invoking the callback and then acquire it again after the call. Also I actually create a callback specific array structure to pass the monitorItemId && monitorContext && DataValue pointer.
The one thing you need to observe in that callback is to NOT touch (or clear) the DataValue pointer nor anything in the Variant in there as it is really owned by the event loop and cleaned by it after the callback returns control to it.

RolfKal:
Well, first why do you want to access client functions in the callback that involve the eventloop? That should not be needed in normal cases?
And what do you mean with filtering? Where do you see that it is filtered?

electro-logic:
Let's suppose we are monitoring a node to check the status of a motor, when the status changes we have to immediately turn off something by writing other nodes. If this scenario is not possible (inside the callback) should be very clear and documented in my opinion.

In ua_client_subscriptions.c we have this code at line 1235 of the PR

```
UA_DataItemsChangeNotification *items = (UA_DataItemsChangeNotification *)
                             UA_calloc(numItems, sizeof(UA_DataItemsChangeNotification));
        if(items) {
          // run callback
        }
```

Is the if instruction filtering (preventing) the callback if there are no items?

RolfKal:
> Is the if instruction filtering (preventing) the callback if there are no items?

Good question. :-) The C standard says about what calloc() should do when called with a 0 size only this:

> If the size of the space requested is zero, the behavior is implementation-defined: either a null pointer is returned, or the behavior is as if the size were some nonzero value, except that the returned pointer shall not be used to access an object.

So the answer is simply: It depends on your compiler!

The other question is: Does OPC UA even generate 0 size notifications? Why? There are also statusChangeCallbacks that could serve as a sort of heartbeat if you absolutely want to have something like that.

electro-logic:
> So the answer is simply: It depends on your compiler!

The behavior should be consistent across different platforms and compilers, would be better to check the number of items before allocating the memory to solve this situation

`if(sub->dataChangeCallback && (numItems > 0))`

> The other question is: Does OPC UA even generate 0 size notifications? Why? There are also statusChangeCallbacks that could serve as a sort of heartbeat if you absolutely want to have something like that.

According to the official standard https://reference.opcfoundation.org/Core/Part4/v104/docs/5.13.5 a PublishResponse is sent anyway.

> This Service [Publish] is used for two purposes. First, it is used to acknowledge the receipt of NotificationMessages for one or more Subscriptions. Second, it is used to request the Server to return a NotificationMessage or a keep-alive Message. Since Publish requests are not directed to a specific Subscription, they may be used by any Subscription. 5.13.1.2 describes the use of the Publish Service.

Anyway I checked the calling code and the PublishResponse message is already filtered if notificationDataSize is zero here:

https://github.com/open62541/open62541/blob/3636fef49a7f48d705f316c63fa557abda5b572a/src/client/ua_client_subscriptions.c#L1383

electro-logic:
> As to accessing other client functions inside the client callback: it should work, as long as those functions do not involve another event loop roundtrip (I think the async APIs should still work if you are absolutely inclined to do that).
> 
> But one of the core principles of callbacks should always be to make them lean and mean. If your callback does take more than a few CPU beats to process, you are totally and utterly degrading the system. Don't do that!! A network roundtrip where your callback is synchronously waiting on a response from the remote server (not possible with UA APIs as explained above but you could add some other nasty long lasting calls in your callback that will need to suspend the calling thread to wait for something) is a completely sure way to make your system crawl and sooner or later crash down because of event queues filling up as the events can't be processed. By invoking an async request you can at least eliminate the synchronous waiting on the response as that response will be processed in your async create callback. But it still feels wrong to burden the open62541 library with managing your process logic dependencies. The proper way is to post the notifications from your callback to a queue that another thread in your application monitors and deques the items and then processes them there.

I think we can decouple this topic to another Issue, because is more general and related to the EventLoop and not to this PR.

My main point is not about being right or wrong, if the library is currently working this way, should be clear and documented with something like "Callbacks [callback list] are running inside the EventLoop and by design they can't call other functions calling the Eventloop to avoid nested calls. You can call async functions or a workaround can be etc.. etc..". We can of course think and evaluate other options for a future version to get rid of this limitation.

RolfKal:
> @RolfKal I can test this PR if ready, is it already merged to 1.4 branch?

As far as the implementation is concerned, I believe it to be finished. I was planning to also look at adding some tests, shamelessly stealing from the previous PR submission but still struggling to get the unit test setup properly working on my Windows machine. That checkTest program is really not very well maintained for Windows installation. I simply can't get it to install in a way that CMake would be happy to find it on my systems.

Refering to 1.4 branch, so far I have exclusively developed against the master branch. Not exactly sure what is the proper workflow in this project. Is it more useful to make PRs against 1.4 or is it better to do them agains master?

RolfKal:
> If you need further assistance in configuring the solution to compile unit tests, please let me know

I'll have to tackle a few other things first but will get back to this in a week or so. Probably will rebase to 1.4 and add at least some basic tests.

RolfKal:
> Hi @RolfKal, any update about this PR? Would be great to have it merged

Been somewhat distracted with my own projects and just kept this monitored at the edge. My actual plan was to instead make a patch for 1.4 and try to get that in instead. But that also requires a few other minor things in 1.4 being updated.

jpfr:
Hey @RolfKal 

The following idea for how to get this done:

We would like to have "local sessions".
Where the local user - via the C API - can use a local session instead of being always the "admin user".

That allows interesting application swhere users can be logged in both local or remote.

Then the local sessions can define a "local subscription" that aggregates value changes like it would happen over the network.

Issue #4896, opened 2022-01-10
IPv6 LDS-ME


Muddasir-Shakil:
## Description
I was working on the discovery using LDS-ME specifically for ipv6 networks and I have few points to discuss and questions to ask.

1. Open62541 relies on the third party library to support ipv4, which is anyhow a forked repo by @Pro. It is also behind the original upstream (mdnsd). mdnsd is planning to add ipv6 support in release v1.0. 
Q. Is this a sustainable solution in a long run in your opinion? 
2.  mdnsd has planned to add ipv6 support in release v1.0, which has currently no due date. However, there are other solutions for mdns implementation. One implementation might be [mjansson/mdns](https://github.com/mjansson/mdns). It supports both ipv4 and ipv6 and it provides a single header only cross-platform file. 
Q. what do you think of switching from mdnsd to another implementation?
Q. what are the pitfalls and showstoppers in doing so?
Q. What other implementations are suitable for open62541 in your opinion?
Q. Apart from external deps, size and ip network support, what could be the criteria to select mdns implementation for open62541?
3. Current implementation of LDS-ME is tightly-coupled with mdnsd implementation. Switching to other implementation would require a great effort and re-implementation of some parts in Discovery manager in open62541. I think better approach and effort well spent  would be to define a common interface for discovery manager and have plugins to integrate external mdsn implementation (we have for security modules). This will also create opportunities to integrate avahi like implementations, if the system allows it.
Q. What do you think about it and if this a solid approach how should we start working on it?

Looking forward to your feedback and discussion.

jpfr:
Hey there, thanks for the remarks.

right now the LDS-ME subsystem has quite some historical baggage.
It should be moved over to the new EventLoop subsystem so it doesn't have to open a separate thread and pull in system dependencies.

Do you have the bandwith to work on this?
Otherwise we can keep the notes for posterity and @JanSurft who is also looking into this.

I have two must criteria. The rest is up for debate:
- Integrate with the new EventLoop subsystem for the communication.
- Use an actively maintained upstream for the mdns library.

Using the EventLoop subsystem means first to create UDP support.
That is needed for mdns and for PubSub.
The TCP support should be easy enough to copy/port over to UDP.

If the upstream mdnsd project is going to supoort IPv6, great!
We can move to mjansson/mdns if that simplifies the code.
I would trade off less complexity on our end against more complexity if the mdns library is well-maintained or stable enough.

Muddasir-Shakil:
[...]

PR #6888, opened 2024-11-20
[Review] Add avahi as mDNS option

V-Strassheim:
With this PR I would like to pick up the following discussion and work of @Muddasir-Shakil:
- https://github.com/open62541/open62541/issues/4896
- https://github.com/open62541/open62541/pull/5121

As already mentioned, mDNS should migrate from the [libmdnsd](https://github.com/Pro/mdnsd) fork to another implementation.
So far the the following options have already been proposed, to which I have added my opinion:
1. Switch to upstream mdnsd
-> A simple switch to upstream mdnsd is not possible, as the extension for CMake and msvc will not be adopted (https://github.com/troglobit/mdnsd/discussions/81). As the implementation of ipv6 has also come to a standstill, a switch would not bring any significant advantages.
2. Switch to an alternative solution, e.g. https://github.com/mjansson/mdns
-> Switching to an alternative stack has the disadvantage that the functionality provided by a standalone daemon/service for mdns must be re-implemented in open62541. This is therefore at most an option for simple devices that do not need to have the complete mdns resolver functionality.
3. Implementation of the discovery manager as a plugin to make it easier to connect different mdns solutions.
-> A clean interface for plugins is missing and there is no standardized API for different mdns libraries.

Since I also could not understand what advantage own event loop connections bring for mdns, I decided to go with a comination of 2 and 3 and change the current implementation to avahi to improve LDS-ME at least for some of the systems.
I tested the behavior mainly with the open62541 discovery examples on two Linux computers in a local network and used UaExpert as a client. When testing the current implementation, I noticed a few other little things besides the missing ipv6 support and https://github.com/open62541/open62541/issues/6805 that will be fixed with this PR. This also makes https://github.com/open62541/open62541/pull/5121 obsolete, as avahi brings support for ipv6.

As I have stuck to the previous range of functions, the following points remain open:
- mDNS collisions are only logged, not handled
- Only opcua.tcp servers are supported

The following new questions arise:
- mdns was previously bound to a specific interface.  I suspect it's just due to the custom implementation of the sockets/connections. Avahi registers the service on all suitable interfaces, could this cause a problem?
- What do we do with systems that implement a different mdnsd service instead of avahi (Windows, Zephyr, libmdnsd, systemd-resolve, ...)?
- How about systems that may not have an external mdnsd service running at all?

closes https://github.com/open62541/open62541/pull/5121
fixes https://github.com/open62541/open62541/issues/6805


jpfr:
This is a very good initiative.
Let's talk about the integration and also how to handle other operating systems (Windows, Zephyr, etc.).
The Windows build doesn't currently work, as avahi is Linux only.

From what I see in the avahi code base, there are three ways to integrate:
- Use dbus (Linux only)
- Integrate with another event loop by an `AvahiWatch` structure - but still using POSIX sockets (https://github.com/avahi/avahi/tree/master/avahi-libevent)
- Handle sockets and timeouts manually and use the avahi core API directly

The core avahi API looks to be platform-independent.
So my hunch is that we should manage the sockets ourselves.
And then replicate the few dozen lines of code for handling the received packets (which builds on the core API).
That way we reuse the existing platform-independent multicast UDP in our EventLoop.
That keeps the socket handling in one place and with uniform logging.

See here for the main server functionality we would have to replicate:
https://github.com/avahi/avahi/blob/1dade81cbf4d3e3f49784e44e543cd77046c79df/avahi-core/server.c#L363

V-Strassheim:
mDNS is a protocol that is independent of OPC UA. If a corresponding service is running on the system, then in my opinion it should also be used. Having all the core mDNS code (like sending/receiving mDNS messages) in open62541 itself only increases the effort for open62541 (as we already see with the IPv6 issue).

Actually, what open62541 needs is only an interface for (de-)initialization, service registration and querying which OPC UA servers are known on the network.

My suggestion would rather be to create a more standardized interface in open62541 to be able to use different mDNS services. Is there anything against introducing mDNS as a plugin, similar to what is already done for encryption?
open62541 could then offer a suitable solution for Avahi for Linux (based on in my draft), [win32](https://learn.microsoft.com/en-us/uwp/api/windows.networking.servicediscovery.dnssd?view=winrt-26100) and [apple](https://github.com/apple-oss-distributions/mDNSResponder/tree/main).
With mainline [libmdnsd](https://github.com/troglobit/mdnsd/tree/master/libmdnsd) plugin the user could then compile this for his system and use it almost as before.
If that is a possible way, I would adjust my current request to provide an mdns-avahi plugin.

jpfr:
> Actually, what open62541 needs is only an interface for (de-)initialization, service registration and querying which OPC UA servers are known on the network.

That sounds good. Please propose such an API.

How would that integrate with the "server thread"?
Just a cyclic interval to retrieve events from the mdsnd layer?

V-Strassheim:
Ok, I'll get back to you as soon as I have something.

The integration with the server thread was previously a UA_DiscoveryManager_cleanupTimedOut timer callback, which was also used to pass and receive packets to and from mdnsd.
For avahi, I [reused](https://github.com/open62541/open62541/pull/6888/commits/bc07509f1a560c64def0da6ae132a8acb944352a) this timer to call a non-blocking avahi_simple_poll_iterate to look for new opc ua services on the network. Is that what you meant?
I think that the other libraries can be used in the same way.

V-Strassheim:
I have retained most of the existing interfaces and tidied them up so that it is possible to implement corresponding functions in the “mdns-module” for other solutions.
Instead of replacing the previous implementation for the libmdnsd fork, there is now a selection option for mDNS with an additional avahi realization.
I am not sure if it is necessary to move the corresponding files to the plugins folder.
Would it be a possibility to continue with mDNS this way?

V-Strassheim:
The tests are now running again, the code can now be reviewed.

When testing the ua_discovery_mdns example with AddressSanitizer enabled, some aligment errors occurred in libmdnsd (_rrparse in 1035.c) which I didn't see with the mainline libmdnsd.
This is one more reason why I would suggest to abandon the previous integration of libmdnsd into the build system in favor of easier maintenance and use it as an external library as well. This is what I did in this PR for avahi. Then it would also be easier to switch to libmdnsd mainline, whereby, if I see it correctly, only a few adjustments to the libmdnsd API are necessary.
Of course, it would have the disadvantage that open62541 would not provide an all-in-one solution and therefore the user would have to take care of how mDNS is implemented on one's own system. But I don't see this as an obligation to build everything into an OPC UA stack.
I have not tested it, but the solution for Windows users could be a self-built libmdnsd with minGW/Cygwin. Or one can use WSL right away.

Briefly summarized, I see two options:
1. as before, include mDNS code in open62541
    -    main disadvantage: increased care and maintenance effort for non-OPC UA specific code.
    -    main advantage: easier to build/use (when it is well maintained)
2. integration of mDNS libraries (my favorite)
   -    main disadvantage: possibly increased effort for users if they want to use LDS-ME.
   -    main advantage: less maintenance effort in open62541, as only connection to defined, mostly stable APIs.

How should we proceed?

V-Strassheim:
> > Actually, what open62541 needs is only an interface for (de-)initialization, service registration and querying which OPC UA servers are known on the network.
> 
> That sounds good. Please propose such an API.
> 
> How would that integrate with the "server thread"? Just a cyclic interval to retrieve events from the mdsnd layer?

@jpfr I have adjusted this PR accordingly, should I rather start a new one?

CMakeLists.txt
```diff
@@ -252,7 +288,7 @@ endif()
 
 if(UA_ENABLE_DISCOVERY_MULTICAST AND NOT UA_ENABLE_DISCOVERY)
     MESSAGE(WARNING "UA_ENABLE_DISCOVERY_MULTICAST is enabled, but not UA_ENABLE_DISCOVERY. UA_ENABLE_DISCOVERY_MULTICAST will be set to OFF")
-    SET(UA_ENABLE_DISCOVERY_MULTICAST OFF CACHE BOOL "Enable Discovery Service with multicast support (LDS-ME)" FORCE)
+    SET(UA_ENABLE_DISCOVERY_MULTICAST_MDNSD OFF CACHE BOOL "Enable Discovery Service with multicast support (LDS-ME)" FORCE)
```
jpfr, line 291: Instead of manually rewriting the config, throw a fatal error.
It is better to let the user know about a problem than to do magic changes in the background.

The principle of least surprise is important for a project of this complexity.

jpfr (approved):
Looks good.
I’m merging this as-is.

We will do some changes in the future that will allow Mdns to become an exchangeable “plugin”.

Then this code gets moved out of the core library into /plugins.

jpfr:
Thanks for the good work!

PR #6892, opened 2024-11-22
Client-side namespace management

NoelGraf:
This allows the configuration of fixed namespace identifiers.
The mapping is performed automagically in the background.


jpfr:
Looks very good.

- Please fold the commit `fix(tests): Add missing binary encoding parameter` into the commit with the API change.
- Re-enable the SecureChannel_renew unit test

Can be merged otherwise.

PR #6893, opened 2024-11-22
[Draft] Add a MetaType structure

NoelGraf:
This pull request is about defining a MetaType structure that can be used to process variables whose data type is unknown at compile time. As there are no generated C structures for this data type, processing is based on the MetaType structure instead.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -83,6 +83,8 @@ endif()
 
 add_example(tutorial_server_variabletype tutorial_server_variabletype.c)
 
+add_example(tutorial_server_metatype tutorial_server_metatype.c)
+
 add_example(tutorial_server_object tutorial_server_object.c)
 
 add_example(tutorial_server_reverseconnect tutorial_server_reverseconnect.c)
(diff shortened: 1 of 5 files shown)


RolfKal:
I was today just checking out how to possibly support reading arbitrary structure data values from a server. Would this be used for that or is that something else?

NoelGraf:
@RolfKal I'm not quite sure what you mean by "arbitrary structures". This PR is about simplifying the writing and reading of data types that are only known at runtime. For example, when a nodeset is loaded at runtime, there are no generated C structures for the data types to simplify access. The MetaType structure is intended to simplify the writing and reading of variable nodes when the data type does not have a generated C structure.

RolfKal:
@NoelGraf thanks. I was looking into how to support reading and writing complex (structure type, possibly recursive) nodes from my client. I have in the meantime written something of my own to construct and or parse the according UA_DataType tree, with non-zero type->membersSize value on the fly and it seems to work fairly well in my isolated test cases. Will have to test with real data from servers in the wild.

andreasebner:
This PR is obsolete due to the integration of UA_Client_getRemoteDataTypes in the master. Tank's for share the concept.

PR #6904, opened 2024-11-26
Updated Visual Studio build instructions on Windows

electro-logic:
Updated Visual Studio 2022 build instructions on Windows. Now CMake is integrated and a separate download is not required anymore.

Note: are Debian packages link still updated? 
For example the last daily build is from 2022: https://launchpad.net/~open62541-team/+archive/ubuntu/daily

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -1,0 +1,117 @@
+﻿{
+  "configurations": [
+    {
+      "name": "x86-RelWithDebInfo",
+      "generator": "Ninja",
+      "configurationType": "RelWithDebInfo",
+      "inheritEnvironments": [ "msvc_x86_x64" ],
+      "buildRoot": "${projectDir}\\build\\${name}",
+      "installRoot": "${projectDir}\\out\\install\\${name}",
+      "cmakeCommandArgs": "",
+      "buildCommandArgs": "",
+      "ctestCommandArgs": "",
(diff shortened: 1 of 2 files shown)


jpfr:
The previous documentation also works for MinGW. That is now lost.

Please split and have a dedicated documentation each for standalone cmake and visual studio.

electro-logic:
MinGW is nowadays not used much. To compile with GCC under Windows is now possible to use WSL (Windows Subsystem for Linux). Many companies (ex. Intel) already moved away from Cygwin/MinGW in their development environment (ex. Quartus).

I can re-add it, but I think would be easier to drop support for MinGW in the next release.

jpfr:
There are also some cross-compilers for embedded that integrate with cmake.

Native cmake is not specific for any one compiler.

but it is good to have native visual studio also.

jpfr:
Loks good.
Please the conventions for the git message heading.
Cf. The contributing.md file.
Can be merged then.

electro-logic:
Changed the latest commit message to:

docs: building instruction for Visual Studio 2022

jpfr:
This is better but not the style we expect.

please rebase to get rid of cleanup commits.
And use the style you see documented and in the git history.

electro-logic:
recreated here https://github.com/open62541/open62541/pull/6910

PR #6907, opened 2024-11-26
Fix clearing accesscontrol plugin in UA_AccessControl_default()

### LINKED ISSUE #6901: Invalid cleanup in `UA_AccessControl_default()`
The cleanup of an existing access control instance in `UA_AccessControl_default()` seems wrong:

https://github.com/open62541/open62541/blob/3636fef49a7f48d705f316c63fa557abda5b572a/plugins/ua_accesscontrol_default.c#L303-L304

This implies that the previously set access control instance is always clearable by `clear_default()` which is only the case when it has been created with `UA_AccessControl_default()` in the first place. For other access control instances, this would probably fail with a segfault.

Shouldn't it be:

```c
    if(ac->clear)
        ac->clear(ac);
```

sgoll:
Fixes #6901. See there for rational and explanation.

### DIFF
--- a/plugins/ua_accesscontrol_default.c
+++ b/plugins/ua_accesscontrol_default.c
@@ -301,7 +301,7 @@ UA_AccessControl_default(UA_ServerConfig *config,
     UA_AccessControl *ac = &config->accessControl;
 
     if(ac->clear)
-        clear_default(ac);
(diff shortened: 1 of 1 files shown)


jpfr:
Merged. Thanks.

PR #6915, opened 2024-11-27
fix(tools): ua2json compilation issue when <unistd.h> is not available

electro-logic:
<unistd.h> is not available in Visual Studio, but it's not really required because can be removed and ssize_t can be replaced by int

### DIFF
--- a/tools/ua2json/ua2json.c
+++ b/tools/ua2json/ua2json.c
@@ -10,7 +10,6 @@
 #include <open62541/types_generated_handling.h>
 
 #include <stdio.h>
-#include <unistd.h>
 
 /* Internal headers */
 #include "ua_pubsub_networkmessage.h"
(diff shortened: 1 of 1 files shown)


RolfKal:
> TODO: CI GCC 9 Error

on line 15 in ua2json.c there is a c define to associate ssize_t with SSIZE_T defined in the standard header BaseTsd.h. Does your version of Visual C not come with that header?

RolfKal:
> Hi, Visual Studio 2022 is compiling fine. The error is from the CI server where GCC 9 failed to compile.
> 
> https://github.com/open62541/open62541/actions/runs/12051516706/job/33602724925

Sure, because the GCC CI setup is very picky about non-explicit datatype type casts, and rightly so. What I mean is that the code you show for ua2json.c doesn't match my version of the code which is latest master.

'/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 *
 *    Copyright 2018 (c) Fraunhofer IOSB (Author: Julius Pfrommer)
 */

#include <open62541/types.h>
#include <open62541/types_generated.h>
#include <open62541/pubsub.h>

#include <stdio.h>
#if defined(_MSC_VER)
# include <BaseTsd.h>
typedef SSIZE_T ssize_t;
#else
#include <unistd.h>
#endif
'

Issue #6909, opened 2024-11-26
BadDataEncodingInvalid on UA_Client_HistoryRead_raw


ffreality:
## Description

`UA_Client_HistoryRead_raw` returns `BadDataEncodingInvalid`.

## Background Information / Reproduction Steps

This is my C++ code in Unreal Engine 5 to access historical values. As you see I used `UA_Client_HistoryRead_raw`
https://gist.github.com/ffreality/00f6d69030811021244982202efe459a

But it returns `BadDataEncodingInvalid` error just like this issue.
https://github.com/open62541/open62541/issues/6386

I put breakpoints on my `Callback_History (Line 36)` and its return values but none of them triggered.

But if I commented out this line, it works.

```
(Line: 64075 in open62541.c)
item.dataEncoding = UA_QUALIFIEDNAME(0, "");
```

v1.4.6 doesn't have this problem. Didn't try v1.4.7

## Used CMake options:
Amalgamation and Historical Access enabled.
I used RelWithDebInfo build.

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): v1.4.8
 - [x] Other OPC UA SDKs used (client or server): ProSYS Simulation Server Free + Visual Studio 22 + UE5.5 (C++ 20)
 - [x] Operating system: Windows 11
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [x] Self-contained code example attached
 - [x] Critical issue

PR #6919, opened 2024-11-28
bugfix: BadDataEncodingInvalid on Historical Access

ffreality:
defining dataEncoding as `(0, "")` cause `BadDataEncodingInvalid` on `ProSYS Simulation Server` when we try to access historical value. We need to commented out that setting or set NULL to solve this issue.

Reported Issue:
https://github.com/open62541/open62541/issues/6909


jpfr:
Merged. Thanks.
I also adjusted the commit message title.

PR #6922, opened 2024-11-29
refactor(docs): update example code in README.md

electro-logic:
Examples were outdated and not compiling

Note: PR for 1.4 only, API in the Master branch are changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,22 +148,22 @@ socket library).
 int main(int argc, char** argv)
 {
     /* Create a server listening on port 4840 (default) */
-    UA_Server *server = UA_Server_new();
+    UA_Server* server = UA_Server_new();
 
     /* Add a variable node to the server */
 
     /* 1) Define the variable attributes */
     UA_VariableAttributes attr = UA_VariableAttributes_default;
-    attr.displayName = UA_LOCALIZEDTEXT("en-US", "the answer");
+    attr.displayName = UA_LOCALIZEDTEXT_ALLOC("en-US", "the answer");
     UA_Int32 myInteger = 42;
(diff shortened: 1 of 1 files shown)


jpfr:
The change leads to memleaks.
What makes you think the change is necessary?

jpfr:
Yeah. But then shouldn't we fix the implementation instead of introducing memleaks?

The definition is right here:
https://github.com/open62541/open62541/blob/0cc73feb099d3c412ac3caf048ecea12e7251ceb/include/open62541/types.h#L643

So it should work. Please show error logs.

jpfr:
By the way, on master this would not fly.
There we have added tests where all examples are checked for memleaks with valgrind in the CI.

jpfr:
The resulting UA_String contains a non-const pointer.
Even if we make the resulting UA_String itself const.

The only clean way would be to malloc and make a copy.
But we definitely want to be able to have strings that are inline-defined and do not need cleanup afterwards.

So casting to (char*) and documenting the behavior still seems like the lesser evil.
I cannot imagine a compiler that exploits any UB resulting from a cast to a char-ptr..
It is a very common pattern.

jpfr:
C developers are very aware when a String is allocated on the Stack and when on the heap.

And our documentation is explicit when the Input to a method is const and when it gets freed eventually.

We need to support stack allocated strings.
We will not introduce a dedicated struct for const strings.

I repeat my question. Is it possible for MSVC to do  a hard cast inside the UA_STRING macro?

in addition with documentation for the usage patterns.

jpfr:
This needs to be reopened when it is less invasive and has the memleaks removed.

for example, modify the original UA_STRING instead of overriding.

PR #6927, opened 2024-12-02
feat(pubsub): Make DataSetFieldId and field Description configurable.

MurzynJ:
In some use cases it's necessary to configure them. There is no reason why they should not be part of the configuration.


jpfr:
Looks good. Please add a usage of this feature to a fitting unit test.

jpfr:
Nice! Merged.

PR #6930, opened 2024-12-02
fix(pubsub): Fix invalid DataType comparison on creation of DSR

MurzynJ:
UA_TYPES_* are indexes of the types in the type array (they are starting from 0).
We need to compare the builtInType to the well-known numeric identifier of the type (they are starting from 1).


jpfr:
👍

PR #6932, opened 2024-12-03
fix[pubsub]: Properly remove key rollover callback

maciek5991:


jpfr:
Looks good.
Would this be valid for 1.4 already?

maciek5991:
Yes, there is the same issue in 1.4

jpfr:
Then please rebase this for 1.4 and we pull it up to master from there.

jpfr:
👍

PR #6933, opened 2024-12-03
fix[pubsub]: Remove the existing key rollover callback before creatin…

maciek5991:
…g a new one

### DIFF
--- a/src/pubsub/ua_pubsub_keystorage.c
+++ b/src/pubsub/ua_pubsub_keystorage.c
@@ -208,7 +208,12 @@ UA_PubSubKeyStorage_addKeyRolloverCallback(UA_Server *server,
     if(!server || !keyStorage || !callback || timeToNextMs <= 0)
         return UA_STATUSCODE_BADINVALIDARGUMENT;
 
-    UA_LOCK_ASSERT(&server->serviceMutex, 1);
+    // Check if the callback already exist
+    if(*callbackID != 0) {
+        return UA_STATUSCODE_GOOD;
+    }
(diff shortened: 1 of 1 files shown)


jpfr:
Thanks.
This could be handled inside `UA_PubSubKeyStorage_addKeyRolloverCallback`.
Then other users of this method have the same protection...

src/pubsub/ua_pubsub_keystorage.c
```diff
@@ -165,6 +165,12 @@ UA_PubSubKeyStorage_addKeyRolloverCallback(UA_PubSubManager *psm,
     if(!psm || !ks || !callback || timeToNextMs <= 0)
         return UA_STATUSCODE_BADINVALIDARGUMENT;
 
+    // First remove the existing callback
+    if(*callbackID != 0) {
```
jpfr, line 169: On the contrary. When the callback already exists, then we don't have to re-add it and can do an early exit.

jpfr (commented):
Comment inline.
Please rebase this for 1.4 and we pull it up to master from there.

src/pubsub/ua_pubsub_keystorage.c
maciek5991: I don't agree. By not re-registering the callback, we risk missing timing adjustment.

maciek5991: ok, I agree. We don't have to re-add it

jpfr:
👍

PR #6941, opened 2024-12-05
[Discussion]: Add user context parameter to UA_Server_createCondition Public API

m-meingast:
Adding a custom context to a Condition Object the same way as for regular Objects would be a useful feature since it increases the versatility of Alarms and Conditions. Especially the State Callback functions, Enabled, Acknowledged, Confirmed, Active, will be more usable.

The proposed change would have an impact on the Public API, that's why I left the current function and created a new one around it. The name `UA_Server_createConditionWithContext` might be bulky, but since A&C is still marked as experimental I suggest to put this method in place of the current one. In this way, the additional qualifier would not be necessary and the same design would be used as for the creation of regular Objects.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -1623,17 +1623,29 @@ typedef UA_StatusCode
  * @param conditionType The NodeId of the node representation of the ConditionType
  * @param conditionName The name of the condition to be created
  * @param conditionSource The NodeId of the Condition Source (Parent of the Condition)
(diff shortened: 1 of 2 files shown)


m-meingast:
PR Update:

I kindly ask again, if there is anything against having a user context parameter. It would be beneficial for us as we use OPC UA Alarms & Conditions in our IEC 61499 project (https://github.com/eclipse-4diac/4diac-forte). The context variable of the condition object would allow to access the 61499 domain from the respective condition callback function.

Issue #6803, opened 2024-10-29
Make open62541 a Zephyr module


xR3b0rn:
There is an existing discussion about whether placing the `zephyr` folder at the top level of this repository is advisable: [PR #6792](https://github.com/open62541/open62541/pull/6792).

After further research ([Zephyr Modules Doc](https://docs.zephyrproject.org/latest/develop/modules.html#modules-vs-west-projects)) and consulting the [discussion forum](https://github.com/zephyrproject-rtos/zephyr/discussions/80553) on Zephyr's official GitHub, it appears that including the `zephyr` folder with a `module.yml` file is the only recognized method to integrate external libraries as a Zephyr module.

Without this folder, Zephyr’s build system cannot automatically detect this library as a module. Alternatively, you could use `west` to clone the `open62541` repository and manually add it as a module by listing `open62541` in the `EXTRA_ZEPHYR_MODULES` list of the project's `CMakeLists.txt`, similar to what is done in the [server example](https://github.com/open62541/open62541/blob/master/examples/zephyr/server/CMakeLists.txt#L7). However, this is unusuale and not the normal way of including Zephyr modules.

jpfr:
Pleas add a small shell-script to /arch/zephyr that adds folders and symlinks.
So that it takes only a single command to locally make open62541 detectable as a zephyr module.

xR3b0rn:
This approach won't work. In the Zephyr build system, a tool called `west` manages Git dependencies similarly to CMake's `FetchContent`. So, if someone wants to integrate `open62541` into their project, they would add the Git link to the `west.yml` file at the project's root. `West` then handles cloning the repository and checking out the correct branch. Once that's done, CMake can automatically locate `open62541`, making it available as a target immediately after running `find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})`.

However, this mechanism will not work if there is no `zephyr/module.yml` file at the root of the `open62541` repository.

PR #6951, opened 2024-12-09
refactor(zephyr): Make open62541 a zephyr module

xR3b0rn:


xR3b0rn:
Please squash the commits.

xR3b0rn:
However, the proposed solution of creating a script (`arch/zephyr/prepare_west.sh`) to set up a symlink (`/zephyr -> /arch/zephyr`) would not solve the issue. This is because `west` strictly manages Zephyr modules by looking for a `module.yml` file at `zephyr/module.yml` relative to the repository root. 

When `west` clones a repository, it expects this specific structure. If the `module.yml` file is not present at `zephyr/module.yml`, `west` cannot recognize the repository as a valid Zephyr module, and the dependency management process fails. Unfortunately, `west` does not follow symlinks or allow alternative paths to locate the `module.yml` file.

To ensure compatibility with `west` and proper handling of this library as a Zephyr module, the `zephyr` folder needs to be at the top level of the repository. While I understand the concern about maintaining a clean top-level folder, this structure is a requirement for `west` to manage the project correctly.

If the `zephyr` folder cannot be placed in the root directory, we must accept that `open62541` will not be able to function as a Zephyr module—at least not until the Zephyr team relaxes the restrictions for module structures.

RolfKal:
What's the chance to make west smarter to work with different projects rather than expecting each dependency to match the limited west model?

xR3b0rn:
@RolfKal I'm not a Zephyr expert, nor am I aware of any discussions that led to the decision to make this `cmake` and `west` symbiosis a core part of the Zephyr's build system. However, everything I've learned about how Zephyr utilizes `cmake` feels completely incorrect and flawed.

RolfKal:
> @RolfKal I posed this specific question in the official Q&A of the Zephyr project: how to handle a maintainer who refuses to include the `zephyr` folder in the root directory of their repository? I don't have high expectations from someone who responds to such a question with, "start maintaining a fork."

That kind of makes me wonder about the project itself. It sounds a lot like my way or the highway and if they want to provide an OS for a larger audience they will have to be somewhat flexible. It especially amazes me since support for symlinks it pretty much not work for a reading application. It takes actually extra work to avoid supporting them under Unix, all normal file IO functions will simply treat them as if they exist for real. So this west tool specifically not working with them sounds like some very weird implementation choice. Even on Windows it takes effort to not support them as most file IO function will simply treat them as if the real thing by default. Creating them under Windows is a different story, but that would not be the task of this west tool.

xR3b0rn:
@RolfKal actually idk if west supports symlinks. The point is, even if symlinks are supported, it wouldn't solve anything.

RolfKal:
> @RolfKal actually idk if west supports symlinks. The point is, even if symlinks are supported, it wouldn't solve anything.

Are you sure? I'm by far no GIT guru, just barely manage to use it in the most basic form, but there probably is some way to have a specific script run when updating a dependency in the GIT flow. It seems to me that GIT has grown to be the every dragon and monster slayer tool by now, that has about at least 3 different, and usually rather involved, methods for every monster to slay them.

jpfr:
You don‘t need to fork.
You can instead make a „package“ repository that references the upstream.

Examples:
https://github.com/microsoft/vcpkg/tree/master/ports/open62541
https://salsa.debian.org/jpfr/open62541

That can be very lightweight if you include the upstream source as a bit submodule.

dm5tt:
Any progress on this pull request? Without this patch I wasn't able to get open62541 running as a west module.

xR3b0rn:
@dm5tt not from my side. As a workaround you can do what is described here: https://github.com/open62541/open62541/issues/6803#issue-2621414324

PR #6952, opened 2024-12-09
improve(tools): Remove unneeded header includes

RolfKal:
these are not needed anymore after the earlier commit b1f5e0e


jpfr:
Nice. Squash-merged and refactored the commit message.

PR #6956, opened 2024-12-10
feat(ci): build and upload amalgamation sources for releases

marwinglaser:
Add a workflow that automatically builds and uploads an amalgamation build based on the debian (salsa) configuration for every release.
To ensure we catch failures before the actual release, this workflow builds on every push, but only tries to upload when it was triggered by a release being published.

Here are example runs for [push events](https://github.com/marwinglaser/open62541/actions/runs/12259608912/job/34202174116) and [release events](https://github.com/marwinglaser/open62541/actions/runs/12259890365/job/34203154860), as well as the corresponding [release](https://github.com/marwinglaser/open62541/releases/tag/v1.4.8-test) with the attached files.

(This needs some small changes when merging back into master to adapt to e907a0ed91427e449ed711bada7a5db3947cd644 and 12bd03568b4b0c8e16b8a6fefd18d03fbbb2e759)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,18 @@
+name: Build Release Artifacts
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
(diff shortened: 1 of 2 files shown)


jpfr:
👍

PR #6959, opened 2024-12-11
fix(plugin): Monitor 'trusted' subdirectory in PKI

MichalTyndel:
inotify doesn't work recursively, subdirectories need to be monitored separately. As a result adding certificate to 'trusted' directory directly didn't refresh the trustlist.


jpfr:
Good find!

PR #6960, opened 2024-12-11
fix(client): Fixed log message crashing application

MichalTyndel:
Wrong data were passed to logger, which resulted in segmentation fault.


jpfr:
👍

PR #6961, opened 2024-12-11
fix(server): Allow empty certificateGroupId in create CSR callback

MichalTyndel:
Default value (DefaultApplicationGroup) is used in this case.


jpfr:
@NoelGraf please check with the spec and confirm.
Do we need to adjust the documentation?

NoelGraf:
Looks good and conforms to specification.
I believe we should outline the behavior in the definition of UA_Server_createSigningRequest.

tomitkl:
Based on my testing there is similar issue in UA_Server_updateCertificate. Maybe that could be fixed at the same time.

NoelGraf:
I have created a new PR with your changes, as there were problems with the merge due to the rebases/adaptations.
But you are still the author. I hope that's okay.
https://github.com/open62541/open62541/pull/7132

andreasebner:
Replaced with #7132 (commit cherry-picked)

PR #6962, opened 2024-12-11
fix(plugin): Fixed handling of empty new private key

MichalTyndel:
Don't create empty key file when new certificate without private key is received (old key is used in this case).


jpfr:
@NoelGraf please confirm.

andreasebner:
Thank you for sharing this change. In recent versions, this behavior has been implemented.

Issue #6914, opened 2024-11-27
[BUG] ua2json not compiling with Visual Studio


electro-logic:
PR will follow shortly, is CI working fine? Compilation errors should be caught

electro-logic:
https://github.com/open62541/open62541/pull/6915

RolfKal:
The real fix is actually this #6950. It avoids Microsoft specific header magic entirely.

RolfKal:
> Thanks for the PR, to catch these issues in future would make sense to fix the CI too

Which CI do you refer to? The SKS one, which completely failed to compile after the latest commit? Or something to test the ua-cli tool itself? Or adding the tools compilation to the cmake configuration used on all CI systems including the Microsoft Windows ones?

RolfKal:
> * move to Visual Studio 2022 (vmImage: 'windows-2022' in tools/ci/azure-devops/azure-pipelines-win.yml)
> * set UA_BUILD_TOOLS=true in tools/ci/azure-devops/win/build.ps1 to build tools under Windows too

No comment about the first one. ;-) I still compile using VS 2015 and 2017 on my two main systems. Simply historical reasons and as I'm mainly doing C programming work, when I'm programming in text languages at all, those versions are way enough. In fact for a few projects I still have a VS 2005 installation in a VM. Lightweight, lean and mean in comparison to what we have now.

The second I fully agree, those tools regularly break. Would be also good to actually setup a build environment that actually tests that SHARED_LIBRARY correctly builds and runs under Windows too. That is currently a part that breaks almost with every second commit. Shared Library support under Windows has a few particularities that make it more restrictive than under the Linux ELF system.

RolfKal:
Yes and no! Newer versions support newer features such a C11 and newer things, while open62541 so far strives to be C99 compliant. With newer compilers it's very easy to have such new compliance features slip into the code and it is useful to have a compiler in the CI toolchain that will bork on them. That is unless we want to give up C99 compliance and move to C11 or newer but considering that open62541 should also compile on embedded targets that is likely one bridge to far to go.

The older compiler toolchain in the CI tests should not be a limitation, other than catching such non-conformant features sneaking into the code base.

RolfKal:
> The point was about "user experience". If a new user want to jump into open62541 using Windows, must have a tested path with currently available tools. CI with old compilers can be kept of course, but new compilers should also be added to CI in my opinion to ensure that everything is working fine.

[...]

PR #6965, opened 2024-12-12
ci: update to VS2022 and build 'tools, unit tests and examples'

electro-logic:
- Update the azure image from VS2019 to VS2022
- Compile 'tools, unit tests and examples'

Note: this PR is to check any VS2022 issue only

Follow-up of https://github.com/open62541/open62541/issues/6914#issuecomment-2538385687

From my understanding:
- Code is compiled against many compilers and likely GCC is catching any C99 issue.
- Azure hosted VM are managed by Microsoft and old versions may be removed, updating time to time should ensure continuity https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml
- CI should also compile against the minimum officially supported Visual Studio version. A custom VM may be required in this case.


jpfr:
Rebased to see if the CI issues are gone.

jpfr:
It doesn't find mbedTLS.
That's bad and probably affects other users of MSVC 2022.

```
Could NOT find MbedTLS 
```

When we install mbedTLS through vcpkg, it should be able to find it reliably.
Provided that the vcpkg cross-compilation setup is selected for CMake.
You could reproduce locally and add path hints to find_package.

jpfr:
It is strange that MSVC2022 actually builds.
It are the other compilers (mingw, clang) that are affected.

andreasebner:
Thanks for sharing. We reviewed the changes, and it appears they’re deprecated in the current codebase.

PR #6968, opened 2024-12-13
fix(pubsub): copy paste error in last commit for LOCK_ASSERT macro

RolfKal:


RolfKal:
Attempting to run a CI test run with just the obvious compilation error fixed in respect to current 1.4 in repository.

jpfr:
Overall very nice!

Why do you need `#if !defined(UA_UNITTEST)`?
I think we should instead abort with `message(FATAL_ERROR ...)` in the CMakeLists.txt if the user wants to build the unit tests with dynamic linking.

That reduces the amount of surprise compared to changing flags in the background.

RolfKal:
> Overall very nice!
> 
> Why do you need `#if !defined(UA_UNITTEST)`? I think we should instead abort with `message(FATAL_ERROR ...)` in the CMakeLists.txt if the user wants to build the unit tests with dynamic linking.
> 
> That reduces the amount of surprise compared to changing flags in the background.

Good question. It's a matter of taste I think. Personally I intend to use open62541 as dynamic library in my project. So I would welcome if the compilation supports that. But if I really have to, I will build the open62541 once to do unit tests and once to use in my own project.

This is anyhow a strange duck in the soup. The unit tests link directly with the object libraries in order to be able to access internal APIs of the open62541 library. And I fully understand that many of those APIs should not be exported if not absolutely necessary, so there is not much else to do about that. But that clashes on Windows when the UA_Export define is indicating that the officially exported APIs are DLL imports. So that define disables the UA_Export definition when compiling unit test sources. One alternative would be to do some config.h vodoo, which is even worse.
But rather than doing a FATAL_ERROR() I would probably opt for the same measure as is done for other components such as certain examples that produce a cmake warning about not being able to compile with shared library support and being then automatically disabled.

jpfr:
I get where you are aiming at.
But this current condition can lead people to disable dynamic linking by mistake.
Just because unit tests are activated. That should not happen.

i se most changes from this pr and e now covered elsewhere.
Can we close?

RolfKal:
> i se most changes from this pr and e now covered elsewhere. Can we close?

Sure! I left the change to config.h.in out of that other patch. It's not really critical for me.

Issue #6724, opened 2024-09-23
UA_Session.attributes seems to be NULL in a valid session


lukasberbuer:
## Description

`UA_Server_setSessionAttribute` fails with `UA_STATUSCODE_BADINVALIDARGUMENT` when working with a valid session.
The status code is generated in the call `UA_KeyValueMap_set(session->attributes, key, value)`:

```c
UA_StatusCode
UA_KeyValueMap_set(UA_KeyValueMap *map,
                   const UA_QualifiedName key,
                   const UA_Variant *value) {
    if(map == NULL || value == NULL)  // <-- map is NULL here
        return UA_STATUSCODE_BADINVALIDARGUMENT;
    // ...
}
```

## Background Information / Reproduction Steps

Following minimal example tries to write a session attribute in a random AccessControl function.
Just run the server, connect with a client and try to access any node.
It will print the status code of the `UA_Server_setSessionAttribute` call: "Status: BadInvalidArgument"

```c
#include <stdio.h>

#include <open62541/plugin/accesscontrol_default.h>
#include <open62541/server.h>

static UA_Byte getUserAccessLevel(
    UA_Server *server, UA_AccessControl *ac,
    const UA_NodeId *sessionId, void *sessionContext,
    const UA_NodeId *nodeId, void *nodeContext
) {
    UA_QualifiedName key = UA_QUALIFIEDNAME_ALLOC(1, "test");
    UA_Variant variant;
    UA_Variant_init(&variant);
    double value = 11.11;
    UA_Variant_setScalar(&variant, &value, &UA_TYPES[UA_TYPES_DOUBLE]);
    UA_StatusCode status = UA_Server_setSessionAttribute(server, sessionId, key, &variant);
    printf("Status: %s\n", UA_StatusCode_name(status));  // prints "Status: BadInvalidArgument"
    UA_QualifiedName_clear(&key);
    return 0xFF;
}

int main() {
    UA_Server* server = UA_Server_new();
    UA_ServerConfig* config = UA_Server_getConfig(server);

    UA_String encryptionPolicy = config->securityPolicies[config->securityPoliciesSize-1].policyUri;
    config->accessControl.clear(&config->accessControl);
    UA_AccessControl_default(config, true, &encryptionPolicy, 0, NULL);
    config->accessControl.getUserAccessLevel = getUserAccessLevel;

    UA_StatusCode retval = UA_Server_runUntilInterrupt(server);
    // now connect with any client and try to access a node

[...]

PR #6970, opened 2024-12-15
fix(server): Missing session attribute init and KVM removal

chandryan:
Fixed issues related to:

- Initializing a session's attribute key value map
- Removing values from a key value map (mapSize was reduced twice)

Also added test cases that check against the fixed issues.

Related to:

- https://github.com/open62541/open62541/issues/6724
- https://github.com/open62541pp/open62541pp/issues/362


jpfr:
Looks good!
Please change the commit title from fix(session) to fix(server), etc.
Can be merged when green.

chandryan:
> Looks good! Please change the commit title from fix(session) to fix(server), etc. Can be merged when green.

Thanks! I checked the contribution guidelines again and changed the commit titles, I hope they're correct now.

jpfr:
Nice, merged!

Issue #6777, opened 2024-10-21
Documentation/examples update [Segmentation fault occurs when clearing a browse request for a stack allocated NodeId]


costinzanfir:
## Description
Segmentation fault occurs when clearing a browse request for a stack allocated NodeId that is identified by a STRING.

## Background Information / Reproduction Steps
Segmentation fault occurs when clearing a browse request for a stack allocated NodeId that is identified by a STRING.
```c
UA_NodeId nodeid = UA_NODEID_STRING(3, "CpuLoad");
UA_BrowseRequest_init(&breq);
breq.requestedMaxReferencesPerNode = 0;
breq.nodesToBrowse = UA_BrowseDescription_new();
breq.nodesToBrowseSize = 1;
breq.nodesToBrowse[0].nodeId = nodeid;
breq.nodesToBrowse[0].resultMask = UA_BROWSERESULTMASK_ALL;

UA_BrowseRequest_clear(&breq);
```
The cause has been tracked down to be the release of memory in case the NodeId is a STRING or a BYTESTRING. Creating the node dynamically
```c
/* UA_NodeId nodeid = UA_NODEID_STRING(3, "CpuLoad"); */
UA_NodeId nodeid = UA_NODEID_STRING_ALLOC(3, "CpuLoad");
```
or using
```c
UA_NodeId_copy(&nodeid, &breq.nodesToBrowse[0].nodeId);
```
clears the issue.

Would it be possible to update the examples (open62541-1.4.6/examples/client.c) with a comment as follows:
```
 72     bReq.requestedMaxReferencesPerNode = 0;
 73     bReq.nodesToBrowse = UA_BrowseDescription_new();
 74     bReq.nodesToBrowseSize = 1;
 75     /* In case the NodeId is identified by a string or bytestring, make a dinamically
 76        allocated copy of it (UA_NodeId_copy(..., &breq.nodesToBrowse[0].nodeId)). */
 77     bReq.nodesToBrowse[0].nodeId = UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER); /* browse objects folder */
 78     bReq.nodesToBrowse[0].resultMask = UA_BROWSERESULTMASK_ALL; /* return everything */
```

Thank you
Costin

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

Please provide the following information:

[...]

PR #6971, opened 2024-12-16
fix(examples): https://github.com/open62541/open62541/issues/6777

costinzanfir:
Segmentation fault occurs when clearing a browse request for a stack allocated NodeId. The example has been updated by adding a comment to warn the user about the use of STRINGS/BYTESTRINGS in case of node browsing. See: https://github.com/open62541/open62541/issues/6777


jpfr:
- typo: dynamically
- you can use `UA_NODEID_STRING_ALLOC` as a shorthand that will do the allocation

jpfr:
Making a copy works in general.
But knowing the shorthand (and that you only need to allocate for STRING and BYTESTRING) can make for cleaner code.

Issue #6871, opened 2024-11-14
[BUG] When encryption is enabled, reverse connect does not succeed [BadTcpEndpointUrlInvalid]


costinzanfir:
## Description
When enabling encryption, reverse connect (client side: UA_Client_startListeningForReverseConnect(client, &listenHost, 1, 4841);) does not succeed.

Direct connection (client side: UA_Client_connect(client, server_ep);) is not affected.

## Background Information / Reproduction Steps
Logs from the client. The last 3 lines are of interest.
```bash
[2024-11-14 11:37:48.815 (UTC+0000)] info/client        Rejecting endpoint 0: security mode does not match
[2024-11-14 11:37:48.815 (UTC+0000)] info/client        Rejecting endpoint 1: security policy does not match the configuration
[2024-11-14 11:37:48.815 (UTC+0000)] info/client        Rejecting endpoint 2: security mode does not match
[2024-11-14 11:37:48.815 (UTC+0000)] info/client        Rejecting endpoint 4: security mode does not match
[2024-11-14 11:37:48.815 (UTC+0000)] info/client        Rejecting endpoint 5: security policy does not match the configuration
[2024-11-14 11:37:48.815 (UTC+0000)] info/client        Selected endpoint with EndpointUrl opc.tcp://host:4840, SecurityMode SignAndEncrypt and SecurityPolicy http://opcfoundation.org/UA/SecurityPolicy#Aes256_Sha256_RsaPss
[2024-11-14 11:37:48.815 (UTC+0000)] info/channel       TCP 5   | SC 1  | SecureChannel closed
[2024-11-14 11:37:48.815 (UTC+0000)] warn/client        skip verifying ApplicationURI for the SecurityPolicy http://opcfoundation.org/UA/SecurityPolicy#None
[2024-11-14 11:37:48.816 (UTC+0000)] info/securitypolicy        The Aes256Sha256RsaPss security policy channel with openssl is created.
[2024-11-14 11:37:48.816 (UTC+0000)] warn/network       OPC UA URL is invalid:
[2024-11-14 11:37:48.816 (UTC+0000)] info/client        Client Status: ChannelState: Fresh, SessionState: Closed, ConnectStatus: BadTcpEndpointUrlInvalid
[2024-11-14 11:37:48.816 (UTC+0000)] info/network       TCP 5   | Socket closed
```
[...]

PR #6973, opened 2024-12-16
fix(client): Read the Url from the remote endoint

costinzanfir:
In case 'reverse connect' is used, read the Url
of the remote endpoint instead of using the
local static configuration.

See https://github.com/open62541/open62541/issues/6871


andreasebner:
Thank's for share'ing. Closing as obsolete on current 1.5: initConnect intentionally opens the TCP connection from config.endpointUrl, while getEndpointUrl (preferring discoveryUrl) is reserved for the HEL/GetEndpoints phases.

PR #6986, opened 2024-12-19
fix(client example) client_subscription_loop.c: Async functions used in callbacks

ThomasNittel:
Async functions have to be used in callbacks else the error message "Cannot run EventLoop from the run method itself" occurs under Windows 10 with Microsoft Visual Studio 17 2022.


jpfr:
Thanks 👍

Issue #6738, opened 2024-09-30
compilation warning in __ZIP_INSERT


StalderT:
I have been using open62541 v1.3 without any issues for years in one of my applications. I tried upgrading to version 1.4 (amalgamation), and when I compile, I get the error:

```
In function '__ZIP_INSERT',
    inlined from 'MonitorItemsTree_ZIP_INSERT' at ../../src/libs/open62541.c:57844:1,
    inlined from 'ua_MonitoredItems_create' at ../../src/libs/open62541.c:58309:9:
../../src/libs/open62541.c:60757:11: error: 'prev_order' may be used uninitialized [-Werror=maybe-uninitialized]
60757 |         if(prev_order == ZIP_CMP_MORE)
      |           ^
../../src/libs/open62541.c: In function 'ua_MonitoredItems_create':
../../src/libs/open62541.c:60740:29: note: 'prev_order' was declared here
60740 |     enum ZIP_CMP cur_order, prev_order;
      |                             ^~~~~~~~~~
In function '__ZIP_INSERT',
    inlined from 'UA_ServerComponentTree_ZIP_INSERT' at ../../src/libs/open62541.c:5084:1,
    inlined from 'addServerComponent' at ../../src/libs/open62541.c:22914:5,
    inlined from 'addServerComponent' at ../../src/libs/open62541.c:22908:1:
../../src/libs/open62541.c:60757:11: error: 'prev_order' may be used uninitialized [-Werror=maybe-uninitialized]
60757 |         if(prev_order == ZIP_CMP_MORE)
      |           ^
../../src/libs/open62541.c: In function 'addServerComponent':
../../src/libs/open62541.c:60740:29: note: 'prev_order' was declared here
60740 |     enum ZIP_CMP cur_order, prev_order;
      |                             ^~~~~~~~~~
../../src/libs/open62541.c: In function '__ZIP_INSERT':
../../src/libs/open62541.c:60757:11: error: 'prev_order' may be used uninitialized [-Werror=maybe-uninitialized]
60757 |         if(prev_order == ZIP_CMP_MORE)
      |           ^
../../src/libs/open62541.c:60740:29: note: 'prev_order' was declared here
60740 |     enum ZIP_CMP cur_order, prev_order;
      |                             ^~~~~~~~~~
In function '__ZIP_INSERT',
    inlined from 'RefHead_ZIP_INSERT' at ../../src/libs/open62541.c:28738:1,
[...]

PR #6994, opened 2024-12-30
fix(deps): Remove maybe-uninitialized warning from ziptree

fedepell:
Warning appears (at least) with GCC 11.5.0 and 14.2.1 when using -O2, so reproduce for example with both -O2 and -Werror=maybe-uninitialized and get:
```
‘prev_order’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
```
Could fix also #6738


jpfr:
Nice, thanks

PR #6995, opened 2025-01-01
refactor(examples): do not set an invalid IP in server_multicast.c

laroche:
Currently an IP 42.42.42.42 is set and the comment suggests that the network stack resets this to 0.0.0.0 in server_multicast.c. Just set 0.0.0.0 directly.

This is a backport of 253fc8abd5855851acc0f0d7fa62c98d2487f4f2 to the 1.3 branch.


jpfr:
👍

PR #6996, opened 2025-01-01
refactor(build): enable -Wshadow compiler warnings

laroche:
The compiler warnings for -Wshadow are not part of the -Wall or -Wextra compiler warnings and need to be enabled separately.


jpfr:
Good idea.
This uncovers a lot of problem areas.

Issue #6937, opened 2024-12-04
The examples do not work with UA_NAMESPACE_ZERO FULL


V-Strassheim:
## Description
The examples e.g. /examples/tutorial_server_firststeps do not work if built with the option UA_NAMESPACE_ZERO set to FULL. The following error occurs when trying to start it:
Initialization of Namespace 0 (after bootstrapping) failed with BadNodeIdUnknown.

## Background Information / Reproduction Steps

The error came in with the commit 21e9c240ecad1bd8021977827a8778a4a2c1ec4f.

Used CMake options:
BUILD_SHARED_LIBS:BOOL=OFF
CMAKE_BUILD_TYPE:STRING=Debug
CMAKE_INSTALL_PREFIX:PATH=/usr/local
UA_ARCHITECTURE:STRING=posix
**UA_BUILD_EXAMPLES:BOOL=ON**
UA_BUILD_TOOLS:BOOL=OFF
UA_BUILD_UNIT_TESTS:BOOL=OFF
UA_ENABLE_AMALGAMATION:BOOL=OFF
UA_ENABLE_DA:BOOL=ON
UA_ENABLE_DATATYPES_ALL:BOOL=ON
UA_ENABLE_DIAGNOSTICS:BOOL=ON
UA_ENABLE_DISCOVERY:BOOL=ON
UA_ENABLE_ENCRYPTION:STRING=OFF
UA_ENABLE_ENCRYPTION_TPM2:STRING=OFF
UA_ENABLE_GDS_PUSHMANAGEMENT:BOOL=OFF
UA_ENABLE_HISTORIZING:BOOL=ON
UA_ENABLE_JSON_ENCODING:BOOL=ON
UA_ENABLE_METHODCALLS:BOOL=ON
UA_ENABLE_NODESETLOADER:BOOL=OFF
UA_ENABLE_PUBSUB:BOOL=ON
UA_ENABLE_PUBSUB_INFORMATIONMODEL:BOOL=ON
UA_ENABLE_SUBSCRIPTIONS:BOOL=ON
UA_ENABLE_SUBSCRIPTIONS_EVENTS:BOOL=ON
UA_ENABLE_XML_ENCODING:BOOL=OFF
UA_LOGLEVEL:STRING=100
UA_MULTITHREADING:STRING=100
**UA_NAMESPACE_ZERO:STRING=FULL**
UA_NODESET_DIR:STRING=/open62541/deps/ua-nodeset

## Checklist

Please provide the following information:

[...]

PR #6999, opened 2025-01-02
[Review] Fix UA_FILE_NS0_PRIVATE variable unset

V-Strassheim:
According to the CMake [documentation](https://cmake.org/cmake/help/latest/command/unset.html): "unsetting a normal variable can expose a cache variable that was previously hidden".
This can lead to wrong build results if UA_NAMESPACE_ZERO is changed to FULL after configuring the project with default REDUCED value. This happens e.g. if ccmake is used like shown in the documentation.

Set UA_FILE_NS0_PRIVATE to "" like proposed in the CMake documentation to prevent this behavior.

This fixes https://github.com/open62541/open62541/issues/6937


jpfr:
👍

PR #7000, opened 2025-01-02
Fix memory leaks when using custom data types with string typeId or b…

martlaak-gmail:
…inaryEncodingId

### DIFF
--- a/src/ua_types.c
+++ b/src/ua_types.c
@@ -99,6 +99,8 @@ UA_cleanupDataTypeWithCustom(const UA_DataTypeArray *customTypes) {
                 const UA_DataType *type = &customTypes->types[i];
 #ifdef UA_ENABLE_TYPEDESCRIPTION
                 UA_free((void*)(uintptr_t)type->typeName);
+                UA_NodeId_clear((UA_NodeId*)&type->typeId);
+                UA_NodeId_clear((UA_NodeId*)&type->binaryEncodingId);                    
                 for(size_t j = 0; j < type->membersSize; ++j) {
                     const UA_DataTypeMember *m = &type->members[j];
                     UA_free((void*)(uintptr_t)m->memberName);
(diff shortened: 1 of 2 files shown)


src/ua_types.c
```diff
@@ -99,6 +99,8 @@ UA_cleanupDataTypeWithCustom(const UA_DataTypeArray *customTypes) {
                 const UA_DataType *type = &customTypes->types[i];
 #ifdef UA_ENABLE_TYPEDESCRIPTION
                 UA_free((void*)(uintptr_t)type->typeName);
+                UA_NodeId_clear((UA_NodeId*)&type->typeId);
```
jpfr, line 102: No need to do the casting here.

jpfr (commented):
Good find.
Please click the CLA.

Minor comment inline.

src/ua_types.c
martlaak-gmail: It seems needed to my compiler at least, as without it throws this warning: 
```ua_types.c:102:33: Passing 'const UA_NodeId *' to parameter of type 'UA_NodeId *' discards qualifiers
types_generated_handling.h:617:28: passing argument to parameter 'p' here```

martlaak-gmail: No, the problem is that in UA_DataTypeArray the `const UA_DataType *types;` is const. IMO, at the API level it is generally quite good that it is const. Library internals can just cast away constness in rare places where needed, UA_cleanupDataTypeWithCustom, as I did here.

jpfr: Yeah, I see it the same.

jpfr:
Please fix the remaining build errors.
MSVC does not like "casting away the const".
Force it by casting through a uintptr_t as `UA_NodeId_clear((UA_NodeId*)(uintptr_t)&type->typeId);`.

martlaak-gmail:
> Please fix the remaining build errors. MSVC does not like "casting away the const". Force it by casting through a uintptr_t as `UA_NodeId_clear((UA_NodeId*)(uintptr_t)&type->typeId);`.
Did it.

andreasebner:
Thanks for sharing! The proposed change has been integrated over time in recent versions (1.5 and master).

PR #7001, opened 2025-01-02
Add methods to get the member of opt-struct and union datatype

martlaak-gmail:


src/ua_types.c
```diff
@@ -2141,16 +2141,50 @@ UA_Array_delete(void *p, size_t size, const UA_DataType *type) {
 }
 
 #ifdef UA_ENABLE_TYPEDESCRIPTION
+
+UA_Boolean
```
jpfr, line 2145: Make this function static.

jpfr (commented):
This is a good feature.

- Please click the CLA so that we can use the contributed code
- The union and optstruct code should be accompanied by methods to test whether a member is active for the current instance of the type.

A small comment inline.

martlaak-gmail:
> This is a good feature.
>     * The union and optstruct code should be accompanied by methods to test whether a member is active for the current instance of the type.

From UA_DataType* we can only detect if the optional member is defined there, as a choice. To be able to detect if the member is currently active, by value, we would need to have UA_Byte* pointing to the value data as well, but by adding this parameter we would lose the symmetry with the original struct function. One anyway needs some traversing on value data ptr functions when working with structs, so can easily then detect if the member is active by looking at the member data ptr is NULL or not.

martlaak-gmail (commented):
Make internal _UA_DataType_getStructMember() function static

martlaak-gmail (commented):
Make internal _UA_DataType_getStructMember() function static

PR #7004, opened 2025-01-07
[Review] fix(build): only try to read open62541 version number from git when it is the main CMake project

sbreuss:
When open62541 gets included into another CMake project using CPM, it still tries to read the version info from a git tag. This leads to wrong version numbers within the config.h header. This PR fix fixes the problem and only changes the version number from git when open62541 is the main CMake project.


jpfr:
Nice, thanks.

I rewrote the commit message header to conform with our CONTRIBUTING.md guideline.

PR #7005, opened 2025-01-08
[Review] fix(plugins): return BadSecurityChecksFailed on untrusted cert or unknown issuer

jepa-hms:
In OPC UA CTT "Security ->Security Certificate Validation -> cases 005 and 009" expects BadSecurityChecksFailed when certificate is untrusted or issuer is unknown.


jpfr:
Internally I would say `UA_STATUSCODE_BADCERTIFICATECHAININCOMPLETE` is still the cause.

Note that we "rewrite" the StatusCode into `BadSecurityChecksFailed`.

Is this an assumed problem or an actual problem that occurs with the current CTT?
Because we have done a full CTT check end of last year.

https://github.com/open62541/open62541/blob/185bcf2f9543656e8205fefe0f1d960c43a54e0b/src/ua_securechannel.c#L94
https://github.com/open62541/open62541/blob/185bcf2f9543656e8205fefe0f1d960c43a54e0b/src/ua_securechannel.c#L71

jepa-hms:
I see, that makes sense.  Turns out this was an issue in 1.4.8 which is what I ran the CTT on, and it has been solved since. Sorry for taking up your time!

PR #7006, opened 2025-01-08
[Review] feat(plugin): Add filestore Windows implementation

tomitkl:
This pull request adds a filesystem porting interface to ua_filestore_common.h similar way than eventloop_posix.h does for sockets and then implements the interface for Windows in addition to the one that existed for Linux.

MIT-licensed dirent.h is added as a dependency for Windows builds.

The are couple of restrictions:
- Filesystem monitoring is not implemented for Windows
- dirname implementation for Windows doesn't fulfill the posix requirements and most probably doesn't work in all possible use cases but works, at least, with the unit tests

I originally wanted to test filestore backend with our own embedded device with GDS so I don't need filesystem monitoring as files cannot be copied directly to the device. We also have functions to implement dirname on our platform. That's why I didn't want to spend too much time to implement those features.

I don't have a Linux environment so the changes have only been tested locally on Windows (and similar changes with our embedded device).

Would there be interest to get this to the master branch? Is the implementation approach OK or should some other approach be selected?


jpfr:
This is great!

@NoelGraf please review.

Please remove the additional build flag.
We can just always have the filestore backend for Windows.

Any size-constrained application will be statically linked.
And that removes the code if it is unused without requiring build flags.

mcpat:
> Any size-constrained application will be statically linked. And that removes the code if it is unused without requiring build flags.

Fair enough. But how do you compile the open62541 stack if the target does not provide the required headers and/or functions?

tomitkl:
I would prefer to keep the build flag as mentioned in the previous comment. That will help compiling for other targets even if those targets are not as well supported as in 1.3.x and before.

jpfr:
> I would prefer to keep the build flag as mentioned in the previous comment.

You already have `UA_ARCHITECTURE_WIN32`.
You can activate the win32 filestore whenever you are on windows and have encryption.

No need for an additional flag.

tomitkl:
OK, I will check what I can do.

jpfr:
Yes!

jpfr:
Merged. Thanks for this. 👍

PR #7009, opened 2025-01-09
feat(server): add context free function to be called when UA_Node_cle…

dpazj:
…ar is performed


jpfr:
For this we have a destructor callback in the server.

If that is not what you need, please explain your use case and why the destructor does not fit the need.

jpfr:
Let's continue the discussion in https://github.com/open62541/open62541/issues/7010.

Issue #7010, opened 2025-01-09
Node context use after free


dpazj:
## Description

A use after free on a node's context is possible if using a Lifecycle destructor to free the context (see example). If an operation e.g a write, is performed and the node is deleted from the server.

In the example a node is written to. During the write the node is deleted from the server and its context free'd. However the write still holds a pointer to the context. 

I have created a PR with a possible solution where the user can register a callback for the context to be free'd when the UA_Node structure itself is cleared. The UA_Node will only be cleared once all outstanding operations are complete.  
see https://github.com/open62541/open62541/pull/7009

## Background Information / Reproduction Steps

```c


#include <open62541/plugin/log_stdout.h>
#include <open62541/server.h>
#include <stdio.h>
#include <pthread.h>
#include <time.h>
#include <unistd.h>
#include "open62541/client_config_default.h"
#include "open62541/client_highlevel.h"

# define sleep_ms(ms) usleep(ms * 1000)

typedef struct my_node_context {
    char * string;
    UA_Int32 val;
}my_node_context ;

static UA_StatusCode on_write(UA_Server *server, const UA_NodeId *sessionId,
                       void *sessionContext, const UA_NodeId *nodeId,
                       void *nodeContext, const UA_NumericRange *range,
                       const UA_DataValue *value)
{
    my_node_context *ctx = (my_node_context *) nodeContext;
    if (!ctx) return UA_STATUSCODE_BAD;
    fprintf (stdout, "on write called - doing write operation\n");
    sleep (3); //operation takes some arbitrary time
    ctx->val++;
    fprintf (stdout, "ctx: %s %u\n", ctx->string, ctx->val);
    return UA_STATUSCODE_GOOD;
}

static void destructor(UA_Server *server,
                   const UA_NodeId *sessionId, void *sessionContext,
                   const UA_NodeId *typeNodeId, void *typeNodeContext,
                   const UA_NodeId *nodeId, void **nodeContext)
{
    my_node_context *ctx = (my_node_context *) *nodeContext;
    if (ctx) {
        fprintf (stdout, "free context\n");
        free (ctx->string);
        free (ctx);
    }
    *nodeContext = NULL;
}


static void
addVariables(UA_Server *server) {

    UA_NodeTypeLifecycle lifecycle;
    lifecycle.destructor = destructor;
    lifecycle.constructor = NULL;
[...]

PR #7013, opened 2025-01-09
Recursive locking

jpfr:
By using recursive locking we do not have to release the locks before entering into user-supplied callbacks.
This solves #7010.

### DIFF
--- a/arch/common/ua_timer.c
+++ b/arch/common/ua_timer.c
@@ -204,9 +204,7 @@ processEntryCallback(void *context, UA_TimerEntry *te) {
      * Instead, whenever t->processTree != NULL, the entries are only marked for
      * deletion by setting elm->callback to NULL. */
     if(te->callback) {
-        UA_UNLOCK(&t->timerMutex);
         te->callback(te->application, te->data);
-        UA_LOCK(&t->timerMutex);
     }
 
     /* Remove and free the entry if marked for deletion or a one-time timed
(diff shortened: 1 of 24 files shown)


dpazj:
I understand it may be a bit contrived, but consider the simple example:

```c
typedef struct my_node_context {
    char * string;
    UA_Int32 val;
}my_node_context ;

static UA_StatusCode on_write(UA_Server *server, const UA_NodeId *sessionId,
                       void *sessionContext, const UA_NodeId *nodeId,
                       void *nodeContext, const UA_NumericRange *range,
                       const UA_DataValue *value)
{
    my_node_context *ctx = (my_node_context *) nodeContext;
    if (!ctx) return UA_STATUSCODE_BAD;
    UA_Server_deleteNode(server, *nodeId, true)
    ctx->val++;
    fprintf (stdout, "ctx: %s %u\n", ctx->string, ctx->val);
    return UA_STATUSCODE_GOOD;
}
```

I belive this would lead to a use after free. At what point does this library consider it to be on the user to account for scenarios like this?

jpfr:
For what you describe, we could delay the execution of the destructor.
So that it gets called once the access counter reaches zero.

jpfr:
Already now, the node only gets free'd when the access counter is zero.

dpazj:
Yeah that would make sense - that was my intention with https://github.com/open62541/open62541/pull/7009/files but I guess ideally we would want to use the destructor rather than adding a new callback?

PR #7015, opened 2025-01-10
fix(server): Use length of private key to check its validity when certificate is updated by GDS

tomitkl:
Use length of UA_ByteString instead of data to check if valid private key is received from GDS.

Data is not necessarily NULL even if length is 0 as it is cast from input buffer. See the below screenshot.

Näyttökuva 2025-01-10 143802

This is how UA_ByteString validity is checked elsewhere too.

Fix also "Declaration of 'i' hides previous local declaration" warning in applyChangesToServer.


NoelGraf:
Nice!

jpfr:
@tomitkl, please use fix(plugins): as the commit message prefix according to our conventions.
Can be merged then.

tomitkl:
Should it be fix(plugins) or fix(server)? The change is under server and the latest change in the same file also has fix(server) as commit message. See #6963.

jpfr:
Yes, fix(server).

src/server/ua_server_ns0_gds.c
```diff
@@ -297,7 +297,7 @@ updateCertificate(UA_Server *server,
 
     /* Verify that the privateKey is in a supported format and
      * that it matches the specified certificate */
-    if(privateKey && privateKey->data) {
+    if(privateKey && privateKey->length > 0) {
```
AdanRuoff, line 300: Just a small side note:
I have used function `UA_String_isEmpty` for cases like this when an empty bytestring can either be signaled by a length of zero or a data pointer that is NULL. It felt like the safest way to check for an empty bytestring, but this will only work as long as type UA_ByteString remains an alias of type UA_String, and sadly there is no `UA_ByteString_isEmpty` (yet?).

tomitkl: Yes, I also used it originally but as it was for UA_String specifically, I decided to go the way it is done elsewhere.

jpfr:
Something appears to have changed in the clang-analyzer package.
Seems to be unrelated from you...
grafik

tomitkl:
My previous pull request caused also a lot of linux builds to fail. I have a fix for that problem in #7018.

tomitkl:
I rebased the changes. There seems to be still issues with check_update_trustlist in some setups. The changes in this pull request shouldn't affect those.

jpfr:
👍

PR #7018, opened 2025-01-13
fix(plugins): Fix build issues caused by dirent.h

tomitkl:
Fix build issues introduced by #7006.

Rename dirent.h with the same logic as mp_printf.h/c to tr_dirent.h as some Linux builds were using dirent.h from deps directory instead of system dirent.h.

Add also include guard to ua_filestore_common.h.


NoelGraf:
I have looked into the issue and submitted a pull request. Once it's merged, you can just rebase.
https://github.com/open62541/open62541/pull/7021

Issue #7014, opened 2025-01-10
GDS PUSH trustlist file interface


tomitkl:
## Description

What is the reason that Trustlist file method nodes are replaced with FileType method nodes in initNS0PushManagement?

For example:
````
/* Open */
retval |= deleteNode(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTAPPLICATIONGROUP_TRUSTLIST_OPEN),true);
retval |= addRef(server, UA_NODEID_NUMERIC(0, UA_NS0ID_SERVERCONFIGURATION_CERTIFICATEGROUPS_DEFAULTAPPLICATIONGROUP_TRUSTLIST),
                     UA_NODEID_NUMERIC(0, UA_NS0ID_HASCOMPONENT),
                     UA_NODEID_NUMERIC(0, UA_NS0ID_FILETYPE_OPEN), true);
````

## Background Information / Reproduction Steps

I have tested the functionality with Unified Automation's UaGDS and it, at least, uses the node ids from the Trustlist interface directly. Therefore, the current open62541 implementation doesn't work with it without modifying the code.

Because I don't have access to any other GDS implementation at the moment, I haven't been able to test this with other GDS servers, so don't know how they behave.

Maybe someone knows if the current implementation is correct or should it be changed/improved somehow.

The problem should be easily seen if tested with UaGDS.

Used CMake options:

UA_ENABLE_ENCRYPTION: ON
UA_ENABLE_GDS_PUSHMANAGEMENT: ON
UA_NAMESPACE_ZERO: "FULL"

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): master
 - [x] Other OPC UA SDKs used (client or server): UaGDS
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

jpfr:
@NoelGraf this is for you

tomitkl:
Fixed by #7137.

Issue #6740, opened 2024-10-01
Method Call: Resolution issue when calling method from supertype on an object node


mcpat:
## Description

There is a problem when checking the relation between an object node and a given method node. Consider the following hierarchy:

> FileType `<|-hasSubType-` TrustListType `<-hasTypeDef-` DefAppGroup/TrustList

Further consider the following method relations:

> FileType `-hasComp->` Read (i=11585)
> TrustListType `-hasComp->` Read (i=12532)
> DefAppGroup/TrustList (i=12642)  `-hasComp->` Read (i=12652)

So now we have the following situation:

* Calling i=12652 on the TrustList object (i=12642) -> works
* Calling i=12532 on the object -> should work (if I understand the code correctly)
* Calling i=11585 on the object -> **does not work** - but it shall

See [also the spec](https://reference.opcfoundation.org/Core/Part4/v105/docs/5.11.2):
>In case of an [Object](https://reference.opcfoundation.org/search/357?t=Object) the [Object](https://reference.opcfoundation.org/search/357?t=Object) or the [ObjectType](https://reference.opcfoundation.org/search/357?t=ObjectType) of the [Object](https://reference.opcfoundation.org/search/357?t=Object) or a super type of that [ObjectType](https://reference.opcfoundation.org/search/357?t=ObjectType) shall be the source of a [HasComponent](https://reference.opcfoundation.org/search/357?t=HasComponent) [Reference](https://reference.opcfoundation.org/search/357?t=Reference) (or subtype of [HasComponent](https://reference.opcfoundation.org/search/357?t=HasComponent) [Reference](https://reference.opcfoundation.org/search/357?t=Reference)) to the [Method](https://reference.opcfoundation.org/search/357?t=Method) specified in [methodId](https://reference.opcfoundation.org/search/357?t=methodId).

In the code there is the following:
https://github.com/open62541/open62541/blob/9d615c2ce2efb9a5bf2314fa5e5107877c457978/src/server/ua_services_method.c#L236-L246

So the comment already describes what to do here. Yet the code only checks the objectType part and does not walk up the hierarchy to check all supertypes as well.

## Checklist

Please provide the following information:

 - [x] open62541 Version (1.4 and master):
 - [x] Operating system: Ubuntu 20.04.6 LTS

mcpat:
As always with the the OPC UA spec it is not sufficient to read just at one location to (sadly) guess what is actually desired here.

[...]

PR #7026, opened 2025-01-16
[Discussion] fix(server): Use trustlist methods instead of filetype methods with GDS

tomitkl:
Related to issue #7014.

At least UaGDS uses trustlist method nodes when it writes the trustlist to the server.

I'm not sure what is the correct approach but with these changes I'm able to get our server to work with UaGDS.

I don't have access to any other GDS so I'm not sure how they work. I'm happy to test with other ones too if someone knows any alternatives.

I think, CTT is happier with this approach too although I haven't run it just now.


jpfr:
@NoelGraf please review

NoelGraf:
This solution is not compatible with the UAExpert GDS PushView.
I will take a look at the problem and think about a solution as well.

tomitkl:
OK, I wasn't aware of UaExpert's GDS PushView. I will take a look at that too.

However, UaGDS PUSH Model seems to work with these changes.

I have one additional change in applyChangesToServer that I need for UaGDS to work properly with our device i.e. I don't call secureChannel_delayedClose from delayed callback but from a timer of 1 second. That gives time for UaGDS to close the session after ApplyChanges and be happy. This might be an issue on UaGDS side. I think, it should be enough if it gets a reply to ApplyChanges and don't fail it it can't close the session. I haven't investigated that in detail so I'm not sure if it gets the reply or not but with the additional delay it seems to work.

mcpat:
@NoelGraf: I think this refers to issue #6740. We hat the same modifications in our code base and realised that UA expert did not work anymore (because it uses the FileType functions).

NoelGraf:
@mcpat: As you mentioned in your issue, the specification doesn't seem to be clearly defined. I noticed that you asked again on Mantis, so I suggest we wait a moment and see if we get more clarity on this topic.

PR #7033, opened 2025-01-20
refactor build_linux workflow

marwinglaser:
GitHub Actions [support objects as matrix values](https://github.com/github/docs/issues/26469#issuecomment-1760332235) (see below), so lets use this instead of `include:` to simplify the build_linux matrix. The remaining commits are some minor cleanup in the same file.

Note that the linter in the GitHub web editor complains about this, but in this case [it appears like the linter is wrong](https://github.com/github/docs/issues/26469#issuecomment-1892388775)

---

From https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-using-a-multi-dimension-matrix:

> A variable configuration in a matrix can be an array of objects.
>
> ```yaml
> matrix:
>   os:
>     - ubuntu-latest
>     - macos-latest
>   node:
>     - version: 14
>     - version: 20
>       env: NODE_OPTIONS=--openssl-legacy-provider
> ```

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -69,40 +69,40 @@ jobs:
 
           - name: "Encryption (LibreSSL) Build & Unit Tests (gcc)"
             cmd_deps: |
-                sudo apt-get install -y -qq curl
-                curl https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.9.1.tar.gz --output libressl.tar.gz
-                tar -xvz -f libressl.tar.gz
-                cd libressl-3.9.1
-                ./configure
-                sudo make install
+              sudo apt-get install -y -qq curl
+              curl https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.9.1.tar.gz --output libressl.tar.gz
+              tar -xvz -f libressl.tar.gz
+              cd libressl-3.9.1
+              ./configure
+              sudo make install
             cmd_action: unit_tests_encryption LIBRESSL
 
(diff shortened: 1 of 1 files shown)


marwinglaser:
I have no idea how this reduced code coverage when it didn't touch any code or the coverage workflow.

jpfr:
Good to review from your end?

marwinglaser:
Yes :+1:

jpfr:
Top. Rebased.
Any ideas why the codecov coverage statistics doesn't update?
This change was coverage-neutral. But it gets assessed as -.05% coverage change.

marwinglaser:
Codecov used https://github.com/open62541/open62541/commit/f7deafa6ae327d42c80abc19dc12be636dd2ca44 from November last year as a base for the comparison, so it has to be accumulated change since that commit. I have no clue why it did that, though.

https://app.codecov.io/gh/open62541/open62541/pull/7033:
> Coverage data is based on head [cbe2339](https://app.codecov.io/gh/open62541/open62541/commit/cbe2339c97a73b403d7640f459e901117f9fbc8b) compared to base [f7deafa](https://app.codecov.io/gh/open62541/open62541/commit/f7deafa6ae327d42c80abc19dc12be636dd2ca44)

marwinglaser:
I think I figured out why. f7deafa is the last commit codecov has data on for the master branch. I assume this is because codecov now requires a token to upload reports. This is also what the [logs for this merge indicate](https://github.com/open62541/open62541/actions/runs/12886001969/job/35925647442#step:6:942).
I can prepare a patch to pass the token to the codecov action in the relevant workflow, but I don't have the permissions required to configure secrets.

PR #7035, opened 2025-01-21
fix(ci): pass CODECOV_TOKEN to codecov-action

marwinglaser:
I *think* this works and the docs/examples indicate it should, but I don't know how to properly test it.
Also, this requires a GitHub actions secret called `CODECOV_TOKEN` that contains the repository upload token found [here](https://app.codecov.io/gh/open62541/open62541/config/general) (requires authentication). I don't have the permission to modify GitHub secrets, so someone else will have to create that.

The same is the case for the 1.4 release branch. What is the policy for changes that should go into multiple branches? I see there is a [PR to merge 1.4 back into master](https://github.com/open62541/open62541/pull/6946), but that is relatively old and Codecov is broken on both now. 

Context: https://github.com/open62541/open62541/pull/7033#issuecomment-2604607901

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -28,3 +28,5 @@ jobs:
           tree .
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}


jpfr:
Ok. I just generated and added the CODECOV_TOKEN to the secrets.

jpfr:
👍

marwinglaser:
This unfortunately didn't fix it (completely). The workflow now [fails with a different error](https://github.com/open62541/open62541/actions/runs/12890704978/job/35940913953#step:6:957): `error - 2025-01-21 15:57:34,905 -- Upload failed: {"message":"Repository not found"}`

Unfortunately, googling this doesn't produce much useful information. From the bits I could find, this seems to indicate that either the repository isn't configured properly with codecov or the token for GitHub actions is misconfigured.

@jpfr Just to make sure, you said you generated a token? The secret needs to contain a "Repository upload token", which can be found [here](https://app.codecov.io/gh/open62541/open62541/config/general) and does *not* need to be generated, not an "API token" which can be found in your codecov account settings and has to be generated.

jpfr:
I just (again) updated the token to be exactly this.
Please verify.

grafik

Also, when I just checked codecov, the /tests and /examples folders were counting towards the coverage percentage.
They should not!

marwinglaser:
Can you restart the coverage workflow for the latest commit on master? I don't have the permissions to do so and can't verify without a new run.

Also, I setup codecov for my fork to test this and it [works as expected](https://github.com/marwinglaser/open62541/actions/runs/12907883118/job/35994359570#step:6:957), so it has to be the codecov config outside the repo or the token. I don't know which yet (or if it still doesn't work after your last change).

PR #7046, opened 2025-01-24
Move the file handling definitions to the arch folder

NoelGraf:
Move the file handling definitions to the arch folder so that file system operations can be used elsewhere in the future.


jpfr:
👍

NoelGraf:
I'm not sure what you mean. Could you provide more context?

NoelGraf:
Ok all right. We should use <limits.h> and <sys/params.h> instead of <linux/limits.h> as this is not available on all Linux distributions.

NoelGraf:
https://github.com/open62541/open62541/pull/7124

omarhogni:
This seems to get further but stops on 

image

To avoid confusion, this is mac os not a linux distribution.

PR #7047, opened 2025-01-24
Fixed 2 bugs. pubsub parsing, added skipUnknownItem()

Berendej:
If **parseJSONConfig()** met unknown complex 1st level item (array or object) it couldn't parse following items correctly since ctx->index was pointing to the internal elements of unknown item instead of the next 1st level item. So **skipUnknownItem()** function was added which skip all internal elements of unknown 1st level item, be it an object, array of objects, array of array or anything else. The unknown item may happen when some functionality is not supported, but related parameters present in config file. E.g. if **UA_ENABLE_ENCRYPTION** is not defined, but **securityPolicies** item is present in config file.

Also PUBSUBCONFIGURATION]((&ctx, **config**, NULL) was a mistake
it was replaced by
PUBSUBCONFIGURATION](&ctx, **&config->pubSubConfig**, NULL)

And 
#define MAX_TOKENS 256
was small it doesn't allow me to test all variants of skipped items. And number of tokens in server_config.json5 given as an example is already near the MAX_TOKENS. So it can be an obstacle in future.
I propose
#define MAX_TOKENS 1024


jpfr:
Thanks.

we have a skip function here:
https://github.com/open62541/open62541/blob/575d0adca79a90b939befa34cb8790ca92c61fd8/src/ua_types_encoding_json_105.c#L1143

That should work for both objects and arrays.
And is faster because it doesn’t need case distinctions and recursion.

Berendej:
Thank you. I didn't notice it. In order skipObject()  to work it need to be declared as non static.
To be able to use pointer to **ParsingCtx ctx**  as a pointer to **ParseCtx ctx**
inside of skipObject() starting part of ParsingCtx must be identical to starting part of ParseCtx
so I changed it as:
```
typedef struct {
    const char *json;
    const cj5_token *tokens;
    size_t tokensSize; /* set identical type to tokensSize */
    size_t index;
    UA_Byte depth;
    cj5_result result; /* shift result to the end */
} ParsingCtx;
```

jpfr:
That looks good.
Please just copy/paste/rename the skip function for the plugin.
Otherwise this will break in non-visible ways the next time we make changes to the internal ctx struct.

Berendej:
Done

jpfr:
Thanks. Squash-merged to clean up the commit history.

PR #7049, opened 2025-01-25
refactor(cmake): `UA_ENABLE_PUBSUB_INFORMATIONMODEL` defaults to `UA_…

fedefrancescon:
…ENABLE_PUBSUB`

`UA_ENABLE_PUBSUB_INFORMATIONMODEL`  depends on `UA_ENABLE_PUBSUB` .   Defaulting to `UA_ENABLE_PUBSUB` makes it clear that when compiling with PUBSUB off we are also disabling informationmodel without the need to explicitly set both flags


jpfr:
The intention is good. But somehow a lot of things broke for the CI...

fedefrancescon:
Sorry for the late response. Should be fine now, I typoed the variable expansion, not noticing immediately as I was using the UA_ENABLE_PUBSUB=OFF  and seems that mistyped variables default to OFF.

fedefrancescon:
The CI seems to fail on `deps/nodesetLoader/CMakeLists.txt` with ubuntu-20.04 with tcc  
This is out of scope of this PR but, if needed I might take a quick look if useful

fedefrancescon:
Now that master branch fixes some compiler warning/hangs, seems everything is fine

jpfr:
Nice, thanks.

PR #7050, opened 2025-01-25
Lock order

jpfr:


GoetzGoerisch:
@jpfr commit 309079069f0e7c53d5e4914e4411e67355ac01b6 basically reverts 10256ed52842762016c738b88cd133272ae995d3 and breaks compilation on musl libc systems again. Is this intentional?

jpfr:
Yes, because PTHREAD_MUTEX_INITIALIZER is _NOT_ the recursive lock.
However there is only one place in the entire project where this gets used.
So maybe we can do entirely without.

GoetzGoerisch:
Thanks, yes, you are correct, I overlooked this in the first place.
I created an issue to track this, as we need a portable approach IMHO.

PR #7054, opened 2025-01-28
fix(build): build amalgamation target by default if enabled

marwinglaser:
It's reasonable to assume that users that set `UA_ENABLE_AMALGAMATION=ON` expect that to build with a simple `make` (or equivalent for other build systems).


jpfr:
👍

PR #7057, opened 2025-01-29
Send Multicast to multiple hardware interfaces

engel-bichlerh:
If the UA_ServerConfig.mdnsInterfaceIP is not set, take all hardware interfaces and send a multicast to each.


V-Strassheim:
Hi, your PR would break LDS  based on avahi mdns.

I added Avahi a few weeks ago as an alternative mDNS backend. In order to change the existing solution as little as possible, I considered among others the following “discovery API”:
- ua_discovery.c/.h and ua_services_discovery.c/.h -> OPC UA and internal related code
- ua_discover_mdns.c and ua_discovery_mdns_avahi.c -> mdsnd and avahi related implementations

@engel-bichlerh could you also implement the API changes in ua_discovery_mdns_avahi.c?

Now that avahi implementation is in, I should make sure that at least a build test is integrated into the CI for this mdns option.

engel-bichlerh:
Hi, now I added the API changes in the ua_discovery_mdns_avahi.c. I hope it is working fine because I couldn't test the avahi mdns version.

src/server/ua_discovery_mdns.c
```diff
@@ -547,7 +549,7 @@ mdns_find_record(mdns_daemon_t *mdnsDaemon, unsigned short type,
 
 /* set record in the given interface */
 static void
-mdns_set_address_record_if(UA_DiscoveryManager *dm, const char *fullServiceDomain,
+mdns_set_address_record_if(const char *fullServiceDomain,
                            const char *localDomain, char *addr, UA_UInt16 addr_len) {
```
V-Strassheim, line 553: Style: The parameters could be distributed differently over two lines. Now it looks strange.
```suggestion
mdns_set_address_record_if(const char *fullServiceDomain, const char *localDomain,
                           char *addr, UA_UInt16 addr_len) {
```

src/server/ua_discovery_mdns.c
```diff
@@ -976,6 +992,86 @@ addMdnsRecordForNetworkLayer(UA_DiscoveryManager *dm, const UA_String serverName
...
+
+static void
+discovery_createMultiConnections(UA_Server *server, UA_DiscoveryManager *dm,
+                                 UA_KeyValuePair params[], UA_KeyValueMap kvm) {
+    char sourceAddr[NI_MAXHOST];
+    kvm.mapSize++;
+    params[5].key = UA_QUALIFIEDNAME(0, "interface");
+#ifdef _WIN32
+    ULONG outBufLen = 15000;
+    UA_STACKARRAY(char, addrBuf, 15000);
```
V-Strassheim, line 1026: 15000 looks like a magic number. Maybe some define would make sense?
I found this in eventloop_posix_udp.c:
`#define ADDR_BUFFER_SIZE 15000 /* recommended size in the MSVC docs */`

src/server/ua_discovery_mdns.c
```diff
@@ -580,7 +582,7 @@ mdns_set_address_record(UA_DiscoveryManager *dm, const char *fullServiceDomain,
             int family = address->Address.lpSockaddr->sa_family;
             if(AF_INET == family) {
                 SOCKADDR_IN* ipv4 = (SOCKADDR_IN*)(address->Address.lpSockaddr); /* IPv4 */
-                mdns_set_address_record_if(dm, fullServiceDomain,
+                mdns_set_address_record_if(fullServiceDomain,
                                            localDomain, (char *)&ipv4->sin_addr, 4);
```
V-Strassheim, line 586: Style: same like above and a few changes bellow. The parameters could be distributed differently over two lines.
```suggestion
                mdns_set_address_record_if(fullServiceDomain, localDomain,
                                           (char *)&ipv4->sin_addr, 4);
```

V-Strassheim (commented):
Just some minor comments on code style.

V-Strassheim:
> Hi, now I added the API changes in the ua_discovery_mdns_avahi.c. I hope it is working fine because I couldn't test the avahi mdns version.

Thank you very much, I have tested it with avahi and it works (except for a null-pointer access, which I added once, I'll fix that separately)

I also briefly reviewed other changes and added small comments.

engel-bichlerh:
Thanks for review, I will add the issues.

engel-bichlerh:
Hi, it is not possible to use the ADDR_BUFFER_SIZE from eventloop_posix_udp.c in ua_discovery_mdns.c.
Some of my new functions use low level network functions, so I am thinking about moving them to eventloop_posix_udp.c, is this okay for you?

V-Strassheim:
> Hi, it is not possible to use the ADDR_BUFFER_SIZE from eventloop_posix_udp.c in ua_discovery_mdns.c. Some of my new functions use low level network functions, so I am thinking about moving them to eventloop_posix_udp.c, is this okay for you?

Sorry, I meant that instead of the number you could define and use a define like in ua_discovery_mdns.c. Just as an example. Not that you can use the same one since it is defined "locally" in the C file.
There's no need to move anything because of me. The maintainers will add their comments if there is still something to be fixed.

jpfr:
@engel-bichlerh What do you want to move to the UDP layer?

please describe if this will add methods the the API or can be handled in the key-value maps with more optional fields.

PR #7064, opened 2025-01-30
[Review] Fix null pointer dereference in ua_services_discovery.c

V-Strassheim:
UA_DiscoveryManager_getNextServerOnNetworkRecord returns NULL if there is no entry.
Add a break to the entry loop to avoid dereferencing a NULL pointer.


V-Strassheim:
I will update the code if https://github.com/open62541/open62541/pull/7057 is merged first.

jpfr:
Thanks, merged.

PR #7077, opened 2025-02-03
refactor(ci): generate a coverage report for all linux gcc builds

marwinglaser:
This PR changes the CI coverage setup completely. Instead of having one dedicated workflow upload coverage for the subset of features it has enabled, we let all unit test workflows (using 64-bit gcc on linux) generate and upload coverage information. This way all actually tested configurations are run through coverage testing, and we can reuse the existing runs. Codecov is then able to [merge all uploads](https://docs.codecov.com/docs/merging-reports) and produce one aggregated report.

A note on gcc only. The current CMake coverage logic is somewhat of a mess and sets coverage flags in multiple locations and with different conditions. This results in `UA_ENABLE_COVERAGE=ON` [passing `--coverage` ](https://github.com/open62541/open62541/blob/6915f463ed6724e77260c97bbdeb54c2167db225/CMakeLists.txt#L639) to clang, even though it doesn't support this flag (`check_c_compiler_flag` doesn't pass the remaining flags). Clang then fails, because we also pass `-Werror,-Wunused-command-line-argument`. tcc doesn't seem to support coverage at all (and correctly fails the check above), and for some reason I can't get 32-bit gcc to find the gcov library when linking.

Considering gcc runs all tests as well and there is barely any compiler-specific code (at least that I found), this shouldn't have a significant impact.

In the future, the CMake coverage logic should probably be cleaned up a bit to set the proper flags for different toolchains and clang coverage enabled.


jpfr:
Nice. Yes, please clean up CMake where you can.

With this commit the coverage however goes down.
Shouldn't it go up by aggregating more builds together?

marwinglaser:
As far as I can tell, coverage goes down because the ratio of covered/uncovered lines that are now newly visible to codecov are below the current master value.
Here you can see that the absolute number of covered lines does indeed go up (as expected) but so does the number of uncovered lines.
|||
|-|-|
| [master](https://app.codecov.io/gh/open62541/open62541/tree/master) | image |
| [this branch](https://app.codecov.io/gh/open62541/open62541/tree/marwinglaser%2Fopen62541%3Acoverage) |  image |

jpfr:
Got it. Merging.

PR #7078, opened 2025-02-03
fix(ci): remove python-six dependency from doc_upload workflow

marwinglaser:
Backport https://github.com/open62541/open62541/pull/6978 to 1.4 to fix doc upload worflow.


jpfr:
Thanks, merged.

PR #7079, opened 2025-02-03
fix(server): Rename UA_ServerConfig_clean to UA_ServerConfig_clear

RolfKal:
to match other functions that clear an object structure. This has been already applied to master but should be also backported to the 1.4 branch.

### DIFF
--- a/examples/pubsub/sks/server_pubsub_central_sks.c
+++ b/examples/pubsub/sks/server_pubsub_central_sks.c
@@ -515,7 +515,7 @@ cleanup:
     if(server)
         UA_Server_delete(server);
     else
-        UA_ServerConfig_clean(&config);
+        UA_ServerConfig_clear(&config);
 
     UA_ByteString_clear(&certificate);
 #if defined(UA_ENABLE_ENCRYPTION)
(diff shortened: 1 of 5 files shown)


andreasebner:
In general, it's a good idea to have a consistent naming scheme. As the function is part of the exported API, this change would break existing code. Due to our release strategy, only major releases are allowed to break existing code. Therefore, this change will be part of the next 1.5 release, but will not be backported.

RolfKal:
It could be argued that that is only relevant for when open62541 is used as shared library, since recompilation of the program would simply use the also provided UA_DEPRECATED inline wrapper. And considering that shared library support is still not fully functional as shown by the fact that tools/ua2json and examples/server_pubsub_file_configuration both use non-exported APIs in 1.4, I kind of wonder how many people build open62541 for shared library use.

But I will refrain from sending in patches for ua2json and server_pubsub_file_configuration then, as they both would involve renaming of functions to make them the same as the already exported functions in open62541_master.

Another question then would be when can we expect 1.5 to get released?

jpfr:
I agree with @andreasebner .
At least we should not deprecate a function in a minor release.

let’s do this on master.
V1.5 is expected in a few months.

RolfKal:
Well this one is already in master. It was sent by me as a patch last year and committed. I was just trying to compile my own library which I had developed against master with 1.4 open62541 and I came across this.

RolfKal:
Closing as Don't Change.

Issue #7074, opened 2025-02-03
Deadlock with separate PubSub event loop


keba-estr:
## Description

In my software I am using two separate event loops for the server and the PubSub part. The second event loop is created with UA_EventLoop_new_POSIX and has a separate event loop mutex.
The following tasks use the open62541 interfaces and are in a deadlock state:

- Server task:
   calls UA_Server_run with server event loop
   takes event loop mutex 1 in UA_EventLoopPOSIX_run
   unlocks event loop mutex 1 in UA_EventLoopPOSIX_pollFDs (before epoll)
   waits for event loop mutex 1 in UA_EventLoopPOSIX_pollFDs (after epoll)

- Other task:
   wants to configure PubSub by calling UA_Server_removePubSubConnection
   Callstack: UA_PubSubConnection_delete -> UA_PubSubConnection_setPubSubState -> UA_PubSubConnection_disconnect -> UDP_shutdownConnection -> waits for EventLoop mutex 2
 
- PubSub task:
   calls separate PubSub EventLoop (created with UA_EventLoop_new_POSIX) -> UA_EventLoopPOSIX_run (locks event loop mutex 2)
   Callstack: processDelayed -> UDP_delayedClose -> UDP_close -> PubSubChannelCallback -> lockServer -> UA_EventLoopPOSIX_lock -> waits for same EventLoop mutex 1 as the server task

-> this callstack should use the event loop mutex 2 of the PubSub EventLoop, not the server event loop!

Still worked with version from commit 42b58674bcf6afc3a3465e81bf95523379c45ade.

## Checklist

Please provide the following information:

 - [X] open62541 Version (release number or git tag): 1.4.10

keba-estr:
When calling PubSubChannelCallback the event loop mutex 2 is already locked, so only the server mutex has to be taken into account here.

keba-estr:
UA_Server_removePubSubConnection locks server mutex
UDP_shutdownConnection tries to lock event loop mutex 2

PubSub-Task with UA_EventLoopPOSIX_run locks event loop mutex 2
PubSubChannelCallback tries to lock server mutex

-> deadlock

keba-estr:
Event loop mutex must always be locked before server mutex.

Possible solution:
Use separate functions lockPubSubServer / unlockPubSubServer that locks the PubSub connection eventloop mutex and the server mutex.

To check:
- is connection available at every PubSub call, else: lock all connections
- Does this work?

keba-estr:
[...]

PR #7084, opened 2025-02-04
[Review] fix(eventloop): Fix deadlock with event loop mutex

keba-estr:
As specific connection is not known yet, lock event loop mutexes of
all connections.
When adding and removing a connection from the server list, lock/unlock
the mutex to match the locking counter.

Resolves https://github.com/open62541/open62541/issues/7074

### DIFF
--- a/examples/pubsub_realtime/server_pubsub_publisher_rt_level.c
+++ b/examples/pubsub_realtime/server_pubsub_publisher_rt_level.c
@@ -176,9 +176,10 @@ int main(void) {
     pubSubEL->stop(pubSubEL);
     while(pubSubEL->state != UA_EVENTLOOPSTATE_STOPPED)
         pubSubEL->run(pubSubEL, 0);
-    pubSubEL->free(pubSubEL);
 
     UA_Server_delete(server);
+
+    pubSubEL->free(pubSubEL);
 
     return retval == UA_STATUSCODE_GOOD ? EXIT_SUCCESS : EXIT_FAILURE;
 }
--- a/src/pubsub/ua_pubsub.h
+++ b/src/pubsub/ua_pubsub.h
@@ -184,7 +184,7 @@ UA_PubSubConnection_create(UA_Server *server,
 void
 UA_PubSubConnectionConfig_clear(UA_PubSubConnectionConfig *connectionConfig);
 
-void
+UA_StatusCode
 UA_PubSubConnection_delete(UA_Server *server, UA_PubSubConnection *c);
 
 UA_StatusCode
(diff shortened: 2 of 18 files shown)


jpfr:
Thanks for this.
PR #7085 is a bit more lightweight.
Does that work for you also?

keba-estr:
> Thanks for this. PR #7085 is a bit more lightweight. Does that work for you also?

No, PR #7085 is not enough, it does not work for me.
The problem is that the dc callback is still called with the locked mutex, there are no changes for the called functions in UA_PubSubConnection_disconnect or processDelayed with that PR.

Also, if a separate eventloop is used for PubSub, what is the cause for also locking/unlocking the server eventloop mutex with the functions lockServer/unlockServer in the PubSub files?

keba-estr:
See discussion in issue #7074

keba-estr:
@jpfr: What is the cause of the failing builds?

jpfr:
Two quick comments:

- You can rename ua_pubsub.h to ua_pubsub_internal.h. It is used exactly like that already.
- There is a lot of repetition in the added code. You could pull it into a commonly used function.
For example ``lockPubSubServerWithConnection``. Where the EventLoop to be used is selected inside that function.

keba-estr:
> Two quick comments:
> 
> * You can rename ua_pubsub.h to ua_pubsub_internal.h. It is used exactly like that already.
> * There is a lot of repetition in the added code. You could pull it into a commonly used function.
>   For example `lockPubSubServerWithConnection`. Where the EventLoop to be used is selected inside that function.

I would prefer to add the functions to ua_pubsub.h then without renaming any files. This minimizes the needed changes.

keba-estr:
The order of the mutexes is still not used consistently.
In lockPubSubServer the event loop mutex is locked and unlocked before the server mutex.
Yet the function callstack UA_Server_run -> UA_Server_run_shutdown -> UA_Server_run_shutdown -> UA_PubSubManager_shutdown -> UA_PubSubConnection_setPubSubState -> UA_PubSubConnection_disconnect ->UDP_shutdownConnection
first locks the server mutex in UA_Server_run_shutdown and then the event loop mutex in UDP_shutdownConnection, which sill causes a deadlock.

keba-estr:
> @andreasebner What is the reason for the failed builds? Not all failures are caused by the changed code. Do I have to update my profile to be able to work with docker?

Build problems seem to be resolved.

keba-estr:
@andreasebner: https://open62541.org/libmbedtls/libmbedcrypto0_2.6.0-1_amd64.deb seems to be missing on the new homepage acoording to the build failure.

jpfr:
> @andreasebner: https://open62541.org/libmbedtls/libmbedcrypto0_2.6.0-1_amd64.deb seems to be missing on the new homepage acoording to the build failure.

Yes. The fix will arrive shortly.

keba-estr:
Ready for review and merge.

jpfr:
TODO: Review existing PR after discussion with @keba-estr and @keba-ebr.

```
	1.	PubSub EventLoop Lock
	2.	Server EventLoop Lock
	3.	Server Lock (ServiceMutex)

1. Idea: Fixed order for taking all locks

2. Idea: PubSub EventLoop gets released before calling into the server
```

keba-estr:
According to Mr. Julius Pfrommer usage of a separate PubSub event loop is a special case which is only used by a small number of companies, there will be no effort taken for this issue.
This problem does not appear on the 1.5 brach as the specific event loop has been removed and the user is responsible for handling the customstatemachine.

PR #7090, opened 2025-02-04
improve(tests): building tests under Windows

RolfKal:
findCheck() under Windows failed to find a normally installed check.lib library
make ua_add_test() macro more versatile to allow extra parameters


jpfr:
👍

PR #7105, opened 2025-02-12
fix(ci): fix release workflow after bad merge

marwinglaser:
These are the small required changes mentioned in https://github.com/open62541/open62541/pull/6956, which [broke the release workflow on master](https://github.com/open62541/open62541/actions/workflows/release.yml?query=branch%3Amaster) after https://github.com/open62541/open62541/pull/6946


jpfr:
👍

Issue #6344, opened 2024-03-13
Documentation version number: master branch


jeroendoggen:
## Description
The [documentation on the master branch ](https://www.open62541.org/doc/master/) currently shows the version number '1.3'
The URL makes it clear that this actually is the documentation for the 'master' branch. (currently ~v1.4.0-rc2)
Changing the number to '1.4 beta' or 'master' or something similar might avoid some confusion.

## Checklist
Please provide the following information:
 - [x] open62541 Version: v1.4.0-rc2

andreasebner:
Thank you for reporting. Fixed in #7106 .

PR #7106, opened 2025-02-13
rewrite sphinx configuration

marwinglaser:
The current sphinx configuration has been unmaintained, was broken in some regards and has gotten out of date. This PR recreates it from scratch, cleaning it up and adapting it to current sphinx and sphinx_rtd_theme versions.

This includes roughly the following changes:
- install sphinx and sphinx_rtd_theme using pip and add version constraints to prevent breaking changes hitting the ci build without notice (and configure dependabot to open PRs for that)
- recreate a base config using `sphinx-quickstart` (this already no longer contains all options by default)
- port over the custom configuration that actually matters (e.g., no default values)
- fix search by actually including the theme as an extension
- fix missing copyright in html output
- remove unused source files from the html output (this was introduced in 33f1054b3cf20c1cce898ee51767bd175abe4213 without reasoning; the `rst` files in `_sources` still exists)

(This is not all cleanly separated into separate commits to recreate and fix because I didn't want to artificially create a broken middle commit.)

I didn't copy over all the commented or defaulted config options as this does not make any sense if no one maintains the file. Part of the reason I recreated the entire thing instead of doing the minimally invasive fix is that I found multiple options that didn't exist anymore or were renamed but were either still being set or in the file commented out. I assume this is also the reason why `sphinx-quickstart` doesn't generate them anymore.

I compared the resulting html and pdf documentation to verify I didn't break anything using the versions specified in `requirements.txt` installed in a python virtual environment as well as in a debian container using the versions from debian's stable repository (sphinx 5.3, sphinx_rtd_theme 1.2).
According to diffpdf, the resulting pdf looks identical to the one on current 1.4 minus the version string on the title page (obviously, this is a different commit), and, manually comparing exiftool output, metadata is the same as well.
For the html output I manually reviewed the recursive diff (`diff -r ...`) and it is identical except for the expected changes (copyright, js included for search, permalink symbol from the rtd theme instead of sphinx) and changes in the doctree cache files.

I'm unsure if this should actually go to the 1.4 branch. I verified that nothing breaks as noted above and it does fix some longstanding issues that might be nice having in 1.4, but I can re-target to master and open a PR with minimal fixes for 1.4 if preferred.

---

~WIP while I am still checking whether we can somehow get the version back (broken upstream in sphinx_rtd_theme 3).~


jpfr:
Nice. Much better.
Since the config is in Python, can't you read out the git meta-data and do a case-distinction for the master branch?
Like `branch (commit-hash) date` whenever the current commit is not also a version tag.

marwinglaser:
I guess I could, I didn't think about that. In case git isn't available (or it's built without the repo) I guess I can just fallback to `version` again.

I've gone through some historical revisions in the www repo and as far as I can tell master always showed the current version (see also https://github.com/open62541/open62541/issues/6344), as it has probably been using the same version variable.

> `branch (commit-hash) date`

That would get awfully long, though. As can be seen in the images we are working with limited space. As a quick test I tried with `release` (which is just reconstructed `git describe`) which should be similar length, and I'm not sure this is a good idea:
image

Maybe just current branch and abbreviated commit is enough for now? This is only a workaround for a missing dedicated field after all.

marwinglaser:
The latest commit tries to get the branch from git, falling back to the version cmake passes on any error (like git not being available or not being built from a repository). This should still work for release branches, as they are named according to the release. Since the commit is in `release` and therefor the page titles, this may be enough?
I'm not sure if this would better be done from CMake, though, so all git version parsing happens in one file?

image

PR #7112, opened 2025-02-18
feat(ci): remove unused azure pipeline configs

marwinglaser:
WIP. Opening to test against live azure config.


marwinglaser:
I've re-marked this as WIP because I remembered that at least this line https://github.com/open62541/open62541/blob/master/azure-pipelines.yml#L12 and other potential references (I haven't `grep`-ed yet) should be removed as well, and I haven't gotten around to it so far. Apart from that, it's fine to merge, as far as I know.

jpfr:
Nice, merged.

PR #7114, opened 2025-02-18
feat(ci): run valgrind checks on ubuntu-24.04 (current latest) only

marwinglaser:
Targeting master as that's where all the resources are used. This should likely be backported to 1.4 as well, but that needs to be done manually, as there are a bunch of conflicts because https://github.com/open62541/open62541/pull/7033 happened on master only (my bad). Maybe I should backport that as well?


jpfr:
It’s okay to have this on master only for now.

PR #7115, opened 2025-02-18
[WIP] feat(ci): remove ubuntu-20.04 ci

marwinglaser:
GitHub is deprecating ubuntu-20.04 runners. See https://github.com/actions/runner-images/issues/11101.

1.4: https://github.com/open62541/open62541/pull/7134
1.3: https://github.com/open62541/open62541/pull/7133

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04, ubuntu-24.04]
+        os: [ubuntu-22.04, ubuntu-24.04]
         build:
           - name: "Debug Build & Unit Tests (gcc)"
             cmd_deps: sudo apt-get install -y -qq mosquitto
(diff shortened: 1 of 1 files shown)


andreasebner:
Since the ubuntu-20 runners are no longer available from github, should we not remove them from all maintends starting with 1.3?

jpfr:
Yes, let’s remove from 1.3 already.

marwinglaser:
> Since the ubuntu-20 runners are no longer available from github, should we not remove them from all maintends starting with 1.3?

Absolutely, this was part of why this is marked WIP. Unfortunately, the workflow config is very different between 1.3/1.4 and master. I have opened additional PRs for both branches with the respective changes and linked between all three (including this one) in the descriptions. This should make it easy to resolve the inevitable merge conflicts when merging back by just using the version that's already there.

marwinglaser:
(I noticed I already broke that promise in https://github.com/open62541/open62541/pull/7114, but the fix would still fit here and this also still applies to the release branches.)

marwinglaser:
Can this be merged? The updated schedule says 20.04 images will be removed on the 15th, we are running out of time. The failing Azure Pipelines job should be unrelated.

jpfr:
We regularly merge **all commits** from 1.3 -> 1.4 -> master.

Since we also need this for the release branches, it needs to go there first.

Otherwise this just generates double work.

Any breakage is resolved during the merge, not in separate commits.

andreasebner:
Replaced by #7198

PR #7116, opened 2025-02-18
Move the server json configuration example to the examples folder

NoelGraf:


andreasebner:
The change is great. Perhaps we should use the revision to name the folder more logically. e.g. filebased_json_confguration or configuration_json and put both the config and the example in the folder.

jpfr:
/examples/json_config

PR #7118, opened 2025-02-19
Support defining UA_LOGLEVEL in an external CMakefile

bisbastuner:
Allows to compile the library as a amalgamated open62541 .c/.h pair, and then change the UA_LOGLEVEL in the CMakefile of the project that's using the amalgamated files.

This was already pulled with https://github.com/open62541/open62541/pull/6068, but then was reverted by https://github.com/open62541/open62541/commit/e907a0ed91427e449ed711bada7a5db3947cd644 .
I just tried creating the amalgamated files from the official 1.3.13 release, and this little edit is still needed before the open62541 .h/.c files can be compiled without errors.
This would, allow to use untouched files from the official open62541 repository.

### DIFF
--- a/include/open62541/config.h.in
+++ b/include/open62541/config.h.in
@@ -39,7 +39,9 @@
  * ---------------
  * Changing the feature options has no effect on a pre-compiled library. */
 
+#ifndef UA_LOGLEVEL /* allows overriding UA_LOGLEVEL from an external Makefile */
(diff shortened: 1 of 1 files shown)


bisbastuner:
I have seen 1.4.12 is out, but does not include this PR? When do you think this would be included? Should I need to wait for 1.5? @jpfr

PR #7121, opened 2025-02-19
fix(plugins): remove unnecessary function call to clear the CertificateVerification structure

RolfKal:
- UA_CertificateVerification_AcceptAll() and UA_CertificateVerification_Trustlist() internally already call the clear function.
- Only call UA_CertificateVerification_Trustlist() if we have a trustlist and/or revocation list to add to the client config structure

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -439,12 +439,8 @@ setDefaultConfig(UA_ServerConfig *conf, UA_UInt16 portNumber) {
 
     /* Certificate Verification that accepts every certificate. Can be
      * overwritten when the policy is specialized. */
-    if(conf->secureChannelPKI.clear)
-        conf->secureChannelPKI.clear(&conf->secureChannelPKI);
     UA_CertificateVerification_AcceptAll(&conf->secureChannelPKI);
 
-    if(conf->sessionPKI.clear)
-        conf->sessionPKI.clear(&conf->sessionPKI);
     UA_CertificateVerification_AcceptAll(&conf->sessionPKI);
 
     /* * Global Node Lifecycle * */
@@ -992,8 +988,6 @@ UA_ServerConfig_setDefaultWithSecurityPolicies(UA_ServerConfig *conf,
         return retval;
     }
 
-    if(conf->sessionPKI.clear)
-        conf->sessionPKI.clear(&conf->sessionPKI);
     retval = UA_CertificateVerification_Trustlist(&conf->sessionPKI,
                                                   trustList, trustListSize,
                                                   issuerList, issuerListSize,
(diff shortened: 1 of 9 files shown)


jpfr:
I see what you are changing.
But why?

This needs more motivation.
And the change needs to be reflected in the public API documentation.
As you are changing behavior which is not a bug per se.

RolfKal:
> I see what you are changing. But why?
> 
> This needs more motivation. And the change needs to be reflected in the public API documentation. As you are changing behavior which is not a bug per se.

The removal of the clear function should be totally non controversial. They just do what the functions later on will attempt to do anyhow.

As to calling the UA_CertificateVerification_Trustlist() function inside UA_ClientConfig_setDefaultEncryption() with an empty trustlist and revocation list: It will result in a connection that trusts nobody and will result always in an 0x801A0000 error as there is nothing to trust and the trust check will simply fail as it finds no trusted certificate to match the server certificate to. It's basically destructing the UA_CertificateVerification_AcceptAll() call earlier performed in the UA_ClientConfig_setDefault() for no good reason.

jpfr:
> The removal of the clear function should be totally non controversial. They just do what the functions later on will attempt to do anyhow.

These are plugins. The default plugins call _clear beforehand.
But users might have developed their own plugins which don't do that.
If it is now expected that plugins call _clear internally, then this needs to be added to the interface documentation.
And this is a change I would rather do an master and not for a patch release.

> As to calling the UA_CertificateVerification_Trustlist() function inside UA_ClientConfig_setDefaultEncryption() with an empty trustlist and revocation list:

We can do that. But it needs a big UA_LOG_WARNING when the checks are actually disabled.
Plus an entry to the changelog and to the API documentation.

RolfKal:
You bring up some interesting points indeed.


There are already calls in current 1.4 to
UA_CertificateVerification_AcceptAll() and
UA_CertificateVerification_Trustlist() that are NOT preceeded by a call to
the clear function (e.g.
UA_ServerConfig_setDefaultWithSecureSecurityPolicies(),
tests\encryption\check_username_connect_none.c,
\tests\pubsub\check_pubsub_sks_pull.c,
tests\pubsub\check_pubsub_sks_push.c, \tests\server\check_discovery.c
(all),  tests\client\check_client_authentication.c
(all), tests\client\check_client_encryption.c (setup), and all the other
encryption specific tests in the setup() routine. So I felt it was already
safe to not call the clear before calling any of these.

So it would seem that removal of the other clear calls simply makes it more
consistent.

Adding an UA_LOG_WARNING would seem an easy enough addition. Current
operation is definitely very weird as calling
UA_ClientConfig_setDefaultEncryption() with empty lists is a valid
operation, but will render the connection uncapable of connection to any
server without calling UA_CertificateVerification_AcceptAll() again
afterwards.

jpfr:
The documentation of the plugin interfaces is here: https://github.com/open62541/open62541/blob/1.4/include/open62541/plugin/pki.h

It says explicitly:
> The ``clear`` method is called automatically when the config is destroyed.

The documentation of the default plugin implementations is here: https://github.com/open62541/open62541/blob/1.4/plugins/include/open62541/plugin/pki_default.h

jpfr:
You could add to the API documentation:

During initialization, a CertificateVerification-plugin should _clear any previous plugin instance sitting in the same memory location.

But that will of course crash if the memory was not zeroed out before the first plugin instance is put into place.

RolfKal:
> You could add to the API documentation:
> 
> > During initialization, a CertificateVerification-plugin should _clear any previous plugin instance sitting in the same memory location.
> 
> But that will of course crash if the memory was not zeroed out before the first plugin instance is put into place.

The logic of that somehow escapes me. Those default plugin implementations now do a _clear call on entry. So they WILL crash if the memory passed in is not zeroed already. Calling a _clear before calling into these functions as it is done in the examples and some other places won't fix that. It would be different if it was an _init call, which just zeroes out the memory; that would not be redundant, but as it is now, there are quite a few _clear calls before calling the plugin method which again does a _clear call. It's not a problem or will it cause a crash, but it is simply unnecessary. The only thing as you correctly point out, is that the current documentation doesn't state that the plugin methods are supposed to clear/clearing the information before doing anything else.

Of course memory management is always tricky, but adding _clear calls before calling a plugin which in its default implementation clears that memory too, just because someone somewhere might implement a plugin that does not do that clearing, sounds pretty patronizing.

Issue #7127, opened 2025-02-22
Error Masking after `UA_SecureChannel_receive` During Certificate Validation Failure


aaghash17:
## Description  

In v1.3.15, when testing negative scenarios for certificate validation, the error returned by `UA_SecureChannel_receive` is overridden with `UA_STATUSCODE_BADCONNECTIONCLOSED`.  

This behavior results in misleading error propagation, making it difficult to diagnose certificate validation failures programmatically.  

## Background Information / Reproduction Steps  

1. Configure the client to connect to a server with an invalid or untrusted certificate.  
2. Observe the log output, which correctly captures the failure reason from `UA_SecureChannel_receive`.  
3. However, the returned status to the `connect` method is always `UA_STATUSCODE_BADCONNECTIONCLOSED`, rather than the actual certificate validation failure code.  

### Reference Code (from `ua_client.c`)  
[Source Link](https://github.com/open62541/open62541/blob/86aa417aabe9e52f519119edd7a794a49d3907cb/src/client/ua_client.c#L431-L442)  

### Example Code
```c
int main(void) {
    UA_StatusCode retval;
    retval = create_Certificate();
    if (retval != UA_STATUSCODE_GOOD)
    {
        printf("Error creating certificate, status code: %s\n", UA_StatusCode_name(retval));
        return retval;
    }
    UA_Client* client = UA_Client_new();
    UA_ClientConfig* config = UA_Client_getConfig(client);
    config->securityMode = UA_MESSAGESECURITYMODE_SIGNANDENCRYPT;
    config->securityPolicyUri = UA_STRING_ALLOC("http://opcfoundation.org/UA/SecurityPolicy#Basic256Sha256");
    retval = UA_ClientConfig_setDefaultEncryption(config, certificate, privateKey, nullptr, 0, nullptr, 0);
    if (retval != UA_STATUSCODE_GOOD)
    {
        printf("Error setting encryption, status code: %s\n", UA_StatusCode_name(retval));
        return retval;
    }
    retval = UA_Client_connectUsername(client, "opc.tcp://localhost:51510/UA/DemoServer", "usr", "pwd");
    if (retval != UA_STATUSCODE_GOOD)
    {
        printf("Error connecting to server, status code: %s\n", UA_StatusCode_name(retval));
        return retval;
    }
    return 0;
}
```

### Log
```
[...]

PR #7130, opened 2025-02-24
Update ua_client.c to fix error masking

aaghash17:


jpfr:
Please describe the change.
Can we make a unit test that breaks before, but doesn’t with the change?

aaghash17:
@jpfr #7127

aaghashr:
#7223 
New PR created for the same

PR #7133, opened 2025-02-24
feat(ci): remove ubuntu-20.04 ci (1.3)

marwinglaser:
GitHub is deprecating ubuntu-20.04 runners. See https://github.com/actions/runner-images/issues/11101.

This also removes the now unnecessary ubuntu2204 workflow that only built one configuration.

master: https://github.com/open62541/open62541/pull/7115
1.4: https://github.com/open62541/open62541/pull/7134


jpfr:
Let's do the change for 1.3 first.
And then do a general merge 1.3 -> 1.4 -> master.

marwinglaser:
1.3 -> 1.4 will be easy, but master completely changed the linux workflow(s). I would suggest fixing master separately now, or whoever merges them will have to recreate #7115 during conflict resolution anyway.

jpfr:
We regularly merge "all" commits from 1.3 -> 1.4 -> master.
So the conflict resolution happens anyways.

marwinglaser:
I know about the periodic merges, but as I said, master changed the config a lot. The diffs for 1.3/1.4 and master don't share a single hunk and master has some additional ones. This wouldn't be a regular merge conflict, but essentially throwing out the REMOTE side and recreating it manually, as I had to do when I tried cherry-picking from #7115 to this branch. I wont push it though, in the end I don't really care how it ends up in master.

I've also been working on supporting older ubuntu versions through containerization (no PR yet) so we can build on an ancient version. This currently struggles with a too old version of git in 16.04, but it should almost work for ubuntu 20.04. We may want to consider this as an alternative to moving versions, considering the ci failures, only some of which I can explain so far (e.g., openssl 3.0).

jpfr:
> I've also been working on supporting older ubuntu versions through containerization 

Yes please!
We have users who are stuck with rather old build chains.

jpfr:
Restarted the build.

Are the failing builds a systematic problem or only bit-rot due to version dependencies?

marwinglaser:
I did some googling and found the exact same valgrind pattern suppressed in other projects. I went ahead and added a suppression for our tests (and tests only) so this PR passes CI, but would appreciate another look by someone else.

I don't think the container branch will be ready in time because of some weird networking issues on github's runners (it works locally in containers), so this PR is the likely candidate to avoid breaking CI when github removes 20.04 runners.

jpfr:
👍

marwinglaser:
@jpfr Can you merge 1.3 -> 1.4 to get this into 1.4, schedule says 20.04 images will be removed on the 15.? #7134 contains two additional commits that are necessary for 1.4 ci to work on 1.4 (93b3d281658e455a27fc69c346959ad84e18c33c and ed9d9695d1045be08397a29610892748644cc6dd).

PR #7137, opened 2025-02-26
[REVIEW] Use trustlist methods in gds informationmodel instead of filetype methods

NoelGraf:
1) If the object does not have a direct hasComponent reference to the method node, traverse its object type and all supertypes to check for a valid reference. This ensures that methods can be invoked on instances of the object type or its subtypes.
2) Use trustlist methods in gds informationmodel instead of filetype methods


tomitkl:
Thanks for working on this. I tried the changes and it seems to work ok with UaGDS now but I have an issue with UaExpert's GDS Push View. UaExpert doesn't seem to be able to read the trustlist. OpenWithMasks method is found but not Read and Close. See the methods calls that UaExpert calls in the attached Wireshark capture: [read_trust_list.zip](https://github.com/user-attachments/files/19005617/read_trust_list.zip)

tomitkl:
With the latest change it seems to work with both UaGDS and UaExpert 👍. I will close my PR #7026 as this should fix the issue once the changes are in master.

NoelGraf:
Thanks for your feedback!

jpfr:
Nice, merged. 👍

PR #7150, opened 2025-03-06
Fix batch timers

jpfr:


src/server/ua_server_async.c
```diff
@@ -178,7 +178,7 @@ UA_AsyncManager_init(UA_AsyncManager *am, UA_Server *server) {
     /* Add a regular callback for cleanup and sending finished responses at a
      * 100s interval. */
```
keba-estr, line 179: The time stated in the comment is wrong.

PR #7154, opened 2025-03-07
Removing the mbedtls plugin's architecture dependency

NoelGraf:


jpfr:
This will likely fail in the linker on systems where the standard library has inet_pron and is statically linked. Please prefix to make the method unique.

jpfr:
Looks good.
MSVC is still complaining.

jpfr:
image

jpfr:
👍

PR #7165, opened 2025-03-15
Update config.h.in

jackybek:
#define UA_UNLIKELY(x) (!x)


jpfr:
This is wrong.
The condition should not be inverted in the macro.

systems with no special treatment of likely/unlikely should just ignore the annotation.

jackybek:
But if I don't make this change, it will keep preventing me to instantiate the opc server, because it will keep returning False.

PR #7167, opened 2025-03-15
Update CMakeLists.txt

jackybek:
added the following:
create_certificate.c
certificategroup.c


NoelGraf:
The files you have added are already included.

PR #7172, opened 2025-03-19
Removing architecture dependency from the filestore backend plugin

NoelGraf:
Removed inclusion of unguarded architecture-specific headers from the filestore backend plugin


jpfr:
👍

PR #7173, opened 2025-03-19
fix(core): add missing symbol _UA_Client_readValueAttribute

jbbjarnason:
Stumbled upon this missing symbol when building examples on mac

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -30,6 +30,9 @@ jobs:
             cmd_deps: brew install mbedtls
             cmake_options: -DBUILD_SHARED_LIBS=ON -DUA_ENABLE_ENCRYPTION=MBEDTLS -DUA_ENABLE_AMALGAMATION=ON -DUA_ENABLE_INLINABLE_EXPORT=ON
 
+          - name: "Build examples with INLINABLE_EXPORT=ON"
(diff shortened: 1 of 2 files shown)


andreasebner:
Could you provide more details about your configuration? We've recently added macOS to our CI pipeline, and the build has been consistently green there without the proposed change.

jbbjarnason:
@andreasebner here is a failing workflow before including this header, let me know when you have recognized it than I will put it back in and clean the commits

https://github.com/open62541/open62541/actions/runs/13961091730/job/39082414431?pr=7173#step:4:194

jbbjarnason:
@andreasebner I have removed the commit which showed this failing, the include fixes that. Please review when you have time. Thanks.

jpfr:
Fixing this is good.
But the include should go to a different place.

We will not introduce a dependency to /client from /util.
/util should be independent from both client and server.

jbbjarnason:
btw, this is error on linux aswell, the symbol is just missing but there is not a compilation error.

jbbjarnason:
@jpfr What do you think about creating new source file like ua_inlinable_impl.c which would include headers and with the `UA_INLINABLE_IMPL` definition to expose symbols for inlinable methods? That would be used for util & server & client.

jpfr:
We took this to heart and completely removed the UA_INLINABLE definitions from the client API.
So solved on master.

PR #7177, opened 2025-03-20
dllexport default to inlinable cmake option

jbbjarnason:
I presume when a user of this library chooses to export inlinable methods he would like to get those exports even though he is not building shared library.

At least this is beneficial for me when I static link open62541 to the executable than I can dynamically load the executable and use the library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -459,7 +459,10 @@ endif()
 ##########################
 
 # Building shared libs (dll, so). This option is written into ua_config.h.
-set(UA_DYNAMIC_LINKING OFF)
+# Default is to use inlinable export, so if you choose to export inlinable methods,
+# it will dllexport them. This means that if you build a static lib you will be able to
+# link and include all symbols, like using /WHOLEARCHIVE.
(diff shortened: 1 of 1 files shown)


jpfr:
When you choose to export inlinable methods, they get a UA_EXPORT attribute automatically.

https://github.com/open62541/open62541/blob/cae846b1e4504af6ee4953a131bb5fe96eaf38af/include/open62541/config.h.in#L194

I would prefer to keep the two mechanisms for inlining and Symbol Export orthogonal / decoupled.

jbbjarnason:
> When you choose to export inlinable methods, they get a UA_EXPORT attribute automatically.
> 
> https://github.com/open62541/open62541/blob/cae846b1e4504af6ee4953a131bb5fe96eaf38af/include/open62541/config.h.in#L194
> 
> I would prefer to keep the two mechanisms for inlining and Symbol Export orthogonal / decoupled.

How would you propose setting UA_EXPORT to `define UA_EXPORT __declspec(dllexport)` when statically linked library is made on windows?

PR #7180, opened 2025-03-22
Fix busy loop

jpfr:


MarkusBeerwerth (approved):
Thanks, that should fix it👍

PR #7192, opened 2025-04-03
[Review] fix(plugin): Login callback should be called also with anonymous user

maikrou:
If a user has logged in, and then changes (does not logout) to anonymous user, then no callback is given to handle this case, possibly having invalid user access permissions on application level.


jpfr:
Does this apply to 1.4?

maikrou:
Yes, this also applies to 1.4

PR #7195, opened 2025-04-03
fix(plugin): access control default with login callback

maikrou:
- Access control default with login callback and no hard coded users did not increase policy size correctly and so does not work 
- Removed fixed UA_UsernamePasswordLogin from UA_AccessControl_defaultWithLoginCallback argument
  - If a callback is used, there should be no need to hard code any users
- Disable warning print about leaking credentials in case security policy none is only used for discovery
- Removed unnecessary warning print from default access control


maikrou:
Affects also 1.4, closing and created
https://github.com/open62541/open62541/pull/7214

PR #7196, opened 2025-04-03
[Review] fix(core): use of null pointer in certificate verification

maikrou:
In case UA_ENABLE_ENCRYPTION is not set and a client tries to connect with a certificate, channel->certificateVerification is set, but certificateVerification->verityCertificate callback is (most likely) not set, causing null pointer dereference.


maikrou:
Closing this PR as also affects 1.4 branch:
https://github.com/open62541/open62541/pull/7213

PR #7202, opened 2025-04-07
feat(server): make updating private keys with GDS Push optional

maikrou:
- Servers should allow transfer of private keys only if the server does not have access to the entropy necessary for creating private keys

OPC 10000-12:

> The SupportedPrivateKeyFormats specifies the PrivateKey formats supported by the Server. Possible values include “PEM” (see RFC 5958) or “PFX”. The array is empty if the Server does not allow external Clients to update the PrivateKey.

- Fix max trust list size to use the configured maximum size

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -193,6 +193,12 @@ struct UA_ServerConfig {
      * Make sure you really need this before enabling plain text passwords. */
     UA_Boolean allowNonePolicyPassword;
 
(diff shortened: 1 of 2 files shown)


jpfr:
Looks good.
Please set a commit message header according to our /Contributing.md document.

maikrou:
I think the commit message is now correct.
Lots of tests are failing but they are unrelated to the changes

PR #7203, opened 2025-04-07
[Review] fix(server): local variable not referenced

maikrou:
Fix compiler warning for local value not referenced in case some compile flags are not set.


jpfr:
Does this also apply to 1.4?

maikrou:
Yes, also applies to 1.4

jpfr:
Okay. Please make all fixes that apply to the 1.4 branch against that.
We will then pull all commits from 1.4 up to master.

jpfr:
You can force-push to the branch and reassign the PR target. No need to create a new PR to change the target branch.

PR #7207, opened 2025-04-08
[Review] fix(server): fix duplicate entries in discoveryUrls list

maikrou:
Fix issue where the same URL is in the discovery URLs list multiple times:
image


maikrou:
Created PR for 1.4 branch as it affects that as well:
https://github.com/open62541/open62541/pull/7212

and closing this one.

PR #7208, opened 2025-04-08
[Review] fix(server): local variable not referenced

maikrou:
Fix compiler warning for local value not referenced in case some compile flags are not set.


jpfr:
Merged. Thanks!

PR #7209, opened 2025-04-08
[Review] fix(plugin): login callback for anonymous user

maikrou:
If a user has logged in, and then changes (does not logout) to anonymous user, then no callback is given to handle this case, possibly having invalid user access permissions on application level.


andreasebner:
Merged with 8166

PR #7212, opened 2025-04-09
[Review] fix(server): fix duplicate entries in discoveryUrls list

maikrou:
Fix issue where the same URL is in the discovery URLs list multiple times:
image


jpfr:
Merged. Thanks!

PR #7220, opened 2025-04-12
[Review] fix(build): adds C++ support for C11 atomics

### LINKED ISSUE #7076: Cannot use the miltithreaded version in C++ project
## Description

I try to use the 1.4.10 version in my C++ project. When the library is built as single-threaded (UA_MULTITHREADING=99), everythins is OK. However, having set UA_MULTITHREADING=101, the library is built succesfully, but when compiling my project, I get an error "‘_Atomic’ does not name a type". To fix this error, it seems to be enough to edit the **config.h** file: the line

`#    include <stdatomic.h>`

replace with the following text:

`#  ifdef __cplusplus`
`#    include <atomic>`
`     using namespace std;`
`#  else // def __cplusplus`
`#    include <stdatomic.h>`
`#  endif // def __cplusplus`

After this change, I succeeded to build my C++ project.

 - open62541 Version (release number or git tag): 1.4.10
 - Operating system: Linux Mint Cinnamon 22

idzm:
Adds support for C++ compilation when using C11 atomics. It achieves this by including the `atomic` header in C++ and defining aliases for the C-style atomic types and functions to their C++ counterparts.

Closes #7076.

### DIFF
--- a/include/open62541/config.h.in
+++ b/include/open62541/config.h.in
@@ -234,7 +234,13 @@
 #  endif
 # elif defined(__has_include) && __has_include(<stdatomic.h>)
 #  define UA_HAVE_C11_ATOMICS
-#  include <stdatomic.h>
+#  ifdef __cplusplus
+#    include <atomic>
+#    define _Atomic(T) std::atomic<T>
+#    define atomic_uintptr_t std::atomic_uintptr_t
+#  else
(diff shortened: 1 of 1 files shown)


jpfr:
Merged. Thanks!

PR #7222, opened 2025-04-14
Fix ASAN crash in removeFieldFromEncoding by validating memory bounds

Shivam7-1:
issue link: https://issues.oss-fuzz.com/issues/387317444
In this PR a high-severity vulnerability in `removeFieldFromEncoding `where an invalid memory range (negative size) could be passed to `memmove`(), triggering an ASAN crash (Negative-size-param). The fix adds bounds checks to ensure that start and end are within the valid range of the UA_ByteString and that end >= start. This prevents undefined behavior when removing fields during JSON decoding in ExtensionObjects.


jpfr:
Merged. Thanks!

PR #7223, opened 2025-04-14
Updated ua_client.c to fix error masking

aaghash17:


aaghash17:
This change addresses the issue logged: #7127 

The code has been modified to resolve the problem. 
The unit test failures have also been fixed, ensuring that the tests now pass with the new changes.

jpfr:
The unit tests now show a BadInternalError instead of ConnectionClosed.

How is that better?
That said, I am all in favor for returning better error codes.

aaghash17:
@jpfr ,

Updated the test case accordingly

aaghash17:
@jpfr
@andreasebner
@marwinglaser

Any chance someone could take a look at this when time allows? Thanks!

tests/client/check_client_securechannel.c
```diff
@@ -205,17 +205,23 @@ START_TEST(SecureChannel_cableunplugged) {
     client->connection.recv = UA_Client_recvTesting;
 
     /* Simulate network cable unplugged (no response from server) */
-    UA_Client_recvTesting_result = UA_STATUSCODE_BADINTERNALERROR;
```
jpfr, line 208: This tests a different condition now.
Because, when we pause the server, the socket will still accept packets.
Only they don't get processed.

This test aims to simulate an "uplugged cable".
Where the connection breaks with an error on the client side also.
Not just a timeout.

aaghash17: Understood,
And reverted the test condition as before

jpfr (approved):
Looks good 👍

jpfr:
The cha he looks good.
But the CI does not complete.
Can be merged when the CI is happy.

aaghash17:
> The change looks good. But the CI does not complete. Can be merged when the CI is happy.

The change is simple and i don't get why CI is unhappy, any inputs to check ?

aaghash17:
@jpfr 

The change is minimal, and I’m not sure why the CI is failing. Do you have any suggestions on what I should check?

aaghash17:
@jpfr ,
Awaiting for your reply

aaghash17:
@jpfr ,

Awaiting for the response

jpfr:
There was a CI fix pushed to 1.3 last week.
Please rebase and it should run.

aaghashr:
@jpfr Few workflows are awaiting approval

PR #7225, opened 2025-04-15
Run linux CI in an ubuntu container

marwinglaser:
This reverts the changes from #7133 and instead uses an ubuntu 20.04 container. This allows us to keep using this older Ubuntu version as long as we need to. A similar PR for the rewritten CI on the master branch that works with the version matrix is also in the works, where it can also be expanded to build on even older ubuntu versions or even different distributions.

This depends on a minimally modified custom docker image, which is currently built in a new repository (https://github.com/marwinglaser/images) and stored in GitHub's container registry. Before this PR is merged these images should be migrated to the open62541 organization. Interestingly enough we already have a similar repository that currently seems unused (https://github.com/open62541/open62541-ci), which would be a nice location.

### DIFF
--- a/tests/pubsub/check_pubsub_connection_mqtt.c
+++ b/tests/pubsub/check_pubsub_connection_mqtt.c
@@ -17,7 +17,7 @@
 #include <check.h>
 
 //#define TEST_MQTT_SERVER "opc.mqtt://test.mosquitto.org:1883"
-#define TEST_MQTT_SERVER "opc.mqtt://localhost:1883"
+#define TEST_MQTT_SERVER "opc.mqtt://mosquitto:1883"
 
 UA_Server *server = NULL;
 UA_ServerConfig *config = NULL;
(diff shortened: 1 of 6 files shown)


tests/pubsub/check_pubsub_connection_mqtt.c
```diff
@@ -17,7 +17,7 @@
 #include <check.h>
 
 //#define TEST_MQTT_SERVER "opc.mqtt://test.mosquitto.org:1883"
-#define TEST_MQTT_SERVER "opc.mqtt://localhost:1883"
+#define TEST_MQTT_SERVER "opc.mqtt://mosquitto:1883"
```
marwinglaser, line 20: This should probably be read from an environment variable, especially since change this breaks local tests with a broker running on localhost.

marwinglaser: Moved to env var in fb170a0042c41380d4e57117c780b2e865c70148

PR #7226, opened 2025-04-15
[REVIEW]Add fallback definitions for AF_INET and AF_INET6 only if not already defined

NoelGraf:


marwinglaser:
Closes https://github.com/open62541/open62541/pull/7221

PR #7231, opened 2025-04-22
Rename UA_String_printf to UA_String_format

jpfr:


jpfr:
This makes an update to the avahi discovery.
@V-Strassheim please validate this still works.

src/server/ua_discovery_mdns_avahi.c
```diff
@@ -881,32 +855,29 @@ UA_Discovery_addRecord(UA_DiscoveryManager *dm, const UA_String servername,
...
+    UA_String serviceDomain = {64, serviceDomainBuf};
+    createServiceDomain(&serviceDomain, servername, hostname);
 
-    UA_Boolean exists = UA_Discovery_recordExists(dm, serviceDomain,
-                                                  port, protocol);
+    UA_Boolean exists = UA_Discovery_recordExists(dm, serviceDomain, port, protocol);
     if(exists == true)
         return UA_STATUSCODE_GOOD;
 
     UA_LOG_INFO(dm->sc.server->config.logging, UA_LOGCATEGORY_DISCOVERY,
```
V-Strassheim, line 866: @jpfr  I get a segmentation fault in this logging function when I start /examples/discovery_server_multicast with avahi option.
Here is a screenshot from the whole call stack: 
image

I can investigate further and give more information tomorrow.

src/server/ua_discovery_mdns_avahi.c
```diff
@@ -881,32 +855,29 @@ UA_Discovery_addRecord(UA_DiscoveryManager *dm, const UA_String servername,
...
                 "Multicast DNS: add record for domain: %s", serviceDomain);
```
V-Strassheim, line 867: serviceDomain is now an UA String. This will fix the [segmentation fault](https://github.com/open62541/open62541/pull/7231/files#r2054378169) in the format_string function:
```suggestion
                "Multicast DNS: add record for domain: %S", serviceDomain);
```

I'm still testing a couple of things, but so far I haven't seen any other problems.

jpfr: 👍

jpfr:
Thanks, updated.

V-Strassheim:
Haven't found anything else, looks good so far :+1: 

Independently of these changes, I have found some issues with LeakSanitizer in ua_discovery_mdns_avahi.c. I will fix them in a separate PR as soon as this PR is merged.

RolfKal:
Instead of removing the check in the beginning of mp_printf() completely it may be actually better to change it to:

  `  // Check that the inputs are sane`
  `  if(!s && n)`
  `      return -1;`

This protects from the caller passing in a NULL pointer but indicating a length > 0 which will crash in the function.

PR #7236, opened 2025-04-22
Merge 1.3 to 1.4

marwinglaser:
Closes #7229

### DIFF
--- a/tests/check_eventloop_mqtt.c
+++ b/tests/check_eventloop_mqtt.c
@@ -6,15 +6,34 @@
 #include <open62541/plugin/log_stdout.h>
 #include "open62541/types.h"
 #include "open62541/types_generated.h"
-
-/* The test suite requires an MQTT server listening at localhost:1883 */
+#include "open62541/util.h"
 
 #include "testing_clock.h"
+#include <stdio.h>
 #include <time.h>
 #include <stdlib.h>
 #include <check.h>
 
 unsigned int messageCount = 0;
+
+UA_String broker_hostname = {0, NULL};
+UA_UInt16 broker_port = 0;
+
+static char* get_mqtt_broker_address(void) {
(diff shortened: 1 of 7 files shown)


tests/check_eventloop_mqtt.c
```diff
@@ -10,12 +10,45 @@
...
+    if (!broker_hostname.data) goto failed;
+
+    return;
+
+    failed:
+        printf("Failed to parse OPEN62541_TEST_MQTT_BROKER with value \"%s\"", broker);
+        exit(1);
+}
+
+static void teardown(void) {}
```
marwinglaser, line 50: This is really hacky, I'm open for any suggestions. We could use separate variables, but everywhere else uses one.

marwinglaser:
The ethernet tests are flaky and fail with `UA_STATUSCODE_BADINTERNALERROR` sometimes, but I'm not sure why. I don't think I touched anything related. They fail outside of containers as well (see [Coverage Test](https://github.com/open62541/open62541/actions/runs/14594158935/job/40935970043?pr=7236) of this PR)
```
	 21 - check_eventloop_eth (Failed)
	 68 - check_pubsub_connection_ethernet (Failed)
```

tests/check_eventloop_mqtt.c
marwinglaser: Now using `UA_parseEndpointUrl`

marwinglaser:
@jpfr I consider this complete now. Waiting for CI to run, but I don't expect anything new.

marwinglaser:
It seems that the tests fail when using image version `20250420.1.0`, while working on `20250413.1.0`.

jpfr:
Merged. As the 1.4 branch was even "more red" before.
Let's handle the remaining issues in a follow-up PR.

PR #7238, opened 2025-04-22
fix(client): use after free in client asyncServiceCalls

bluhm:
Commit de9d691906547c9fc99e0a77198a80fe5bba54b1 fixed a use after free in UA_Client_AsyncService_cancel().  Move all elements to a local list before iterating over the callbacks.  Commit 713cdb419f0f844517afc549c7c1ef48d8498cfb fixed the same bug in in asyncServiceTimeoutCheck().  After timeout, remove the async service calls from the global list and store them in a local one.  Call UA_Client_AsyncService_cancel() and UA_free() while traversing this local list.  Otherwise the callback could free a global list element that asyncServiceTimeoutCheck() is referencing.

Backport these fixes from 1.4 to 1.3.  I have been carrying them in my local patch set for too long.


jpfr:
Nice, merged.

PR #7252, opened 2025-05-07
Generate callback stubs

basyskom-dege:
This PR picks up on the idea of #3199 (and the closed PR #3219). It allows to generate callback stubs for methods in the nodeset compiler. The body of the callback will instantiate the input and output arguments. Additionally, comments with infos about the members of the datatypes/enums are printed as well. 

An additional argument `--method-callbacks` enables this feature. When using CMake, the feature can be enabled with `METHOD_CALLBACKS` in the `ua_generate_nodeset_and_datatypes` command:

```
    ua_generate_nodeset_and_datatypes(
        NAME "autoid"
        FILE_NS "${FILE_NS_DIRPREFIX}/AutoID/Opc.Ua.AutoID.NodeSet2.xml"
        DEPENDS "di"
        METHOD_CALLBACKS
    )
```

The result lookes like this (for the Scan method i=7013 from the AutoId Nodeset):

```
UA_StatusCode 
namespace_autoid_generated_methodCallback_Types_ObjectTypes_BaseObjectType_TopologyElementType_ComponentType_DeviceType_AutoIdDeviceType_RfidReaderDeviceType_Scan
(UA_Server *server, const UA_NodeId *sessionId, void *sessionHandle, const UA_NodeId *methodId, void *methodContext, const UA_NodeId *objectId, void *objectContext, size_t inputSize, const UA_Variant *input, size_t outputSize, UA_Variant *output) {
/* InputArguments arrayDimensions=['1'] */
UA_ScanSettings setting = *(UA_ScanSettings*)input[0].data;
/* INFO Members of UA_ScanSettings:
* duration : Duration 
* cycles : Int32 
* dataAvailable : Boolean 
* locationType : LocationTypeEnumeration  (optional)
*/

/* OutputArguments arrayDimensions=['2'] */
UA_RfidScanResult results;
UA_init(&results, &UA_TYPES_AUTOID[UA_TYPES_AUTOID_RFIDSCANRESULT]);
/* INFO Members of UA_RfidScanResult:
* sighting : RfidSighting 
*/

UA_AutoIdOperationStatusEnumeration status;
UA_init(&status, &UA_TYPES_AUTOID[UA_TYPES_AUTOID_AUTOIDOPERATIONSTATUSENUMERATION]);
/* INFO Members of UA_AutoIdOperationStatusEnumeration:
* SUCCESS = 0 (Successful operation)
* MISC_ERROR_TOTAL = 1 (The operation has not be executed in total.)
* MISC_ERROR_PARTIAL = 2 (The operation has been executed only partial.)
* PERMISSON_ERROR = 3 (Password required)
* PASSWORD_ERROR = 4 (Password is wrong.)
* REGION_NOT_FOUND_ERROR = 5 (Memory region not available for the actual tag.)
* OP_NOT_POSSIBLE_ERROR = 6 (Operation not supported by the actual tag.)
* OUT_OF_RANGE_ERROR = 7 (Addressed memory not available for the actual tag.)
* NO_IDENTIFIER = 8 (The operation cannot be executed because no tag or code was inside the range of the AutoID Device or the tag or code has been moved out of the range during execution.)
* MULTIPLE_IDENTIFIERS = 9 (Multiple tags or codes have been selected, but the command can only be used with a single tag or code.)
* READ_ERROR = 10 (The tag or code exists and has a valid format, but there was a problem reading the data (e.g. still CRC error after maximum number of retries).)
* DECODING_ERROR = 11 (The (optical) code or plain text has too many failures and cannot be detected.)
* MATCH_ERROR = 12 (The code doesn’t match the given target value.)
* CODE_NOT_SUPPORTED = 13 (The code format is not supported by the AutoID Device.)
* WRITE_ERROR = 14 (The tag exists, but there was a problem writing the data.)
* NOT_SUPPORTED_BY_DEVICE = 15 (The command or a parameter combination is not supported by the AutoID Device.)
* NOT_SUPPORTED_BY_TAG = 16 (The command or a parameter combination is not supported by the tag.)
* DEVICE_NOT_READY = 17 (The AutoID Device is in a state not ready to execute the command.)
* INVALID_CONFIGURATION = 18 (The AutoID Device configuration is not valid.)
* RF_COMMUNICATION_ERROR = 19 (This error indicates that there is a general error in the communication between the transponder and the reader.)
* DEVICE_FAULT = 20 (The AutoID Device has a hardware fault.)
* TAG_HAS_LOW_BATTERY = 21 (The battery of the (active) tag is low.)
*/

return UA_STATUSCODE_BADNOTIMPLEMENTED;
}
```


basyskom-dege:
Right now, the function body ends up in the generated file, which gets overwritten every time you rebuild. That makes it hard to find the function, especially in a large generated file, and it's not really practical for writing your own logic there.

One idea would be to generate a stub that just calls an external function (with the resolved input and output arguments). The user would then implement that function in their own code and we could provide a header with the declaration. 

Another option might be to move user callbacks into a separate .c file?

What do you think is the best way to handle this?

jpfr:
I would generate the methods only in the header definitions and add the callbacks to the information model.

Generating stub implementations should be a second build flag. Users might want the implementation to live in existing project files.

we will have another look when the build is green again.

jimmy-loyola:
I'd really need this feature too. Let me know if I can contribute. Thanks.

PR #7253, opened 2025-05-07
mbedtls createcertificate: fix length calculation

jennes:
In createcertificate, strlen() was used on a not null terminated string. This could result in using uninitialized memory for SAN attributes or even segfaults.


jpfr:
Thanks 👍

jennes:
Maybe this should also be included into the `master` branch. So far we only use the `1.4` branch, that's why my pull request was for `1.4` only. If it helps, I can create a PR for `master`, too. It would be unfortunate to have this as a regression once the first stable releases come out of `master`.

jpfr:
We regularly merge all commits from 1.4 to master.

PR #7261, opened 2025-05-16
Shell-Interface for ua-cli tool

jpfr:


RolfKal:
It would be probably useful to update:

`line 67:  UA_String_vprintf(&out, msg, args);`

as that was changed to UA_String_vformat() recently.

Another instance where the old API name is used is in ua_types.c line 203.

Issue #7265, opened 2025-05-19
Memory leak at ua_pki_openssl.c malloc of scandir is not freed


HUmgnad:
## Description

The namelist argument of scandir needs to be freed after use see (also the example at the bottom):
https://man7.org/linux/man-pages/man3/scandir.3.html

This is not done in the code at open62541/plugins/crypto/openssl/ua_pki_openssl.c
There are three locations where scandir() is called.

## Background Information / Reproduction Steps

I used the UA_CertificateVerification_Verify Method during the ActivateSession call.

My valgrind output:
```
==28774== 720 (480 direct, 240 indirect) bytes in 6 blocks are definitely lost in loss record 30 of 33
==28774==    at 0x4A2D7A9: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==28774==    by 0x5D15CCA: __scandir64_tail (scandir-tail-common.c:62)
==28774==    by 0x51F20C8: UA_ReloadCertFromFolder (ua_pki_openssl.c:305)
==28774==    by 0x51F23B9: UnknownInlinedFun (ua_pki_openssl.c:649)
==28774==    by 0x51F23B9: UA_CertificateVerification_Verify (ua_pki_openssl.c:639)
```

## Checklist

Please provide the following information:

 - [X] open62541 Version (release number or git tag): v1.4.10
 - [ ] Other OPC UA SDKs used (client or server): 
 - [X] Operating system: Linux
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [X] Critical issue

jpfr:
Closed via [#7266](https://github.com/open62541/open62541/pull/7266).

PR #7266, opened 2025-05-19
Bugfix/add free for scandir

HUmgnad:
Fix for https://github.com/open62541/open62541/issues/7265
Is it fine to use free() directly or does this break any code of the library itself?


jpfr:
@NoelGraf Please verify

HUmgnad:
I have not really noticed the continue statement, that is correct.
I updated the commit accordingly.

PR #7267, opened 2025-05-19
Add socket limit configuration flag

NoelGraf:
Currently, there is no way to limit the number of open sockets. This PR introduces a new server configuration flag that limits the maximum number of open sockets.

Once a connection has been accepted, the server verifies whether the maximum limit has been reached. If so, the listening sockets are deregistered from the socket list.

### DIFF
--- a/arch/posix/eventloop_posix_tcp.c
+++ b/arch/posix/eventloop_posix_tcp.c
@@ -45,6 +45,12 @@ typedef struct {
 static void
 TCP_shutdown(UA_ConnectionManager *cm, TCP_FD *conn);
 
+static UA_StatusCode
+TCP_registerListenSockets(UA_POSIXConnectionManager *pcm, const char *hostname,
+                          UA_UInt16 port, void *application, void *context,
+                          UA_ConnectionManager_connectionCallback connectionCallback,
+                          UA_Boolean validate, UA_Boolean reuseaddr);
+
 /* Do not merge packets on the socket (disable Nagle's algorithm) */
 static UA_StatusCode
 TCP_setNoNagle(UA_FD sockfd) {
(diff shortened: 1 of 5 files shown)


include/open62541/plugin/eventloop.h
```diff
@@ -454,12 +454,17 @@ struct UA_InterruptManager {
...
- * POSIX EventLop Implementation
+ * POSIX EventLoop Implementation
  * -----------------------------
  * The POSIX compatibility of Win32 is 'close enough'. So a joint implementation
  * is provided. The configuration paramaters must be set before starting the
  * EventLoop.
  *
+ * **Socket configuration**
+ *
+ * 0:max-sockets [uint32]
```
jpfr, line 465: This is now even "more global".
You added the limit
- not to a specific server socket
- not to the TCP subsystem
but to the entire EventLoop.

In that case we need to also ensure that we do not exceed the number of sockets across TCP and UDP!

arch/posix/eventloop_posix_tcp.c
```diff
@@ -231,6 +252,19 @@ TCP_connectionSocketCallback(UA_ConnectionManager *cm, TCP_FD *conn,
                         &UA_KEYVALUEMAP_NULL, response);
 }
 
+static void *
+removeListenSockets(void *application, UA_RegisteredFD *rfd) {
```
jpfr, line 256: You also need to make sure that new server sockets are not getting registered...

arch/posix/eventloop_posix_tcp.c
```diff
@@ -68,6 +68,21 @@ TCP_checkStopped(UA_POSIXConnectionManager *pcm) {
     }
 }
 
+static UA_Boolean maxSocketsLimitReached = false;
```
jpfr: This doesn't work as a global variable.
We might have several EventLoop instances in the same process.

arch/posix/eventloop_posix_tcp.c
```diff
@@ -329,6 +363,17 @@ TCP_listenSocketCallback(UA_ConnectionManager *cm, TCP_FD *conn, short event) {
...
     pcm->fdsSize++;
 
+    /* Verify whether the maximum socket limit has been exceeded.
+     * If true, remove listen sockets from the socket list to stop
+     * accepting additional connection requests */
+    const UA_UInt32 *maxSockets = (const UA_UInt32 *)UA_KeyValueMap_getScalar(
+        &cm->eventSource.params, UA_QUALIFIEDNAME(0, "max-sockets"),
+        &UA_TYPES[UA_TYPES_UINT32]);
+    if(maxSockets && *maxSockets != 0 && pcm->fdsSize >= (size_t)*maxSockets) {
+        ZIP_ITER(UA_FDTree, &pcm->fds, removeListenSockets, el);
```
jpfr: This can still be DDOS'ed.
I would remove the listen socket entirely when we have maxed out the open connections.
So new connections get rejected in the kernel already.

include/open62541/plugin/eventloop.h
```diff
@@ -494,6 +494,9 @@ UA_EventLoop_new_POSIX(const UA_Logger *logger);
  *
  * **Configuration parameters for the ConnectionManager (set before start)**
  *
+ * 0:max-sockets [uint32]
```
jpfr: Rename this to max-connections.
Sockets are a UNIX API for managing connections.
But lwip-raw doesn't have sockets, for example.

include/open62541/plugin/eventloop.h
```diff
@@ -494,6 +494,9 @@ UA_EventLoop_new_POSIX(const UA_Logger *logger);
...
+ *    max sockets (default: 0 -> unbounded).
```
jpfr: You should document what happens when max-connections is reached.
That is, the server sockets get deactivated.

jpfr:
Merge. Nice!

PR #7274, opened 2025-05-20
fix(core): Fixing some typos in diagnostics

m-obi:


jpfr:
This doesn't work due to auto-generated definitions that come from the standard XML.
We will update this when the upstream makes the fix.
Closing this.

PR #7282, opened 2025-05-25
refactor(tools): fix possible NULL-pointer dereference in ua-cli

laroche:
Fix possible NULL-pointer dereference in ua-cli as reported by clang scan-build.


jpfr:
Thanks 👍

PR #7284, opened 2025-05-25
feat(client): add combined subscription callbacks

RolfKal:
Add an option to register data change and event callbacks on subscription level


jpfr:
Thanks for this.

Please provide your motivation, what is lacking in the current functionality?

The CI has some red tests.
They need to go green before we can merge.

RolfKal:
> Thanks for this.
> 
> Please provide your motivation, what is lacking in the current functionality?

Basically when a client application passes this event on to some other component, with each sub event reported individually, this results in potentially a lot of small event messages being generated and having to be marshalled to a potentially different context. With the possibility to receive the whole aggregated bunch of events at once this overhead can significantly be reduced.

This is incidentally also the functionality provided by the original OPC UA sample implementation as it passes on the aggregated network event list directly on subscription niveau instead of reporting them as individual monitored item events.

Without the use of the createEx() APIs and passing a valid dataChange and/or event callback pointer, everything works exactly as it did before. I chose to not pass the events on individually per monitored item if there is a callback on subscription level. It would be possible to do that anyways if there was a valid monitored item callback registered when creating such an item, but I'm feeling undecided about that detail.
 
> The CI has some red tests. They need to go green before we can merge.

I noticed and found the problem and am running final tests to make sure it works fine.

RolfKal:
@jpfr Is there anything else I can do here?
While the Linux Ubuntu checks now seem to consistently succeed, there seems to be since several days a problem with the Windows runs to get aborted consistently because of the Windows 2019 server images being about to be discontinued.

RolfKal:
@jpfr has this any change to be committed to the master tree? My use case pretty much depends on this being available to provide a reasonable performance in subscription callback handling.

jpfr:
@RolfKal,

We have thought about your use case. And how we can get a generalized solution that works across different services.

Please check if the following works for your use case:

There is now a "notification callback" in the client for all kinds of internal events.
Here is the current list of "notification types".

https://github.com/open62541/open62541/blob/c36c59ecab17a4de0e35f986c629327caef1fa61/include/open62541/common.h#L206

And here is the callback in the client config:

https://github.com/open62541/open62541/blob/c36c59ecab17a4de0e35f986c629327caef1fa61/include/open62541/client.h#L684

One of the notification types is to signal the beginning and the end of processing a response message.

That way, you can "collect" the incoming individual MonitoredItem-notifications.
And you know to get a signal when you can "batch-process" the collected MonitoredItem-notifications.

Does that work for you?

jpfr:
No response from @RolfKal .
This was solved with the described strategy.

PR #7293, opened 2025-05-28
Fix build with diffrent encoding settings

NoelGraf:
- The build error has been fixed when PubSub is switched off due to missing directives
- No value attribute extraction from the XML if XML encoding is deactivated.
- Correctly added the target dependency. The check for dependencies was incorrect, which prevented the dependencies from being added to the target.


jpfr:
👍

Issue #7279, opened 2025-05-21
nodeset_compiler: LocalizedText does not parse the Locale from the XML correctly


julepp:
## Description
According to the OPC UA XML Schema Localized Texts have the following format:

    <DisplayName Locale="en-US">Hello</DisplayName>

or without a locale.
The Class LocalizedText expects the Locale as an additional element and not as an attribute as the Code and the comment in the code shows.

    def parseXML(self, xmlvalue):
        # Expect <LocalizedText> or <AliasName>
        #          <Locale>xx_XX</Locale>
        #          <Text>TextText</Text>
        #        <LocalizedText> or </AliasName>
        if not isinstance(xmlvalue, dom.Element):
            self.text = xmlvalue
            return
        self.checkXML(xmlvalue)
        tmp = xmlvalue.getElementsByTagName("Locale")
        if len(tmp) > 0 and tmp[0].firstChild != None:
            self.locale = tmp[0].firstChild.data.strip(' \t\n\r')
        tmp = xmlvalue.getElementsByTagName("Text")
        if len(tmp) > 0 and tmp[0].firstChild != None:
            self.text = tmp[0].firstChild.data.strip(' \t\n\r')

This would also require a change when passing the XML Element in the creation of the LocalizedTexts for the attributes of a Node.

## Checklist

Please provide the following information:

 - [ ] open62541 Version (release number or git tag):
 - [ ] Other OPC UA SDKs used (client or server): 
 - [ ] Operating system:
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

PR #7294, opened 2025-05-28
fix(node_comp): Fixed parsing of LocalizedText (#7279)

julepp:
The Format of the LocalizedText XML-Element that the parser expected was not to specification. It has been changed to expect the following:
```xml
<LocalizedText locale="en">Text</LocalizedText>
```
Whenever a LocalizedText is created, the whole XML Element has to be passed now to be able to parse the locale. Before, only the child element was passed. If no Locale is specified, it remains None.

The class LocalizedText in datatypes.py was adjusted.
The class Node and ReferenceTypeNode in nodes.py
There is also a bunch of formatting that was done.

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -7,61 +7,103 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - build_name: "Debug Build & Unit Tests (gcc)"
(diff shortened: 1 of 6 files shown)


jpfr:
Please split functional and cosmetic changes into separate commits.
This is harder than necessary to review otherwise.

PR #7295, opened 2025-05-29
refactor(crypto): Clean up Certificate Verification

RolfKal:
- Document that the crypto plugin initialization routines should clear an incoming certificate verification and remove the unnecessary clear() calls in the code
- avoid calling UA_CertificateVerification_Trustlist() with an empty trustlist, Doing so results in a certificate verification configuration that can not accept any possible server
- refactor certificate verification code on the client side to move duplicate code into its own local function and call it where needed


jpfr:
Seems to be a problem of the Github Action Runners.
I restarted the tests.

jpfr:
Looks good!

Clang-Analyzer still complains, can be merged then.

grafik

RolfKal:
Windows builds all seem to get canceled because of a scheduled Windows Server 2019 brownout!

jpfr:
Thanks for this.
We update the Windows CI in a separate PR.

PR #7298, opened 2025-06-02
[WIP] Implement query service

jpfr:


vvpnet:
This is a very necessary feature. Tell me when can I expect it to be experimental in the release version?

PR #7299, opened 2025-06-03
feat(ci): add windows build and test workflow based on github runner

andreasebner:
This PR is a preparation for removing the Windows Azure build pipeline in order to create a more homogeneous CI environment.

In addition, we have to update the Windows version because Server 2019 is no longer supported. This could certainly also be done in the current Azure environment.


andreasebner:
@jpfr This is now up and running. The failed CI tasks just require a minor adjustment to the path. You can take a look to see if everything looks OK.

PR #7300, opened 2025-06-04
refactor(examples,tools,server,plugins,pubsub): clang compiler warnings

laroche:
- unused macros
- sign conversions
- adding "static" modifier
- printf formatting

### DIFF
--- a/deps/dtoa.c
+++ b/deps/dtoa.c
@@ -35,7 +35,7 @@
 #define fracmask  0x000FFFFFFFFFFFFFU
 #define expmask   0x7FF0000000000000U
 #define hiddenbit 0x0010000000000000U
-#define signmask  0x8000000000000000U
+//#define signmask  0x8000000000000000U
 #define expbias   (1023 + 52)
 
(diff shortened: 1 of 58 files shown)


jpfr:
All of this code gets tested in the CI.
Future regressions should be caught automatically.

Which additional build settings did you used with clang to get these warnings exposed?

laroche:
I've just used the big switch "-Weverything" and then again disabled a few less interesting and noisy options.
As this list gets quite long, maybe changing this to a positive list is possible, but for most options there
are also many items that are not worth fixing.
E.g. unused macros occur often depending on features selected for compiling. I've just fixed a few ones
that are leftovers from deleted code.

best regards,

Florian La Roche

laroche:
Just to add to previous comment: Warnings that can be enabled beyond -Wall often need review and
only some reports actually need fixing. The warning output from "-Weverything" then contains the exact
warning option that could be enabled separately ontop of "-Wall".

best regards,

Florian La Roche

jpfr:
Nice, merged.

Issue #6237, opened 2024-01-18
Published json via MQTT is mixed up 


MarcelArmgart:
## Description
Hi Team,

I am currently workling with the sources of v1.4 (#5e76d257fd0ab4bc2558bf5e24ba4d5f0bab0aec)
I enabled MQTT publish via json-encoding on a server. Therefore I used the example "tutorial_pubsub_mqtt_publish.c".
In the example I changed the line 50 from
`#define TRANSPORT_PROFILE_URI        "http://opcfoundation.org/UA-Profile/Transport/pubsub-mqtt"`
to 
`#define TRANSPORT_PROFILE_URI        "http://opcfoundation.org/UA-Profile/Transport/pubsub-mqtt-json"`
as suggested in issue [#6129](https://github.com/open62541/open62541/pull/6129).

After that the example works perfectly and publishes the current server time in the expected interval.
But the json-string which is send to the mqtt broker is mixed up.

I subscribed the broker with MQTT Explorer and mosquitto_sub.exe via PowerShell this is what was send by the server:
`{null"MessageId":"ua-data","MessageType":,"PublisherId":2234,"Messages":[{62541"DataSetWriterId":2,"SequenceNumber":{"MajorVersion":464029280,"MinorVersion":464029280},"MetaDataVersion":"2024-01-18T12:46:00.289Z","Timestamp":0,"Status":,"Payload":{"Server localtime":{"Type":13,"Body":"2024-01-18T12:46:00.289Z"}}}]}`

Correct would be:
`{
    "MessageId":null,
    "MessageType":"ua-data",
    "PublisherId":2234,
    "Messages": [{
        "DataSetWriterId":62541,
        "SequenceNumber":2
        "MetaDataVersion":{
            "MajorVersion":464029280,
            "MinorVersion":464029280
        },
        "Timestamp":"2024-01-18T12:46:00.289Z",
        "Status":0,
        "Payload":{
            "Server localtime":{
                "Type":13,
                "Body":
                "2024-01-18T12:46:00.289Z"
            }
        }
    }]
}`

What am I doing wrong?

## Background Information / Reproduction Steps
Sources are buid from current v1.4 (Git-Hash: #5e76d257fd0ab4bc2558bf5e24ba4d5f0bab0aec)

Used CMake options:

```bash
[...]

PR #7305, opened 2025-06-05
Fix json encoding problem when compiling as Release

migueltorres:
Following up the issue #6237 , I applied a fix on both src/ua_types_encoding_json.c and src/ua_types_encoding_json_105.c. Tested it with mqtt pub sub and json encoding, as Debug and as Release.


jpfr:
Thanks! Merged.

PR #7306, opened 2025-06-05
Support zoned DateTime representations with UTC offset

NoelGraf:


jpfr:
The code is flawless.
Add a unit test example and we are good to go.

jpfr:
I think we can also expose this as
UA_DateTime_parse
In the public api.
With a signature similar to the other parsing methods.

jpfr:
Nice! Merged.

PR #7309, opened 2025-06-06
feat(pubsub) extend pubsub configuration with enable flags

andreasebner:


include/open62541/server_pubsub.h
```diff
@@ -257,6 +257,7 @@ typedef struct {
     UA_Variant address;
     UA_KeyValueMap connectionProperties;
     UA_Variant connectionTransportSettings;
+    UA_Boolean enabled;
```
jpfr, line 260: Add some documentation what enabled means and when it takes effect.

Will this be written into when the component is manually enabled?

jpfr (commented):
Please add documentation to the public header. Then there can be a review against the defined behavior.

include/open62541/server_pubsub.h
andreasebner: See the description and added documentation in #7379 which replaces this PR.

PR #7315, opened 2025-06-12
feat: Add pause to historical data gathering

behrangyou:
Stopping a node from in memory historizing removes both recording and the ability to read historical data. There is no way to stop recording new values while keeping the existing historical data available for clients.

I propose adding a paused flag to the backend's context. When paused is true, new values are not recorded, but historical reads still work. This is implemented by checking the flag in setValue_gathering_default and providing a function to set the flag.


jpfr:
Looks good.
Rebased on the current master to retrigger the CI run.

behrangyou:
> Looks good. Rebased on the current master to retrigger the CI run.

Done!

PR #7322, opened 2025-06-18
fix(client): Fix user token policy validation

basyskom-jvoe:
If there is a separate authSecurityPolicies list, it should not be required to have the token's policy in the securityPolicies list in the client config if the endpoint uses a different policy.


basyskom-jvoe:
The connect failed with `Rejecting endpoint 6: No matching UserTokenPolicy` for the following scenario:
- The server exposes an endpoint with security policy `None` and an user/password auth token with `Basic256Sha256`
- The client config only has the `None` policy in `securityPolicies` and `Basic256Sha256` in `authSecurityPolicies`.

basyskom-jvoe:
The updated commit should be ok, at least I can no longer reproduce my problem locally.

Maybe it would be a good idea to also skip the check in `findUserTokenPolicy()` if the token is anonymous?
Most servers I have encountered use anonymous tokens with an empty securityPolicyUri, so using `activateSessionAsync()` with anonymous auth on an endpoint  with a policy != None fails if the policy is not also available  in `authSecurityPolicies`.

jpfr:
Thanks for this.
In order to make this robust against future changes, please add a unit test for the scenario.

basyskom-jvoe:
@jpfr I have added two tests.
Could we get a v1.4.13 release with this change?

jpfr:
Merged. Thanks for this.

PR #7327, opened 2025-06-24
Set the parent scope for some open62541 variables

NoelGraf:
When the stack is added to a CMake project as a dependency using add_subdirectory, some CMake variables are not properly propagated. This PR updates the relevant variables to be set in the parent scope, ensuring they are accessible from the main project's CMake configuration.


jpfr:
👍

PR #7328, opened 2025-06-24
Async operations update

jpfr:


basyskom-jvoe:
I have implemented a small test server using the new API as shown in `check_server_asyncop.c` and reading/writing using explicit requests from the client works great.

@andreasebner The problem outlined in my comment in PR #6220 is still the same if it comes to monitored items. If the callback returns `UA_STATUSCODE_GOODCOMPLETESASYNCHRONOUSLY` and the data acquisition from the source is done somewhere else asynchronously, a monitored item on such a node never gets a value during sampling.
Dropping the node on the Data Access View in UaExpert results in a broken table entry with the status code `BadWaitingForResponse`. 

Do you have an idea for an architecture where sampling would also be possible for nodes with asynchronous callbacks? As the sampling is done using the session of the owner of the subscription, there is no way to recognize a sampling related request in the callback and to answer it synchronously (in case the data source's response times would allow it).

andreasebner:
> I have implemented a small test server using the new API as shown in `check_server_asyncop.c` and reading/writing using explicit requests from the client works great.
> 
> @andreasebner The problem outlined in my comment in PR #6220 is still the same if it comes to monitored items. If the callback returns `UA_STATUSCODE_GOODCOMPLETESASYNCHRONOUSLY` and the data acquisition from the source is done somewhere else asynchronously, a monitored item on such a node never gets a value during sampling. Dropping the node on the Data Access View in UaExpert results in a broken table entry with the status code `BadWaitingForResponse`.
> 
> Do you have an idea for an architecture where sampling would also be possible for nodes with asynchronous callbacks? As the sampling is done using the session of the owner of the subscription, there is no way to recognize a sampling related request in the callback and to answer it synchronously (in case the data source's response times would allow it).

@jpfr, since you've been working on the latest asynchronous behaviour, I'd like to add you to this thread. The aforementioned behaviour seems crucial and should perhaps be addressed in the upcoming release.

PR #7339, opened 2025-06-30
[Review] [1.4] refactor(deps): bump libressl to 3.9.2

GoetzGoerisch:


GoetzGoerisch:
@andreasebner the GitHub CI is disabled on purpose?

jpfr:
We are working on that.
I think Github doesn't like the volume of compute our CI generates...

GoetzGoerisch:
clang on windows took longer then usable and was canceled. Don't see any reason that this change is causing this.
Already the case on `1.4` since merge of #7295.

jpfr:
👍

Issue #7146, opened 2025-03-03
Accessing DataSetMessage and/or NetworkMessage timestamp in active pubsub communication


lmlcvc:
## Description

I’ve been working on a OPC-UA UDP PubSub subscriber using open62541, which is communicating with a PLC. The PLC continuously constructs and sends NetworkMessages (on byte level). Because of certain limitations in our design, I am able to put a timestamp into the NetworkMessage and DataSetMessage - but not that of the field.
Also, using the proper mask, the fields are on the Publisher side defined to be UA_Variant, not UA_DataValue - containing just the value, not status and timestamp.

Now obviously, since `data->sourceTimestamp`  is accessing something that's not there, it will be 0. **But I would like to see if I can get the timestamp of the Network or DataSet Message here (or somewhere near).**
I tried logging the `serverTimestamp` too just in case, but that's not it.

## Background Information / Reproduction Steps

As Subscriber, I construct an address space of subscribed variables and add an onWrite callback to each DataSetMessageField.
```cpp
// ...
  UA_ValueCallback callback;
  callback.onWrite = [](UA_Server *server, const UA_NodeId *sessionId, void *sessionContext,
                      const UA_NodeId *nodeId, void *nodeContext, const UA_NumericRange *range,
                      const UA_DataValue *data) {
      auto impl = static_cast<SubscriberImplementation::Impl*>(sessionContext);
      impl->dataSetCallback(server, sessionId, sessionContext, nodeId, nodeContext, range, data);
  };
  
  retval = UA_Server_setVariableNode_valueCallback(m_server, newNode, callback);
```

This is the logging callback previous code is pointing to:
```cpp
void SubscriberImplementation::Impl::dataSetCallback(UA_Server *server,
                const UA_NodeId *sessionId, void *sessionContext,
                const UA_NodeId *nodeId, void *nodeContext,
                const UA_NumericRange *range, const UA_DataValue *data) {
    // ...

    std::string logMessage = fmt::format(
        "NODE={} VALUE={} SOURCE_TIME={}",
        context->fullPath, 
        receivedValue, 
        data->sourceTimestamp);
}
```

The logs from the callback look something like this (note the `SOURCE_TIME=0` - of course, because that is the `sourceTimestamp` of the field, not DataSet or NetworkMessage):

```
2025-03-03 16:00:47.090930 [Subscriber.cpp:171, INF] NODE=Objects/Subscribed Variables/DataSet 100/Field_73 
[...]

PR #7344, opened 2025-07-03
Feature/pass datasetmassagetimestamp to variants

Schify:
As discussed in issue #7146 timestamps that are on the message level of the PubSub protocol are not made available on the subscriber level. This means that data fields containing variants, or raw data payloads are stripped of their temporal information, and are unable to access related higher level timestamps (namely the NetworkMessag's timestamp, and the DataSetMessage's timestamp). I believe this is not completely in line with the specification laid out in the OPC UA PubSub standard. 

In the OPC Foundation's document at [5.3.2 DataSetMessage field](https://reference.opcfoundation.org/Core/Part14/v105/docs/5.3.2) the description of Raw Data and Variant to me implies that the timestamp should be at least somehow provided when it is communally shared within a DataSetMessage. However, I am aware that at the same time the definition of the [sourceTimestamp](https://reference.opcfoundation.org/search/407?t=sourceTimestamp) runs contrary to this:
> The [sourceTimestamp](https://reference.opcfoundation.org/search/407?t=sourceTimestamp) is only returned with a [Value](https://reference.opcfoundation.org/search/407?t=Value) [Attribute](https://reference.opcfoundation.org/search/407?t=Attribute). For all other [Attributes](https://reference.opcfoundation.org/search/407?t=Attributes) the returned [sourceTimestamp](https://reference.opcfoundation.org/search/407?t=sourceTimestamp) is set to null.

My solution, for now, would be to just assign the DatasetMessage timestamp to variant fields as a sourceTimestamp (or their DataValue representation on the subscriber side). But perhaps a more invasive, but also more correct, adjustment would be to make the DataSetMessage available for individual DataValue variables.


jpfr:
What about the raw encoding?
What about DataValue with the timestamp not included?

Schify:
Hi @jpfr ,
Thanks for the reply. To answer your questions:

> What about DataValue with the timestamp not included?

I would not change the behaviour of the DataValue-containing messages at all, since there an explicit choice had been made by the publisher not to include a sourceTimestamp.

> What about the raw encoding?

The same form of quick solution could be inserted for the `UA_FIELDENCODING_RAWDATA` case, because this behaviour is explicitly called for in [5.3.2 DataSetMessage field](https://reference.opcfoundation.org/Core/Part14/v105/docs/5.3.2). However, I wanted to keep the scope of this pull request as limited as possible.

Schify:
Hi @jpfr ,

Is this feature is ever realistically getting merged into master? Or would there be some other workaround for this problem of utilizing DatasetMessage timestamps?

PR #7356, opened 2025-07-13
fix(server): use UA_malloc instead of malloc

max65482:
Replace malloc with UA_malloc in ua_server_ns0_gds.c


jpfr:
Thanks. I adjusted the commit title per our CONTRIBUTING.md

PR #7355, opened 2025-07-12
fix(cmake): respect user-defined mbedTLS paths in FindMbedTLS.cmake

JianbinLiu-CFLab:
Dear Maintainers,

### 1. Fix Bug in `FindMbedTLS.cmake` That Ignores User-Specified Paths

When using `open62541` with `mbedTLS` encryption support enabled, if users pass path parameters through CMake command line arguments like:

```bash
-DMBEDTLS_INCLUDE_DIRS=...
-DMBEDTLS_LIBRARY=...
-DMBEDX509_LIBRARY=...
-DMBEDCRYPTO_LIBRARY=...
```

These should be prioritized and used. However, the original `open62541/tools/cmake/FindMbedTLS.cmake` has a **design flaw that overrides user paths**, causing it to ignore user-specified custom paths and always force the use of system default locations. This is particularly problematic when `UA_BUILD_OSS_FUZZ=ON`.

The failure result is as follows:

```bash
CMake Warning: 
Manually-specified variables were not used by the project: MBEDCRYPTO_LIBRARY 
MBEDTLS_INCLUDE_DIRS 
MBEDTLS_LIBRARY 
MBEDX509_LIBRARY
```

#### 1.1 📌 Original Code (Flawed)

```cmake
# Check environment variable
if("$ENV{MBEDTLS_FOLDER_INCLUDE}")
    set(MBEDTLS_FOLDER_INCLUDE "$ENV{MBEDTLS_FOLDER_INCLUDE}")
endif()
if("$ENV{MBEDTLS_FOLDER_LIBRARY}")
    set(MBEDTLS_FOLDER_LIBRARY "$ENV{MBEDTLS_FOLDER_LIBRARY}")
endif()

find_path(MBEDTLS_INCLUDE_DIRS mbedtls/ssl.h HINTS ${MBEDTLS_FOLDER_INCLUDE})

if(UA_BUILD_OSS_FUZZ)
    # oss-fuzz requires static linking of libraries
    set(MBEDTLS_LIBRARY /usr/lib/x86_64-linux-gnu/libmbedtls.a)
    set(MBEDX509_LIBRARY /usr/lib/x86_64-linux-gnu/libmbedx509.a)
    set(MBEDCRYPTO_LIBRARY /usr/lib/x86_64-linux-gnu/libmbedcrypto.a)
else()
    find_library(MBEDTLS_LIBRARY mbedtls HINTS ${MBEDTLS_FOLDER_LIBRARY})
[...]

PR #7359, opened 2025-07-14
Fix: Prevent macro redefinition and improve MbedTLS path discovery

JianbinLiu-CFLab:
Hi @jpfr,

Thank you for the thorough review! I've updated the patch according to your suggestions:

**For the macro `MBEDTLS_ASN1_CHK_CLEANUP_ADD`:**
- Removed the `#undef` and replaced it with a proper `#ifndef` guard
- This avoids redefinition warnings in amalgamation builds while respecting any external definitions

**For `FindMbedTLS.cmake`:**
- Removed all hardcoded fallback paths (including those under `UA_BUILD_OSS_FUZZ`)
- Fixed the logic to properly honor:
  - User-specified variables (like `MBEDTLS_LIBRARY`)
  - Environment hints (like `MBEDTLS_FOLDER_LIBRARY`) 
  - `find_library()` as the final fallback when nothing else is defined

The changes should make the build system more portable and follow better practices. Please let me know if anything needs further adjustment!

Best regards
Jianbin Liu
CFLab


jpfr:
This is great now.
With the caveat that OSS-Fuzz no longer builds.
It appears the library lookup doesn't work there.

What you could do is re-establish the "hard coded paths" with the HINTS keyword.
Then they only get used if the paths exist and resolve to the library.
That is much better than true hardcoded paths.

JianbinLiu-CFLab:
Hi @jpfr,

Thanks again for your guidance and patience.

Just to clarify: I initially opened two separate PRs (#7355 and this one) mainly due to my lack of experience with contributing to open-source workflows. I now understand it's better to keep iterations within the same PR and branch, and I’ll definitely follow that in the future.

As for the latest update:
- I've replaced all `set(...)` statements with `find_library(...)` using `HINTS`, as you suggested.
- The fallback static paths (`/usr/lib/x86_64-linux-gnu`) are now included via `HINTS`, so they only apply if present.
- Verified that `cmake ..` runs successfully locally.

Let me know if further improvements are needed. I really appreciate your feedback—it’s been a valuable learning experience.

Best regards,  
Jianbin

JianbinLiu-CFLab:
Hi @jpfr,

Thanks again for your feedback and review.

After thinking it through a bit more, I felt there were still some areas that could be improved—especially in terms of robustness and alignment with standard CMake practices. So I went ahead and made one last update.

In particular:
- I removed the use of `libmbedtls.a` as a `NAMES` entry, and instead now use `CMAKE_FIND_LIBRARY_SUFFIXES` to enforce static linking in the OSS-Fuzz case.
- I added more portable fallback paths (`/usr/lib`, `/usr/lib64`, etc.) to help in more varied environments.
- The dynamic library discovery path was also simplified slightly to avoid redundancy.

If you feel the previous version was already good enough, feel free to disregard this one—I just wanted to submit something I’d feel more confident about maintaining long-term, should this stay in the project.

Thank you again for your time and support throughout this! I’ve learned a lot in this process.

Best regards,  
Jianbin

jpfr:
Nice, merged.

PR #7365, opened 2025-07-18
fix multiple run of UA_ClientConfig_setDefaultEncryption

niunuinui:
1. set config->certificateVerification.logging back after clear.
2. Clean up old SecurityPolicies before realloc. Add #None as first Policy

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -1187,16 +1187,16 @@ static UA_StatusCode
 clientConfig_setAuthenticationSecurityPolicies(UA_ClientConfig *config,
                                                UA_ByteString certificateAuth,
                                                UA_ByteString privateKeyAuth) {
+    /* Clean up old SecurityPolicies */
+    for(size_t i = 0; i < config->authSecurityPoliciesSize; i++) {
+        config->authSecurityPolicies[i].clear(&config->authSecurityPolicies[i]);
+    }
+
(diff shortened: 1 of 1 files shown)


niunuinui:
It seems some parts were already fixed in master

jpfr:
Your branch is called 1.4. but you merge to master. Is this intended?

i don’t think we need the none policy for the authentication.
That leaves the risk to have no encryption for both the securechannel and also the user identity token.

what is the use case here?

niunuinui:
Sorry It's not intended. I'm new to github pull request. I forked from v1.4 tag but didn't find the v1.4 branch, so I add a pull request to the master.

I'm also new to Open62541. So I was trying to use the lib on my project but find ``` UA_ClientConfig_setDefaultEncryption``` had some clean bug when called mutiple time.

The none policy was added for the consistance between the original code(v1.4) and my fix. In the v1.4 code, ```securityPoliciesSize``` was 1 when ```UA_ClientConfig_setDefaultEncryption``` was first called, then the added encryption policy was added after #none. So in my 'fix' code I add it back after my cleaning.

I do think you are right about the new behavior of the funcion. It seems my code is not needed in the master branch as the cleanning was already fixed by others.

jpfr:
No problem. I changed the target to 1.4.
This is still open then? Otherwise please close the PR yourself.

Issue #7326, opened 2025-06-23
Build failed with UA_ENABLE_SUBSCRIPTIONS_EVENTS=OFF


engycz:
## Background Information / Reproduction Steps

Used CMake options:

```bash
cmake -DUA_ENABLE_SUBSCRIPTIONS_EVENTS=OFF ..
```

```
In file included from /tmp/open62541/src/server/ua_session.c:12:
/tmp/open62541/src/server/ua_server_internal.h:42:9: error: unknown type name ‘UA_Server_EventNotificationCallback’; did you mean ‘UA_Server_DataChangeNotificationCallback’?
   42 |         UA_Server_EventNotificationCallback eventCallback;
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |         UA_Server_DataChangeNotificationCallback
```

## Checklist

Please provide the following information:

 - [ ] open62541 Version (release number or git tag): master

andreasebner:
Issue should be solved by merging #7326. This change is also included in 1.5.

PR #7370, opened 2025-07-21
fix(server): Fixed issue building with UA_ENABLE_SUBSCRIPTIONS_EVENTS=OFF (#7326)

Killian4096:
Fixed (#7326). Issue seemed to be that struct callback would attempt to include UA_Server_EventNotificationCallback in a union when it was not defined. Additionally, headers for both UA_Server_createEventMonitoredItem and UA_Server_createEventMonitoredItemEx had correct preprocessor dirrectives, buyt function implementation did not, causing another compilation error.


jpfr:
👍

PR #7374, opened 2025-07-25
refactor(examples,tools,server,plugins): clang compiler warnings

laroche:
- not-needed semicolons
- unused macros
- sign conversion in ua-cli


jpfr:
Nice, good finds.

PR #7381, opened 2025-08-01
fix(server): Wrong out value for asyncOperationCancelCallback() for the call service

basyskom-jvoe:


jpfr:
Uhh. Good find.
Thanks for putting this under test also.

PR #7384, opened 2025-08-03
Fix: Deregistration timing, config cleanup, and invalid mDNS IP in server_multicast.c

JianbinLiu-CFLab:
Hi @jpfr,

This patch addresses three critical issues in `examples/discovery/server_multicast.c`:

1. **Incorrect shutdown order**:  
   `UA_Server_deregisterDiscovery()` was called after `UA_Server_run_shutdown()`,  
   which leads to undefined behavior, as the server instance is already partially destroyed.
2. **Missing UA_ClientConfig cleanup on failure**:  
   If `getRegisterClient()` fails, `UA_ClientConfig_clear()` was not called, causing potential memory leaks.
3. **Invalid mDNS interface IP**:  
   The example used `0.0.0.0` as the outbound interface IP for libmdnsd.  
   This is not valid for multicast sending. It has been replaced with `127.0.0.1` by default,  
   along with a comment explaining when to use a real interface IP (e.g., `192.168.x.x`) for deployment.

All fixes have been tested under local and loopback conditions.  
Let me know if you'd prefer the commits to be split or if anything needs clarification.

Best regards,  
Jianbin Liu (CFLab)


JianbinLiu-CFLab:
Closing in favor of a cleaner PR containing only the intended fix for `server_multicast.c`.
Apologies for the noise — will open a new one shortly!

PR #7391, opened 2025-08-07
Add functions to get PubSub configuration component informations

andreasebner:


include/open62541/server_pubsub.h
```diff
@@ -289,6 +296,54 @@ typedef struct {
     UA_PubSubSecurityPolicy *securityPolicies;
 } UA_PubSubConfiguration;
 
+/**
+ * PubSub Components
```
jpfr: The documentation and #define are redundant.
See ca 100 lines above.

PR #7399, opened 2025-08-10
refactor(examples): set allocated ApplicationName in server_lds example

JianbinLiu-CFLab:
Hi @jpfr,

- Add a custom `applicationName` to the LDS example.
- Use `UA_LOCALIZEDTEXT_ALLOC` to ensure correct memory ownership so `UA_Server_delete()` frees it safely (avoid invalid free on shutdown).
- Add NULL-check for `serverCapabilities` allocation.

Thanks, 
Jianbin

### DIFF
--- a/examples/discovery/server_lds.c
+++ b/examples/discovery/server_lds.c
@@ -1,19 +1,25 @@
 /* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
- * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. 
+ * Copyright (c) 2025 Construction Future Lab gGmbH (Author: Jianbin Liu)
+ */
 /*
  * Server representing a local discovery server as a central instance.
  * Any other server can register with this server (see server_register.c). Clients can then call the
  * find servers service to get all registered servers (see client_find_servers.c).
+ * NOTE: Requires a local mDNS/Bonjour service (e.g., Avahi on Linux) to be running.
+ * Without it, multicast discovery will not work.
  */
 
 #include <open62541/server.h>
 #include <open62541/server_config_default.h>
+#include <open62541/plugin/log_stdout.h>
 
 #include <signal.h>
 #include <stdlib.h>
 
 static UA_Boolean running = true;
 static void stopHandler(int sig) {
+    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "received Ctrl-c");
     running = false;
 }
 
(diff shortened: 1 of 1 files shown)


jpfr:
Just for my understanding:

> Use UA_LOCALIZEDTEXT_ALLOC to ensure correct memory ownership so UA_Server_delete() frees it safely (avoid invalid free on shutdown).

Wasn't that the case before?
I don't see any stack-allocation of that name being removed from the code.

examples/discovery/server_lds.c
```diff
@@ -23,17 +27,36 @@ int main(void) {
...
             UA_String_fromChars("urn:open62541.example.local_discovery_server");
+    
+    /* Set a custom ApplicationName
+     * IMPORTANT: Must use an allocated LocalizedText (e.g. UA_LOCALIZEDTEXT_ALLOC or
+     * UA_LocalizedText_fromChars) so that UA_Server_delete() can free the memory safely.
+     * Using UA_LOCALIZEDTEXT with string literals will cause an invalid free() on shutdown. */
+    UA_LocalizedText_clear(&config->applicationDescription.applicationName);
+    config->applicationDescription.applicationName =
+        UA_LOCALIZEDTEXT_ALLOC("en-US", "Open62541 Local Discovery Server");
+    /* Alternative:
```
jpfr: The alternative is not needed.

examples/discovery/server_lds.c
```diff
@@ -23,17 +27,36 @@ int main(void) {
...
     // This is an LDS server only. Set the application type to DISCOVERYSERVER.
     // NOTE: This will cause UaExpert to not show this instance in the server list.
     // See also: https://forum.unified-automation.com/topic1987.html
+    // NOTE: In production, the ApplicationURI must be globally unique and stable for this application instance.
     config->applicationDescription.applicationType = UA_APPLICATIONTYPE_DISCOVERYSERVER;
     UA_String_clear(&config->applicationDescription.applicationUri);
     config->applicationDescription.applicationUri =
             UA_String_fromChars("urn:open62541.example.local_discovery_server");
+    
+    /* Set a custom ApplicationName
```
jpfr: The comment is not needed.
When you first call _clear the applicationName must have been heap-allocated already.
So you do it to get a "nice" name. But there is no "memory fix" here.

examples/discovery/server_lds.c
```diff
@@ -23,17 +27,36 @@ int main(void) {
...
+    UA_LocalizedText_clear(&config->applicationDescription.applicationName);
+    config->applicationDescription.applicationName =
+        UA_LOCALIZEDTEXT_ALLOC("en-US", "Open62541 Local Discovery Server");
+    /* Alternative:
+    config->applicationDescription.applicationName =
+        UA_LocalizedText_fromChars("en-US", "Open62541 Local Discovery Server");
+    */
 
     // Enable the mDNS announce and response functionality
+    // NOTE: Requires a local mDNS/Bonjour service (e.g., Avahi on Linux) to be running.
```
jpfr: This comment is good.
But it should go to the explanation at the beginning of the file.

jpfr (commented):
Comments inline.

JianbinLiu-CFLab:
Hi @jpfr, 
I have updated the code according to your request. Understood — the change is indeed not a memory fix for the current default behavior. My motivation came from a previous case where I changed the applicationName to a static literal (not heap-allocated), and later UA_Server_delete() unconditionally called UA_LocalizedText_clear(), which then tried to free() it → causing an invalid free. So here I used UA_LOCALIZEDTEXT_ALLOC as a safe pattern for customization, to avoid that pitfall when users modify the example, and I appreciate your precise review and clarification.

jpfr:
Merged, nice!

Issue #7390, opened 2025-08-06
Nodeset compiler error in examples/server_nodeset when changing UA_NAMESPACE_ZERO to FULL


V-Strassheim:
## Description
When following the build instructions and enabling examples with the full namespace zero, the build fails at examples/server_nodeset with the following error:

```bash
Exception: Reference ns=1;i=15235--[ns=0;i=40]-->ns=0;i=18791 has an unknown target
make[2]: *** [examples/nodeset/CMakeFiles/open62541-generator-ns-testnodeset.dir/build.make:84: src_generated/open62541/namespace_testnodeset_generated.c] Error 1
make[1]: *** [CMakeFiles/Makefile2:1955: examples/nodeset/CMakeFiles/open62541-generator-ns-testnodeset.dir/all] Error 2
```

## Background Information
I was able to trace the issue back to changes introduced in [PR #7333](https://github.com/open62541/open62541/pull/7333).

The root cause appears to be that the CMake cache for namespace zero (NS0) is not updated after the initial cmake call, resulting in outdated or incomplete nodeset dependencies during build.

Reproduction Steps:

```bash
mkdir build
cd build
cmake ..
ccmake ..          # Set: UA_BUILD_EXAMPLES=ON, UA_NAMESPACE_ZERO=FULL
make

```
**Note:** The error occurs **only when changing `UA_NAMESPACE_ZERO` between CMake configurations** (e.g., switching from `REDUCED` to `FULL` using `ccmake` or `cmake-gui`). It does **not** happen when using `cmake` directly with the preferred options from the beginning.

@NoelGraf Do you have any idea how best to fix this?

Additional used CMake options:

UA_BUILD_EXAMPLES=ON, UA_NAMESPACE_ZERO=FULL

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): master
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Debian 12
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

r-sin:
This looks like an issue that was fixed on 1.4:
https://github.com/open62541/open62541/issues/6622

V-Strassheim:
> This looks like an issue that was fixed on 1.4: [#6622](https://github.com/open62541/open62541/issues/6622)


[...]

PR #7404, opened 2025-08-11
Avoid caching open62541_* cmake variables

NoelGraf:
The PR fixes the problem described in the issue https://github.com/open62541/open62541/issues/7390 .


jpfr:
👍

PR #7405, opened 2025-08-12
refactor(examples,server): clang compiler warnings

laroche:
- sign conversions
- not-needed semicolons
- adding "static" to file-local vars


jpfr:
Nice, merged.

PR #7407, opened 2025-08-13
Fix APPLE || MACH ifdef to only trigger on APPLE && MACH

Zopolis4:
This is intended to only trigger on OSX, but since Hurd uses Mach as well it overmatches when it should be checking for both `__APPLE__` and `__MACH__`.

https://www.gnu.org/software/hurd/hurd/porting/guidelines.html

```
#ifdef __MACH__
Some applications put Apple Darwin-specific code inside #ifdef __MACH__ guards. Such guard is clearly not enough, since not only Apple uses Mach as a kernel. This should be replaced by #if defined(__MACH__) && defined(__APPLE__)
```

### DIFF
--- a/arch/posix/clock_posix.c
+++ b/arch/posix/clock_posix.c
@@ -18,7 +18,7 @@
 #include <time.h>
 #include <sys/time.h>
 
-#if defined(__APPLE__) || defined(__MACH__)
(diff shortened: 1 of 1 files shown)


jpfr:
Nice, merged.

PR #7410, opened 2025-08-13
Correct the invalid XML values in the testnodeset example

NoelGraf:
* The server_testnodeset example is currently unable to start successfully. The following error message is displayed:

Bild

* Fix flaky client async test case with lwip

### DIFF
--- a/tools/ci/linux/examples_with_valgrind.py
+++ b/tools/ci/linux/examples_with_valgrind.py
@@ -55,7 +55,6 @@ server_needed_examples = {
         "tutorial_client_events":"tutorial_server_events",
         "tutorial_client_firststeps":"tutorial_server_firststeps",
         "tutorial_pubsub_connection":"tutorial_pubsub_subscribe",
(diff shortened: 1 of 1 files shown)


jpfr:
The lwip build is flaky: `check_activateSessionAsync` 

Please add the testnodeset to the CI.
This should have failed long before.

NoelGraf:
This example is already being executed in the tests. However, when run under Valgrind, if it exits with a code of 1, nothing happens and the remaining tests continue.
grafik


One option is to update the Python script so that it terminates if the exit code is not 0.

Alternatively, we could add a dedicated CI test to run the examples without Valgrind and validate their exit codes.

jpfr:
> Alternatively, we could add a dedicated CI test to run the examples without Valgrind and validate their exit codes.

I thought we already had that.
There is a python script to execute alle examples.

Please reactivate or add such a functionality.

jpfr:
Nice 😀

PR #7421, opened 2025-08-20
add function to register remote server record to lds

engel-bichlerh:
Add a public function to ua_services_discovery to add a programmatically created ua server record to the own discovery list.


.github/workflows/build_linux.yml
```diff
@@ -10,44 +10,49 @@ jobs:
         os: [ubuntu-20.04, ubuntu-22.04, ubuntu-24.04]
         build:
           - name: "Debug Build & Unit Tests (gcc)"
+            cmd_deps: sudo apt-get install -y -qq curl
```
jpfr: Lets install curl globally for all builds.
So we don’t Need to make the change in 20 different places….

include/open62541/server.h
```diff
@@ -1397,6 +1397,25 @@ void UA_EXPORT UA_THREADSAFE
 UA_Server_setRegisterServerCallback(UA_Server *server,
                                     UA_Server_registerServerCallback cb, void* data);
 
+/* Register a server with this discovery server.
```
jpfr, line 1677: Document the return code.
Why not return a global statuscode?

include/open62541/server.h
```diff
@@ -1397,6 +1397,25 @@ void UA_EXPORT UA_THREADSAFE
...
+ * @param requestDiscoveryConfiguration the discovery configuration of the server record
+ * @param responseHeader the response header to fill
+ * @param responseConfigurationResultsSize the size of the configuration results
+ * @param responseConfigurationResults the configuration results to fill
+ */
+UA_Boolean UA_EXPORT UA_THREADSAFE
+UA_Server_registerServer(UA_Server *server, const UA_RegisteredServer *requestServer,
+                         const size_t requestDiscoveryConfigurationSize,
+                         const UA_ExtensionObject *requestDiscoveryConfiguration,
+                         UA_ResponseHeader *responseHeader,
```
jpfr: If we return a global statuscode instead of boolean, then the response header is not needed.

include/open62541/server.h
```diff
@@ -1397,6 +1397,25 @@ void UA_EXPORT UA_THREADSAFE
...
+ * @param requestDiscoveryConfigurationSize the size of the configuration
+ * @param requestDiscoveryConfiguration the discovery configuration of the server record
+ * @param responseHeader the response header to fill
+ * @param responseConfigurationResultsSize the size of the configuration results
+ * @param responseConfigurationResults the configuration results to fill
+ */
+UA_Boolean UA_EXPORT UA_THREADSAFE
+UA_Server_registerServer(UA_Server *server, const UA_RegisteredServer *requestServer,
+                         const size_t requestDiscoveryConfigurationSize,
+                         const UA_ExtensionObject *requestDiscoveryConfiguration,
```
jpfr: ExtensionObject is basically untyped.
Document the different supported options for the content.

How about we define a union type just for the C API?

jpfr (commented):
Very nice. Some comments online.

include/open62541/server.h
engel-bichlerh: Changed to a global statuscode and removed unused input, output parameters to simplify function.

PR #7321, opened 2025-06-17
fix(tools): Fixed parsing of LocalizedText in nodeset_compiler (#7279)

julepp:
The Format of the LocalizedText XML-Element that the parser expected was not to specification. It has been changed to expect the following:
```xml
<LocalizedText locale="en">Text</LocalizedText>
```
Whenever a LocalizedText is created, the whole XML Element has to be passed now to be able to parse the locale. Before, only the child element was passed. If no Locale is specified, it remains None.

The class LocalizedText in datatypes.py was adjusted.
The class Node and ReferenceTypeNode in nodes.py

PR #7422, opened 2025-08-20
fix(tools) Fixes bug created in PR #7321

julepp:
Pull Request 7321 introduced a bug which caused an error in parsing some LocalizedText elements.
This was caused by whitespace being interpreted as a valid Text node in the LocalizedText element. 

To fix the bug, it will now be checked first if an element with the name Text exists. If it does, then the content of this element will be used as the text for the LocalizedText.
Only if such an element does not exist will the text node be used.


jpfr:
Thanks for this.
The code is non-obvious.
Please add a comment for the different syntax it supports.
And which is the „true“ syntax prescribed by the standard.

jpfr:
Merged. Thanks!

PR #7430, opened 2025-08-28
refactor(server): check return value of malloc()

laroche:
Check the return value of malloc() to prevent a
Null Pointer Exception (NPE).


jpfr:
Thanks for this!

PR #7443, opened 2025-09-11
Fix dead link to certified SDK page

paskozdilar:
Current index.rst document contains the following line:

```
https://open62541.org/certified-sdk for more details.
```

However, https://open62541.org/certified-sdk is a dead link.

This PR replaces it with https://open62541.org/certification which contains the relevant details.


paskozdilar:
Seems like this is changed already. Closing.

PR #7451, opened 2025-09-16
Update encryption examples to demonstrate secure-only policy usage

NoelGraf:
This PR improves the documentation and the encryption examples to demonstrate the usage of the UA_ServerConfig_setDefaultWithSecureSecurityPolicies function.
When using this function, the None security policy is not added to the list of policies.

As a result:
Clients cannot retrieve the server certificate or discover all available server endpoints.
A connection is only possible if one or more endpoints are manually configured on the client side.
Connecting with UAExpert is not possible, even in Advanced Mode. This is because UAExpert requires discovery to obtain the server certificate.
To enable connections with UAExpert, the None security policy must be included in the list.
However, exposing an actual None endpoint is not required.

### DIFF
--- a/examples/encryption/client_encryption.c
+++ b/examples/encryption/client_encryption.c
@@ -7,6 +7,7 @@
 #include <open62541/plugin/securitypolicy.h>
 #include <open62541/server.h>
 #include <open62541/server_config_default.h>
+#include <open62541/plugin/create_certificate.h>
 
 #include <stdlib.h>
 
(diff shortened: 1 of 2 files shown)


neojaw:
There is a field `securityPolicyNoneDiscoveryOnly` in the server config. Should that be set in case `onlySecure` was requested?

NoelGraf:
Yes, exactly—like the UA_ServerConfig_setDefaultWithSecureSecurityPolicies function does at the end.

neojaw:
Indeed. But this change also affects `UA_ServerConfig_addAllSecureSecurityPolicies` on its own (and anything else I missed?) which doesn't do that. Is that intended?

NoelGraf:
You mean that the flag securityPolicyNoneDiscoveryOnly should already be set in the function UA_ServerConfig_addAllSecureSecurityPolicies?

neojaw:
Even further, I think `addAllSecurityPolicies` called with `onlySecure=true` should either
* not add the None policy at all and leave it to higher level callers like `UA_ServerConfig_setDefaultWithSecureSecurityPolicies` or
* add the None policy but then set the `securityPolicyNoneDiscoveryOnly` flag at the same time.

I don't know, having an `onlySecure` parameter that still adds a fully functional insecure mode feels like a trap for future callers, even if it's an internal function.

PR #7457, opened 2025-09-23
refactor(common,deps,server,util,tests): fix "gcc -Wconversion" warnings

laroche:
Fix most warnings from "gcc -Wconversion".


jpfr:
Nice, merged.

Does it make sense to activate -Wconversion everywhere now?

laroche:
A few warnings about -Wconversion are still open, so as of now I'd keep this option away from default settings.

best regards

Florian La Roche

PR #7465, opened 2025-09-29
feat(server): Extended the access control plugin with check for receiving events

mcpat-pilz:
The ReceiveEvents permission from the OPC UA specification was missing in the AccessControl plugin. This change adds it now.


mcpat-pilz:
@jpfr This should work now.

andreasebner:
Thank's for your input. The master already enforces per-MonitoredItem event-receive authorization
natively via RBAC — covering both EventType and SourceNode. This PR seems therefore to be oudated.

mcpat-pilz:
@andreasebner May be I misunderstand you, but the purpose of this PR is to support checking for this permission without enabling RBAC.

PR #7467, opened 2025-09-30
feat(server): New allowCreateSubscription check for access control plugin

mcpat-pilz:
Subscriptions consume resources on the server. It should be possible to limit this for example to forbid anonymously authenticated sessions to create subscriptions at all.


andreasebner:
Solved via #7918

PR #7479, opened 2025-10-07
refactor(deps): Update to latest NodesetLoader as dependency

laroche:
The nodesetLoader is currently pointing to an old hash. ".gitmodules" is pointing to the default
branch, so I've updated to the newest commit of it.

The other submodules should probably also be updated to newer commits?

best regards,

Florian La Roche


laroche:
The submodule directories in deps/* all contain individual hashes for the individual subprojects.
But the ".gitmodules" file contains the branchname "master" and "latest" instead of these commit hashes.
Should they also be changed to these commit hashes as branchnames?

(Submodules are confusing to me.)

best regards,

Florian La Roche

jpfr:
👍

PR #7504, opened 2025-10-25
Remove enabling of all PubSub components

jackybek:
Removed enabling of all PubSub components before server run.


andreasebner:
See the comment in #7510.

PR #7505, opened 2025-10-25
Remove enabling of all PubSub components

jackybek:
Removed enabling of all PubSub components in the server.


andreasebner:
See the comment in #7510.

PR #7507, opened 2025-10-25
Remove redundant call to enable all PubSub components

jackybek:


andreasebner:
See the comment in #7510.

PR #7508, opened 2025-10-25
Remove unnecessary enableAllPubSubComponents call

jackybek:


andreasebner:
See the comment in #7510.

PR #7509, opened 2025-10-25
Update tutorial_pubsub_connection.c

jackybek:


andreasebner:
See the comment in #7510.

PR #7510, opened 2025-10-25
Remove redundant enableAllPubSubComponents call

jackybek:


andreasebner:
Thank you for your active involvement in open62541. You have submitted a large number of pull requests on one topic, which makes it difficult for us as maintainers. I am therefore closing your other PRs, but you are welcome to integrate all your changes into this one. Please also explain what the changes are intended to achieve and why they are necessary. Best regards.

jackybek:
this function does not exist anymore.  i checked the latest v1.4.14 pubsub examples and they did not have this function.
so this function is defunct already. 
but the examples in github still calls this function so it is
misleading.

neojaw:
Code in master is newer than releases. That function was added in #6687 when the PubSub mechanism was changed drastically. This change will not become part of v1.4.x because it changed the public API, but it will be part of future releases. Calling that function is necessary when using the new code.

jackybek:
are you sure ? i already downloaded v1.4.4 and compiled the pubsub example, which did not include enablelPubSubComponent.

i also tried to compile with this function call in my
program but it failed - function not found.

so what you said do not make sense.

neojaw:
Both revisions are consistent.
* v1.4.x does not have the function and the examples in v1.4.x don't call it.
* master has the function and the examples in master call it. (Since the PR and the Pubsub overhaul I mentioned.)

What you have to do in your own code depends on which version you are using.

jpfr:
There seems to have been a misunderstanding with the versions.
Please reopen when required.

PR #7516, opened 2025-10-31
[WIP] customisable certificate verification

mcpat-pilz:
This PR introduces an API change to allow the configuration of the certificate validation.
It can be configured which level of validation is requested as well as what types of certificates (Instance Certificate, User Certificate or Issuer Certificate) are allowed.

Further this PR will add the missing certificate and trust list checks for the GDS push management.

### DIFF
--- a/drivers/gds_receiver/gds_receiver.c
+++ b/drivers/gds_receiver/gds_receiver.c
@@ -658,15 +658,6 @@ UA_GDSReceiver_addCertificate(UA_GDSReceiverContext *ctx,
     UA_Server *server = ctx->drv.server;
     UA_ServerConfig *sc = UA_Server_getConfig(server);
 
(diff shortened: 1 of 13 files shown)


mcpat-pilz:
@jpfr Is this something you can imagine to merge in the future? Then I will also adapt the OpenSSL code to have it complete.

jpfr:
Yes, this is a good direction.
We then can move many more checks inside the certificate group and out of the core library.

a few comments:

Let’s allow fine-grained check selection with a uint32 bitmask for the enum.
Then the output can return the same bitmask to identify which test failed. We need this detailed information to trigger the correct audit event.

Fields in a bitfield struct only merge when they have the same type.

jpfr:
Give us a trigger for an eventual review.

PR #7523, opened 2025-11-06
refactor(server): use MDNSD_free() instead of UA_free() for "packet"

laroche:
"packet" is allocated within sd2txt() and with MDNSD_malloc(), so use MDNSD_free() instead of UA_free().


jpfr:
Merged, thanks!

PR #7537, opened 2025-11-13
[Review] fix(arch): fix IEEE 802.1Q VLAN TCI Creation/Parsing

fgeiger-est:
The VLAN TCI field is now correctly created and parsed in eventloop_posix_eth.c


jpfr:
Thanks for this.
Can you make the PR against 1.4 or 1.5?
Whichever is the first that applies.

jpfr:
👍

PR #7538, opened 2025-11-13
[Review] fix(pubsub): adjust UADP over Ethernet connection function

fgeiger-est:
Now the vid and pcp are correctly parsed from the endpoint url and will be set in the keyvaluemap for eventloop handling, also this will add the ethertype for uadp over ethernet


jpfr:
The same, please make this a PR against 1.4.

jpfr:
Merged. Thanks.

PR #7540, opened 2025-11-13
Fix conditionRefresh method callback

basyskom-jvoe:
The refreshMethodCallback() function attempted to call refreshLogic() with a NULL monitoredItem parameter which triggered an assert().


jpfr:
👍

Issue #7543, opened 2025-11-14
Check certifcate chain broken for windows


soekkle:
## Description

For a pure windows compilation the check of certifacate cain with pem files is brocken.

## Background Information / Reproduction Steps

I implement a test case check_ca_chain  at https://github.com/soekkle/open62541/commit/a2ecb6f372d29cc9f1b14901de22952a91708082. It fails on windows with the used CMake options:
```
BUILD_SHARED_LIBS:BOOL=OFF
CLANG_FORMAT_EXE:FILEPATH=CLANG_FORMAT_EXE-NOTFOUND
CMAKE_BUILD_TYPE:STRING=Debug
CMAKE_CXX_COMPILER:FILEPATH=C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe
CMAKE_INSTALL_PREFIX:PATH=C:/Users/soren/Source/Repos/soekkle/open62541/out/install/x64-Debug
CMAKE_MAKE_PROGRAM:FILEPATH=C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/Ninja/ninja.exe
UA_AMALGAMATION_MULTIARCH:BOOL=OFF
UA_ARCHITECTURE:STRING=win32
UA_BUILD_EXAMPLES:BOOL=OFF
UA_BUILD_TOOLS:BOOL=OFF
UA_BUILD_UNIT_TESTS:BOOL=TRUE
UA_ENABLE_AMALGAMATION:BOOL=OFF
UA_ENABLE_DA:BOOL=ON
UA_ENABLE_DATATYPES_ALL:BOOL=ON
UA_ENABLE_DIAGNOSTICS:BOOL=ON
UA_ENABLE_DISCOVERY:BOOL=ON
UA_ENABLE_ENCRYPTION:STRING=OPENSSL
UA_ENABLE_ENCRYPTION_TPM2:STRING=OFF
UA_ENABLE_GDS_PUSHMANAGEMENT:BOOL=OFF
UA_ENABLE_HISTORIZING:BOOL=ON
UA_ENABLE_JSON_ENCODING:BOOL=ON
UA_ENABLE_METHODCALLS:BOOL=ON
UA_ENABLE_NODESETLOADER:BOOL=OFF
UA_ENABLE_PUBSUB:BOOL=ON
[...]

PR #7544, opened 2025-11-14
fix(encryption) Fix certifacate cain check for windows

soekkle:
Fixes issue https://github.com/open62541/open62541/issues/7543 for v 1.4.14

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -141,6 +141,49 @@ isFullyConnected(UA_Client *client) {
 
 #ifdef UA_ENABLE_ENCRYPTION
 
+/* Find binary substring. Taken and adjusted from
+ * http://tungchingkai.blogspot.com/2011/07/binary-strstr.html */
+
+static const unsigned char *
+bstrchrClient(const unsigned char *s, const unsigned char ch, size_t l) {
+    /* find first occurrence of c in char s[] for length l*/
+    for(; l > 0; ++s, --l) {
+        if(*s == ch)
+            return s;
+    }
+    return NULL;
+}
+
+static const unsigned char *
+UA_BstrstrClient(const unsigned char *s1, size_t l1, const unsigned char *s2, size_t l2) {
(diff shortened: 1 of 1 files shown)


jpfr:
Thanks for this.
The new file encryption/check_ca_chain.c is missing though.

soekkle:
I hope it is under test/encryption/check_ca_chain.c

include/open62541/util.h
```diff
@@ -287,6 +287,10 @@ UA_constantTimeEqual(const void *ptr1, const void *ptr2, size_t length);
 UA_EXPORT void
 UA_ByteString_memZero(UA_ByteString *bs);
 
+/* Find binary substring in a string*/
+UA_EXPORT const unsigned char *
+UA_Bstrstr(const unsigned char *s1, size_t l1, const unsigned char *s2, size_t l2);
```
jpfr: We can pull out the bstrstr functionality. But do it in a second PR:

- The change should go into v1.5 or master. Because many distributions (e.g. Debian) complain when we change the exported symbols in a patch release
- Call the function `UA_ByteString_findSubstring` or similar. And work on UA_ByteString instead of naked pointers and sizes.

src/client/ua_client_connect.c
```diff
@@ -162,18 +165,38 @@ signClientSignature(UA_Client *client, UA_ActivateSessionRequest *request) {
     if(retval != UA_STATUSCODE_GOOD)
         return retval;
 
+    UA_ByteString rc;
+    UA_ByteString_copy(&client->channel.remoteCertificate, &rc);
+    {
```
jpfr: Why is this wrapped in curly braces?

This needs a comment to document what is going on and why.

jpfr:
Thanks for the update.
I made some small changes and force-pushed on the same branch.
Your are still the author of all the commits of course.

Added more context to the commit message and more comments to the functional logic.
This is pretty nice and tight now!

PR #7546, opened 2025-11-19
Fix segfault triggered by createEvent() debug output

basyskom-jvoe:
The %su placeholder treats the severity as a string which leads to a segfault if the event's severity is not 0.


jpfr:
👍
Commit title adjusted to our format.

PR #7563, opened 2025-11-26
[WIP] fix(server): Update safe iterator before deleting

daniel-petrovic:
Getting random crashes on a RT arm-linux embedded caused by double free in ua_subscription.c.
521338937-a47fd0d3-60e5-4547-8cb9-256598a92ce9

We are stuck at the version 1.3, so I cannot verify that the crash happens in the current master too, but the code is very similar.
In this PR are analog changes applied to our 1.3 version, which seems to work, but we are still testing it.

It could be though, that the real cause is somewhere else.

### DIFF
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -543,17 +543,21 @@ prepareNotificationMessage(UA_Server *server, UA_Subscription *sub,
          * local queue, remove all of them. These are earlier Notifications that
          * are non-reporting. And we don't want them to show up after the
          * current Notification has been sent out. */
-        UA_Notification *prev;
-        while((prev = TAILQ_PREV(n, NotificationQueue, monEntry))) {
-            UA_Notification_delete(prev);
+        UA_MonitoredItem *mon = n->mon;
+        UA_Notification *curr = TAILQ_FIRST(&mon->queue);
+        while(curr && curr != n) {
+            UA_Notification* next = TAILQ_NEXT(curr, monEntry);
+            if (curr == n_tmp) {
+                /* Update the safe iterator before deleting it */
+                n_tmp = TAILQ_NEXT(curr, subEntry);
+            }
 
-            /* Help the Clang scan-analyzer */
-            UA_assert(prev != TAILQ_PREV(n, NotificationQueue, monEntry));
+            UA_Notification_delete(curr);
+            curr = next;
         }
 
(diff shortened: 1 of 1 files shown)


jpfr:
Is this a theoretical bug or does it just occur in the wild?
It would be great to have a unit test as reproducer.
To avoid future breakage…

daniel-petrovic:
We are getting random crashes on our devices. Crashes are hard reproducable (e.g. after 2/3 h running).
But we still have to use  version 1.3. 

Because the relevant code looks very similar compared to the current master I assume there could be potential issue with the  current master too, but I cannot verify that.

@jpfr  Drop some info if I should add some specific tests to this.

daniel-petrovic:
UPDATE: The fix reduced the bug but did not eliminate it. Crashes are now reproducible only in the release build; the debug build works fine.

I suspect the real cause lies elsewhere. From examining the protocol, it looks like we have a misbehaving client which causes the bug in open62541 server.
 The client subscribes to the node but also issues ReadRequests. The polling interval for these ReadRequests is 100 ms, which matches the current publishingInterval.

It seems that the ReadRequests are consuming notifications from the publishing queue.

@jpfr  Could you provide some guidance on what is the expacted behaviour in this case and where to look? I only occasionally dive into the open62541 internals. Thanks!

jpfr:
@daniel-petrovic , are you on a Linux-based system?
Then you could let a debug build run with valgrind. That gives you log messsages and insights whenever a bad memory access happens.
Also when it doesn't crash the server.

If you want to exchange outside the public forum, write me at julius.pfrommer@o6-automation.com

daniel-petrovic:
> @daniel-petrovic , are you on a Linux-based system? Then you could let a debug build run with valgrind. That gives you log messsages and insights whenever a bad memory access happens. Also when it doesn't crash the server.
> 
> If you want to exchange outside the public forum, write me at [julius.pfrommer@o6-automation.com](mailto:julius.pfrommer@o6-automation.com)

@jpfr yes, we are on embedded RT linux . It will probably be hard with valgrind, because of it's big impact on runtime performance and we have some strict RT requirements. I already tried with gcc address sanitizers, but it influences a system a lot, so device is not really usable (at least not in the same way) - and valgrind is typically slower. but I will  check what I can do. otherwise I'll reach out. thanks

PR #7572, opened 2025-12-02
fix(plugins): Extend expiresInDays range for OpenSSL cert creation on 32 bit

basyskom-jvoe:
X509_time_adj() takes seconds as an int value. On a 32 bit build, this artificially limits the possible time range.

This should also be relevant for the 1.5 branch.


jpfr:
Nice, good improvement.

Issue #7216, opened 2025-04-09
Memory leak in ua_pki_openssl.c, UA_ReloadCertFromFolder function.


xydan83:
Please provide the following information:

 - [x] open62541 Version (release number or git tag): **v1.4.11**
 - [x] Other OPC UA SDKs used (client or server): **Server**
 - [x] Operating system: **Ubuntu 22.04.1**
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

Hi!

When using the `UA_CertificateVerification_CertFolders` function I get a **valgrind** warning about a memory leak.
The leak is in the `UA_ReloadCertFromFolder` function, which is used in `UA_CertificateVerification_Verify`, which is set as a callback (`cv->verifyCertificate`) in the `UA_CertificateVerification_Trustlist` and `UA_CertificateVerification_CertFolders` functions.

![Image](https://github.com/user-attachments/assets/35691d85-4b32-4657-a35b-801a07dadcd8)

I think if we use `scandir` we create `struct dirent ** dirlist = NULL;` but I didn't find where we clear that double pointer.

Then when I add `free` to the code, the **leak stops** and Valgring no longer warns.. 

![Image](https://github.com/user-attachments/assets/f1f34dc3-6808-4e2c-ac46-578346c32c1d)

I found three places in the code where we don't clear this double pointer in the `UA_ReloadCertFromFolder` function.
1) https://github.com/open62541/open62541/blob/2aa764b640c3f049e71e64661ff10eb1f513878c/plugins/crypto/openssl/ua_pki_openssl.c#L305
2) https://github.com/open62541/open62541/blob/2aa764b640c3f049e71e64661ff10eb1f513878c/plugins/crypto/openssl/ua_pki_openssl.c#L340
3) https://github.com/open62541/open62541/blob/2aa764b640c3f049e71e64661ff10eb1f513878c/plugins/crypto/openssl/ua_pki_openssl.c#L375

xydan83:
Update: New bug in the same module...

Double allocate memory in **UA_CertificateVerification_Verify** function. Define **UA_ENABLE_CERT_REJECTED_DIR**.

![Image](https://github.com/user-attachments/assets/21f85f0b-bd6f-48c7-bc82-982e543e62ad)

[...]

PR #7582, opened 2025-12-09
fix(plugin): double allocation of certificate thumbprint

JohannesKauffmann:
This code was introduced in 805c1ab17a1d4b4ecca2f1c9b11b9185cac2346a. UA_Openssl_X509_GetCertificateThumbprint() will allocate the thumbprint UA_ByteString itself, unlike mbedtls_thumbprint_sha1() (only when the last argument is true). As such, for the OpenSSL plugin, do not allocate the certificate thumbprint ourselves to fix double allocation.

Reported-by: xydan83 <aleksprog@hotmail.com>
Ref #7216


jpfr:
Good find. However you should still initialize the thumbprint variable.
Otherwise, in case something goes wrong, bad memory gets cleaned up.

JohannesKauffmann:
Thanks for the review, indeed you're right. Changed.

jpfr:
Merged. Thanks!

PR #7584, opened 2025-12-09
fixc(build): Use OpenSSL::Crypto target in CMake (fixes #7574)

### LINKED ISSUE #7574: OpenSSL link flags wrong using open62541.pc
## Description
When OpenSSL is enabled for open62541, the `${OPENSSL_LIBRARIES}` are appended to `${open62541_LIBRARIES}`, to be able to do:
```
target_link_libraries(open62541 PRIVATE ${open62541_LIBRARIES})
```
But (at least on my system), `${OPENSSL_LIBRARIES}` contain full paths to the .so files:
```
/usr/lib/x86_64-linux-gnu/libssl.so
/usr/lib/x86_64-linux-gnu/libcrypto.so
```

Since cec93e44d3c3589b32fee0e79e868aac1a5a90da started to append `${open62541_LIBRARIES}` to `${pkgcfgpubliclibs}` and `${pkgcfglibs}` was removed, the full paths to OpenSSL libraries is added with the `-l` prefix to the `Libs:` section of `open6251.pc`.

Which ends up looking like so:
```
libdir=/my/install/path/lib
sharedlibdir=${libdir}
includedir=/my/install/path/include

Name: open62541
Description: open62541 is an open source C (C99) implementation of OPC UA
Version: 1.4.14
Libs: -L${libdir} -lopen62541 -lpthread -lm -lrt -l/usr/lib/x86_64-linux-gnu/libssl.so -l/usr/lib/x86_64-linux-gnu/libcrypto.so 
Cflags: -I${includedir}
```

So when using this .pc file, it causes a link error because `-l` in front of full paths is not understood/supposed to be that way:
```
/usr/bin/ld: cannot find -l/usr/lib/x86_64-linux-gnu/libssl.so: No such file or directory
/usr/bin/ld: cannot find -l/usr/lib/x86_64-linux-gnu/libcrypto.so: No such file or directory
```

Used CMake options:

```bash
cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DUA_MULTITHREADING=100 -DUA_NAMESPACE_ZERO=REDUCED -DUA_ENABLE_SUBSCRIPTIONS=ON -DUA_ENABLE_ENCRYPTION=OPENSSL -DCMAKE_INSTALL_PREFIX=/my/install/path ..
```

## Checklist

 - [x] open62541 Version (release number or git tag): v1.4.14
 - [x] Operating system: Ubuntu 24.04
 - [x] Critical issue

jpfr:


JohannesKauffmann:
Hi, thanks for the work. I just did a test build, the .pc now looks like:
```
libdir=/my/install/path/open62541/build/install/lib
sharedlibdir=${libdir}
includedir=/my/install/path/open62541/build/install/include

Name: open62541
Description: open62541 is an open source C (C99) implementation of OPC UA
Version: 1.4.14-14-gd7ebaf1d8-dirty
Libs: -L${libdir} -lopen62541 -lpthread -lm -lrt -lOpenSSL::Crypto 
Cflags: -I${includedir}
```
Notice the `-lOpenSSL::Crypto`. The linker doesn't like it:
```
/usr/bin/ld: cannot find -lOpenSSL::Crypto: No such file or directory
```

JohannesKauffmann:
I was thinking, since open62541 doesn't expose its crypto plugin backend library interface in its own public headers (not for openssl, not for mbedtls), can't we simply add the names of the respective library .pc file to open62541.pc? E.g. on my Ubuntu 24 system I have `openssl.pc` and `mbedtls.pc`:
```
Requires.private: openssl
```
```
Requires.private: mbedtls
```
This should pull in the correct dependencies when linking statically, but when simply linking shared libs it shouldn't modify the link flags required to link to open62541 itself.

jpfr:
Just pushed an updated.
Please check on your end, works for me.

JohannesKauffmann:
Hi, yes this does indeed fix the linking issue for OpenSSL.

However, like I mentioned in the issue, the same problem occurs when using MBedTLS. I think that the same fix can apply. I am unable to test with LibreSSL but I suspect that it is also affected.

JohannesKauffmann:
It seems that the same (kind of) patch for LibeSSL does break linking to LibreSSL, at least in the CI (which does static linking). As an alternative, I propose #7588 which does not modify `${open62541_PUBLIC_LIBRARIES}` or `${open62541_LIBRARIES}`, so the existing CMake behaviour isn't modified, only the .pc generation is changed.

jpfr:
Closed via #7588

PR #7585, opened 2025-12-09
Extend the NS0 initialisation logic to include pre-configured namespaces

andreasebner:
This PR extends the server namespace initialisation logic to allow implementations with a preconfigured NS0 to start. The current behaviour for standard server start-up remains unchanged.


jpfr:
Nice!

PR #7588, opened 2025-12-10
fix(build): List private dependencies in .pc file

JohannesKauffmann:
Alternative proposal to #7584. Please let me know if this should be done on `master` or `1.5` first.

---

In the "normal" case of linking to open62541.so, only -lopen62541 is needed in `Libs` because open62541's dependencies are not exposed via its public header files.

All the other dependencies (`Requires.private` and `Libs.private`) only matter using open62541.a (linking statically
via `pkg-config --static`). The crypto libraries and TMP2 library provide .pc files themselves, so we can list them in
`Requires.private`. The other (direct) open62541 dependencies are listed in `Libs.private`.

Also reorder the .pc file (`Libs` comes last and `Requires(.private) comes` before `Cflags`).

Fixes: #7574


jpfr:
Merged, thanks.

PR #7599, opened 2025-12-22
src: client: ua_client_discovery.c: fix avoid double lock in UA_Client_findServersOnNetwork

AntonMoryakov:
In UA_Client_findServersOnNetwork, a second call to lockClient() was erroneously made on the error path after connectSecureChannel() failed. Since the client mutex is already locked at the beginning of the function, this leads to a double-lock attempt on a non-recursive mutex, causing undefined behavior (typically a deadlock).

The erroneous lockClient() call is replaced with unlockClient() to properly release the mutex before returning, consistent with the error handling in UA_Client_getEndpoints and UA_Client_findServers.

Fixes a potential deadlock in multithreaded client usage.


jpfr:
Ouch. Good find.
Please rebase and target the pr for the 1.5 branch.

PR #7600, opened 2025-12-22
src: pubsub: ua_pubsub_ns0.c: add null check after UA_Array_new in addDataSetReaderConfig

AntonMoryakov:
In addDataSetReaderConfig, the result of UA_Array_new was used without checking for NULL when copying DataSetMetaData fields. If memory allocation failed (e.g., due to OOM), the subsequent loop would dereference a null pointer, causing a crash.

Added a check for allocation failure when fieldsSize > 0, consistent with other parts of the codebase that handle UA_Array_new. Returns UA_STATUSCODE_BADOUTOFMEMORY and cleans up partially initialized config.

This improves robustness under low-memory conditions in PubSub subscriber setup.


jpfr:
Very nice.
This should go into the 1.5 branch.
Please rebase and change the pr.

PR #7601, opened 2025-12-26
src: client: ua_client_connect.c: fix prevent potential use of close depollfd in synchronous disconnect

AntonMoryakov:
In UA_Client_disconnectSecureChannel, the synchronous wait loop calls el->run() repeatedly until the secure channel is closed. However, the EventLoop may be stopped (and its epollfd closed) inside one of these run() calls — e.g., due to connection shutdown callbacks. A subsequent el->run() invocation would then operate on a closed file descriptor, triggering warnings from static analyzers and risking undefined behavior in edge cases.

Fix by checking the return status of el->run(). If it returns an error (e.g., BADINTERNALERROR when epollfd < 0), break out of the loop instead of continuing to call run().

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -2483,7 +2483,13 @@ disconnectSecureChannel(UA_Client *client, UA_Boolean sync) {
        el->state != UA_EVENTLOOPSTATE_FRESH &&
        el->state != UA_EVENTLOOPSTATE_STOPPED) {
         while(client->channel.state != UA_SECURECHANNELSTATE_CLOSED) {
-            el->run(el, 100);
+            UA_StatusCode runStatus = el->run(el, 100);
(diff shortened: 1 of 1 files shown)


jpfr:
Very nice.
This should go into the 1.4 branch.
Please rebase and update the PR.
We then pull it up to more recent branches.

AntonMoryakov:
> Very nice. This should go into the 1.4 branch. Please rebase and update the PR. We then pull it up to more recent branches.

I decided to avoid a lengthy rebase since this fix consists of a single, self-contained commit.
Instead, I’ve opened a new PR against the 1.4 branch as requested:
https://github.com/open62541/open62541/pull/7604

This new PR is cleanly based on upstream/1.4 and includes the same fix with the proper Signed-off-by.
Please consider the new PR for merging.

Thank you!

PR #7610, opened 2026-01-02
refactor(tools): Use pylint suggestions to refactor python code

laroche:
Use pylint suggestions to refactor python code.


jpfr:
👍

PR #7613, opened 2026-01-03
refactor(server,test,example): fix clang compiler warnings

laroche:
Fix a few clang compiler warnings from "-Weverything".

### DIFF
--- a/examples/custom_datatype/client_types_custom.c
+++ b/examples/custom_datatype/client_types_custom.c
@@ -7,8 +7,6 @@
 #include <stdlib.h>
 #include <stdio.h>
 
-#define STRING_BUFFER_SIZE 20
-
 int main(void) {
     UA_Client *client = UA_Client_new();
     UA_ClientConfig *cc = UA_Client_getConfig(client);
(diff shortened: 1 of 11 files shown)


jpfr:
Hey @laroche,
the changes look functionally good.
But style-wise I prefer to have a semicolon after macros like CHECK_ATTRIBUTES in code like this:

```c
     case UA_NODECLASS_OBJECT:
          CHECK_ATTRIBUTES(OBJECTATTRIBUTES);
          retval = copyObjectNodeAttributes(&node->objectNode,
                                            (const UA_ObjectAttributes*)attributes);
          break;
```

Also my editor starts to make indenting mistakes if lines don't end with a semicolon.

In the PR #7623 I use to the `do { } while(0)` trick to keep the final semicolon.
Please do the same for the other macros also.
Sometimes this might require a `goto` if the `break`-statement of the enclosing switch-case gets clobbered by the `do-while`...

I will merge #7623 when green, you can build on top.

laroche:
Agreed, this is much better to have a closing/trailing semicolon. I'll remove the parts of this patch that
are now obsolete und push a patch with the remaining changes.

best regards,

Florian La Roche

laroche:
I've added a few similar patches that came up via 'grep -r ";;"'.

best regards,

Florian La Roche

jpfr:
Great! Merged.

PR #7630, opened 2026-01-12
src: pubsub: ua_pubsub_dataset.c: prevent NULL dereference in generateFieldMetaData

AntonMoryakov:
The function UA_findDataTypeWithCustom may return NULL if the DataType referenced in a PublishedDataSet field is not found in the standard or custom type dictionaries. Previously, the code unconditionally accessed currentDataType->typeName (in debug log) and currentDataType->typeKind, which could lead to a crash when using unknown or unregistered data types.

Added explicit NULL checks before all accesses to currentDataType. In debug builds, a helpful message is logged when the type is not found.

This improves robustness against misconfigured or malicious PubSub configurations and aligns with defensive coding practices.


jpfr:
Looks good!
But please make the logging in the failure case a warning.

so problem can be quickly detected and resolved.

PR #7631, opened 2026-01-12
src: server: ua_server_async.c: check return value of addRepeatedCallback in UA_AsyncManager_init

AntonMoryakov:
The function addRepeatedCallback may fail (e.g. due to out-of-memory) and return a non-good StatusCode. Previously, the return value was ignored in UA_AsyncManager_init, which could lead to the timeout cleanup callback (checkTimeouts) never being registered.

As a result:
- Async operations would not be cleaned up on timeout,
- Stale operations could accumulate in queues,
- Memory leaks and performance degradation could occur silently.

Now the return status is checked. On failure, a warning is logged and checkTimeoutCallbackId is set to zero to ensure safe cleanup.

This improves robustness and observability of the async method infrastructure


jpfr:
👍

PR #7632, opened 2026-01-12
src: pubsub: ua_pubsub_dataset.c: prevent NULL dereference in UA_PubSubDataSetField_sampleValue

AntonMoryakov:
The function UA_NODESTORE_GET may return NULL if the referenced node no longer exists (e.g., was deleted after PubSub configuration). Previously, the code unconditionally dereferenced the returned pointer to access rtNode->valueBackend.backend.external.value, which could cause a crash.

Added a NULL check after UA_NODESTORE_GET. If the node is not found, the function returns a DataValue with status UA_STATUSCODE_BADNODEIDUNKNOWN, consistent with OPC UA error handling practices.

This improves robustness of PubSub message generation when source nodes are removed or misconfigured.


jpfr:
Merged, thanks!

PR #7633, opened 2026-01-12
src: pubsub: ua_pubsub_ns0.c: prevent NULL dereference in onReadLocked for PubSubConnection

AntonMoryakov:
The function UA_PubSubConnection_findConnectionbyId may return NULL if the connection no longer exists. In the onReadLocked callback, the result was dereferenced without a NULL check when reading the PublisherId property, which could lead to a crash.

Added a missing NULL check, consistent with the handling of other PubSub entities (DataSetReader, WriterGroup, etc.) in the same function.


jpfr:
Merged, thanks!

PR #7650, opened 2026-01-19
reproduce crash & example fix

dpazj:


jpfr:
Thanks for this.
We'll see if this also applies to 1.3.

jpfr:
Reworked the commit (same author).
Continued in #7667. Thanks for providing this.

PR #7654, opened 2026-01-20
Add enum fields support to the nodeset compiler

basyskom-jvoe:
UaExpert doesn't fall back to the EnumStrings property if the DataTypeDefinition attribute can be read but has no fields. This breaks the method call interface for enum parameters.

This PR adds enum fields support to the nodeset compiler and updates ua_services_attribute.c to return UA_STATUSCODE_BADATTRIBUTEIDINVALID for the DataTypeDefinition attribute if the membersSize of the data type is 0.

### DIFF
--- a/tools/generate_datatypes.py
+++ b/tools/generate_datatypes.py
@@ -249,6 +249,7 @@ class CGenerator:
         before = None
         size = len(datatype.elements) if isEnum else len(datatype.members)
         if isEnum:
+            # Print all enumerators as UA_DataTypeMember
             for i, name in enumerate(datatype.elements):
                 m = "\n{\n"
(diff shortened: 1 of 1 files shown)


basyskom-jvoe:
The nodeset compiler change omits enumerations with `IsOptionSet="true"` for now because these types are generated with a type kind != `UA_DATATYPEKIND_ENUM`.

@jpfr Do you have an idea how the implementation in `ua_types_definition.c` could be updated to support this?

jpfr:
We‘re trying to get our heads around the proposed change.
@basyskom-jvoe Please DM me for a quick phone call.

basyskom-jvoe:
@jpfr I have sent you an e-mail

jpfr:
Nice, thanks

PR #7659, opened 2026-01-21
wipe private key before free in UA_Server_createSigningRequest

AdanRuoff:
This PR zeroes the memory where a created private key was stored during UA_Server_createSigningRequest before freeing the memory. As far as I know, this is best practice with security-sensitive information.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -1070,7 +1070,7 @@ cleanup:
     if(newPrivateKey)
     {
         /* wipe private key before freeing its memory */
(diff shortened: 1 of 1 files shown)


jpfr:
Good idea.
But please use UA_ByteString_memZero.

AdanRuoff:
> Good idea. But please use UA_ByteString_memZero.

Done.

Thanks for the hint btw. - I wasn't aware this function existed until now 😲 .

jpfr:
👍

PR #7660, opened 2026-01-21
fix use-after-free bug in UA_Server_updateCertificate/secureChannel_delayedClose

AdanRuoff:
This PR fixes a use-after-free bug that occurs between "UA_Server_updateCertificate" and "secureChannel_delayedClose":

In the current code, a pointer to the function parameter "certificateTypeId" of the function "UA_Server_updateCertificate" is written into the "certificateTypeId" field of the struct "UpdateCertInfo". Since the parameter has a scope like a local variable, it gets freed as soon as the "UA_Server_updateCertificate" function returns.
Later, in the callback "secureChannel_delayedClose" (i.e. the function that closes the SecureChannel(s) whose Application Instance Certificate changed), this pointer is used for comparing the certificateTypeId stored in a SecurityPolicy struct against it, but since the memory has been freed, this pointer now points to un-allocated memory.

Storing a copy of the NodeId fixes this problem.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -901,7 +901,7 @@ UA_Server_removeCertificates(UA_Server *server,
 
 typedef struct UpdateCertInfo {
     UA_Server *server;
-    const UA_NodeId *certificateTypeId;
+    UA_NodeId certificateTypeId;
 } UpdateCertInfo;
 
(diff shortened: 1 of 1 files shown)


jpfr:
👍

PR #7661, opened 2026-01-21
[Review] fix memory leak in UA_Server_delete when cleaning up "fileInfoContext"s

AdanRuoff:
When calling "UA_Server_delete" while a TrustList is open, the "UA_GDSManager_clear" function loops over all "fileInfoContext" structs and frees their memory. However, the "fileContext" structs stored inside "fileInfoContext" were not freed currently.

As the "fileInfoContext" struct is defined in "ua_server_ns0_gds.c", I had to move the code freeing the internal "fileContext" from "ua_server.c" to "ua_server_ns0_gds.c". Another option would have been to make "fileInfoContext" struct accessible from "ua_server.c", but I think moving the clean-up code is clearer; also, it matches the pattern of "init_*" and "clear_*" functions.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -337,24 +337,6 @@ void UA_GDSTransaction_delete(UA_GDSTransaction *transaction) {
     UA_free(transaction);
 }
 
-/********************/
-/*   GDS Manager    */
-/********************/
-
-void
-UA_GDSManager_clear(UA_GDSManager *gdsManager) {
-    if(!gdsManager)
-        return;
-    gdsManager->checkSessionCallbackId = 0;
-    UA_GDSTransaction_clear(&gdsManager->transaction);
-    void *fileInfoContext = gdsManager->fileInfoContext;
-    while(fileInfoContext) {
-        void *next = *((void **)fileInfoContext);
-        UA_free(fileInfoContext);
-        fileInfoContext = next;
-    }
-}
-
 /*********************/
 /* Server Components */
 /*********************/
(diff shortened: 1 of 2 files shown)


AdanRuoff:
Okay - my bad: I should have done more intensive testing before opening a PR 🤦 .

Compiling currently does not work when UA_ENABLE_GDS_PUSHMANAGEMENT is not defined, as the function declaration in "ua_server_internal.h" will not get included in that case. Seeing that the previous call to "UA_GDSManager_clear" was made unconditional of UA_ENABLE_GDS_PUSHMANAGEMENT, I could simply move the declaration in "ua_server_internal.h" out of the " #ifdef UA_ENABLE_GDS_PUSHMANAGEMENT" block.

A cleaner approach, however, would be to drop _all_ GDS-related code from "ua_server.c" when GDS push management is not enabled.

What is your take on that?

jpfr:
> A cleaner approach, however, would be to drop all GDS-related code from "ua_server.c" when GDS push management is not enabled.

Yes please

AdanRuoff:
In the end, the fix was embarrassingly simple as there are no further references to the GDS Manager in ua_server.c, except for the call to UA_GDSManager_clear() that I removed for the PR.

Enclosing the clean-up call in an "#ifdef UA_ENABLE_GDS_PUSHMANAGEMENT" block fixed the CI for this problem; however, there are still some tests failing with issues that I think are not related to the PR (tests "check_nodeset_loader_compare_di" and "check_csr_generation").

From my side, I think the patch should be good by now.

jpfr:
Lets reduce the indirections.
You can call UA_GDSManager_clear(&server->gdsManager); from server.c.

AdanRuoff:
> Lets reduce the indirections. You can call UA_GDSManager_clear(&server->gdsManager); from server.c.

Okay - done.

Just to share my initial motivation for the indirection: I thought I'd follow the concept of "init" and "clear" functions for features, in case something else besides UA_GDSManager_clear might become necessary during GDS Push cleanup in the future.

andreasebner:
Solved via #7906

PR #7662, opened 2026-01-21
allow removal of CA certificates without valid CRL with UA_Server_removeCertificates

AdanRuoff:
The "UA_Server_removeCertificates" function in "ua_server.c" (which is currently unused, I know 😄 ) calls "getCertificateCrls" when removing a CA certificate to obtain a list of the associated CRLs. However, this function may return "Bad_NoMatch" when either no or no valid CRL exists (e.g. the CRL has expired). In that case, trying to remove a CA certificate via "UA_Server_removeCertificates" will fail.

My assumption however is that _removing_ a certificate should always be possible, even if it is a CA certificate that currently has no valid associated CRL stored in the TrustList. Hence I propose to make "UA_Server_removeCertificates" accept "Bad_NoMatch" from "getCertificateCrls", so CA certificates that do not have an associated valid CRL can still be removed.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -877,7 +877,9 @@ UA_Server_removeCertificates(UA_Server *server,
     UA_StatusCode retval = UA_STATUSCODE_GOOD;
     for(size_t i = 0; i < certificatesSize; i++) {
         retval = certGroup->getCertificateCrls(certGroup, &certificates[i], isTrusted, &crls, &crlsSize);
-        if(retval != UA_STATUSCODE_GOOD) {
+        /* Tolerate "Bad_NoMatch" to support removing CA certificates that do
+         * not have an associated CRL. */
+        if((retval != UA_STATUSCODE_GOOD) && (retval != UA_STATUSCODE_BADNOMATCH)) {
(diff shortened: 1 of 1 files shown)


mcpat-pilz:
@AdanRuoff but it would also make sense to remove all associated CRLs that are not valid anymore too.

AdanRuoff:
> @AdanRuoff but it would also make sense to remove all associated CRLs that are not valid anymore too.

@mcpat-pilz:
I assume by "associated CRLs" you mean the CRLs associated with the certificate to delete. If that is the case, you may rest assured: "getCertificateCrls" returns all CRLs associated with the passed-in certificate, even if they are expired. When getCertificateCrls returns any CRLs (i.e. a non-empty list of CRLs), it will also return UA_STATUSCODE_GOOD, so the CRLs will definitely get removed.

What may have triggered your comment: I just noticed that I write something about "this includes expired CRLs" in the comment above the fix - this is vastly misleading, I will certainly remove it 😲 .

jpfr:
Is this still needed after https://github.com/open62541/open62541/pull/7917?

AdanRuoff:
> Is this still needed after #7917?

No, #7917 solves the issue just fine 😄 . Sorry, I forgot that this PR was still open - will close it.

PR #7663, opened 2026-01-21
[Review] Add size check of uploaded TrustList to "writeTrustList"

AdanRuoff:
The OPC specification states that writing a TrustList that exceeds the maximum configured size should fail with "Bad_RequestTooLarge" (see https://reference.opcfoundation.org/GDS/v105/docs/7.8.2.5 ).

This PR adds a corresponding check at the "Write" stage when updating a TrustList. I am not completely sure if this is the best place to put the check, seeing that the spec only mentions the size check in "CloseAndUpdate"; however, we have all necessary information at hand already during "Write" and I personally see no advantage in postponing this check to the "CloseAndUpdate" stage.


jpfr:
Merged via https://github.com/open62541/open62541/pull/7905

PR #7664, opened 2026-01-21
[Review] Re-use functions from "ua_server.c" in GDS code for adding and removing single certificates

AdanRuoff:
Before version 1.5, the GDS code in "ua_server_ns0_gds.c" used the functions "UA_Server_addCertificates" and "UA_Server_removeCertificates" defined in "ua_server.c" for implementing the "AddCertificate" and "RemoveCertificate" Methods.

In newer versions, the functionality is duplicated between "ua_server.c" (the "UA_Server_addCertificates" and "UA_Server_removeCertificates" functions still exist there) and "ua_server_ns0_gds.c", which provides independent implementations "addCertificate" and "removeCertificate".

This removes the code duplication for adding and removing single certificates between "ua_server.c" and "ua_server_ns0_gds.c". In my opinion, having only a single implementation for a certain task simplifies maintaining the code, and in this special case, it also does away with the creation and immediate execution of a transaction (which is unnecessary in my opinion, as it does not have any advantages over removing the certificate right away). It further allows calling RemoveCertificate while the TrustList is currently open for "Read", something which the OPC specification (if I interpret it correctly) does not explicitly forbid (cf. https://reference.opcfoundation.org/GDS/v105/docs/7.8.2.7 - they only state that "This Method returns Bad_NotWritable if the TrustList Object is read only.", which I think applies to the case when the TrustList is open for "Write").

### DIFF
--- a/src/server/ua_server_ns0_gds.c
+++ b/src/server/ua_server_ns0_gds.c
@@ -468,16 +468,17 @@ addCertificate(UA_Server *server,
     if(fileInfo->openCount > 0)
         return UA_STATUSCODE_BADINVALIDSTATE;
 
-    UA_TrustListDataType trustList;
-    memset(&trustList, 0, sizeof(UA_TrustListDataType));
-    UA_ByteString certificates[1];
-    certificates[0] = certificate;
+    /* use UA_Server_addCertificates() function to do the actual work */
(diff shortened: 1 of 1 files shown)


AdanRuoff:
Superseded by PR #8105.

Closing.

PR #7666, opened 2026-01-22
Fix event filter modification in the client

### LINKED ISSUE #7616: EventFilter modification breaks event notification processing in the client
While porting Qt OPC UA to open62541 v1.5.0-rc2, I encountered a failing test for event filter modification.
After some analysis, this is my understanding of what happens:

If the filter of an existing EventNotifier monitored item is modified by the client, the modification is not applied to the local `UA_Client_MonitoredItem`. This causes the following problems:

- If the new filter has a different number of select clauses, the user callback is not called at all because `processEventNotification()` aborts processing the event due to the continue in the `if(mon->eventFields.mapSize != efl->eventFieldsSize)` block.
- If  the number of select clauses is the same but the fields are not the same as in the initial filter, the event field values will be assigned to a `UA_KeyValuePair` entry with the wrong name.

## Checklist

 - [X] open62541 Version (release number or git tag): v1.5.0-rc2
 - [ ] Operating system:
 - [ ] Critical issue

basyskom-jvoe:
Fixes #7616

### DIFF
--- a/src/client/ua_client_subscriptions.c
+++ b/src/client/ua_client_subscriptions.c
@@ -160,7 +160,7 @@ UA_Client_Subscriptions_create_async(UA_Client *client, const UA_CreateSubscript
     return res;
 }
 
-static UA_Client_Subscription *
+UA_Client_Subscription *
 findSubscriptionById(const UA_Client *client, UA_UInt32 subscriptionId) {
     UA_Client_Subscription *sub = NULL;
     LIST_FOREACH(sub, &client->subscriptions, listEntry) {
@@ -499,6 +499,15 @@ typedef struct {
     UA_ClientAsyncServiceCallback userCallback;
     void *userData;
 } MonitoredItems_CreateData;
+
+typedef struct {
+    /* The request that was made to the server */
+    UA_ModifyMonitoredItemsRequest request;
+
+    /* Notify the user that the async callback was processed */
+    UA_ClientAsyncServiceCallback userCallback;
+    void *userData;
+} MonitoredItems_ModifyData;
 
 static void
 MonitoredItems_CreateData_clear(UA_Client *client, MonitoredItems_CreateData *data) {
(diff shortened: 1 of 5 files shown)


src/client/ua_client_subscriptions.c
```diff
@@ -1048,6 +1057,67 @@ UA_MonitoredItem_change_clientHandle(UA_Client_Subscription *sub,
...
+                                           UA_ModifyMonitoredItemsRequest *request,
+                                           UA_ModifyMonitoredItemsResponse *response) {
+    if(request->itemsToModifySize != response->resultsSize)
+        return;
+
+    for(size_t i = 0; i < request->itemsToModifySize; ++i) {
+        if(response->results[i].statusCode != UA_STATUSCODE_GOOD)
+            continue;
+
+        ZIP_ITER(MonitorItemsTree, &sub->monitoredItems,
```
jpfr: The MonitoredItem has a unique id.
To a search in the ZipTree instead of calling your callback for every MonitoredItem.

src/client/ua_client_subscriptions.c
```diff
@@ -1048,6 +1057,67 @@ UA_MonitoredItem_change_clientHandle(UA_Client_Subscription *sub,
     }
 }
 
+static void *
+UA_MonitoredItem_update_localMonitoredItems_wrapper(void *data, UA_Client_MonitoredItem *mon) {
+    UA_MonitoredItemModifyRequest *monitoredItemModifyRequest =
+        (UA_MonitoredItemModifyRequest *)data;
+    if(monitoredItemModifyRequest &&
+       mon->monitoredItemId == monitoredItemModifyRequest->monitoredItemId) {
+        // Make sure there is an event filter in the modification request
```
jpfr: We prefer C-Style comments /* */ over // C++ comments.

src/client/ua_client_subscriptions.c
```diff
@@ -1088,17 +1160,25 @@ UA_Client_MonitoredItems_modify_async(UA_Client *client,
         return UA_STATUSCODE_BADSUBSCRIPTIONIDINVALID;
     }
 
-    UA_ModifyMonitoredItemsRequest modifiedRequest;
-    UA_ModifyMonitoredItemsRequest_copy(&request, &modifiedRequest);
-    UA_MonitoredItem_change_clientHandle(sub, &modifiedRequest);
+    MonitoredItems_ModifyData *data = (MonitoredItems_ModifyData *)UA_calloc(1, sizeof(MonitoredItems_ModifyData));
```
jpfr, line 1214: Always check if malloc/calloc fails.

src/client/ua_client_subscriptions.c
```diff
@@ -1088,17 +1160,25 @@ UA_Client_MonitoredItems_modify_async(UA_Client *client,
...
+    data->userData = userdata;
+    data->userCallback = (UA_ClientAsyncServiceCallback)callback;
+
+    UA_ModifyMonitoredItemsRequest_copy(&request, &data->request);
```
jpfr: Check the return value

jpfr (commented):
Looks good. Comments inline.
Please add a boolean flag "dirty" to the MonitoredItem structure.

A MonitoredItem is dirty when a ModifyMonitoredItem is currently in flight.

For a dirty MonitoredItem we also don't deliver notifications.
Because we cannot know if they are for the old or for the new configuration.
Instead there should be a warning.

Also we should prohibit two overlapping ModifyMonitoredItem.
So we return an error for ModifyMonitoredItem on a MonitoredItem that is already dirty.

src/client/ua_client_subscriptions.c
basyskom-jvoe: Done

src/client/ua_client_subscriptions.c
basyskom-jvoe: Done

src/client/ua_client_subscriptions.c
basyskom-jvoe: Done

src/client/ua_client_subscriptions.c
basyskom-jvoe: Done

basyskom-jvoe:
@jpfr Thanks for the review.

What return status code would you suggest for attempting to modify a dirty monitored item?
It might be a bit confusing for the user if a modification request with valid values fails.

Would you use this approach for all monitored items, or just EventNotifier monitored items?

And how do you suggest to handle a service call for multiple monitored items if not all are dirty? Fail the entire request? Splicing the offending monitored items out of the request and reassembling a response with a dummy entry at the right index should be avoided IMO.

jpfr:
Decide on a good statuscode.
BadInternalError if nothing fits.

Yes, all MonitoredItems. Not just events.

reject the entire request if one monitireditem is dirty.

basyskom-jvoe:
This will help against event notifications based on the new filter arriving before we have received the modify response.
But depending on the scheduling inside the server, there might still be event notifications in the queue that were created using the old filter and will be delivered after we have updated the event fields map.

The only way I see to avoid any mismatch between local and remote event fields would be to change the client handle with every monitored item modification.

Would you be available for a second call so we can find a solution without any race conditions?

jpfr:
That’s a great idea!
So instead of setting the MI dirty, we can update to a new unique clienthandle.

And of course reject notifications that don’t match.

but we need the option to roll back if the modify request fails.

thjs can be done similar to the „rollover logic“ in the SecureChannel.

if we encounter a ClientHandle in the notification that is unknown, test if matching ClientHandle is queued to replace the config.

basyskom-jvoe:
My current draft just makes the update in the request. The client handle in the local monitored item stays valid until a response with a successful status code arrives. Only then, the client handle in the local monitored item is updated.
This will filter out notifications that arrive befor the response arrives, but until then, all notifications for the old handle will be processed.

The remaining question is if we still need to lock monitored items for modification. Would you deem the following scenario possible?
- We send a modification request and update the handle to x
- We send a second modification request before the response has arrived  and update the handle to y
- The server processes the first request
- The server processes the second request
- The client receives the response for the second request and updates the local handle to y
- The client receives the response for the first request and updates the local handle to x
- The server keeps sending notifications with client handle y while we assume the client handle to be x

jpfr:
The issue is with the application processing the notifications.
It needs to know if it receives data from the new or the old version.

Either we stop processing in the "dirty" state or we provide a mechanism to notify the application when the rollover happens. We can use the existing UA_ClientNotificationCallback mechanism for that.

basyskom-jvoe:
@jpfr The CI seems to be happy now (except random failing tests on LWIP which seem unrelated)

basyskom-jvoe:
@jpfr With #7778 merged, this PR's changes to `ua_client_subscriptions.c` would need a complete rewrite.
Do you still consider the approach valid or is there a better way to do it with your updated architecture?

basyskom-jvoe:
@jpfr @andreasebner Is this issue still present or have there been further changes to the events implementation in the client which took care of it?

jpfr:
Resolved via https://github.com/open62541/open62541/pull/8244

Based on the discussion we had earlier this year.

PR #7677, opened 2026-01-28
fix checking result value in Operation_SetMonitoringMode

AdanRuoff:
In Operation_SetMonitoringMode, the "result" _pointer_ is compared against UA_STATUSCODE_GOOD instead of its value.


jpfr:
👍

PR #7678, opened 2026-01-28
include type casts in UA_StatusCode definitions to improve type safety

AdanRuoff:
This PR adds explicit type casts "(UA_StatusCode)" to the UA_STATUSCODE_* DEFINEs to improve type safety. This may help to spot issues when e.g. accidentally comparing a pointer to a UA_StatusCode with an UA_STATUSCODE_* literal, instead of the value pointed to.

### DIFF
--- a/tools/generate_statuscode_descriptions.py
+++ b/tools/generate_statuscode_descriptions.py
@@ -39,22 +39,22 @@ printh('''/** .. _statuscodes:
  * with the OPC UA standard. */
 
 /* These StatusCodes are manually generated. */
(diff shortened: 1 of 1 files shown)


AdanRuoff:
CI currently fails due to bug addressed in PR #7677 - I will re-run the tests when the bug has been fixed.

AdanRuoff:
> CI currently fails due to bug addressed in PR #7677 - I will re-run the tests when the bug has been fixed.

Looks much better now 😄

PR #7692, opened 2026-02-02
build(core): make zephyr use k_alloc instead of malloc

ge0053:
Adding a macro block to specify that UA_free, UA_malloc, UA_calloc and and UA_realloc is to use the versions provided by zephyr. this makes it so that the default heap is used, meaning that "CONFIG_COMMON_LIBC_MALLOC_ARENA_SIZE" isn't mandatory.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,6 +1,6 @@
 name: Coverity Scan
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   coverity:


jpfr:
You include <zephyr/kernel.h>.
Is that generally available to "userland" applications?

ge0053:
when compiling with west it is available. I am using the same "syntax" as used in the example code for compiling open62541 using zephyr and west([link](https://github.com/open62541/open62541/blob/master/examples/zephyr/server/src/main.c)).

looked a bit on zephyrs documentation and found this note:

"Native Zephyr applications should use the [memory management API](https://docs.zephyrproject.org/latest/kernel/memory_management/index.html#memory-management-api) supported by the Zephyr kernel such as [k_malloc()](https://docs.zephyrproject.org/latest/doxygen/html/group__heap__apis.html#gaa8edf1e63e5d5dd78d7adcfd787394ee) in order to take advantage of the advanced features that they offer.

C standard dynamic memory management interface functions such as [malloc()](https://docs.zephyrproject.org/latest/doxygen/html/stdlib_8h.html#a9c36d0fe3ec4675cbffdc9b52f5fb399) should be used only by the portable applications and libraries that target multiple operating systems.
"
[link](https://docs.zephyrproject.org/latest/develop/languages/c/index.html#dynamic-memory-management)

andreasebner:
I’ve rebased your work onto the latest master branch. After the changes, the Zephyr pipeline is failing. Aside from that, the modifications and the documentation references appear correct.

ge0053:
k_realloc was introduced in zephyr v3.7.0, while the example code uses v2.7.0. to fix this i added a check if zephyr is v3.7.0. or older. 
the LTS version v.2.7.6 is marked as end of life, so it would be applicable to change the version used in the example to the current LTS v3.7.0 or the upcomming LTS v4.6.0 [link](https://docs.zephyrproject.org/latest/releases/index.html#previous-lts).

andreasebner:
The changes look fine now. Since the change is small and limited to a single file, please perform an interactive rebase on your side onto our latest master and create a single commit. Ensure your code does not remove the RBAC lag. Then I will merge it.

jpfr:
You remove the line #cmakedefine UA_ENABLE_RBAC.
Please revert that change. Then we can merge.

jpfr:
@ge0053 bump

PR #7693, opened 2026-02-02
fix(arch): compile error related to zsock_fd_set

ge0053:
when compiling with West version: v1.5.0, zephyr v4.3.0, and gcc (GCC) 15.2.0 ther is an error related to wrong type for UA_fd_set. seems like the type is wrapped twice in a struct.


jpfr:
Thanks. Merged.

PR #7694, opened 2026-02-02
doc(ex): indicate that native_sim may be wrong for some systems

ge0053:
native_sim builds for 32bit systems. for 64bit systems there are some requirements regarding multilib or a 32bit superspace (documented by zephyr)


jpfr:
Thanks, that's a good hint for other users of Zephyr.

PR #7705, opened 2026-02-04
[Review] feat(plugin): add parsing from json client configuration and tests

muesli4:
* Add a feature to parse client configurations from json files.
* Rename loading functions.
* Update documentation.
* Update parameter names.
* Provide tests.
* Remove unused dynamic dispatch in parser.


muesli4:
I did not manage to specify a valid `UserIdentityToken` that is accepted by the test server.

muesli4:
I fixed the previous issues.  The following is a list of what could be done further.

- [x] A JSON file for a client that signs in with a certificate and a test case.
- [x] The JSON file for the example could be extended depending on whether we want a maximal or minimal example.
- [x] Some parts still have missing error code propagation.
- [x] There were discussions about putting certificates and passwords into extra files.  Then there needs to be support in the JSON parser.

examples/client_json_config.c
```diff
@@ -0,0 +1,70 @@
...
+                    "The connection failed with status code %s",
+                    UA_StatusCode_name(retval));
+        UA_Client_delete(client);
+        UA_ByteString_clear(&json_config);
+        return 0;
+    }
+
+    /* Read the value attribute of the node. UA_Client_readValueAttribute is a
+     * wrapper for the raw read service available as UA_Client_Service_read. */
+    UA_Variant value; /* Variants can hold scalar values and arrays of any type */
```
andreasebner: Comment not needed with respect to example scope.

examples/client_json_config.c
```diff
@@ -0,0 +1,70 @@
...
+        /* Load client config */
+        json_config = loadFile(argv[1]);
+    } else {
+        UA_LOG_FATAL(UA_Log_Stdout, UA_LOGCATEGORY_APPLICATION,
+                     "Missing argument. Argument are "
+                     "<client-config.json5>");
+        return EXIT_FAILURE;
+    }
+
+    UA_Client *client = UA_Client_newFromFile(json_config);
```
andreasebner, line 41: You can add a comment for the alternative API to load init a client config with an JSON and then generate a client based on the existing config.

plugins/include/open62541/client_config_file_based.h
```diff
@@ -0,0 +1,31 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information.
```
andreasebner: Integrate all this definitions into the client.h

plugins/include/open62541/server_config_file_based.h
andreasebner: Lets also integrate this into server.h to keep the "surface" in includes small

src/server/ua_services_session.c
```diff
@@ -755,8 +718,8 @@ selectTokenPolicy(UA_Server *server, UA_SecureChannel *channel,
         if(!UA_String_equal(&policyPrefix, &pol->policyId))
             continue;
 
-        /* Get the SecurityPolicy for the endpoint from the postfix */
-        UA_String utPolPostfix = securityPolicyUriPostfix(token->policyId);
+        /* Get the SecurityPolicy */
+        UA_String utPolPostfix = securityPolicyUriPostfix(pol->securityPolicyUri);
```
andreasebner: What is the aim here? We should not integrate a server change into a PR regarding the client config.

tests/client/client_json/client_json_config_username.json5
```diff
@@ -0,0 +1,98 @@
...
+  allowNonePolicyPassword: true,
+
+  // only if encryption enabled
+  maxTrustListSize: 20,
+  maxRejectedListSize: 20,
+
+  namespaces: [],
+
+  outStandingPublishRequests: 0
+}
```
andreasebner, line 105: We should include a Client/File Configuration section in our documentation and provide guidance on the two API functions, along with a complete example of a valid JSON configuration. As a proposal, we could share one version containing only the minimum required fields and another version including all available fields. Finally, it might also be useful to add a JSON Schema for both client and server configuration to the tools folder for validation purposes.

plugins/include/open62541/client_config_file_based.h
muesli4: I assume that you want me to add them in `#ifdef UA_ENABLE_JSON_ENCODING` blocks.  Correct?

src/server/ua_services_session.c
muesli4: I noticed this change is wrong.  I need to add `#None` to the `policyId`.  The `policyId` is something implementation specific.  I wonder how the user is supposed to know it such that it can be specified in the configuration.

tests/client/client_json/client_json_config_username.json5
muesli4: Everything was implemented.

examples/client_json_config.c
```diff
@@ -0,0 +1,82 @@
+/* This work is licensed under a Creative Commons CCZero 1.0 Universal License.
+ * See http://creativecommons.org/publicdomain/zero/1.0/ for more information. */
+
+#include <open62541/client.h>
+#include <open62541/client_highlevel.h>
+#include <open62541/client_config_default.h>
+#include <open62541/plugin/log_stdout.h>
+#include <open62541/client.h>
```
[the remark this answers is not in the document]
muesli4: Removed.

muesli4:
Since I generate certificates during the build, the script requires `python3-netifaces`.  There are 2 choices:
* Check in the certificates, or
* create them during the build (which requires installing the dependency on the runners).

@andreasebner, do you have any preference?

doc/json_config.rst
```diff
@@ -0,0 +1,452 @@
...
+- There are no trailing commas where they shouldn't be (though JSON5 allows them in most places)
+- The file encoding is UTF-8
+
+**Q: The client/server starts but doesn't use my configuration**
+
+**A:** Ensure that:
+
+- The configuration file is being loaded successfully (check return values)
+- The ``UA_ByteString`` contains the actual file contents
+- You're not overriding configuration options after loading the file
```
andreasebner: Shorten the documentation to make the information more concise. It’s obviously inflated by AI.

tools/validate_configs.py
```diff
@@ -0,0 +1,133 @@
...
+
+import json
+import json5
+import sys
+from pathlib import Path
+from jsonschema import Draft7Validator
+
+def load_json5(filepath):
+    """Load a JSON5 file and convert to Python dict."""
+    with open(filepath, 'r') as f:
```
andreasebner: If this file isn’t directly used in our CI, let’s remove it. With the schema in place, anyone can validate it, and this file looks very AI‑unfiltered.

doc/json_config.rst
```diff
@@ -0,0 +1,212 @@
+.. _json_config:
+
+JSON Configuration
```
jpfr: The added-value of this file is not enough to bloat the documentation structure.
Rather move the explanation to the client.h and server.h where the json config api is defined.

Instead of showing example json5 inside a markdown, beef up the json5 files in the example folder.
Use all available options there and put explanations and alternatives with comments.
The json5 in the examples folder is a much better documentation than the json schema.
We need discoverability of the features for user!

Refer to the json5 in the examples folder from the server.h and client.h.

muesli4:
Yesterday, I changed the `examples/client_json_config.c` to fail properly when the config is missing.  This exposed another missing setup command in a CI job about examples.  I fixed it and hopefully it should be fine now.

tests/client/check_client_json_config.c
```diff
@@ -0,0 +1,579 @@
...
+
+#include "../common.h"
+#include "test_helpers.h"
+#include "thread_wrapper.h"
+
+#include <check.h>
+
+#if defined(_MSC_VER)
+# pragma warning(disable: 4146)
+#endif
```
andreasebner, line 23: What is it? Check if needed, if yes, please document the reason.

PR #7706, opened 2026-02-05
refactor(ci): change order on removing/adding build dir

laroche:
Change order on removing/adding build dir.


andreasebner:
Thanks for sharing this change. Could you explain the motivation for switching from “deleting all files inside the existing directory” to “removing and recreating the directory”?
I assume the goal was to ensure hidden files are removed as well. In which situation did the previous behavior cause issues?

laroche:
Correct, I want to make sure hidden files are removed as well. I haven't seen any real problems with the current code.

best regards,

Florian La Roche

PR #7714, opened 2026-02-09
plugins: ua_log_syslog.c: remove unreachable code in syslog logger

AntonMoryakov:
The condition `if(logLevelSlot < 0 || logLevelSlot > 5)` is never true because:
- UA_LOGLEVEL_TRACE (which would yield logLevelSlot = -1) is handled in the switch statement and causes an early return.
- All other valid log levels produce logLevelSlot values in [0, 4].

Thus, the fallback assignment `logLevelSlot = 5` is dead code and can be safely removed without changing runtime behavior.


jpfr:
👍 Commit title adjusted

PR #7716, opened 2026-02-09
src: pubsub: ua_pubsub_ns0.c: prevent null pointer dereference in addDataSetReaderLocked

AntonMoryakov:
The function UA_ReaderGroup_findRGbyId may return NULL if the ReaderGroup does not exist. Previously, the returned pointer was dereferenced without a NULL check when accessing rg->configurationFrozen, which could cause a segmentation fault on invalid or malicious method calls.

Added an explicit NULL check and return UA_STATUSCODE_BADNODEIDUNKNOWN with proper logging using UA_NODEID_PRINT for safety and diagnostics.

This aligns with error handling patterns used elsewhere in the PubSub information model (e.g., addDataSetWriterLocked).

### DIFF
--- a/src/pubsub/ua_pubsub_ns0.c
+++ b/src/pubsub/ua_pubsub_ns0.c
@@ -950,6 +950,12 @@ addDataSetReaderLocked(UA_Server *server,
     UA_LOCK_ASSERT(&server->serviceMutex, 1);
     UA_StatusCode retVal = UA_STATUSCODE_GOOD;
     UA_ReaderGroup *rg = UA_ReaderGroup_findRGbyId(server, *objectId);
+    if(!rg) {
+        UA_LOG_ERROR(server->config.logging, UA_LOGCATEGORY_SERVER,
+                     "ReaderGroup not found for NodeId %S",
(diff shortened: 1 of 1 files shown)


src/pubsub/ua_pubsub_ns0.c
```diff
@@ -950,6 +950,12 @@ addDataSetReaderLocked(UA_Server *server,
     UA_LOCK_ASSERT(&server->serviceMutex, 1);
     UA_StatusCode retVal = UA_STATUSCODE_GOOD;
     UA_ReaderGroup *rg = UA_ReaderGroup_findRGbyId(server, *objectId);
+    if(!rg) {
+        UA_LOG_ERROR(server->config.logging, UA_LOGCATEGORY_SERVER,
+                     "ReaderGroup not found for NodeId %S",
```
jpfr, line 955: The uppercase %S fomat string is available on 1.5/master, but not yet for 1.4.
UA_NODEID_PRINT is not defined for 1.4 either from what I see.

jpfr (commented):
Good find.

jpfr:
@AntonMoryakov are you on this?

PR #7717, opened 2026-02-09
src: server: ua_server.c: prevent null pointer dereference in UA_Server_run_startup

AntonMoryakov:
The function getServerComponentByName may return NULL if the "binary" component was not successfully created (e.g., due to memory allocation failure). Previously, the returned pointer was dereferenced without a NULL check, leading to a potential segmentation fault during server startup.

Added explicit NULL check and proper error handling with logging and cleanup.


jpfr:
👍 Commit title adjusted

PR #7718, opened 2026-02-09
src: pubsub: ua_pubsub_writergroup.c: prevent null pointer dereference in UA_WriterGroup_publishCallback

AntonMoryakov:
When writerGroup->linkedConnection is NULL, the publish callback previously called UA_WriterGroup_setPubSubState, which internally dereferences writerGroup->linkedConnection (e.g., in UA_PubSubConnection_setPubSubState). This leads to a segmentation fault if the connection has been removed or was never set.

Fixed by directly setting writerGroup->state = UA_PUBSUBSTATE_ERROR instead of invoking UA_WriterGroup_setPubSubState when the connection is missing. This avoids unsafe pointer dereference while preserving correct error state and logging.


jpfr:
Setting the state variable directly is not enough.
We need to drop into the state machine function to
- Clean up the internal state (e.g. timers)
- Notify the application
- Cascade the state transition into the writer's state machine

PR #7719, opened 2026-02-09
plugins: crypto: openssl: securitypolicy_openssl_common.c: remove unreachable code in UA_Openssl_RSA_Private_Decrypt

AntonMoryakov:
The second NULL check for 'privateKey' is redundant because the first condition already covers it:

    if (data == NULL || privateKey == NULL)
        return ...;

Thus, the subsequent 'if (privateKey == NULL)' block is unreachable.


jpfr:
👍 
Commit title adjusted

PR #7720, opened 2026-02-09
arch: eventloop_posix_udp.c: check return value of getsockname in UDP_registerListenSocket

AntonMoryakov:
The return value of getsockname() was not checked when retrieving the dynamically assigned port after binding a UDP socket. If getsockname() fails (e.g., due to EBADF or memory issues), the port variable would be set from uninitialized memory, leading to undefined behavior.

Added proper error checking and logging using UA_LOG_SOCKET_ERRNO_WRAP. Also improved robustness by using sockaddr_storage to support both IPv4 and IPv6 address families.


jpfr:
👍

PR #7721, opened 2026-02-09
src: pubsub: ua_pubsub_networkmessage.c: initialize 'rv' in UA_NetworkMessage_decodePayload to avoid undefined behavior

AntonMoryakov:
The variable 'rv' was declared but not initialized. In the case where dst->payloadHeaderEnabled is true and count == 0, no assignment to 'rv' occurs before UA_CHECK_STATUS(rv, ...), leading to reading uninitialized stack data.

Fixed by initializing 'rv' to UA_STATUSCODE_GOOD and explicitly handling the count == 0 case. Also added error checking inside the multi-message loop.


jpfr:
Thanks, merged.
I had to adjust the commit title to our conventions though.

PR #7726, opened 2026-02-11
refactor(plugins,pubsub,tests): cleanup "gcc -Wshadow -Wconversion"

laroche:
Cleanup compiler warnings from "gcc -Wshadow -Wconversion".


jpfr:
👍

PR #7755, opened 2026-02-20
FileSystem Implementierung

LRausch01:
Implementierung eines FileSystems


LRausch01:
@andreasebner

andreasebner:
Please ensure that the code from SKYredSKY is rewritten and that his commits are removed from the history due to the missing CLA. After this, the CI should also run without requiring my approval.

LRausch01:
For some Reson it doesnt automatically run the ci @andreasebner.

PR #7756, opened 2026-02-20
[Review]: fix fetching common name [CN=] from cert

Akul123:
- When there is something after common name [CN=] in subject string it will be added to common name and possibly used for certificate fileName during in getCertFileName function.
If i.e. email is added after CN= in subject string content of common name is common name + email which results in bad internal error.
- This commit introduces fetching common name directly from certitificate, similar to fetching thumbprint, instead of using pointer arithemetic on subject string.
- New test case which uses certificate with email after common name in subject string is added to confirm this behaviour.


Akul123:
Can anyone take a look at this pull req please?

PR #7762, opened 2026-02-22
plugins: crypto: fix UNCHECKED_FUNC_RES.STAT in ua_securitypolicy_none.c

AntonMoryakov:
The functions UA_OpenSSL_LoadLocalCertificate and
UA_mbedTLS_LoadLocalCertificate return a UA_StatusCode that was previously ignored in the SecurityPolicy#None implementation. This could lead to undefined behavior if certificate loading failed (e.g., due to invalid format, parsing errors, or memory allocation failure), while the policy continued initialization with a partially initialized or empty certificate.

Now, the return value is explicitly checked for both OpenSSL/LibreSSL and MbedTLS backends. On failure:
  • The partially initialized certificate is cleared via policy_clear_none()
  • The error status is propagated upward to the caller
  • The policy is left in a safe, clean state

This change brings UA_SecurityPolicy_None in line with other security policies (e.g., Basic256, Aes128Sha256, Basic128Rsa15) which already perform this check, ensuring consistent error handling across all crypto backends.

### DIFF
--- a/plugins/crypto/ua_securitypolicy_none.c
+++ b/plugins/crypto/ua_securitypolicy_none.c
@@ -127,16 +127,29 @@ policy_clear_none(UA_SecurityPolicy *policy) {
 UA_StatusCode
 UA_SecurityPolicy_None(UA_SecurityPolicy *policy, const UA_ByteString localCertificate,
                        const UA_Logger *logger) {
+
+    UA_StatusCode retval = UA_STATUSCODE_GOOD;
+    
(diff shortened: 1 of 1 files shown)


jpfr:
The change is good.
But there is a surprising number of CI runs which fails now.

jpfr:
@AntonMoryakov would you take a look at the failing CI tests?

andreasebner:
Merged in #8160

PR #7763, opened 2026-02-22
fix(core): UA_fileExists for zephyr

ge0053:
zephyr does not provide access(). since we only
 want to know if the file exists it is adequate
to use fs_stat().
the missing access() is only a problem iff UA_ENABLE_DISCOVERY and
UA_ENABLE_DISCOVERY_SEMAPHORE is active.
not properly tested.


jpfr:
@ge0053 what is your response to the comments for this PR?

PR #7766, opened 2026-02-23
Return Bad Stauts Code for empty non-nullable variables

andreasebner:
The OPC UA specification Parts 4 and 6 define non‑nullable data types and describe how services should behave when a value is NULL even though the type is non‑nullable. Currently, it is possible that we return a Good status code together with an empty value for a non‑nullable type. This PR adds checks for nullable types and updates the attribute service to return an appropriate status code.


jpfr:
👍

Issue #7691, opened 2026-02-02
Source File Duplication in Amalgamation in v1.5.0


ogultan:
I have been using v1.5.0-rc2 previously and had no issues with amalgamation process. It seems with the new release of 1.5.0 there is a bug or oversight in cmake.

Previously arch/ contents have been amalgamated like this:
```
if(UA_ARCHITECTURE_POSIX OR UA_ARCHITECTURE_WIN32)
    list(APPEND plugin_sources
         ${PROJECT_SOURCE_DIR}/arch/common/timer.h
         ${PROJECT_SOURCE_DIR}/arch/common/timer.c
         ${PROJECT_SOURCE_DIR}/arch/posix/clock_posix.c
         ${PROJECT_SOURCE_DIR}/arch/common/eventloop_common.h
         ${PROJECT_SOURCE_DIR}/arch/common/eventloop_common.c
         ${PROJECT_SOURCE_DIR}/arch/posix/eventloop_posix.h
         ${PROJECT_SOURCE_DIR}/arch/posix/eventloop_posix.c
         ${PROJECT_SOURCE_DIR}/arch/posix/eventloop_posix_tcp.c
         ${PROJECT_SOURCE_DIR}/arch/posix/eventloop_posix_udp.c
         ${PROJECT_SOURCE_DIR}/arch/posix/eventloop_posix_eth.c
         ${PROJECT_SOURCE_DIR}/arch/posix/eventloop_posix_interrupt.c)
endif()

if(UA_ARCHITECTURE_ZEPHYR)
    list(APPEND plugin_sources
         ${PROJECT_SOURCE_DIR}/arch/common/timer.h
         ${PROJECT_SOURCE_DIR}/arch/common/timer.c
         ${PROJECT_SOURCE_DIR}/arch/zephyr/clock_zephyr.c
         ${PROJECT_SOURCE_DIR}/arch/common/eventloop_common.h
         ${PROJECT_SOURCE_DIR}/arch/common/eventloop_common.c
         ${PROJECT_SOURCE_DIR}/arch/zephyr/eventloop_zephyr.h
         ${PROJECT_SOURCE_DIR}/arch/zephyr/eventloop_zephyr.c
         ${PROJECT_SOURCE_DIR}/arch/zephyr/eventloop_zephyr_tcp.c)
endif()

if(UA_ARCHITECTURE_LWIP)
    list(APPEND plugin_sources
            ${PROJECT_SOURCE_DIR}/arch/common/timer.h
[...]

PR #7767, opened 2026-02-23
Fix amalgamation build without selected architecture

basyskom-jvoe:
Currently, the clock functions for posix and win32 are only included into the amalgamation if one of the two architectures is selected.

This extends bc6e26b 

The last time I did a similar modification, it broke amalgamation for one of the more exotic architectures (see #7691). I think this won't happen this time because the concerned file has an architecture based ifdef, but I don't have a way to test this.

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -70,6 +70,9 @@ jobs:
           - name: "Amalgamation Build with Multithreading"
             cmd_action: build_amalgamation_mt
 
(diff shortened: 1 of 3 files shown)


basyskom-jvoe:
@andreasebner Could you please have a look at the modifications to make sure I didn't recreate the issue that was fixed by bc6e26b?

basyskom-jvoe:
Fixed via #7965

PR #7769, opened 2026-02-23
fix(client): Fix authSecurityPolicies usage for username identity token

basyskom-jvoe:
If the token is not an X.509 identity token, the securityPolicies array in the client config was searched instead of authSecurityPolicies.

### DIFF
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -1779,6 +1779,23 @@ UA_ClientConfig_setDefault(UA_ClientConfig *config) {
         config->securityPoliciesSize = 1;
     }
 
+    /* TODO: This is only necessary if we don't have special cases for authentication tokens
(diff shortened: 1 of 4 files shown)


jpfr:
We will review this.
In the meantime, please some more comments at the top of the new "code paragraphs".

basyskom-jvoe:
@jpfr The additional paragraph in `ua_config_default.c` is only necessary because there is no special handling for anonymous tokens/None policy in the security policy lookup in `ua_client_connect.c`.

Would this be an option or are there too many other code paths that rely on having a valid security policy pointer?

basyskom-jvoe:
This is basically the same scenario as in #7322
We had special handling for Anonymous back then, but I'm not sure if this is still applicable due to architectural changes.

andreasebner:
Rework made in #7920 and merged.

PR #7776, opened 2026-02-23
fix(server): Handle null-pointer case in processRequest.

r-sin:
Handle the case where the session pointer is NULL.


jpfr:
There are many unintended commits in here.
Please rebase.

r-sin:
Ok, re-opened in #7839.

PR #7779, opened 2026-02-25
[Review] Conditionally generalize certificate failures

muesli4:
Fix 2 CTT tests by generalizing the return value.

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -854,6 +854,16 @@ checkActivateSessionX509(UA_Server *server, UA_Session *session,
                               session->channel, session, "ActivateSession",
                               NULL, token->certificateData);
 
+    /* Expected result for CTT */
+    if(res == UA_STATUSCODE_BADCERTIFICATEUNTRUSTED ||
+       res == UA_STATUSCODE_BADCERTIFICATETIMEINVALID ||
+       res == UA_STATUSCODE_BADCERTIFICATEREVOKED ||
+       res == UA_STATUSCODE_BADCERTIFICATEUSENOTALLOWED ||
+       res == UA_STATUSCODE_BADCERTIFICATEISSUERUSENOTALLOWED ||
+       res == UA_STATUSCODE_BADCERTIFICATECHAININCOMPLETE) {
+        res = UA_STATUSCODE_BADIDENTITYTOKENREJECTED;
+    }
+
  out:
     /* Delete the temporary channel context */
     sp->deleteChannelContext(sp, tempChannelContext);


src/server/ua_server_utils.c
```diff
@@ -416,6 +417,44 @@ validateCertificate(UA_Server *server, UA_CertificateGroup *cg,
...
+                res = UA_STATUSCODE_BADINTERNALERROR;
+            }
+        }
+        /* This is required for compliance with CTT (Security/Security Certificate Validation/002.js). */
+        else if(res == UA_STATUSCODE_BADCERTIFICATEISSUERREVOCATIONUNKNOWN) {
+            UA_TrustListDataType trustList;
+            UA_StatusCode trustListRes = cg->getTrustList(cg, &trustList);
+
+            if(trustListRes == UA_STATUSCODE_GOOD) {
+                if (UA_TrustListDataType_contains(&trustList, &certificate, UA_TRUSTLISTMASKS_TRUSTEDCERTIFICATES)) {
```
AdanRuoff, line 445: Thank you for making the return codes of the certificate verification specification-conforming 👍- it took me quite some time figuring out that the "trusted" state of the certificate must be considered for returning the proper status code (see e.g. tests 007.js vs. 033.js).

Just a quick question: Shouldn't this check be made against the "list of _issuer_ certificates", i.e.
```
if (!UA_TrustListDataType_contains(&trustList, &certificate, UA_TRUSTLISTMASKS_ISSUERCERTIFICATES)) {
...
```
since test 002.js targets certificates that are "not trusted but known" (i.e. certs. in the "list of issuer certificates|CRLs")?

muesli4: > Thank you for making the return codes of the certificate verification specification-conforming 👍- it took me quite some time figuring out that the "trusted" state of the certificate must be considered for returning the proper status code (see e.g. tests 007.js vs. 033.js).

I have to give @andreasebner credit for the idea.

> Just a quick question: Shouldn't this check be made against the "list of _issuer_ certificates", i.e.
> 
> ```
> if (!UA_TrustListDataType_contains(&trustList, &certificate, UA_TRUSTLISTMASKS_ISSUERCERTIFICATES)) {
> ...
> ```

Good that you brought it up.  With my limited knowledge in certificates, it was very difficult for me to translate the textual description into a meaningful condition.  In the end I looked at the name of the certificate `Settings.Advanced.Certificates.ApplicationInstanceCertificates.ctt_ca1TC_ca2I_appT` and the scheme from the UaCtt documentation.  The last certificate in the chain is trusted, which is the certificate that is verified here.  Please let me know if there is a misunderstanding on my side.

So far I did not have any test failures there, which might simply indicate that the test coverage is insufficient.  It might very well be that my condition is not sufficient at all.

AdanRuoff: Oops - you are of course quite right - it was _me_ thinking in the wrong direction: It's the last certificate in the chain that matters. Sorry for causing confusion!

Still, I wonder if the check for "TrustList contains leaf certificate" is correct here: If the TrustList _does_ contain the leaf (i.e. Application Instance) certificate, it is considered "trusted" and hence the more specific status code (here: Bad_CertificateIssuerRevocationUnknown) should be returned, should it not? If so, I think the condition here should be the same as above where you handle "Bad_CertificateTimeInvalid", i.e. "TrustList does _not_ contain leaf certificate":
````
if (!UA_TrustListDataType_contains(&trustList, &certificate, UA_TRUSTLISTMASKS_TRUSTEDCERTIFICATES)) {
...
````

Comment (muesli4) [reply src/server/ua_server_utils.c:445]: > If the TrustList does contain the leaf (i.e. Application Instance) certificate, it is considered "trusted" and hence the more specific status code (here: Bad_CertificateIssuerRevocationUnknown) should be returned, should it not?

I thought so too in the beginning, but the generalization should happen in this specific case.  It might be also necessary to generalize it if the condition does not apply.  But there is no corresponding test yet.  I debugged the test case in UaCtt and I had GDB run into this condition.  I checked the trust list there and basically this was the only list it was in, i.e. `UA_TRUSTLISTMASK_TRUSTEDCERTIFICATES`.

What I do not know yet:
1. It was not found in all other trust lists.  Can it be in other trust lists at the same time?  It does not make sense to me.
2. Is it sufficient to just check the leaf?  Is the return code from the validation sufficient for the condition?

I was not able to find anything in the specification.

File: src/server/ua_server_utils.c [hunk 2]
```diff
@@ -416,6 +417,43 @@ validateCertificate(UA_Server *server, UA_CertificateGroup *cg,
                          "in the CertificateGroup with StatusCode %s",
                          logPrefix, descr);
         }
+
+        /* Conform to the specification by generalizing errors */
+
+        if(res == UA_STATUSCODE_BADCERTIFICATETIMEINVALID) {
```
jpfr, line 423: It doesn't seem right to have special case code for every different status code.
That is very brittle, e.g. when new statuscodes are introduced.
And there is a lot of duplication.

Why don't we move the UA_TrustListDataType_contains test **inside** cg->verifyCertificate?

It would be good for cg->verifyCertificate to return the detailed statuscodes.
And then we replace that with a generic BADSECURITYCHECKSFAILED (or similar) within this `validateCertificate`.

The reason is that we are working on the audit events feature.
And the audit events need the detailed statuscode.

As a nitpick, keep to the coding style of the repository.
You are using too much vertical space with the braces on a newline.

src/server/ua_server_utils.c
AdanRuoff: > I thought so too in the beginning, but the generalization should happen in this specific case. It might be also necessary to generalize it if the condition does not apply. But there is no corresponding test yet. I debugged the test case in UaCtt and I had GDB run into this condition.

I don't argue about this specific case - it is just the condition of the "if" I am wondering about. You get the detailed status code from `cg->verifyCertificate` (line 401) and want to replace it with the generic "Bad_SecurityChecksFailed" when the certificate to be checked is not trusted, correct?

This is in line with how I read the OPC specification: The steps to determine if a certificate is considered "trusted" (defined here: https://reference.opcfoundation.org/Core/Part4/v105/docs/6.1.3) are to be executed in the given order, i.e. first "Certificate Structure", then "Build Certificate Chain" and so on. If any check fails, the process of certificate validation is exited prematurely.

In every step until and including "Trust List Check", the specification explicitly states that "If this check fails on the Server side, the error Bad_SecurityChecksFailed shall be reported back to the Client.". As I see it, this means that any detailed status code returned by the following steps:
1. Certificate Structure
2. Build Certificate Chain
3. Signature	
4. Security Policy Check
5. Trust List Check

should be masked with "Bad_SecurityChecksFailed".

When reading further, only the descriptions of the steps "Find Revocation List" and "Revocation Check" state that "Bad_SecurityChecksFailed should be reported back to the Client." I interpret this in the way such that the detailed status codes of the steps:
7. Validity Period
8. Host Name
9. URI
10. Certificate Usage
should **not** get masked.

However, if you perfom certificate validation as you currently do (i.e. performing all defined steps without exiting prematurely - we do it the same way in our homebrewn certificate validation implemention btw. :smile:), you will have to replace the detailed error codes generated in these steps with the generic "Bad_SecurityChecksFailed" when the certificate is not trusted.

In summary, this means the detailed status codes of the following verification steps should get masked if (and only if) the certificate is considered "not trusted":
- Validity Period:
    * Bad_CertificateTimeInvalid
    * Bad_CertificateIssuerTimeInvalid
- Host Name (only done on Client side):
    * Bad_CertificateHostNameInvalid
- URI:
    * Bad_CertificateUriInvalid
- Certificate Usage:
    * Bad_CertificateUseNotAllowed
    * Bad_CertificateIssuerUseNotAllowed

Your code tries to achieve that for the two status codes "Bad_CertificateTimeInvalid" (lines 423ff), and "Bad_CertificateIssuerRevocationUnknown" (lines 440ff). However, in my understanding, the condition in line 445 is wrong: You need to mask the detailed status code when the certificate is **not** trusted, but you check for:
```
if (UA_TrustListDataType_contains(&trustList, &certificate, UA_TRUSTLISTMASKS_TRUSTEDCERTIFICATES)) 
```
Note the absence of any negation operator - if I understand the code correctly, you are checking here if the certificate is trusted, meaning you replace the detailed status code with the generic "Bad_SecurityChecksFailed" when the certificate is trusted, and keep the detailed code when the certificate is not trusted. This is why I think that the condition should be the same here as in the case of "Bad_CertificateTimeInvalid", i.e. it should be:
```
if (!UA_TrustListDataType_contains(&trustList, &certificate, UA_TRUSTLISTMASKS_TRUSTEDCERTIFICATES)) 
```
Note the negation operator before `UA_TrustListDataType_contains()`.

Please excuse my very verbose explanation...

AdanRuoff: > What I do not know yet:
>
> 1. It was not found in all other trust lists. Can it be in other trust lists at the same time? It does not make sense to me.

I assume by "trust lists" you mean the sub-lists ("list of trusted certificates", "list of trusted CRLs",  "list of issuer certificates" and "list of issuer CRLs"), correct? While (at least to the best of my knowledge :smile: ) the specification does not explicitly rule out the case that a certificate is present in both the "list of trusted certificates" and the "list of issuer certificates" at the same time, it would not make much sense from my perspective, as the "list of trusted certificates" is also considered when building up a "chain of trust".

In our certificate verification process, we do it the following way:
- To verify the "chain of trust" of a Client Application Instance Certificate, we merge the "list of trusted certificates" and the "list of issuer certificates" and search this combined list. Having a certificate in both lists would just yield a duplicate entry in the merged list and make no difference when validating the "chain of trust".
- In a second step, we check if we "trust" a certificate by traversing the "chain of trust" constructed in the previous step backwards. If any of the certificates is present in the "list of trusted certificates", we consider the Client certificate "trusted".

As to your statement "It was not found in all other trust lists": Which CTT test case are you referring to?
- In "002.js", the Client uses `ctt_ca1TC_ca2I_appT` as its Application Instance Certificate (AIC). This is a certificate signed by the CA `ctt_ca1TC_ca2I_appT`. The AIC is present in the Server's "list of trusted certificates" (file `copyToServer/ApplicationInstance_PKI/trusted/certs/ctt_ca1TC_ca2I_appT.der` in CTT), its issuer is present in the Server's "list of issuer certificates" (file `copyToServer/ApplicationInstance_PKI/issuers/certs/ctt_ca1TC_ca2I.der`). Hence, the AIC is considered "trusted".
-  In "007.js", the Client uses the self-signed certificate `ctt_appTE` as its AIC; this certificate is present in the Server's "list of trusted certificates" (file `copyToServer/ApplicationInstance_PKI/trusted/certs/ctt_appTE.der`). Hence, the AIC is considered "trusted".
-  In "033.js", the Client uses the self-signed certificate `ctt_appUE` as its AIC; this certificate is *not* present in any sub-list of the Server's TrustList, as there is no file `ctt_appUE.der` in any of `copyToServer`'s subdirectories. Hence, the AIC is *not* trusted.


> 2. Is it sufficient to just check the leaf? Is the return code from the validation sufficient for the condition?

As I tried to explain in the previous comment, I think it is unfortunately *not* sufficient to check the leaf only 😢 : When you put a CA certificate in the "list of trusted certificates", it basically means that you (recursively!) trust *any* certificate that was signed by this CA. So it is possible that the Client AIC that you are trying to verify is not even present in any of the TrustList's sub-lists - if it's issuing CA is present in the "list of trusted certificates", you trust any certificate of that CA, thus also the particular Client AIC at hand.

I think you will have to pass the "trust" state of the AIC-to-be-verified out from `cg->verifyCertificate` in some way, so you can mask the return codes stated in the previous comment when "isTrusted == false".

This whole certificate validation process is just so damn complicated 😵

src/server/ua_server_utils.c
AdanRuoff: I agree with the "duplication issue"; however, as to having special cases for each different status code: I remember that during your GDS implementation for us, you came up with a pretty similar piece of code. IIRC, it was a C function for masking return values that used a "switch/case" statement with fall-throughs to avoid code duplication, i.e. something in the way:
```
switch (res) {
    UA_STATUSCODE_BADCERTIFICATETIMEINVALID:
    UA_STATUSCODE_BADCERTIFICATEISSUERTIMEINVALID:
    UA_STATUSCODE_BADCERTIFICATEHOSTNAMEINVALID:
    UA_STATUSCODE_BADCERTIFICATEURIINVALID:
    UA_STATUSCODE_BADCERTIFICATEUSENOTALLOWED:
    UA_STATUSCODE_BADCERTIFICATEISSUERUSENOTALLOWED:
        if (!isTrusted) {
            res = UA_STATUSCODE_BADSECURITYCHECKSFAILED;
        }
        break;
    default:
        break;
}
```

Frankly, I personally wouldn't know how to implement it another way, but I'm open to learn :smile:.

BTW: Moving the masking into `cg->verifyCertificate` sounds like a good idea to me :smile:.

muesli4: > It doesn't seem right to have special case code for every different status code. That is very brittle, e.g. when new statuscodes are introduced. And there is a lot of duplication.

The inner condition is different.  ~That is why it is not a duplication.~  I can merge the conditions if you want.  I would love to implement it in a general manner but it seems the specification does not mention this behavior.

> Why don't we move the UA_TrustListDataType_contains test **inside** cg->verifyCertificate?

I initially wanted to, but:
1. there are different certificate groups, which means I have to duplicate code, and
2. the server might want to report the specific error code but still return the generic one.

> It would be good for cg->verifyCertificate to return the detailed statuscodes. And then we replace that with a generic BADSECURITYCHECKSFAILED (or similar) within this `validateCertificate`.
> 
> The reason is that we are working on the audit events feature. And the audit events need the detailed statuscode.

How would this be different from this PR?

> As a nitpick, keep to the coding style of the repository. You are using too much vertical space with the braces on a newline.

I am happy to comply.  But I am afraid I don't understand what you want me to do.

muesli4:
I took all of the feedback under consideration and discussed the approach with @jpfr.  Also, thank you @AdanRuoff for your valuable feedback.  I did read every part of it.  The result is https://github.com/open62541/open62541/pull/7902.

AdanRuoff:
> I did read every part of it. The result is #7902.

Thank you for bearing with me 😄 .

I had a look at #7902 and I must say that I wouldn't have come up with the idea of keeping a struct with both StatusCodes (an internal one and a "masked" one as return value for the Client). IMHO that is a quite creative solution - I like it. This way, any code calling "verifyCertificate" has access to the detailed StatusCode (e.g. for logging), but the SecureChannel connection logic does not have to care about masking out possibly sensitive StatusCodes.

I understand that this will not find its way into the current 1.5 release because of the API changes, but I look forward to seeing this in a future version 👍 .

PR #7780, opened 2026-02-25
fix(plugin): Enable parsing Alternative Subject Names for mbedtls.

laroche:
mbedtls_x509_subject_alternative_name is no define, so use MBEDTLS_X509_CRT_PARSE_C and MBEDTLS_X509_CSR_PARSE_C to enable this.

### DIFF
--- a/plugins/crypto/mbedtls/certificategroup.c
+++ b/plugins/crypto/mbedtls/certificategroup.c
@@ -694,7 +694,7 @@ cleanup:
     return retval;
 }
 
-#if !defined(mbedtls_x509_subject_alternative_name)
+#if MBEDTLS_VERSION_NUMBER < 0x03040000
 
 static const unsigned char *
 UA_Bstrstr(const unsigned char *s1, size_t l1, const unsigned char *s2, size_t l2) {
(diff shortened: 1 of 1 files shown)


laroche:
Seems to work with very light testing. Not sure these defines are the best to look at.
Is there a reason to keep this out per default and assume a special enable-define?

best regards,

Florian La Roche

laroche:
One more data point: Seems with openssl this checking of Alternative Names is always compiled in.

AdanRuoff:
> mbedtls_x509_subject_alternative_name is no define, so use MBEDTLS_X509_CRT_PARSE_C and MBEDTLS_X509_CSR_PARSE_C to enable this.

Correct, `mbedtls_x509_subject_alternative_name` is no define. My guess is that the original developer intended to check for the availability of the `mbedtls_x509_subject_alternative_name` struct, seeing as this only got introduced into MbedTLS v3.4.0. Prior to that version, MbedTLS had virtually no support for SANs at all.

Instead of looking for the defines MBEDTLS_X509_CRT_PARSE_C / MBEDTLS_X509_CSR_PARSE_C, I think the cleanest solution would be to check for an MbedTLS version >= 3.4.0 here.

I can look into how to do this as soon as I find the time (i.e. probably never 😄 )...

laroche:
Yes, seems all older Ubuntu releases fail this build. I've only tested on a development Debian-14 machine
and there this change works ok.

With your suggestion maybe looking at MBEDTLS_VERSION_NUMBER should be possible:
"MBEDTLS_VERSION_NUMBER >= 0x03040000"

best regards,

Florian La Roche

AdanRuoff:
> With your suggestion maybe looking at MBEDTLS_VERSION_NUMBER should be possible: "MBEDTLS_VERSION_NUMBER >= 0x03040000"

Yep - that was the "define" I was looking for 😄 . Looks good to me now - thanks ❤️ .

andreasebner:
Thanks for contributing. The change looks good to me. The check in the old preprocessor always leads to the fallback implementation.

PR #7781, opened 2026-02-25
[Review] fix(server): permit untrusted certificate in CreateSession

muesli4:
If the channel has no security, the server should ignore the certificate in CreateSession.  This means it will skip certificate validation and will also avoid including the certificate in the response.

I assumed `channel->securityMode == UA_MESSAGESECURITYMODE_NONE` follows from the _none_ security policy.  Please tell me whether this assumption is false.

The CTT test `Session Services/Session Base/012.js` now passes.

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -468,7 +468,8 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
      * checked for the SecureChannel. But here we use the Session
      * CertificateGroup and we now also have the ApplicationDescription to check
      * the ApplicationUri. */
-    if(request->clientCertificate.length > 0) {
+    if(channel->securityMode != UA_MESSAGESECURITYMODE_NONE &&
+       request->clientCertificate.length > 0) {
         rh->serviceResult =
             validateCertificate(server, &server->config.secureChannelPKI,
                                 channel, NULL, "CreateSession",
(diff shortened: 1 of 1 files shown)


src/server/ua_services_session.c
```diff
@@ -468,7 +468,8 @@ Service_CreateSession(UA_Server *server, UA_SecureChannel *channel,
      * checked for the SecureChannel. But here we use the Session
      * CertificateGroup and we now also have the ApplicationDescription to check
      * the ApplicationUri. */
-    if(request->clientCertificate.length > 0) {
+    if(channel->securityMode != UA_MESSAGESECURITYMODE_NONE &&
```
muesli4, line 471: This may accept more than just untrusted certificates, including some error conditions.

muesli4: In [OpenSecureChannel](https://reference.opcfoundation.org/Core/Part4/v104/docs/5.5.2.2):
> If the securityPolicyUri is None, the Server shall ignore the ApplicationInstanceCertificate. 

In [CreateSession](https://reference.opcfoundation.org/Core/Part4/v104/docs/5.6.2.2):
> If the securityPolicyUri is None, the Server shall ignore the ApplicationInstanceCertificate.

Therefore I would argue this behavior is desired.

muesli4:
A similar change was introduced in e20bac48751ed62dd4f516d6779d08d27c4978c8.  However, it misses the change about copying of the certificate.  I will have to test again whether it works in CTT.  I believe it needs both.

muesli4:
I tested the current repository state and it works in UaCtt even without the second change.

PR #7798, opened 2026-03-02
feat(client): Add readDatatypeDefinitionAttribute convenience function.

r-sin:
I don’t know the reason why the convenience functions stop at attribute ID 22 and whether this one should be added. This PR adds another one that we need for a use case.


jpfr:
Nice 👍

PR #7803, opened 2026-03-03
style(plugins): Remove duplicated code path

vil02:
The original code could have a bug or a confusing _typo_. I hope it is the second case.


jpfr:
Merged, thanks.

PR #7812, opened 2026-03-05
client/server: guard method call APIs with UA_ENABLE_METHODCALLS

StalderT:
### Problem

With `UA_ENABLE_METHODCALLS=OFF`, some public headers still expose Call/Method
APIs and callbacks referencing `UA_CallResponse`, `UA_CallMethodRequest` and
`UA_CallMethodResult`. In amalgamation builds this leads to compilation errors
(e.g. `UA_CallResponse` / `UA_CallMethodResult` unknown type).

### Fix

Wrap the affected declarations (UA_Client_Service_call, UA_Client_call_async,
and server async method callback/API) in `#ifdef UA_ENABLE_METHODCALLS` so
they are only visible when method calls are enabled.

### DIFF
--- a/include/open62541/client.h
+++ b/include/open62541/client.h
@@ -349,8 +349,12 @@ UA_Client_Service_historyUpdate(UA_Client *client,
  * Methods represent the function calls of Objects. The Method Service Set
  * defines the means to invoke Methods. */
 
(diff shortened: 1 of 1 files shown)


jpfr:
I‘d rather remove the build flag and always enable methods.
The binary size difference is negligible.

PR #7815, opened 2026-03-05
fix(client): copy type into typeString to avoid it being uninitialized on read

### LINKED ISSUE #7800: Read from uninitialized variable in `ua-cli`
The variable `typeString` is read uninitialized in lines [299](https://github.com/open62541/open62541/blob/89020b5e31ba1d337db80b55a220cabe7b883f27/tools/ua-cli/ua.c#L299) and [310](https://github.com/open62541/open62541/blob/89020b5e31ba1d337db80b55a220cabe7b883f27/tools/ua-cli/ua.c#L310):

https://github.com/open62541/open62541/blob/89020b5e31ba1d337db80b55a220cabe7b883f27/tools/ua-cli/ua.c#L287-L321

I suppose the content of `type` should be copied to `typeString`.

o6-nbeier:
Fixes #7800.

### DIFF
--- a/tools/ua-cli/ua.c
+++ b/tools/ua-cli/ua.c
@@ -286,14 +286,12 @@ parseVariant(UA_Variant *v, UA_String valstr) {
     const UA_DataType *datatype = &UA_TYPES[UA_TYPES_VARIANT];
     if(valstr.data[0] == '(') {
         char typeString[512];
-        UA_STACKARRAY(char, type, valstr.length);
-        int elem = sscanf((char*)valstr.data, "(%511[^)])", type);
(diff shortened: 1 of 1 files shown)


jpfr:
I don’t get it.
The sscanf could directly go into the typString buffer.
Do we even need the indirection?

o6-nbeier:
I think this grew upon `type` being a VLA and `typeString` being introduced as a FLA replacement which is safer from a security perspective. But `typeString` did never actually replace `type` since it was never initialized correctly but both were used nevertheless.
My suggestion would be to correctly replace `type` with `typeString`.

PR #7816, opened 2026-03-05
fix(core): fix QualifiedName parsing when name contains a semicolon

### LINKED ISSUE #7801: JSON deserialization of QualifiedName fails when name contains semicolon
## Summary
JSON serialization/deserialization of `UA_QualifiedName` fails when the name contains a semicolon (`;`). After decoding, the namespace index is lost and set to 0, and the decoded name includes the namespace prefix.

## Environment
- **Branch:** 1.5
- **Commit:** [97a4ee6](https://github.com/open62541/open62541/tree/97a4ee6c1c687ba1909ca59369e2b73b1c938d07)
- **OS:** Linux x64
- **Compiler:** gcc 13

## Steps to Reproduce

```
UA_QualifiedName value;
UA_QualifiedName_init(&value);
value.name = UA_STRING_ALLOC("te;st");
value.namespaceIndex = 123;
printf("value: ns=%u, name=\"%.*s\"\n", (unsigned)value.namespaceIndex, (int)value.name.length, (const char*)value.name.data);

// ---------------- encode ---------------- 
UA_ByteString encoded;
UA_ByteString_init(&encoded);
const UA_StatusCode encStatus =
    UA_encodeJson(&value, &UA_TYPES[UA_TYPES_QUALIFIEDNAME], &encoded, NULL);
ck_assert_uint_eq(encStatus, UA_STATUSCODE_GOOD);
printf("encoded: \"%.*s\"\n", (int)encoded.length, (const char*)encoded.data); UA_QualifiedName decoded;

// ---------------- decode ---------------- 
UA_QualifiedName_init(&decoded);
const UA_StatusCode decStatus =
    UA_decodeJson(&encoded, &decoded, &UA_TYPES[UA_TYPES_QUALIFIEDNAME], NULL);
ck_assert_uint_eq(decStatus, UA_STATUSCODE_GOOD);
printf("decoded: ns=%u, name=\"%.*s\"\n", (unsigned)decoded.namespaceIndex, (int)decoded.name.length, (const char*)decoded.name.data);

ck_assert_uint_eq(value.namespaceIndex, decoded.namespaceIndex);
ck_assert_uint_eq(value.name.length, decoded.name.length);
ck_assert_mem_eq(value.name.data, decoded.name.data, value.name.length);

UA_ByteString_clear(&encoded);
UA_QualifiedName_clear(&value);
UA_QualifiedName_clear(&decoded);
```

## Expected Behavior
- Decoded namespace index: 123
- Decoded name: `"te;st"`

## Actual Behavior

Console output:
```
value: ns=123, name="te;st"
encoded: ""123:te;st""
decoded: ns=0, name="123:te;st"
0%: Checks: 1, Failures: 1, Errors: 0
... Assertion 'value.namespaceIndex == decoded.namespaceIndex' failed: value.namespaceIndex == 123, decoded.namespaceIndex == 0
```

- Decoded namespace index: **0** (should be 123)
- Decoded name: **`"123:te;st"` (should be `"te;st"`)** - includes namespace prefix

## Potential Root Cause
The JSON deserialization logic may need investigation around [src/util/ua_types_lex.re](https://github.com/open62541/open62541/blob/97a4ee6c1c687ba1909ca59369e2b73b1c938d07/src/util/ua_types_lex.re#L323), where the semicolon is handled as a delimiter during parsing.

## Observations
- Removing the semicolon makes round-trip serialization/deserialization work as expected (value == decoded)
- My analysis found that [ua_types_lex.re@323](https://github.com/open62541/open62541/blob/97a4ee6c1c687ba1909ca59369e2b73b1c938d07/src/util/ua_types_lex.re#L323) deals with `;` as a delimiter, which may be a good start for further investigations.

## Questions for Maintainers
1. Can you confirm this behavior with the provided reproduction code?
2. Is this considered a bug (i.e., should `QualifiedName` strings with semicolons round-trip correctly through JSON serialization/deserialization)?
3. If semicolons are not allowed in `QualifiedName` values, should the encoder reject such inputs or escape them?

o6-nbeier:
The re2c grammar for parse_qn defined the namespace-URI alternative as [^;\000]*";" which matches any characters up to the first ";". Due to re2c's longest-match rule this pattern would beat the numeric namespace index rule [0-9]+":". For example, "123:te;st" was tokenized as URI="123:te" followed by name="st" instead of nsIndex=123, name="te;st".

Fix by restricting the URI alternative to inputs that start with a non-Digit character. All valid OPC UA namespace URIs (e.g. "http://..." or "urn:...") begin with a letter, so this does not affect legitimate namespace URIs.

Fixes #7801.

### DIFF
--- a/src/util/ua_types_lex.c
+++ b/src/util/ua_types_lex.c
@@ -712,10 +712,10 @@ parse_qn(UA_QualifiedName *qn, const u8 *pos, const u8 *end,
     
 {
 	u8 yych;
-	unsigned int yyaccept = 0;
 	yych = YYPEEK();
 	switch (yych) {
-		case 0x00: goto yy41;
+		case 0x00:
+		case ';': goto yy41;
 		case '0':
 		case '1':
 		case '2':
@@ -726,7 +726,6 @@ parse_qn(UA_QualifiedName *qn, const u8 *pos, const u8 *end,
 		case '7':
 		case '8':
 		case '9': goto yy44;
-		case ';': goto yy45;
 		default: goto yy43;
 	}
 yy41:
@@ -734,19 +733,16 @@ yy41:
 yy42:
 	{ pos = begin; goto match_name; }
 yy43:
-	yyaccept = 0;
 	YYSKIP();
 	YYBACKUP();
 	yych = YYPEEK();
 	if (yych <= 0x00) goto yy42;
-	goto yy47;
+	goto yy46;
 yy44:
-	yyaccept = 0;
 	YYSKIP();
 	YYBACKUP();
 	yych = YYPEEK();
 	switch (yych) {
-		case 0x00: goto yy42;
 		case '0':
 		case '1':
 		case '2':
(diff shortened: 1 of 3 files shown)


jpfr:
Nice.
Please add a test case to the parsing unit test…

PR #7818, opened 2026-03-06
refactor(core): Simplify logic of `UA_StatusCode_isUncertain`

vil02:
For every `x`: if `x == 0x01`, then `x < 0x02`.

I checked the related [docs](https://reference.opcfoundation.org/Core/Part4/v104/docs/7.34): the _uncertain_ status is represented by a value `01`, so the original code is just a _confusing typo_.

### DIFF
--- a/include/open62541/types.h
+++ b/include/open62541/types.h
@@ -182,7 +182,7 @@ UA_StatusCode_name(UA_StatusCode code);
 UA_EXPORT UA_Boolean
 UA_StatusCode_isBad(UA_StatusCode code);
 
-/* ((code >> 30) == 0x01) && ((code >> 30) < 0x02) */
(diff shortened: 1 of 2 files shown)


jpfr:
Good find.
Please make this a PR against the 1.5 branch. You can change the target of the existing PR.

vil02:
> Please make this a PR against the 1.5 branch. You can change the target of the existing PR.

Doing it created a mess. I recreated by branch on top of 1.5 and cherry-picked my commit.

jpfr:
> Doing it created a mess. I recreated by branch on top of 1.5 and cherry-picked my commit.

Yes, that is standard practice.

jpfr:
Nice, thanks

PR #7822, opened 2026-03-06
docs(ex): specify codeblock as sh

ge0053:
specify that the code block is of type .sh. this is to have color formatting.


jpfr:
Thanks for the improvement

PR #7823, opened 2026-03-06
fix: make sure tens[] is optimized into read only array

ge0053:
it does not properly optimized into a read only array. most likely it is since it is accessed as a pointer. 
the problem is present for  Zephyr version: 4.3.0 using west.


jpfr:
This broke a lot of unit tests.
Please check.

PR #7826, opened 2026-03-06
refactor(server): avoid undefined behaviour with memcpy()

laroche:
Sometimes memcpy() is called with the second parameter set to NULL. This only happens in cases where the length parameter 3 is also set to 0. So add checks for the third parameter before calling memcpy() to avoid undefined behaviour.


andreasebner:
Solved via #7919

PR #7827, opened 2026-03-06
fix(server) Resolve issues detected during coverage improvements

andreasebner:
During the latest push of large unit test sets, a few minor issues surfaced. This PR resolves those issues and adds additional tests to prevent future regressions.

1. parseDouble read-beyond-bounds: parseDouble() copied input to a null-terminated local buffer but passed the original (potentially unterminated) pointer to strtod(), causing an out-of-bounds read when the input was not null-terminated.
2. UA_KeyValueMap_setScalarShallow crash on free: The variant stored a pointer to caller-owned data without setting storageType = UA_VARIANT_DATA_NODELETE, so UA_Variant_clear() would later call free() on a stack variable.
3. UA_base64 invalid pointer on empty input: UA_base64() returned UA_EMPTY_ARRAY_SENTINEL (0x1) for zero-length input; callers that passed the result to UA_free() would crash because 0x1 is not a valid heap pointer.

### DIFF
--- a/src/util/ua_util.c
+++ b/src/util/ua_util.c
@@ -315,7 +315,7 @@ UA_StatusCode
 UA_ByteString_toBase64(const UA_ByteString *byteString,
                        UA_String *str) {
     UA_String_init(str);
-    if(!byteString || !byteString->data)
+    if(!byteString || !byteString->data || byteString->length == 0)
         return UA_STATUSCODE_GOOD;
 
     str->data = (UA_Byte*)
@@ -479,6 +479,7 @@ UA_KeyValueMap_setShallow(UA_KeyValueMap *map,
     }
 
     *target = *value;
+    target->storageType = UA_VARIANT_DATA_NODELETE;
     return UA_STATUSCODE_GOOD;
 }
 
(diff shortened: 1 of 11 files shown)


tests/check_xml_encoding_roundtrip.c
```diff
@@ -234,10 +234,11 @@ START_TEST(xml_bytestring) {
...
 
 START_TEST(xml_bytestring_empty) {
-    /* Uses BYTESTRING_NULL here. A non-NULL data pointer with length=0
-     * triggers BUG: UA_base64() returns UA_EMPTY_ARRAY_SENTINEL (0x1)
-     * which crashes on UA_free(). See COVERAGE_BUGS.md #8. */
-    UA_ByteString src = UA_BYTESTRING_NULL;
+    /* With the fix for COVERAGE_BUGS #8, UA_base64() returns NULL for
+     * zero-length input instead of UA_EMPTY_ARRAY_SENTINEL. A non-NULL
+     * data pointer with length=0 no longer crashes. */
+    UA_Byte dummy = 0;
```
jpfr, line 239: This was fixed at the root cause inside encodeXml.
The workaround in the tests can be removed-

src/util/ua_util.c
```diff
@@ -507,6 +507,7 @@ UA_KeyValueMap_setScalarShallow(UA_KeyValueMap *map, const UA_QualifiedName key,
     v.type = type;
     v.arrayLength = 0;
     v.data = p;
+    v.storageType = UA_VARIANT_DATA_NODELETE;
```
jpfr: This is not needed and/or should be handled inside setShallow

PR #7829, opened 2026-03-07
win32: fix case-sensitive includes for mingw cross-compilation

StalderT:
When cross-compiling on Linux with mingw-w64, the Windows headers are installed with lowercase filenames (e.g. wtypes.h, winbase.h).

The current includes use mixed case (WTypes.h, WinBase.h) which works on Windows but fails on case-sensitive filesystems during mingw cross-compilation.

This patch changes the includes to lowercase to match the actual header names and ensure successful cross-compilation.

### DIFF
--- a/src/util/ua_util.c
+++ b/src/util/ua_util.c
@@ -25,8 +25,8 @@
 #include "itoa.h"
 
 #if defined(UA_ARCHITECTURE_WIN32)
-#include <WTypes.h>
-#include <WinBase.h>
+#include <wtypes.h>
+#include <winbase.h>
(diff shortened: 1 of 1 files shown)


jpfr:
Thanks for this

StalderT:
Thanks for merging the fix into master.

Would it be possible to also backport this change to the 1.5 branch so it can be included in future 1.5.x releases?

Thanks!

jpfr:
Please open a pr for that

StalderT:
done via #7874

PR #7832, opened 2026-03-07
posix: avoid double fclose in writeByteStringToFile

StalderT:
writeByteStringToFile closed the file twice when UA_fwrite returned a short write:

- once inside the error path
- once again before returning

This can trigger a use-after-free warning with GCC when building with -Werror. Only close the file once before returning.


jpfr:
Nice

PR #6750, opened 2024-10-03
Support custom datatypes namespaces

segotdko:
Called by ua_generate_datatypes() Python script does not recognise datatypes defined in any namespace.
CSV file definitions of type myVar,"ns=1;s=myStr",DataType cannot be used.

PR #7834, opened 2026-03-07
fix(core): Correct WRITEMASK typo and support custom datatype namespaces

andreasebner:
Based on PR #6750 by @segotdko, with extensions.


jpfr:
This is good for 1.5 also?

andreasebner:
> This is good for 1.5 also?

It depend's:
-> The recent version with

` "#define UA_WRITEMASK_ARRRAYDIMENSIONS        UA_WRITEMASK_ARRAYDIMENSIONS /* legacy typo alias */" 
 `

could also be merged into 1.5 since we are not breaking the API. If we prefer to remove this compatibility define (which I would prefer for master), then we need to separate the behavior between versions or make the change master‑only. What’s your opinion on that?

jpfr:
Let’s add the api compatible changes on 1.5 also.

PR #7835, opened 2026-03-07
style(core): Add missing standard headers

vil02:
One should not rely on transitive includes.


jpfr:
We do rely on the headers coming from config.h.
So there is a single place to handle the system includes.
Which often rely on macro definitions before, for example to unlock newer Unix features.

PR #7841, opened 2026-03-09
openssl: fix unused return value from BIO_reset with -Werror

StalderT:
Follow-up to #7831

When building open62541 with -Werror, OpenSSL's BIO_reset macro produces a warning because the return value is not used:

value computed is not used [-Wunused-value]

Cast the call to (void) to mark the return value as intentionally unused and avoid the warning.


jpfr:
Thanks

Issue #7842, opened 2026-03-09
Dangling pointer warning in notifySubscription with GCC -Wdangling-pointer


StalderT:
https://github.com/open62541/open62541/blob/26a10956e9c56077cfbe750614c7cd8ada68db0e/src/server/ua_services_subscription.c#L91

# Bug: Dangling pointer warning in `notifySubscription` with GCC -Wdangling-pointer

## Description

When compiling **open62541** with recent GCC versions and strict warning
settings (`-Wall -Wextra -Werror`), the build fails due to a dangling
pointer warning in `notifySubscription()`.

The compiler reports that the address of a stack variable is stored
inside a `UA_Variant`, which may outlive the function scope.

Example compiler error:

    error: storing the address of local variable 'enabled' in 'createSubData[7].value.data' [-Werror=dangling-pointer=]

## Code location

Function:

    notifySubscription()

Relevant code:

``` c
UA_NodeId sessionId = (sub->session) ? sub->session->sessionId : UA_NODEID_NULL;
UA_Boolean enabled = (sub->state = UA_SUBSCRIPTIONSTATE_ENABLED);

UA_Variant_setScalar(&createSubData[0].value, &sessionId,
                     &UA_TYPES[UA_TYPES_NODEID]);

UA_Variant_setScalar(&createSubData[7].value, &enabled,
                     &UA_TYPES[UA_TYPES_BOOLEAN]);
```

Both `sessionId` and `enabled` are local variables whose addresses are
passed to `UA_Variant_setScalar()`.

## Compiler output

Example output:

    In function 'UA_Variant_setScalar',
    inlined from 'notifySubscription':
    error: storing the address of local variable 'enabled'
    in 'createSubData[7].value.data' [-Werror=dangling-pointer]

    note: 'enabled' declared here
    note: 'createSubData' declared here

## Environment

-   OS: Linux
-   Compiler: GCC 13 / GCC 14
-   Build flags: `-Wall -Wextra -Werror`
-   open62541 version: 1.5

## Additional context

The build fails when `-Werror` is enabled because the compiler detects a
potential dangling pointer when storing the address of a stack variable
in a `UA_Variant`.

PR #7847, opened 2026-03-09
fix(server)  #7842 : fix potential dangling pointer in subscription notification data and add test

StalderT:
## Additionnal question

must be :
    enabled = (sub->state = UA_SUBSCRIPTIONSTATE_ENABLED);
or
    enabled = (sub->state == UA_SUBSCRIPTIONSTATE_ENABLED);

## Description

Fix a build failure with recent GCC versions when compiling with
`-Werror`.

The function `notifySubscription()` stores the address of a stack
variable (`enabled`) in a `UA_Variant` via `UA_Variant_setScalar()`.
Newer GCC versions detect this as a potential dangling pointer
(`-Wdangling-pointer`).

To avoid storing a pointer to a stack variable, the value is now taken
directly from the subscription object instead of using a temporary local
variable.

## Example compiler error

    error: storing the address of local variable 'enabled' in
    'createSubData[7].value.data' [-Werror=dangling-pointer]

## Changes

-   Remove the temporary stack variable `enabled`
-   Use the subscription state directly when populating the `UA_Variant`

### Before

``` c
UA_Boolean enabled = (sub->state = UA_SUBSCRIPTIONSTATE_ENABLED);

UA_Variant_setScalar(&createSubData[7].value, &enabled,
                     &UA_TYPES[UA_TYPES_BOOLEAN]);
```

### After

``` c
sub->enabled = (sub->state == UA_SUBSCRIPTIONSTATE_ENABLED);

UA_Variant_setScalar(&createSubData[7].value, &sub->enabled,
                     &UA_TYPES[UA_TYPES_BOOLEAN]);
```

## Environment

-   OS: Linux
-   Compiler: GCC 13 / GCC 14
-   Build flags: `-Wall -Wextra -Werror`

### DIFF
--- a/src/server/ua_services_subscription.c
+++ b/src/server/ua_services_subscription.c
@@ -71,8 +71,10 @@ notifySubscription(UA_Server *server, UA_Subscription *sub,
     };
     UA_KeyValueMap createSubMap = {8, createSubData};
 
-    UA_NodeId sessionId = (sub->session) ? sub->session->sessionId : UA_NODEID_NULL;
-    UA_Boolean enabled = (sub->state = UA_SUBSCRIPTIONSTATE_ENABLED);
+    static UA_THREAD_LOCAL UA_NodeId sessionId;
+    sessionId = (sub->session) ? sub->session->sessionId : UA_NODEID_NULL;
+    static UA_THREAD_LOCAL UA_Boolean enabled;
+    enabled = (sub->state == UA_SUBSCRIPTIONSTATE_ENABLED);
 
     UA_Variant_setScalar(&createSubData[0].value, &sessionId,
                          &UA_TYPES[UA_TYPES_NODEID]);
--- a/tests/server/check_services_subscriptions.c
+++ b/tests/server/check_services_subscriptions.c
@@ -89,6 +89,50 @@ static void teardown(void) {
 
 static UA_UInt32 subscriptionId;
 static UA_UInt32 monitoredItemId;
+static UA_Boolean subscriptionNotificationCalled = false;
+static UA_ApplicationNotificationType subscriptionNotificationType;
+static UA_UInt32 subscriptionNotificationId = 0;
+static UA_Boolean subscriptionNotificationEnabled = false;
+static size_t subscriptionNotificationMapSize = 0;
+
+static const UA_Variant *
+findNotificationValue(const UA_KeyValueMap *map, const char *key) {
+    if(!map || !map->map)
+        return NULL;
+
+    size_t keylen = strlen(key);
+    for(size_t i = 0; i < map->mapSize; i++) {
+        const UA_String *name = &map->map[i].key.name;
+        if(name->length == keylen &&
+           memcmp(name->data, key, keylen) == 0)
+            return &map->map[i].value;
+    }
+
+    return NULL;
+}
+
+static void
+testSubscriptionNotificationCallback(UA_Server *server,
+                                     UA_ApplicationNotificationType type,
+                                     UA_KeyValueMap data) {
+    (void)server;
+
+    subscriptionNotificationCalled = true;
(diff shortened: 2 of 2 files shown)


jpfr:
Uh. Good find.
The enabled variable is only used locally.
Please remove from the structure, it can be inferred from the state.

StalderT:
Thanks for the review.

While looking into this I noticed that `createSubData` is declared as `static UA_THREAD_LOCAL`. This means the `UA_Variant` entries inside the map store pointers to external data whose lifetime must outlive the function call.

If callbacks are strictly synchronous and must not retain the map or any of its values, this approach works.

However, if callbacks are allowed to retain the map or access it later, the pointer-backed variants could become invalid (since they reference fields from `sub`, `session`, or temporary values).

Would it make more sense to allocate the `UA_KeyValuePair` array locally for the duration of the notification and free it  fterwards, instead of using a thread-local static buffer?

That would avoid lifetime issues and make ownership clearer, at the cost of a small allocation per notification.

Happy to adjust the patch depending on the intended lifetime model.

StalderT:
## Summary

Fix a potential dangling pointer in `notifySubscription()` and add a unit test
verifying that the subscription notification callback is invoked with valid
data.

The previous implementation used a `static UA_THREAD_LOCAL` array for the
`UA_KeyValuePair` storage used to build the notification metadata map.

This caused the map storage to outlive the function scope while some variant
values could reference temporary stack variables. In particular, GCC recently
started warning about storing the address of a stack variable in the map
(`-Wdangling-pointer`).

The patch removes the `static UA_THREAD_LOCAL` storage and instead allocates
the `UA_KeyValuePair` array locally in the function. This aligns the lifetime
of the map storage with the function scope and avoids the possibility of
storing pointers to stack variables in longer-lived memory.

Since the notification callback is expected to consume the `UA_KeyValueMap`
synchronously, a local array is sufficient and simplifies the lifetime model.

## Changes

- remove `static UA_THREAD_LOCAL` from the notification metadata storage
- allocate the `UA_KeyValuePair` array locally in `notifySubscription()`
- add a unit test in `tests/server/check_services_subscriptions.c`
  verifying that the notification callback is invoked and that the expected
  metadata entries are present

## Test

A new test (`Server_createSubscription_notificationCallback`) was added to
`check_services_subscriptions.c`.

The test:

- installs a subscription notification callback
- creates a subscription
- verifies that the callback is called
- verifies that metadata entries such as:
  - `subscription-id`
  - `publishing-enabled`
  are present and contain valid values.

The test follows the existing Check-based test structure used in the file.

## Notes

The notification metadata contains pointers to server-owned data (e.g.
subscription fields), so callbacks are expected to consume the map
synchronously and not retain the pointers after the callback returns.

This patch keeps that contract while removing the lifetime mismatch between
temporary stack variables and the static thread-local metadata buffer.

jpfr:
Please re-add the thread-local tagging.
We want this code path to be maximally efficient.
From the initialization just below, we know that all dangling pointers are reset at the next iteration.

StalderT:
If we keep the thread-local caching here, recent GCC versions complain with `-Werror` enabled.

With GCC 14 for example, this triggers `-Wdangling-pointer` because `UA_Variant_setScalar()` stores the address of a local variable (`enabled`) in the thread-local `createSubData` array:

    error: storing the address of local variable ‘enabled’ in ‘createSubData[7].value.data’ [-Werror=dangling-pointer=]

So even if this was previously accepted, it now breaks builds with newer compilers and strict warning settings.

StalderT:
Makes sense.

I can keep the thread-local caching and avoid the GCC 14 `-Wdangling-pointer` warning by moving the temporary values referenced by the variants (`sessionId` and `enabled`) to thread-local storage as well.

That keeps the fast path unchanged while avoiding pointers to stack locals.

PR #7856, opened 2026-03-11
fix(client): prevent infinite loop in UA_ClientConfig_clear

### LINKED ISSUE #7771: after unsuccessful UA_Client_getEndpoints(): 100% CPU during UA_ClientConfig_clear
Sometimes after interrupt, my UA client process hangs with 100% CPU usage. `perf top` shows me:

Image

It seems that it's a busy loop in `UA_ClientConfig_clear()` that also tries to `UA_LOG_ERROR`. 

It's a Ruby binding, where a `struct ClientData` wraps the native client. Ruby calls `ClientData_free()` when the corresponding `Open62541::Client` object is garbage collected. Unfortunately, calling `UA_Client_free()` will result in logging. To avoid logging via a potentially GC'd Ruby logger object (segfault) or creating Ruby objects during GC phase (not allowed), my `ClientData_free()` first unsets `clientConfig.logging.log` before actually freeing the native UA client:

```c
static void ClientData_free(void* data)
{
  ClientData *clientData = (ClientData *)data;

  if (clientData->endpointDescriptions) {
    UA_free(clientData->endpointDescriptions);
    clientData->endpointDescriptions = NULL;
    clientData->endpointDescriptionsSize = 0;
  }

  if (clientData->client) {
    UA_ClientConfig *config = UA_Client_getConfig(clientData->client);

    // NOTE: avoid segfault from logging via RbLog during UA_Client_delete()
    if (config->logging && config->logging->log == RbLog) {
      config->logging->log = NULL;
      config->logging->context = NULL;
    }

    UA_Client_delete(clientData->client);
    clientData->client = NULL;
  }

  free(clientData);
}
```

Looking at the implementation of `UA_LOG_ERROR`, it should not matter whether I unset the `clientConfig.logging` or just `clientConfig.logging.log`:
https://github.com/open62541/open62541/blob/master/include/open62541/plugin/log.h#L132-L135

Is it possible that `UA_ClientConfig.logging` and/or `UA_Logger.log` should be **marked `volatile`**, so that my removal of the logger is actually respected?

Used CMake options:

```sh
/usr/bin/cmake -S /home/user/dev/oss/open62541-ruby/ext/open62541/deps/open62541 -B /home/user/dev/oss/open62541-ruby/tmp/x86_64-linux/open62541/3.2.2build/open62541 -DCMAKE_INSTALL_PREFIX=/home/user/dev/oss/open62541-ruby/tmp/x86_64-lin
ux/open62541/3.2.2/stage --log-level=VERBOSE -DUA_ENABLE_DIAGNOSTICS=ON -DUA_NAMESPACE_ZERO=FULL -DUA_MULTITHREADING=100 -DUA_LOGLEVEL=200 -DUA_ENABLE_ENCRYPTION=MBEDTLS -DCMAKE_BUILD_TYPE=Debug -DUA_DEBUG=1 -DCMAKE_C_FLAGS_DEBUG='-fPIC -O0 -g' -DCM
AKE_INTERPROCEDURAL_OPTIMIZATION=OFF
```

open62541 version: 1.4.16

OS: Ubuntu Linux 24.04

paddor:
## Summary

- After an unsuccessful `UA_Client_getEndpoints()` (e.g. TCP connects but server never responds), calling `UA_Client_delete()` causes 100% CPU usage because `UA_ClientConfig_clear` loops forever waiting for the event loop to reach `STOPPED`
- Added two safety valves to the event loop shutdown loop: check `el->run()` return value (matching the existing `disconnectSecureChannel` pattern) and limit iterations to 10 (1s max)
- `el->free()` is still called after the loop, so resources are force-cleaned regardless

Fixes #7771

## Test plan

1. Start `nc -l 4840` as a fake unresponsive OPC UA server
2. Run a client that calls `UA_Client_getEndpoints("opc.tcp://localhost:4840")`, let it timeout, then `UA_Client_delete()`
3. Before fix: hangs at 100% CPU in `UA_ClientConfig_clear`
4. After fix: exits cleanly within ~1s

### DIFF
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -232,6 +232,11 @@ UA_Client_clear(UA_Client *client) {
     client->sessionState = oldState;
 
     UA_Client_disconnect(client);
+
+    /* Prevent reconnection attempts during the EventLoop teardown
+     * in UA_ClientConfig_clear */
+    client->connectStatus = UA_STATUSCODE_BADSHUTDOWN;
+
     UA_String_clear(&client->discoveryUrl);
     UA_EndpointDescription_clear(&client->endpoint);
 


jpfr:
Lets try to fix the root cause.
Can we set the Status differently before droppping into the _clear?

waiting a second seems … Dirty

paddor:
You're right. PR updated.

PR #7850, opened 2026-03-10
Fix inverted condition in UA_Client_getNamespaceUri

paddor:
## Summary

The condition in `UA_Client_getNamespaceUri` (`src/client/ua_client.c`) is inverted:

```c
if(index > client->namespacesSize)
    res = UA_String_copy(&client->namespaces[index], nsUri);
else
    res = UA_STATUSCODE_BADNOTFOUND;
```

This causes:
- `UA_STATUSCODE_BADNOTFOUND` for all **valid** indices (`index <= namespacesSize`)
- Out-of-bounds memory access for all **invalid** indices (`index > namespacesSize`)

PR #7857, opened 2026-03-11
Fix inverted condition in UA_Client_getNamespaceUri

paddor:
## Summary

- The condition in `UA_Client_getNamespaceUri` is inverted (`>` instead of `<`), causing `UA_STATUSCODE_BADNOTFOUND` for valid indices and out-of-bounds memory access for invalid ones
- Same fix as #7850, retargeted to `1.5` per maintainer request

Supersedes #7850

## Test plan

- Valid namespace index (e.g. 0) should return the namespace URI
- Invalid namespace index (e.g. 9999) should return `UA_STATUSCODE_BADNOTFOUND` without memory corruption


jpfr:
Thanks

PR #7858, opened 2026-03-11
fix(tests): Remove Null pointer dereference in `client.c`

vil02:
There is no need to run `fclose` at this point.

Minimal example: https://godbolt.org/z/zbs4jE63o


jpfr:
Thanks

Issue #7652, opened 2026-01-19
Documentation how to compile 1.5 branch not up to date


keba-estr:
## Description

Migrating from 1.4 to 1.5 branch, I was looking for a documentation on how to compile the 1.5 branch on debian using a docker container.
In the documentation from the development branch (https://www.open62541.org/doc/master/building.html#building-debian-packages-inside-docker-container-with-cmake-on-ubuntu-or-debian), the documentation still references the file "./tools/prepare_packaging.py", which does not exist neither on the 1.5 branch nor on the development branch any more.

Please update the documentation on how to compile the 1.5 (or development) branch.

## Checklist

Please provide the following information:

 - [X] open62541 Version (release number or git tag): 1.5-rc2
 - [X] Operating system: debian

keba-estr:
Resolved with PR #7865

PR #7865, opened 2026-03-12
docs(core): Add reference to external salsa repository for debian build

keba-estr:


keba-estr:
Should resolve issue #7652

PR #7866, opened 2026-03-12
fix(build) fix missing clock implementation in Win32/POSIX amalgamation and dirent in Win32

StalderT:
## Problem

When generating a compatible Win32/POSIX amalgamation with:

```
-DUA_ENABLE_AMALGAMATION=ON
-DUA_ARCHITECTURE=none
```

the generated `open62541.c` is missing the clock implementation, leading to unresolved symbols at link time.

Example build error:

```
/usr/bin/ld: open62541.c:(.text+0x11a5c): undefined reference to `UA_DateTime_now`
/usr/bin/ld: open62541.c:(.text+0x94dad): undefined reference to `UA_DateTime_localTimeUtcOffset`
/usr/bin/ld: open62541.c:(.text+0x96cdb): undefined reference to `UA_DateTime_nowMonotonic`
collect2: error: ld returned 1 exit status
```

These functions are normally provided by:

```
arch/posix/clock_posix.c
```

However this source file is not included when `UA_ARCHITECTURE=none` is used to generate the compatible amalgamation.

## Root Cause

`config.h` correctly resolves the architecture at compile time:

- `_WIN32` → `UA_ARCHITECTURE_WIN32`
- otherwise → `UA_ARCHITECTURE_POSIX`

And `clock_posix.c` already contains both implementations guarded by preprocessor checks.

However the build system only includes the file when `UA_ARCHITECTURE_POSIX` or `UA_ARCHITECTURE_WIN32` are set at configure time.  
This excludes the `UA_ARCHITECTURE=none` amalgamation case.

## Fix

Always include `arch/posix/clock_posix.c` when generating an amalgamation.  
The correct implementation will still be selected at compile time via the existing preprocessor guards.

```diff
-if(UA_ARCHITECTURE_POSIX OR UA_ARCHITECTURE_WIN32)
+if(UA_ARCHITECTURE_POSIX OR UA_ARCHITECTURE_WIN32 OR UA_ENABLE_AMALGAMATION)
     list(APPEND plugin_sources ${PROJECT_SOURCE_DIR}/arch/posix/clock_posix.c)
 endif()
```

## Reproduction

```
mkdir build
cd build

cmake ..   -DUA_ENABLE_AMALGAMATION=ON   -DUA_ARCHITECTURE=none

cmake --build .
```

Compile a simple program using the generated `open62541.c`.

Linker errors appear for:

- `UA_DateTime_now`
- `UA_DateTime_localTimeUtcOffset`
- `UA_DateTime_nowMonotonic`

## Result

The compatible Win32/POSIX amalgamation builds correctly again and works as intended with runtime architecture selection.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -856,7 +856,7 @@ set(lib_headers ${PROJECT_SOURCE_DIR}/deps/open62541_queue.h
                 ${PROJECT_SOURCE_DIR}/src/server/ua_server_internal.h
                 ${PROJECT_SOURCE_DIR}/src/client/ua_client_internal.h)
 
-if(UA_ENABLE_ENCRYPTION AND UA_ARCHITECTURE_WIN32)
+if(UA_ENABLE_ENCRYPTION AND (UA_ARCHITECTURE_WIN32 OR UA_ENABLE_AMALGAMATION))
     list(APPEND lib_headers ${PROJECT_SOURCE_DIR}/deps/tr_dirent.h)
 endif()
 
--- a/deps/tr_dirent.h
+++ b/deps/tr_dirent.h
@@ -9,6 +9,8 @@
 #ifndef DIRENT_H
 #define DIRENT_H
 
+#ifdef _WIN32
+
 /* Hide warnings about unreferenced local functions */
 #if defined(__clang__)
 #	pragma clang diagnostic ignored "-Wunused-function"
@@ -1237,4 +1239,7 @@ dirent_set_errno(int error)
 #ifdef __cplusplus
 }
 #endif
+
+#endif /*_WIN32*/
+
 #endif /*DIRENT_H*/


basyskom-jvoe:
This is also very important for my use of the open62541 amalgamation.
In my PR #7767 I have added a CI check for this on Linux to keep the none architecture working for future versions.
Maybe you have an idea how we could check this for Windows too? I made an attempt but failed to find a viable solution.

StalderT:
Yes, this is important for my use of the compatible amalgamation as well.

For Linux I replaced the previous `gcc -c open62541.c` check with a small
smoke test that actually compiles and links a minimal program using the
public API. This catches missing symbols that only appear at link time.

For Windows I added a similar check in the MinGW CI job. It builds the
amalgamation with:

    -DUA_ENABLE_AMALGAMATION=ON
    -DUA_ARCHITECTURE=none

and then compiles a minimal application against the generated
`open62541.c/.h`. This ensures the compatible Win32/POSIX amalgamation
also works in the Windows toolchain.

---

# Small Suggestion Regarding Release Amalgamation

Small suggestion regarding the release amalgamation build.

It might be beneficial to generate the official amalgamated release with:

    -DUA_ARCHITECTURE=none

instead of forcing a specific architecture.

With `UA_ARCHITECTURE=none`, the generated `open62541.h` keeps all
architecture macros undefined and lets the compiler select the correct
one at build time:

```c
/* #undef UA_ARCHITECTURE_WIN32 */
/* #undef UA_ARCHITECTURE_POSIX */
/* #undef UA_ARCHITECTURE_ZEPHYR */
/* #undef UA_ARCHITECTURE_LWIP */
/* #undef UA_ARCHITECTURE_FREERTOS */

#if !defined(UA_ARCHITECTURE_WIN32) && !defined(UA_ARCHITECTURE_POSIX) &&     !defined(UA_ARCHITECTURE_ZEPHYR) && !defined(UA_ARCHITECTURE_LWIP)
# ifdef _WIN32
#  define UA_ARCHITECTURE_WIN32
# else
#  define UA_ARCHITECTURE_POSIX
# endif
#endif
```

This allows a single amalgamated source to build correctly on both
Windows and POSIX systems without regenerating the amalgamation.

Currently the released amalgamation appears to force POSIX:

```c
/* #undef UA_ARCHITECTURE_WIN32 */
#define UA_ARCHITECTURE_POSIX
/* #undef UA_ARCHITECTURE_ZEPHYR */
/* #undef UA_ARCHITECTURE_LWIP */
/* #undef UA_ARCHITECTURE_FREERTOS */
```

Using `UA_ARCHITECTURE=none` would make the release amalgamation more
portable and consistent with the intended architecture auto-selection
logic already present in `open62541.h`.

Just a thought — there may be reasons for the current setup that I'm not aware of.

StalderT:
From my side everything looks good now. The PR should be ready to merge.

StalderT:
For this patch, I’d like to ask whether it would make sense to consider it for inclusion in the 1.5 release.

The main motivation here is to ensure that the amalgamation in 1.5 properly supports multi-architecture setups (both Win32 and POSIX), which seems increasingly important for downstream users.

Additionally, for the master branch, would there be interest in renaming the "none" option to "auto"? This might better reflect the behavior and improve clarity.

If there is interest from the maintainers, I’d be happy to further refine and formally propose this patch.

Thanks for your feedback!

andreasebner:
Thanks for contributing. Merged in #7965

PR #7874, opened 2026-03-13
feat(ci) add MinGW cross-compilation test for amalgamation

StalderT:
Add a Linux CI check that cross-compiles the compatible amalgamation with MinGW-w64.

This verifies that the amalgamation can also be compiled for Windows targets and helps prevent regressions like #7829.


jpfr:
Thanks

PR #7884, opened 2026-03-16
perf(server_subscription) index monitored items by id for fast lookup

StalderT:
## Problem

When deleting a large number of MonitoredItems (e.g. tens of thousands), the server becomes very slow.

Profiling shows that `Operation_DeleteMonitoredItem()` spends most of its time in `UA_Subscription_getMonitoredItem()`, which currently performs a linear scan over the subscription's monitored item list.

With large subscriptions (e.g. 50k items), this results in O(n²) behavior when deleting many items.

## Solution

Introduce an additional index of monitored items inside `UA_Subscription` using the existing `ziptree` structure.

The linked list is kept unchanged for iteration, while the new tree is used for fast lookup by `monitoredItemId`.

Changes:
- add a `ziptree` index (`monitoredItemsById`) in `UA_Subscription`
- add a tree entry in `UA_MonitoredItem`
- insert items into the tree on creation
- remove items from the tree on deletion
- replace the linear scan in `UA_Subscription_getMonitoredItem()` with a tree lookup

## Result

Lookup of monitored items becomes **O(log n)** instead of **O(n)**.

This significantly improves performance when deleting large numbers of monitored items (e.g. from clients such as UA Expert with tens of thousands of monitored items).

No functional behavior changes.

This change does not modify existing APIs and keeps the original list structure for iteration compatibility.

### DIFF
--- a/src/server/ua_server_ns0.c
+++ b/src/server/ua_server_ns0.c
@@ -687,6 +687,30 @@ resendData(UA_Server *server, const UA_NodeId *sessionId, void *sessionContext,
     unlockServer(server);
     return UA_STATUSCODE_GOOD;
 }
+
+static void *
+countMonitoredItemsVisitor(void *context, UA_MonitoredItem *monitoredItem) {
+    UA_UInt32 *sizeOfOutput = (UA_UInt32*)context;
+    (void)monitoredItem;
+    ++(*sizeOfOutput);
+    return NULL;
+}
+
+struct FillHandlesContext {
+    UA_UInt32 *clientHandles;
+    UA_UInt32 *serverHandles;
+    UA_UInt32 *i;
+};
+
+static void *
+fillHandlesVisitor(void *context, UA_MonitoredItem *monitoredItem) {
+    struct FillHandlesContext *ctx = (struct FillHandlesContext*)context;
+    ctx->clientHandles[*ctx->i] = monitoredItem->parameters.clientHandle;
+    ctx->serverHandles[*ctx->i] = monitoredItem->monitoredItemId;
+    ++(*ctx->i);
+    return NULL;
+}
+
 static UA_StatusCode
 readMonitoredItems(UA_Server *server, const UA_NodeId *sessionId, void *sessionContext,
                    const UA_NodeId *methodId, void *methodContext, const UA_NodeId *objectId,
@@ -727,10 +751,8 @@ readMonitoredItems(UA_Server *server, const UA_NodeId *sessionId, void *sessionC
 
     /* Count the MonitoredItems */
     UA_UInt32 sizeOfOutput = 0;
-    UA_MonitoredItem* monitoredItem;
-    LIST_FOREACH(monitoredItem, &subscription->monitoredItems, listEntry) {
-        ++sizeOfOutput;
-    }
+    ZIP_ITER(UA_MonitoredItemIdTree, &subscription->monitoredItemsById,
+             countMonitoredItemsVisitor, &sizeOfOutput);
     if(sizeOfOutput == 0) {
         unlockServer(server);
         return UA_STATUSCODE_GOOD;
@@ -753,11 +775,12 @@ readMonitoredItems(UA_Server *server, const UA_NodeId *sessionId, void *sessionC
 
     /* Fill the array */
     UA_UInt32 i = 0;
-    LIST_FOREACH(monitoredItem, &subscription->monitoredItems, listEntry) {
-        clientHandles[i] = monitoredItem->parameters.clientHandle;
-        serverHandles[i] = monitoredItem->monitoredItemId;
-        ++i;
-    }
+    struct FillHandlesContext ctx;
+    ctx.clientHandles = clientHandles;
+    ctx.serverHandles = serverHandles;
+    ctx.i = &i;
+    ZIP_ITER(UA_MonitoredItemIdTree, &subscription->monitoredItemsById,
+             fillHandlesVisitor, &ctx);
     UA_Variant_setArray(&output[0], serverHandles, sizeOfOutput, &UA_TYPES[UA_TYPES_UINT32]);
     UA_Variant_setArray(&output[1], clientHandles, sizeOfOutput, &UA_TYPES[UA_TYPES_UINT32]);
 
--- a/src/server/ua_server_ns0.c
+++ b/src/server/ua_server_ns0.c
@@ -699,6 +699,30 @@ resendData(UA_Server *server, const UA_NodeId *sessionId, void *sessionContext,
     unlockServer(server);
     return UA_STATUSCODE_GOOD;
 }
+
+static void *
+countMonitoredItemsVisitor(void *context, UA_MonitoredItem *monitoredItem) {
+    UA_UInt32 *sizeOfOutput = (UA_UInt32*)context;
+    (void)monitoredItem;
+    ++(*sizeOfOutput);
+    return NULL;
+}
+
+struct FillHandlesContext {
+    UA_UInt32 *clientHandles;
+    UA_UInt32 *serverHandles;
+    UA_UInt32 *i;
+};
+
+static void *
+fillHandlesVisitor(void *context, UA_MonitoredItem *monitoredItem) {
+    struct FillHandlesContext *ctx = (struct FillHandlesContext*)context;
+    ctx->clientHandles[*ctx->i] = monitoredItem->parameters.clientHandle;
+    ctx->serverHandles[*ctx->i] = monitoredItem->monitoredItemId;
+    ++(*ctx->i);
+    return NULL;
+}
+
 static UA_StatusCode
 readMonitoredItems(UA_Server *server, const UA_NodeId *sessionId, void *sessionContext,
                    const UA_NodeId *methodId, void *methodContext, const UA_NodeId *objectId,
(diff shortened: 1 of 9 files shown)


jpfr:
Good change!
Can we remove the linked list then?
Right now the list remains and we have the ziptree…

StalderT:
Yes, I think removing the linked list is possible.

However, that would turn this change into a broader refactor. The list is still used in several iteration paths (e.g. subscription teardown and cleanup), and replacing it would require updating all those paths to use the ziptree.

One tricky point is iteration combined with deletion: the current code relies on LIST_FOREACH_SAFE patterns, while ZIP_ITER does not directly provide the same guarantees, so this would need some care (e.g. collecting nodes first or adding a safe iteration helper).

My initial goal here was to keep the patch minimal and only introduce the tree as a fast index for monitoredItemId lookups, without changing existing iteration logic.

If you prefer, I can go further and remove the linked list entirely and switch to the ziptree as the primary container. 

Depending on your preference, I can work on that next.

jpfr:
If you remove all elements, zip_iter is safe.
Because it goes bottom up and will only process a parent once all its children are done.

I prefer not having double both lists and tree.

StalderT:
I looked into removing the linked list entirely.

For `UA_Subscription_delete()`, this part is actually straightforward. The current safe list iteration can be replaced with epeated removal of the minimum element from the ziptree:

```diff
--- a/src/server/ua_subscription.c
+++ b/src/server/ua_subscription.c
@@ -334,8 +334,12 @@ UA_Subscription_delete(UA_Server *server, UA_Subscription *sub) {

     /* Delete monitored Items */
     UA_assert(server->monitoredItemsSize >= sub->monitoredItemsSize);
-    UA_MonitoredItem *mon, *tmp_mon;
-    LIST_FOREACH_SAFE(mon, &sub->monitoredItems, listEntry, tmp_mon) {
+
+    while(sub->monitoredItemsSize > 0) {
+        UA_MonitoredItem *mon =
+            ZIP_MIN(UA_MonitoredItemIdTree, &sub->monitoredItemsById);
+        if(!mon)
+            break;
         UA_MonitoredItem_delete(server, mon);
     }
     UA_assert(sub->monitoredItemsSize == 0);
```

So for deletion alone, removing the list is easy.

However, the linked list is also used in several other places for full iteration,
for example in `Service_ModifySubscription()`. Converting those paths means adding
extra code with visitor callbacks and context structs, for example:

```diff
+struct UpdateSamplingContext {
+    UA_Server *server;
+    UA_Double oldPublishingInterval;
+};
+
+static void *
+updateSamplingIntervalVisitor(void *context, UA_MonitoredItem *mon) {
+    struct UpdateSamplingContext *ctx =
+        (struct UpdateSamplingContext*)context;
+
+    if(mon->parameters.samplingInterval == mon->subscription->publishingInterval ||
+       mon->parameters.samplingInterval == ctx->oldPublishingInterval) {
+        UA_MonitoredItem_unregisterSampling(ctx->server, mon);
+        UA_MonitoredItem_registerSampling(ctx->server, mon);
+    }
+
+    return NULL;
+}
+
+
+
 UA_Boolean
 Service_ModifySubscription(UA_Server *server, UA_Session *session,
                            const UA_ModifySubscriptionRequest *request,
@@ -216,14 +237,12 @@ Service_ModifySubscription(UA_Server *server, UA_Session *session,
         /* For each MonitoredItem check if it was/shall be attached to the
          * publish interval. This ensures that we have less cyclic callbacks
          * registered and that the notifications are fresh. */
-        UA_MonitoredItem *mon;
-        LIST_FOREACH(mon, &sub->monitoredItems, listEntry) {
-            if(mon->parameters.samplingInterval == sub->publishingInterval ||
-               mon->parameters.samplingInterval == oldPublishingInterval) {
-                UA_MonitoredItem_unregisterSampling(server, mon);
-                UA_MonitoredItem_registerSampling(server, mon);
-            }
-        }
+        struct UpdateSamplingContext ctx;
+        ctx.server = server;
+        ctx.oldPublishingInterval = oldPublishingInterval;
+
+        ZIP_ITER(UA_MonitoredItemIdTree, &sub->monitoredItemsById,
+                 updateSamplingIntervalVisitor, &ctx);
     }
```

The same happens in `UA_Subscription_resendData()`:

```diff
+static void *
+resendDataMonitoredItemVisitor(void *context, UA_MonitoredItem *mon) {
+    UA_Server *server = (UA_Server*)context;
+
+    /* Create only DataChange notifications */
+    if(mon->itemToMonitor.attributeId == UA_ATTRIBUTEID_EVENTNOTIFIER)
+        return NULL;
+
+    /* Only if the mode is monitoring */
+    if(mon->monitoringMode != UA_MONITORINGMODE_REPORTING)
+        return NULL;
+
+    /* If a value is queued for a data MonitoredItem, the next value in
+     * the queue is sent in the Publish response. */
+    if(mon->queueSize > 0)
+        return NULL;
+
+    /* Create a notification with the last sampled value */
+    UA_MonitoredItem_createDataChangeNotification(server, mon, &mon->lastValue);
+    return NULL;
+}
+
 void
 UA_Subscription_resendData(UA_Server *server, UA_Subscription *sub) {
     UA_LOCK_ASSERT(&server->serviceMutex);
@@ -966,24 +992,8 @@ UA_Subscription_resendData(UA_Server *server, UA_Subscription *sub) {
      * queued for a data MonitoredItem, the next value in the queue is sent in
      * the Publish response. If no value is queued for a data MonitoredItem, the
      * last value sent is repeated in the Publish response. */
-    UA_MonitoredItem *mon;
-    LIST_FOREACH(mon, &sub->monitoredItems, listEntry) {
-        /* Create only DataChange notifications */
-        if(mon->itemToMonitor.attributeId == UA_ATTRIBUTEID_EVENTNOTIFIER)
-            continue;
-
-        /* Only if the mode is monitoring */
-        if(mon->monitoringMode != UA_MONITORINGMODE_REPORTING)
-            continue;
-
-        /* If a value is queued for a data MonitoredItem, the next value in
-         * the queue is sent in the Publish response. */
-        if(mon->queueSize > 0)
-            continue;
-
-        /* Create a notification with the last sampled value */
-        UA_MonitoredItem_createDataChangeNotification(server, mon, &mon->lastValue);
-    }
+    ZIP_ITER(UA_MonitoredItemIdTree, &sub->monitoredItemsById,
+             resendDataMonitoredItemVisitor, server);
 }
```

And there are more places like this, for example in `src/server/ua_server_ns0.c`.

So removing the list is possible, but it is not just a small cleanup. It means touching several iteration paths and adding a fair amount of extra code.

Because of that, my recommendation would be to keep both structures here:
- the ziptree as a fast index for `monitoredItemId` lookup
- the linked list for the existing full-iteration paths

If you still want the linked list removed, I can continue working on that, but it will turn this into a broader refactor rather than just the lookup performance fix.

StalderT:
Good point.

After looking at it more closely, I agree the linked list should be removed.

Keeping both the list and the ziptree means we would have to keep two different containers perfectly in sync across several iteration / modification paths.

Given the number of LIST_FOREACH_SAFE-style usages and places where monitored items are modified or removed, that feels too error-prone.

So I think the right fix is to remove the linked list entirely and make the ziptree the single source of truth.

I’ll rework this properly and come back with a complete patch updating all affected iteration paths.

StalderT:
The patch is now ready for review.

It does touch several places since LIST usage is quite widespread, but the changes are mechanical: LIST_FOREACH / LIST_* are replaced by the existing ZIP tree equivalents. The overall logic, behavior and control flow remain unchanged.

Even if this targets 1.5, I think it would be worth considering. In its current state, the LIST-based iteration does not scale well and becomes a real limitation with larger numbers of monitored items (~10k and above). In practice, the server can no longer keep up and clients like UaExpert start disconnecting.

So for higher loads, the current implementation is difficult to use reliably.

Some quick benchmarks (removing monitored items):

Without the patch:
- 20,000 → ~3 s
- 30,000 → ~7 s
- 40,000 → UaExpert disconnects after ~20 s

With the patch:
- 20,000 → instantaneous
- 30,000 → instantaneous
- 40,000 → instantaneous
- 100,000 → still instantaneous

With the ZIP-based iteration, behavior stays identical but performance improves drastically and remains stable even with large numbers of monitored items. Operations that previously took seconds (or led to timeouts) become almost instantaneous.

So while the diff is not small, it is mostly mechanical and addresses a real scalability issue.

Let me know what you think.

StalderT:
I added a dedicated test to cover the changes introduced in this patch and ensure the behavior remains correct.

The test specifically targets the new code path and helps prevent regressions.

StalderT:
Hi! I noticed that this PR now has merge conflicts because it has been open for a while.

Before I spend time rebasing and updating it, I wanted to check whether you are still interested in this patch being merged. If so, I can update the branch and resolve the conflicts. Otherwise, no problem.

I just wanted to confirm before doing the extra work.

jpfr:
@StalderT there are a few additional commits on top of this now.
Please check. If okay from your side and the CI is green, then I will squash-merge.
This leaves you as the author.

StalderT:
Thanks for the refactor commits, the full switch to the ziptree is much cleaner than keeping both structures.

Reviewed and tested locally: all subscription tests pass (incl. the 3 new ones), Valgrind clean, and the delete path is back to O(n·log n):

| N   | before | with this PR |
|-----|--------|--------------|
| 40k | 6.6 s  | 0.31 s       |
| 80k | 50 s   | 0.72 s       |

Good to merge from my side. I really think this patch is a genuine improvement for open62541

jpfr:
Thanks @StalderT, merged. 👍

PR #7885, opened 2026-03-16
perf(timer): improve batching scalability for large numbers of timers

StalderT:
## Problem

When creating a large number of cyclic timers, the current batching logic scans all existing timers to find a batching candidate.

With many timers, this becomes very expensive and significantly slows down operations such as creating large numbers of monitored items.

## Solution

Replace the global scan with a bounded search in the time-ordered timer tree.

The new implementation keeps the existing batching semantics:
- only batch within the allowed time window
- only batch compatible intervals
- prefer exact interval matches

This improves scalability while preserving timer behavior.

## Result

Creating large numbers of monitored items becomes much faster, while the existing batching behavior is preserved.

### DIFF
--- a/arch/common/timer.c
+++ b/arch/common/timer.c
@@ -54,14 +54,12 @@ static UA_DateTime earliest, latest, adjustedNextTime;
 
 static void *
 findTimer2Batch(void *context, UA_TimerEntry *compare) {
-    UA_TimerEntry *te = (UA_TimerEntry*)context;
-
-    /* NextTime deviation within interval? */
-    if(compare->nextTime < earliest || compare->nextTime > latest)
-        return NULL;
+    /* Invariance of ZIP_ITER_KEY  */
+    UA_assert(compare->nextTime >= earliest && compare->nextTime <= latest);
 
     /* One-shot timers have interval == 0.
      * They cannot participate in the modulo-based batching check. */
+    UA_TimerEntry *te = (UA_TimerEntry*)context;
     if(te->interval == 0 || compare->interval == 0)
         return NULL;
 


jpfr:
The previous code did a simple iterate over the timers.
Now you do tree splitting with some recursion. That looks pretty dangerous.

Can’t we simply look for the entry-point, that is the earliest timer in the window?
And then do a linear scan until we arrive at the end of the window (and then break).

That should be doable by a specially crafted cmp method for the tree lookup of the earliest candidate.

jpfr:
With a special cmp method, all we need is ZIP_ITER_KEY.
And the cmp method tells us if we are in the window (equal), before or after.

jpfr:
The we can also break the iter once a good enough candidate is found.

StalderT:
I reworked it in the original direction.

This version removes the recursive tree walk and instead defines a batching window key for ZIP_ITER_KEY. It starts at the first timer inside the window and then scans linearly through that window only.

The batching selection logic stays intentionally simple and close to the previous code:
- keep the last compatible candidate
- stop immediately on an exact interval match

StalderT:
Tested with large datasets (up to 100k monitored items).

The behavior remains consistent with the original implementation, but avoids scanning the entire timer set. The batching still follows the same logic (last compatible match, early exit on exact interval), just restricted to the relevant time window.

Performance is significantly improved while keeping the same observable behavior.

jpfr:
Sounds good.
Since this goes on the stable release 1.5, keep the patch at a minimum.

Keep the static global variables.
Then you don’t need to define an additional struct.
For the main code, only the line with ZIP_ITER needs to change.

How big is the speed improvement?

StalderT:
I updated the patch to keep it minimal as suggested:
- kept the existing static globals
- kept the original batching logic unchanged
- only replaced the ZIP_ITER with ZIP_ITER_KEY on the time-ordered tree

Here are the benchmarks from my setup (UaExpert, monitored items creation):

Without the patch:
- 10k: ~1 s
- 20k: ~1.4 s
- 30k: ~3.6 s
- 40k: ~7 s
- 50k: UaExpert disconnects after ~20 s
- 100k: UaExpert disconnects after ~20 s

With the patch:
- 10k: ~0.4 s
- 20k: ~0.5 s
- 30k: ~1 s
- 40k: ~1.3 s
- 50k: ~1.5 s
- 100k: ~3 s

The improvement is significant, especially for large numbers of monitored items where the previous implementation leads to timeouts and client disconnects.

jpfr:
This is great now.
Small nitpick: This check inside findTimer2Batch is now redundant - it only gets called when that is true.

```c
    /* NextTime deviation within interval? */
    if(compare->nextTime < earliest || compare->nextTime > latest)
        return NULL;
```

jpfr:
Even better: Make it into an assert that allow timers check via ZIP_ITER_KEY are in the correct time range.

jpfr:
The change is added directly to the PR.

jpfr:
Thanks @StalderT

StalderT:
Thanks for merging!

Small patch, but a big performance improvement 🙂

PR #7887, opened 2026-03-16
fix(plugin): remove stat pre-check and handle mkdir races via errno

o6-nbeier:
This fixes a TOCTOU of `mkpath` which was caused by using `UA_stat` for checking on the dir. This approach was switched to a safe `mkdir`-approach that returns success on `UA_mkdir == 0` and if dir already exists. If parent directories do not exist, it first recurses on their creation and then retries the actual `mkdir`.


jpfr:
👍

PR #7888, opened 2026-03-16
fix(pubsub): Timeout when disabling all PubSub components

keba-estr:
When calling UA_Server_disableAllPubSubComponents, a timeout might occur if the writers are disabled before the readers in case of a loopback configuration on the same server. So disable the reader side before the writer side.


jpfr:
The motivation is clear.
Add this as a description to the comments.
Otherwise future developers will not have the insight.

keba-estr:
> The motivation is clear. Add this as a description to the comments. Otherwise future developers will not have the insight.

I rewrote the comment in the header file.

PR #7891, opened 2026-03-16
fix(server): correct range check in UA_CAST_SIGNED macro

o6-nbeier:
The bounds check `(t)i > T##_MAX` cast the source value to the narrower target type before comparing, making the expression always false (e.g. `(UA_Byte)i > 255`). Out-of-range values therefore passed the check silently and were stored with truncation. Now, we compare the original UA_Int64 value against the bounds directly.


jpfr:
👍

PR #7894, opened 2026-03-16
feat(test): add dictionary for fuzz_xml_decode_encode

DavidKorczynski:
Adds an OPC UA XML encoding dictionary for Variant decoding, including XML syntax tokens, built-in scalar types, and commonly assembled tags.


jpfr:
Nice

Issue #7893, opened 2026-03-16
server : slow performance server with many monitoreditems


StalderT:
To reproduce the issue I used the following minimal server with 50'000 variables:

``` c
#include "open62541.h"

#define NODE_COUNT 50000

int main() {
    UA_Server *server = UA_Server_new();
    UA_ServerConfig *config = UA_Server_getConfig(server);
    UA_ServerConfig_setDefault(config);

    for(size_t i = 0; i < NODE_COUNT; i++) {

        char name[64];
        snprintf(name, sizeof(name), "Var_%zu", i);

        UA_VariableAttributes attr = UA_VariableAttributes_default;

        UA_Int32 value = (UA_Int32)i;
        UA_Variant_setScalar(&attr.value, &value, &UA_TYPES[UA_TYPES_INT32]);

        attr.displayName = UA_LOCALIZEDTEXT("en-US", name);
        attr.dataType = UA_TYPES[UA_TYPES_INT32].typeId;

        UA_NodeId nodeId = UA_NODEID_STRING(1, name);
        UA_QualifiedName nodeName = UA_QUALIFIEDNAME(1, name);
        UA_NodeId parentNodeId = UA_NODEID_NUMERIC(0, UA_NS0ID_OBJECTSFOLDER);
        UA_NodeId parentReferenceNodeId = UA_NODEID_NUMERIC(0, UA_NS0ID_ORGANIZES);

        UA_Server_addVariableNode(server,
                                  nodeId,
                                  parentNodeId,
                                  parentReferenceNodeId,
                                  nodeName,
                                  UA_NODEID_NUMERIC(0, UA_NS0ID_BASEDATAVARIABLETYPE),
                                  attr,
                                  NULL,
                                  NULL);
    }

    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_USERLAND,
                "Server started with %d variables", NODE_COUNT);

    UA_Server_runUntilInterrupt(server);

    UA_Server_delete(server);
    return 0;
}
```

This simply creates 50'000 variables.

Then I connect with UA Expert and try to monitor all variables.

## Problem 1 - very slow creation of monitored items

Creating the 50'000 monitored items takes more than 20 seconds. UA Expert disconnects during that time.

The main reason is the linear traversal of the monitored item list when looking up monitored items.

Using a tree index instead of a list makes this operation much faster.

This is implemented in: https://github.com/open62541/open62541/pull/7885

[...]

PR #7898, opened 2026-03-17
fix(mt): fix data races in timer implementation

o6-nbeier:
Replace the three file-scope static variables (`earliest`, `latest`, `adjustedNextTime`) used as a communication channel between `batchTimerEntry` and `findTimer2Batch` with a stack-allocated `BatchContext` struct passed via the callback context argument. The statics were only safe when a single `UA_Timer` instance existed. Concurrent calls on different instances would race on these globals even though each held its own mutex.

Move the `batchTimerEntry` call in `UA_Timer_add` to inside the `timerMutex` critical section. Previously it was invoked before `UA_LOCK`, causing `ZIP_ITER` to traverse `idTree` without protection while another thread could concurrently modify it under the lock.

Release `timerMutex` before invoking the user callback in `processEntryCallback` and re-acquire it afterwards, as required by the contract documented in `timer.h`. Calling the callback while holding the lock would deadlock any callback that re-enters the timer API (`UA_Timer_add`, `UA_Timer_modify`, `UA_Timer_remove`).

### DIFF
--- a/arch/common/timer.c
+++ b/arch/common/timer.c
@@ -49,20 +49,15 @@ UA_Timer_init(UA_Timer *t) {
     UA_LOCK_INIT(&t->timerMutex);
 }
 
-struct BatchContext {
-    UA_TimerEntry *te;
-    UA_DateTime earliest;
-    UA_DateTime latest;
-    UA_DateTime adjustedNextTime;
-};
+/* Global variables, only used behind the mutex */
+static UA_DateTime earliest, latest, adjustedNextTime;
 
 static void *
 findTimer2Batch(void *context, UA_TimerEntry *compare) {
-    struct BatchContext *bc = (struct BatchContext*)context;
-    UA_TimerEntry *te = bc->te;
+    UA_TimerEntry *te = (UA_TimerEntry*)context;
 
     /* NextTime deviation within interval? */
-    if(compare->nextTime < bc->earliest || compare->nextTime > bc->latest)
+    if(compare->nextTime < earliest || compare->nextTime > latest)
         return NULL;
 
     /* One-shot timers have interval == 0.
@@ -76,7 +71,7 @@ findTimer2Batch(void *context, UA_TimerEntry *compare) {
     if(te->interval > compare->interval && te->interval % compare->interval != 0)
         return NULL;
 
-    bc->adjustedNextTime = compare->nextTime; /* Candidate found */
+    adjustedNextTime = compare->nextTime; /* Candidate found */
 
     /* Abort when a perfect match is found */
     return (te->interval == compare->interval) ? te : NULL;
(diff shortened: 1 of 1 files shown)


StalderT:
Hi,

I just wanted to raise a concern regarding a related issue that is currently quite critical for us: https://github.com/open62541/open62541/issues/7893

The problem described there makes open62541 very difficult to use when dealing with a large number of items. I previously proposed a fix here: https://github.com/open62541/open62541/pull/7885 which specifically targets that issue.

The approach proposed in this PR seems to be in conflict with the solution implemented in that PR, and might prevent it from being integrated or working correctly.

Since this issue is currently blocking for us, would it be possible to take it into account when working on this change? It would be really helpful to ensure that this use case is still addressed, or to find a way to make both approaches compatible.

Thanks a lot for your time and consideration.

Best regards,

arch/common/timer.c
```diff
@@ -137,11 +144,12 @@ UA_Timer_add(UA_Timer *t, UA_Callback callback,
     te->nextTime = nextTime;
     te->timerPolicy = timerPolicy;
 
+    /* Insert into the timer */
+    UA_LOCK(&t->timerMutex);
```
jpfr, line 141: It looks like moving the lock solves the core issue.
The rest is just ceremony.

arch/common/timer.c
```diff
@@ -235,12 +243,16 @@ processEntryCallback(void *context, UA_TimerEntry *te) {
     struct TimerProcessContext *tpc = (struct TimerProcessContext*)context;
     UA_Timer *t = tpc->t;
 
-    /* Execute the callback */
-    if(te->cb) {
+    /* Release the lock before calling the user callback.
+     * The timer can be modified from within the callback (see timer.h). */
+    UA_UNLOCK(&t->timerMutex);
```
jpfr: Our locks are reentrant from the same thread.
Don’t unlock around the callbacks into the application.

But we need to ensure that locks are always taken in the same order to prevent deadlocks.

arch/common/timer.c
```diff
@@ -49,15 +49,20 @@ UA_Timer_init(UA_Timer *t) {
     UA_LOCK_INIT(&t->timerMutex);
 }
 
-/* Global variables, only used behind the mutex */
-static UA_DateTime earliest, latest, adjustedNextTime;
+struct BatchContext {
+    UA_TimerEntry *te;
```
jpfr: The global variables are fine if the code using it is in a critical section/behind a mutex.

We also have UA_THREAD_LOCAL.
But thread local variables have some overhead also.

jpfr:
@StalderT  Let’s merge a reduced version of this PR for the locking issues first.
And then do your speed improvement. Both for 1.5.

PR #7900, opened 2026-03-17
refactor(arch): change functions related to time for zephyr

ge0053:
…fset

- changing UA_DateTime_now to use clock_gettime instead of gettimeofday.
- make UA_DateTime_localTimeUtcOffset return 0 since zephyr, as of now, doesn't support local and UTC time.
  - this is mostly to reduce code size since mktime() adds allot of  code without providing any functionality.

### DIFF
--- a/arch/zephyr/clock_zephyr.c
+++ b/arch/zephyr/clock_zephyr.c
@@ -19,22 +19,20 @@
 
 UA_DateTime
 UA_DateTime_now(void) {
-    struct timeval tv;
-    gettimeofday(&tv, NULL);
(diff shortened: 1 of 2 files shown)


jpfr:
UA_DATETIME_NSEC I have a bad feeling about.
Like it might get cast to an integer zero in some situations.
This is a potential bug source.
Let’s do without.

ge0053:
did some more reading and found out that the libc variants (picolivc, newlib, newlib-nano) does add support to set and calculate time zones.
I need to rethink the pull request as the main "problem" is that UA_DateTime_localTimeUtcOffset() currently adds alot of code to calculate local time offset.

PR #7902, opened 2026-03-18
WIP - fix(core): separate certificate validation status code

muesli4:
Add a separate UA_StatusCode for certificate validation client responses.  In the validation algorithm, hide the client response for the steps and return codes mentioned in the specification.

# Advantages
* It enables a more verbatim implementation of the specification.  In the [specification](https://reference.opcfoundation.org/Core/Part4/v105/docs/6.1.3) table 100 lists all of the steps and the corresponding return values including the result for the client.  This should be very maintainable if the specification changes.  (I also updated the code with AI, which I checked line by line, using a link to the specification.)
* Code using the verification result can pick the status code it wants.
* The code trades a few conditional jumps, if the status code is replaced conditionally instead, against a few additional bytes in stack frames, from the return value.

# Disadvantages
* In 1.5, the ABI is broken, since one part of the CertificateGroup plugin API is exposed, and we also need to expose the type itself. _I could base this on master instead._
* Some minor duplication for different crypto backends.

# Future Work
* The resulting `UA_SplitStatusCode` might need to be pushed down further to reach all potential responses, where necessary.

# Testing Done
I ran a full CTT run with the standard server profile and all of the failures related to return values, except for those related to `Bad_CertificateTimeInvalid`, went away.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1125,7 +1125,8 @@ endif()
 
 if(UA_ENABLE_ENCRYPTION_MBEDTLS OR UA_ENABLE_AMALGAMATION)
     list(INSERT plugin_sources 0
-         ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/securitypolicy_common.h)
+         ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/securitypolicy_common.h
+         ${PROJECT_SOURCE_DIR}/plugins/crypto/certificategroup_common.h)
     list(APPEND plugin_sources
          ${PROJECT_SOURCE_DIR}/plugins/crypto/mbedtls/securitypolicy_common.c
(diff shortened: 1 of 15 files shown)


muesli4:
I discussed this PR with @jpfr and we cannot merge this PR for version 1.5 because it breaks the API.  I will try to investigate whether the masking can happen at the end of `validateCertificate` (i.e. https://github.com/open62541/open62541/pull/7779).  

I should check whether there are status codes that are returned as `Bad_SecurityChecksFailed` in some steps of the certificate validation algorithm but not in others.  I believe this is the case and also why it did not work by simply inspecting the status codes before.

muesli4:
This is the previous PR: https://github.com/open62541/open62541/pull/7779

muesli4:
I found a better and less invasive way to resolve the issues.

PR #7908, opened 2026-03-19
fix(core): handle the case where the calloc implementation returns NULL.

r-sin:
Depending on the implementation, calloc may return NULL when n_elements is 0.
Our implementation returns NULL, causing UA_DataType_toEnumDescription to report the misleading status code UA_STATUSCODE_BADOUTOFMEMORY when selecting the Types/DataTypes/BaseDataType/Enumeration node in UAExpert.

The related fixes in UA_DataType_fromStructureDescription, UA_DataType_toStructureDescription, and UA_DataType_fromEnumDescription address the same issue, even though I was unable to trigger the n_elements == 0 case in those paths.


jpfr:
👍

PR #7915, opened 2026-03-19
refactor(nc): resolve logging-fstring-interpolation pylint warning

vil02:
Similar to #7903.


jpfr:
I like f-strings.
What is wrong here?

vil02:
> I like f-strings.

I also like them.

> What is wrong here?

In the old code the string is always constructed, despite the logging level. In the new code, the string is constructed only, when it is needed (i.e. the logging level is such, that this is string has to be written into the log).

jpfr:
Ok, cool. If that makes pylint happy.

PR #7917, opened 2026-03-19
 fix(server): allow removal of CA certificates without valid CRL

andreasebner:
When removing a CA certificate from the trust store, getCertificateCrls may return BadNoMatch if no CRL is associated with that certificate. Previously, this was treated as a hard failure, preventing the certificate from being removed.

This PR extends and replaces #7662


jpfr:
👍

PR #7923, opened 2026-03-20
feat(test): add harness for unfuzzed type parsing routines

DavidKorczynski:


jpfr:
Looks good. Check if you can build on this: #7925
Essentially the same, yet a bit more "linear".

Since we expect the numerical fields to be at least >=2 characters long, we can early abort and keep the size-field for parseInt64 constant.

DavidKorczynski:
> Looks good. Check if you can build on this: #7925 Essentially the same, yet a bit more "linear".
> 
> Since we expect the numerical fields to be at least >=2 characters long, we can early abort and keep the size-field for parseInt64 constant.

thanks -- works well. Have rebased the new harness on top of that.

jpfr:
Cool, merged.

PR #7927, opened 2026-03-23
fix(core): fix the case where the last chunk may exceed the max message size.

r-sin:
If a transferred message exceeds the configured tcpMaxMsgSize, the server must return UA_STATUSCODE_BADTCPMESSAGETOOLARGE. This did not occur in our tests with a slightly oversized message. This modification ensures that the final chunk is correctly validated against the maximum message size.


jpfr:
Good find. We will take your commit and apply it to all previous affected versions.

PR #7942, opened 2026-03-26
feat(core): add ANSI color support for Windows 10 and above

m-obi:
Added Windows support for ANSI color output and initialized colors based on the OS version.


jpfr:
That's a good change.
But note that a lot of usage goes through the global logger variable.
So it will not pass through `UA_Log_Stdout_new`.
Can you detect >win10 at compile time?

m-obi:
Ah yes. Forgot.

Let's discard this PR and I will write something what get called before the main() from the user runs.
Using this neat principle:
https://stackoverflow.com/questions/1113409/attribute-constructor-equivalent-in-vc

Detecting Win10 while compile-time doesn't make sense, cause we doesn't know if the user is maybe running the application on Win7 for example.

jpfr:
@m-obi you have a status?

m-obi (commented):
Sorry for the delay. Was much other stuff ongoing and forgot to give an update. But I got it working. 

I found a way to run the initialization of the coloring before the user's main() run.

jpfr:
Thanks.
Please have a look at the remaining ci fails

PR #7814, opened 2026-03-05
fix(ex): fix key copy and memory leaks in pubsub TPM keystore publisher

o6-nbeier:
Replace out-of-bounds single-byte key assignments with memcpy for the signing and encrypting keys. Also free allocated output buffers on the early-return error paths to avoid memory leaks.
Fixes #7809.

Issue #7809, opened 2026-03-04
Memory issues in `pubsub_publish_encrypted_tpm_keystore.c`


vil02:
I observed some _memory_ like issues in [`pubsub_publish_encrypted_tpm_keystore.c`](https://github.com/open62541/open62541/blob/52053b84243d285067ca532a3501d27a92b600c4/examples/pubsub/pubsub_publish_encrypted_tpm_keystore.c), namely:
1. _index out of bounds_ in lines:
https://github.com/open62541/open62541/blob/52053b84243d285067ca532a3501d27a92b600c4/examples/pubsub/pubsub_publish_encrypted_tpm_keystore.c#L544-L545
please note the definitions:
https://github.com/open62541/open62541/blob/52053b84243d285067ca532a3501d27a92b600c4/examples/pubsub/pubsub_publish_encrypted_tpm_keystore.c#L20-L25

This can be probably fixed by replacing these lines with something like:
```c
memcopy(signingKey, sign_out_data->data, UA_AES128CTR_SIGNING_KEY_LENGTH); 
memcopy(encryptingKey, encrypt_out_data->data, UA_AES128CTR_KEY_LENGTH); 
````

2. There is a possible memory leak in the lines:
https://github.com/open62541/open62541/blob/52053b84243d285067ca532a3501d27a92b600c4/examples/pubsub/pubsub_publish_encrypted_tpm_keystore.c#L536
https://github.com/open62541/open62541/blob/52053b84243d285067ca532a3501d27a92b600c4/examples/pubsub/pubsub_publish_encrypted_tpm_keystore.c#L541
Note the allocations: https://github.com/open62541/open62541/blob/52053b84243d285067ca532a3501d27a92b600c4/examples/pubsub/pubsub_publish_encrypted_tpm_keystore.c#L525
https://github.com/open62541/open62541/blob/52053b84243d285067ca532a3501d27a92b600c4/examples/pubsub/pubsub_publish_encrypted_tpm_keystore.c#L529

jpfr:
Fixed via #7814

PR #7946, opened 2026-03-28
fix(ex): fix key copy and memory leaks in pubsub TPM keystore subscriber

vil02:
Similar to #7814.

The clean-up part could be simplified but I think it is worth to have this file similar to `pubsub_publish_encrypted_tpm_keystore.c`.

I missed that while creating #7809.


jpfr:
👍

PR #7950, opened 2026-03-29
feat(server): Add const option for generated UA_DataType

andreasebner:
UA_DataTypeArray.types and UA_DataType.members become const; UA_TYPES and all generated member arrays are const. Type arrays of additional nodesets can be made const by pinning their namespace indices at generation time
(NAMESPACE_MAP / --namespaceMap). The generated init code verifies the runtime namespace order and fails with BadInternalError on mismatch. Unpinned arrays keep the existing load-time rewrite.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,8 +21,7 @@ The `types` field in `UA_DataTypeArray` changed from `UA_DataType *` to
 them `const` makes this explicit in the API, protects against accidental
 modification and allows the toolchain to place the definitions in
 read-only memory. DataType arrays generated for additional nodesets
-remain mutable, as their NamespaceIndex gets adjusted when the nodeset
-is loaded into a server.
+remain mutable by default (see below).
 
 The `members` field in `UA_DataType` changed from `UA_DataTypeMember *`
 to `const UA_DataTypeMember *`. Generated member arrays are declared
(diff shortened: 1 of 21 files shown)


jpfr:

diplfranzhoepfinger:
Just wanted to flag that we're actively relying on this for an embedded (ESP32-P4/RISC-V) OPC UA server+client project, and it solves a real problem for us better than what we were attempting separately (closing #8322 and #8323 in favor of this).

Verified locally on our actual target toolchain: both `UA_TYPES`/`UA_TYPES_TRANSPORT` and the generated `*_members` arrays end up in flash-mapped `.rodata` -- no separate `.data.rel.ro` section, no runtime relocation, since ESP-IDF apps link non-PIC with absolute addressing. That's real internal RAM and PSRAM freed up for us, for free, with none of the opt-in flags or runtime-constructor machinery we'd built to get a much smaller part of the way there. The namespace-index-pinning approach for companion nodesets is also exactly the right design for a problem we only found by hitting it on hardware (self-referencing/companion-spec types breaking once the array is naively made const).

I'll write up our testing/results here in a bit more detail shortly.

In the meantime -- a very friendly ask: any chance of bumping the priority on this one? I know it currently has merge conflicts and no review decision yet, but from where we're sitting the design is sound and the payoff (both for us and probably for anyone else running open62541 on RAM-constrained targets) is significant. Happy to help test or rebase if that's useful.

diplfranzhoepfinger:
Quick follow-up on our earlier testing, in case it's useful: we hit a scary-looking crash today (Guru Meditation / load access fault inside `UA_ServerConfig_setDefault`'s default SecurityPolicy/endpoint setup) right after switching our project to this branch, and initially suspected a regression here.

Root-caused it to our own build process, not this branch: our downstream integration keeps a separately-generated amalgamation (`open62541.c`/`.h`) that we hadn't regenerated after bumping the submodule, so our own precompiled library (built against the old struct layout) ended up linked against a freshly-compiled `open62541.c` from this branch -- a plain ABI mismatch on our end. Once we regenerated the amalgamation and did a full rebuild of our own code against it, everything works cleanly on real ESP32-P4 hardware: local OPC UA server (including our optional WebSocket transport) starts up fine, both triggered at server startup and via a deployed application, no crash.

So: no issue found in this branch itself from our side -- just flagging in case anyone else hits something that looks similar during their own testing, so it's not mistaken for a regression here.

franz-ms-muc:
Wow. Cool

PR #7963, opened 2026-03-31
fix(server): guard securityPolicyUriPostfix against empty URI

lonelybones:
securityPolicyUriPostfix iterates backward using raw pointer arithmetic. When uri.length is 0 and uri.data is NULL, the initial pointer wraps to 0xFFFFFFFFFFFFFFFF and the loop dereferences wild memory. Add an early return for empty strings.


jpfr:
👍

PR #7969, opened 2026-04-02
fix(server): support Good_PasswordChangeRequired from AccessControl

mcpat-pilz:
The fixes the issue where Service_ActivateSession() overwrites the serviceResult after activateSession() from the AccessControl plugin returned either Good or Good_PasswordChangeRequired.

Only in case of an error in any subsequent step will the serviceResult be properly adjusted.

### DIFF
--- a/src/server/ua_services_session.c
+++ b/src/server/ua_services_session.c
@@ -1093,7 +1093,7 @@ Service_ActivateSession(UA_Server *server, UA_SecureChannel *channel,
         activateSession(server, &server->config.accessControl, ed,
                         &channel->remoteCertificate, &session->sessionId,
                         &req->userIdentityToken, &session->context);
-    if(rh->serviceResult != UA_STATUSCODE_GOOD) {
+    if(!UA_StatusCode_isGood(rh->serviceResult)) {
         UA_LOG_ERROR_SESSION(server->config.logging, session,
                              "ActivateSession: The AccessControl plugin "
                              "denied the activation with the StatusCode %s",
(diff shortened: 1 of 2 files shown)


src/server/ua_services_session.c
```diff
@@ -1085,14 +1085,17 @@ Service_ActivateSession(UA_Server *server, UA_SecureChannel *channel,
         activateSession(server, &server->config.accessControl, ed,
                         &channel->remoteCertificate, &session->sessionId,
                         &req->userIdentityToken, &session->context);
-    if(rh->serviceResult != UA_STATUSCODE_GOOD) {
+    if((rh->serviceResult != UA_STATUSCODE_GOOD) && (rh->serviceResult != UA_STATUSCODE_GOODPASSWORDCHANGEREQUIRED)) {
```
jpfr: It looks like UA_StatusCode_isGood would also apply here.

src/server/ua_services_session.c
```diff
@@ -1085,14 +1085,17 @@ Service_ActivateSession(UA_Server *server, UA_SecureChannel *channel,
...
         UA_LOG_ERROR_SESSION(server->config.logging, session,
                              "ActivateSession: The AccessControl plugin "
                              "denied the activation with the StatusCode %s",
                              UA_StatusCode_name(rh->serviceResult));
         UA_SECURITY_REJECT;
     }
 
+    // We must not touch the serviceResult if everything works fine!
```
jpfr: This can be moved to below - where stepResult is used for the first time.
Don't use C++ // comments.

jpfr (commented):
Looks good. 👍 
However some builds break with
```
/__w/open62541/open62541/src/server/ua_services_session.c:1088:75: error: 'UA_STATUSCODE_GOODPASSWORDCHANGEREQUIRED' undeclared (first use in this function)
 1088 |     if((rh->serviceResult != UA_STATUSCODE_GOOD) && (rh->serviceResult != UA_STATUSCODE_GOODPASSWORDCHANGEREQUIRED)) {
      |                                                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/__w/open62541/open62541/src/server/ua_services_session.c:1088:75: note: each undeclared identifier is reported only once for each function it appears in
```

Try to use UA_StatusCode_isGood instead of checking for a specific status code.

mcpat-pilz:
@jpfr Issues should be fixed now.

PR #7976, opened 2026-04-07
feat(posix): expose per-iteration work-performed flag via EventLoop params

paddor:
No way to tell whether a `run()` iteration did useful I/O work. Applications that want adaptive scheduling (back off when idle, tick faster under load) have to guess or poll timers.

The POSIX EventLoop now sets a `UA_Boolean` under `"work-performed"` in `el->params` after each `run()` iteration. `true` if any `eventSourceCB` fired, `false` otherwise. Applications read it with `UA_KeyValueMap_getScalar()`. No new public types or struct fields. Scope is I/O dispatch only (timers not tracked).


paddor:
Ready for second review

paddor:
@o6-jpfr @jpfr I've just rebased this PR. Please give your feedback. We've been using this patch in production since April.

jpfr:
@paddor You make a change in the public API (the kv-map is part of the API) but don't document it.
Are you also going to align the other eventloop implementations?

Another way to find out if work was performed is to track the duration spent in the _run method.
Because the eventloop returns early from its "sleep duration" when it has been woken up by an event.

paddor:
@o6-jpfr @jpfr It's documented now.

Since we develop and run our software on typical Linux systems, I don't think I can update the other event loops. Is it necessary or will you accept incremental development?

Believe me, I tried using the runtime of `_run` method as a work signal. It does not work. Systems are too different, and even the best solution will just "guess" based on timing, which is fragile.

PR rebased.

PR #7977, opened 2026-04-08
refactor(arch): Add lwIP EventLoop mutex

r-sin:
We repeatedly encounter a deadlock when the server thread and an app thread contend for the timer mutex and the service mutex, with each thread holding one of them.
This change aligns the lwIP implementation with the POSIX implementation.


jpfr:
Yes! This aligns to the locking behavior of the POSIX implementation.

Deadlocks are avoided by an absolute ordering of mutexes.
Always first the EventLoop lock and only then the server/ client lock.

This is why the EventLoop exposes its lock in the api.

Issue #7844, opened 2026-03-09
subscriptions alarms conditions: memory allocation problems


laroche:
## Description

If UA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS is enabled on the 1.5 branch and
clang is used together with some memory checking options like
UA_ENABLE_STATIC_ANALYZER=FULL or -fsanitize=undefined -fsanitize=address
-fsanitize=memory -fsanitize-memory-track-origins

## Background Information / Reproduction Steps

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=<YOUR_OPTION> <ANY_OTHER_OPTIONS> ..
```

## Checklist

Please provide the following information:

 - [x] open62541 Version (release number or git tag): current 1.5 branch
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Debian-14 (current development branch) with clang-19
 - [x] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

<pre>
==406910==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x55b299a7c92d in calloc (/home/flaroche/data/open62541/open62541-1.5/build/bin/tests/check_server_alarmsconditions+0x58192d) (BuildId: 11b05ea59c43437eab39684965a8b6035029b675)
    #1 0x55b299ac8346 in UA_Array_copy /home/flaroche/data/open62541/open62541-1.5/src/ua_types.c:2638:12
    #2 0x55b299ad01e6 in Variant_copy /home/flaroche/data/open62541/open62541-1.5/src/ua_types.c:1376:28
    #3 0x55b299ad0005 in DataValue_copy /home/flaroche/data/open62541/open62541-1.5/src/ua_types.c:1820:28
    #4 0x55b299ac76fb in UA_copy /home/flaroche/data/open62541/open62541-1.5/src/ua_types.c:2044:28
    #5 0x55b299c4607c in UA_DataValue_copy /home/flaroche/data/open62541/open62541-1.5/build/src_generated/open62541/types_generated.h:781:1
    #6 0x55b299c4607c in readInternalValueAttribute /home/flaroche/data/open62541/open62541-1.5/src/server/ua_services_attribute.c:163:9
[...]

PR #7978, opened 2026-04-08
[REVIEW] fix(alarms): clear temporary Variants in limit readers

StalderT:
fix for #7844

The reported leak is visible through getHighLimit in the issue trace, but the same temporary-Variant pattern is used by all four limit readers (LowLimit/LowLowLimit/HighLimit/HighHighLimit). This patch clears the local UA_Variant in each helper and keeps the change minimal.


jpfr:
Thanks.
Good to see you also improved the CI!

Issue #7807, opened 2026-03-04
RemoveCertificate memory problem


laroche:
## Description
Compiling the current 1.5 branch on Linux+mbedtls with clang together with
sanitize-options on Debian-14 (current development branch), the included testsuite
has one failing test within RemoveCertificate accessing the wrong memory:
AddressSanitizer: stack-use-after-scope

Showing more info:
<pre>
==263499==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7f79496d6180 at pc 0x558ff0c4ded4 bp 0x7f794b3fd080 sp 0x7f794b3fd078
READ of size 8 at 0x7f79496d6180 thread T6
    #0 0x558ff0c4ded3 in stringOrder /home/flaroche/data/open62541/open62541-1.5/src/ua_types.c:2215:26
    #1 0x558ff0c7d9a8 in UA_ByteString_equal /home/flaroche/data/open62541/open62541-1.5/build/src_generated/open62541/types_generated.h:524:1
    #2 0x558ff0c7d9a8 in UA_TrustListDataType_remove /home/flaroche/data/open62541/open62541-1.5/src/util/ua_util.c:1418:20
    #3 0x558ff0ce4124 in removeCertificate /home/flaroche/data/open62541/open62541-1.5/src/server/ua_server_ns0_gds.c:592:18
    #4 0x558ff0ce4124 in removeCertificateAction /home/flaroche/data/open62541/open62541-1.5/src/server/ua_server_ns0_gds.c:1568:25
    #5 0x558ff0d7f00a in callWithResolvedMethodAndObject /home/flaroche/data/open62541/open62541-1.5/src/server/ua_services_method.c:243:12
    #6 0x558ff0d7f00a in callWithMethodAndObject /home/flaroche/data/open62541/open62541-1.5/src/server/ua_services_method.c:319:11
    #7 0x558ff0d7f00a in Operation_CallMethod /home/flaroche/data/open62541/open62541-1.5/src/server/ua_services_method.c:367:9
    #8 0x558ff0d1ba83 in Service_Call /home/flaroche/data/open62541/open62541-1.5/src/server/ua_server_async.c:808:27
    #9 0x558ff0d60cc0 in processServiceInternal /home/flaroche/data/open62541/open62541-1.5/src/server/ua_services.c:290:12
    #10 0x558ff0d60cc0 in processRequest /home/flaroche/data/open62541/open62541-1.5/src/server/ua_services.c:348:23
[...]

PR #7979, opened 2026-04-08
[REVIEW] fix(gds): avoid stack-use-after-scope in removeCertificate

StalderT:
Possible fix for #7807

removeCertificate stored the address of a loop-local UA_ByteString in the temporary TrustList passed to removeFromTrustList(). The pointer was then used after the loop-local variable went out of scope, triggering ASan in RemoveCertificate. Move the temporary certificate variable to the outer scope.


laroche:
Confirmed fixed from my side.

Thanks a lot for creating this patch,

Florian La Roche

jpfr:
Nice, thanks

PR #7985, opened 2026-04-09
fix(server): prevent infinite loop in UA_ServerConfig_clear

paddor:
`UA_Server_run_shutdown` unconditionally sets `server->state = STOPPED` after its drain loops, even when those loops bailed early (`res != GOOD`). This lets `UA_Server_delete` reach `UA_ServerConfig_clear`, whose unbounded `while(el->state != STOPPED) el->run(el, 100)` loop spins forever if the EventLoop is stuck in STOPPING.

Fix: `setServerLifecycleState()` now refuses the transition to STOPPED while the EventLoop is still draining. Companion to the client-side fix in #7856.

### DIFF
--- a/src/server/ua_server.c
+++ b/src/server/ua_server.c
@@ -1222,6 +1222,14 @@ setServerLifecycleState(UA_Server *server, UA_LifecycleState state) {
     if(server->state == state)
         return;
 
+    /* Refuse to mark STOPPED while the EventLoop is still draining */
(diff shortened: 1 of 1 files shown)


jpfr:
This should be handled inside setServerLifecycleState

paddor:
@o6-jpfr @jpfr I've just rebased this PR. Please give your feedback. We've been using this patch in production since April.

paddor:
@o6-jpfr @jpfr PR rebased.

PR #7989, opened 2026-04-13
fix(server): method call status codes

mcpat-pilz:
* "executable" attribute is not set? ->  BadNotExecutable
* the user does not have permissions? -> BadUserAccessDenied

See also https://reference.opcfoundation.org/Core/Part4/v104/docs/5.11.2 and later.


jpfr:
👍

Issue #5778, opened 2023-05-10
UA_Server_newWithConfig does not copy "cv" pointer of the CertContext structure


Alexk12:
## Description

UA_Server_newWithConfig does not copy "cv" pointer of the CertContext structure

After call to this function CertContext structure's cv pointer is equal to &config.certificateVerification, while it must be equql to &cfg->certificateVerification, where UA_ServerConfig* cfg = UA_Server_getConfig(server).

In more details.
The idea of UA_Server_newWithConfig function is:
1) prepare config structure
2) pass this structure to UA_Server_newWithConfig
3) UA_Server_newWithConfig function makes a copy of "config" and memset(s) initial "config" to 0.

This idea is bad as the UA_Server_newWithConfig function does not know the members and the contents of CertContext structure (prepared within "config") and has no idea how to copy it correctly.

UA_Server_newWithConfig function is used, for example, in open62541/examples/server_ctt.c
Below is an extraction of code that illustrates the problem and shows how to fix it.

`
#ifdef _DEBUG
    // Debug notice: CertContext structure is defined in ua_pki_openssl.c (structure is used only within ua_pki_openssl.c)
    // We defined CertContext1 structure above (absolute equivalent to CertContext) just for debugging purposes,
    // to check members of CertContext structure.
#define CertContext CertContext1
#endif

#ifdef _DEBUG
    {
        UA_CertificateVerification* configBefore_certificateVerification = &config.certificateVerification;
        CertContext* contextBefore = (CertContext*)config.certificateVerification.context;
        UA_CertificateVerification* contextBefore_cv = (UA_CertificateVerification*)contextBefore->cv;
        assert(contextBefore_cv == &config.certificateVerification);
        CertContext* contextBefore_cv_context = (CertContext*)contextBefore->cv->context;
        UA_CertificateVerification* pCV = contextBefore_cv_context->cv;
        assert(pCV == &config.certificateVerification);
    }
#endif
    server = UA_Server_newWithConfig(&config);
#ifdef _DEBUG
    {
        UA_ServerConfig* cfg = UA_Server_getConfig(server);
        UA_CertificateVerification* configAfter_certificateVerification = &cfg->certificateVerification;
        CertContext* contextAfter = (CertContext*)cfg->certificateVerification.context;
        UA_CertificateVerification* contextAfter_cv = (UA_CertificateVerification*)contextAfter->cv;
[...]

PR #7995, opened 2026-04-20
Fix crash when logging in v1.4

karzhenkov:
The following code may cause the server to crash, depending on the actual configuration:
```c++
static void initialize_server_config(UA_ServerConfig *config) { ... }

static UA_Server *create_server(void) {
    UA_ServerConfig config;
    memset(&config, 0, sizeof(config));
    initialize_server_config(&config);
    return UA_Server_newWithConfig(&config);
}
```

Note that the above `create_server` function is almost identical to [`UA_Server_new`](https://github.com/open62541/open62541/blob/v1.4.16/plugins/ua_config_default.c#L72-L80). The proposed test differs in that it creates a non-default configuration, which cannot be copied correctly. The configuration in the test is initialized using `UA_ServerConfig_setDefaultWithSecurityPolicies`. The bug has already been reported in #5778.


jpfr:
Merged. thanks!

PR #8001, opened 2026-04-22
fix(server) monitoreditem: allow samplingInterval=0 and restore value change notifications

### LINKED ISSUE #8000: Sampling interval = 0 does not behave as expected (regression?)
Hi,

I noticed a potential issue regarding the handling of a `samplingInterval` set to `0`.

According to the OPC UA specification, a sampling interval of `0` should mean *“fastest practical rate”*, i.e. the server should revise it to the minimum supported sampling interval.

However, since commit https://github.com/open62541/open62541/commit/9b759bbfb7a9d267ac097c1542ff387a7553a79f  it appears that using `samplingInterval = 0` no longer works as expected.

The behavior seems to be linked to this check in the code: 
https://github.com/open62541/open62541/blob/8d060e94c11ba5025438bc92bd3ae2df4ba87ed0/src/server/ua_services_monitoreditem.c#L311  

## Observed behavior  
- Creating a MonitoredItem with `samplingInterval = 0` does not result in a valid revised sampling interval  
- **Value changes are no longer reported** when using `samplingInterval = 0`  
- The monitored item appears inactive even when the underlying variable is changing  

## Expected behavior  
- The server should revise `0` to a valid minimum sampling interval  
- Value changes should still be reported at the fastest supported rate  
- The MonitoredItem should behave as if configured with the minimum possible sampling interval  

## How to reproduce  
This is very easy to test using UaExpert:
1. Connect to a server using open62541  
2. Subscribe to a variable that changes over time  
3. Set `samplingInterval = 0`  
4. Observe that value changes are not reported  

## Questions  
- Is this behavior intentional?  
- Should `samplingInterval = 0` still be supported according to the specification?  
- Would it make sense to remove or adjust the check rejecting `0` (see link above)?  

Thanks in advance for your clarification!

## Checklist

 - [X] open62541 Version 1.5:
 - [X] Other OPC UA SDKs used (UaExpert): 
 - [X] Critical issue

StalderT:
According to the OPC UA specification, a sampling interval of 0 means "fastest practical rate" and should be revised by the server to the minimum supported sampling interval.

Currently, due to a check introduced in https://github.com/open62541/open62541/commit/9b759bbfb7a9d267ac097c1542ff387a7553a79f, a samplingInterval of 0 prevents proper monitoring behavior. In particular, value changes are no longer reported when samplingInterval is set to 0.

This patch restores the expected behavior by allowing samplingInterval=0 and ensuring it is revised to a valid minimum interval, so that monitored items continue to report value changes.

Reproducible using UaExpert by creating a subscription with samplingInterval=0.

Fixes #8000

### DIFF
--- a/src/server/ua_services_monitoreditem.c
+++ b/src/server/ua_services_monitoreditem.c
@@ -307,14 +307,12 @@ checkAdjustMonitoredItemParams(UA_Server *server, UA_Session *session,
     if(mon->subscription && params->samplingInterval < 0.0)
         params->samplingInterval = mon->subscription->publishingInterval;
 
-    /* Adjust non-null sampling interval to lie within the configured limits */
-    if(params->samplingInterval != 0.0) {
-        UA_BOUNDEDVALUE_SETWBOUNDS(server->config.samplingIntervalLimits,
-                                   params->samplingInterval, params->samplingInterval);
-        /* Check for NaN */
-        if(mon->parameters.samplingInterval != mon->parameters.samplingInterval)
-            params->samplingInterval = server->config.samplingIntervalLimits.min;
-    }
+    /* Adjust sampling interval to lie within the configured limits */
+    UA_BOUNDEDVALUE_SETWBOUNDS(server->config.samplingIntervalLimits,
+                               params->samplingInterval, params->samplingInterval);
+    /* Check for NaN */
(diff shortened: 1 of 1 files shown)


jpfr:
Also fixed via https://github.com/open62541/open62541/pull/7753

PR #8002, opened 2026-04-24
fix(network): Ensure vhost data is initialized in websocket callback

diplfranzhoepfinger:
- Initialize per-vhost data if it is missing during the connection establishment callback to prevent null pointer access.
- Remove the hardcoded 10-second WebSocket ping-pong interval.


diplfranzhoepfinger:
https://github.com/diplfranzhoepfinger/opcua-dashboard
https://github.com/diplfranzhoepfinger/opcua-dashboard/tree/v1
Test Code.

PR #8014, opened 2026-04-28
fix(sec): guard plaintext block size underflows

o6-nbeier:
This PR fixes five RSA plaintext block size underflow issues in the OpenSSL security policy implementations and adds regression coverage for each case.

The affected policies are `basic128rsa15`, `aes128sha256rsaoaep`, `basic256sha1`, `basic256sha256` and `aes256sha256rsapss`. In each `getRemotePlainTextBlockSize` implementation, the code now checks the return value of `UA_Openssl_RSA_Public_GetKeyLength` and rejects RSA key lengths that are too small before subtracting the policy-specific padding size. This prevents wraparound when OpenSSL fails to provide a valid key length or returns a length below the minimum required threshold.

Each fix is accompanied by a regression test that uses a zeroed fake channel context to drive the failure path and verify that the function returns `0` instead of an underflowed size.

All of those were findings by a Coverity static code analysis.


jpfr:
👍

PR #8015, opened 2026-04-28
fix(server): Ensure non-zero condition event Time across all state callbacks

### LINKED ISSUE #6391: Condition Time set to 0 generating an Alarm from a timed callback
## Description
I have a timed callback that read a temperature value and, if this value is over a threshold, a Condition
is set to active state setting ActiveState->Id filed to true. Before the ActiveState->Id field is set to true,
I set the "Time" of the condition using

`UA_DateTime ts = UA_DateTime_now();`

`UA_Server_writeObjectProperty_scalar(
       server, conditionNodeId, UA_QUALIFIEDNAME(0, "Time"), &ts, &UA_TYPES[UA_TYPES_DATETIME]
);`

But in UAExpert, where the Alarm message, comment and status is showed correctly, I see 
2000-01-01T00:00:00.000Z as time value.

The correct time is set instead, in the Enabled, Acked and Confirmed Callback, using the same code as above.
It seems that, after I set the time in the callback, the Time is set again to 0 when the event is generated.

Used CMake options:

```bash
cmake .. -DCMAKE_BUILD_TYPE=MinSizeRel -DUA_LOGLEVEL=99 -DBUILD_SHARED_LIBS=yes 
-DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_SUBSCRIPTIONS_EVENTS=ON 
-DUA_ENABLE_SUBSCRIPTIONS_ALARMS_CONDITIONS=ON
```

## Checklist

 - [x] open62541 Version (release number or git tag): 1.4
 - [ x] Operating system: Debian 12
 - [ ] Critical issue

michael-itq:
PR #7555 fixed the Time fallback for the EnabledState/Disabled callback. The same root cause exists in five other condition state-change callbacks: when the application updates a condition state via the high- level helpers (setConditionVariableFieldProperty / setConditionField), the resulting DataValue carries sourceTimestamp = 0. Each after-write callback then copies that 0 into the condition's Time property, which propagates into every event subsequently fired -- both the events that auto-fire on state change and explicit UA_Server_triggerConditionEvent calls. Clients render this as the OPC UA epoch (1601-01-01).

Apply the same `if (sourceTimestamp == 0) eventTime = UA_DateTime_now();` guard that #7555 introduced to the remaining call sites:

  * afterWriteCallbackAckedStateChange      (Acked -> false)
  * afterWriteCallbackConfirmedStateChange  (Confirmed -> false, -> true)
  * afterWriteCallbackActiveStateChange     (Active -> true, -> false)
  * afterWriteCallbackSeverityChange        (Severity change)

After this change, condition events fired via the standard ActiveState toggle carry a wall-clock Time matching the activation moment, both for auto-fired events (severity-changed, active-changed) and for explicit triggerConditionEvent calls.

Verification (live capture against a server with this patch applied):

  Before:                                              After:
    severity has increased  Time=1601-01-01             Time=2026-04-28 18:49:41.646663+00:00
    Fault active            Time=1601-01-01             Time=2026-04-28 18:49:41.646892+00:00
    severity has decreased  Time=1601-01-01             Time=2026-04-28 18:49:36.646521+00:00
    Fault cleared           Time=1601-01-01             Time=2026-04-28 18:49:36.646521+00:00

Closes #6391

### DIFF
--- a/src/server/ua_subscription_alarms_conditions.c
+++ b/src/server/ua_subscription_alarms_conditions.c
@@ -962,8 +962,11 @@ afterWriteCallbackAckedStateChange(UA_Server *server,
      * That check makes it possible to set ackedState/Id to false, without triggering an event */
     if(*((UA_Boolean *)data->value.data) == false) {
         /* Set unacknowledging time */
+        UA_DateTime eventTime = data->sourceTimestamp;
+        if (eventTime == 0)
+            eventTime = UA_DateTime_now();
         retval = UA_Server_writeObjectProperty_scalar(server, conditionNode, fieldTimeQN,
-                                                      &data->sourceTimestamp,
+                                                      &eventTime,
                                                       &UA_TYPES[UA_TYPES_DATETIME]);
         CONDITION_ASSERT_RETURN_VOID(retval, "Set deactivating Time failed",
                                      UA_NodeId_clear(&conditionNode););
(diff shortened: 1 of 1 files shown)


jpfr:
Thanks for providing this.

PR #8018, opened 2026-04-29
[Review]: fix fetching common name [CN=] from cert

Akul123:
- When there is something after common name [CN=] in subject string it will be added to common name and possibly used for certificate fileName during in getCertFileName function.
If i.e. email is added after CN= in subject string content of common name is common name + email which results in bad internal error.
- This commit introduces fetching common name directly from certificate, similar to fetching thumbprint, instead of using pointer arithemetic on subject string.
- openssl and mbedtls implementations of the function UA_CertificateUtils_getCertCommonName
- New test case which uses certificate with email after common name in subject string is added to confirm this behaviour.


Akul123:
Hi @jpfr, now when all tests are passing can you please take a look when you find time?

jpfr:
Please add some comments where this behavior is described / mandated in the specification.
The changes themselves look good.

Akul123:
Here is short explanation of the commit. 
When the certificate subject string contains additional fields after the Common Name (CN=), those fields are incorrectly included in the extracted common name. The extracted value can later be used as the certificate filename in getCertFileName().
For example, if an email address appears after CN= in the subject string, the parsed common name becomes <common name> + <email> instead of only <common name>. This produces an invalid certificate filename and can lead to a BadInternalError.
This commit fixes that behavior by introducing UA_CertificateUtils_getCertCommonName(), which correctly extracts only the Common Name (CN) from the certificate subject string.
The implementation of UA_CertificateUtils_getCertCommonName() has been added for both OpenSSL and mbedTLS certificate groups in certificategroup.c. It uses the corresponding OpenSSL and mbedTLS library functions to extract the Common Name directly from the certificate subject instead of parsing the subject string manually.

Hope that now is more understandable.

jpfr:
I understand the change.
My worry is that the standard might be explicit about it.
And that we get bitten the next time we do the official certification.

Please add comments to the code with references to the specification in relation to the cn extraction.

Akul123:
Hi @jpfr,
according to the OPC UA reference for Application Certificate (https://reference.opcfoundation.org/specs/OPC-10000-6/6.2.2) and User Certificate (https://reference.opcfoundation.org/specs/OPC-10000-6/6.2.3) both contain subject as distinguished name. Subject is defined by RFC5280 (https://www.ietf.org/rfc/rfc5280.txt). OPC UA reference notes that Common Name (CN) attribute shall be specified in the subject field for both Application Instance Certificates and User Certificates, other subject attributes may also be specified. So the Common Name, in my opinion, must be correctly extracted as a single subject attribute instead of parsing all characters after "CN=" from the subject string. 

I added this comment in plugin/certificategroup.h‎ so it can stay as function explanation.

jpfr:
Great. 👍 
Merged.

PR #8020, opened 2026-04-30
tests(fuzz): add fuzz targets for oss-fuzz consumption

DavidKorczynski:
Adds three new harness for fuzzing server services, client, and pubsub connection config. This unlocks new areas for code coverage from the fuzzers.


jpfr:
👍

PR #8034, opened 2026-05-11
fix(ci): execute Coverity scans only on push to master and exclude dependabot branches explicitly

o6-nbeier:
Since dependabot creates a new branch directly inside the repository, the Coverity Scans are executed during its PRs but fail due to missing authentication information.
This PR changes the workflow trigger to be only executed on pushes to master and explicitly excludes all dependabot branches.


jpfr:
👍

PR #8036, opened 2026-05-11
fix(core): copy membersSize to destination in UA_DataType_copy

### LINKED ISSUE #7999: `UA_DataType_copy`  does not copy the `membersSize` field to the destination
see title
https://github.com/open62541/open62541/blob/1.5/src/ua_types.c#L128

 - [x] open62541 Version (release number or git tag): v1.5.4

o6-nbeier:
`UA_DataType_copy` zeroed `membersSize` but never updated it after successfully copying members. This left the destination struct in an inconsistent state where members pointed to valid allocated memory but `membersSize` was 0.
This fixes #7999.

### DIFF
--- a/src/ua_types.c
+++ b/src/ua_types.c
@@ -167,6 +167,7 @@ UA_DataType_copy(const UA_DataType *t1, UA_DataType *t2) {
             *(void**)(uintptr_t)&m2->memberName = mName;
 #endif
         }
+        t2->membersSize = t1->membersSize;
     }
 
  errout:


jpfr:
👍

RolfKal:
I'm not sure that is fully correct. The code in UA_DataType_copy() can theoretically bail out inside the loop that creates the memberName for each element with a UA_STATUSCODE_BADOUTOFMEMORY. The memberSize parameter will then be 0 but the members element was properly allocated in line 150 and there may have been successful memberName allocations. The subsequent UA_DataType_clear() that is called at the end when an error occured will not try to clear any of the typenames that may have succeeded before the outofmemory error.

I admit that this is a bit of a edge case, so it likely will never occur. But I think removing the setting of memberSize=0 and instead checking in the UA_DataType_clear() if members is not NULL before trying to free any strings in the members array would be safer.

Issue #7998, opened 2026-04-21
FindServers() fails if the server has no None mode


basyskom-jvoe:
If a server is configured without None security mode, calling `UA_Client_findServers()` fails with status `BadIdentityTokenRejected`.

This is my configuration:

<img width="977" height="319" alt="Image" src="https://github.com/user-attachments/assets/8a730792-1d53-4ac0-9f7c-e737c8494487" />

This is what Wireshark shows:

<img width="543" height="136" alt="Image" src="https://github.com/user-attachments/assets/21b0b305-038d-458f-a3c8-b5e16d13703b" />

All information displayed by Wireshark is unencrypted and returns the information that would be required to answer the `UA_Client_findServers()` request. When I enable the None security mode, it works and I see a second FindServers request after the FindServers and GetEndpoints requests shown in the first trace:

<img width="518" height="128" alt="Image" src="https://github.com/user-attachments/assets/7b3d052a-4f9a-42b0-9c0e-0b5c8f1d5c0e" />

This minimal example can be used to reproduce the issue:

```
#include <open62541/client.h>
#include <open62541/client_config_default.h>

#include <stdio.h>

int main(int argc, char *argv[])
{
    UA_Client *client = UA_Client_new();
    UA_ClientConfig_setDefault(UA_Client_getConfig(client));

    size_t registeredServersSize = 0;
    UA_ApplicationDescription *registeredServers = NULL;

    UA_StatusCode result = UA_Client_findServers(client, "opc.tcp://zosma:53530/OPCUA/SimulationServer",
                                                 0, NULL, 0, NULL, &registeredServersSize, &registeredServers);

    printf("Request returned %s and got %zu servers\n", UA_StatusCode_name(result), registeredServersSize);

    return 0;
}
```
This is the output:
```
[2026-04-21 09:24:31.529 (UTC+0200)] info/network	TCP 6	| Opening a connection to "zosma" on port 53530
[2026-04-21 09:24:31.557 (UTC+0200)] info/channel	TCP 6	| SC 1	| SecureChannel opened with SecurityMode None for SecurityPolicy http://opcfoundation.org/UA/SecurityPolicy#None and a revised lifetime of 600.0s
[2026-04-21 09:24:31.557 (UTC+0200)] info/client	Client Status: ChannelState: Open, SessionState: Closed, ConnectStatus: Good
[...]

PR #8042, opened 2026-05-13
Skip endpoint selection for discovery-only secure-channel connects

hit-labvn:
## Summary

 Related to #7998, as this issue affects from version 1.4.x

  Fix discovery-only client connections so `UA_Client_findServers()` can complete without selecting a session endpoint first.

  `connectSecureChannel()` sets `client->config.noSession = true`. In that mode the client only needs to open a SecureChannel, run discovery, and stop before endpoint selection and session setup. That makes `UA_Client_findServers()` fail on servers that allow discovery over a None SecureChannel but do not expose a usable None endpoint.

  ## Why

  `FindServers` is a discovery service. It should not depend on the number of endpoints, endpoint security policies, message security modes, or user token policies. Those are relevant for normal session connections, not for discovery-only secure-channel connects.

  ## Verification

  Tested against Prosys OPC UA Simulation Server:

image

  Before fix:

  ```text
    Rejecting endpoint 0: security policy not available
    No suitable endpoint found
    UA_Client_findServers status=0x80210000 servers=0
```
  After fix:
```
  UA_Client_findServers status=0x00000000 servers=1
  server[0] uri=urn:Username:OPCUA:SimulationServer discoveryUrls=1
  url[0]=opc.tcp://Username:53530
```

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -127,23 +127,25 @@ isFullyConnected(UA_Client *client) {
     if(client->channel.state != UA_SECURECHANNELSTATE_OPEN)
         return false;
 
-    /* GetEndpoints handshake ongoing or not yet done */
-    if(client->endpointsHandshake || endpointUnconfigured(&client->endpoint))
-        return false;
-
     /* FindServers handshake ongoing or not yet done */
     if(client->findServersHandshake || client->discoveryUrl.length == 0)
         return false;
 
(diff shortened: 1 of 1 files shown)


jpfr:
Please make this PR against the 1.4 branch if that is already affected.

Also add a unit test to prevent future regressions and to make the merge onto 1.5 and master testable.

PR #8045, opened 2026-05-13
feat(examples): client: allow configurable server

thorstenl312:
* Enable a remote server URL to be provided as an argument.
* Default to localhost if no arguments are provided.


RolfKal:
Helpful addition, but if you allow for a remote url, are you sure that server still uses the fixed username and password?

jpfr:
👍

PR #8047, opened 2026-05-14
fix(client): distinguish secure channel renewal from sending requests

### LINKED ISSUE #8046: Bad debug assertion condition in async client
## Description
`__Client_AsyncService()` asserts that any `sendRequest()` failure must leave the SecureChannel in `CLOSING` or `CLOSED`:
https://github.com/open62541/open62541/blob/0f638da3936609136baab9b917debd8901d0d163/src/client/ua_client.c#L830-L835

But `sendRequest()` can fail before attempting to send:
https://github.com/open62541/open62541/blob/0f638da3936609136baab9b917debd8901d0d163/src/client/ua_client.c#L362-L370

In this path, `closeSecureChannel()` is not called, so `client->channel.state` may still be `UA_SECURECHANNELSTATE_OPEN`.

This condition seems to be triggered when attempting to drive the event loop across sleep/wake cycles of the client machine, where the state of the connection goes stale due to client non-responsiveness, and secure channel renewal will fail outright.

1. Connect a client successfully.
2. Verify `client->channel.state == UA_SECURECHANNELSTATE_OPEN`.
3. Synthesize the edge case via `client->connectStatus = UA_STATUSCODE_BADTIMEOUT`, simulating the development of a stale connection.
4. Call `__UA_Client_AsyncService()`.

`sendRequest()`, in this case, returns `UA_STATUSCODE_BADTIMEOUT`, but the channel is still `OPEN`, causing the assertion to fail.

I added a repro test case to the client secure channel test suite in [this branch](https://github.com/cartercanedy/open62541/tree/cartercanedy/bad-assertion-repro).

```sh
git clone -b cartercanedy/bad-assertion-repro git@github.com:cartercanedy/open62541 && cd open62541
mkdir build && cd $_
cmake -DCMAKE_BUILD_TYPE=Debug -DUA_BUILD_UNIT_TESTS=ON ..
make -j
ctest -R securechannel --output-on-failure
```

~In my mind, `sendRequest()` should call `closeSecureChannel()` (or some new routine with clearer semantics that it's simply reconciling internal state) after the failed call to `__Client_renewSecureChannel()` that leaves the client in a disjoint state.~ The actual fix that I implemented just inlined the renewal of the secure channel, so that any errors produced by the secure channel renewal are disambiguated from errors produced by an IO cycle.

## Checklist

 - [x] open62541 Version (release number or git tag): `0f638da39`
 - [x] Other OPC UA SDKs used (client or server): Beckhoff OPC-UA server
 - [x] Operating system: Client = macOS, server = Windows
 - [x] Logs (with `UA_LOGLEVEL` set as low as necessary) attached: logs able to be reproduced with the [reproduction branch](https://github.com/cartercanedy/open62541/tree/cartercanedy/bad-assertion-repro)
 - [ ] Wireshark network dump attached: will attach if requested
 - [ ] Critical issue

cartercanedy:
Closes #8046

## Summary
- Move SecureChannel renewal and subsequent `connectStatus` check out of `sendRequest()` and into the sync/async service callers.
- Keep `sendRequest()` focused on the actual service-message send, so a non-good return from it still means the send path failed and the SecureChannel
  should be closing or closed.
- Fix the async assertion failure that can occur when the client has a stale logical connection status while the SecureChannel state still appears open.

## Details
`sendRequest()` could previously return a pre-existing `client->connectStatus` before attempting `UA_SecureChannel_sendMSG()`. The async caller then asserted that any non-good return meant the SecureChannel was already closing or closed, which is not true for pre-send client status failures.

This change performs the renewal/status preflight before calling `sendRequest()` in both `__Client_Service()` and `__Client_AsyncService()`.

### DIFF
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -364,11 +364,6 @@ sendRequest(UA_Client *client, const void *request,
             const UA_DataType *requestType, UA_UInt32 *requestId) {
     UA_LOCK_ASSERT(&client->clientMutex);
 
-    /* Renew SecureChannel if necessary */
-    __Client_renewSecureChannel(client);
-    if(client->connectStatus != UA_STATUSCODE_GOOD)
-        return client->connectStatus;
-
     UA_EventLoop *el = client->config.eventLoop;
 
     /* Adjusting the request header. The const attribute is violated, but we
@@ -677,6 +672,14 @@ __Client_Service(UA_Client *client, const void *request,
      * reconnection within the EventLoop run method. */
     UA_UInt32 channelId = client->channel.securityToken.channelId;
 
+    /* Renew SecureChannel if necessary */
+    __Client_renewSecureChannel(client);
+    if(client->connectStatus != UA_STATUSCODE_GOOD) {
+        notifyClientState(client);
+        respHeader->serviceResult = client->connectStatus;
+        return;
+    }
+
     /* Send the request */
     UA_UInt32 requestId = 0;
     UA_StatusCode retval = sendRequest(client, request, requestType, &requestId);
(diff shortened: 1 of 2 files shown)


jpfr:
This might be resolved via #8040.
There we have a setter method where a bad conbectStatus is always combined with a closeSecureChannel.

jpfr:
Is this still an issue?
If yes, it would be great to have a reproducer in the unit tests.
The coding agents could come up with one.

cartercanedy:
Hey @jpfr, haven't had a chance to take a look. Hopefully I'll have some time this coming week

PR #8054, opened 2026-05-19
fix(server): delete non-activated sessions when their SecureChannel closes

andreasebner:
Per spec a server must automatically delete any session that was created but never activated when the SecureChannel it was created on closes. The previous code detached all sessions unconditionally, leaving non-activated ones in the session list until timeout — exhausting maxSessions in CI scenarios with repeated failed-credential attempts.

Activated sessions are still only detached (enabling session transfer to a new channel). Non-activated sessions are now removed immediately.


jpfr:
A good change.
Please use UA_SHUTDOWNREASON_ABORT as the shutdown reason.

PR #8058, opened 2026-05-20
[Review] rejectedList modification API

Akul123:
Add rejectedList modification API
- OPC UA Part 6 defines the rejected certificates store as a certificate
store that applications may read and change. This change adds explicit
rejectedList modification support to UA_CertificateGroup, following the
same plugin interface pattern already used for trusted certificates.
- The new API adds following functions:
    setRejectedList
    removeFromRejectedList
- Implementations are provided for both mbedTLS and OpenSSL certificate
stores. Tests were added to verify setting, retrieving, removing and
memory ownership behavior for rejected certificates. Also 1 test for
checking filestore persistence between server restart is added to
confirm that server correctly loads rejected certificates on restart.


mcpat-pilz:
@Akul123 could you please provide some more details about where in Part 6 a modification API is proposed/mandated?

The only thing I could find is [this regarding "RejectedCertificateStore"](https://reference.opcfoundation.org/specs/OPC-10000-6/annex-e):

> The location of the shared CertificateStore containing the Certificates of applications which were rejected.

>
> Applications allow this value to be read or changed.

>
> Applications add the DER encoded Certificate into this store whenever it rejects a Certificate because it is untrusted or if it  failed one of the validation rules which can be suppressed (see Clause [E.6](https://reference.opcfoundation.org/specs/OPC-10000-6/e-6)).

>
>Applications do not add a Certificate to this store if it was rejected for a reason that cannot be suppressed (e.g. Certificate revoked).

Yes it states that this value "to be read or changed". But "change" does not necessarily mean to "set" a complete rejected list or remove a certificate from it.

The usual workflow is the following:
1. The application rejects a connection attempt from an untrusted client and places the certificate into the list of rejected certificates.
2. Another client may than obtain the list of rejected certificates (an operation that is already available), check the desired certificate and place it into the list of trusted certificates.
3. The original client is now trusted and can establish a connection.

As the list of rejected certificates (currently) only serves an informational purpose and not a semantical one (as opposed to a CRL) it is also not required to discard the now-trusted certificate from the list of rejected certificates. And further the number of certificates in this list can be limitted to avoid disk space exhaution and to have some kind of rotation.

So could you please explain what use case you have here?

Akul123:
Hi @mcpat-pilz,

Thanks for the detailed answer. My idea is to allow certificates to be moved between the trusted and untrusted/rejected stores in both directions.
Currently, the rejectedList keeps growing until maxRejectedListSize is reached, at which point the entire rejectedList is cleared. Since both the trusted and rejected certificate stores are described as being possible to “read or change,” I was thinking of adding the same API support for the rejected certificate store. That would make my original use case much easier to implement.
From the usual workflow, it seems that the rejected list and trusted list may contain the same certificate. However, there does not seem to be a way to explicitly reject a certificate that was previously trusted. The only option I currently see is to revoke trust by removing it from the trusted store.

Akul123:
@mcpat-pilz can you please suggest way forward? Would you like me to modify commit, if yes in which way? Or to close the PR?

mcpat-pilz:
It is not up to me to decide. But my oppinion is, that such an API is not required. OPC UA does not specify the rejected list to be evaluated by the server when determining trust so placing additional certificates in there is honestly of no use.

Akul123:
Hi @mcpat-pilz in that case I will close PR. Thanks for the response.

Akul123:
Equal handling of trusted and rejected certificates is not required in OPC UA API.

PR #8068, opened 2026-05-26
[Review] Fix coverity issues and compiler warnings

tomitkl:
Fix some issues reported by Coverity as high impact:
 - Memory leaks in error cases
 - top decremented from zero in xml_tokenize
 - Handle error (-1) from mbedtls_pk_write_key_der instead of casting the return value directly to size_t

And couple of compiler warnings:
 - C4702: unreachable code in the end of the function
 - C4456: declaration of 'pos' hides previous local declaration


jpfr:
The changes are all good.
Please add our typical pattern to the commit title.

- fix(core): ...
- fix(arch): ...
- fix(server): ...

This allows us to keep track of what is happending.
Details are in the CONTRIBUTING.md file.

tomitkl:
Updated the commit titles

jpfr:
Thanks! Merged.

PR #8076, opened 2026-06-02
[Review] Coverity issue fixes

tomitkl:
Fixes for some medium impact Coverity issues reported in our internal Coverity runs:

- Check UA_NODESTORE_GET and getCertGroup return values in ua_server_ns0_gds.c like elsewhere (Deference null return value)
- Check return value of UA_realloc in append_operand (Deference null return value)
- Check return value of UA_ftell in readFileToByteString to avoid -1 to be cast to size_t for buffer allocation (Improper use of negative value)
- Add some asserts to protect against Division or module by zero issues in ua_securechannel_crypto.c

Are the asserts enough in ua_securechannel_crypto.c? At least there are some other asserts in the same functions.

### DIFF
--- a/plugins/crypto/ua_filestore_common.c
+++ b/plugins/crypto/ua_filestore_common.c
@@ -38,7 +38,12 @@ readFileToByteString(const char *const path, UA_ByteString *data) {
 
     /* Get the file length, allocate the data and read */
     UA_fseek(fp, 0, UA_SEEK_END);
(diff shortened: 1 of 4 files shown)


jpfr:
Thanks, merged.
We won't be perfect with the assertions. Feel free to add those that make sense and/or capture an important invariance or assumptions in the code.

PR #8078, opened 2026-06-02
build(core): disable new clang default warnings which error out

bluhm:


jpfr:
Which clang versions are impacted?
Maybe we can remove the warnings before this clang version becomes too mainstream.

bluhm:
OpenBSD clang version 22.1.6
Target: amd64-unknown-openbsd7.9

In file included from /usr/ports/pobj/open62541-1.3.17/open62541-1.3.17/src/ua_types.c:19:
/usr/ports/pobj/open62541-1.3.17/build-amd64/src_generated/open62541/types_generated.h:204:1: error: '__COUNTER__' is a C2y extension [-Werror,-Wc2y-extensions]
  204 | UA_STATIC_ASSERT(sizeof(UA_NodeClass) == sizeof(UA_Int32), enum_must_be_32bit);
      | ^
/usr/ports/pobj/open62541-1.3.17/open62541-1.3.17/include/open62541/architecture_definitions.h:163:42: note: expanded from macro 'UA_STATIC_ASSERT'
  163 |     } UA_CTASTR(static_assertion_failed_,__COUNTER__)
      |                                          ^

/usr/ports/pobj/open62541-1.3.17/open62541-1.3.17/plugins/crypto/openssl/ua_pki_openssl.c:43:16: error: implicit conversion when returning 'void *' from a function with result type 'const unsigned char *' is not permitted in C++ [-Werror,-Wimplicit-void-ptr-cast]
   43 |         return (NULL);
      |                ^~~~~~

jpfr:
I just added a few minor commits on top of the 1.5 branch.
It compiles clean on OpenBSD 7.9 (default Clang) and the unit tests run with encryption enabled (LibreSSL).

I see you are the port maintainer: https://openports.eu/ports/misc/open62541
How about doing an update to 1.5?

We still release security fixes for 1.3. But overall this is pretty outdated.

jpfr:
By the way, I recommend calling the port libopen62541-1.5 and doing separetely libopen62541-1.5-dev.
The lib prefix shows this is a library, not an application.
And with the SONAME in the .so we can install multiple versions of the library in parallel.
Only the header files will conflict if we do 1.3/1.4/1.5 in parallel.

bluhm:
Usually we only have one version of a port in OpenBSD.  Otherwise maintaining effort gets even higher.  The clear way would be to upgrade.  But 1.4 is not API compatible and updating requirements memory management changes for every version.  I need https://github.com/bluhm/p5-opcua-open62541 as a Perl wrapper.  I managed to support the new logging API.  But what breaks my regression test is the different behavior regarding signals.  There was a change in 1.4, that is when I tried last time.  I know that I have to invest time to adapt to the current code at some point.  But for now the old library works well enough.

jpfr:
Please rebase this PR on the current 1.3.
1.3 received security fixes and CI improvements.

I understand that your application is locked to 1.3.
But keeping the OpenBSD port at 1.3 is not a good idea.

Yes, the API changed. But for good reason changes are generally slow.
The EventLoop change in particular has been widely digested by now.
A coding agent would probably one-shot the fallout from the upgrade to 1.5.

I suggest you move your application to non-packaged "oldstable" version and keep OpenBSD aligned with "stable".
If you need a package, this is exactly why I suggested the libopen62541-1.5 naming, so multiple versions of the library can coexist.

Issue #5882, opened 2023-07-18
Server constantly updates timestamp of monitored items with data change trigger = UA_DATACHANGETRIGGER_STATUSVALUETIMESTAMP


paddor:
## Background Information / Reproduction Steps

When I connect UaExpert (v1.6.3) to a Open62541 Server (v1.3.6) and drag a variable (e.g. `i=2266 [Server_ServerStatus_BuildInfo_BuildDate]`) into the Data Access View and change the Monitored Item Settings so that Data Change Trigger is `Status/Value/Timestamp`, both Source Timestamp and Server Timestamp keep getting updated.

This also happens to my own variables (added dynamically, not imported from a nodeset). I tried with a writable Bool variable node. Interestingly, as soon as I write into the variable using UaExpert, the updating of the timestamps stops.

Using Ubuntu Linux 22.04.

Used CMake options:

```
-DUA_NAMESPACE_ZERO=FULL -DUA_MULTITHREADING=100 -DCMAKE_BUILD_TYPE=Debug -DUA_DEBUG=1 -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF -DCMAKE_C_FLAGS_DEBUG='-fPIC -O0 -g'
```

paddor:
To reproduce, I compiled the library:

```
git clone https://github.com/open62541/open62541.git
cd open62541
git submodule update --init --recursive
mkdir build && cd build
cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug  -DUA_ENABLE_DIAGNOSTICS=ON -DUA_NAMESPACE_ZERO=FULL ..
make
sudo make install
sudo ldconfig
```

And then just run the example program https://github.com/open62541/open62541/blob/master/examples/tutorial_server_variable.c.
```
gcc -std=c99 tutorial_server_variable.c -lopen62541 -o tutorial_server_variable
./tutorial_server_variable
```

Connect with UaExpert, expand `ServerDiagnostics`, drag and drop the `BuildInfo`, set to `Status/Value/Timestamp`, and watch the timestamp continually get updated.

paddor:
More findings:

* `detectVAlueChange()` seems to behave correctly, but the `value.sourceTimestamp` just keeps getting updated
* even read operations update the timestamps (Refresh button in UaExpert)
* it's not just a bug in UaExpert. WireShark shows the variables repeatedly in `PublishResponse` messages.
[...]

PR #8082, opened 2026-06-04
fix(server): opt-in stable source timestamps for data source variables

paddor:
## Context

Reopening (from the application side) a problem first reported in #5882.

When a data source read callback does not set `hasSourceTimestamp`, the server fills in `dateTime_now()` on every read (`readValueAttributeComplete`, line 283 in `ua_services_attribute.c`). With `UA_DATACHANGETRIGGER_STATUSVALUETIMESTAMP` subscriptions this causes a flood of spurious notifications even when neither value nor status have changed.

This has bitten us again, this time specifically with **data source variables** (not just nodeset-imported ones). Our OPC UA server exposes ~80 data source variables to an external SCADA system. Each read callback returns the current process value but does not set a source timestamp (because the source has no meaningful timestamp to provide). With the current behavior, every sampling interval produces a notification, flooding the SCADA client.

## Fix

Add `UA_ServerConfig::stableDataSourceTimestamps` (default `false`). When enabled, the server caches the last value and status per data-source variable node and only updates the source timestamp when one of them actually changes.

This is fully backward compatible: the default behavior is unchanged. Applications that want stable timestamps opt in with a single config flag.

### Changes

- **`include/open62541/server.h`**: new `UA_Boolean stableDataSourceTimestamps` field on `UA_ServerConfig`
- **`include/open62541/plugin/nodestore.h`**: cache fields on `UA_VariableNode` (`cachedSourceValue`, `cachedSourceStatus`, etc.)
- **`src/server/ua_services_attribute.c`**: in `readValueAttributeComplete`, compare against cached state before stamping
- **`src/server/ua_nodes.c`**: cleanup in `UA_Node_clear`, deep copy in `UA_VariableNode_copy`

## Workaround (current)

We work around this in our application layer by caching value + status + sourceTimestamp in the data source read callback itself and setting `hasSourceTimestamp = true` only when something changes. This works but pushes a concern into every application that uses data source variables with timestamp-sensitive subscriptions. A server-level opt-in is cleaner.

## Test plan

- [ ] Existing unit tests pass (no behavior change with flag off)
- [ ] Manual verification with UaExpert: subscribe to a data source variable with StatusValueTimestamp trigger, confirm no spurious notifications when value is unchanged

### DIFF
--- a/include/open62541/plugin/nodestore.h
+++ b/include/open62541/plugin/nodestore.h
@@ -335,6 +335,14 @@ typedef struct {
                            * background. Only dynamic variables conserve source
                            * and server timestamp for the value attribute.
                            * Static variables have timestamps of "now". */
+
+    /* Cached source timestamp for data-source variables
+     * (UA_ServerConfig::stableDataSourceTimestamps). */
+    UA_Variant cachedSourceValue;
+    UA_StatusCode cachedSourceStatus;
+    UA_Boolean cachedSourceHasStatus;
+    UA_DateTime cachedSourceTimestamp;
+    UA_Boolean hasCachedSourceTimestamp;
 } UA_VariableNode;
 
 /**
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -2105,6 +2105,14 @@ struct UA_ServerConfig {
 
     UA_RuleHandling allowAllCertificateUris;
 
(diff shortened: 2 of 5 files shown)


paddor:
Probably not necessary. Callback can set its own stable `sourceTimestamp`.

PR #8088, opened 2026-06-07
test: extend unit-test coverage for core, server and client

andreasebner:
Adds unit tests covering previously-uncovered paths across the stack on top of 1.5.

**Core / types**
- Variant and DataType lookup, `UA_clear()` edge cases, utility functions
- Batch 2 & 3 coverage for high-impact uncovered functions
- Config, syslog, certificate-group and auditing coverage

**Server**
- `UA_Server_getConfig` returns a live pointer; attribute-read edge cases
- Node-context API and `findChildByBrowsename`
- Browse service error paths
- Async operations, NS0 diagnostics and session coverage

**Client**
- High-level API error paths
- `findServersOnNetwork` paging
- Client config and namespace-lookup paths

**Security (incl. fix)**
- Coverage tests for `plugins/crypto/`
- fix: heap-allocate the IP buffer for the SAN list to avoid a stack-use-after-scope (caught by the new tests)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,8 @@ tests/interop/node-opcua/package-lock.json
 
 # Local doc-review helper (not part of the project)
 gen_review_docs.sh
+
+# Coverage artifacts
+*.gcov
+*.gcda
+*.gcno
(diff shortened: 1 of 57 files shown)


jpfr:

PR #8099, opened 2026-06-10
fix(client): correct secure channel state output

neojaw:
The enum was changed in b19cc6d but the names weren't.


neojaw:
I don't think the failed checks are my fault.

jpfr:
Good find.
Please change the target of this PR to the 1.4 branch (already affected from what I see).
And rebase.

We will then merge/shepherd this upwards until master.

neojaw:
Done.

jpfr:
👍

PR #8105, opened 2026-06-11
[WIP] Fix RemoveCertificate Method

AdanRuoff:
This PR aims to fix two issues in the current implementation of the RemoveCertificate Method in ua_server_ns0_gds.c:

### 1. The temporary transaction created at the beginning is not cleaned up on errors.
Reproduction steps:
- compile open62541 with "examples"
- run the "server_encryption_filestore" example; populate the own certificate and TrustList so that you can connect with an external Client (e.g. UaExpert)
- From the Client, call RemoveCertificate with a non-existing or dummy Certificate Thumbprint parameter. This call will fail as expected with "Bad_InvalidArgument".
- From the Client, call RemoveCertificate a second time with a non-existing or dummy Certificate Thumbprint parameter.

**Expected behavior:**
Call fails with "Bad_InvalidArgument".

**Observed behavior:**
This call (and all subsequent Method calls that act on the TrustList) will now fail with "Bad_TransactionPending".

### 2. RemoveCertificate may not be used to remove issuers on which other certificates depend.
The OPC specification states for RemoveCertificate:

> This Method returns Bad_CertificateChainIncomplete if the Certificate is a CA Certificate needed to validate another Certificate in the TrustList.

(https://reference.opcfoundation.org/specs/OPC-10000-12/v1.05.07/7.8.2.7)

However, the current implementation allows removing issuers on which other certificates inside the TrustList depend.

Reproduction steps:
- compile open62541 with "examples"
- Populate the own certificate and TrustList:
  - provide a certificate and private key for the Server instance ("own" certificate)
  - "list of trusted certificates": certificate of the Client you are using for the test
  - "list of issuer certificates":
    - "Root CA" certificate (used to sign "Intermediate CA" certificate)
    - "Intermediate CA" certificate (signed by "Root CA" certificate)
  - "list of issuer CRLs":
    - empty CRL for "Root CA"
    - empty CRL for "Intermediate CA"
- run the "server_encryption_filestore" example
- From the Client, call RemoveCertificate with the following parameters:
  - Thumbprint: thumbprint (SHA-1 hash) of the "Root CA" certificate
  - isTrustedCertificate: false (unticked in UaExpoert)

**Expected behavior:**
Call fails with "Bad_CertificateChainIncomplete" and certificate is still inside the TrustList.

**Observed behavior:**
Call succeeds with "Good" and certificate (plus its CRL) has been removed from the TrustList.

### DIFF
--- a/include/open62541/plugin/certificategroup.h
+++ b/include/open62541/plugin/certificategroup.h
@@ -115,6 +115,16 @@ UA_CertificateUtils_checkKeyPair(const UA_ByteString *certificate,
 UA_EXPORT UA_StatusCode
 UA_CertificateUtils_checkCA(const UA_ByteString *certificate);
 
+/* Checks if "issuer" is the issuer of "certificate".
+ * Returns the result in "isIssuer".
+ *
+ * Returns UA_STATUSCODE_BADINVALIDARGUMENT if any parameter is NULL or when
+ * issuer or certificate is no valid x509 certificate. */
+UA_EXPORT UA_StatusCode
+UA_CertificateUtils_isIssuer(const UA_ByteString *issuer,
+                             const UA_ByteString *certificate,
+                             UA_Boolean *isIssuer);
+
 /* Decrypt a private key in PEM format using a password. The output is the key
  * in the binary DER format. Also succeeds if the PEM private key does not
  * require a password or is already in the DER format. The outDerKey memory is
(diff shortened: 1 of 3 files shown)


AdanRuoff:
Currently, this fixes only the mbedTLS plugin. I will port the issuer checking code to OpenSSL as time permits 😄 .

AdanRuoff:
Note: This PR is intended as a replacement for PR #7664.

jpfr:
Give us a trigger for an eventual review.

PR #8112, opened 2026-06-16
fix(server): reject negative TrustList read lengths

MrAlaskan:
This PR updates the TrustList `Read` implementation to reject negative `length` arguments.

Before this change, negative `Int32` values were cast to `size_t` during the boundary checks in `readTrustList()`. As a result, requests such as `length = -1` could be treated like oversized reads and then clamped to the remaining file size, instead of being rejected as invalid input.

The OPC UA FileType `Read` definition specifies that only positive values are allowed for the `Length` argument. This change aligns the current behavior more closely with that definition by explicitly rejecting negative values and returning `UA_STATUSCODE_BADINVALIDARGUMENT`.

A regression test is included to verify that calling the TrustList `Read` method with `length = -1` no longer returns file data and instead reports the invalid request.


jpfr:
Merged. Thanks.

PR #8113, opened 2026-06-16
fix(server): validate AddReferences target node class

MrAlaskan:
This PR adds validation for `targetNodeClass` in AddReferences when the target node is local.

Previously, AddReferences could succeed even if the provided `targetNodeClass` did not match the actual NodeClass of the target node. This change returns `UA_STATUSCODE_BADNODECLASSINVALID` for such requests.

The OPC UA AddReferences definition expects `targetNodeClass` to match the TargetNode NodeClass for local targets. To avoid changing existing helper paths that do not set this field, the check is only applied when a non-zero `targetNodeClass` is provided.

A regression test is included to cover this case.

### DIFF
--- a/src/server/ua_services_nodemanagement.c
+++ b/src/server/ua_services_nodemanagement.c
@@ -2317,6 +2317,17 @@ Operation_addReference(UA_Server *server, UA_Session *session, void *context,
             *retval = UA_STATUSCODE_BADTARGETNODEIDINVALID;
             return;
         }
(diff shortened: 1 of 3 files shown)


jpfr:
Thanks for this.
I see a lot of CI runs break with
```
2026-06-16T11:18:39.7567573Z The following tests FAILED:
2026-06-16T11:18:39.7567946Z 	122 - check_client_highlevel (Failed)
```

Maybe there are some wrong assumptions in check_client_highlevel.

MrAlaskan:
The wrong assumptions have been fixed.

PR #8118, opened 2026-06-16
test(interop): fail when no .NET ECC test actually runs

andreasebner:
Cover the new ECC_nistP256_AesGcm/_ChaChaPoly policies in C server, C client and .NET tests and fail instead of silently skipping when no ECC session is established.


jpfr:

Issue #8062, opened 2026-05-21
Access Level checks too restrictive for writing a value


mcpat-pilz:
## Description

PR #7768 adds checks for the StatusWrite and TimestampWrite masks. Nevertheless, the checks are incomplete according to [https://reference.opcfoundation.org/specs/OPC-10000-3/8.57](https://reference.opcfoundation.org/specs/OPC-10000-3/8.57):

> StatusWrite: Indicates if the current StatusCode of the value is writeable (**0 means only StatusCode Good is writeable**, 1 means any StatusCode is writeable).

and

> TimestampWrite: Indicates if the current SourceTimestamp is writeable (**0 means only null timestamps are writeable**, 1 means any timestamp value is writeable).

So before checking the AccessLevel it must be checked, if a status is given **and its value != Good** as well as whether a source timestamp is given **and its value != Null**. The bold parts are what is missing in PR #7768.

PR #8122, opened 2026-06-18
[Review] Fix the checks for writing the StatusCode and SourceTimestamp of a value

AdanRuoff:
Commit 59d47fc introduced checks for writing the StatusCode and SourceTimestamp of a Variable. However, these checks are currently implemented too restrictive, as they only allow writing these fields when the corresponding AccessLevelType bits are set.

The OPC specification, however, states for the AccessLevelType bits:
- `StatusWrite`: "Indicates if the current StatusCode of the value is writeable (*0 means only StatusCode Good is writeable*, 1 means any StatusCode is writeable)."
- `TimestampWrite`: "Indicates if the current SourceTimestamp is writeable (*0 means only null timestamps are writeable*, 1 means any timestamp value is writeable)."

(see https://reference.opcfoundation.org/specs/OPC-10000-3/v1.05.06/8.57)

Thus, according to the specification, writing a StatusCode of "Good" and a "null" timestamp must always be possible, even when the `StatusWrite` and `TimestampWrite` are not set.

This PR completes the checks such that they behave in an OPC-conforming way and fixes issue #8062.

### DIFF
--- a/src/server/ua_services_attribute.c
+++ b/src/server/ua_services_attribute.c
@@ -1780,8 +1780,12 @@ copyAttributeIntoNode(UA_Server *server, UA_Session *session,
                 retval = UA_STATUSCODE_BADUSERACCESSDENIED;
                 break;
             }
-            /* Writing the StatusCode requires the StatusWrite bit */
-            if(wvalue->value.hasStatus) {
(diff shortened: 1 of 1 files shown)


jpfr:
Thanks. Merged.

PR #8123, opened 2026-06-18
fix(deps): fix dtoa buffer offsets

neojaw:
`buffer` is already iterated to the current write position, writing at `pos` may skip a byte. Current code effectively prints negative infinity as `-?in` with `?` being memory garbage.

### DIFF
--- a/deps/dtoa.c
+++ b/deps/dtoa.c
@@ -351,7 +351,6 @@ unsigned dtoa(double d, char* buffer) {
     if(sign) {
         buffer[0] = '-';
         pos++;
-        buffer++;
     }
 
     if(exponent == ((1u << exponent_bits) - 1u)) {
@@ -368,5 +367,5 @@ unsigned dtoa(double d, char* buffer) {
     char digits[18];
     memset(digits, 0, 18);
     unsigned ndigits = grisu2(bits, digits, &K);
-    return pos + emit_digits(digits, ndigits, buffer, K, sign);
+    return pos + emit_digits(digits, ndigits, &buffer[pos], K, sign);
 }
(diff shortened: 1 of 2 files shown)


jpfr:
Good find.
There might be a bug in the NAN case also.
Where it puts in "-nan" and returns length 3.

Please check and add a unit test for the cases you are solving.

mcpat-pilz:
Looking at the code it would be best not to increment "buffer" at all but to keep the indexing with the "pos". Of course also in the NAN case.

neojaw:
You're right about the "-nan", I fixed that as well.

Whether you increment `buffer` or index with `pos` is a matter of taste, you just have to make sure to stick to one. I went this way because it required only one changed line while the other would have been three. I can do it differently if you prefer.

While writing the tests, I noticed that the `NAN` macro that is used everywhere locally actually produces *negative* NaN on byte level, so I'm using it with a minus sign. Does it make sense to change the other ones as well (in a separate PR) or would you prefer to refactor it out to a global place?
Edit: Uh, some compilers disagree, apparently.

neojaw:
Seems like in tcc there is no such thing as positive NaN. Both the macro from `<math.h>` and the manually defined one, with and without a minus sign in front, produce negative NaN bit pattern.

Also, gcc and clang behave differently on the sign for the manually defined macro, but identically for the one from `<math.h>`.

Hmm... as far as I understand, there is no explicit definition of positive/negative NaN. Any number that matches the bit pattern is NaN and applications/libraries are not required to make a difference. Maybe it would be the cleanest solution for `dtoa` to give the same output in all cases. Alternatively we could allow both outputs in the tests. Any preference?

neojaw:
Changed the output to "nan" for any type of NaN. Also using indexing because now it makes more sense than iterating the buffer.
The failing check looks unrelated.

jpfr:
Thanks for this. Merged.

PR #8130, opened 2026-06-22
feat(server): enum datatypedefinition property fallback

andreasebner:
Reading the DataTypeDefinition attribute of an Enumeration currently only works for enums that are compiled into the C code. Enums that come in through a NodeSet at runtime have no compiled members (membersSize == 0), so the server answers BadAttributeIdInvalid — which trips up CTT and clients like UaExpert, even though Part 3 says the attribute is mandatory for every Enumeration.

This adds a small fallback for that case: when there are no compiled members, the server builds the EnumDefinition from the node's EnumValues property (or EnumStrings if that's all there is, using implicit 0..N-1 values). The existing compiled-enum path is untouched, and a node with neither property still returns BadAttributeIdInvalid as before.


jpfr:

PR #8135, opened 2026-06-22
fix(server): return runtime auditing state in Server.Auditing

MrAlaskan:
This PR updates `Server.Auditing` to reflect the runtime auditing configuration.

Previously, `readAuditing()` always returned `false`, even when auditing was enabled in the server configuration. This could make the standard `Server.Auditing` variable report a disabled auditing state while the server was actually configured to emit audit events.

The OPC UA Part 5 definition of `Server.Auditing` says that it is a Boolean specifying whether the Server is currently generating audit events, and that it shall be set to `TRUE` if the Server generates audit events, otherwise `false`. This change returns `server->config.auditingEnabled` instead of a constant `false`, so the exposed state matches the runtime configuration more closely.

A regression test is included to verify that `Server.Auditing` returns `false` with the default configuration and `true` after auditing is enabled.


jpfr:
Looks good. Thanks!

PR #7983, opened 2026-04-09
[REVIEW] fix(server): require Browse permission for common attribute reads

StalderT:
Possible fix for #7935

Issue #7935, opened 2026-03-25
Reading attributes seems to be too permissive


mcpat-pilz:
Currently, most node attributes (like NodeId, NodeClass, BrowseName, DisplayName, Description, etc.) can be read without any permission being checked.

Looking into the Spec at [PermissionType](https://reference.opcfoundation.org/Core/Part3/v105/docs/8.55) there is the following line for the "Browse" permission:

> The [Client](https://reference.opcfoundation.org/search/462?t=Client) is allowed to see the references to and from the [Node] https://reference.opcfoundation.org/search/462?t=Node).
>
> This implies that the [Client](https://reference.opcfoundation.org/search/462?t=Client) is able to Read to [Attributes] https://reference.opcfoundation.org/search/462?t=Attributes) other than the [Value](https://reference.opcfoundation.org/search/462?t=Value) or the [RolePermissions](https://reference.opcfoundation.org/search/462?t=RolePermissions) [Attribute](https://reference.opcfoundation.org/search/462?t=Attribute).
>
> This [Permission](https://reference.opcfoundation.org/search/462?t=Permission) is valid for all [NodeClasses](https://reference.opcfoundation.org/search/462?t=NodeClasses).

So I basically understand, that reading the common attributes (except RolePermissions and Value) should only be possible if the current user has the Browse permission on the node. Do I miss something?

PR #8144, opened 2026-06-23
fix(server): Apply allowBrowseNode access control in TranslateBrowsePathsToNodeIds

o6-nbeier:
The TranslateBrowsePathsToNodeIds service did not call the allowBrowseNode access-control callback when resolving path elements or terminal targets. A client denied direct Browse access to a node could still use that node as a waypoint or receive its NodeId as a path translation result, disclosing hidden nodes and their children.

Add the same allowBrowseNode gate already applied in the Browse service to both walkBrowsePathElement (intermediate nodes) and
Operation_TranslateBrowsePathToNodeIds (terminal nodes).

See: open62541-SA-2026-0004

### DIFF
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -436,6 +436,13 @@ receiveResponse(UA_Client *client, void *response, const UA_DataType *responseTy
             UA_LOG_WARNING_CHANNEL(&client->config.logger, &client->channel,
                                    "Receiving the response failed with StatusCode %s",
                                    UA_StatusCode_name(retval));
+            /* The receive itself can succeed while the channel was closed
+             * remotely (e.g. an ERR message during the handshake). Forward the
+             * detailed status code captured from the close so the reconnect is
+             * not silent. */
+            if(retval == UA_STATUSCODE_GOOD)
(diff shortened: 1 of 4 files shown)


mcpat-pilz:
The TranslateBrowsePathsToNodeIds now seem to behave as expected. Still I want to point out, that the Browse-Service itself is inconsistent to this new behaviour.

Imagine the following tree structure A -> B -> C -> D where Node C is not browsable. The TranslateBrowsePathsToNodeIds will return the following:

* Request "A :: B" -> NodeId of B
* Request "A :: B :: C" -> BadNoMatch
* Request "A :: B :: C :: D" -> BadNoMatch

Now using the Browse Service the following happens:

* Request "Browse A" -> [B]
* Request "Browse B" -> [A, C]
* Request "Browse C" -> BadUserAccessDenied
* Request "Browse D" -> [C]

Hence Node "C" is not really hidden here even though one would expect that. This may relate to #7935 or #7983.

jpfr:
@mcpat-pilz , we will touch that as part of the RBAC access control.
What can happen is that a node appears in browse, but is itself not readable.
I would not touch that in a patch release. Could break some code that depends on the exact behavior of browse,

Issue #8124, opened 2026-06-19
UA_NAMESPACE_ZERO=MINIMAL behaves like NONE and never creates the minimal server object


Ubberx:
### Description

When building with `UA_NAMESPACE_ZERO=MINIMAL`, the server appears to follow the
same initialization path as `UA_NAMESPACE_ZERO=NONE`.

As a result, the `minimalServerObject()` path in `ua_server_ns0.c` is never
reached during normal server startup.

### Relevant code

In `CMakeLists.txt`, both `MINIMAL` and `NONE` disable
`UA_GENERATED_NAMESPACE_ZERO`:

```cmake
if(UA_NAMESPACE_ZERO STREQUAL "MINIMAL" OR UA_NAMESPACE_ZERO STREQUAL "NONE")
    set(UA_GENERATED_NAMESPACE_ZERO OFF)
else()
    set(UA_GENERATED_NAMESPACE_ZERO ON)
endif()
```

Then in `src/server/ua_server.c`, namespace 0 initialization is selected only
based on `UA_GENERATED_NAMESPACE_ZERO`:

```c
#ifdef UA_GENERATED_NAMESPACE_ZERO
    res = initNS0(server);
#else
    res = initNS0_dataSources(server);
#endif
```

Therefore, for `UA_NAMESPACE_ZERO=MINIMAL`, the server calls:

```c
initNS0_dataSources(server);
```

which is the same path used for `NONE`.

However, the minimal namespace zero object is only created inside `initNS0()`
when `UA_GENERATED_NAMESPACE_ZERO` is not defined:

```c
#ifdef UA_GENERATED_NAMESPACE_ZERO
    retVal |= namespace0_generated(server);
#else
    retVal |= minimalServerObject(server);
#endif
```

This branch seems unreachable during normal server startup, because `initNS0()`
is only called when `UA_GENERATED_NAMESPACE_ZERO` is defined.

### Expected behavior

`UA_NAMESPACE_ZERO=MINIMAL` should initialize a minimal namespace zero, including
calling `minimalServerObject(server)`.

`UA_NAMESPACE_ZERO=NONE` should skip namespace zero creation and only
configure/connect data sources for externally preloaded namespace zero nodes.

### Actual behavior

`UA_NAMESPACE_ZERO=MINIMAL` behaves like `UA_NAMESPACE_ZERO=NONE`.

The `minimalServerObject(server)` function is not called.

### Suggested direction

The startup logic likely needs to distinguish between:

[...]

PR #8147, opened 2026-06-26
Fix: UA_NAMESPACE_ZERO=MINIMAL behaves like NONE and never creates the minimal server object #8124

Ubberx:
## Related Issue

UA_NAMESPACE_ZERO=MINIMAL behaves like NONE and never creates the minimal server object
https://github.com/open62541/open62541/issues/8124

## Problem Description

`UA_NAMESPACE_ZERO` is the user-facing build option that selects how complete
Namespace 0 should be:

```text
NONE, MINIMAL, REDUCED, FULL
```

with the following expected behavior:

| `UA_NAMESPACE_ZERO` value | Expected startup behavior |
| --- | --- |
| `FULL` | Create Namespace 0 and load the full generated nodeset. |
| `REDUCED` | Create Namespace 0 and load the reduced generated nodeset. |
| `MINIMAL` | Create Namespace 0 and add only the hand-written minimal `Server` object. |
| `NONE` | Do not create Namespace 0 during normal startup. Assume it is provided externally, for example by a ROM nodestore. |

Before the fix, `MINIMAL` and `NONE` were accidentally treated the same during
server startup.

The problem was in `src/server/ua_server.c`. The startup logic used
`UA_GENERATED_NAMESPACE_ZERO` to decide whether Namespace 0 should be created at
all:

```c
#ifdef UA_GENERATED_NAMESPACE_ZERO
    res = initNS0(server);
#else
    res = initNS0_dataSources(server);
#endif
```

This made the server behave like this:

| Mode | `UA_GENERATED_NAMESPACE_ZERO` | Old startup behavior |
| --- | --- | --- |
| `FULL` | ON | `initNS0(server)` |
| `REDUCED` | ON | `initNS0(server)` |
| `MINIMAL` | OFF | `initNS0_dataSources(server)` |
| `NONE` | OFF | `initNS0_dataSources(server)` |

That is wrong for `MINIMAL`.

`MINIMAL` still needs to call `initNS0(server)`. Inside `initNS0(server)`,
generated Namespace 0 is disabled, so the code takes the minimal branch:

```c
#ifdef UA_GENERATED_NAMESPACE_ZERO
    retVal |= namespace0_generated(server);
#else
    retVal |= minimalServerObject(server);
#endif
```

But because the old startup logic skipped `initNS0(server)` for `MINIMAL`,
`minimalServerObject(server)` was never reached during normal server startup.

## Applied Fix

The fix was to add a separate macro for the `NONE` case:

```c
UA_NAMESPACE_ZERO_NONE
```

This macro has a different meaning from `UA_GENERATED_NAMESPACE_ZERO`.

`UA_GENERATED_NAMESPACE_ZERO` answers:

> Should generated Namespace 0 code be compiled and called?

`UA_NAMESPACE_ZERO_NONE` answers:

> Should normal Namespace 0 creation be skipped because Namespace 0 is provided externally?

Separating these two questions fixes the `MINIMAL` vs `NONE` confusion.

### CMakeLists.txt

The existing generated-namespace logic stays unchanged:

```cmake
if(UA_NAMESPACE_ZERO STREQUAL "MINIMAL" OR UA_NAMESPACE_ZERO STREQUAL "NONE")
    set(UA_GENERATED_NAMESPACE_ZERO OFF)
else()
    set(UA_GENERATED_NAMESPACE_ZERO ON)
endif()
```

This remains correct:

| Mode | `UA_GENERATED_NAMESPACE_ZERO` |
| --- | --- |
| `FULL` | ON |
| `REDUCED` | ON |
| `MINIMAL` | OFF |
| `NONE` | OFF |

The fix adds a new `NONE`-specific CMake variable:

```cmake
if(UA_NAMESPACE_ZERO STREQUAL "NONE")
    set(UA_NAMESPACE_ZERO_NONE ON)
else()
    set(UA_NAMESPACE_ZERO_NONE OFF)
endif()
```

This creates the missing distinction:

| Mode | `UA_NAMESPACE_ZERO_NONE` |
| --- | --- |
| `FULL` | OFF |
| `REDUCED` | OFF |
| `MINIMAL` | OFF |
| `NONE` | ON |

### config.h.in

The new CMake variable is exported into the generated configuration header:

```c
#cmakedefine UA_NAMESPACE_ZERO_NONE
```

After CMake configures the build, C code can now distinguish `NONE` from
`MINIMAL`.

### ua_server.c

The startup decision was changed from checking whether Namespace 0 is generated
to checking whether Namespace 0 should be skipped entirely.

Old logic:

```c
#ifdef UA_GENERATED_NAMESPACE_ZERO
    /* Standard configuration: generate NS0 nodes at runtime */
    res = initNS0(server);
#else
    /* NONE configuration: NS0 pre-loaded by external nodestore (e.g., ROM).
     * Only connect data sources for dynamic values like ServerTime, ServerStatus, etc. */
    res = initNS0_dataSources(server);
#endif
```

New logic:

```c
#ifdef UA_NAMESPACE_ZERO_NONE
    /* NONE configuration: NS0 pre-loaded by external nodestore (e.g., ROM).
     * Only connect data sources for dynamic values like ServerTime, ServerStatus, etc. */
    res = initNS0_dataSources(server);
#else
    /* Standard configuration: generate NS0 nodes at runtime */
    res = initNS0(server);
#endif
```

This means:

| Mode | New startup function |
| --- | --- |
| `FULL` | `initNS0(server)` |
| `REDUCED` | `initNS0(server)` |
| `MINIMAL` | `initNS0(server)` |
| `NONE` | `initNS0_dataSources(server)` |

### ua_server_ns0.c

No change is needed in the existing `initNS0(server)` branch:

```c
#ifdef UA_GENERATED_NAMESPACE_ZERO
    /* Load nodes and references generated from the XML ns0 definition */
    retVal |= namespace0_generated(server);
#else
    /* Create a minimal server object */
    retVal |= minimalServerObject(server);
#endif
```

Once `MINIMAL` reaches `initNS0(server)`, this branch works correctly:

| Mode | `initNS0(server)` behavior |
| --- | --- |
| `FULL` | Calls `namespace0_generated(server)` generated from the full XML nodeset. |
| `REDUCED` | Calls `namespace0_generated(server)` generated from the reduced XML nodeset. |
| `MINIMAL` | Calls `minimalServerObject(server)`. |
| `NONE` | Does not call `initNS0(server)`. |


andreasebner:
Thank's for raising the issue. The fix was made independently in #8231

Issue #5131, opened 2022-05-18
How to get the sampling interval that the client has set for subscription/monitored items?


serega1103:
## Description
Hi! A simple question from a beginner.

I designed the server so that:
1. The server gets new values for node variables from the database only for the Monitored Items list, and not for all node variables on the server (when the client creates new MonitoredItems, my Monitored Items list on the server is formed thanks to config->monitoredItemRegisterCallback);
2. The sampling interval for new node variable values from the database is set in the my "settings.json" file.

I want to use the sampling interval that the client set in the subscription in order to use it to get new values of node variables from the database.

How do I get the sampling interval on the server side that the client has set for the subscription and included Monitored Items?

## Background Information / Reproduction Steps
Used CMake options:
```bash
cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_AMALGAMATION=ON -G "Visual Studio 17 2022" -A Win32 -S .. -B "build32"
```

## Checklist
 - [x] open62541 Version: v1.2
 - [ ] Other OPC UA SDKs used (client or server): 
 - [x] Operating system: Windows 10
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached
 - [ ] Wireshark network dump attached
 - [ ] Self-contained code example attached
 - [ ] Critical issue

serega1103:
I am interested in the possibility of getting Monitored Item Sampling Interval on the Server side.

paperino1983:
Looking in the code I suppose that this is not possible today (looking code in tag `v1.4.12`)
In order to propagate the service and service's data specifics (for instance `sampling rate` for the monitored items service) a non trivial change shall be done.

PR #8154, opened 2026-06-29
feat(monitored items): provide sampling interval to the application

paperino1983:
### Summary
Currently (up to the openstack v1.5.3) `UA_ServerConfig` structure allows the application to be notified when a node is added (or removed) to a monitor items list. The signature is as follows :

```c
    /* Register MonitoredItem in Userland
     *
     * @param server Allows the access to the server object
     * @param sessionId The session id, represented as an node id
     * @param sessionContext An optional pointer to user-defined data for the
     *        specific data source
     * @param nodeid Id of the node in question
     * @param nodeidContext An optional pointer to user-defined data, associated
     *        with the node in the nodestore. Note that, if the node has already
     *        been removed, this value contains a NULL pointer.
     * @param attributeId Identifies which attribute (value, data type etc.) is
     *        monitored
     * @param removed Determines if the MonitoredItem was removed or created. */
    void (*monitoredItemRegisterCallback)(UA_Server *server,
                                          const UA_NodeId *sessionId,
                                          void *sessionContext,
                                          const UA_NodeId *nodeId,
                                          void *nodeContext,
                                          UA_UInt32 attibuteId,
                                          UA_Boolean removed);
```
here in the [HTML documentation](https://open62541.org/doc/1.5/server.html#server-configuration)

The current callback signature does not allows to notify the application about the sampling interval the node is monitored. This is a missing information that could be useful at application level.

So, the main idea is to propose a second API, called `monitoredItemRegisterCallbackEx` which acts as extension of the original one, providing this information to the application.

### Why this PR
This PR try to resolve some issues, like this: #5131 and the intent is to collect your feedback in order to be, I think, better implemented

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -2418,6 +2418,32 @@ struct UA_ServerConfig {
                                           void *nodeContext,
                                           UA_UInt32 attibuteId,
                                           UA_Boolean removed);
+
+    /* Register MonitoredItem in Userland with sampling interval report
+     *
+     * Extended version of monitoredItemRegisterCallback
+     *
+     * @param server Allows the access to the server object
+     * @param sessionId The session id, represented as an node id
+     * @param sessionContext An optional pointer to user-defined data for the
+     *        specific data source
+     * @param nodeid Id of the node in question
+     * @param nodeidContext An optional pointer to user-defined data, associated
+     *        with the node in the nodestore. Note that, if the node has already
+     *        been removed, this value contains a NULL pointer.
+     * @param attributeId Identifies which attribute (value, data type etc.) is
+     *        monitored
+     * @param samplingInterval Sampling interval of the node id
+     * @param removed Determines if the MonitoredItem was removed or created. */
+
+    void (*monitoredItemRegisterCallbackEx)(UA_Server *server,
+                                            const UA_NodeId *sessionId,
+                                            void *sessionContext,
+                                            const UA_NodeId *nodeId,
+                                            void *nodeContext,
+                                            UA_UInt32 attibuteId,
+                                            UA_Double samplingInterval,
+                                            UA_Boolean removed);
 #endif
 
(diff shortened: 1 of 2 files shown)


jpfr:
You can get the sampling interval via the ApplicationNotification callback.
The callback you mentioned is deprecated and will be removed.

https://github.com/open62541/open62541/blob/9f715bbb33c77bf256144485742bfbb6162e2066/include/open62541/server.h#L205
https://github.com/open62541/open62541/blob/9f715bbb33c77bf256144485742bfbb6162e2066/include/open62541/common.h#L360

paperino1983:
> You can get the sampling interval via the ApplicationNotification callback. The callback you mentioned is deprecated and will be removed.
> 
> https://github.com/open62541/open62541/blob/9f715bbb33c77bf256144485742bfbb6162e2066/include/open62541/server.h#L205
> 
> 
> https://github.com/open62541/open62541/blob/9f715bbb33c77bf256144485742bfbb6162e2066/include/open62541/common.h#L360

That's a great thing ! Thanks for your answer so. I will take a look.

paperino1983:
Hi Everyone, I took a look to the `ApplicationNotification` mechanism, more in details on the `UA_APPLICATIONNOTIFICATIONTYPE_SERVICE` and I have some point to clarify:

in my application I would really have the need to follow the list of nodes that are required for READ or WRITE operations. Using the application notification I am able to be notified about the `request-id` and the `session-id` and the Servuce ID but not more about the nodes that are involved in the request itself, so I am guessing if:
- Can I retrieve this somehow via the `request-id` ?
- Shall we add new API targeting a given service, for instance for read node to add a specific callaback `UA_APPLICATIONNOTIFICATIONTYPE_SERVICE_NODE_READ`  where we can give the list of nodeid that are required to be read ?

Thanks for your feedbacks !

Lorenzo

Issue #8126, opened 2026-06-19
Zephyr CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS maps to deprecated CMake option and leaves encryption disabled


Ubberx:
## Description
When open62541 is used as a Zephyr module, enabling the Zephyr mbedTLS option (`CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS=y`) for open62541 does not enable encryption in the generated open62541 configuration.

## Expected behavior

When the Zephyr configuration enables:

```conf
CONFIG_MBEDTLS=y
CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS=y
```

open62541 should be configured with mbedTLS encryption enabled. In CMake terms, the Zephyr option should result in:

```cmake
UA_ENABLE_ENCRYPTION=MBEDTLS
```

and the generated open62541 configuration should contain encryption support.

## Actual behavior

The Zephyr Kconfig option is translated to the deprecated backend-specific CMake variable:

```cmake
UA_ENABLE_ENCRYPTION_MBEDTLS=1
```

but `UA_ENABLE_ENCRYPTION` remains at its default value:

```cmake
UA_ENABLE_ENCRYPTION=OFF
```

The main CMake logic then resets the backend-specific encryption flags, so the generated open62541 configuration has encryption disabled.

This can be verified in the generated CMake cache or generated open62541 configuration. The relevant actual values look like this:

```text
UA_ENABLE_ENCRYPTION:STRING=OFF
UA_ENABLE_ENCRYPTION_MBEDTLS:STRING=0
```

or equivalently, the generated open62541 config does not define the expected mbedTLS encryption support.

## Reproduction

1. Use open62541 as a Zephyr module.
2. Add the following options to the Zephyr application configuration:

```conf
CONFIG_MBEDTLS=y
CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS=y
```

3. Build the Zephyr application, for example:

```sh
west build -b <board> <application>
```

4. Inspect the generated CMake cache or generated open62541 configuration.

Expected result:

```text
UA_ENABLE_ENCRYPTION=MBEDTLS
```

Actual result:

```text
UA_ENABLE_ENCRYPTION=OFF
```

and mbedTLS encryption support is not enabled in the generated open62541 configuration.

## In depth analysis
[...]

PR #8158, opened 2026-06-29
Fix: deprecated encryption CMake options are ignored

Ubberx:
# Fix: deprecated encryption CMake options are ignored

## Problem

The deprecated backend-specific encryption options (`UA_ENABLE_ENCRYPTION_MBEDTLS`,
`UA_ENABLE_ENCRYPTION_OPENSSL`, `UA_ENABLE_ENCRYPTION_LIBRESSL`) and the old boolean
form (`UA_ENABLE_ENCRYPTION=ON`) were silently ignored by the CMake configuration logic.

When a user configured open62541 with a deprecated option such as:

```bash
cmake -DUA_ENABLE_ENCRYPTION_MBEDTLS=ON ...
```

encryption was not enabled in the generated `config.h`.

### Root cause

The normalization block in `CMakeLists.txt` unconditionally reset all backend-specific
flags to `OFF` before the compatibility path could inspect them:

```cmake
list(FIND UA_ENCRYPTION_PLUGINS ${UA_ENABLE_ENCRYPTION} _tmp)
if(UA_ENABLE_ENCRYPTION STREQUAL "OFF" OR ${_tmp} GREATER -1)
    set(UA_ENABLE_ENCRYPTION_OPENSSL OFF)
    set(UA_ENABLE_ENCRYPTION_MBEDTLS OFF)
    set(UA_ENABLE_ENCRYPTION_LIBRESSL OFF)
    ...
else()
    # deprecated conversion — never reached when UA_ENABLE_ENCRYPTION=OFF
    if(UA_ENABLE_ENCRYPTION_MBEDTLS)
        set(UA_ENABLE_ENCRYPTION "MBEDTLS")
    ...
```

Since `UA_ENABLE_ENCRYPTION` defaults to `"OFF"`, the first branch was always taken,
clearing the deprecated flags before the `else` compatibility path could run. The
deprecated conversion was therefore unreachable in the common case.

The same ordering problem affected `UA_ENABLE_ENCRYPTION=ON` (the old boolean style
used before the selector API was introduced): it was handled by an `elseif` fallback
that silently defaulted to mbedTLS without a clear deprecation message.

## Applied fix

The deprecated-to-current conversion is now performed **before** the unconditional
reset. The backend flags are always reset to `OFF` after the conversion, so the
remaining logic has a clean state to work with.

The `UA_ENABLE_ENCRYPTION=ON` case is also handled explicitly: if a backend-specific
flag is set alongside it, that flag determines the backend; otherwise it defaults to
`MBEDTLS`. An explicit `DEPRECATION` message is emitted in both cases.

Invalid values now produce a `WARNING` and are reset to `"OFF"` instead of falling
through silently.

```cmake
# Convert deprecated backend-specific options to the current selector option.
if(UA_ENABLE_ENCRYPTION STREQUAL "OFF" OR UA_ENABLE_ENCRYPTION STREQUAL "ON")
    if(UA_ENABLE_ENCRYPTION_MBEDTLS)
        message(DEPRECATION "Set UA_ENABLE_ENCRYPTION=MBEDTLS instead of UA_ENABLE_ENCRYPTION_MBEDTLS=ON.")
        set(UA_ENABLE_ENCRYPTION "MBEDTLS")
    elseif(UA_ENABLE_ENCRYPTION_OPENSSL)
        message(DEPRECATION "Set UA_ENABLE_ENCRYPTION=OPENSSL instead of UA_ENABLE_ENCRYPTION_OPENSSL=ON.")
        set(UA_ENABLE_ENCRYPTION "OPENSSL")
    elseif(UA_ENABLE_ENCRYPTION_LIBRESSL)
        message(DEPRECATION "Set UA_ENABLE_ENCRYPTION=LIBRESSL instead of UA_ENABLE_ENCRYPTION_LIBRESSL=ON.")
        set(UA_ENABLE_ENCRYPTION "LIBRESSL")
    elseif(UA_ENABLE_ENCRYPTION STREQUAL "ON")
        message(DEPRECATION "Set UA_ENABLE_ENCRYPTION=MBEDTLS instead of UA_ENABLE_ENCRYPTION=ON.")
        set(UA_ENABLE_ENCRYPTION "MBEDTLS")
    endif()
endif()

set(UA_ENABLE_ENCRYPTION_OPENSSL OFF)
set(UA_ENABLE_ENCRYPTION_MBEDTLS OFF)
set(UA_ENABLE_ENCRYPTION_LIBRESSL OFF)

list(FIND UA_ENCRYPTION_PLUGINS ${UA_ENABLE_ENCRYPTION} _tmp)

if(UA_ENABLE_ENCRYPTION STREQUAL "OFF" OR ${_tmp} GREATER -1)
    if(UA_ENABLE_ENCRYPTION STREQUAL "MBEDTLS")
        set(UA_ENABLE_ENCRYPTION_MBEDTLS ON)
    elseif(UA_ENABLE_ENCRYPTION STREQUAL "OPENSSL")
        set(UA_ENABLE_ENCRYPTION_OPENSSL ON)
    elseif(UA_ENABLE_ENCRYPTION STREQUAL "LIBRESSL")
        set(UA_ENABLE_ENCRYPTION_LIBRESSL ON)
    endif()
else()
    message(WARNING "Invalid UA_ENABLE_ENCRYPTION value: ${UA_ENABLE_ENCRYPTION}. Proceeding without encryption")
    set(UA_ENABLE_ENCRYPTION "OFF")
endif()
```

## File changed

- `CMakeLists.txt`

## Related Issue

Zephyr CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS maps to deprecated CMake option and leaves encryption disabled
https://github.com/open62541/open62541/issues/8126


jpfr:
Merged, thanks

PR #8159, opened 2026-06-29
Fix: Zephyr mbedTLS Kconfig option does not enable open62541 encryption

Ubberx:
## Problem

When open62541 is used as a Zephyr module, enabling the mbedTLS encryption option
in Zephyr Kconfig:

```conf
CONFIG_MBEDTLS=y
CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS=y
```

did not enable encryption in the generated open62541 `config.h`.

### Root cause

In `arch/zephyr/zephyr.cmake`, the Kconfig option was mapped to the deprecated
backend-specific CMake variable using the existing bool translation macro:

```cmake
open62541_translate_zephyr_bool_option(
    UA_ENABLE_ENCRYPTION_MBEDTLS
    CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS
)
```

This sets `UA_ENABLE_ENCRYPTION_MBEDTLS=1` but leaves `UA_ENABLE_ENCRYPTION` at
its default value of `"OFF"`. The main `CMakeLists.txt` then enters the `"OFF"`
branch of the normalization block, which resets `UA_ENABLE_ENCRYPTION_MBEDTLS`
back to `OFF` before the deprecated compatibility path can convert it to
`UA_ENABLE_ENCRYPTION=MBEDTLS`.

The result is that encryption is silently disabled even though the Zephyr project
explicitly enables `CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS=y`.

## Applied fix

A new macro `open62541_translate_zephyr_bool_to_string_option` was added to
`arch/zephyr/zephyr.cmake`. Unlike the existing bool macro (which maps `y` → `1`),
this macro maps a Zephyr bool Kconfig option to an arbitrary string value on a
named CMake variable:

```cmake
macro(open62541_translate_zephyr_bool_to_string_option name value option)
    if(DEFINED ${option} AND "${${option}}" STREQUAL "y")
        set(${name} ${value} CACHE STRING "" FORCE)
    endif()
endmacro()
```

The encryption mapping now uses this macro to set the current primary CMake option
directly:

```cmake
open62541_translate_zephyr_bool_to_string_option(
    UA_ENABLE_ENCRYPTION "MBEDTLS"
    CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS
)
```

When `CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS=y`, this sets
`UA_ENABLE_ENCRYPTION=MBEDTLS` in the CMake cache before the main
`CMakeLists.txt` normalization block runs. The normalization block then correctly
sets `UA_ENABLE_ENCRYPTION_MBEDTLS=ON` and encryption is enabled in the generated
`config.h`.

## Related Pull Request

The `deprecated_encryption_options_patch` PR ( Fix: deprecated encryption CMake options are ignored#8158) fixes the same area of `CMakeLists.txt` for non-Zephyr users who still pass deprecated options such as
`-DUA_ENABLE_ENCRYPTION_MBEDTLS=ON`. This Zephyr fix sidesteps that bug entirely
by setting `UA_ENABLE_ENCRYPTION=MBEDTLS` directly, never going through the
deprecated path. The two fixes are independent but address the same root area.

## File changed

- `arch/zephyr/zephyr.cmake`

## Related Issue

Zephyr CONFIG_OPEN62541_ENABLE_ENCRYPTION_MBEDTLS maps to deprecated CMake option and leaves encryption disabled
https://github.com/open62541/open62541/issues/8126


jpfr:
Thanks.
Please rebase and make this a PR against the 1.5 branch.

Ubberx:
Hello,
I rebased onto 1.5 as requested

jpfr:
Thanks, merged.

PR #8163, opened 2026-06-30
tests(fuzz): add datatype description harness

DavidKorczynski:
This harnses is to be consumed by OSS-Fuzz and targets UA_DataType_fromDescription. This function currently has 0 code coverage in OSS-Fuzz.


jpfr:
Rebase with two commits on top

PR #8174, opened 2026-07-07
update mdnsd to v1.0 tag

matkonnerth:


jpfr:
Thanks for the update.
It seems the commit was done with a new user.
Can you sign the CLA again?

matkonnerth:
sry, had to add the commit again, otherwise cla would fail

PR #7974, opened 2026-04-07
Role-Based Access Control (RBAC): Batch 8 - GDS securityadmin, standard roles and docs

andreasebner:
Work on implementing Role-Based Access Control according to Part 18 of the OPC UA specification began in PR https://github.com/open62541/open62541/pull/7458. As this is a substantial subsystem, the mechanism will be merged into the main branch over a series of ~7 PRs

This PR: 

- Register the SecurityKeyServer well-known roles (Admin, Push, Access) when available in the nodeset
- Restrict all GDS ServerConfiguration methods to the SecurityAdmin role per Part 12 §7.2
- Document known RBAC limitations in the source file headers (TrustedApplication gap, write support, filters, audit events)

Batch 1: Configuration option, core RBAC types and initialisation. (merged)
Batch 2: Add Roles and Permissions confugration structures and handling API.
Batch 3: Role Management and RBAC initialization
Batch 4: RBAC Information Model
Batch 5: Roles and Permssions Node Attribute
Batch 6: Session role auto-assignment
Batch 7: Namespace default role permissions API
Batch 8: GDS securityadmin, standard roles and docs

PR #8186, opened 2026-07-10
feature(server): Extended the access control plugin to read role permissions

mcpat-pilz:
This allows a vendor specific Role Permission model to be mapped to the standard attributes.

### DIFF
--- a/include/open62541/plugin/accesscontrol.h
+++ b/include/open62541/plugin/accesscontrol.h
@@ -142,6 +142,24 @@ struct UA_AccessControl {
                                                       UA_DateTime endTimestamp,
                                                       bool isDeleteModified);
 #endif
+
+#ifndef UA_ENABLE_RBAC
+    /* Allow reading of RolePermissions attribute */
+    UA_Boolean (*allowReadRolePermissions)(UA_Server *server,  UA_AccessControl *ac,
+                                           const UA_NodeId *sessionId, void *sessionContext,
+                                           const UA_NodeId *nodeId, void *nodeContext);
+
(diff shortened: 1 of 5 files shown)


jpfr:
@andreasebner Please check with @mcpat-pilz on this.

andreasebner:
Thank's for your contribution. Your work (with authorship) is now part of #7974.

andreasebner:
After reviewing this again, I've decided to remove this work from the scope of PR #7974. This proposal is currently designed explicitly for non-RBAC use cases. When RBAC is not enabled, we currently return BadAttributeIdInvalid. However, my initial concern is that this feature breaks flag-based ABI compatibility of the access control layer. Since RBAC is specification-defined and closely tied to the information model, I'm uncertain whether we can merge this into the mainline.

mcpat-pilz:
I think I understand what you mean. I can change this PR to always include those new callbacks into the AccessControl plugin unguarded.

The default implementation of the new callbacks are straight forward then for the case where RBAC is enabled - just following the scheme of the other callbacks that are already implemented based on RBAC.

For the case where RBAC is disabled, the new callbacks are just NULL by default. This case shall be properly handled in ua_services_attribute.c

So the ABI compatibility is ensured regardless of whether RBAC is enabled or not. And we have the freedom to fill the RolePermission attributes based on a vendor-specific mapping which is also a valid use case in the OPC UA specification.

Is it worth a try?

mcpat-pilz:
@andreasebner I reworked this and now the plugin is extended unconditionally as I lined out.

PR #8204, opened 2026-07-17
fix(client): include endpointUrl in FindServers request

MrAlaskan:
### Summary

This PR propagates the requested Discovery Endpoint URL in `UA_Client_findServers()` and add a unit test that verifies the returned `discoveryUrls` mirror the URL used by the client.

### Problem

According to the OPC UA specification, the client should pass the Discovery Endpoint URL it used in `FindServersRequest.endpointUrl`, and servers may use that value to return matching `discoveryUrls`.

Previously, `UA_Client_findServers()` did not set `FindServersRequest.endpointUrl`. As a result, servers could fall back to their default discovery URL selection logic and return `discoveryUrls` that did not match the actual Discovery Endpoint URL used by the client.

### Changes

- Set `request.endpointUrl` from the `serverUrl` argument in `UA_Client_findServers()`.
- Extend `check_client_discovery` to assert that the returned `discoveryUrls` match the requested URL.

### DIFF
--- a/src/client/ua_client_discovery.c
+++ b/src/client/ua_client_discovery.c
@@ -117,6 +117,7 @@ UA_Client_findServers(UA_Client *client, const char *serverUrl,
     }
 
     UA_StatusCode retval;
+    const UA_String url = UA_STRING((char*)(uintptr_t)serverUrl);
     if(!connected) {
         retval = connectSecureChannel(client, serverUrl);
         if(retval != UA_STATUSCODE_GOOD) {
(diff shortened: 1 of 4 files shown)


MrAlaskan:
Hi, I’ve submitted a few PRs, including this one, that I think could be helpful to the project. When you have a chance, could you please take a look? I’d be happy to make any changes if needed. Thanks!

jpfr:
Merged. Thanks!

PR #8208, opened 2026-07-17
fix(pubsub): keep delta frame field count consistent

MrAlaskan:
### Summary

This PR preserves the actual number of changed fields when generating UADP `DataDeltaFrame` messages.

### Problem

According to the OPC UA specification, a `DataDeltaFrame` only carries the subset of fields that changed since the last key frame, and `FieldCount` must match the number of encoded delta fields.

Previously, `UA_PubSubDataSetWriter_generateDeltaFrameMessage()` correctly counted changed fields first, but later overwrote `fieldCount` with the total number of fields in the `PublishedDataSet`. As a result, the generated `FieldCount` could exceed the number of valid delta entries, and later encoding logic could process uninitialized tail elements as if they were part of the delta frame.

### Changes

Keep `fieldCount` equal to the actual number of changed fields in `UA_PubSubDataSetWriter_generateDeltaFrameMessage()`.

### DIFF
--- a/src/pubsub/ua_pubsub_writer.c
+++ b/src/pubsub/ua_pubsub_writer.c
@@ -480,7 +480,6 @@ UA_PubSubDataSetWriter_generateDeltaFrameMessage(UA_PubSubManager *psm,
     if(!deltaFields)
         return UA_STATUSCODE_BADOUTOFMEMORY;
 
-    dsm->fieldCount = counter;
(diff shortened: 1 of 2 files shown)


jpfr:
Please add a unit test for the new behavior

MrAlaskan:
A unit test is added.

PR #8209, opened 2026-07-17
fix(pubsub): correct delta frame source picoseconds mask

MrAlaskan:
DeltaFrame generation used the `SOURCEPICOSECONDS` mask to clear `hasServerPicoseconds`, which mixed up the source and server picoseconds semantics. This change makes the `SOURCEPICOSECONDS` mask clear `hasSourcePicoseconds` instead.


jpfr:
Thanks. Merged.

PR #8211, opened 2026-07-17
fix(gds): publish configured max trust list size

MrAlaskan:
`ServerConfiguration.MaxTrustListSize` was always advertised as `0` (unlimited), even when TrustList writes were enforced against `server->config.maxTrustListSize`. This change publishes the configured limit in the information model so the advertised TrustList size stays aligned with the runtime write limit.


jpfr:
rebased and moved to the new gds receive driver location

PR #8214, opened 2026-07-17
fix(discovery): filter before limiting FindServersOnNetwork

MrAlaskan:
### Summary

This PR fixes `FindServersOnNetwork` so `maxRecordsToReturn` is applied after capability filtering, and adds a regression test that covers the case where matching servers appear only after earlier non-matching records.

### Problem

According to OPC UA Specification, `FindServersOnNetwork` first considers all records with `recordId > startingRecordId`, then filters that candidate set with `serverCapabilityFilter[]`, and finally limits the response to at most `maxRecordsToReturn` matching entries.

Previously, `process_FindServersOnNetwork()` truncated the candidate window to `maxRecordsToReturn` before calling `entryMatchesCapabilityFilter()`. As a result, the implementation only scanned the first N raw records after `startingRecordId` instead of the full candidate set. If those early records did not match the requested capabilities, later matching entries were skipped entirely and never made it into the response.

### Changes

- Keep scanning all candidate `ServerOnNetwork` records after the `startingRecordId` cutoff.
- Apply `entryMatchesCapabilityFilter()` before enforcing the `maxRecordsToReturn` response limit.
- Add a regression test in `check_discovery_mdnsd` that verifies a later matching record is still returned even when an earlier non-matching record consumes the first raw slot.

### DIFF
--- a/src/server/ua_services_discovery.c
+++ b/src/server/ua_services_discovery.c
@@ -357,17 +357,16 @@ process_FindServersOnNetwork(UA_Server *server, UA_Session *session,
             break;
     }
 
-    /* Compute the max number of records to return */
+    /* Compute the number of candidate records after the record id cutoff */
     size_t recordCount = server->serversOnNetworkSize - recordOffset;
-    if(maxRecordsToReturn > 0 && maxRecordsToReturn < recordCount)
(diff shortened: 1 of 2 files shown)


jpfr:
Thanks. Merged.

PR #8227, opened 2026-07-21
fix(deps): always NUL-terminate cj5_get_str() output buffer

o6-nbeier:
cj5_get_str() only wrote the terminating NUL on its success path. Every
error return (incomplete/invalid escape sequences, truncated \u
codepoints, invalid surrogate pairs) left the caller-supplied buffer
unterminated.

Callers in ua_config_json.c (e.g. BuildInfo_parseJson()) allocate the
buffer with UA_malloc() and pass it to strcmp()/UA_LOG_ERROR("%s", ...)
without checking cj5_get_str()'s return code, so a malformed escape in
a JSON5 config field name (e.g. "\u12") causes a heap out-of-bounds read
past the allocation.

Restructure cj5_get_str() around a single exit point so buf[outpos] = 0
is written unconditionally, before every return -- success or error.
Callers should still check the returned error code, but can no longer
read past the buffer if they don't.

Extended the current test suite accordingly as well.

Reported by Martin Mohl (@Whit3rose)

### DIFF
--- a/deps/cj5.c
+++ b/deps/cj5.c
@@ -671,19 +671,26 @@ cj5_error_code
 cj5_get_str(const cj5_result *r, unsigned int tok_index,
             char *buf, unsigned int *buflen) {
     const cj5_token *token = &r->tokens[tok_index];
-    if(token->type != CJ5_TOKEN_STRING)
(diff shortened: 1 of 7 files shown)


jpfr:
Thanks. Merged.

PR #8235, opened 2026-07-23
fix(core): reject overflowing arrayDimensions product in Variant en/d…

o6-nbeier:
…ecoding

Multiplying attacker-controlled arrayDimensions values into a size_t accumulator without an overflow check allows the product to wrap and match a small arrayLength. Subsequent ranged read or write operations then access memory outside the allocated array bounds.

Add checked multiplication at all sites where the dimension product is computed: Variant binary decode, Variant binary encode, checkAdjustRange, and computeStrides (as a defence-in-depth measure).

Internal Vulnerability Advisory: open62541-SA-2026-0012 and open62541-SA-2026-0013
Reported by Asher Davila and Malav Vyas from Palo Alto Networks

### DIFF
--- a/src/ua_types.c
+++ b/src/ua_types.c
@@ -636,8 +636,11 @@ checkAdjustRange(const UA_Variant *v, UA_NumericRange *range) {
     /* Check that the number of elements in the variant matches the array
      * dimensions */
     size_t elements = 1;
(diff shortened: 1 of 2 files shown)


jpfr:
👍

PR #8237, opened 2026-07-23
fix(pubsub): check message length before signature size subtraction

o6-nbeier:
Add bounds check in verifyAndDecryptNetworkMessage() to prevent unsigned integer underflow when buffer.length is smaller than the expected signature size. Without this check, the subtraction wraps around producing a large size_t that causes mbedtls to read unmapped memory.

Internal Vulnerability Advisory: open62541-SA-2026-0014
Reported by Abhinav Agarwal


jpfr:
Unrelated fuzzing issues. Merged.

PR #8239, opened 2026-07-23
fix(server): migrate samplingMonitoredItems in TransferSubscriptions

o6-nbeier:
Operation_TransferSubscription() copies the UA_Subscription struct with memcpy but only migrates monitoredItems, notificationQueue, and retransmissionQueue. The samplingMonitoredItems list was not re-initialized in the new subscription, leaving stale le_prev backpointers from items with samplingType==PUBLISH pointing into the freed old subscription struct. When UA_MonitoredItem_unregisterSampling() calls LIST_REMOVE, it writes to freed memory (heap-use-after-free).

Add migration of samplingMonitoredItems to the new subscription after the point of no return, matching the pattern used for monitoredItems.

Internal Vulnerability Advisory: open62541-SA-2026-0015
Reported by Abhinav Agarwal

### DIFF
--- a/src/server/ua_services_subscription.c
+++ b/src/server/ua_services_subscription.c
@@ -555,6 +555,15 @@ Operation_TransferSubscription(UA_Server *server, UA_Session *session,
     }
     sub->monitoredItemsSize = 0;
 
+    /* Move over the samplingMonitoredItems and adjust the backpointers */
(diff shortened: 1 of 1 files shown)


jpfr:
Unrelated fuzzing issues. Merged.

PR #8250, opened 2026-07-27
feat(discovery): Add mDNS option to send announcements to all interfa…

engel-bichlerh:
…ces (port of PR #7057)


jpfr:
0:send-to-all-interfaces 

Can't we have the same if the interfaces argument is empty?

engel-bichlerh:
Yes I can remove the 0:send-to-all-interfaces flag and check if the interface argument is empty, and use that instead.

PR #8255, opened 2026-07-29
fix(build): Fix amalgamation build for win32

basyskom-jvoe:
The file tr_dirent.h should always be included in the amalgamation or setting UA_ENABLE_ENCRYPTION later will result in build errors.


jpfr:
Thanks, merged.

PR #8256, opened 2026-07-30
fix(build): Fix UA_THREAD_LOCAL for MSVC 14.36.32532 (and older?)

basyskom-jvoe:
The existing code works for MSVC 14.43.34808 but 14.36.32532 fails with errors like: C2146: syntax error: missing ';' before identifier 'UA_KeyValuePair'.


basyskom-jvoe:
@jpfr The WIP check seems to be stuck and also doesn't restart on title change. Could you please restart it manually?

jpfr:
👍

Issue #7480, opened 2025-10-08
Access to underlying posix fds to combine event loops


xR3b0rn:
Would be nice to have access to the underlying fds to combine event loops. With the current implementation I could not find a way of doing so.

xR3b0rn:
To add some context:

In my specific case, I want to integrate the open62541 event loop into the event loop of another library  (or vice versa), which uses the `sd_event` implementation from **systemd**.

Systemd provides [sd_event_get_fd](https://www.freedesktop.org/software/systemd/man/latest/sd_event_get_fd.html), which is explicitly intended to make it possible to use its file descriptors in combination with `epoll` in order to merge two event loops:

> This file descriptor may itself be polled for POLLIN/EPOLLIN events. This makes it possible to embed an [sd-event(3)](https://www.freedesktop.org/software/systemd/man/latest/sd-event.html) event loop into another, possibly foreign, event loop.

However, because open62541 manages its file descriptors internally and does not expose them, it is currently impossible to efficiently integrate the open62541 event loop with another event loop in the same thread.

xR3b0rn:
Could we add a `getEpollFd` function pointer to the `EventLoop` struct, which is set only if `epoll` is available, and otherwise is `NULL`?

kh-schneider:
Could also make use of it: In my case to handle connections to unix domain sockets.

xR3b0rn:
## Proposed GLib-based EventLoop backend for open62541

Since I need this feature to avoid two-thread solution, busy polling, or timer-based polling, I spent some time understanding the internal structure of the open62541 EventLoop as well as the event loop model of GLib.

### Observations

`open62541` deliberately hides file descriptors (FDs) from the public API. This is consistent with its design goals:

* Platform independence (POSIX, Windows, RTOS, embedded)
* Backend-agnostic networking
* Avoiding exposure of OS-specific concepts in the public API

Because of this abstraction, it is not currently possible to embed the existing default open62541 event loop into another external event loop in a prepare/dispatch style (as e.g. GLib supports). The current model is ownership-based:

* `UA_Server_run()` / `UA_Server_run_iterate()` drive the loop
* The backend (posix/win32) owns polling and FD registration

---

## Proposal

To get around this limitation a third-party event loop library backend such as a GLib-based `UA_EventLoop` backend could be implemented.

This does not require changing the public API or exposing FDs. Instead, it provides an alternative backend implementation using `GMainContext` and `GSource`.

---

## Visible Changes for Users

[...]

PR #8258, opened 2026-07-30
feat(arch): glib eventloop

xR3b0rn:
Implements #7480

Vibe-coded GLib event loop. With this MR, open62541 can be fully driven by GLib or vice versa (see `examples`).

If you don't like how claude touched the POSIX implementation, we can change it quite easily!


jpfr:
Nice. Will take a look.
In the end, claude should not be a named contributor.
You run the agent, it should be your code in the end.

    signal(SIGINT, stopHandler);
    signal(SIGTERM, stopHandler);

If you integrate the interrupt handling via glib, then the loop abortion can be done fully by glib.
That helps for GUI applications

xR3b0rn:
Added `arch/posix/eventloop_glib_interrupt.c` and removed claude as co-contributor.

jpfr:
Nice, merged.

PR #8271, opened 2026-08-04
Decouples WebSocket transport from UA_ENABLE_LWS

diplfranzhoepfinger:
Separates server-side WebSocket transport wiring from reliance on UA_ENABLE_LWS.

Adds support for plain opc.ws:// WebSocket server, enabling non-TLS WebSocket connections alongside the existing opc.wss:// support. This enhancement allows non-POSIX platforms to implement a custom "websocket" UA_ConnectionManager using native HTTP/WS stacks.

Adressing: https://github.com/o6-automation/open62541-dev/pull/2#issuecomment-5175143291

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,7 @@ mark_as_advanced(UA_ENABLE_LWS)
 option(UA_ENABLE_LWS_MQTT "Enable the libwebsockets MQTT ConnectionManager" OFF)
 mark_as_advanced(UA_ENABLE_LWS_MQTT)
 
+
 if(UA_ENABLE_LWS_MQTT AND NOT UA_ENABLE_LWS)
     message(FATAL_ERROR "UA_ENABLE_LWS_MQTT requires UA_ENABLE_LWS")
 endif()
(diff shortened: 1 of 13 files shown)


jpfr:
The opc.ws:// without tls is not defined in the standard.
It should only be enabled with a configuration flag that defaults to false.
And with additional warnings during startup.

diplfranzhoepfinger:
> The opc.ws:// without tls is not defined in the standard.
> It should only be enabled with a configuration flag that defaults to false.
> And with additional warnings during startup.

i Agree. this should be only for very first local Tests in a DMZ or so. Only for People who are aware what they are doing. 

`Unencrypted WebSocket transport is non-standard and should only be used for testing/debugging!`
this is the Warning. 

`webSocketAllowUnencrypted` is the config, defaults to false.

jpfr:
Looks good.

- UA_ENABLE_WEBSOCKET_TRANSPORT can be removed entirely from CMake and the config.h.in
- Extend the UA_ENABLE_LWS documentation in /doc/build.rst with the protocols that are enabled by it.

diplfranzhoepfinger:
> Looks good.
> 
> * UA_ENABLE_WEBSOCKET_TRANSPORT can be removed entirely from CMake and the config.h.in
> * Extend the UA_ENABLE_LWS documentation in /doc/build.rst with the protocols that are enabled by it.

i did a sqash, so the no more existent UA_ENABLE_WEBSOCKET_TRANSPORT is no more visible in Git commit History.

jpfr:
Merged. Nice!

diplfranzhoepfinger:
Updated my Demo here: 

https://github.com/diplfranzhoepfinger/opcua-dashboard
https://github.com/diplfranzhoepfinger/opcua-dashboard/tree/v4

Issue #7796, opened 2026-03-02
JSON decode of null QualifiedName fails with BADDECODINGERROR


fischjo3SICKAG:
## Summary

Attempting to round-trip encode and decode a null `UA_QualifiedName` value via JSON encoding fails. The encoding succeeds and produces a JSON `null` value, but decoding that `null` back to a `UA_QualifiedName` returns `UA_STATUSCODE_BADDECODINGERROR` instead of `UA_STATUSCODE_GOOD`.

## Minimal Reproducible Example

```c
UA_QualifiedName nullValue;
UA_QualifiedName_init(&nullValue);
ck_assert(UA_QualifiedName_isNull(&nullValue));

UA_ByteString encoded;
UA_ByteString_init(&encoded);
ck_assert_uint_eq(UA_STATUSCODE_GOOD, UA_encodeJson(&nullValue, &UA_TYPES[UA_TYPES_QUALIFIEDNAME], &encoded, NULL)); // succeeds, encoded as 'null'

UA_QualifiedName decoded;
UA_QualifiedName_init(&decoded);
ck_assert_uint_eq(UA_STATUSCODE_GOOD, UA_decodeJson(&encoded, &decoded, &UA_TYPES[UA_TYPES_QUALIFIEDNAME], NULL)); // FAILS
ck_assert(UA_QualifiedName_isNull(&decoded));
```

## Actual Error

```
Assertion '((UA_StatusCode) 0x00000000) == UA_decodeJson(&encoded, &decoded, &UA_TYPES[19], ((void *)0))' failed: 
((UA_StatusCode) 0x00000000) == 0, UA_decodeJson(&encoded, &decoded, &UA_TYPES[19], ((void *)0)) == 2147942400
```

The error code `2147942400` is `UA_STATUSCODE_BADDECODINGERROR` (`0x80070000`).

## Expected Behavior

Decoding a JSON `null` value should successfully produce a null `UA_QualifiedName` instance (i.e., return `UA_STATUSCODE_GOOD`).

## Investigation

My reading of OPC UA specification part 6 (v1.0.5), section 5.4.2.1, suggests that null encoding is defined only for array elements and fields. For other cases (such as scalars), the specification appears to delegate to RFC 8259, which uses `null` as the JSON representation. However, I acknowledge that this may not be the correct interpretation of the spec.

[...]

PR #8278, opened 2026-08-07
fix(core): fix JSON decoding of null QualifiedName values

DirikTi:
Related to #7796

UA_encodeJson() already serializes a null UA_QualifiedName as JSON null, but
UA_decodeJson() failed because QualifiedName decoding forwarded the null token
to String_decodeJson().

Handle JSON null values in DECODE_JSON(QualifiedName) and preserve the existing
null state behavior. Add a regression test for the JSON encode/decode
round-trip.


jpfr:
Thanks, looks good.
Please check the failing CI.

PR #7937, opened 2026-03-25
fix(arch): remove bits/stdio_lim.h

20162026:
Fix #7936

if there are some obscure or old systems that require this header we could wrap the include with `#ifdef __GLIBC__`

PR #8279, opened 2026-08-09
fix(posix): remove non-portable stdio_lim include from 1.5

### LINKED ISSUE #8197: 1.5 branch: inclusion of non-standard bits/stdio_lim.h causes compilation failure
When updating the FreeBSD port I have discovered that the latest 1.5 branch release includes the non-standard header `bits/stdio_lim.h` in `arch/posix/eventloop_posix.h` which simply doesn't exist on FreeBSD. The real fix would be to only include this file when compiling under conditions where this is required (gcc/glibc I guess?) and nowhere else. For now, I have patched this in ports by just adding a FreeBSD inclusion guard, but this is merely a workaround than a proper fix.

I have seen https://github.com/open62541/open62541/pull/7937, don't know if this can easily be backported.

PS: I also see that sys/inotify.h is included with a guard for `__linux__`. Likely a configure test is more appropriate here since FreeBSD has support for the inotify API since 15.x. and it would make sense to use it if it is available.

Daniiiil1:
## What changed

- remove the non-standard `<bits/stdio_lim.h>` include from the POSIX event-loop header
- keep the portable standard headers that already provide the file/path limits used by this header

## Why

`bits/stdio_lim.h` is a glibc implementation detail and is unavailable on FreeBSD. The include is unused and prevents the 1.5 branch from compiling there. This backports the relevant part of #7937 to the 1.5 branch.

Fixes #8197.

## Validation

- `git diff --check`
- `cmake -S . -B build -DUA_BUILD_EXAMPLES=OFF -DUA_BUILD_UNIT_TESTS=OFF -DCMAKE_BUILD_TYPE=Release`
- `cmake --build build -j4`

### DIFF
--- a/arch/posix/eventloop_posix.h
+++ b/arch/posix/eventloop_posix.h
@@ -275,10 +275,6 @@ typedef int SOCKET;
 #endif /* __linux__ */
 #include <sys/stat.h>
 
-#if !defined(__ANDROID__) && !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__OHOS__)
-#include <bits/stdio_lim.h>
-#endif
-
 #define UA_STAT stat
 #define UA_DIR DIR
 #define UA_DIRENT dirent


jpfr:
Thanks, merged.

PR #8280, opened 2026-08-09
fix: eliminate uninitialized-variable and potential overflow warnings under GCC LTO

zhaoxi-scut:
#### Background

When open62541 is built with GCC optimization and link-time optimization (LTO), the compiler reports several warnings in the Client, WebSocket Server, and PubSub NS0 code paths:

- `-Wmaybe-uninitialized`
- `-Wstringop-overflow`

The affected code includes a potentially uninitialized `UA_MessageType`, PubSub node-context pointers, and a theoretical integer wraparound in `path.length + 1` when allocating the WebSocket path buffer.

These warnings create noise for downstream projects using strict compiler-warning settings and may hide unrelated issues in the same build log.

#### Changes

- Explicitly initialize the message type with `UA_MESSAGETYPE_INVALID` in the Client network callback so that the value passed to `processServiceResponse` always has a defined state.
- Check for `SIZE_MAX` before incrementing the WebSocket path length, preventing the allocation size from wrapping to zero. Continue to use `UA_ByteString_allocBuffer` and preserve the existing out-of-memory status code.
- Add a shared node-context cleanup helper for the PubSub NS0 destructors. A context is released only when the NodeId is valid and `getNodeContext` succeeds.
- Preserve the behavior of successful execution paths while making state and resource ownership explicit on error paths.

#### Impact

- No public API or ABI changes.
- No intended behavioral changes on successful code paths.
- More robust handling of error paths.
- Removes the corresponding GCC optimization/LTO warnings.

#### Verification

- Completed a Release + LTO build with GCC 15.2.0.
- Built with the Client, PubSub, and PubSub Information Model enabled, including compilation of the WebSocket Server source.
- Completed the final LTO link of `tutorial_client_firststeps` and `tutorial_server_firststeps`.
- Confirmed that the reported `-Wmaybe-uninitialized` and `-Wstringop-overflow` warnings no longer occur.

### DIFF
--- a/src/client/ua_client_connect.c
+++ b/src/client/ua_client_connect.c
@@ -2218,7 +2218,7 @@ __Client_networkCallback(UA_ConnectionManager *cm, uintptr_t connectionId,
     /* Received a message. Process the message with the SecureChannel. */
     UA_StatusCode res = UA_SecureChannel_loadBuffer(&client->channel, msg);
     while(UA_LIKELY(res == UA_STATUSCODE_GOOD)) {
-        UA_MessageType messageType;
+        UA_MessageType messageType = UA_MESSAGETYPE_INVALID;
         UA_UInt32 requestId = 0;
         UA_ByteString payload = UA_BYTESTRING_NULL;
         UA_Boolean copied = false;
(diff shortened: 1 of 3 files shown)


jpfr:
Thanks, merged.

PR #8283, opened 2026-08-10
fix(util): Avoid a dangling pointer.

r-sin:
Calling UA_KeyValueMap_clear directly leaves a dangling pointer. Reusing the map via UA_KeyValueMap_set passes a stale non-NULL pointer to UA_Array_appendCopy.


jpfr:
Thanks, merged.

PR #8309, opened 2026-08-21
fix(plugins): check UserAccessLevel on HistoryRead request

mcpat-pilz:


mcpat-pilz:
What I do not like about this PR is, that the permission check is done in the plug-in itself. I think it is better located in the server-core - where all other permission checks are performed.

PR #8314, opened 2026-08-24
Fixed string offset table

jpfr:


jpfr:
already merged on master

PR #8316, opened 2026-08-24
fix(server): emit LIFECYCLE_STOPPED instead of duplicate STOPPING

o6-dop:
UA_APPLICTATIONNOTIFICATIONTYPE_LIFECYCLE_STOPPED never fired, instead LIFECYCLE_STOPPING was fired twice


jpfr:
👍  Merged

PR #8318, opened 2026-08-25
fix(arch): fix monotonic clock unit on zephyr

grassnick:
k_uptime_get() returns milliseconds, but UA_DateTime is measured in 100ns intervals, so the monotonic clock ran 10000x slow.
Derive the value from k_uptime_ticks() converted to nanoseconds instead.

See https://docs.zephyrproject.org/latest/kernel/services/timing/clocks.html


jpfr:
Great, thanks!

PR #8321, opened 2026-08-26
feature(server): Allow to limit the max size of untrusted messages

mcpat-pilz:
"Untrusted" messages are hereby messages that are sent over an unsecured Secure Channel (neither encrypted nor signed).

This complements PR #8142 to provide a separate knob for untrusted messages.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -2158,6 +2158,11 @@ struct UA_ServerConfig {
                               * (default: 0 -> unbounded) */
     UA_UInt32 tcpMaxChunks;  /* Max number of chunks per message
                               * (default: 0 -> unbounded) */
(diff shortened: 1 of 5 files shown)


jpfr:
This is nice!

However it limits the #None SecureChannel in a big way.

How about we check whether at least one session was activated for the SecureChannel?
The session-pointers are directly available in the SecureChannel.

mcpat-pilz:
Hm, I did this on purpose: Anonymous Sessions are still not trustworthy and therefore my thinking is to be strict with the limit.

Or to rephrase it: even though there is an activated Session a #None SecureChannel is still competely unsecured and the Client is not validated or trusted. So also in this state I want a stricter limit compared to that of validated and trusted Clients.

PR #8322, opened 2026-08-26
fix(nc): declare generated UA_TYPES-style arrays const

diplfranzhoepfinger:
generate_datatypes.py emits the per-namespace type description array
(UA_TYPES, and any other namespace's equivalent, e.g. UA_TRANSPORT)
as a plain, non-const global. Nothing in the codebase writes to these
arrays or their elements at runtime -- the public API is already
const-correct throughout (UA_DataType_clear/UA_DataType_fromDescription
are the only two functions taking a non-const UA_DataType*, and
neither is ever called with a UA_TYPES entry), and every internal use
already goes through a `const UA_DataType *`.

Declaring them const lets the compiler/linker place them in .rodata
(mappable to flash/PSRAM on embedded targets) instead of writable
.data/internal RAM. Measured on an ESP32-S3 embedded build using the
generated amalgamation: UA_TYPES alone is 32592 bytes, moving from
internal DRAM to flash-mapped .rodata with this change, verified
compiling and linking cleanly with no behavior change.

### DIFF
--- a/tests/nodeset-compiler/check_nodeset_compiler_testnodeset.c
+++ b/tests/nodeset-compiler/check_nodeset_compiler_testnodeset.c
@@ -12,10 +12,14 @@
 #include "tests/namespace_tests_testnodeset_generated.h"
 #include "namespace_tests_di_generated.h"
 #include <limits.h>
+#include <stdint.h>
 #include <stdlib.h>
 
 UA_Server *server = NULL;
-UA_DataTypeArray customTypesArray = { NULL, UA_TYPES_TESTS_TESTNODESET_COUNT, UA_TYPES_TESTS_TESTNODESET, UA_FALSE};
+/* UA_TYPES_TESTS_TESTNODESET is const (never mutated here); the cast only
+ * matches the non-const UA_DataTypeArray::types pointer type. */
+UA_DataTypeArray customTypesArray = { NULL, UA_TYPES_TESTS_TESTNODESET_COUNT,
+    (UA_DataType*)(uintptr_t)UA_TYPES_TESTS_TESTNODESET, UA_FALSE};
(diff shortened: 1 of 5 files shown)


andreasebner:
This appears to reimplement or duplicate #7950. The const types will land soon, but they have broader implications than this PR addresses.

diplfranzhoepfinger:
> This appears to reimplement or duplicate #7950. The const types will land soon, but they have broader implications than this PR addresses.

ah, i could test then the code of PR https://github.com/open62541/open62541/pull/7950 if it has the same effect. 
so far THANKS for already working on that, and SORRY for duplicating this.

diplfranzhoepfinger:
Closing in favor of #7950, per the comment above -- thanks for pointing at it.

Pulled that branch and checked it against our embedded use case (const
`UA_TYPES`/`UA_TYPES_TRANSPORT`, no companion nodesets). It covers what this
PR does and more:

- `UA_TYPES`/`UA_TYPES_TRANSPORT` end up const automatically (namespace 0 is
  always in the generator's namespace map), no opt-in flag needed at all --
  simpler than the `MUTABLE_TYPES`-style opt-out this PR would have needed
  for nodeset-compiler-generated companion types.
- It also makes the generated `*_members` arrays unconditionally const, which
  we hadn't attempted here. That turned out to matter for us beyond this PR's
  scope too -- see the closing note on #8323.
- The `NAMESPACE_MAP`/pinning mechanism for companion nodesets is a properly
  designed solution to a problem we only discovered by hitting it on
  hardware: making a nodeset-compiler-generated type array const breaks its
  in-place namespace-index patch unless the index is fixed at generation
  time and verified at load time instead. #7950 does exactly that, with test
  coverage for both the matching and mismatching case.

Good to close this one as superseded.

PR #8323, opened 2026-08-26
feat(nc): optionally place generated *_members arrays in external RAM

diplfranzhoepfinger:
Adds a --members-extram-attr <attrMacro> option to generate_datatypes.py
(wired through ua_generate_datatypes()'s new MEMBERS_EXTRAM_ATTR
argument, and a UA_TYPES_MEMBERS_EXTRAM_ATTR CMake cache option for
the two builtin invocations -- "types"/UA_TYPES and
"transport"/UA_TRANSPORT).

Default (unset): no change, *_members arrays stay exactly as before --
a plain compile-time-initialized static array.

When set (e.g. to EXT_RAM_BSS_ATTR on ESP-IDF, with
CONFIG_SPIRAM_ALLOW_BSS_SEG_EXTERNAL_MEMORY enabled): the array is
instead declared zero-initialized with the given attribute, and a
generated per-type init function assigns each element -- the exact
same link-time-constant values (pointers into the const
UA_TYPES-style arrays, offsetof()/sizeof() expressions) as before,
just via assignment instead of an aggregate initializer. All the
per-type init functions are called once from a single
__attribute__((constructor)) function per output file, so they run
automatically before any code (server, client, or standalone type
encode/decode) executes -- no explicit call required from any
existing or new open62541 user. Order between the init functions
doesn't matter: each only writes to its own array.

This exists because ESP-IDF's EXT_RAM_BSS_ATTR (the mechanism to
attribute-place a static variable in external PSRAM) only supports
zero-initialized (.bss) data; a regular non-zero initializer stays in
internal RAM regardless of the attribute. Companion to the earlier
`const UA_TYPES` fix: that one is free (no runtime cost, works
everywhere) and gets UA_TYPES itself (32592 bytes on our target) into
flash-mapped .rodata; the *_members arrays can't take that path since
UA_DataType.members is a non-const pointer (needed for legitimate
runtime custom-type registration), so moving them requires this
opt-in, runtime-init approach instead.

Verified on an ESP32-S3 embedded build using the generated
amalgamation with --members-extram-attr=EXT_RAM_BSS_ATTR: compiles
and links cleanly, a new 220KB .ext_ram.bss section appears (confirms
the real ESP-IDF attribute resolved, not the portable no-op
fallback), internal .dram0.data drops by a further 16912 bytes on top
of the const-UA_TYPES savings, and UA_TYPES_members_init is emitted
as .text.startup.UA_TYPES_members_init (GCC's constructor-function
naming convention) alongside the toolchain's normal
__libc_init_array/init_array machinery, confirming it runs
automatically at startup.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,6 +273,16 @@ if(UA_ENABLE_TPM2_KEYSTORE)
     endif()
 endif()
 
+# Generated *_members arrays (UA_DataTypeMember[]) placement. Empty
+# (default): compile-time-initialized as always. Set to an attribute macro
+# name (e.g. EXT_RAM_BSS_ATTR on ESP-IDF, with
+# CONFIG_SPIRAM_ALLOW_BSS_SEG_EXTERNAL_MEMORY enabled) to instead declare
+# them zero-initialized with that attribute and fill them in via a
+# generated __attribute__((constructor)) function -- lets embedded targets
+# place them in external RAM, since only zero-initialized (.bss) data can
+# be attribute-placed there. See tools/generate_datatypes.py.
+set(UA_TYPES_MEMBERS_EXTRAM_ATTR "" CACHE STRING "Attribute macro to zero-init-declare and runtime-populate generated *_members arrays with (e.g. EXT_RAM_BSS_ATTR); empty disables this and keeps them compile-time-initialized")
+
 # Namespace Zero
 set(UA_NAMESPACE_ZERO "REDUCED" CACHE STRING "Completeness of the generated namespace zero (none/minimal/reduced/full)")
 SET_PROPERTY(CACHE UA_NAMESPACE_ZERO PROPERTY STRINGS "NONE" "MINIMAL" "REDUCED" "FULL")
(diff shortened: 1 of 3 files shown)


diplfranzhoepfinger:
Closing this in favor of #7950.

That PR declares the generated `*_members` arrays `static const` unconditionally
(regardless of whether the owning type array itself is const), since member
records never carry namespace-dependent data -- only the type's own `typeId`/
`binaryEncodingId` do, and only for types belonging to a nodeset whose
namespace index isn't fixed at generation time. The `.memberType` pointers
inside `_members` (`&UA_TYPES[k]`-style) are ordinary link-time-constant
addresses, valid whether or not the target array ends up const.

I assumed `UA_DataType.members` had to stay a non-const pointer to allow the
existing runtime custom-type-registration path, which is why this PR went the
opt-in `--members-extram-attr` + runtime-constructor route instead. That
assumption doesn't hold: nothing about legitimate runtime registration needs
`members` itself to be mutable, only the top-level array.

Practical effect for our embedded use case: I verified on our actual
ESP32-P4/RISC-V toolchain (non-PIC, absolute addressing, as any ESP-IDF app is
linked) that a `const` array of structs containing pointers to other globals
is placed straight in `.rodata` -- no separate `.data.rel.ro` section, no
runtime relocation, zero RAM cost. So #7950's plain `const` `_members` gets
the same arrays we were moving into PSRAM here into flash instead, for free:
no opt-in flag, no `CONFIG_SPIRAM_ALLOW_BSS_SEG_EXTERNAL_MEMORY` dependency,
no constructor running before `main()`, and it doesn't spend any PSRAM budget
at all (versus the ~220KB `.ext_ram.bss` this PR's approach used on our
target).

Given that, this PR's opt-in runtime-init mechanism doesn't pull its weight
next to #7950 -- closing as superseded. Thanks for pointing at #7950 on the
related #8322.

PR #8328, opened 2026-08-28
fix(server/types): enable retaining of ServerLog object and honor optional fields in NodeSets 

mlgiraud:
This PR makes it possible to configure the NodeSet injector to retain the standard ServerLog subtree when providing a custom implementation.

It also makes generated data types honor optional fields declared in NodeSet DataTypeDefinitions. This fixes LogRecord encoding. Its optional fields were treated as required and encoded unconditionally instead of being controlled by an encoding mask. This led to encoding errors when attempting to fetch log records with UaExpert for example.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,14 +131,6 @@ endif()
 
 if(UA_INFORMATION_MODEL_AUTOLOAD AND NOT UA_BUILD_FUZZING)
     set(UA_ENABLE_NODESET_INJECTOR ON)
-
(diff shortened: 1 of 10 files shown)


jpfr:
The private macro definitions will be lost for the amalgamation build.
Any chance to detect a dependency on the log object at runtime?
By the existence of a new reference maybe?

mlgiraud:
This should be a more generic solution now, including all the other deleted nodes. Sounds like a reasonable approach to keep anything the user modified, right?

PR #8331, opened 2026-08-28
fix(plugin): open an outbound socket for mDNS sends

mlgiraud:
The mDNSd driver used listen=true for both UDP connections. The POSIX connection manager therefore created two receive sockets. Calls to sendto on the second socket failed with EINVAL.

Set listen=false before the driver opens the send connection. The regression test verifies one receive connection and one send connection at startup.

### DIFF
--- a/drivers/discovery_mdns_mdnsd.c
+++ b/drivers/discovery_mdns_mdnsd.c
@@ -1054,6 +1054,7 @@ createMulticastSocket(MdnsdDriver *md) {
     }
 
     /* Open the send connection */
+    listen = false;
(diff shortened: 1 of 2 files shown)


jpfr:
Thanks @mlgiraud for the one-line fix. 😆 
I added a second end-to-end test on top.

mlgiraud:
This seems to be a regression between 1.5.6 and master btw. Worked on that tag and i updated to master for some newer features, but that broke mDNS. Luckily the fix was easy as you can tell :D

jpfr:
Thanks, merged.

Yes. There was big refactor.

Mdnsd is a 'driver' now, and no longer in the core library.

PR #8333, opened 2026-08-28
fix: preserve digits in NodeSet symbolic names

mlgiraud:
NodeSet BrowseNames can contain digits that are part of the type name. Removing all digits causes ThreeDFrame to lose its NodeId.


jpfr:
Thanks, merged.

PR #8339, opened 2026-08-29
fix(client): do not run the connectivity check before the Session is …

### LINKED ISSUE #8195: [Question, Bug?] ConnectivityCheckInterval read request before SessionCreateResponse
# Problem description
While trying to use the client's `UA_Client_connect` I'm encountering the following issue:

With a particular plc, the secure channel creation takes quite some time (which i have worked around by setting the client's timeout to a high enough value). 
However, the session creation also takes some time. 
By enabling debug log level, I've found, that the read-request on the server status caused by ua_client.c 's `__Client_backgroundConnectivity` gets fired, even before the server returns CreateSessionResponse (so also before the client tries to activate the session).
This happens shortly after sending CreateSessionRequest, regardless of the value of `connectivityCheckInterval`, only disabling it by setting it to 0 fixes the connect (albeit of course breaking connection monitoring). 

The ReadRequest then returns a ServiceFault, leading to the session being scrapped and the SessionCreateResponse being ignored, as the request for that session has been cleaned up already as well.

According to the opc reference, it states that:
[The Session created with this Service shall not be used until the Client calls the ActivateSession Service and proves possession of its ApplicationInstanceCertificate and any user identity token that it provided.](https://reference.opcfoundation.org/specs/OPC-10000-4/5.7)

And in my understanding any read-request requires a session. So is this a bug or intended behaviour?
(Unfortunately I'm not that prolific of a c programmer, so I have some trouble understanding the source at times)

# Proposed fix
If my assumption is correct, the connectivity-check should wait for the session to be `client->sessionState == UA_SESSIONSTATE_ACTIVATED` before sending the first read-request

StalderT:
…activated

__Client_backgroundConnectivity sent its ServerStatus Read regardless of the Session state. When the SecureChannel and Session setup take a while, the housekeeping timer can fire the Read before the server has returned ActivateSessionResponse; the resulting ServiceFault aborts the connect.

Guard the check on client->sessionState == UA_SESSIONSTATE_ACTIVATED: per the spec a Session must not be used before ActivateSession completes.

Fixes #8195.

### DIFF
--- a/src/client/ua_client.c
+++ b/src/client/ua_client.c
@@ -1019,6 +1019,12 @@ __Client_backgroundConnectivity(UA_Client *client) {
     if(!client->config.connectivityCheckInterval)
         return;
 
+    /* Only probe the server once the Session is activated. A Read sent
+     * before the server returns ActivateSessionResponse triggers a
+     * ServiceFault that aborts a slow connect (issue #8195). */
+    if(client->sessionState != UA_SESSIONSTATE_ACTIVATED)
+        return;
+
     if(client->pendingConnectivityCheck)
         return;
 


StalderT:
Some extra context on the impact, since it's easy to miss:

With `connectivityCheckInterval` set and the server unreachable, the background connectivity `Read` gets retried on every housekeeping tick (~1s) instead of once per interval: `lastConnectivityCheck` is only advanced in the response callback, which never runs when the send fails on a closed channel. The result is a steady stream of `SecureChannel must be connected to send request`.

Gating on `UA_SESSIONSTATE_ACTIVATED` stops it cleanly — in a quick test against a dead endpoint the error log went from **~1/s to 0**.

Ready to merge from my side.

jpfr:
Thanks, merged 👍

Issue #8327, opened 2026-08-28
Marking UA_Server_setAsyncCallMethodResult UA_THREADSAFE  seems pointless and missleading


xR3b0rn:
In the header `UA_Server_setAsyncCallMethodResult` is marked as 'thread-safe', which is highly misleading and sets a dangerous trap for users.

Because this function uses the output pointer as the identifier to map the result back to the client's method call, it must be called from within the open62541 event loop. If called from a worker thread, it is impossible to avoid a race condition: If the method call is cancelled (e.g., due to a timeout) just before the worker thread delivers the result, the event loop frees the UA_Variant memory. The worker thread will then pass a dangling pointer to this function, resulting in a Use-After-Free and a server crash.

Thread-safety inside the function implementation doesn't help if the API design inherently requires event loop synchronization to ensure pointer validity.

xR3b0rn:
Could we change the lifecycle behavior of async method calls so that the ownership of the async context (or the output pointer) is transferred from the event loop to the user's callback handler? Ownership would then be transferred back to the event loop once `UA_Server_setAsyncCallMethodResult` is called. This way, users wouldn't be forced to implement complex dispatching logic (like an event loop trampoline) just to avoid race conditions and use-after-free bugs.

xR3b0rn:
Btw, regarding the official async example (examples/tutorial_server_method_async.c):
This example is highly misleading regarding how asynchronous methods should actually be implemented. It doesn't demonstrate real concurrency; it merely defers the execution to the next iteration of the main event loop.

Because the 'mocked worker' runs in the exact same thread as `UA_Server_run_iterate`, the mutex used to lock the workQueue is completely pointless - the cancel callback and the worker can never run concurrently anyway. This gives developers the false impression that a simple mutex is enough to prevent race conditions in a real multi-threaded scenario (which it is not, due to API boundaries).

It might be worth rewriting this example to demonstrate a real background worker thread to show the actual complexity. If rewritten, it would also be necessary to demonstrate delegating the final check and result delivery back to the open62541 event loop. As I described above, because the event loop can free the output pointer at any time upon a timeout, simply locking an "alive"-queue in the worker thread is entirely insufficient to prevent a use-after-free.

jpfr:
This is solved via https://github.com/open62541/open62541/pull/8470.

The async examples already showed how this can be used safely.

With the new PR the same will then be possible without an additional mutex.
At the expense of more moving parts inside the core library.

PR #8343, opened 2026-08-29
feat(async): defer freeing async operation memory until the worker acknowledges completion

xR3b0rn:
Related to #8327.

With this MR no complicated dispatch logic is required anymore to get rid of the race condition between a canceled async op and the worker thread which processes the op.

While implementing this ownership transfer, I considered the edge case of a complete server shutdown (UA_Server_delete). If the server is torn down while background worker threads are still processing async operations and holding onto the pointers, a lifecycle conflict occurs.

However, this shouldn't be an issue as long as it is clearly documented: Users are responsible for gracefully stopping and joining their background worker threads before fully shutting down the open62541 server instance.


jpfr:
Hey @xR3b0rn,

the idea to simplify the user application logic is great.

I ran this by an LLM and it gave me a few hints.

----

Verdict: request changes. The PR merges cleanly with current upstream `master`, but several ownership races remain.

### Findings

1. **[P1] Canceled results are consumed while the worker may still modify them.**  
   [`[directOpCallback](https://github.com/open62541/open62541/pull/8343/files)`](codex://review?pr=https%3A%2F%2Fgithub.com%2Fopen62541%2Fopen62541%2Fpull%2F8343&path=src%2Fserver%2Fua_server_async.c&line=265&side=right) and [`[sendAsyncResponse](https://github.com/open62541/open62541/pull/8343/files)`](codex://review?pr=https%3A%2F%2Fgithub.com%2Fopen62541%2Fopen62541%2Fpull%2F8343&path=src%2Fserver%2Fua_server_async.c&line=280&side=right) read/serialize the result before it is parked as a zombie. The code explicitly assumes the worker may still be writing. Retaining the allocation prevents freeing it, but does not prevent this concurrent read/write race. Cancellation needs separate response storage or a synchronization/ownership handoff before the result is consumed.

2. **[P1] Reentrant acknowledgement from the cancellation callback corrupts the queues.**  
   [`[UA_AsyncOperation_cancel](https://github.com/open62541/open62541/pull/8343/files)`](codex://review?pr=https%3A%2F%2Fgithub.com%2Fopen62541%2Fopen62541%2Fpull%2F8343&path=src%2Fserver%2Fua_server_async.c&line=85&side=right) invokes the user callback while the operation is still in `waitingOps`, then updates its status. If that callback calls `UA_Server_setAsync*Result`, the result path moves the operation; the outer cancellation path subsequently processes it again. A targeted timeout test aborted at `sendAsyncResponse` with `ar->opCountdown == 0` violated. The operation must be transitioned to a stable cancellation state before invoking user code.

3. **[P1] Several cancellation paths bypass zombie tracking entirely.**  
   The new tracking occurs only inside [`[processOperationResult](https://github.com/open62541/open62541/pull/8343/files)`](codex://review?pr=https%3A%2F%2Fgithub.com%2Fopen62541%2Fopen62541%2Fpull%2F8343&path=src%2Fserver%2Fua_server_async.c&line=325&side=right), but:

   - synchronous `UA_Server_cancelAsync(..., true)` still deletes the direct operation immediately;
   - session cancellation and self-closing-session cleanup unlink operations without adding zombies;
   - queue-overflow handling cancels and immediately releases/returns the storage.

   A synchronous variant of the PR’s direct-read test failed because `zombieOps` was empty and `opsCount` was already zero. These paths preserve the original use-after-free risk.

4. **[P2] The required late acknowledgement contradicts the public API contract.**  
   The public header still says canceled output memory is freed and “should not be touched afterwards” ([[server.h](https://github.com/open62541/open62541/blob/ac011438e66e734aade9d349d80f8cc2246f264d/include/open62541/server.h#L1471-L1480)](https://github.com/open62541/open62541/blob/ac011438e66e734aade9d349d80f8cc2246f264d/include/open62541/server.h#L1471-L1480)). Existing compliant workers therefore will not call `UA_Server_setAsync*Result` after cancellation. Direct zombies then retain memory and queue capacity until server deletion, eventually exhausting `maxAsyncOperationQueueSize`. The example also says late acknowledgement returns `BADNOTFOUND`, while the implementation returns `GOOD`.

5. **[P2] The tutorial accepts only eight calls during the server’s lifetime.**  
   [`[activeThreadCount](https://github.com/open62541/open62541/pull/8343/files)`](codex://review?pr=https%3A%2F%2Fgithub.com%2Fopen62541%2Fopen62541%2Fpull%2F8343&path=examples%2Ftutorial_server_method_async.c&line=160&side=right) only increments; completed threads are not joined or reclaimed until shutdown. After eight completed calls, every subsequent call returns `BadTooManyOperations`. Thread creation and `UA_String_copy` failures are also ignored.

----

So we still need to ensure there is no data race. 
Without a mutex, the async worker might be editing the results data structure when we encode the response message.

What we could do is
- First shallow copy the results array into the response message
- Iterate the async operations and null out all results entries in the response message that are not-yet-done.

Still there needs to be a detailed analysis to rule out data races.

Thanks for this!

xR3b0rn:
I see. I made some wrong assumptions in my patch. However, besides that, while digging into the code, I think I found another critical problem with the current implementation: a possible deadlock scenario.

Looking at the [example](https://github.com/open62541/open62541/blob/master/examples/tutorial_server_method_async.c):

```c
static void
asyncOperationCancelCallback(UA_Server *server, const void *out) {
    /* This blocks if a worker thread currently processes the async callback */
    UA_LOCK(&asyncLock);

    if((void*)workQueue[0] == out)
        workQueue[0] = NULL; /* Disable active async callback */

    UA_UNLOCK(&asyncLock);
}

static void
asyncCall(UA_Server *server, void *data) {
    UA_LOG_INFO(UA_Server_getConfig(server)->logging, UA_LOGCATEGORY_APPLICATION, "call");

    /* Process async result if still active */
    UA_LOCK(&asyncLock);
    if((void*)workQueue[0] == data)
        UA_Server_setAsyncCallMethodResult(server, workQueue[0], UA_STATUSCODE_GOOD);
    workQueue[0] = NULL;
    UA_UNLOCK(&asyncLock);
}

```

If `asyncCall` is called from a thread other than the server thread, a deadlock may occur:

1. `asyncCall` executes up to the line `if((void*)workQueue[0] == data)`, meaning it has acquired the `asyncLock`.
2. The server calls `asyncOperationCancelCallback`, which always occurs while holding the server lock. The server thread then tries to acquire the `asyncLock`, but because the other thread is holding it, the server must wait.
3. Meanwhile, the non-server thread calls `UA_Server_setAsyncCallMethodResult` and attempts to acquire the server lock, resulting in a deadlock.

In the provided example, this can never happen because `asyncCall` runs from within the event loop (which, incidentally, makes the lock pointless). However, in a real-world application where `UA_Server_setAsyncCallMethodResult` is called from a separate worker thread, this creates a classic deadlock resulting from lock inversion.

My patch will fix this (once I remove the data race scenarios).

xR3b0rn:
Committing as a draft for now. I've applied fixes for the mentioned data races and included AI-generated tests, but haven't had the time for a thorough review to be 100% confident that all concurrency issues are resolved or that the tests are fully accurate.

The basic idea now is to pass around the ownership between the application and the server. While the application owning the memroy, the servers uses a pointer to a dummy.

xR3b0rn:
This turned out to be much more complex than I initially anticipated. I essentially rewrote the entire implementation compared to my previous version, getting the code to about 95% complete. To close the gap to 100%, I used the following loop:

1. Have Claude review the code and flag potential bugs.
2. Have Claude write tests that trigger the bugs found.
3. Run the tests under ASan and Valgrind.
4. Fix the bugs iteratively until tests went green (during the first few iterations, Claude went in a pretty odd direction fairly often, so this required close supervision).
5. Open a new Claude session.
6. Repeat.

### Note on AI usage

~95% of the production code is handwritten by me, and the remainder came out of the loop described above. 100% of the tests were written by Claude.

### General note

I kept the semantics of the previous implementation intact wherever possible. The only semantic changes introduced relate to the API contract:
1. Writing to the buffer is now permitted even after an operation has been canceled.
2. The application must invoke the `setAsync*Result`-callback exactly once. Failing to do so will leave the operation in a zombie state and cause a memory leak, which is cleaned up only on server shtudown
3. The application must ensure that no external worker threads are still reading from or writing to the buffer before the application cleans up the open62541-server.

This brings significant benefits for applications using the async API. The biggest advantage is that developers no longer need to handle synchronization manually. Because the server lock is automatically held both when passing the buffer pointer to the application and when returning it to open62541, thread safety is guaranteed by design.

Some parts were hard to implement because I could no longer rely on the contract: *"you are not allowed to write into the aliased buffer, once the associated operation was cancelled."*, but still wanted to keep the old semantics. For example, I had to create a copy of the results array to keep the operations in a `Service`-request alive in case the session is already closed and some operations get zombified. Previously, this wasn't an issue, since calling a `Service_call` on a closed session triggered an immediate call to the application's notification callback.

However, as described in other comments, using the old open62541 async approach and implementing it correctly with a worker thread is extremely difficult. Even the official open62541 examples can introduce deadlock scenarios when translated to a real-world, multi-threaded environment. Because of this, I believe the simplification in the usage of the API fully justifies the added complexity on the implementation side.

@jpfr, kindly asking for a review.

jpfr:
I played with these ideas a bit.
Out came https://github.com/open62541/open62541/pull/8470
It even reduces the size of /src.

I'll separate out the drive-by fixes so the PR can be compared better.
Are you available for a quick call to consolidate?

xR3b0rn:
sure thing, I'll reach out to you by mail

PR #8358, opened 2026-08-31
examples: fix client_subscription_loop and add a robust reconnecting client + test

StalderT:
Three related changes around the reconnecting subscription client.

### 1. Fix `client_subscription_loop.c`

The example stopped connecting to the server. `8a34b4d7e` replaced the old `while { UA_Client_connect(...); run_iterate(); }` loop with `UA_Client_runUntilInterrupt()` but dropped the connect call, and `runUntilInterrupt` does not connect on its own (as its documentation states). This re-adds the async connect.

### 2. New example: `client_subscription_reconnect_async.c`

A non-blocking client that recovers on its own from every failure mode of a remote server:

- graceful shutdown / restart,
- abrupt disconnect,
- a **hung** server that keeps the TCP connection open but stops answering — only detectable through the `subscriptionInactivityCallback`.

`UA_Client_run_iterate()` drives the state machine; the (re)connection is issued from the main loop with `connectAsync` / `disconnectAsync`, and the subscription is re-created from the `stateCallback` on every activated session.

### 3. Regression test: `tests/client/check_client_reconnect.c`

Drives the same loop and asserts recovery after a server restart **and** after a hung server: reconnect, subscription re-created, and a write round-trip back through the subscription.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -98,6 +98,7 @@ endif()
 
 if(UA_ENABLE_SUBSCRIPTIONS)
     add_example(client_subscription_loop client_subscription_loop.c)
+    add_example(client_subscription_reconnect_async client_subscription_reconnect_async.c)
 endif()
 
 ####################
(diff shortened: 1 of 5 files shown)


jpfr:
Thanks, merged.

PR #8360, opened 2026-09-01
perf(client): skip the notification re-key scan when none is pending

StalderT:
findMonitoredItemForNotification falls back to a full linear scan of the MonitoredItem ziptree when a notification clientHandle is not found, to promote a pending re-key from a ModifyMonitoredItems. A malicious or buggy server can send a PublishResponse full of notifications with unknown clientHandles: each triggers the full O(n) scan, i.e. O(N*n) CPU burn on the single client event-loop thread, stalling reconnection and every other subscription.

Track the number of pending re-keys per subscription (pendingRekeys) and skip the scan when it is zero, which is the common case since a re-key is pending only briefly after a ModifyMonitoredItems. The counter is maintained where pendingParameters is set (modify), cleared (rollback, delete) and promoted (on the authoritative notification).

Extend Client_subscription_modifyMonitoredItem_doubleBuffer to assert the counter (0 -> 1 -> 0) and that a notification with an unknown handle is dropped without a scan while nothing is pending.

### DIFF
--- a/src/client/ua_client_internal.h
+++ b/src/client/ua_client_internal.h
@@ -63,6 +63,10 @@ typedef struct UA_Client_Subscription {
     UA_UInt32 sequenceNumber;
     UA_DateTime lastActivity;
     MonitorItemsTree monitoredItems;
+    size_t pendingRekeys; /* MonitoredItems with a pending clientHandle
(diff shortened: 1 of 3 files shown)


jpfr:
Thanks, merged. 👍

PR #8361, opened 2026-09-01
perf(server): O(1) MonitoredItem removal via the intrusive LIST macros

StalderT:
The per-node MonitoredItem back-pointer list (UA_NodeHead.monitoredItems) was a hand-rolled singly-linked list. Removing an item scanned the list to find the predecessor, so tearing down the MonitoredItems of a node was O(n^2).

Switch the list to the intrusive BSD-style LIST_* macros from deps/open62541_queue.h (already used for the subscription sampling list). Removal uses the le_prev back-pointer and is O(1); no search is needed.

The list head is part of the public UA_NodeHead in plugin/nodestore.h. It is declared as a plain struct { UA_MonitoredItem *lh_first; } (structurally a LIST_HEAD) instead of the LIST_HEAD macro, so the public header and the generated amalgamation (open62541.h) stay independent of open62541_queue.h.

UA_Node_copy copies the list head by value and re-anchors the first element's le_prev to the destination head, keeping the list consistent when a node is copied.

Add a regression test covering head, interior and tail removal, the on-write iterator, and the copy-on-write duplication of a Node that carries attached MonitoredItems (it asserts the *le_prev == element invariant, which detects a stale back-pointer after the copy).

Tearing down N MonitoredItems attached to a single node (samplingInterval 0) was O(n^2). Deleting them now scales linearly:

    N         delete before    delete after
    20000       669 ms           29 ms   (x23)
    50000      6477 ms           71 ms   (x92)
    80000     25756 ms          120 ms  (x215)
   100000     66072 ms          169 ms  (x391)

(amalgamation of each revision built identically with -O2 -DNDEBUG.)

### DIFF
--- a/src/server/ua_nodes.c
+++ b/src/server/ua_nodes.c
@@ -577,6 +577,13 @@ copyNode(const UA_Node *src, UA_Node *dst, UA_Boolean copyMonitoredItems) {
     dsthead->context = srchead->context;
     dsthead->constructed = srchead->constructed;
 #ifdef UA_ENABLE_SUBSCRIPTIONS
+    /* The MonitoredItem list head is opaque here: an actual copy (new NodeId)
+     * owns none, an editable replacement keeps the very same head by value.
+     * The intrusive list anchors the first element's le_prev in the node that
+     * holds the head. Nothing has to be re-anchored here because the server
+     * edits nodes in place (getEditNode), so a live node never moves. A
+     * nodestore that instead relocates a node must re-anchor le_prev onto the
+     * replacement after a successful replaceNode. */
     dsthead->monitoredItems = copyMonitoredItems ? srchead->monitoredItems : NULL;
 #else
     (void)copyMonitoredItems;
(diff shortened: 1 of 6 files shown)


include/open62541/plugin/nodestore.h
```diff
@@ -292,8 +292,14 @@ struct UA_NodeHead {
     void *context;
     UA_Boolean constructed; /* Constructors were called */
 #ifdef UA_ENABLE_SUBSCRIPTIONS
-    UA_MonitoredItem *monitoredItems; /* MonitoredItems for Events and immediate
```
jpfr, line 295: You can keep the public header as-is and internally cast to LIST_HEAD.

StalderT: Good idea, done. The public header is left untouched (no API/ABI change). The list is manipulated internally by casting the field to a LIST_HEAD.

jpfr:
There is a subtle bug in src/server/ua_nodes.c.

We have to differentiate between write-on-copy (replacement) and making an actual copy of a node (with a new NodeId).

This PR for 1.5 makes the differences explicit.
Please rebase on top of that and handle the fallout.
https://github.com/open62541/open62541/pull/8363

Thanks 👍

StalderT:
> There is a subtle bug in src/server/ua_nodes.c.
> 
> We have to differentiate between write-on-copy (replacement) and making an actual copy of a node (with a new NodeId).
> 
> This PR for 1.5 makes the differences explicit. Please rebase on top of that and handle the fallout. #8363
> 
> Thanks 👍

Done !

jpfr:
There was another change already on 1.5 which simplifies this one.
Merging this via #8367.
You remain author of the commit there.

PR #8373, opened 2026-09-03
feat(server): let applications tighten the message size limit on a channel

o6-nbeier:
Related to #8321, but instead of open62541 enforcing a fixed default limit on untrusted messages, adds a generic `SecureChannel` attribute map (mirroring the existing per-Session attributes) with `UA_Server_{get,set,delete}SecureChannelAttribute`. Writing the reserved `0:maxMessageSize` key caps `tcpMaxMsgSize` for that one channel, it can only tighten, never loosen, the configured ceiling. The application decides when to write it, using the already-existing `secureChannelNotificationCallback` rather than a new dedicated callback.

Also includes unit/integration tests covering the attribute API and the size-limit enforcement.

### DIFF
--- a/include/open62541/server.h
+++ b/include/open62541/server.h
@@ -237,6 +237,59 @@ UA_Server_closeSecureChannel(UA_Server *server, UA_UInt32 channelId,
                              UA_ShutdownReason reason);
 
 /**
+ * A SecureChannel carries application-defined attributes in a key-value map,
+ * mirroring the :ref:`session attributes <server-session-handling>` below.
+ * This lets the application attach its own per-channel state, for example
+ * from the secureChannelNotificationCallback once it has looked at the
+ * channel's SecurityMode.
+ *
+ * One attribute key is interpreted by the server itself:
+ *
+ * - ``0:maxMessageSize`` (``UA_UInt32``): If set, tightens (but never
+ *   loosens) UA_ServerConfig.tcpMaxMsgSize for this specific channel. Useful
+ *   to cap messages on a channel the application does not yet consider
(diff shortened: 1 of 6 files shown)


jpfr:
1) This is good. But the following comment for _getScalar is misleading:
> Returns NULL if the attribute is not defined or not a scalar or not of the right datatype.

2) Extend the exposed attributes to include what we have here:
https://github.com/open62541/open62541/blob/6021f825f8a8d87798b30d1d1a1bdc77332f74f4/include/open62541/common.h#L221

Most of those will be readable, but not writeable.

3) Add an API to manually close a securechannel via its id.

mcpat-pilz:
This approach is not optimal. A key-value map is very flexible and extensible - which can be good. But it is also vulnerable for typos - which in this case is bad.

If one fails to hit the key for modifying this setting either because of sloppyness or because the key is renamed it goes by unnotified. And this leaves the server vulnerable to DoS situations from untrusted sources.

Instead I would recommend to introduce a public SecureChannel configuration structure which can be adjusted by the application when being notified about changes of the SecureChannel's state. The compiler will then properly handle typos or renames.

jpfr:
I understand that you value type-safety.
But we have also heard the complete opposite - that people value the directness of kv-maps.
Not only do we have easy extensibility, we also allow applications to define a convention for their own metadata they then attach to a SecureChannel.
And this is now a pattern we use in many places that are facing the "userland".

jpfr:
- For key-QualifiedName in ns0, only allow the predefined set. For both read and write. Other ns are free to set.
- Dont prepopulate the kv map. Generate / copy the values from the SecureChannel on access.

PR #8375, opened 2026-09-03
1.4+: fix(historydata): prevent UAF in memory backend range-limited history…

o6-nbeier:
… reads

UA_DataValue_backend_copyRange() performed a shallow memcpy of the source UA_DataValue before attempting the range copy. Because the destination's UA_Variant fields were not reinitialized first, a failing UA_Variant_copyRange() (for example when the client-supplied indexRange exceeds UA_MAX_ARRAY_DIMS) left the destination value as an uninitialized alias of the backend-owned storage instead of an independent copy.

copyDataValues_backend_memory() also discarded the status code from both UA_DataValue_backend_copyRange() and UA_DataValue_copy(), so the aliased, partially-copied value was still counted as a successful result and forwarded to the HistoryRead response. Freeing that response later released the backend's live storage, leaving the backend with a dangling pointer that was dereferenced on a subsequent HistoryRead.

Reinitialize the destination variant before the range copy and clear it again on failure, mirroring the safe pattern already used by UA_DataValue_copyRange(). Propagate copy failures out of copyDataValues_backend_memory() immediately instead of counting a failed entry as provided.

Internal Vulnerability Advisory: open62541-SA-2026-0021
Reported by Chuan Wei [SCU]

### DIFF
--- a/plugins/historydata/ua_history_data_backend_memory.c
+++ b/plugins/historydata/ua_history_data_backend_memory.c
@@ -322,9 +322,15 @@ UA_DataValue_backend_copyRange(const UA_DataValue *src, UA_DataValue *dst,
                                const UA_NumericRange range)
 {
     memcpy(dst, src, sizeof(UA_DataValue));
-    if (src->hasValue)
-        return UA_Variant_copyRange(&src->value, &dst->value, range);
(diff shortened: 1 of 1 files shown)


jpfr:
Superseded by #8378

PR #8379, opened 2026-09-05
server: sync cached logger pointers in UA_ServerConfig_setLoggerserver: add UA_ServerConfig_setLogger to sync cached logger pointers

### LINKED ISSUE #7307: Logger in eventloop is copied from the client or server configuration
Currently when creating a default Server or Client configuration, the according eventloop is also initialized and during that the current logger is copied into the eventloop by value. This goes however wrong when someone wants to change the logger in the client or server later on. The logger in the config is cleared with log->clear() to avoid a memory leak and then a new logger is initialized. The logger in the eventloop however still contains the old value which is now invalid and causes a crash the first time the eventloop tries to log something.

Ideally the logger is not copied into the eventloop but rather a reference to it. Or there could be a helper function for the server and client to update the logger. Or at least mention in the documentation that the logger in the config structure can't be just replaced once the configure structure has been initialized with one of the setDefault() or similar functions.

ParthTiwari2599:
This adds `UA_ServerConfig_setLogger()` as discussed in #7307.

## Problem
`config->logging` is copied by value into several places when the config is
first initialized:
- `config->eventLoop->logger` (set once in `UA_EventLoop_new_POSIX()` and
  the other architecture backends)
- `config->secureChannelPKI.logging`
- `config->sessionPKI.logging`

None of these are updated if the application later replaces
`config->logging` (e.g. `config->logging->clear(config->logging)` followed
by assigning a new logger). This leaves stale pointers to freed memory,
causing a use-after-free the next time the EventLoop or a CertificateGroup
logs something.

## Fix
`UA_ServerConfig_setLogger()` frees the old logger and updates every cached
copy in a single call, so the application doesn't need to know about each
internal cache point.

I've tested this locally: without the fix, replacing the logger and then
calling `UA_Server_run_startup()` reliably segfaults. With
`UA_ServerConfig_setLogger()` used instead of a raw field assignment,
startup completes cleanly and the EventLoop's logger pointer matches the
new one.

Open to feedback on the approach -- happy to adjust (e.g. add an
equivalent `UA_ClientConfig_setLogger()`) if preferred.

Closes #7307

### DIFF
--- a/plugins/include/open62541/server_config_default.h
+++ b/plugins/include/open62541/server_config_default.h
@@ -129,6 +129,17 @@ UA_ServerConfig_setDefault(UA_ServerConfig *config) ,{
  *
  * @param conf The configuration to manipulate
  */
+
+/* Replace the logger in an already-configured UA_ServerConfig. This
+ * correctly frees the old logger and keeps every cached copy of the
+ * logger pointer in sync (the EventLoop and the certificate-verification
+ * groups), so nothing is left pointing at freed memory.
+ *
+ * @param config The configuration to manipulate
+ * @param newLogger The new logger to install
+ */
+UA_EXPORT UA_StatusCode
+UA_ServerConfig_setLogger(UA_ServerConfig *config, UA_Logger *newLogger);
 UA_EXPORT UA_StatusCode
 UA_ServerConfig_setBasics(UA_ServerConfig *conf);
 
--- a/plugins/ua_config_default.c
+++ b/plugins/ua_config_default.c
@@ -573,6 +573,38 @@ setDefaultConfig(UA_ServerConfig *conf, UA_UInt16 portNumber) {
     return UA_STATUSCODE_GOOD;
 }
 
+
+/* Replace the logger in an already-configured UA_ServerConfig. This
+ * correctly frees the old logger and keeps every cached copy of the
+ * logger pointer in sync (the EventLoop and the certificate-verification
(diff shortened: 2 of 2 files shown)


jpfr:
Looks good.
Please do a sweep for existing places where we should reuse this.

for example in the json config parsing/setting.
And in the examples and tests.

ParthTiwari2599:
Addressed the review feedback:

Swept remaining direct ->logging = assignments in examples/JSON parsing/tests (ClientConfig usages correctly left untouched)
Fixed a use-after-free bug in the JSON config loader where a fresh logger was silently replacing the user's already-configured one
Added UA_ServerConfig_setLogger() with guards against a NULL clear callback and self-assignment (replacing the logger with itself)
Fixed examples/server_loglevel.c, which was passing a stack-allocated logger into the ownership-taking setter — switched to UA_Log_Stdout_new() for a proper heap allocation
Added a unit test (setLogger_syncsCachedPointers) covering the EventLoop/PKI sync, fixed a duplicate test registration

All local tests pass (check_server_config_default, check_server_config_json), and server_loglevel builds and runs cleanly.

ParthTiwari2599:
Hi @jpfr ,

Just following up on this PR. I've addressed your feedback by sweeping the remaining direct logger assignments across examples, tests, and JSON parsing, along with adding unit test coverage.

All local tests pass and the CLA is signed. Could you please take a look when you have a moment or trigger the workflow approval so CI can run?

Thanks!

PR #8384, opened 2026-09-05
Fix NULL dereference in EventFilter parser on allocation failure

### LINKED ISSUE #7675: NULL Pointer Dereference in EventFilter Parser causes SEGV
## Summary

The EventFilter parser's `newOperand()` function does not check if memory allocation succeeds before using the pointer. When `UA_calloc()` returns NULL (under memory pressure), `LIST_INSERT_HEAD` dereferences the NULL pointer, causing a SEGV crash.

## Root Cause

The vulnerable code in `src/util/ua_eventfilter_parser.c:27-33`:

```c
static Operand *
newOperand(EFParseContext *ctx) {
    Operand *op = (Operand*)UA_calloc(1, sizeof(Operand));
    LIST_INSERT_HEAD(&ctx->operands, op, entries);  // BUG: op may be NULL!
    ctx->operandsSize++;
    return op;
}
```

The `LIST_INSERT_HEAD` macro (from `sys/queue.h`) expands to access `op->entries.le_prev`, which dereferences NULL when allocation fails.

## Fuzzer

```cpp
#include "custom_memory_manager.h"

#include <open62541/types.h>
#include <open62541/util.h>

extern "C" int
LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    if(size < 4)
        return 0;

    /* Use last 4 bytes to set memory limit for OOM testing */
    if(!UA_memoryManager_setLimitFromLast4Bytes(data, size))
        return 0;
    size -= 4;

    UA_EventFilter filter;
    UA_EventFilter_init(&filter);

    UA_EventFilterParserOptions options;
    memset(&options, 0, sizeof(options));

    UA_ByteString content;
    content.data = (UA_Byte*)(void*)data;
    content.length = size;

    UA_StatusCode rv = UA_EventFilter_parse(&filter, content, &options);
    if(rv == UA_STATUSCODE_GOOD)
        UA_EventFilter_clear(&filter);

    return 0;
}
```

### Crash input (hex)

```
23 31 00 00 00
```

This is `#1` followed by 3 null bytes (memory limit marker for fuzzer).

### How to generate crash_input.bin

```bash
printf '#1\x00\x00\x00' > crash_input.bin
```

---

## Reproduction

```bash
cd /path/to/open62541/build
./bin/fuzz_parse_eventfilter crash_input.bin
```

**Output:**
```
==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000050
The signal is caused by a WRITE memory access.
Hint: address points to the zero page.
    #0 newOperand /src/util/ua_eventfilter_parser.c:30:5
    #1 create_operand /src/util/ua_eventfilter_parser.c:103:19
    #2 UA_EventFilter_lex /src/util/ua_eventfilter_lex.c:8506:14
    #3 UA_EventFilter_parse /src/util/ua_eventfilter_grammar.c:1415:19
```

---

## Impact

| Aspect | Details |
|--------|---------|
| **Type** | Denial of Service (DoS) |
| **Severity** | Medium |
| **Attack Vector** | Malformed EventFilter expression under memory pressure |
| **Affected API** | `UA_EventFilter_parse()` (public, `UA_EXPORT`) |
| **CWE** | CWE-476 (NULL Pointer Dereference) |

---

## Suggested Fix

Add NULL check after memory allocation:

```c
static Operand *
newOperand(EFParseContext *ctx) {
    Operand *op = (Operand*)UA_calloc(1, sizeof(Operand));
    if(!op)
        return NULL;  // Handle allocation failure
    LIST_INSERT_HEAD(&ctx->operands, op, entries);
    ctx->operandsSize++;
    return op;
}
```

Callers of `newOperand()` (e.g., `create_operand()` at line 103) should also check for NULL return values and propagate the error.

---

## Contribution

This issue was discovered by **FuzzingBrain @O2Lab**, an AI-based fuzzing system. I'm happy to submit a pull request with the fix if that would be helpful.

afonsojanu:
Fixes #7675.

`newOperand()` in `ua_eventfilter_parser.c` calls `UA_calloc` and hands the result straight to `LIST_INSERT_HEAD` without checking it. When the allocation fails, that macro dereferences a NULL pointer. The two wrappers built on top of it, `create_operand()` and `create_operator()`, have the same problem one level up, and so do all five places in the lexer (`ua_eventfilter_lex.re` / the generated `.c`) that call `create_operand()` and immediately write into the operand it returns.

I added the missing checks along that whole chain: `newOperand()` now sets `ctx->error = UA_STATUSCODE_BADOUTOFMEMORY` and returns NULL instead of inserting garbage into the list, `create_operand()`/`create_operator()` bail out early if the operand is NULL, and each of the five lexer sites sets `tokenId = 0` and jumps to `finish` in that case, which is exactly the existing convention this file already uses for a failed parse.

While in there I noticed `save_string()` has the identical calloc-then-use pattern (calloc followed by an unconditional memcpy), a few lines above these functions, so I gave it the same one-line guard.

To verify, I built the library with `UA_ENABLE_MALLOC_SINGLETON` on and wrote a small standalone program linking against the repo's own `tests/fuzz/custom_memory_manager.c`, which lets you cap available memory and force allocations to fail at a chosen point. Parsing a normal, valid filter string (`SELECT $1 WHERE $1 := 1 == 1`) with the memory cap set low enough to fail partway through:

- Before this change: segfaults inside `newOperand`, same stack as the one in the issue (`newOperand` -> `create_operand` -> `UA_EventFilter_lex` -> `UA_EventFilter_parse`).
- After this change: `UA_EventFilter_parse` returns `BadInternalError` and nothing crashes.

I also ran the existing `check_eventfilter_parser` suite against the fix, still 21/21.

This is against the `1.5` branch per the request in the issue thread.

### DIFF
--- a/src/util/ua_eventfilter_lex.c
+++ b/src/util/ua_eventfilter_lex.c
@@ -8453,11 +8453,13 @@ binary_op:
 
 make_op:
     *token = create_operand(ctx, OT_OPERATOR);
+    if(!*token) { tokenId = 0; goto finish; }
     (*token)->operand.op.filter = f;
     goto finish;
 
 namedoperand:
     *token = create_operand(ctx, OT_REF);
+    if(!*token) { tokenId = 0; goto finish; }
     size_t nameSize = (uintptr_t)(pos - b);
     (*token)->operand.ref = (char*)UA_malloc(nameSize+1);
     memcpy((*token)->operand.ref, b, nameSize);
@@ -8470,6 +8472,7 @@ json:
     match.length = (uintptr_t)(end-b);
     match.data = (UA_Byte*)(uintptr_t)b;
     *token = create_operand(ctx, OT_LITERAL);
+    if(!*token) { tokenId = 0; goto finish; }
     UA_DecodeJsonOptions options;
     memset(&options, 0, sizeof(UA_DecodeJsonOptions));
     size_t jsonOffset = 0;
@@ -8485,6 +8488,7 @@ lit:
     match.length = (uintptr_t)(pos-b);
     match.data = (UA_Byte*)(uintptr_t)b;
     *token = create_operand(ctx, OT_LITERAL);
+    if(!*token) { tokenId = 0; goto finish; }
     (*token)->operand.literal.data = UA_new(lt);
     (*token)->operand.literal.type = lt;
     if(lt == &UA_TYPES[UA_TYPES_NODEID]) {
@@ -8504,6 +8508,7 @@ sao:
     match.length = (uintptr_t)(pos-b);
     match.data = (UA_Byte*)(uintptr_t)b;
     *token = create_operand(ctx, OT_SAO);
+    if(!*token) { tokenId = 0; goto finish; }
     res = UA_SimpleAttributeOperand_parse(&(*token)->operand.sao, match);
     tokenId = (res == UA_STATUSCODE_GOOD) ? EF_TOK_SAO : 0;
 
--- a/src/util/ua_eventfilter_lex.re
+++ b/src/util/ua_eventfilter_lex.re
@@ -170,11 +170,13 @@ binary_op:
 
 make_op:
     *token = create_operand(ctx, OT_OPERATOR);
+    if(!*token) { tokenId = 0; goto finish; }
     (*token)->operand.op.filter = f;
     goto finish;
 
 namedoperand:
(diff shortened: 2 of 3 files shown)


jpfr:
Thanks. Please also click the CLA.

PR #8392, opened 2026-09-07
fix(xml): escape the markup delimiters in string content

### LINKED ISSUE #8356: XML encoding produces unescaped strings that decoding fails to decode
`"<Variant><Value><String><![CDATA[<]]></String></Value></Variant>"` fails to roundtrip

This bug was found by Claude, which produced the following long issue report, description, and patch.

<details><summary>Claude's report</summary>

## Description

The XML encoder (`UA_ENABLE_XML_ENCODING`) writes `UA_String` content verbatim, with no XML-metacharacter escaping, while the XML decoder is a strict yxml-based tokenizer that rejects malformed XML. As a consequence, a `decode → encode → decode` round-trip of a Variant carrying a String is not stable whenever the string value contains a raw `<` or a bare `&`: the re-encoded buffer is not well-formed XML and the second decode fails.

Mechanism:

- The decoder captures element content as a raw byte span of the original input buffer (tokenizer loop in `src/ua_types_encoding_xml.c`, `YXML_CONTENT` handling); yxml is used as a validating tokenizer, and no entity/CDATA unescaping is applied on the String path. Since yxml reports `YXML_CONTENT` only for the payload bytes inside a `<![CDATA[...]]>` section, decoding `<String><![CDATA[<]]></String>` yields a `UA_String` whose content is the single literal byte `<`.
- On the way back out, `ENCODE_XML(String)` in `src/ua_types_encoding_xml.c` writes the string bytes via `xmlEncodeWriteChars`, which is a plain `memcpy` with no escaping. The encoder therefore emits `<String><</String>`.
- Re-decoding that buffer fails: the raw `<` opens a bogus element, yxml signals an error, and `UA_decodeXml` returns a non-GOOD status.

The same applies to every field that routes through the String codec, not just a bare `String` Variant: `QualifiedName`'s `Name`, `LocalizedText`'s `Locale` and `Text`, and `DiagnosticInfo`'s `AdditionalInfo` are all encoded and decoded as `UA_TYPES_STRING` and share the trigger condition. The CDATA form is the reliable trigger, because it is what places a genuine metacharacter into the decoded value; entity references do not trigger it, for a separate reason described below.

### Related: entity references in String content are silently corrupted

While investigating the above, a second and arguably worse defect in the same raw-span scheme turned up. yxml decodes an entity reference into `x->data` and signals `YXML_CONTENT` only once, at the closing `;` (`deps/yxml.c`: `yxml_refstart` and `yxml_ref` both return `YXML_OK`; only `yxml_refcontent` returns `YXML_CONTENT`). Because the tokenizer in `src/ua_types_encoding_xml.c` records `val_begin`/`end` from the *input position* of that event and never reads yxml's decoded `x->data`, the resulting span is just the `;`:

- `<String>&lt;</String>` decodes to a one-byte `UA_String` containing `;` — not `<`.
- With surrounding text, `<String>x&lt;y</String>` decodes to the literal six bytes `x&lt;y` — the entity is left unexpanded rather than decoded to `x<y`.
- Content spanning two CDATA sections keeps the intervening markup: `<String><![CDATA[a]]]]><![CDATA[>b]]></String>` should decode to `a]]>b`, but the span captures the raw source `a]]]]><![CDATA[>b` (17 bytes). Because that contains a `<`, this one does also break the round trip.

Both were confirmed against the same Debug build (`&lt;` and `&amp;` each yield a one-byte `;`; `x&lt;y` yields six bytes). Both round-trip *stably* (`;` re-encodes and re-decodes as `;`), so the round-trip oracle in the fuzz harness is structurally blind to them — the value is simply wrong, consistently. This is a decode-side value-correctness bug, distinct from the encode-side escaping bug above, though both stem from the decoder using raw input spans instead of yxml's decoded output. Happy to split this into its own issue if you prefer.

For comparison, the JSON encoder does not have this problem: `writeJsonStringContent` in `src/ua_types_encoding_json.c` escapes quotes, backslashes, and control characters. The XML encoder never got an equivalent escaper — the `memcpy` write path is unchanged since the XML encoder was introduced (f54f1961b, #5437); the strict yxml decoder that makes the mismatch observable landed later with #7140.

Severity, stated honestly: per the comment in `include/open62541/types.h`, the XML encoding extensions are not intended for the OPC UA protocol on the network but for configuration file formats carrying data in the OPC UA type system. So this is not a network-protocol issue — but it does break the feature's documented use case: a String value containing `<` or `&` is written out in a form the library itself cannot read back. In a Debug / OSS-Fuzz build (`UA_DEBUG` on), the existing fuzz target `tests/fuzz/fuzz_xml_decode_encode.cc` hits this and aborts on its round-trip `UA_assert`s. In a release build those asserts are compiled out and the corruption is silent: `UA_encodeXml` returns `UA_STATUSCODE_GOOD` but produces output that `UA_decodeXml` rejects.

### Why the existing fuzz target has not caught this

Worth noting since `fuzz_xml_decode_encode` has been running on OSS-Fuzz since March 2025 with these asserts live. The oracle is armed and working — in the 2026-08-30 per-target coverage report, 1,520 of 6,650 executions reach and pass all four round-trip asserts, and `ENCODE_XML(String)` itself runs 372 times. The bug needs a specific *value*, not an unreached *path*, so coverage guidance gives no gradient.

Concretely, in the published ClusterFuzz corpus (6,435 units) 94 files contain `String` and 80 contain `CDATA`, but **none contain both** — and combining them earns no new edge, since a `<String>` wrapping a harmless CDATA section executes exactly the same code as one wrapping plain text. Two things would likely close that gap:

- Adding CDATA tokens to `tests/fuzz/fuzz_xml_decode_encode.dict`. The dictionary currently supplies tag scaffolding (`<Variant>`, `<Value>`, `<String>`) but no CDATA or entity tokens, so the payload half of the trigger has to be invented by chance. A single combined token such as `"<![CDATA[<]]>"` is more effective than separate `"<![CDATA["` and `"]]>"` entries, since it needs no coordinated placement of two insertions. Rough modelling suggests this moves time-to-trigger from "not in practice" to minutes-to-hours on a single core.
- Adding a `tests/fuzz/fuzz_xml_decode_encode_corpus/` seed directory. The target currently ships no seed corpus (`oss-fuzz-copy.sh` only zips one when that directory exists), so its corpus is entirely fuzzer-derived and never inherited realistic OPC UA XML — which is exactly the kind of data that routinely carries escaped text.

Self-contained example (link against a plain build of master with `UA_ENABLE_XML_ENCODING` on, which is the default):

```c
#include <open62541/types.h>
#include <stdio.h>
#include <string.h>

int main(void) {
    const char *xml =
        "<Variant><Value><String><![CDATA[<]]></String></Value></Variant>";

    UA_ByteString in;
    in.data = (UA_Byte*)(uintptr_t)xml;
    in.length = strlen(xml);

    /* First decode: succeeds, value is the one-byte string "<" */
    UA_Variant v;
    UA_Variant_init(&v);
    UA_StatusCode rc = UA_decodeXml(&in, &v, &UA_TYPES[UA_TYPES_VARIANT], NULL);
    printf("decode #1: %s\n", UA_StatusCode_name(rc));

    /* Re-encode: succeeds, but emits the '<' unescaped */
    size_t n = UA_calcSizeXml(&v, &UA_TYPES[UA_TYPES_VARIANT], NULL);
    UA_ByteString out = UA_BYTESTRING_NULL;
    UA_ByteString_allocBuffer(&out, n);
    rc = UA_encodeXml(&v, &UA_TYPES[UA_TYPES_VARIANT], &out, NULL);
    printf("encode:    %s\n", UA_StatusCode_name(rc));
    printf("output:    %.*s\n", (int)out.length, out.data);
    /* -> <Variant><Value><String><</String></Value></Variant>  (not well-formed) */

    /* Second decode of the encoder's own output: fails */
    UA_Variant v2;
    UA_Variant_init(&v2);
    rc = UA_decodeXml(&out, &v2, &UA_TYPES[UA_TYPES_VARIANT], NULL);
    printf("decode #2: %s\n", UA_StatusCode_name(rc));

    UA_Variant_clear(&v);
    UA_Variant_clear(&v2);
    UA_ByteString_clear(&out);
    return 0;
}
```

Expected: the encoder escapes the metacharacter (or emits CDATA), so decode #2 returns `UA_STATUSCODE_GOOD` and the round-trip is stable.

Actual output, from running the above against a Debug build of master:

```
decode #1: Good
  decoded value: len=1 bytes=[<]
encode:    Good
output:    <Variant><Value><String><</String></Value></Variant>
decode #2: BadDecodingError
```

The existing fuzz target reproduces it directly too. Compiling `tests/fuzz/fuzz_xml_decode_encode.cc` against a `CMAKE_BUILD_TYPE=Debug` build and feeding it the same 64 bytes aborts:

```
Assertion `retval == ((UA_StatusCode) 0x00000000)' failed.
fuzz_xml_decode_encode.cc:51
Aborted (rc 134)
```

Line 51 is the `UA_assert` after the *second* decode. A plain-text control (`<String>hello</String>`) returns normally, and the following variants were checked against the same build:

| decoded value | round trip |
|---|---|
| `<` (via `<![CDATA[<]]>`) | **fails** — `BadDecodingError` |
| `&` (via `<![CDATA[&]]>`) | **fails** — `BadDecodingError` |
| same `<` inside `LocalizedText`/`Text` | **fails** |
| same `<` inside `QualifiedName`/`Name` | **fails** |
| `a]]b` | round-trips |
| `hello` | round-trips |

So the trigger set is a raw `<` or a bare `&` in the decoded value. (`]` and `]]` are harmless; yxml's content state treats `]` as an ordinary character.)

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_XML_ENCODING=ON -DCMAKE_BUILD_TYPE=Debug ..
```

`UA_ENABLE_XML_ENCODING` is ON by default but named here for unambiguous reproduction. `CMAKE_BUILD_TYPE=Debug` enables `UA_DEBUG` and thus the round-trip asserts in the fuzz harness (this is also the OSS-Fuzz configuration). With a release build type the defect is still present but silent: encode reports GOOD, re-decode fails.

## Proposed fix

A patch is below. It was written against `12b72510`, applies cleanly, builds
without warnings, and I verified it as described under "Verification" after the
diff.

Escaping the encoder alone is *not* sufficient, which is the main design point.
Because the decoder captures a raw input span rather than yxml's decoded output,
emitting `&lt;` would make the output well-formed but decode back to `;` — the
round-trip would then fail at the `UA_order` assert (`:53`) instead of the
decode assert (`:51`), turning a visible error into silent value corruption. So
the patch fixes both sides:

1. `ENCODE_XML(String)` routes through a new `xmlEncodeWriteEscapedChars`
   helper escaping `<`, `&` and `>`. A *separate* helper is required because
   `xmlEncodeWriteChars` also writes tag names, enum member names, booleans,
   digits, `NaN`/`INF`, `null`, DateTime and base64, none of which may be
   escaped. `XmlElement` deliberately keeps the raw writer.
2. The tokenizer backtracks from a character reference's terminating `;` to its
   `&`, so the reference is inside the captured content span.
3. `DECODE_XML(String)` unescapes the five predefined entities and numeric
   character references.
4. The ExtensionObject XML body switches from the String codec to the
   XmlElement codec, since that body is markup and must not be escaped.

Because the new writer is built on `xmlEncodeWriteChars`, which bounds-checks
and advances `ctx->pos` in both modes, `UA_calcSizeXml` (`calcOnly = true`) and
`UA_encodeXml` count identical bytes — worth stating explicitly, since escaping
that applied only on the real pass would under-allocate.

```diff
--- a/src/ua_types_encoding_xml.c
+++ b/src/ua_types_encoding_xml.c
@@ -18,6 +18,7 @@
 #include "../deps/base64.h"
 #include "../deps/libc_time.h"
 #include "../deps/dtoa.h"
+#include "../deps/utf8.h"
 #include "../deps/yxml.h"
 
 /* Replicate yxml_isNameStart and yxml_isName from yxml. But differently we
@@ -106,8 +107,25 @@
         }
         case YXML_CONTENT:
         case YXML_ATTRVAL:
-            if(val_begin == 0)
+            if(val_begin == 0) {
                 val_begin = pos;
+                /* yxml signals a character reference in a single event at its
+                 * terminating ';'. Backtrack to the '&' so that the raw
+                 * reference is contained in the content and can be unescaped
+                 * during decoding. */
+                if(xml[pos] == ';') {
+                    unsigned amp = pos;
+                    while(amp > 0) {
+                        char rc = xml[amp-1];
+                        if((rc < 'a' || rc > 'z') && (rc < 'A' || rc > 'Z') &&
+                           (rc < '0' || rc > '9') && rc != '#')
+                            break;
+                        amp--;
+                    }
+                    if(amp > 0 && amp < pos && xml[amp-1] == '&')
+                        val_begin = amp - 1;
+                }
+            }
             stack[top]->end = pos;
             break;
         case YXML_ELEMEND:
@@ -200,6 +218,33 @@
     return UA_STATUSCODE_GOOD;
 }
 
+/* Escape XML metacharacters in the content of String values. Element names and
+ * preformatted values (numbers, DateTime, base64, ...) are written with
+ * xmlEncodeWriteChars directly. The escaped bytes are also counted when
+ * ctx->calcOnly is set, so UA_calcSizeXml matches the actual encoding. */
+static status UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT
+xmlEncodeWriteEscapedChars(CtxXml *ctx, const char *c, size_t len) {
+    status ret = UA_STATUSCODE_GOOD;
+    size_t start = 0;
+    for(size_t i = 0; i < len && ret == UA_STATUSCODE_GOOD; i++) {
+        const char *esc;
+        if(c[i] == '<')
+            esc = "&lt;";
+        else if(c[i] == '>')
+            esc = "&gt;";
+        else if(c[i] == '&')
+            esc = "&amp;";
+        else
+            continue;
+        ret |= xmlEncodeWriteChars(ctx, &c[start], i - start);
+        ret |= xmlEncodeWriteChars(ctx, esc, strlen(esc));
+        start = i + 1;
+    }
+    if(start < len)
+        ret |= xmlEncodeWriteChars(ctx, &c[start], len - start);
+    return ret;
+}
+
 static status UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT
 writeXmlElemNameBegin(CtxXml *ctx, const char* name) {
     if(ctx->depth >= UA_XML_ENCODING_MAX_RECURSION - 1)
@@ -361,7 +406,7 @@
 /* String */
 ENCODE_XML(String) {
     const UA_String *src = (const UA_String*)src_;
-    return xmlEncodeWriteChars(ctx, (const char*)src->data, src->length);
+    return xmlEncodeWriteEscapedChars(ctx, (const char*)src->data, src->length);
 }
 
 /* XmlElement */
@@ -505,8 +550,8 @@
         if(src->encoding == UA_EXTENSIONOBJECT_ENCODED_BYTESTRING)
            ret |= writeXmlElement(ctx, "ByteString", &src->content.encoded.body,
                                   &UA_TYPES[UA_TYPES_BYTESTRING]);
-        else
-            ret |= ENCODE_DIRECT_XML(&src->content.encoded.body, String);
+        else /* The body is XML markup and must not be escaped */
+            ret |= ENCODE_DIRECT_XML(&src->content.encoded.body, XmlElement);
         ret |= writeXmlElemNameEnd(ctx, UA_XML_EXTENSIONOBJECT_BODY);
     } else {
         /* Write the decoded value */
@@ -1096,6 +1141,74 @@
     return res;
 }
 
+/* Unescape character references in String content. Returns the unescaped
+ * length; writes the unescaped bytes if out is set. The tokenizer (yxml) has
+ * already validated all references outside of CDATA sections, so anything that
+ * does not parse as a reference is kept verbatim. Every replacement shortens
+ * the content, so the unescaped length never exceeds the input length. */
+static size_t
+unescapeXmlString(const UA_Byte *data, size_t length, UA_Byte *out) {
+    size_t pos = 0;
+    for(size_t i = 0; i < length; i++) {
+        unsigned cp = 0;
+        size_t refEnd = i;
+        if(data[i] == '&') {
+            /* Find the terminating ';'. Valid references have at most eight
+             * characters between '&' and ';'. */
+            refEnd = i + 1;
+            while(refEnd < length && refEnd - i <= 9 && data[refEnd] != ';')
+                refEnd++;
+        }
+        if(refEnd > i && refEnd < length && data[refEnd] == ';') {
+            const UA_Byte *ref = &data[i+1];
+            size_t refLen = refEnd - i - 1;
+            if(refLen == 2 && memcmp(ref, "lt", 2) == 0)
+                cp = '<';
+            else if(refLen == 2 && memcmp(ref, "gt", 2) == 0)
+                cp = '>';
+            else if(refLen == 3 && memcmp(ref, "amp", 3) == 0)
+                cp = '&';
+            else if(refLen == 4 && memcmp(ref, "apos", 4) == 0)
+                cp = '\'';
+            else if(refLen == 4 && memcmp(ref, "quot", 4) == 0)
+                cp = '"';
+            else if(refLen >= 2 && ref[0] == '#') {
+                /* Numeric character reference, decimal or hexadecimal */
+                UA_Boolean hex = (ref[1] == 'x');
+                size_t k = hex ? 2 : 1;
+                UA_Boolean valid = (k < refLen);
+                unsigned v = 0;
+                for(; k < refLen && valid; k++) {
+                    UA_Byte d = ref[k];
+                    if(d >= '0' && d <= '9')
+                        v = v * (hex ? 16 : 10) + (unsigned)(d - '0');
+                    else if(hex && (d|32) >= 'a' && (d|32) <= 'f')
+                        v = v * 16 + (unsigned)((d|32) - 'a' + 10);
+                    else
+                        valid = false;
+                    if(v > 0x10FFFF)
+                        valid = false;
+                }
+                if(valid)
+                    cp = v;
+            }
+        }
+        if(cp > 0 && cp <= 0x10FFFF) {
+            unsigned char utf8[4];
+            unsigned utf8Len = utf8_from_codepoint(utf8, cp);
+            if(out)
+                memcpy(&out[pos], utf8, utf8Len);
+            pos += utf8Len;
+            i = refEnd;
+            continue;
+        }
+        if(out)
+            out[pos] = data[i];
+        pos++;
+    }
+    return pos;
+}
+
 DECODE_XML(String) {
     UA_String *dst = (UA_String*)dst_;
     CHECK_DATA_BOUNDS;
@@ -1109,8 +1222,19 @@
         return UA_STATUSCODE_GOOD;
     }
 
-    UA_String str = {length, (UA_Byte*)(uintptr_t)data};
-    return UA_String_copy(&str, dst);
+    /* No character reference was replaced? Every replacement shortens. */
+    size_t unescapedLength = unescapeXmlString(data, length, NULL);
+    if(unescapedLength == length) {
+        UA_String str = {length, (UA_Byte*)(uintptr_t)data};
+        return UA_String_copy(&str, dst);
+    }
+
+    dst->data = (UA_Byte*)UA_malloc(unescapedLength);
+    if(!dst->data)
+        return UA_STATUSCODE_BADOUTOFMEMORY;
+    dst->length = unescapedLength;
+    unescapeXmlString(data, length, dst->data);
+    return UA_STATUSCODE_GOOD;
 }
 
 DECODE_XML(DateTime) {
--- a/tests/check_types_builtin_xml.c
+++ b/tests/check_types_builtin_xml.c
@@ -1013,7 +1013,7 @@
     status s = UA_encodeXml(&src, type, &buf, NULL);
     ck_assert_int_eq(s, UA_STATUSCODE_GOOD);
 
-    char *result = "<String>he\\zsdl\alo€ \x26\x3A asdasd</String>";
+    char *result = "<String>he\\zsdl\alo€ &amp;\x3A asdasd</String>";
     buf.data[size] = 0; /* zero terminate */
     ck_assert_str_eq(result, (char*)buf.data);
     UA_ByteString_clear(&buf);
```

### Verification

Built out-of-tree at `CMAKE_BUILD_TYPE=Debug` with `UA_ENABLE_XML_ENCODING=ON`
(so `UA_DEBUG` is set and the harness asserts are live), then:

- **The reported abort is gone.** The real `fuzz_xml_decode_encode` harness now
  exits 0 on `<![CDATA[<]]>`, `<![CDATA[&]]>` and `<![CDATA[x<y&z]]>`, which
  previously aborted rc 134 at `:51`.
- **A 20-case round-trip battery passes** (decode → encode → decode → encode,
  asserting `UA_order` equality and bit-equal re-encoding), covering `<`, `&`,
  `>`, `]]>`, literal `&lt;` text (must survive as `&amp;lt;` and not be
  double-unescaped), numeric references, invalid references, bare `&`, quotes,
  UTF-8 including astral-plane codepoints, and the empty string. The same
  battery fails 12 of 20 before the patch.
- **`UA_calcSizeXml` equals the encoded length** in every case above.
- **No over-escaping regressions.** Values containing `>`, `"`, `'` and `;`
  still round-trip; `XmlElement` output stays raw markup; `Int32`, `Boolean`,
  `Double`, `Guid` and tag names are byte-identical to before.
- The one existing test the patch touches
  (`UA_String_escapeutf_xml_encode` in `tests/check_types_builtin_xml.c`) was
  checked directly: its updated expectation matches the new output exactly. I
  could not run the full `check`-based suite as libcheck was unavailable in my
  environment, so that suite is worth running before merge.

### Two silent-corruption cases the patch also fixes

Neither produces an error today; both simply return the wrong value.

| value | encoded today | decodes back as |
|---|---|---|
| `&lt;` (literal 4 chars) | `&lt;` | `;` |
| `<String>hi</String>` (as a *value*) | written raw | the **empty string** |

### Separate pre-existing bug found while testing (not addressed here)

`LocalizedText` does not round-trip when a sub-field is absent, independent of
escaping and reachable with no metacharacters at all:
`<Variant><Value><LocalizedText><Text>hello</Text></LocalizedText></Value></Variant>`
aborts at `:53` on unmodified master. The decoder leaves `locale.data == NULL`,
the encoder unconditionally emits `<Locale></Locale>`, and re-decoding yields a
non-NULL zero-length string, so `UA_order` differs. Happy to file separately.

## Checklist

 - [x] open62541 Version (release number or git tag): master, v1.5.6-514-g12b72510f (commit 12b72510); still present on current master
 - [x] Other OPC UA SDKs used (client or server): none
 - [x] Operating system: Linux
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached — N/A, API-level encode/decode defect, no server involved
 - [ ] Wireshark network dump attached — N/A, does not affect on-the-wire protocol traffic
 - [ ] Critical issue — not a crash in release builds and not a protocol issue, but note this is a silent round-trip / data-fidelity break in the XML encoding's documented use case (configuration files): the encoder produces output its own decoder rejects

</details>

swimmingbrain:
The XML encoder copies string content into the buffer as it is, so a string carrying `<` or `&` comes out as XML that no parser will read back:

```
<Variant><Value><String>a<b</String></Value></Variant>
```

String content now goes out with `&`, `<` and `>` replaced by their entities. `>` only has to be escaped where it would end a CDATA section, doing it everywhere is simpler and decodes back the same. `XmlElement` keeps writing raw, that type is markup by definition.

One thing had to move with it. An ExtensionObject with an XML encoded body keeps that body in a `UA_String` and wrote it through the string encoder, so escaping would have turned the body into text. It goes through `XmlElement` now, which is what it is. That is the first commit, and on its own it changes nothing, the two encoders were identical before.

While checking the report I could not reproduce the decoder half of it. On master here, entity references and CDATA all decode correctly:

```
<String>&lt;</String>              -> <
<String>x&lt;y</String>            -> x<y
<String><![CDATA[<]]></String>     -> <
```

So only the encoder needed changing, and with it the round trip from the issue works.

`UA_String_escapeutf_xml_encode` expected a raw `&` in its output, so that expectation is now `&amp;`. The new tests cover the three delimiters and a round trip back through the decoder. `ctest` fails the same seven tests before and after here (pubsub connection, custom enum, nodeset loader), all of them wanting things my container does not provide.

Closes #8356

### DIFF
--- a/src/ua_types_encoding_xml.c
+++ b/src/ua_types_encoding_xml.c
@@ -586,7 +586,8 @@ ENCODE_XML(ExtensionObject) {
            ret |= writeXmlElement(ctx, "ByteString", &src->content.encoded.body,
                                   &UA_TYPES[UA_TYPES_BYTESTRING]);
         else
-            ret |= ENCODE_DIRECT_XML(&src->content.encoded.body, String);
+            /* An XML encoded body is markup already, it goes out as it is */
+            ret |= ENCODE_DIRECT_XML(&src->content.encoded.body, XmlElement);
         ret |= writeXmlElemNameEnd(ctx, UA_XML_EXTENSIONOBJECT_BODY);
     } else {
         /* Write the decoded value */
--- a/src/ua_types_encoding_xml.c
+++ b/src/ua_types_encoding_xml.c
@@ -280,6 +280,31 @@ xmlEncodeWriteChars(CtxXml *ctx, const char *c, size_t len) {
     return UA_STATUSCODE_GOOD;
 }
 
+/* Element content cannot carry '&' and '<' as they are. '>' only has to be
+ * escaped where it ends a CDATA section, escaping it everywhere is simpler and
+ * decodes back the same. */
+static status UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT
+xmlEncodeWriteEscapedChars(CtxXml *ctx, const char *c, size_t len) {
+    if(len == 0)
+        return UA_STATUSCODE_GOOD;
+    status ret = UA_STATUSCODE_GOOD;
+    size_t start = 0;
+    for(size_t i = 0; i < len; i++) {
+        const char *entity;
+        size_t entityLen;
+        switch(c[i]) {
+        case '&': entity = "&amp;"; entityLen = 5; break;
+        case '<': entity = "&lt;";  entityLen = 4; break;
+        case '>': entity = "&gt;";  entityLen = 4; break;
+        default: continue;
+        }
+        ret |= xmlEncodeWriteChars(ctx, &c[start], i - start);
+        ret |= xmlEncodeWriteChars(ctx, entity, entityLen);
+        start = i + 1;
+    }
+    return ret | xmlEncodeWriteChars(ctx, &c[start], len - start);
+}
+
 static status UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT
 writeXmlElemNameBegin(CtxXml *ctx, const char* name) {
     if(ctx->depth >= UA_XML_ENCODING_MAX_RECURSION - 1)
@@ -441,7 +466,7 @@ ENCODE_XML(Double) {
 /* String */
 ENCODE_XML(String) {
     const UA_String *src = (const UA_String*)src_;
-    return xmlEncodeWriteChars(ctx, (const char*)src->data, src->length);
+    return xmlEncodeWriteEscapedChars(ctx, (const char*)src->data, src->length);
 }
 
 /* XmlElement */
--- a/src/ua_types_encoding_xml.c
+++ b/src/ua_types_encoding_xml.c
@@ -280,9 +280,10 @@ xmlEncodeWriteChars(CtxXml *ctx, const char *c, size_t len) {
     return UA_STATUSCODE_GOOD;
 }
 
-/* Element content cannot carry '&' and '<' as they are. '>' only has to be
- * escaped where it ends a CDATA section, escaping it everywhere is simpler and
- * decodes back the same. */
+/* Element content cannot carry '&' and '<' as they are. The other three
+ * predefined entities are only needed in attributes or where a CDATA section
+ * would end, escaping all five everywhere is simpler and decodes back the
+ * same. */
 static status UA_INTERNAL_FUNC_ATTR_WARN_UNUSED_RESULT
 xmlEncodeWriteEscapedChars(CtxXml *ctx, const char *c, size_t len) {
     if(len == 0)
@@ -293,9 +294,11 @@ xmlEncodeWriteEscapedChars(CtxXml *ctx, const char *c, size_t len) {
         const char *entity;
         size_t entityLen;
         switch(c[i]) {
-        case '&': entity = "&amp;"; entityLen = 5; break;
-        case '<': entity = "&lt;";  entityLen = 4; break;
-        case '>': entity = "&gt;";  entityLen = 4; break;
+        case '&':  entity = "&amp;";  entityLen = 5; break;
+        case '<':  entity = "&lt;";   entityLen = 4; break;
+        case '>':  entity = "&gt;";   entityLen = 4; break;
+        case '"':  entity = "&quot;"; entityLen = 6; break;
+        case '\'': entity = "&apos;"; entityLen = 6; break;
         default: continue;
         }
         ret |= xmlEncodeWriteChars(ctx, &c[start], i - start);
--- a/tests/check_types_builtin_xml.c
+++ b/tests/check_types_builtin_xml.c
@@ -996,7 +996,7 @@ START_TEST(UA_String_escapesimple_xml_encode) {
     status s = UA_encodeXml(&src, type, &buf, NULL);
     ck_assert_int_eq(s, UA_STATUSCODE_GOOD);
 
-    char *result = "<String>\b\th\"e\fl\nl\\o\r</String>";
+    char *result = "<String>\b\th&quot;e\fl\nl\\o\r</String>";
     buf.data[size] = 0; /* zero terminate */
     ck_assert_str_eq(result, (char*)buf.data);
     UA_ByteString_clear(&buf);
@@ -1022,7 +1022,7 @@ START_TEST(UA_String_escapeutf_xml_encode) {
 END_TEST
 
 START_TEST(UA_String_markup_xml_encode) {
-    UA_String src = UA_STRING("a<b & c>d");
+    UA_String src = UA_STRING("a<b & c>d \"e\" 'f'");
     const UA_DataType *type = &UA_TYPES[UA_TYPES_STRING];
     size_t size = UA_calcSizeXml((void*)&src, type, NULL);
 
@@ -1032,7 +1032,7 @@ START_TEST(UA_String_markup_xml_encode) {
     status s = UA_encodeXml(&src, type, &buf, NULL);
     ck_assert_int_eq(s, UA_STATUSCODE_GOOD);
 
-    char *result = "<String>a&lt;b &amp; c&gt;d</String>";
+    char *result = "<String>a&lt;b &amp; c&gt;d &quot;e&quot; &apos;f&apos;</String>";
     buf.data[size] = 0; /* zero terminate */
     ck_assert_str_eq(result, (char*)buf.data);
     UA_ByteString_clear(&buf);
@@ -1040,7 +1040,7 @@ START_TEST(UA_String_markup_xml_encode) {
 END_TEST
 
 START_TEST(UA_String_markup_xml_roundtrip) {
-    UA_String src = UA_STRING("a<b & c>d");
+    UA_String src = UA_STRING("a<b & c>d \"e\" 'f'");
     const UA_DataType *type = &UA_TYPES[UA_TYPES_STRING];
     size_t size = UA_calcSizeXml((void*)&src, type, NULL);
 
--- a/tests/check_types_builtin_xml.c
+++ b/tests/check_types_builtin_xml.c
@@ -1021,6 +1021,45 @@ START_TEST(UA_String_escapeutf_xml_encode) {
 }
 END_TEST
 
+START_TEST(UA_String_markup_xml_encode) {
+    UA_String src = UA_STRING("a<b & c>d");
+    const UA_DataType *type = &UA_TYPES[UA_TYPES_STRING];
+    size_t size = UA_calcSizeXml((void*)&src, type, NULL);
+
+    UA_ByteString buf;
+    UA_ByteString_allocBuffer(&buf, size + 1);
+
+    status s = UA_encodeXml(&src, type, &buf, NULL);
+    ck_assert_int_eq(s, UA_STATUSCODE_GOOD);
+
+    char *result = "<String>a&lt;b &amp; c&gt;d</String>";
+    buf.data[size] = 0; /* zero terminate */
+    ck_assert_str_eq(result, (char*)buf.data);
+    UA_ByteString_clear(&buf);
+}
+END_TEST
+
+START_TEST(UA_String_markup_xml_roundtrip) {
+    UA_String src = UA_STRING("a<b & c>d");
+    const UA_DataType *type = &UA_TYPES[UA_TYPES_STRING];
+    size_t size = UA_calcSizeXml((void*)&src, type, NULL);
+
+    UA_ByteString buf;
+    UA_ByteString_allocBuffer(&buf, size);
+    status s = UA_encodeXml(&src, type, &buf, NULL);
+    ck_assert_int_eq(s, UA_STATUSCODE_GOOD);
+
+    UA_String out;
+    UA_String_init(&out);
+    s = UA_decodeXml(&buf, &out, type, NULL);
+    ck_assert_int_eq(s, UA_STATUSCODE_GOOD);
+    ck_assert(UA_String_equal(&src, &out));
+
+    UA_String_clear(&out);
+    UA_ByteString_clear(&buf);
+}
+END_TEST
+
 START_TEST(UA_String_special_xml_encode) {
     UA_String src = UA_STRING("𝄞𠂊𝕥🔍");
     const UA_DataType *type = &UA_TYPES[UA_TYPES_STRING];
@@ -4534,6 +4573,8 @@ static Suite *testSuite_builtin_xml(void) {
     tcase_add_test(tc_xml_encode, UA_String_Null_xml_encode);
     tcase_add_test(tc_xml_encode, UA_String_escapesimple_xml_encode);
     tcase_add_test(tc_xml_encode, UA_String_escapeutf_xml_encode);
+    tcase_add_test(tc_xml_encode, UA_String_markup_xml_encode);
+    tcase_add_test(tc_xml_encode, UA_String_markup_xml_roundtrip);
     tcase_add_test(tc_xml_encode, UA_String_special_xml_encode);
 
     tcase_add_test(tc_xml_encode, UA_DateTime_xml_encode);
--- a/tests/check_types_builtin_xml.c
+++ b/tests/check_types_builtin_xml.c
@@ -1014,7 +1014,7 @@ START_TEST(UA_String_escapeutf_xml_encode) {
     status s = UA_encodeXml(&src, type, &buf, NULL);
     ck_assert_int_eq(s, UA_STATUSCODE_GOOD);
 
-    char *result = "<String>he\\zsdl\alo€ \x26\x3A asdasd</String>";
+    char *result = "<String>he\\zsdl\alo€ &amp;\x3A asdasd</String>";
     buf.data[size] = 0; /* zero terminate */
     ck_assert_str_eq(result, (char*)buf.data);
     UA_ByteString_clear(&buf);


jpfr:
Looks really good.
Make sure to escape all five possibilities enumerated here:
https://www.liquid-technologies.com/Reference/Glossary/XML_EscapingData.html

swimmingbrain:
Done, all five now. `"` and `'` go out as `&quot;` and `&apos;` too. The one existing test that expected a raw quote in its output is updated, and the two new ones carry both quote kinds and round trip through the decoder. `check_types_builtin_xml` passes here, 181 checks.

jpfr:
Thanks, merged.

PR #8398, opened 2026-09-10
fix(plugins): announce the UserName policy when only a login callback is set

StalderT:
### Problem

`UA_AccessControl_defaultWithLoginCallback()` delegates to
`UA_AccessControl_default()` and attaches the callback **afterwards**. The
UserTokenPolicies are built inside that first call, so the callback does not exist
yet when they are assembled — and the UserName policy is only added when
`usernamePasswordLoginSize > 0`.

The consequence is that a caller who verifies credentials in its own callback — the
whole point of this variant — still has to hand over a login list. Without one, no
client is ever told that username/password login is available, so the callback is
never reached.

### Why this is worth fixing rather than documenting

Passing a dummy list looks like a harmless workaround, but the obvious filler — an
empty password — **fails open**:

```c
if(userToken->password.length != upl->password.length) continue;  /* 0 == 0 */
if(!UA_constantTimeEqual(..., ..., 0))                 continue;  /* returns true */
match = true;
```

`UA_constantTimeEqual()` over zero bytes returns true, and the guard above it only
rejects the case where the user name **and** the password are empty. A valid user
name with an empty password would open a session.

So the API, as it stands, pushes integrators toward a credential bypass. That is
what makes it a defect rather than an inconvenience.

### Change

Move the body into a shared static function that receives the callback, so both
entry points build the policies with the same knowledge, and announce the UserName
policy when a login list is configured **or** a callback is set. The post-hoc
attachment is gone; no public signature changes and no behaviour change for
existing callers.

### Test

`Server_usernamePolicyWithLoginCallbackOnly` in `tests/server/check_accesscontrol.c`
asserts the UserName policy is present with an empty list and a callback. Verified
in both directions: it fails on `1.5` without the change
(`Assertion 'foundUserName' failed`) and passes with it. It also asserts the
Anonymous policy, so a configuration that announced nothing at all could not pass
it. Added to the existing `tc_server` case — no server, no fixture, no CMake change.

### DIFF
--- a/plugins/ua_accesscontrol_default.c
+++ b/plugins/ua_accesscontrol_default.c
@@ -364,12 +364,18 @@ static void clear_default(UA_AccessControl *ac) {
     }
 }
 
-UA_StatusCode
-UA_AccessControl_default(UA_ServerConfig *config,
-                         UA_Boolean allowAnonymous,
-                         const UA_String *userTokenPolicyUri,
-                         size_t usernamePasswordLoginSize,
-                         const UA_UsernamePasswordLogin *usernamePasswordLogin) {
+/* Shared implementation. The login callback has to be known *here*, before the
+ * UserTokenPolicies are built: the UserName policy is only announced when
+ * username/password login is actually possible, which is the case either when a
+ * login list is configured or when a callback verifies the credentials itself. */
+static UA_StatusCode
+setDefaultAccessControl(UA_ServerConfig *config,
+                        UA_Boolean allowAnonymous,
+                        const UA_String *userTokenPolicyUri,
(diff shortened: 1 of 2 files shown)


jpfr:
This is a change for a stable release family.
It looks as if this might break existing applications.

Okay to make this for master?

StalderT:
On 1.5 the callback variant cannot be used on its own: without a login list the UserName policy is never announced, so no client can even present credentials to the callback. The only workaround is a dummy list, and the obvious filler — an empty password — fails open (constant-time compare over zero bytes returns true), which is why I see this as a fix rather than a feature. If you still prefer master, I'll rebase it there.

PR #8399, opened 2026-09-12
[REVIEW] fix(arch,plugins): remove the non-threadsafe gmtime and inet_ntoa calls

### LINKED ISSUE #7997: use of non threadsafe function like gmtime
in current version of open62541 (1.5.4) there are some functions used which are not threadsafe. 

e.g. in [plugins](https://github.com/open62541/open62541/tree/master/plugins)/[crypto](https://github.com/open62541/open62541/tree/master/plugins/crypto)/[mbedtls](https://github.com/open62541/open62541/tree/master/plugins/crypto/mbedtls)
/create_certificate.c the method "gmtime" is used twice. "strtok" was also used in the past (v.1.4.2).

is it possible to use only threadsafe functions?

StalderT:
gmtime and lwIP's inet_ntoa return a pointer to a static buffer shared by all threads. The mbedTLS certificate creation now derives its validity dates from UA_DateTime_now/UA_DateTime_toStruct, which also keeps the expiry date in UTC (mktime interpreted the gmtime result as local time). Zephyr uses gmtime_r and lwIP inet_ntoa_r.

Fixes #7997

### DIFF
--- a/arch/lwip/eventloop_lwip.h
+++ b/arch/lwip/eventloop_lwip.h
@@ -83,10 +83,10 @@ lwip_getnameinfo(const struct sockaddr *sa, socklen_t salen,
         if(host) {
             if(flags & NI_NUMERICHOST) {
                 /* Return numeric IP address */
-                strncpy(host, inet_ntoa(addr_in->sin_addr), hostlen);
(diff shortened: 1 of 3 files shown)


jpfr:
Thanks, merged.

PR #8407, opened 2026-09-13
fix(client): bound monitored item delete results

jpfr:
A server can return more DeleteMonitoredItems results than the client requested. The client previously used the response length to index the request array, allowing a malformed response to trigger an out-of-bounds read while updating the local monitored-item state.

Process only request/result pairs and add a regression for an oversized response.


jpfr:
Duplicate from a separate security issue.

Issue #8404, opened 2026-09-13
validateCertificate() computes BadCertificateUriInvalid and then returns BadCertificateInvalid, discarding it


ronaldmcvety:
## Description

When a client presents a certificate whose `uniformResourceIdentifier` SAN does not match the `applicationUri` in its ApplicationDescription, the server **correctly rejects it** — the security control works and nothing here is exploitable. But it answers with the generic `BadCertificateInvalid` where Part 4 specifies `BadCertificateUriInvalid`.

The interesting part is that the stack already *has* the right answer and throws it away.

`UA_CertificateUtils_verifyApplicationUri()` (`plugins/crypto/mbedtls/certificategroup.c`) correctly returns `UA_STATUSCODE_BADCERTIFICATEURIINVALID` on a SAN/applicationUri mismatch. `validateCertificate()` (`src/server/ua_server_utils.c`) captures it in `res`, logs the URI-specific message — *"The client's ApplicationUri could not be verified against the ApplicationUri … from the client's ApplicationDescription"* — and then, with `res` still holding the specific code and never reassigned, returns a hardcoded constant instead.

**v1.5.5:**

```c
if(server->config.allowAllCertificateUris <= UA_RULEHANDLING_ABORT)
    return UA_STATUSCODE_BADCERTIFICATEINVALID;
```

**Current master** — the function has been rewritten since (auditing added, the rule-handling threshold changed), and the same substitution survives the rewrite:

```c
/* Throw an audit event and abort */
if(server->config.allowAllCertificateUris < UA_RULEHANDLING_WARN) {
#ifdef UA_ENABLE_AUDITING
    auditCertificateDataMismatchEvent(server, channel, session, logPrefix,
                                      res, certificate, ad->applicationUri);
#endif
    return UA_STATUSCODE_BADCERTIFICATEINVALID;
}
```

Note that the audit call on master is handed `res` — so the specific status is passed to auditing and discarded from the return, in the same statement block.

## Suggested fix

```c
return res;
```

`res` already holds `BADCERTIFICATEURIINVALID` at that point, so this is a one-token change with no new state and no behaviour change other than the code the client sees. The certificate is still rejected; only the reason it is given changes.

## Why it is worth fixing

The specific status is what tells a client *which* thing was wrong with its certificate — a mismatched URI is an operator configuration error with an obvious fix, and `BadCertificateInvalid` sends someone looking at expiry, chain and trust lists instead. A CTT run is the likely place this surfaces.

[...]

PR #8438, opened 2026-09-15
fix(server): separate return code for invalid certificate uri

keba-estr:
If an invalid certificate uri is found, the separate return code UA_STATUSCODE_BADCERTIFICATEURIINVALID is used instead of the generic UA_STATUSCODE_BADCERTIFICATEINVALID.


keba-estr:
Fixes issue [#8404](https://github.com/open62541/open62541/issues/8404)

PR #8443, opened 2026-09-15
1.4+: fix(client): initialize async response before decode

o6-nbeier:
UA_Response asyncResponse is declared without an initializer in processMSGResponse. When UA_NodeId_decodeBinary fails, control jumps directly to the process label via goto, bypassing the UA_init call that is only reached through the type-verification branch. At the process label, a single field (serviceResult) is written into an otherwise uninitialized object. The user callback is then invoked with this garbage-filled struct, and UA_clear subsequently walks every field per the type descriptor, treating uninitialized pointer-shaped fields as heap allocations, which may cause invalid-free or memory corruption.

Initialize asyncResponse immediately after declaration so that every code path through the function operates on a well-defined object.

Internal Vulnerability Advisory: open62541-SA-2026-0037


jpfr:
Duplicate from #8442

PR #8446, opened 2026-09-15
1.5+: fix(crypto): use constant-time, length-checked comparison for HMAC/MA…

o6-nbeier:
…C verification

The OpenSSL HMAC-SHA1/SHA256/SHA384 verify helpers and the mbedTLS
ECC policies' shared HMAC-SHA256/SHA384 verify helpers compared the
recomputed MAC against the peer-supplied signature with
UA_ByteString_equal(), which reduces to a plain memcmp(). memcmp()
carries no constant-time guarantee and, on a byte-at-a-time
implementation (the norm on the small-footprint libcs open62541 is
routinely built against), returns as soon as it finds a differing
byte -- a timing side channel a remote peer can use to recover a
valid MAC byte-by-byte.

Switch to UA_constantTimeEqual(), already used correctly by every
other symmetric SecurityPolicy (mbedTLS Basic128Rsa15, Basic256,
Basic256Sha256, Aes128Sha256RsaOaep, Aes256Sha256RsaPss, and the
PubSub Aes128/256Ctr policies) and by the default access control
plugin's password comparison. Since it takes a single length and
reads that many bytes from both buffers, first validate
signature->length against the expected MAC length, mirroring the
pattern used by the project's mbedTLS RSA/AES-CTR Verify functions.
The digest length is fixed by the algorithm and policy, not attacker
data, so this early-return introduces no new timing side channel.

Internal Vulnerability Advisory: open62541-SA-2026-0038


jpfr:
Duplicate from #8444

PR #8448, opened 2026-09-15
1.4: fix(client): free the OpenSecureChannelResponse and type NodeId on ev…

o6-nbeier:
…ery processOPNResponse exit

processOPNResponse() decodes the leading type NodeId and, once it matches, the OpenSecureChannelResponse body before validating the response (nonce reuse). Two of its return paths -- the type-NodeId branch when the response type does not match, and the reused-nonce rejection after the response has been fully decoded -- returned without freeing the decoded value. A server that repeatedly rejects a connecting client with a malformed or intentionally-invalid OpenSecureChannelResponse leaks the ResponseHeader, SecurityToken and ServerNonce fields of the response (and, when the type NodeId itself is heap-backed, the NodeId) on every attempt, letting a hostile or misbehaving endpoint grow a reconnecting client's heap usage without bound.

Clear the type NodeId immediately after the type check, since its content is only needed for that one comparison. Clear the decoded OpenSecureChannelResponse at every return point after it has been decoded -- on rejection as well as on success.

Internal Vulnerability Advisory: open62541-SA-2026-0039


jpfr:
Duplicate from #8447

PR #8455, opened 2026-09-17
fix(xml): omit absent LocalizedText and QualifiedName sub-fields

### LINKED ISSUE #8357: LocalizedText without a locale drops the text in round-trip
When round-tripping `"<Variant><Value><LocalizedText><Text>hello</Text></LocalizedText></Value></Variant>"`, the text gets dropped.

<details><summary>Claude's report</summary>

## Description

XML encoding of `LocalizedText` and `QualifiedName` does not preserve the
distinction between an **absent** string sub-field and an **empty** one, so a
`decode → encode → decode` round trip silently changes the value. No XML
metacharacters are involved; this is unrelated to escaping.

Minimal case — a `LocalizedText` with no `Locale` element:

```
<Variant><Value><LocalizedText><Text>hello</Text></LocalizedText></Value></Variant>
```

Mechanism, all in `src/ua_types_encoding_xml.c`:

1. `DECODE_XML(LocalizedText)` decodes via `decodeXmlFields`, which leaves a
   field that was not present in the input at its initialised value — so
   `locale.data == NULL`, `locale.length == 0` (absent).
2. `ENCODE_XML(LocalizedText)` writes **both** sub-elements unconditionally:

   ```c
   ENCODE_XML(LocalizedText) {
       const UA_LocalizedText *src = (const UA_LocalizedText*)src_;
       return writeXmlElement(ctx, UA_XML_LOCALIZEDTEXT_LOCALE,
                              &src->locale, &UA_TYPES[UA_TYPES_STRING]) |
           writeXmlElement(ctx, UA_XML_LOCALIZEDTEXT_TEXT,
                           &src->text, &UA_TYPES[UA_TYPES_STRING]);
   }
   ```

   `writeXmlElement` has no emptiness or NULL guard, so a NULL `locale` is
   emitted as `<Locale></Locale>`.
3. `DECODE_XML(String)` maps an empty element to the **non-NULL** empty
   sentinel:

   ```c
   /* Empty string? */
   if(length == 0) {
       dst->data = (UA_Byte*)UA_EMPTY_ARRAY_SENTINEL;
       dst->length = 0;
       return UA_STATUSCODE_GOOD;
   }
   ```

4. `UA_order` distinguishes a NULL `UA_String` from a non-NULL empty one, so
   the re-decoded value compares unequal to the original.

Note the serialisation itself is *stable* — encoding the re-decoded value gives
a byte-identical buffer. What is lost is the value: absent becomes empty, once,
on the first round trip.

### This contradicts the library's own documented model

`types.h:1436` states the three-valued convention plainly:

> An undefined array has length 0 and the data pointer is `NULL`. An array of
> length 0 also has length 0 but a data pointer `UA_EMPTY_ARRAY_SENTINEL`.

`stringOrder` (`src/ua_types.c:2275`) implements that distinction deliberately,
and says so:

```c
/* For zero-length arrays, every pointer not NULL is considered a
 * UA_EMPTY_ARRAY_SENTINEL. */
if(p1->data == p2->data) return UA_ORDER_EQ;
if(p1->data == NULL) return UA_ORDER_LESS;
```

So `UA_order` is behaving as designed; it is the encoder that collapses the two.

This is also a class of bug you have fixed before, on the binary side:
[#6802](https://github.com/open62541/open62541/issues/6802) "UA_encodeBinary -
Empty Strings encoded as NULL Strings" (fixed by
[#6827](https://github.com/open62541/open62541/pull/6827), Nov 2024). The XML
encoder has the same problem in the opposite direction — NULL encoded as empty.

And the XML encoder *already implements the right rule elsewhere*.
`encodeXmlStructure` skips a member that is NULL rather than empty:

```c
if(!m->isOptional && length == 0 && !data)
    continue;
```

Note that guard tests the **data pointer**, not the length — so an
empty-but-present string still gets its element. `LocalizedText` and
`QualifiedName` are builtins with hand-written `ENCODE_XML` functions, so they
bypass `encodeXmlStructure` and never picked up this rule.

### A caution on copying the JSON encoder

`ENCODE_JSON(LocalizedText)` guards on `src->locale.length > 0`. That fixes the
NULL case but introduces the mirror problem — an empty-but-present string is
also omitted and comes back NULL. Measured, same build:

| source `locale` | JSON emitted | decodes back as | `UA_order` |
|---|---|---|---|
| NULL | `{"UaType":21,"Value":{"Text":"hello"}}` | NULL | EQ |
| empty (sentinel) | `{"UaType":21,"Value":{"Text":"hello"}}` | NULL | **NEQ** |

So the two codecs lose the distinction in opposite directions, and a
length-based guard in XML would trade one round-trip failure for another. The
pointer-based test already used by `encodeXmlStructure` preserves all three
states.

### Observable

The existing fuzz target `tests/fuzz/fuzz_xml_decode_encode.cc` aborts on this
input in a `UA_DEBUG` build (which `CMAKE_BUILD_TYPE=Debug` enables, and which
is the OSS-Fuzz configuration), at line 53:

```
Assertion `UA_order(&value, &value2, &UA_TYPES[23]) == UA_ORDER_EQ' failed.
fuzz_xml_decode_encode.cc:53
Aborted (rc 134)
```

In a release build the assert is compiled out and the value change is silent.

Self-contained example:

```c
#include <open62541/types.h>
#include <stdio.h>
#include <string.h>

int main(void) {
    /* A LocalizedText with no Locale element. */
    const char *xml =
        "<Variant><Value><LocalizedText><Text>hello</Text>"
        "</LocalizedText></Value></Variant>";

    UA_ByteString in;
    in.data = (UA_Byte*)(uintptr_t)xml;
    in.length = strlen(xml);

    UA_Variant v;
    UA_Variant_init(&v);
    UA_StatusCode rc = UA_decodeXml(&in, &v, &UA_TYPES[UA_TYPES_VARIANT], NULL);
    printf("decode #1: %s\n", UA_StatusCode_name(rc));

    UA_LocalizedText *lt = (UA_LocalizedText*)v.data;
    printf("  locale after decode #1: %s (length %zu)\n",
           lt->locale.data ? "non-NULL" : "NULL", (size_t)lt->locale.length);

    size_t n = UA_calcSizeXml(&v, &UA_TYPES[UA_TYPES_VARIANT], NULL);
    UA_ByteString out = UA_BYTESTRING_NULL;
    UA_ByteString_allocBuffer(&out, n);
    rc = UA_encodeXml(&v, &UA_TYPES[UA_TYPES_VARIANT], &out, NULL);
    printf("encode:    %s\n", UA_StatusCode_name(rc));
    printf("output:    %.*s\n", (int)out.length, out.data);

    UA_Variant v2;
    UA_Variant_init(&v2);
    rc = UA_decodeXml(&out, &v2, &UA_TYPES[UA_TYPES_VARIANT], NULL);
    printf("decode #2: %s\n", UA_StatusCode_name(rc));

    UA_LocalizedText *lt2 = (UA_LocalizedText*)v2.data;
    printf("  locale after decode #2: %s (length %zu)\n",
           lt2->locale.data ? "non-NULL" : "NULL", (size_t)lt2->locale.length);

    printf("UA_order(v, v2): %s\n",
           UA_order(&v, &v2, &UA_TYPES[UA_TYPES_VARIANT]) == UA_ORDER_EQ
           ? "UA_ORDER_EQ" : "NOT EQUAL  <-- the round trip changed the value");

    UA_Variant_clear(&v);
    UA_Variant_clear(&v2);
    UA_ByteString_clear(&out);
    return 0;
}
```

Actual output:

```
decode #1: Good
  locale after decode #1: NULL (length 0)
encode:    Good
output:    <Variant><Value><LocalizedText><Locale></Locale><Text>hello</Text></LocalizedText></Value></Variant>
decode #2: Good
  locale after decode #2: non-NULL (length 0)
UA_order(v, v2): NOT EQUAL  <-- the round trip changed the value
```

### Which inputs are affected

Checked against the fuzz harness on an unmodified Debug build (rc 134 = the
`UA_order` assert at `:53` fires):

| input | result |
|---|---|
| `<LocalizedText><Text>hi</Text></LocalizedText>` | **fails** |
| `<LocalizedText><Locale>en</Locale></LocalizedText>` | **fails** |
| `<LocalizedText></LocalizedText>` | **fails** |
| `<LocalizedText><Locale>en</Locale><Text>hi</Text></LocalizedText>` | passes |
| `<LocalizedText><Locale></Locale><Text></Text></LocalizedText>` | passes |
| `<QualifiedName><NamespaceIndex>0</NamespaceIndex></QualifiedName>` | **fails** |
| `<QualifiedName></QualifiedName>` | **fails** |
| `<QualifiedName><Name>hi</Name></QualifiedName>` | passes |
| `<String></String>` | passes |
| `<ByteString></ByteString>` | passes |

So it is any structured type with a `UA_String` sub-field that the input omits.
A bare `String`/`ByteString` Variant is unaffected, because there the value is
the string itself rather than an optional field.

### How this arose

Not a long-standing oversight — it was introduced by a fix for a different bug.
5b8d4593d802 (2025-03-03, "fix(core): Don't XML-encode empty Strings as
null") removed the `if(!src->data) return xmlEncodeWriteChars(ctx, "null", 4);`
branch from `ENCODE_XML(String)`, which correctly stopped the literal text
`null` appearing on the wire, but left NULL and empty encoding identically.

Then 7a07028859d6 (2026-07-23, "feat(core): Don't print 'default' values in
XML structures") added the `!data` skip quoted above — the right rule, but only
on the `encodeXmlStructure` path, which builtins do not take.

### Suggested fix

Guard each `writeXmlElement` in `ENCODE_XML(LocalizedText)` on the data
pointer, matching `encodeXmlStructure`:

```c
ENCODE_XML(LocalizedText) {
    const UA_LocalizedText *src = (const UA_LocalizedText*)src_;
    status ret = UA_STATUSCODE_GOOD;
    if(src->locale.data)
        ret |= writeXmlElement(ctx, UA_XML_LOCALIZEDTEXT_LOCALE,
                               &src->locale, &UA_TYPES[UA_TYPES_STRING]);
    if(src->text.data)
        ret |= writeXmlElement(ctx, UA_XML_LOCALIZEDTEXT_TEXT,
                               &src->text, &UA_TYPES[UA_TYPES_STRING]);
    return ret;
}
```

`ENCODE_XML(QualifiedName)` writes `&src->name` unconditionally in the same way
and needs the same treatment — I confirmed it fails on
`<QualifiedName></QualifiedName>` and on `<QualifiedName><NamespaceIndex>0</NamespaceIndex></QualifiedName>`.

**Two existing tests currently certify the present behaviour and would need
updating**, which is worth knowing before anyone starts: in
`tests/check_types_builtin_xml.c`, `UA_LocalizedText_null_xml_encode` (`:1618`)
asserts that a fully-`init`ed `LocalizedText` encodes to
`<LocalizedText><Locale></Locale><Text></Text></LocalizedText>`, and
`UA_LocalizedText_null_locale_xml_encode` (`:1596`) makes the corresponding
assertion for a NULL `Text`. Under the fix both should expect the elements to
be omitted.

I am happy to prepare the patch if you agree with this direction; I did not
want to send one that changes two of your existing test expectations without
checking first.

Used CMake options:

```bash
cmake -DUA_NAMESPACE_ZERO=FULL -DUA_ENABLE_XML_ENCODING=ON -DCMAKE_BUILD_TYPE=Debug ..
```

`CMAKE_BUILD_TYPE=Debug` enables `UA_DEBUG`, which is what makes the fuzz
harness's `UA_assert`s live. With a release build type the assert is compiled
out and the value change happens silently.

## Checklist

 - [x] open62541 Version (release number or git tag): master, v1.5.6-514-g12b72510f (commit 12b72510)
 - [x] Other OPC UA SDKs used (client or server): none
 - [x] Operating system: Linux
 - [ ] Logs (with `UA_LOGLEVEL` set as low as necessary) attached — N/A, API-level encode/decode defect, no server involved
 - [ ] Wireshark network dump attached — N/A, does not affect on-the-wire protocol traffic
 - [ ] Critical issue — not a crash in release builds, but a silent value change on round trip

</details>

YuEfSaEDU:
The XML encoders of LocalizedText and QualifiedName serialize their string sub-fields unconditionally, so an absent (NULL) locale or name is written as an empty element and comes back as an empty-but-present string on decoding. Because UA_order deliberately distinguishes a NULL string from the empty-array sentinel, a decode → encode → decode round trip silently changed the value and dropped locale-less text comparisons.

This makes the hand-written encoders skip sub-elements whose data pointer is NULL, matching the rule encodeXmlStructure already applies, while empty-but-present strings keep their element; DECODE_XML(String) already maps empty elements to the sentinel, so absent, empty, and present states are all stable across round trips. Two encode tests that certified the old behavior are updated, and round-trip tests for a locale-less LocalizedText and a QualifiedName without a name assert the values survive.

Fixes #8357


jpfr:
Thanks!

Issue #6943, opened 2024-12-05
UA_Client_getEndpoints crashing


electro-logic:
Hi,
The following code is crashing (branch 1.4, latest)

```
#define _CRT_SECURE_NO_DEPRECATE
#include <open62541/client.h>
#include <open62541/client_highlevel.h>
#include <open62541/client_config_default.h>
#include <cstdio>
#include <open62541/plugin/pki_default.h>
int main(int argc, char* argv[])
{
    while (true) {
        UA_EndpointDescription* endpointArray = NULL;
        size_t endpointArraySize = 0;
        UA_Client* client = UA_Client_new();
        UA_Client_getEndpoints(client, "opc.tcp://127.0.0.1", &endpointArraySize, &endpointArray);
        UA_Client_connect(client, "opc.tcp://127.0.0.1");
        UA_ClientConfig* cc = UA_Client_getConfig(client);
        UA_ByteString certificate = loadFile("public.der");
        UA_ByteString privateKey = loadFile("private.pem");
        UA_ClientConfig_setDefaultEncryption(cc, certificate, privateKey, NULL, 0, NULL, 0);
        UA_CertificateVerification_AcceptAll(&cc->certificateVerification);
        cc->securityMode = UA_MESSAGESECURITYMODE_SIGNANDENCRYPT;
        cc->securityPolicyUri = UA_STRING("http://opcfoundation.org/UA/SecurityPolicy#Basic256");
        UA_Client_getEndpoints(client, "opc.tcp://127.0.0.1", &endpointArraySize, &endpointArray);   // crash!
        UA_Client_disconnect(client);
    }
}
```
where loadFile() is the function from examples/common.h.

The last call to UA_Client_getEndpoints() is crashing at UA_SecureChannel_clear(), with an Access read violation.

Can anyone reproduce the issue? I used the Quickstart Reference Server for the test. Any der/pem files can be used to reproduce the issue.

PR #8465, opened 2026-09-18
fix(client): reject SecurityPolicy reconfiguration on a connected client

o6-nbeier:
UA_ClientConfig_setDefaultEncryption reallocated config->securityPolicies while an already-open SecureChannel still held a pointer into it, causing a heap-use-after-free on the next request. Track whether a SecureChannel is bound to the array and refuse to reconfigure while that holds instead of corrupting memory.

Also adds a regression test.

Fixes #6943


include/open62541/client.h
```diff
@@ -160,6 +160,13 @@ struct UA_ClientConfig {
     size_t securityPoliciesSize;
     UA_SecurityPolicy *securityPolicies;
 
+    /* Set by the client while a SecureChannel holds a pointer into
+     * securityPolicies. While true, UA_ClientConfig_setDefaultEncryption
+     * refuses to grow/move that array, since doing so could invalidate the
+     * pointer the SecureChannel already uses. Internal state -- do not set
+     * from application code. */
+    UA_Boolean securityPoliciesInUse;
```
jpfr: I don't see why we would expose this internal state in the public API.

Especially since this is explicitly mentioned to be internal!

PR #8473, opened 2026-09-21
fix(types): use adjusted range for array writes (backport of #8408)

### LINKED ISSUE #8471: Security fix #8408 (Variant_setRange heap OOB write) merged into 1.4 but not backported to master
PR #8408 ("fix(types): use adjusted range for array writes") was merged into the `1.4` branch on 2026-09-14, fixing a heap out-of-bounds write in `Variant_setRange` (`src/ua_types.c`): `checkAdjustRange()` produces a bounds-clamped copy (`thisrange`), but `computeStrides()` was called with the original, unclamped `range` instead -- letting an attacker-controlled `IndexRange` in a `Write` service request (`WriteValue.IndexRange`) drive `count`/`block` past the real array bounds and `memcpy` past the allocated buffer.

Current `master` (HEAD `f223658`) still has the unfixed code:

```c
// src/ua_types.c, Variant_setRange, current master
UA_StatusCode retval = checkAdjustRange(v, &thisrange);
    return retval;

/* Compute the strides */
size_t count, block, stride, first;
computeStrides(v, range, &count, &block, &stride, &first);   // <-- should be &thisrange, matches 1.4's fix
```

Anyone building from `master` (or a submodule/CI pipeline pinned to `master`/"latest" rather than a tagged `1.4.x` release) is still exposed to this heap OOB write via a standard OPC UA `Write` request. Could this fix be cherry-picked/backported to `master` as well? Happy to open a PR with the one-line change if useful.

iliasabk:
Fixes #8471 — backport of the `Variant_setRange` out-of-bounds write fix from `1.4` (#8408, commit `edfe421`) to `master`.

## Root cause

`Variant_setRange()` copies the caller's `UA_NumericRange` into the stack-local `thisrange` and lets `checkAdjustRange()` clamp it to the destination array's real bounds — but then computes `count`/`block`/`stride`/`first` from the *original* `range`. An `IndexRange` whose `max` exceeds the actual dimension (e.g. received through the Write service) keeps its oversized extent: a matching source array passes `count != arraySize`, and the subsequent `memcpy` loop writes `block` elements per `stride` starting at `first` — past the end of the destination Variant allocation (heap out-of-bounds write).

The read path (`Variant_copyRange`) already derives strides from the clamped `thisrange`.

## Fix

One-line cherry-pick of the merged `1.4` fix — `computeStrides(v, thisrange, …)` — so an oversized range is rejected with `UA_STATUSCODE_BADINDEXRANGEINVALID` when its clamped element count no longer matches the source array.

Backports the `1.4` regression test `UA_Variant_setRangeRejectsUnclampedSourceSize` (`{3,15}` range over a 10-element `UInt32` array with a 13-element source): unfixed builds write elements 10–12 out of bounds; fixed builds return `BADINDEXRANGEINVALID`.

## Verification

- The change is context-identical to the merged `1.4` commit; `Variant_setRange` and the `check_types_builtin` anchors are unchanged between branches on current `master` (`f223658`, the same head referenced in the issue).
- The test only uses `UA_Variant_setArrayCopy` / `UA_Variant_setRangeCopy`, both present on `master`.

(cherry picked from commit edfe42154b02422237e88b3d6af18b1b1dca90da)


jpfr:
Thanks for this.
We have however already merged the 1.5 branch into master.
So the patch you refer to is already applied.

